PR #661, opened 2018-08-17
Add serviceId to all services

varunnvs92:
* Add serviceId to all services in the repo
* All modules names should be based on service id. Fix the names of modules that don't follow this pattern
* Throw an exception during code generation if serviceId is missing or empty


.changes/next-release/feature-AWSSDKforJavav2-3344b57.json
```diff
@@ -0,0 +1,5 @@
+{
+    "category": "AWS SDK for Java v2", 
+    "type": "feature", 
+    "description": "Refactored all services to make module names match the service id from the c2j model"
```
zoewangg: nit: instead of saying `c2j`, can we just say `service model files`

PR #666, opened 2018-08-21
add codebuild buildspec files 

zoewangg:
add codebuild buildspec files so that we don't need to copy paste the same command again and again..

also, updated device farm integ test to clean up resource

### DIFF
--- a/buildspecs/integ-test.yml
+++ b/buildspecs/integ-test.yml
@@ -1,0 +1,6 @@
+version: 0.2
+
+phases:
+  build:
+    commands:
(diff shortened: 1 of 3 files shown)


buildspecs/integ-test.yml
```diff
@@ -0,0 +1,6 @@
+version: 0.2
+
+phases:
+  build:
+    commands:
+    - mvn clean verify -Dskip.unit.tests -P integration-tests -Dfindbugs.skip -Dcheckstyle.skip -pl !:dynamodbmapper-v1 -Dfailsafe.rerunFailingTestsCount=1
```
dagnir: I think this might require the `-Dmaven.wagon.httpconnectionManager.maxPerRoute=2` option too, since this command will be pulling down dependencies from Maven as well

zoewangg: It's currently working without it, but yeah, I can add it

PR #678, opened 2018-08-29
Fix paginated response iterable hasNextPage

### LINKED ISSUE #677: Dynamodb batchGetItemPaginator not working
Dynamodb batchGetItemPaginator not working. (There might be other pagination operations facing the same issue.)

## Expected Behavior
Dynamodb batchGetItemPaginator should work

## Current Behavior
It throws `DynamoDbException` when there's no next page.

```
software.amazon.awssdk.services.dynamodb.model.DynamoDbException: 1 validation error detected: Value '{}' at 'requestItems' failed to satisfy constraint: 
Member must have length greater than or equal to 1 (Service: DynamoDB, Status Code: 400, )
```
 Looking at the wirelogs, seems it tries to retrieve the next page even if there's no next page.

`batchGetItemResponse.unprocessedKeys()` is empty and is not an instance of `SdkAutoConstructMap`, so 
`BatchGetItemResponseFetcher.hasNextPage` returns true.

```java
    private class BatchGetItemResponseFetcher implements SyncPageFetcher<BatchGetItemResponse> {
        @Override
        public boolean hasNextPage(BatchGetItemResponse previousPage) {
            return previousPage.unprocessedKeys() != null
                    && !SdkAutoConstructList.class.isInstance(previousPage.unprocessedKeys())
                    && !SdkAutoConstructMap.class.isInstance(previousPage.unprocessedKeys());
        }

``` 

```java 
        BatchGetItemIterable iterable = 
            dynamo.batchGetItemPaginator(b -> b.requestItems(attributes));
        iterable.stream().forEach(System.out::println);
```

## Your Environment

* AWS Java SDK version used: 2.0.0-preview-11
* JDK version used: 10.0.1 
* Operating System and version: Mac OS

zoewangg:
## Description
Fix paginated response iterable hasNextPage to check if the output token is non null or non empty if it is a collection or map type. 

`Dynamodb.batchGetItem` returns empty `unprocessedKeys` when there is no unprocessedKeys.
 
`"{"Responses":{"zoewang-TestTable":[{"attribute_foo":{"S":"test"},"hash":{"N":"0"}}]},"UnprocessedKeys":{}}"`

So unprocessedKeys is an empty map and not an instance of `SdkAutoConstructMap`. (we only create the `SdkAutoConstructMap` when it is null)

Currently, the `hasNextPage` checks if the output token is not null AND it is not instance of SdkAutoConstructMap AND not an instance of SdkAutoConstructList. So `hasNextPage` will always return true for empty list or map, which is not expected.

## Motivation and Context
Fixes #677

## Testing
`mvn clean install` local integration test.

Note: I didn't create the integration tests for the scenario where dynamodb returns us unprocessedKeys as it takes large load to trigger that scenario and the test might be flaky. I did test it for once and verified it worked.

## Types of changes

## Checklist

## License

### DIFF
--- a/.changes/next-release/bugfix-AWSSDKforJavav2-532f520.json
+++ b/.changes/next-release/bugfix-AWSSDKforJavav2-532f520.json
@@ -1,0 +1,5 @@
+{
+    "category": "AWS SDK for Java v2", 
+    "type": "bugfix", 
+    "description": "Fix Response Fetcher hasNextPage to check if the output token is non null or non empty if it is a collection or map type. Related to [#677](https://github.com/aws/aws-sdk-java-v2/issues/677)"
+}
--- a/codegen/src/main/java/software/amazon/awssdk/codegen/poet/paginators/PaginatorsClassSpec.java
+++ b/codegen/src/main/java/software/amazon/awssdk/codegen/poet/paginators/PaginatorsClassSpec.java
@@ -34,8 +34,7 @@ import software.amazon.awssdk.codegen.model.service.PaginatorDefinition;
 import software.amazon.awssdk.codegen.poet.ClassSpec;
 import software.amazon.awssdk.codegen.poet.PoetExtensions;
 import software.amazon.awssdk.codegen.poet.model.TypeProvider;
-import software.amazon.awssdk.core.util.SdkAutoConstructList;
-import software.amazon.awssdk.core.util.SdkAutoConstructMap;
+import software.amazon.awssdk.core.util.PaginatorUtils;
 
 public abstract class PaginatorsClassSpec implements ClassSpec {
 
@@ -178,14 +177,11 @@ public abstract class PaginatorsClassSpec implements ClassSpec {
         }
         // If there is no more_results token, then output_token will be a single value
         return CodeBlock.builder()
-                // Auto construct containers are treated as null so terminate
-                // pagination for those cases as well.
-                .add("return $1N.$2L != null && !$3T.class.isInstance($1N.$2L) && !$4T.class.isInstance($1N.$2L)",
-                              PREVIOUS_PAGE_METHOD_ARGUMENT,
-                              fluentGetterMethodsForOutputToken().get(0),
-                              SdkAutoConstructList.class,
-                              SdkAutoConstructMap.class)
-                .build();
+                        .add("return $3T.isOutputTokenAvailable($1N.$2L)",
+                             PREVIOUS_PAGE_METHOD_ARGUMENT,
+                             fluentGetterMethodsForOutputToken().get(0),
+                             PaginatorUtils.class)
+                        .build();
 
     }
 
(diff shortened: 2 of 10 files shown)


core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/util/PaginatorUtils.java
```diff
@@ -0,0 +1,51 @@
...
+
+        if (outputToken instanceof Map) {
+            return !((Map) outputToken).isEmpty();
+        }
+
+        if (outputToken instanceof Collection) {
+            return !((Collection) outputToken).isEmpty();
+        }
+
+        return true;
```
dagnir: nit: you can remove line 37-39 and replace this with `outputToken != null`

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/util/PaginatorUtils.java
```diff
@@ -0,0 +1,51 @@
...
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.core.internal.util;
+
+import java.util.Collection;
+import java.util.Map;
+import software.amazon.awssdk.annotations.SdkInternalApi;
+
+@SdkInternalApi
```
dagnir: Should this be protected? It's imported by the service modules right?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/util/PaginatorUtils.java
varunnvs92: I like the current way instead of nested if statements. (my personal opinion)

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/util/PaginatorUtils.java
zoewangg: yup, good call. I will update it and move it outside of internal package.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/util/PaginatorUtils.java
zoewangg: Hmm, I kind of like the current way. We can reduce 2 line of code with what you suggested.

```java
    public static <T> boolean isOutputTokenAvailable(T outputToken) {
//        if (outputToken == null) {
//            return false;
//        }

        if (outputToken instanceof Map) {
            return !((Map) outputToken).isEmpty();
        }

        if (outputToken instanceof Collection) {
            return !((Collection) outputToken).isEmpty();
        }

        return outputToken != null;
    }
```

But it seems more intuitive to me with the current way where we check null first and don't need to follow the logic all the way down if it's null.

dagnir: Yup, it's fine with me :)

PR #679, opened 2018-08-30
Add SdkAnnotation to service module and other small fixes

zoewangg:
- Add Sdk annotation to classes in service modules and removed the MissingSdkAnnotationCheck suppressions.
   - Most of the classes are marked with `SdkProtectedApi` as they are interceptors and are used in the generated classes. One service I'm not 100% sure is the stepFunction and I marked some of those classes with`SdkPublicApi`. Let me know if that's not correct.

- Renamed `EC2Interceptor` to `Ec2Interceptor`

- Make classes that are not intended to be extended `final`

### DIFF
--- a/services/rds/src/main/java/software/amazon/awssdk/services/rds/RdsPresignInterceptor.java
+++ b/services/rds/src/main/java/software/amazon/awssdk/services/rds/RdsPresignInterceptor.java
@@ -19,6 +19,7 @@ import static software.amazon.awssdk.auth.signer.internal.AwsSignerExecutionAttr
 
 import java.net.URI;
 import java.time.Clock;
+import software.amazon.awssdk.annotations.SdkPublicApi;
 import software.amazon.awssdk.auth.signer.Aws4Signer;
 import software.amazon.awssdk.auth.signer.params.Aws4PresignerParams;
 import software.amazon.awssdk.awscore.endpoint.DefaultServiceEndpointBuilder;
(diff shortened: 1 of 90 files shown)


services/rds/src/main/java/software/amazon/awssdk/services/rds/RdsPresignInterceptor.java
```diff
@@ -43,6 +44,7 @@
  *
  * @param <T> The request type.
  */
+@SdkInternalApi
```
spfink: Wonder if this one should be public to allow customers to easily add their own presigners...

services/s3/src/main/java/software/amazon/awssdk/services/s3/TaggingAdapter.java
```diff
@@ -15,12 +15,14 @@
...
 package software.amazon.awssdk.services.s3;
 
+import software.amazon.awssdk.annotations.SdkProtectedApi;
 import software.amazon.awssdk.core.adapter.TypeAdapter;
 import software.amazon.awssdk.services.s3.model.Tag;
 import software.amazon.awssdk.services.s3.model.Tagging;
 import software.amazon.awssdk.utils.http.SdkHttpUtils;
 
-public class TaggingAdapter implements TypeAdapter<Tagging, String> {
+@SdkProtectedApi
```
spfink: I'd think this should be public.

zoewangg: yup, updated

services/rds/src/main/java/software/amazon/awssdk/services/rds/RdsPresignInterceptor.java
zoewangg: yeah, updated the class to public

PR #688, opened 2018-09-07
Update HTTP Async SPI

dagnir:
## Description

This commit makes changes to the HTTP Async SPI as well as the
AsyncResponseTransformer interface exposed by SDK clients.

For the HTTP SPI, the main changes are to SdkHttpResponseHandler and
SdkAsyncHttpClient.

- SdkHttpResponseHandler

  The most important change is to complete(). It has been removed as it was a
  potential spot where the async client could get blocked or where the
  implementation would be forced to block, since the handler must return the
  transformed response from this method. We also don't necessarily care about
  what the response handler does at the HTTP client layer, so there's no need to
  return the transformed result.

  Secondly, the interface now guarantees that onStream() is always called, even
  for responses with no content. This should simplify implementation a bit since
  implementers know to always expect a stream.

  Other changes are mostly minor, such as naming: the interface is now
  SdkAsyncResponseHandler to be more intuitive since it follows our current
  naming conventions.

- SdkAsyncHttpClient

  prepareRequest() has been replaced with execute() that takes a single
  parameter object. It now also returns a CompletableFuture rather than a
  RunnableRequest.

- AsyncResponseTransformer

  This continues to be almost identical to SdkAsyncHttpResponseHandler.

  The most important change here is in complete(). Like in
  SdkHttpResponseHandler, this method was problematic since it had the
  protential to block the caller. To address this, complete() has been replaced
  with a method named prepare() that returns a CompletableFuture. We guarantee
  that this method is always called before executing the request to allow
  implementers to do any setup/cleanup. This method could be called more than
  once in the event of retries.

  Another important result of prepare() returning a CompletableFuture is this
  gives users the ability to trigger retries by completing it exceptionally with
  a instance of RetryableException similar to the sync ReponseTransformer.

  Finally, this change decouples response transformation from request
  completion. With the current behavior, the complete() is only called when the
  entire response is received. This makes doing things like returning the
  Publisher from onStream() as the result of the transformer awkward.

## Motivation and Context
(See above)

## Types of changes

- [x] New feature **Breaking change: changes public interfaces**

## Checklist

## License

### DIFF
--- a/codegen/src/main/java/software/amazon/awssdk/codegen/poet/client/AsyncClientClass.java
+++ b/codegen/src/main/java/software/amazon/awssdk/codegen/poet/client/AsyncClientClass.java
@@ -179,7 +179,7 @@ public final class AsyncClientClass extends AsyncClientInterface {
         if (opModel.hasStreamingOutput() || opModel.hasEventStreamOutput()) {
             String paramName = opModel.hasStreamingOutput() ? "asyncResponseTransformer" : "asyncResponseHandler";
             builder.addStatement("runAndLogError(log, \"Exception thrown in exceptionOccurred callback, ignoring\",\n" +
-                                 "() -> $L.exceptionOccurred(t))", paramName);
+                                 "() -> $N.exceptionOccurred(t))", paramName);
         }
 
         return builder.addStatement("return $T.failedFuture(t)", CompletableFutureUtils.class)
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/async/EmptyPublisher.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/async/EmptyPublisher.java
@@ -13,22 +13,28 @@
  * permissions and limitations under the License.
  */
 
-package software.amazon.awssdk.http.async;
+package software.amazon.awssdk.core.async;
 
-import java.nio.ByteBuffer;
-import java.util.Optional;
 import org.reactivestreams.Publisher;
+import org.reactivestreams.Subscriber;
+import org.reactivestreams.Subscription;
 import software.amazon.awssdk.annotations.SdkProtectedApi;
 
-/**
- * A {@link Publisher} of HTTP content data that allows streaming operations for asynchronous HTTP clients.
- */
 @SdkProtectedApi
-public interface SdkHttpRequestProvider extends Publisher<ByteBuffer> {
+public class EmptyPublisher<T> implements Publisher<T> {
+    private static final Subscription SUBSCRIPTION = new Subscription() {
+        @Override
+        public void request(long l) {
+        }
 
-    /**
-     * @return The content length of the data being produced.
-     */
-    Optional<Long> contentLength();
+        @Override
+        public void cancel() {
+        }
+    };
 
+    @Override
+    public void subscribe(Subscriber<? super T> subscriber) {
+        subscriber.onSubscribe(SUBSCRIPTION);
+        subscriber.onComplete();
+    }
 }
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/FileAsyncResponseTransformer.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/FileAsyncResponseTransformer.java
@@ -24,6 +24,7 @@ import java.nio.channels.CompletionHandler;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.StandardOpenOption;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.atomic.AtomicLong;
 import org.reactivestreams.Subscriber;
 import org.reactivestreams.Subscription;
@@ -34,13 +35,13 @@ import software.amazon.awssdk.core.async.SdkPublisher;
 /**
  * {@link AsyncResponseTransformer} that writes the data to the specified file.
  *
- * @param <ResponseT> Response POJO type. Returned on {@link #complete()}.
+ * @param <ResponseT> Response POJO type.
  */
 @SdkInternalApi
 public final class FileAsyncResponseTransformer<ResponseT> implements AsyncResponseTransformer<ResponseT, ResponseT> {
-
     private final Path path;
-    private AsynchronousFileChannel fileChannel;
+    private volatile AsynchronousFileChannel fileChannel;
+    private volatile CompletableFuture<Void> cf;
     private volatile ResponseT response;
 
     public FileAsyncResponseTransformer(Path path) {
@@ -52,7 +53,18 @@ public final class FileAsyncResponseTransformer<ResponseT> implements AsyncRespo
     }
 
     @Override
-    public void responseReceived(ResponseT response) {
+    public CompletableFuture<ResponseT> prepare() {
+        cf = new CompletableFuture<>();
+        cf.whenComplete((r, t) -> {
+            if (t != null && fileChannel != null) {
+                invokeSafely(fileChannel::close);
+            }
+        });
+        return cf.thenApply(ignored -> response);
+    }
+
+    @Override
+    public void onResponse(ResponseT response) {
         this.response = response;
     }
 
@@ -60,7 +72,7 @@ public final class FileAsyncResponseTransformer<ResponseT> implements AsyncRespo
     public void onStream(SdkPublisher<ByteBuffer> publisher) {
         // onStream may be called multiple times so reset the file channel every time
         this.fileChannel = invokeSafely(() -> createChannel(path));
-        publisher.subscribe(new FileSubscriber(this.fileChannel, path));
+        publisher.subscribe(new FileSubscriber(this.fileChannel, path, cf));
     }
 
     @Override
@@ -70,14 +82,7 @@ public final class FileAsyncResponseTransformer<ResponseT> implements AsyncRespo
         } finally {
             invokeSafely(() -> Files.delete(path));
         }
-    }
-
-    @Override
-    public ResponseT complete() {
-        if (fileChannel != null) {
-            invokeSafely(fileChannel::close);
-        }
-        return response;
+        cf.completeExceptionally(throwable);
     }
 
     /**
@@ -88,14 +93,16 @@ public final class FileAsyncResponseTransformer<ResponseT> implements AsyncRespo
 
         private final AsynchronousFileChannel fileChannel;
         private final Path path;
+        private final CompletableFuture<Void> future;
 
         private volatile boolean writeInProgress = false;
         private volatile boolean closeOnLastWrite = false;
         private Subscription subscription;
 
-        FileSubscriber(AsynchronousFileChannel fileChannel, Path path) {
+        FileSubscriber(AsynchronousFileChannel fileChannel, Path path, CompletableFuture<Void> future) {
             this.fileChannel = fileChannel;
             this.path = path;
+            this.future = future;
         }
 
         @Override
@@ -135,6 +142,7 @@ public final class FileAsyncResponseTransformer<ResponseT> implements AsyncRespo
                 @Override
                 public void failed(Throwable exc, ByteBuffer attachment) {
                     subscription.cancel();
+                    future.completeExceptionally(exc);
                 }
             });
 
@@ -155,6 +163,7 @@ public final class FileAsyncResponseTransformer<ResponseT> implements AsyncRespo
                     close();
                 }
             }
+            future.complete(null);
         }
 
         private void close() {
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/TransformingAsyncResponseHandler.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/TransformingAsyncResponseHandler.java
@@ -1,0 +1,36 @@
+/*
+ * Copyright 2010-2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.core.internal.http;
+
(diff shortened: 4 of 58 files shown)


core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/TransformingAsyncResponseHandler.java
```diff
@@ -0,0 +1,36 @@
...
+public interface TransformingAsyncResponseHandler<ResultT> extends SdkAsyncHttpResponseHandler {
+    /**
+     * Return the future holding the transformed response.
+     * <p>
+     * This method is guaranteed to be called before the request is executed, and before {@link
+     * SdkAsyncHttpResponseHandler#onHeaders(software.amazon.awssdk.http.SdkHttpResponse)} is signaled.
+     *
+     * @return The future holding the transformed response.
+     */
+    CompletableFuture<ResultT> transformResult();
```
millems: Is there a better name, like `transformationResultFuture()`? This method isn't actually doing the transforming?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/AsyncRetryableStage.java
```diff
@@ -122,65 +112,75 @@ private RetryExecutor(SdkHttpFullRequest request, RequestExecutionContext contex
...
+            }
+
+            if (err instanceof NonRetryableException) {
+                future.completeExceptionally(err);
+                return;
+            }
+
+            // TODO(dongie): We need to wrap into SdkException in order to call setLastRetriedException. Is this necessary?
+            // Can it just take a Throwable?
+            SdkException sdkException = err instanceof SdkException ?
```
millems: TODO?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/MakeAsyncHttpRequestStage.java
```diff
@@ -91,34 +85,57 @@ public MakeAsyncHttpRequestStage(SdkHttpResponseHandler<OutputT> responseHandler
...
+        // Offload the completion of the future returned from this stage onto
+        // the future completion executor
+        transformFuture.whenCompleteAsync((r, t) -> {
+            if (t == null) {
+                responseFuture.complete(r);
+            } else {
+                if (t instanceof CompletionException) {
+                    t = t.getCause();
+                }
+                responseFuture.completeExceptionally(t);
```
millems: This pattern is all over the place. Is there any way we can prevent its spread?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/MakeAsyncHttpRequestStage.java
```diff
@@ -136,119 +153,83 @@ private boolean shouldSetContentLength(SdkHttpFullRequest request, SdkHttpReques
...
                 SdkStandardLogger.REQUEST_LOGGER.debug(() -> "Received error response: " + response.statusCode());
-                errorResponseHandler.headersReceived(response);
+                errorResponseHandler.onHeaders(response);
+            }
+            try {
+                this.response = (SdkHttpFullResponse) response;
+            } catch (ClassCastException e) {
+                // Make SpotBugs happy. Should never happen since SdkHttpFullResponse is the only concrete impl we pass
+                // around.
+                log.error(() -> response + " is not an instance of SdkHttpFullResponse", e);
```
millems: Ew. If it's the only impl we'll pass around, why can't we make it typed correctly?

core/sdk-core/src/test/java/software/amazon/awssdk/core/internal/http/timers/AsyncHttpClientApiCallTimeoutTests.java
```diff
@@ -128,6 +129,8 @@ public void successfulResponse_SlowBeforeRequestRequestHandler_ThrowsApiCallTime
     }
 
     @Test
+    @Ignore
+    // FIXME(dongie): Partly broken by the SPI changes. This test appears to be broken since introduction, but didn't start failing until now: the code path never calls afterTransmission.
```
millems: ?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/TransformingAsyncResponseHandler.java
dagnir: 👍 Like that.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/AsyncRetryableStage.java
dagnir: Not sure what you mean. Do you mean to remove the "(dongie)" part?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/MakeAsyncHttpRequestStage.java
dagnir: I'm not sure, that's the way it was originally. I think we should be able to fix the types though!

core/sdk-core/src/test/java/software/amazon/awssdk/core/internal/http/timers/AsyncHttpClientApiCallTimeoutTests.java
dagnir: So this test class uses `AmazonAsyncHttpClient` and the code path of its `exectue` method never calls the `afterTransmission()` callback; that's something added by `BaseAsyncClientHandler`: https://github.com/aws/aws-sdk-java-v2/blob/c2005a7b765523a9e6a3c083f6a6d022f3f8bde4/core/sdk-core/src/main/java/software/amazon/awssdk/core/client/handler/BaseAsyncClientHandler.java#L174 not sure how it was passing before

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/MakeAsyncHttpRequestStage.java
dagnir: Yep I'll add a util method in `CompletableFutureUtils`

dagnir: BTW we need this check because the type of `t` seems to depend on how the future was completed:

For a future `f`, if it's completed directly using `f.completeExceptionally(new MyException())`, then `t` is type `MyException`.

but if it's some some other future completed higher up, such as

```java
CompletableFuture<String> f0 = new CompletableFuture<>();

CompletableFuture<String> f = f0.thenApply(s ->upperCase(s));

f0.completeExceptionally(new MyException());

f.whenComplete((r, t) -> { .. });
```
then `t` will be type `CompletionException` and cause is `MyException`

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/FutureCancelHandler.java
```diff
@@ -0,0 +1,60 @@
...
+import io.netty.channel.SimpleChannelInboundHandler;
+import io.netty.util.ReferenceCountUtil;
+import java.io.IOException;
+import software.amazon.awssdk.annotations.SdkInternalApi;
+
+/**
+ * Closes the channel if the execution future has been cancelled.
+ */
+@SdkInternalApi
+public class FutureCancelHandler extends SimpleChannelInboundHandler {
```
zoewangg: can we add some unit tests for this class? it might be trivial for now but we might add more logic in this class in the future.

http-clients/netty-nio-client/src/test/java/software/amazon/awssdk/http/nio/netty/NettyNioAsyncHttpClientSpiVerificationTest.java
```diff
@@ -0,0 +1,230 @@
...
+            }
+        }
+
+        @Override
+        public void cancel() {
+            done = true;
+        }
+    }
+
+    private static class DrainingSubscriber<T> implements Subscriber<T> {
```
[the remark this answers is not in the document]
zoewangg: oh nvm, the other one is in core.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/AsyncRetryableStage.java
```diff
@@ -122,65 +112,75 @@ private RetryExecutor(SdkHttpFullRequest request, RequestExecutionContext contex
...
             beforeExecute();
-            doExecute().handle((resp, err) -> handle(future, resp, err));
+            doExecute().whenComplete((resp, err) -> maybeRetry(future, resp, err));
+            return future;
         }
 
-        private Void handle(CompletableFuture<Response<OutputT>> future,
-                            Response<OutputT> resp,
-                            Throwable err) {
+        private void maybeRetry(CompletableFuture<Response<OutputT>> future,
```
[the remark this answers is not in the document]
dagnir: Hmm I don't really like that because it implies we always want to retry, but we only want to do that if it's a retryable exception, and not if it was a successful response.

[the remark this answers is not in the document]
zoewangg: Fair enough, `retryIfNeeded` then? `maybeRetry` feels a bit like we are retrying based on mood to me ╮(╯_╰)╭

core/aws-core/src/main/java/software/amazon/awssdk/awscore/eventstream/EventStreamResponseHandler.java
```diff
@@ -63,7 +63,7 @@
      *
      * @param throwable Exception that occurred.
      */
-    void exceptionOccurred(Throwable throwable);
```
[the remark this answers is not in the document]
dagnir: Ah right I'll change revert. Are you okay with keeping it `onError` on `AsyncResponseTransformer`?

core/sdk-core/src/main/java/software/amazon/awssdk/core/async/EmptyPublisher.java
```diff
@@ -0,0 +1,55 @@
...
+import org.reactivestreams.Publisher;
+import org.reactivestreams.Subscriber;
+import org.reactivestreams.Subscription;
+import software.amazon.awssdk.annotations.SdkProtectedApi;
+
+@SdkProtectedApi
+public class EmptyPublisher<T> implements Publisher<T> {
+    @Override
+    public void subscribe(Subscriber<? super T> subscriber) {
+        subscriber.onSubscribe(new EmptySubscription<>(subscriber));
```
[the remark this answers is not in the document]
dagnir: Nice I'll do that

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/NettyRequestExecutor.java
```diff
@@ -35,91 +40,127 @@
...
+     */
+    private CompletableFuture<Void> createExecuteFuture(Future<Channel> channelFuture) {
+        final CompletableFuture<Void> future = new CompletableFuture<>();
+
+        future.whenComplete((r, t) -> {
+            if (t == null) {
+                return;
             }
+
+            channelFuture.addListener((Future<Channel> f) -> {
```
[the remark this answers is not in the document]
dagnir: Not necessarily but it's true the other way around: if the future is cancelled, then it's completed exceptionally. There are other cases (like in `handleFailure`) where we complete the future exceptionally and want to close the connection so that handles that here as well.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/NettyRequestExecutor.java
```diff
@@ -130,32 +171,22 @@ private void writeRequest(HttpRequest request) {
                        // Auto-read is turned off so trigger an explicit read to give control to HttpStreamsClientHandler
                        channel.read();
                    } else {
+                       // Are there cases where we can keep the channel open?
```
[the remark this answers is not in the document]
dagnir: `handleFailure` completes the future exceptionally which will close the connection (as a result of firing an exception down then channel pipeline)

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/FileAsyncResponseTransformer.java
```diff
@@ -135,6 +142,7 @@ public void completed(Integer result, ByteBuffer attachment) {
                 @Override
                 public void failed(Throwable exc, ByteBuffer attachment) {
                     subscription.cancel();
+                    future.completeExceptionally(exc);
```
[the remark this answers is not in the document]
dagnir: Cancelling which?

codegen/src/main/java/software/amazon/awssdk/codegen/poet/client/AsyncClientClass.java
```diff
@@ -158,8 +158,9 @@ private MethodSpec closeMethod() {
         // For streaming operations we also want to notify the response handler of any exception.
         if (opModel.hasStreamingOutput() || opModel.hasEventStreamOutput()) {
             String paramName = opModel.hasStreamingOutput() ? "asyncResponseTransformer" : "asyncResponseHandler";
+            String errorMethodName = opModel.hasStreamingOutput() ? "onError" : "exceptionOccurred";
```
millems: Any reason not to just keep it consistent, even if it's not ideal?

dagnir: No reason, other than I think `onError` is more consistent with the rest of the methods, but I don't feel strongly about it. I'll go ahead and make them consistent.

dagnir:
@zoewangg Will do.

dagnir:
I ran the integration tests against 7b11a9b7e1bdaf8bc66bcf2326a09b0f22ab15b0 (current head of this PR) and the tests passed. They're currently failing on master so a rebased/merged test will also fail.

PR #695, opened 2018-09-11
Create AutoRenewable STS/SAML credentialsProvider

mscharp:
Added new Sts credential provider class that auto renews a SAML assertion
Created interface that can be implemented to provide functionality to CredentialProvider

## Motivation and Context
When using the StsAssumeRoleWithSamlCredentialProvider, a non expired saml assertion is needed before the AWS credentials can be refreshed.  Currently, the SDK doesn't provide a way to handle renewing the saml assertion before calling the STS service to obtain new AWS credentials.  This pull requests implements a generic solution for this.

It's possible that the new class could be merged into the current StsAssumeRoleWithSamlCredentialProvider, but I wasn't sure if that would be the best way to handle it.

## Testing
Added unit tests for the new class. 

## Types of changes

- [ X] New feature (non-breaking change which adds functionality)

## Checklist

- [ X ] I have read the **CONTRIBUTING** document
- [ X ] My code follows the code style of this project
- [ X ] I have read the **README** document
- [ X ] I have added tests to cover my changes
- [ X ] All new and existing tests passed

## License

- [ X ] I confirm that this pull request can be released under the Apache 2 license

### DIFF
--- a/.changes/next-release/feature-AWSSecurityTokenService-cdd6371.json
+++ b/.changes/next-release/feature-AWSSecurityTokenService-cdd6371.json
@@ -1,0 +1,5 @@
+{
+    "category": "AWS Security Token Service", 
+    "type": "feature", 
+    "description": "Added supplier functionality to StsAssumeRoleWithSamlCredentialProvider.  This allows for the saml assertion to be refreshed before getting new credentials from STS."
+}
--- a/services/sts/src/main/java/software/amazon/awssdk/services/sts/auth/StsAssumeRoleWithSamlCredentialsProvider.java
+++ b/services/sts/src/main/java/software/amazon/awssdk/services/sts/auth/StsAssumeRoleWithSamlCredentialsProvider.java
@@ -16,6 +16,7 @@
 package software.amazon.awssdk.services.sts.auth;
 
 import java.util.function.Consumer;
+import java.util.function.Supplier;
 import software.amazon.awssdk.annotations.NotThreadSafe;
 import software.amazon.awssdk.annotations.SdkPublicApi;
 import software.amazon.awssdk.annotations.ThreadSafe;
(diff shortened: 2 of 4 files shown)


millems:
+1 to merging this with the existing credentials provider instead of creating a new one. Instead of reactively updating the assertion, can we proactively update it when we create the request?

We've added something similar to the assume-role credentials provider that allows specifying a request OR a request provider (for use with MFA): https://github.com/aws/aws-sdk-java-v2/blob/master/services/sts/src/main/java/software/amazon/awssdk/services/sts/auth/StsAssumeRoleCredentialsProvider.java#L99-L109

If a similar pattern would work here, that seems ideal.

mscharp:
I can merge the changes into the existing provider easily enough.  I'll have an update shortly.

Could you elaborate more on how the request provider would work?

I initially created a solution that would keep track of the expiry time and proactively reauthenticate with the saml, but I didn't like the solution.  It required creating a timer and trying to time the saml reauthentication with the aws reauthentication, and over time got out of sync.  This may not have been the best way to do it, but it was what I had come up with.  

As such, I went with this approach as it shouldn't impact performance or usability, and has the added benefit of being quite simple.

mscharp:
Ok.  I took a deeper look and now understand how the Supplier works. (I haven't used that pattern before).  I think that will work great.  It's similar to what I already implemented, but executes a little sooner (before the ETE is thrown) and doesn't require a timer or custom interface.  Much cleaner!

I should have it updated soon.

millems:
Awesome, it'd be great to have that functionality added to the default SAML provider.

mscharp:
Code has been updated and rebased against master.  Just waiting for review now.

.changes/next-release/feature-AWSSecureTokenService-6de82b5.json
```diff
@@ -0,0 +1,5 @@
+{
+    "category": "AWS SecureTokenService",
```
dagnir: minor: This should be 'AWS Security Token Service'

mscharp:
I've rebased my feature branch with master and squashed the commits.

mscharp:
Bumping to hopefully get it merged in soon...

.changes/next-release/feature-AWSSecureTokenService-6de82b5.json
```diff
@@ -0,0 +1,5 @@
+{
+    "category": "AWS Secure Token Service",
```
dagnir: nit: this should be "Security"

mscharp:
@dagnir 

Rebased and fixed.

dagnir:
@mscharp Thanks for your contribution!

PR #700, opened 2018-09-13
Add support for Aws4UnsignedPayloadSigner

varunnvs92:
**Changes**
* Add support for request signers set through "authtype" trait in c2j model
  * Added a new signer param in [RequestOverrideConfiguration](https://github.com/aws/aws-sdk-java-v2/blob/2e0a6700a3fc2f57841aee4c3ba452a0d22f9cba/core/sdk-core/src/main/java/software/amazon/awssdk/core/RequestOverrideConfiguration.java#L45) class
  * I implemented it this way as it gives option for SDK to set signer based on `authtype` trait and also gives customers ability to override signers at request level
  * In v1, we used [SignerTypeAware](https://github.com/aws/aws-sdk-java/blob/2b447a8287ff04cfddd1fedcf358c9f9b2e9fced/aws-java-sdk-core/src/main/java/com/amazonaws/auth/SignerTypeAware.java) interface on requests that have the 
 authtype trait
* Added Aws4UnsignedPayloadSigner similar to [v1](https://github.com/aws/aws-sdk-java/blob/2b447a8287ff04cfddd1fedcf358c9f9b2e9fced/aws-java-sdk-core/src/main/java/com/amazonaws/auth/AWS4UnsignedPayloadSigner.java)
* Fixed a bug in codegen that fails if operation has both streaming input and streaming output
* Renamed lex -> lexruntime based on serviceId naming convention

**Tests**
* Updated Codegen test models
* All existing tests pass

**Pending**
* ~~Make a integration test call to test PostContent API~~ - Done (Tested via integration test in my local machine)


codegen/src/main/java/software/amazon/awssdk/codegen/IntermediateModelBuilder.java
```diff
@@ -193,21 +189,50 @@ private void linkCustomAuthorizationToRequestShapes(IntermediateModel model) {
...
+            shape.setRequestSignerClassFqcn(model.getMetadata().getAuthPolicyPackageName() + '.' +
+                                            auth.getInterfaceName());
+        } else if (AuthType.IAM.equals(c2jOperation.getAuthType())) {
+            model.getMetadata().setRequiresIamSigners(true);
+            // TODO IamRequestSigner does not exist
+            shape.setRequestSignerClassFqcn("software.amazon.awssdk.opensdk.protect.auth.IamRequestSigner");
+        }
+    }
+
+    private void linkCustomAuthorizationToRequestShapeForAwsProtocol(AuthType authType, ShapeModel shape) {
```
shorea: I wouldn't call this "custom" so much as operation specific.

core/auth/src/main/java/software/amazon/awssdk/auth/signer/Aws4UnsignedPayloadSigner.java
```diff
@@ -0,0 +1,68 @@
...
+import software.amazon.awssdk.core.interceptor.ExecutionAttributes;
+import software.amazon.awssdk.http.SdkHttpFullRequest;
+
+/**
+ * Exactly the same as {@link Aws4Signer} except if the request is being sent
+ * over HTTPS, then it returns the string <code>UNSIGNED-PAYLOAD</code> as the
+ * content SHA-256 so services that support it can avoid needing to calculate
+ * the value when authorizing the request.
+ * <p>
+ * Payloads are still signed for requests over HTTP to preserve the request
```
shorea: Where does this happen?

core/auth/src/main/java/software/amazon/awssdk/auth/signer/Aws4UnsignedPayloadSigner.java
```diff
@@ -0,0 +1,68 @@
...
+ * integrity over a non-secure transport.
+ */
+@SdkPublicApi
+public final class Aws4UnsignedPayloadSigner extends Aws4Signer {
+
+    private Aws4UnsignedPayloadSigner() {
+        super();
+    }
+
+    public static Aws4UnsignedPayloadSigner create() {
```
zoewangg: hmmm, I think we are gonna have the similar issue as before where it has two static `create()` methods: one from here and the other inherited from the parent. Does it have to extend `Aws4Signer`?

codegen/src/main/java/software/amazon/awssdk/codegen/IntermediateModelBuilder.java
varunnvs92: Used the same names from v1. Removed "Custom" from method name. Btw these are private methods, so not worried about naming as long as they are understandable.

core/auth/src/main/java/software/amazon/awssdk/auth/signer/Aws4UnsignedPayloadSigner.java
varunnvs92: For http requests, [this](https://github.com/aws/aws-sdk-java-v2/blob/e8aee9db60fefad5b13ae16ae5540820f7037e72/core/auth/src/main/java/software/amazon/awssdk/auth/signer/internal/AbstractAws4Signer.java#L146-L159) super class method is invoked

core/auth/src/main/java/software/amazon/awssdk/auth/signer/Aws4UnsignedPayloadSigner.java
varunnvs92: Intellij is only showing one create() method. Can you point me to the old issue you are referring to?

aws-sdk-java/pom.xml
```diff
@@ -490,6 +490,11 @@
             <groupId>software.amazon.awssdk</groupId>
             <version>${awsjavasdk.version}</version>
         </dependency>
+        <dependency>
+            <artifactId>lexruntime</artifactId>
```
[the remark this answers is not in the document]
varunnvs92: Good catch. Added a few missing. We need to review all aggregate module poms before GA to ensure all service modules are included.

core/auth/src/main/java/software/amazon/awssdk/auth/signer/Aws4UnsignedPayloadSigner.java
zoewangg: oh, both of them are no-arg, so we are good for now, but if we add a static create method with argument in the parent such as `create(String arg)`, this class will inherit it.

zoewangg: I'd say we should avoid inheritance if not necessary.

zoewangg: Right, customers won't expect polymorphic. let's say we added `Aws4Signer.create(String arg)` in the parent and when a customer trying to create a `Aws4UnsignedPayloadSigner`, there are going to be two static `create` methods popping up, 
- `Aws4UnsignedPayloadSigner.create()` (create `Aws4UnsignedPayloadSigner`)  
- `Aws4UnsignedPayloadSigner.create(String arg)` (create `Aws4Signer`). 

It's a bit confusing to customers which one to use.

I think static members are hidden in a sense that they are not overridable by child classes. If the static methods in both classes have the same signature then it will be hidden in the child class, on the other hand, if they have different signatures, the child class will "inherit" that method.

> To avoid inheritance, we need to use composition and override all public, protected methods in Aws4Signer.


Why do we need to override all the public methods in `Aws4Signer`? can we just create a `Aws4Signer` field in `Aws4UnsignedPayloadSigner`, let's say `delegate`, and call `delegate.sign()`?

varunnvs92: > Let's say we added Aws4Signer.create(String arg)

In this case, we would need to add `create(String arg)` in Aws4UnsignedPayloadSigner too as we would need to construct the base class object

> Why do we need to override all the public methods in Aws4Signer? 

Yep, we would delegate all the calls. I meant we need to define [methods like these](https://github.com/aws/aws-sdk-java-v2/blob/master/core/auth/src/main/java/software/amazon/awssdk/auth/signer/Aws4Signer.java#L85-L104) in Aws4UnsignedPayloadSigner and call delegate in the body. 
In future, we add `protected String calculateSignature(foo)` method to `Aws4Signer`, then we need to add the same method to `Aws4UnsignedPayloadSigner` too and call `delegate.calculateSignature(foo)`

Basically we have similar issues in both approaches. I see scenario 2 (adding method like calculateSignature) likely to happen more than scenario 1 (adding overloaded create method).

zoewangg: Gotcha, I still think we should try to avoid inheritance as it can introduce more issues and is more fragile. 

- Even though we can kind of solve the awkward two static `create` issue by creating the same `create(String arg)` in this class,  but it's going be useless if this class doesn't need that argument for instantiation at all. -> I'd be fine if we know they are always going to have the same constructors.

- Inheritance by nature violates encapsulation. If we add more protected and public methods in `Aws4Signer` in the future, `Aws4UnsignedPayloadSigner` will inherit those methods no matter if they are needed in that class or not. -> I'd be fine if it's guaranteed that all inherited methods in `Aws4UnsignedPayloadSigner` make sense.

- While `Aws4Signer` is not designed to be extended, it can be extended by customers and that could cause unwanted results. I recall we have seen this kind of issue in the past. (We can probably mitigate it a little by making some of the methods `final`).

I think calling delegate and adding more methods in `Aws4UnsignedPayloadSigner` gives us more flexibility and less error prune with the little price of more code.

varunnvs92: > Inheritance by nature violates encapsulation. 

I agree with it but it makes sense in here as essentially `Aws4Signer` and `Aws4UnsignedPayloadSigner` are exactly same except that latter doesn't sign payload using `https`.

I am not fully convinced we should go with delegate approach here. Let's see what others think about this.

varunnvs92: Added a common base class for Aws4Signer and Aws4UnsignedPayloadSigner. Using composition requires more refactoring to signers as a `calculateContentHash` is used in [AbstractAws4Signer.dosign](https://github.com/aws/aws-sdk-java-v2/blob/810f0e14e078aa60b295a229b5109eed27a026c8/core/auth/src/main/java/software/amazon/awssdk/auth/signer/internal/AbstractAws4Signer.java#L74) method and just overriding calculateContentHash method in the impl classes won't work.

core/auth/src/main/java/software/amazon/awssdk/auth/signer/internal/BaseAws4Signer.java
```diff
@@ -0,0 +1,96 @@
...
+import software.amazon.awssdk.auth.signer.params.Aws4PresignerParams;
+import software.amazon.awssdk.auth.signer.params.Aws4SignerParams;
+import software.amazon.awssdk.core.interceptor.ExecutionAttributes;
+import software.amazon.awssdk.http.SdkHttpFullRequest;
+
+/**
+ * Abstract base class for concrete implementations of Aws4 signers.
+ */
+@SdkInternalApi
+public abstract class BaseAws4Signer extends AbstractAws4Signer<Aws4SignerParams, Aws4PresignerParams> {
```
shorea: Can this be package private?

varunnvs92: This is in internal package (different than the impl classes)

Issue #650, opened 2018-08-07
secretsmanager is missing in the BOM


saiya:
Please add `secretsmanager` into BOM (https://github.com/aws/aws-sdk-java-v2/blob/master/bom/pom.xml).

## Expected Behavior

BOM has `<dependencyManagement>` definition for `software.amazon.awssdk:secretsmanager`

## Current Behavior

BOM does not have it, thus user cannot rely on BOM.

## Possible Solution

Please add following into `<dependencyManagement>` of https://github.com/aws/aws-sdk-java-v2/blob/master/bom/pom.xml 

```
            <dependency>
                <artifactId>secretsmanager</artifactId>
                <groupId>software.amazon.awssdk</groupId>
                <version>${awsjavasdk.version}</version>
            </dependency>
```

## Steps to Reproduce (for bugs)

1. Add dependency to BOM
2. Add dependency to `secretsmanager`, without explicit version
3. Failed to resolve dependency because there is no version specification

## Context

When use BOM in maven/gradle.

## Your Environment

* AWS Java SDK version used: 2.0.0-preview-11
* JDK version used: Any (I used 8, 10, 11)
* Operating System and version: Any (I used Mac OS, Linux)

dagnir:
Fixed by #707.

Issue #610, opened 2018-07-10
Organizations Service is missing in BOM


JohannesRudolph:
## Expected Behavior
https://search.maven.org/remotecontent?filepath=software/amazon/awssdk/bom/2.0.0-preview-10/bom-2.0.0-preview-10.pom should contain the organizations service.

## Current Behavior
Gradle imports like this fail to resolve: 
```
dependencyManagement {
  imports {
    mavenBom 'software.amazon.awssdk:bom:2.0.0-preview-10'
  }
}

dependencies {
  compile 'software.amazon.awssdk:organizations'
}
``` 

## Possible Solution
Add organizations service to the BOM. The artifact is already published to maven central individually, it's just missing from the BOM. https://search.maven.org/#artifactdetails%7Csoftware.amazon.awssdk%7Corganizations%7C2.0.0-preview-10%7Cjar

The current workaroudn is to import the organizations service directly and not via BOM like this (gradle): 
```
compile 'software.amazon.awssdk:organizations:2.0.0-preview-10'
```

dagnir:
Thanks for reporting! Looks like it needs to be added here: https://github.com/aws/aws-sdk-java-v2/blob/1fcbaaf1b1c4739856ed285f0c75bed318a10ad5/aws-sdk-java/pom.xml#L37

saiya:
Also please add `secretsmanager` into BOM.

I filed separate issue https://github.com/aws/aws-sdk-java-v2/issues/650 just in case.

dagnir:
Fixed by #707.

PR #707, opened 2018-09-14
Add mediastoredata module

varunnvs92:


dagnir:
Sorry to piggyback but could you include `secretsmanager` and `organizations` in the BOM changes as well? I'd add a separate PR but I don't want to have to wait for integration to pass another time.

#650, #610

PR #713, opened 2018-09-18
Apache configuration cleanup

varunnvs92:
* Current Apache [configuration options](https://github.com/aws/aws-sdk-java-v2/blob/32ea7e74844975dcbb38e99967fa2a0e7b6d9db5/http-clients/apache-client/src/main/java/software/amazon/awssdk/http/apache/ApacheHttpClient.java#L263-L307) and [default values](https://github.com/aws/aws-sdk-java-v2/blob/32ea7e74844975dcbb38e99967fa2a0e7b6d9db5/http-clients/apache-client/src/main/java/software/amazon/awssdk/http/apache/internal/DefaultConfiguration.java) seems fine (matches v1)
* Includes changes to support nonProxyHosts proxy configuration ([v1 reference](https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-core/src/main/java/com/amazonaws/http/apache/SdkProxyRoutePlanner.java))

### DIFF
--- a/core/sdk-core/src/test/java/software/amazon/awssdk/core/internal/http/timers/AsyncHttpClientApiCallTimeoutTests.java
+++ b/core/sdk-core/src/test/java/software/amazon/awssdk/core/internal/http/timers/AsyncHttpClientApiCallTimeoutTests.java
@@ -147,7 +147,7 @@ public class AsyncHttpClientApiCallTimeoutTests {
     }
 
     @Test
-    public void  slowApiAttempt_ThrowsApiCallAttemptTimeoutException() {
+    public void slowApiAttempt_ThrowsApiCallAttemptTimeoutException() {
         httpClient = testAsyncClientBuilder()
             .apiCallTimeout(API_CALL_TIMEOUT)
             .apiCallAttemptTimeout(Duration.ofMillis(100))
--- a/http-clients/apache-client/pom.xml
+++ b/http-clients/apache-client/pom.xml
@@ -68,6 +68,11 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>org.assertj</groupId>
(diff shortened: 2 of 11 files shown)


http-clients/apache-client/src/main/java/software/amazon/awssdk/http/apache/internal/ProxySystemSetting.java
```diff
@@ -0,0 +1,50 @@
...
+import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.utils.SystemSetting;
+
+/**
+ * The system properties related to http proxy
+ */
+@SdkInternalApi
+public enum ProxySystemSetting implements SystemSetting  {
+
+    NON_PROXY_HOSTS("http.nonProxyHosts")
```
millems: Are there any other system properties we should support?

millems:
Are our sync configurations consistent with async, or can they be updated to be more consistent?

varunnvs92:
I don't see a lot of similarity between sync and async. The only options that match are `connectionTimeout`, `maxConnections` (renamed to `maxConcurrency` in async). I don't see any other related options.

millems:
What about acquire timeout in async? Can that be controlled separately in sync?

varunnvs92:
I see there is a "http.connection-manager.timeout" option in Apache but it only applicable for a specific implementation of apache client. It might not be applicable if we change the type of underlying apache client.

millems:
Do we expect to ever change the underlying implementation? It would be nice to expose the equivalent features if we can.

varunnvs92:
Makes sense. Added the option. Ah I really hate that Github doesn't allow replying to general comment directly.

millems:
Nice! Are there any other async options we can also support, like max pending acquires, etc.?

PR #724, opened 2018-09-24
Add timeout feature for synchronous api calls

zoewangg:
## Description
Add timeout feature for synchronous api calls to allow customers to configure how long the api call or api call attempt should timeout if it didn't complete within the configured time. 

~~It works the same way as v1.~~ Updated: the apiCallTimeout works the same way as v1, `ApiCallAttemptTimeout`, on the other hand, behaves differently compared with v1. It will not only abort the request but also interrupt the thread just like apiCallTimeout.

- `ApiCallTimeout`: a timeout task is scheduled to a separate thread pool and it will interrupt and the current thread **and** abort the underlying http request when the total request time exceeds the configured timeout.
- `ApiCallAttemptTimeout`: a timeout task is scheduled to a separate thread pool and it will interrupt the thread and abort the http request when times up.

~~**NOTE**:
`ApiCallAttemptTimeout` will not abort the request if the request is stuck in the `ResponseTransformer` because the timer task is being cancelled in `MakeHttpRequestStage` and `ResponseTransformer` is triggered after that. This is different from asynchronous timeout where the timer task is being cancelled upon the completion of the completable future.~~

## Testing
Integration tests for timeouts are passed.
All integration tests are passed.

## Types of changes

## Checklist

## License


core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/MakeHttpRequestStage.java
```diff
@@ -53,8 +63,20 @@ public MakeHttpRequestStage(HttpClientDependencies dependencies) {
...
-                .prepareRequest(request, SdkRequestContext.builder().build());
+            .prepareRequest(request, SdkRequestContext.builder().build());
+
+        long timeoutInMillis = resolveTimeoutInMillis(context.requestConfig()::apiCallAttemptTimeout, apiCallAttemptTimeout);
+
+        TimeoutTracker timeoutTracker = timeSyncTaskIfNeeded(timeoutExecutor, timeoutInMillis, false);
+        timeoutTracker.abortable(requestCallable);
+
+        context.apiCallAttemptTimeoutTracker(timeoutTracker);
+        context.apiCallTimeoutTracker().abortable(requestCallable);
```
dagnir: Is this redundant with line 71?

test/protocol-tests/src/test/java/software/amazon/awssdk/protocol/tests/timeout/BaseTimeoutTest.java
```diff
@@ -0,0 +1,131 @@
...
+    public static class SlowCustomResponseTransformer implements ResponseTransformer {
+
+        @Override
+        public Object transform(Object response, AbortableInputStream inputStream) throws Exception {
+            wastingTimeInterrupitably();
+            return null;
+        }
+    }
+
+    public static void wastingTimeInterrupitably() throws InterruptedException {
```
dagnir: nit: "interruptibly" mispelled

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/timers/TimerUtils.java
```diff
@@ -59,4 +63,33 @@ private TimerUtils() {
 
         return timeoutTracker;
     }
+
+    /**
+     * Schedule a {@link TimeoutTask} that aborts the task if not otherwise completed before the given timeout.
+     *
+     * @param timeoutExecutor the executor to execute the {@link TimeoutTask}
+     * @return a {@link TimeoutTracker}
```
dagnir: Please add docs for the other params as well. It's hard to tell what `isInterruptedThread` is without looking at the source. I'd also suggest changing it to `interruptCurrentThread`

core/sdk-core/src/main/java/software/amazon/awssdk/core/sync/ResponseTransformer.java
```diff
@@ -59,12 +61,12 @@
  * <p>
  * Implementations should have proper handling of Thread interrupts. For long running, non-interruptible tasks, it is recommended
  * to check the thread interrupt status periodically and throw an {@link InterruptedException} if set. When an {@link
- * InterruptedException} is thrown from a interruptible task, you should either re-interrupt the current thread or throw that
- * {@link InterruptedException} from the {@link #apply(Object, AbortableInputStream)} method. Failure to do these things may
- * prevent the SDK from stopping the request in a timely manner in the event the thread is interrupted externally.
+ * InterruptedException} is thrown from a interruptible task, you should throw that {@link InterruptedException} from the
```
dagnir: Did we remove the ability for customers to just interrupt the thread without (re)throwing an `InterruptedException`?

test/protocol-tests/src/test/java/software/amazon/awssdk/protocol/tests/timeout/BaseTimeoutTest.java
```diff
@@ -0,0 +1,131 @@
...
+
+    protected void verifyTimedOut() {
+        timeoutExceptionAssertion().accept(() -> callable().call());
+    }
+
+    protected void verifyRetryableTimeout() {
+        timeoutExceptionAssertion().accept(() -> retryableCallable().call());
+    }
+
+    public static class SlowBytesResponseTransformer<ResponseT> implements ResponseTransformer<ResponseT, ResponseBytes<ResponseT>> {
```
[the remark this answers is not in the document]
zoewangg: Yeah, those tests are created to make sure we have `InterruptMonitor.checkInterrupted();` in each of the response transformer types to cover the scenarios where the request time exceeds the configured time right after entering our defined response transformer.

https://github.com/aws/aws-sdk-java-v2/pull/724/files#diff-66508e304899c9905bc9d659b856b50dR119

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/MakeHttpRequestStage.java
zoewangg: This is intended. line 71 is setting the abortable for `ApiCallAttemptTimeoutTracker` whereas line 74 is setting the abortable for `ApiCallTimeoutTracker`

core/sdk-core/src/main/java/software/amazon/awssdk/core/sync/ResponseTransformer.java
dagnir: Not sure what you mean? Can we add a check for `Thread.isInterrupted()` after we call `transform`? 

To me it's perfectly valid for customers to do this in their transform methods:

```java
MyObject transform() {
    try {
        // something interruptible
    } catch (InterruptedException e) {
        log.error("Transform was interrupted", e);
        Thread.interrupt();
        return null;
    }
}
```

We have an explicit test for this behavior in the S3 module (though it's currently disabled): https://github.com/aws/aws-sdk-java-v2/blob/2.0.0-preview-12/services/s3/src/it/java/software/amazon/awssdk/services/s3/GetObjectFaultIntegrationTest.java#L108-L129

zoewangg: Right, we can add the check after we call `transform`, but the time when the request actually gets timed out might be far from accurate. 

For example, say we set the request timeout to `1s` and if the `transform` method takes more than `5s`, the request only will throw `TimeoutException` after 5s when it goes to `InterruptMonitor.checkInterrupted()`, or even worse, if the request gets stuck in `transform`, TimeoutException will never get thrown.

Yeah, I can see it's valid for customers to just re-interrupt the thread, but my point is it's highly possible that they get TimeoutException long after the configured timeout.

dagnir: But we already state in the doc that customers should throw an `InterruptedException` if their transform method gets interrupted, so I guess I don't see how that's too different (as far as timing accuracy) if instead of re-throwing the exception, they set the flag and return right away and on the SDK side we have a check for the flag as soon as `transform()` returns.

zoewangg: Yeah, we should be fine for most of the cases. I was thinking of one edge case where the customers just set the interrupted flag without re-throwing any exception and do some time-consuming task after it.

I will update the docs.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/MakeHttpRequestStage.java
dagnir: Ah okay I see it

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/MakeHttpRequestStage.java
```diff
@@ -53,8 +63,20 @@ public MakeHttpRequestStage(HttpClientDependencies dependencies) {
...
 
-        return requestCallable.call();
+        try {
+            return requestCallable.call();
```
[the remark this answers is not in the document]
zoewangg: We don't currently do this in V2, but I guess we can add a `isBufferResponse` in `ExecuteRequest` and pass it to the http clients. This seems a bit out of scope of this PR and I'd say creating another PR for this feature.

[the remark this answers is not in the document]
shorea: One of the goals of the new timers is to make the per attempt and per call timeouts behave more similiarly. So the per attempt timeout would not just be an abort on the HTTP request but a timer that interrupts as well and is scoped to the entirety of the things we do per request (i.e. unmarshalling for example). This would elimnate the need for buffering as we consume the data in unmarshalling which is subject to the timeout. Is this how the timeouts work (haven't looked at the PR)?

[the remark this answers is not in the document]
zoewangg: No, the current timeout implementation is the same as v1, so attempt timeout is just aborting the request. I like the idea of scoping it to the entire thing and will update the PR to see how it looks. One concern I have though is now that both call timemout and attempt timeout can interrupt the thread, handling interrupt exception can get tricky and we might get more issues with uncleared interrupt status.

core/sdk-core/src/main/java/software/amazon/awssdk/core/sync/ResponseTransformer.java
```diff
@@ -81,10 +83,14 @@
      */
     default ReturnT apply(ResponseT response, AbortableInputStream inputStream) throws Exception {
         try {
-            return transform(response, inputStream);
-        } catch (RetryableException e) {
+            InterruptMonitor.checkInterrupted();
```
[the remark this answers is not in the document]
zoewangg: Yeah, same as exception handling logic here. I think we should probably move the whole thing to the `SyncClientHandler`

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/timers/SyncTimeoutTask.java
```diff
@@ -0,0 +1,65 @@
...
+    @Override
+    public void abortable(Abortable abortable) {
+        this.abortable = abortable;
+    }
+
+    @Override
+    public void run() {
+        log.debug(() -> "Timing out, aborting the task");
+        hasExecuted = true;
+        if (!threadToInterrupt.isInterrupted()) {
```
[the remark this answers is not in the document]
zoewangg: Nope, will remove

Issue #607, opened 2018-07-06
Runtime HTTP client resolution in OSGi


dagnir:
Today, HTTP client resolution happens at runtime using Java's [ServiceLoader](https://docs.oracle.com/javase/7/docs/api/java/util/ServiceLoader.html) class. This works fine in most cases, but not in OSGi: https://blog.osgi.org/2013/02/javautilserviceloader-in-osgi.html

We should create an OSGi-aware implementation of [SdkHttpServiceProvider](https://github.com/aws/aws-sdk-java-v2/blob/4413bfda7b77ced3ce9dc43d3cf5f8ddca2495e2/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/loader/SdkHttpServiceProvider.java), and make current and future implementations of the HTTP client SPI loadable in an OSGi environment.

dagnir:
Looks like the Netty Reactive Streams packages aren't available as OSGi bundles: https://github.com/playframework/netty-reactive-streams/issues/5

sarod:
This is still broken in latest SDK version 2.5.17.

The fix made in commit 2614631 has apparently been removed as part of b5acc52ea196eb978d12026af2968c5b65aa6f80.

@dagnir was this forgotten or is it on purpose?

dagnir:
@sarod Thanks for pinging this. I'm going to go ahead and reopen this as the fix made to Netty as part of #1391 should now unblock the Netty client.

We're still tracking OSGi compatibility for the SDK as a whole in #59.

PR #732, opened 2018-09-25
Require HTTP service cap from core

dagnir:
## Description
The core bundle does the actual loading through ServiceLoader so it should
declare the requirement on the capability.

## Motivation and Context
#607.

## Testing
Create a bundle that uses Kinesis and attempt to make a service call.

## Types of changes

## Checklist

## License

### DIFF
--- a/core/sdk-core/pom.xml
+++ b/core/sdk-core/pom.xml
@@ -211,6 +211,24 @@
                     </properties>
                 </configuration>
             </plugin>
+            <plugin>
+                <groupId>org.apache.felix</groupId>
+                <artifactId>maven-bundle-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>bundle-manifest</id>
+                        <phase>process-classes</phase>
+                        <goals>
+                            <goal>manifest</goal>
+                        </goals>
+                    </execution>
+                </executions>
+                <configuration>
+                    <instructions>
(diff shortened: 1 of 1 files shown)


core/sdk-core/pom.xml
```diff
@@ -211,6 +211,24 @@
...
+                        <id>bundle-manifest</id>
+                        <phase>process-classes</phase>
+                        <goals>
+                            <goal>manifest</goal>
+                        </goals>
+                    </execution>
+                </executions>
+                <configuration>
+                    <instructions>
+                        <Require-Capability>osgi.serviceloader; filter:="(|(osgi.serviceloader=software.amazon.awssdk.http.SdkHttpService)(osgi.serviceloader=software.amazon.awssdk.http.SdkAsyncHttpService))";cardinality:=multiple,osgi.extender; filter:="(osgi.extender=osgi.serviceloader.processor)"</Require-Capability>
```
varunnvs92: Do we need the "|" symbol before first "osgi.serviceloader"?

We can also just put ` osgi.extender; filter:="(osgi.extender=osgi.serviceloader.processor)"` which will allow sdk-core to load any service instead of limiting to SdkHttpService and SdkAsyncHttpService

dagnir: The `|` is a logical OR so customers don't need to load both types of HTTP clients if they only intend to use one of them.

> We can also just put osgi.extender; filter:="(osgi.extender=osgi.serviceloader.processor)" which will allow sdk-core to load any service instead of limiting to SdkHttpService and SdkAsyncHttpService

What's the benefit to this? This is the only use of `ServiceLoader` in `core` afaik

varunnvs92: Yep no additional advantage now. Might be useful in case if there is an additional service interface and we don't need to remember to update the pom. I am okay with current change too.

PR #733, opened 2018-09-26
Move cross-module dependencies to protected

### LINKED ISSUE #719: Multiple packages import .internal subpackage from dependencies
## Current Behavior
The following modules import `*.internal.*` packages from the dependencies:
 - `auth`
    
    https://github.com/aws/aws-sdk-java-v2/blob/5f8d7182150fc4256102f3dad4def0a8a24cc63e/core/auth/src/main/java/software/amazon/awssdk/auth/credentials/ContainerCredentialsProvider.java#L35-L36

 - `aws-core`

    https://github.com/aws/aws-sdk-java-v2/blob/5f8d7182150fc4256102f3dad4def0a8a24cc63e/core/aws-core/src/main/java/software/amazon/awssdk/awscore/client/handler/AwsSyncClientHandler.java#L30

 - `codegen`, and by extension, services modules
    
    https://github.com/aws/aws-sdk-java-v2/blob/5f8d7182150fc4256102f3dad4def0a8a24cc63e/codegen/src/test/resources/software/amazon/awssdk/codegen/poet/client/test-async-client-class.java#L22-L23

## Possible Solution
Remove the dependencies, or move the referenced classes outside of `.internal`.

`grep -r '^import .*\.internal' .`

## Your Environment

* AWS Java SDK version used: 2.0.0-preview-12
* JDK version used:

dagnir:
## Description
Move `.internal` classes being imported by other modules to non-internal package. SDK annotations changed accordingly.

## Motivation and Context
Fixes #719

## Testing
- `find ./ -name 'MANIFEST.MF' -type f -exec grep -n '\.internal\.' {} /dev/null \;`

- Load a service client in an OSGi container.

   By default, the BND tool does not export any packages containing '.internal', so if any modules were still importing them, at least one of the bundles in the dependency graph would have failed to load.

## Types of changes

## Checklist

## License


varunnvs92:
What do you think about moving to a package like "protected" so that we can do targeted exports when we add module-info.java

millems:
The problem with targeted exports is that we'd need our runtime to reference each SDK client, which means we need to update runtime every time we add a new service.

PR #758, opened 2018-10-15
Rest Xml marshaller/unmarshaller refactor

varunnvs92:
* Refactored xml marshaller/unmarshaller similar to json/query
* Addressed most of the old comments
* Perf tests results are similar to json (a bit slower than v1). But it is ~10 micro seconds depending on the file size which is minute compared to other latency like network calls

**TODO**
* Will run integ tests


core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/protocol/restxml/QueryParamMarshaller.java
```diff
@@ -0,0 +1,102 @@
...
+            return;
+        }
+
+        final MapTrait mapTrait = sdkField
+            .getOptionalTrait(MapTrait.class)
+            .orElseThrow(() -> new IllegalStateException("SdkField of list type is missing List trait"));
+        final SdkField valueField = mapTrait.valueFieldInfo();
+
+        for (Map.Entry<?, ?> mapEntry : ((Map<?, ?>) map).entrySet()) {
+            SimpleQueryParamMarshaller keyMarshaller = (SimpleQueryParamMarshaller)
```
shorea: I've actually made changes to assume that it's always a Map<String, ?> to avoid some casting. Do we think it's possible to support non string keyed maps in the future?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/protocol/restxml/QueryParamMarshaller.java
```diff
@@ -0,0 +1,102 @@
...
+                context.marshallerRegistry().getMarshaller(MarshallLocation.QUERY_PARAM, mapEntry.getKey());
+
+            if (valueField.containsTrait(ListTrait.class)) {
+                context.request().addParameters(keyMarshaller.convert(mapEntry.getKey(), null),
```
shorea: Do we have to assume it's a list of strings here?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/protocol/restxml/XmlGenerator.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+package software.amazon.awssdk.core.internal.protocol.restxml;
+
+import java.io.StringWriter;
+import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.core.util.xml.XmlWriter;
+
+@SdkInternalApi
+public interface XmlGenerator {
```
shorea: Do we need this abstraction? The reasoning for JSON having it is to support other wire formats like ION and CBOR.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/protocol/restxml/XmlPayloadMarshaller.java
```diff
@@ -0,0 +1,255 @@
...
+import software.amazon.awssdk.core.util.SdkAutoConstructMap;
+
+@SdkInternalApi
+public class XmlPayloadMarshaller {
+
+    public static final XmlMarshaller<String> STRING = new BasePayloadMarshaller<String>(FROM_STRING) {
+        @Override
+        public void marshall(String val, XmlMarshallerContext context, String paramName,
+                                SdkField<String> sdkField, ValueToStringConverter.ValueToString<String> converter) {
+            context.xmlGenerator().xmlWriter().startElement(paramName).value(val).endElement();
```
shorea: In JSON we emit the field name in the base class. Could we do something similiar here? I.e. start and end in the base and emit just the value in the impl.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/protocol/restxml/XmlPayloadMarshaller.java
```diff
@@ -0,0 +1,255 @@
...
+                   (!list.isEmpty() || !(list instanceof SdkAutoConstructList));
+        }
+    };
+
+    public static final XmlMarshaller<Map> MAP = new BasePayloadMarshaller<Map>(null) {
+
+        @Override
+        public void marshall(Map map, XmlMarshallerContext context, String paramName,
+                             SdkField<Map> sdkField, ValueToStringConverter.ValueToString<Map> converter) {
+            context.xmlGenerator().startElement(paramName);
```
shorea: I think we have a test in the protocol suite that tests for flattened maps. I don't see any actual usage of it in the rest-xml services though. Maybe just make a note that we explicitly don't care about the flattened trait for maps.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/protocol/restxml/XmlProtocolMarshaller.java
```diff
@@ -0,0 +1,179 @@
...
+
+    private boolean isBinary(SdkField<?> field, Object val) {
+        return isExplicitPayloadMember(field) && val instanceof SdkBytes;
+    }
+
+    private boolean isExplicitPayloadMember(SdkField<?> field) {
+        return field.containsTrait(PayloadTrait.class);
+    }
+
+    private Object resolveValue(Object val, SdkField<?> marshallingInfo) {
```
shorea: I wonder if this is something we should move into the SdkField class since it's common across all protocols.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/protocol/restxml/XmlProtocolMarshaller.java
```diff
@@ -0,0 +1,179 @@
...
+
+    public void doMarshall(SdkPojo pojo) {
+        for (SdkField<?> field : pojo.sdkFields()) {
+            Object val = resolveValue(field.get(pojo), field);
+
+            if (isBinary(field, val)) {
+                request.setContent(((SdkBytes) val).asInputStream());
+                setContentTypeHeaderIfNeeded("binary/octet-stream");
+
+            } else if (isExplicitPayloadMember(field) && val instanceof String) {
```
shorea: This branch seems very S3 specific. Is there anything we can do to keep it separate?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/protocol/restxml/XmlProtocolMarshallerBuilder.java
```diff
@@ -0,0 +1,57 @@
...
+ * @param <T> Type of the original request object.
+ */
+@SdkInternalApi
+public class XmlProtocolMarshallerBuilder<T extends SdkRequest>  {
+
+    private XmlGenerator xmlGenerator;
+    private T originalRequest;
+    private OperationInfo operationInfo;
+
+    public static <T extends SdkRequest> XmlProtocolMarshallerBuilder<T> standard() {
```
shorea: Maybe just create(). Standard is more a v1 style.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/protocol/restxml/unmarshall/XmlUnmarshaller.java
```diff
@@ -0,0 +1,26 @@
...
+
+package software.amazon.awssdk.core.internal.protocol.restxml.unmarshall;
+
+import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.core.protocol.SdkField;
+
+@SdkInternalApi
+public interface XmlUnmarshaller<T> {
+
+    T unmarshall(XmlUnmarshallerContext context,
```
shorea: NP one liner

pom.xml
```diff
@@ -549,7 +549,7 @@
                     </execution>
                 </executions>
                 <configuration>
-                    <failOnWarning>true</failOnWarning>
+                    <failOnWarning>false</failOnWarning>
```
shorea: What's this?

services/budgets/pom.xml
```diff
@@ -29,4 +29,14 @@
         AWS Budgets Service
     </description>
     <url>https://aws.amazon.com/sdkforjava</url>
+
+    <dependencies>
```
shorea: Remove

services/cloudfront/src/main/resources/codegen-resources/log4j.properties
```diff
@@ -0,0 +1,29 @@
+
+#
```
shorea: Why's this in codegen-resources?

services/s3/pom.xml
```diff
@@ -63,5 +63,9 @@
             <version>${awsjavasdk.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
shorea: Remove

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/protocol/restxml/XmlPayloadMarshaller.java
varunnvs92: For list, we should add the top level startElement/endElement only if it is not flattened. We don't have that information is base class, hence we can't move the start/end tags to base.

shorea: Could we just have a separate impl for list and maps? Seems like we are duplicating a lot of effort just to handle two of the types.

protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/internal/marshall/XmlPayloadMarshaller.java
```diff
@@ -0,0 +1,188 @@
...
+    public static final XmlMarshaller<Boolean> BOOLEAN = new BasePayloadMarshaller<>(ValueToStringConverter.FROM_BOOLEAN);
+
+    public static final XmlMarshaller<Instant> INSTANT =
+        new BasePayloadMarshaller<>(XmlProtocolMarshaller.INSTANT_VALUE_TO_STRING);
+
+    public static final XmlMarshaller<SdkBytes> SDK_BYTES = new BasePayloadMarshaller<SdkBytes>(null) {
+        @Override
+        public void marshall(SdkBytes val, XmlMarshallerContext context, String paramName, SdkField<SdkBytes> sdkField,
+                             ValueToStringConverter.ValueToString<SdkBytes> converter) {
+            context.xmlGenerator().xmlWriter().value(val.asByteBuffer());
```
shorea: Is this base64 encoded?

protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/internal/unmarshall/XmlOperationMetadata.java
```diff
@@ -0,0 +1,56 @@
...
+        this.hasStreamingSuccessResponse = hasStreamingSuccessResponse;
+        return this;
+    }
+
+    public boolean useRootElement() {
+        return useRootElement;
+    }
+
+    /**
+     * This value indicates if the root element in the response xml document should be ignored
```
shorea: Do we need this or can we just use the PayloadTrait?

services/rds/src/test/java/software/amazon/awssdk/services/rds/PresignRequestHandlerTest.java
```diff
@@ -55,7 +55,8 @@
     private static final Region DESTINATION_REGION = Region.of("us-west-2");
 
     private static RdsPresignInterceptor<CopyDbSnapshotRequest> presignInterceptor = new CopyDbSnapshotPresignInterceptor();
-    private final CopyDbSnapshotRequestMarshaller marshaller = new CopyDbSnapshotRequestMarshaller(new AwsQueryProtocolFactory());
+    private final CopyDbSnapshotRequestMarshaller marshaller =
+        new CopyDbSnapshotRequestMarshaller(AwsQueryProtocolFactory.builder().build());
```
shorea: Should we just have a create method?

services/s3/src/main/java/software/amazon/awssdk/services/s3/handlers/AddContentMd5HeaderInterceptor.java
```diff
@@ -0,0 +1,51 @@
...
+import software.amazon.awssdk.utils.IoUtils;
+import software.amazon.awssdk.utils.Md5Utils;
+
+@SdkProtectedApi
+public class AddContentMd5HeaderInterceptor implements ExecutionInterceptor {
+
+    @Override
+    public SdkHttpFullRequest modifyHttpRequest(Context.ModifyHttpRequest context, ExecutionAttributes executionAttributes) {
+        SdkHttpFullRequest request = context.httpRequest();
+        if (request.contentStreamProvider().isPresent() && !request.firstMatchingHeader(CONTENT_MD5).isPresent()) {
```
shorea: Does this add MD5 for all operations? what about PutObject?

protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/internal/marshall/XmlPayloadMarshaller.java
varunnvs92: yes ([here](https://github.com/aws/aws-sdk-java-v2/blob/master/core/sdk-core/src/main/java/software/amazon/awssdk/core/util/xml/XmlWriter.java#L134-L136))

protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/internal/unmarshall/XmlOperationMetadata.java
varunnvs92: As we discussed yesterday, we are setting a customization to enable this value for operation like S3 `GetBucketLocationOutput`. So we need it

services/rds/src/test/java/software/amazon/awssdk/services/rds/PresignRequestHandlerTest.java
varunnvs92: We need to add overload `create` methods if we add state to this class, so i avoided it. We don't use it in many places, so it should be fine. We can always add it later.

protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/internal/marshall/XmlPayloadMarshaller.java
shorea: Is it worth just moving that up to this level? We've got a ValueToString for SDK_BYTES that base64 encodes already.

protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/internal/unmarshall/XmlOperationMetadata.java
shorea: I was thinking that we would use that customization to just apply the payload trait to LocationConstraint. Wouldn't that have the same behavior?

services/rds/src/test/java/software/amazon/awssdk/services/rds/PresignRequestHandlerTest.java
shorea: Sounds good.

protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/internal/marshall/XmlPayloadMarshaller.java
varunnvs92: Makes sense. I will use it

services/s3/src/main/java/software/amazon/awssdk/services/s3/handlers/AddContentMd5HeaderInterceptor.java
varunnvs92: Ignoring PutObject and UploadPart as discussed offline

PR #759, opened 2018-10-15
Generation region/service metadata

spfink:
Introduces two new modules, codegen-lite and codegen-lite-maven-plugin in order to be able to generate region and service metadata for the regions module. 

Could also refactor codegen-lite into codegen-core and move the core non-service related codegen files to this module 

```java
@SdkPublicApi
public final class Batch implements ServiceMetadata {
    private static final String ENDPOINT_PREFIX = "batch";

    private static final Map<String, String> REGION_OVERRIDDEN_ENDPOINTS = new HashMap();

    private static final List<Region> REGIONS = Collections.unmodifiableList(Arrays.asList(Region.AP_NORTHEAST_1,
            Region.AP_NORTHEAST_2, Region.AP_SOUTH_1, Region.AP_SOUTHEAST_1, Region.AP_SOUTHEAST_2, Region.CA_CENTRAL_1,
            Region.EU_CENTRAL_1, Region.EU_WEST_1, Region.EU_WEST_2, Region.EU_WEST_3, Region.SA_EAST_1, Region.US_EAST_1,
            Region.US_EAST_2, Region.US_WEST_1, Region.US_WEST_2));

    private static final Map<String, String> SIGNING_REGION_OVERRIDES = new HashMap();

    @Override
    public List<Region> regions() {
        return REGIONS;
    }

    @Override
    public URI endpointFor(Region region) {
        return URI.create(REGION_OVERRIDDEN_ENDPOINTS.getOrDefault(region.id(), computeEndpoint(ENDPOINT_PREFIX, region)));
    }

    @Override
    public Region signingRegion(Region region) {
        return Region.of(SIGNING_REGION_OVERRIDES.getOrDefault(region.id(), region.id()));
    }
}
```
```java
@SdkPublicApi
public final class UsEast1 implements RegionMetadata {
    private static final String ID = "us-east-1";

    private static final String DOMAIN = "amazonaws.com";

    private static final String NAME = "US East (N. Virginia)";

    private static final String PARTITION = "aws";

    public String name() {
        return NAME;
    }

    public String domain() {
        return DOMAIN;
    }

    public String id() {
        return ID;
    }

    public String partition() {
        return PARTITION;
    }
}
```


shorea:
Could we rename the generated metadata file to something like BatchServiceMetadata?

shorea:
Can the region class itself be generated?

spfink:
Will add diff tests.

Fine with adding ServiceMetadata suffix.

Can switch Region to be generated as well.

core/regions/pom.xml
```diff
@@ -94,4 +89,44 @@
...
+                            </goals>
+                        </execution>
+                    </executions>
+                </plugin>
+            </plugins>
+        </pluginManagement>
+    </build>
+
+    <profiles>
+        <!-- Build steps executed for services with models -->
```
shorea: can you explain what this does? not quite getting it.

codegen-lite/src/test/resources/software/amazon/awssdk/codegen/lite/regions/service-metadata-provider.java
```diff
@@ -0,0 +1,372 @@
...
+import software.amazon.awssdk.regions.servicemetadata.WafServiceMetadata;
+import software.amazon.awssdk.regions.servicemetadata.WorkdocsServiceMetadata;
+import software.amazon.awssdk.regions.servicemetadata.WorkmailServiceMetadata;
+import software.amazon.awssdk.regions.servicemetadata.WorkspacesServiceMetadata;
+import software.amazon.awssdk.regions.servicemetadata.XrayServiceMetadata;
+
+@Generated("software.amazon.awssdk:codegen")
+@SdkPublicApi
+public final class ServiceMetadataProvider {
+    private static final Map<String, ServiceMetadata> SERVICE_METADATA = Collections.unmodifiableMap(Stream.of(
```
shorea: Can we use ImmutableMap.Builder here? I feel like that would cut down on the verbosity.

codegen-lite/src/test/resources/software/amazon/awssdk/codegen/lite/regions/regions.java
```diff
@@ -19,125 +17,82 @@
 import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.ConcurrentHashMap;
+import software.amazon.awssdk.annotations.Generated;
 import software.amazon.awssdk.annotations.SdkPublicApi;
 import software.amazon.awssdk.utils.Validate;
 
-/**
- * An Amazon Web Services region that hosts a set of Amazon services.
```
shorea: Can we keep the docs in the generated version?

codegen-lite/src/test/resources/software/amazon/awssdk/codegen/lite/regions/region-metadata-provider.java
```diff
@@ -0,0 +1,69 @@
...
+import software.amazon.awssdk.regions.regionmetadata.SaEast1;
+import software.amazon.awssdk.regions.regionmetadata.UsEast1;
+import software.amazon.awssdk.regions.regionmetadata.UsEast2;
+import software.amazon.awssdk.regions.regionmetadata.UsGovWest1;
+import software.amazon.awssdk.regions.regionmetadata.UsWest1;
+import software.amazon.awssdk.regions.regionmetadata.UsWest2;
+
+@Generated("software.amazon.awssdk:codegen")
+@SdkPublicApi
+public final class RegionMetadataProvider {
```
shorea: Should we keep this an interface if we decide to support runtime discovery later on?

codegen-lite/src/test/resources/software/amazon/awssdk/codegen/lite/regions/regions.java
spfink: Ah, yes. This was just a miss.

codegen-lite/src/test/resources/software/amazon/awssdk/codegen/lite/regions/service-metadata-provider.java
spfink: Can do.

core/regions/pom.xml
spfink: This comment is incorrect as I just copied it from the other but it's just the configuration for running the codegen-lite-maven-plugin. It's exactly copied from the pom.xml of the services module just with the "lite" version.

codegen-lite/src/test/resources/software/amazon/awssdk/codegen/lite/regions/region-metadata-provider.java
spfink: Makes sense. Will do.

PR #760, opened 2018-10-15
AWS/Query and EC2 marshaller support.

shorea:


core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/protocol/json/JsonProtocolMarshaller.java
```diff
@@ -130,6 +125,24 @@ private static MarshallerRegistry createMarshallerRegistry() {
             .build();
     }
 
+    private static Map<MarshallLocation, TimestampFormatTrait.Format> getDefaultTimestampFormats() {
+        Map<MarshallLocation, TimestampFormatTrait.Format> formats = new HashMap<>();
+        formats.put(MarshallLocation.HEADER, TimestampFormatTrait.Format.ISO_8601);
```
varunnvs92: Header default format is rfc822

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/protocol/query/QueryProtocolMarshaller.java
```diff
@@ -15,12 +15,147 @@
...
+        .marshaller(MarshallingType.FLOAT, SimpleTypeQueryMarshaller.FLOAT)
+        .marshaller(MarshallingType.BOOLEAN, SimpleTypeQueryMarshaller.BOOLEAN)
+        .marshaller(MarshallingType.DOUBLE, SimpleTypeQueryMarshaller.DOUBLE)
+        .marshaller(MarshallingType.LONG, SimpleTypeQueryMarshaller.LONG)
+        .marshaller(MarshallingType.INSTANT, SimpleTypeQueryMarshaller.INSTANT)
+        .marshaller(MarshallingType.SDK_BYTES, SimpleTypeQueryMarshaller.SDK_BYTES)
+        .marshaller(MarshallingType.NULL, SimpleTypeQueryMarshaller.NULL)
+        .marshaller(MarshallingType.MAP, new MapMarshaller())
+        .marshaller(MarshallingType.LIST, new ListMarshaller())
+        .marshaller(MarshallingType.SDK_POJO, (req, path, val, sdkField) -> doMarshall(path, req, val))
```
varunnvs92: Do we need a BigDecimal marshalling type?

codegen/src/main/java/software/amazon/awssdk/codegen/poet/transform/protocols/QueryMarshallerSpec.java
```diff
@@ -0,0 +1,111 @@
...
+        initializationCodeBlockBuilder.add(".requestUri($S)", shapeModel.getMarshaller().getRequestUri())
+                                      .add(".httpMethodName($T.$L)", HttpMethodName.class, shapeModel.getMarshaller().getVerb())
+                                      .add(".hasExplicitPayloadMember($L)", shapeModel.isHasPayloadMember() ||
+                                                                            shapeModel.getExplicitEventPayloadMember() != null)
+                                      .add(".hasPayloadMembers($L)", shapeModel.hasPayloadMembers());
+
+        if (StringUtils.isNotBlank(shapeModel.getMarshaller().getTarget())) {
+            initializationCodeBlockBuilder.add(".operationIdentifier($S)", shapeModel.getMarshaller().getTarget())
+                                          .add(".apiVersion($S)", metadata.getApiVersion())
+                                          .add(".serviceName($S)", metadata.getServiceName());
```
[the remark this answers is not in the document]
shorea: good catch!

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/protocol/json/JsonProtocolMarshaller.java
shorea: Our protocol tests have ISO for marshalling and RFC822 for unmarshalling, not sure why. Did a search and no service is currently using timestamps in input headers so we may just have this wrong in our tests. I'll create a task to investigate.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/protocol/query/QueryProtocolMarshaller.java
shorea: Nah we have that as a customization for budgets. We actually don't support the "true" big decimal type as defined in the aws service framework. It's modeled as a string in budgets.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/protocol/json/JsonProtocolMarshaller.java
zoewangg: Yeah, for v1, we have to keep using iso as default for header marshalling for backwards compatibility reason. For v2, I think we can make it right and keep the same behavior as other SDKs.

varunnvs92: I think we can change it to Rfc822. Or you can add a TODO to investigate

varunnvs92: Add a TODO?

services/elasticloadbalancing/src/it/java/software/amazon/awssdk/services/elasticloadbalancing/ElbIntegrationTest.java
```diff
@@ -189,13 +190,14 @@ public void tearDown() throws Exception {
      */
     @Test
     public void testLoadBalancerInstanceOperations() throws Exception {
+        BasicConfigurator.configure();
```
varunnvs92: Do we need it?

shorea: Nope, will remove

Issue #722, opened 2018-09-23
S3 virtual-style addressing for non AWS endpoints


ebabani:
Any consideration to allowing DNS style bucked addresses for non s3 endpoints?

## Expected Behavior
Expect virtual style addressing to work for non-s3 endpoints as well.

## Current Behavior
Currently DNS style addressing is only enabled if the endpoint starts with s3. (https://github.com/aws/aws-sdk-java-v2/blob/1203086f21355a1fba45d4f0119b090cb3d77995/services/s3/src/main/java/software/amazon/awssdk/services/s3/handlers/EndpointAddressInterceptor.java#L141) 

There are s3 compatible providers that support only DNS style addressing, which makes this SDK harder to use with them. (https://www.alibabacloud.com/help/doc-detail/64919.htm)

## Possible Solution
Remove the restriction on DNS style addresses only working for the AWS s3 endpoints. 

## Context
Trying to use this client on Alicloud OSS.

## Your Environment
<N/A>

ebabani:
The change is fairly straight forward to make, but just want to know what the direction is regarding this. At the moment if feels like for s3 compatible providers this sdk leans towards forcing the path style.

millems:
It's tough for us to test to make sure we're compatible with all of them, but we're definitely not against it. We'd gladly accept a PR that fixes the issue.

blackstorm:
i decide back to sdk v1. Sad!

millems:
Note: This feature request is just a convenience. You can always enable virtual-style addressing manually:

```java
S3Client.builder()
        .serviceConfiguration(c -> c.pathStyleAccessEnabled(true))
        .build();
```

We don't consider this a blocker to moving to V2, so it's not high on our priorities.

yannrouillard:
Currently experiencing this issue as well.

@millems I saw you were not keen to merge the PR https://github.com/aws/aws-sdk-java-v2/pull/763, is there some other way to make it work that you would accept like creating a new option to force virtual-style addressing?

millems:
It's going to be difficult to figure out how to support this, with new features like [S3 access points](https://aws.amazon.com/s3/features/access-points/) adding even more complexity to the URL building process, and the S3 specification being implemented differently by different providers.

Is there a specific non-AWS runtime we can focus in on supporting, so that we don't just have it be "all non-AWS S3 implementations"?

yannrouillard:
Hi @millems,

[...]

PR #763, opened 2018-10-17
Use virtual style addressing for non Amazon S3 endpoints.

ebabani:
## Description
Allow using the dns style (bucket.s3-endpoint) instead or forcing the path style (s3-endpoint/bucket) for non Amazon s3-compatible endpoint. This allows using the sdk with third party providers that only support the dns style for access (eg. Alicloud)

## Motivation and Context

Allows using this sdk with any third party s3 providers that support virtual style addressing.

## Types of changes

## Checklist

## License

### DIFF
--- a/.changes/next-release/feature-AmazonS3-4d7c0ed.json
+++ b/.changes/next-release/feature-AmazonS3-4d7c0ed.json
@@ -1,0 +1,5 @@
+{
+    "category": "Amazon S3", 
+    "type": "feature", 
+    "description": "Use virtual style addressing for third party s3 endpoints."
+}
--- a/services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Configuration.java
+++ b/services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Configuration.java
@@ -80,7 +80,7 @@ public final class S3Configuration implements ServiceConfiguration, ToCopyableBu
      * requests.
      * </p>
      *
-     * @return True is the client should always use path-style access
+     * @return True if the client should always use path-style access
      */
     public boolean pathStyleAccessEnabled() {
         return pathStyleAccessEnabled;
--- a/services/s3/src/main/java/software/amazon/awssdk/services/s3/handlers/EndpointAddressInterceptor.java
+++ b/services/s3/src/main/java/software/amazon/awssdk/services/s3/handlers/EndpointAddressInterceptor.java
@@ -138,8 +138,12 @@ public final class EndpointAddressInterceptor implements ExecutionInterceptor {
      * @param bucketName     Bucket name for this particular operation.
      */
     private void changeToDnsEndpoint(SdkHttpFullRequest.Builder mutableRequest, String bucketName) {
-        if (mutableRequest.host().startsWith("s3")) {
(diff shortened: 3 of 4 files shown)


ebabani:
Related issue: https://github.com/aws/aws-sdk-java-v2/issues/722

ebabani:
@millems Could I get someone to take a look at this PR?

The travis failure seems unrelated, and I don't know if I have a way to re-run the test.

millems:
Sorry, talking with the others on our end, it looks like we've actually standardized this behavior across the latest SDK versions:
> Determining which addressing style to use:
>
> In general, SDK's MUST default to using virtual hosted-style addressing when making S3 API calls that address an S3 bucket. However, SDK's MUST default to path style addressing under the following scenarios:
> 
> * The S3 bucket being addressed is non-DNS compatible. To determine DNS compatibility of a bucket, read the section on DNS compatibility.
> 
> * The S3 bucket being addressed has one or more periods (.) in its name. If the request is being sent over HTTP instead of HTTPS, SDK's MAY chose to default to virtual hosted-style addressing despite the period in the bucket name.
> 
> * The client is explicitly configured to use path style addressing and the request being sent is not to an S3 accelerate endpoint. For more information related to configuration, read the section on client configuration.

You can explicitly disable path-style addressing in the S3 client builder:

```Java
S3Client.builder()
        .serviceConfiguration(c -> c.pathStyleAccessEnabled(true))
        .build();
```

... but in the interest of behavioral parity with the other SDKs, we won't be able to accept this PR.

ebabani:
@millems Totally agree with that behaviour, but at the moment this SDK is NOT using virtual hosted-style addressing for third party endpoints, and there is no way to change that.

https://github.com/aws/aws-sdk-java-v2/blob/master/services/s3/src/main/java/software/amazon/awssdk/services/s3/handlers/EndpointAddressInterceptor.java#L141

This PR addresses that limitation, but does not affect the addressing behaviour above.

millems:
It looks like setting path-style access to enabled should work for third-party endpoints: https://github.com/aws/aws-sdk-java-v2/blob/master/services/s3/src/main/java/software/amazon/awssdk/services/s3/handlers/EndpointAddressInterceptor.java#L63-L69

Is that not the case?

ebabani:
No, the `changeToDnsEndpoint` function has a check that the endpoint must start with `s3`, otherwise it doesn't make the change.

https://github.com/aws/aws-sdk-java-v2/blob/master/services/s3/src/main/java/software/amazon/awssdk/services/s3/handlers/EndpointAddressInterceptor.java#L141
```
if (mutableRequest.host().startsWith("s3")) {
```

millems:
Ah, I misunderstood the request. For some reason I was thinking it was to use path-style addressing, not virtual style addressing. Reopening. I suppose it could be nice to be able to force virtual-style addressing on. We'll need to talk about whether we want to support this, since it will probably need to be a configuration option, not an implicit behavior.

millems:
Just commenting to thank you for your patience on this issue. I know we haven't replied in a while. I'm a little nervous about merging this as-is, because it could break people who need path-style for third-party services and are relying on the current behavior.

ebabani:
No worries, I'm good with closing this PR since it's difficult to change the default behaviour.  I was able to use a fork for my purposes.

PR #765, opened 2018-10-17
Remove JsonPath dependency from sfn module

zoewangg:
## Description
Remove jsonPath dependency from sfn module, which has also been done in v1.

## Motivation and Context
removing another non-modularized dependency

## Testing
`mvn clean install`

## License


shorea:
:( do we need to do this?

zoewangg:
Yeah, unfortunately, they are not modularized. Looking at their commit history, it doesn't seem like it's actively maintained. :( https://github.com/json-path/JsonPath/commits/master We can try to open a PR to add automatic-module-name to their repo, but I don't think it will get merged any time soon.

shorea:
Is it possible to shade it?

zoewangg:
We'd still need to add something like `require jsonpath;` to our module-info file as the following and if the module name for json path ever changes, it will break the customers.

https://github.com/aws/aws-sdk-java-v2/blob/1deb48527d130ab0312e57ce3c5061ad199f0aab/codegen/src/main/resources/templates/common/module-info.ftl#L7

shorea:
Why do we need it in our module info if we are shading it?

zoewangg:
As long as we import the jsonpath packages,
```
import com.jayway.jsonpath.InvalidPathException;
import com.jayway.jsonpath.JsonPath;
```
we need to add them to the module info. 

I tried to create a new module, named `thirdparty` to shade the json-path and relocating it to `software.amazon.awssdk.thirdparty.com.jayway.jsonpath`. Then I made sfn depends on `thirdparty` module and changed the package imports to 
`import software.amazon.awssdk.thirdparty.com.jayway.jsonpath.InvalidPathException;`.

This still doesn't work. `thirdparty` will not be able to export the shaded jsonpath pacakge from its `module-info` because it doesn't know the relocated package at compiler time.

```
module thirdparty {
    exports software.amazon.awssdk.thirdparty.com.jayway.jsonpath.InvalidPathException;
} 
```

There might be some other things we can try, but I don't think we can go full modularization with unmodularized dependencies.

zoewangg:
Yeah, shade is is bound to the package phase and we would get compiler error in `module-info.java` `exports software.amazon.awssdk.sfn.thirdparty.com.jayway.jsonpath.InvalidPathException`  because it doesn't know the relocated path at compile time.

zoewangg:
Correction: if we shade it within `sfn` module, we would get compiler error at `import software.amazon.awssdk.sfn.thirdparty.com.jayway.jsonpath.InvalidPathException` and not from the module-info class because we would not export it.

zoewangg:
I think we still need this because removing `jsonpath` later after GA would be considered backwards incompatible. We can add `jsonpath` in the future if it has `Automatic-Module-Name` added.

zoewangg:
Closing in favor of https://github.com/aws/aws-sdk-java-v2/pull/815

PR #766, opened 2018-10-18
Make request content repeatable w/o mark,reset

dagnir:
## Description
This commit removes the reliance on mark,reset on InputStreams for making them
them repeatable and instead abstracts it to an interface where the exact details
of resetting the content is left up to the implementer.

## Motivation and Context
Make content repeatable.

## Testing
`mvn clean install`

Will run integ tests.

## Types of changes

## Checklist

## License


dagnir:
@shorea I'm not aware of any instances where the SDK only partially reads the stream after accessing it; I considered expanding the interface so that `reset()` and `getStream()` were separate methods but I don't see a compelling reason to do it given how we currently use request InputStreams.

dagnir:
@shorea I think we can get away with just wrapping the providers themselves right? For example: https://github.com/aws/aws-sdk-java-v2/blob/cd4a268a6be017450bf0bc3ab9609c63a1d380c6/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/StreamManagingStage.java#L55

Do you see a case where this wouldn't work?

http-clients/apache-client/src/main/java/software/amazon/awssdk/http/apache/internal/RepeatableInputStreamRequestEntity.java
```diff
@@ -109,7 +110,8 @@ private long parseContentLength(String contentLength) {
      * @return The request content input stream or an empty input stream if there is no content.
      */
     private InputStream getContent(SdkHttpFullRequest request) {
-        return request.content().orElseGet(() -> new ByteArrayInputStream(new byte[0]));
+        return request.contentStreamProvider().map(ContentStreamProvider::reset)
```
shorea: Can we keep the orElseGet so we don't create the stream if we don't need it.

http-client-spi/src/main/java/software/amazon/awssdk/http/ContentStreamProvider.java
```diff
@@ -0,0 +1,36 @@
...
+ * returned from the previous invocation.
+ *
+ */
+@SdkPublicApi
+@FunctionalInterface
+public interface ContentStreamProvider {
+    /**
+     * @return The content stream.
+     */
+    InputStream reset();
```
shorea: Can we name this something other than reset? reset makes it sound side-affecty rather than returning a new stream. getNewStream()? getStream()? something like that.

http-clients/apache-client/src/main/java/software/amazon/awssdk/http/apache/internal/RepeatableInputStreamRequestEntity.java
dagnir: +1 done.

http-client-spi/src/main/java/software/amazon/awssdk/http/ContentStreamProvider.java
dagnir: done

PR #768, opened 2018-10-19
Update the way serviceSpecificHttpConfig customization is specified

varunnvs92:


spfink:
Looks good to me. Background on why this change is being made?

varunnvs92:
Some service teams want service level defaults for read/write timeout. In current design, we take a string which we generate as body in `serviceHttpConfig` method. This PR change the customization to take a FQCN that has `defaultHttpConfig` method similar to what we do for retry policy

PR #772, opened 2018-10-19
event streaming SigV4 signing

varunnvs92:
Pull code from internal repo

### DIFF
--- a/core/auth/src/main/java/software/amazon/awssdk/auth/signer/internal/AsyncSigV4SubscriberAdapter.java
+++ b/core/auth/src/main/java/software/amazon/awssdk/auth/signer/internal/AsyncSigV4SubscriberAdapter.java
@@ -1,0 +1,102 @@
+/*
+ * Copyright 2010-2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.auth.signer.internal;
+
+import java.nio.ByteBuffer;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicLong;
+
+import org.reactivestreams.Subscriber;
+import org.reactivestreams.Subscription;
+import software.amazon.awssdk.annotations.SdkProtectedApi;
+
+
+/**
+ * AsyncSigV4SubscriberAdapter appends an empty frame at the end of original Publisher
+ * <dl>
+ * <dt><b>Backpressure:</b></dt>
+ * <dd>The trailing empty frame is sent only if there is demand from the downstream subscriber</dd>
+ * </dl>
+ */
+@SdkProtectedApi
+final class AsyncSigV4SubscriberAdapter implements Subscriber<ByteBuffer> {
+    private AtomicBoolean upstreamDone = new AtomicBoolean(false);
+    private AtomicLong downstreamDemand = new AtomicLong();
+    private volatile boolean streamTerminate = false;
+    private Object lock = new Object();
+    private Subscriber<? super ByteBuffer> delegate;
+
+    AsyncSigV4SubscriberAdapter(Subscriber<? super ByteBuffer> actual) {
+        this.delegate = actual;
+    }
+
+    @Override
+    public void onSubscribe(Subscription s) {
+        delegate.onSubscribe(new Subscription() {
+            @Override
+            public void request(long n) {
+                if (n <= 0) {
+                    throw new IllegalArgumentException("n > 0 required but it was " + n);
+                }
+
+                downstreamDemand.getAndAdd(n);
+
+                if (upstreamDone.get()) {
+                    sendTrailingEmptyFrame();
+                } else {
+                    s.request(n);
+                }
+            }
+
+            @Override
+            public void cancel() {
+                s.cancel();
+            }
+        });
+    }
+
+    @Override
+    public void onNext(ByteBuffer byteBuffer) {
+        downstreamDemand.decrementAndGet();
+        delegate.onNext(byteBuffer);
+    }
+
+    @Override
+    public void onError(Throwable t) {
+        upstreamDone.compareAndSet(false, true);
+        delegate.onError(t);
+    }
+
+    @Override
+    public void onComplete() {
+        upstreamDone.compareAndSet(false, true);
+
+        if (downstreamDemand.get() > 0) {
+            sendTrailingEmptyFrame();
+        }
+    }
+
+    private void sendTrailingEmptyFrame() {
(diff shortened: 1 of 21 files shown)


core/auth/src/main/java/software/amazon/awssdk/auth/signer/internal/AsyncSigV4SubscriberAdapter.java
```diff
@@ -0,0 +1,102 @@
...
+ * <dd>The trailing empty frame is sent only if there is demand from the downstream subscriber</dd>
+ * </dl>
+ */
+@SdkProtectedApi
+final class AsyncSigV4SubscriberAdapter implements Subscriber<ByteBuffer> {
+    private AtomicBoolean upstreamDone = new AtomicBoolean(false);
+    private AtomicLong downstreamDemand = new AtomicLong();
+    private volatile boolean streamTerminate = false;
+    private Object lock = new Object();
+    private Subscriber<? super ByteBuffer> delegate;
```
shorea: Make final

core/auth/src/main/java/software/amazon/awssdk/auth/signer/internal/AsyncSigV4SubscriberAdapter.java
```diff
@@ -0,0 +1,102 @@
...
+
+        if (downstreamDemand.get() > 0) {
+            sendTrailingEmptyFrame();
+        }
+    }
+
+    private void sendTrailingEmptyFrame() {
+        //when upstream complete, send a trailing empty frame
+        synchronized (lock) {
+            if (!streamTerminate) {
```
shorea: streamTerminate is a bit ambiguous. Maybe something like sentTrailingFrame?

core/auth/src/main/java/software/amazon/awssdk/auth/signer/internal/AsyncSigV4SubscriberAdapter.java
```diff
@@ -0,0 +1,102 @@
...
+                streamTerminate = true;
+                delegate.onNext(ByteBuffer.wrap(new byte[] {}));
```
shorea: Does this actually do anything?

core/auth/src/main/java/software/amazon/awssdk/auth/signer/internal/BaseAsyncAws4Signer.java
```diff
@@ -0,0 +1,106 @@
...
+
+        return transformer.transform(requestPublisher);
+    }
+
+    /**
+     * Create a transformer for requestProvider. The transformer should add signing operator to the original requestProvider
+     *
+     * Can be overriden by subclasses to provide specific signing method
+     */
+    protected abstract RequestProviderTransformer createRequestProviderTransformer(String headerSignature,
```
shorea: Why not just have this method be the transformer instead of it returning a functional interface?

core/sdk-core/src/main/java/software/amazon/awssdk/core/signer/AsyncRequestBodySigner.java
```diff
@@ -0,0 +1,40 @@
...
+public interface AsyncRequestBodySigner {
+    /**
+     * Method that takes in an signed request and async request body provider,
+     * and returns a transformed version the request body provider.
+     *
+     * @param request             The signed request (with Authentication header)
+     * @param executionAttributes Contains the attributes required for signing the request
+     * @return The transformed requestPublisher (with singing operator)
+     */
+    SdkHttpContentPublisher signAsyncRequestBody(SdkHttpFullRequest request, SdkHttpContentPublisher requestPublisher,
```
shorea: Should we use AsyncRequestBody here instead of SdkHttpContentPublisher? They are kinda the same thing and since this is more public I'd err towards AsyncRequestBody. We'd have to do a bit of refactoring to pass AsyncRequestBody further down the stack and transform only at the MakeAsyncHttpRequest stage.

core/auth/src/main/java/software/amazon/awssdk/auth/signer/internal/AsyncSigV4SubscriberAdapter.java
varunnvs92: Will do

core/auth/src/main/java/software/amazon/awssdk/auth/signer/internal/AsyncSigV4SubscriberAdapter.java
varunnvs92: Will do

core/auth/src/main/java/software/amazon/awssdk/auth/signer/internal/AsyncSigV4SubscriberAdapter.java
varunnvs92: It sends an empty bytebuffer which is want we want. Are you asking a different question?

shorea: Is it wrapped with anything? Like it becomes the payload of a event stream message?

varunnvs92: No, it is not converted to flow Message. We just send empty byte array (no headers, no payload).

core/auth/src/main/java/software/amazon/awssdk/auth/signer/internal/BaseAsyncAws4Signer.java
varunnvs92: That should do it. I can make the change

core/sdk-core/src/main/java/software/amazon/awssdk/core/signer/AsyncRequestBodySigner.java
varunnvs92: Do you mean we should return AsyncRequestBody in this method and use that method somewhere [here](https://github.com/aws/aws-sdk-java-v2/blob/master/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/MakeAsyncHttpRequestStage.java#L98-L100)?

varunnvs92:
Updated the PR based on the comments

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/AmazonAsyncHttpClient.java
```diff
@@ -92,12 +92,12 @@ public RequestExecutionBuilder requestExecutionBuilder() {
...
 
         /**
-         * Fluent setter for {@link SdkHttpContentPublisher}
+         * Fluent setter for {@link AsyncRequestBody}
          *
          * @param requestProvider Request provider object
          * @return This builder for method chaining.
          */
-        RequestExecutionBuilder requestProvider(SdkHttpContentPublisher requestProvider);
+        RequestExecutionBuilder requestProvider(AsyncRequestBody requestProvider);
```
shorea: rename setter

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/RequestExecutionContext.java
```diff
@@ -56,7 +56,7 @@ public static Builder builder() {
         return new Builder();
     }
 
-    public SdkHttpContentPublisher requestProvider() {
+    public AsyncRequestBody requestProvider() {
```
shorea: rename

PR #777, opened 2018-10-25
Add functional and integration tests for module path

zoewangg:
## Description
This PR is to make sure we cover the module path testing for Java 9+. The existing tests we will only run on the class path and we cannot guarantee it will work on module path.

- Created `module-path-test` module and add functional and integration tests for module path
    - mock tests: calling xml/json prococol sync/async clients using mock http clients. (I tried Wiremock but there are split packages in wiremock and it throws errors when running in module path)
   - integ tests: calling `S3Client` using `UrlConnectionHttpClient` and `ApacheHttpClient`. (Netty client is not currently working because of two root packages)

- Updated CodeBuild buildspec files to run the tests only when the Java version is 9+

**Note**
- The reason that the "tests" are actually in the main methods and are not Junit tests is because `mvn-surefire-plugin` still doesn't have very good support on Java 9 module path. There are still gaps, eg: when I tried to run the same code using unit tests, I didn't get any errors using the netty client which I would get from running on the main program.

- `module-path-test` is not a module in the parent so it will not run when you do `mvn clean install` because it only compiles in Java 9+ and it will be triggered in our gate keeper CodeBuild jobs and release jobs.

## Motivation and Context
Testing module path using Java 9+

## Testing
`mvn exec:exec -P mock-tests`, `mvn exec:exec -P integ-tests` succeeded locally and on CodeBuild

## Types of changes

## Checklist

## License

### DIFF
--- a/test/module-path-tests/pom.xml
+++ b/test/module-path-tests/pom.xml
@@ -1,0 +1,169 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Copyright 2010-2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+  ~
+  ~ Licensed under the Apache License, Version 2.0 (the "License").
+  ~ You may not use this file except in compliance with the License.
+  ~ A copy of the License is located at
+  ~
+  ~  http://aws.amazon.com/apache2.0
+  ~
+  ~ or in the "license" file accompanying this file. This file is distributed
+  ~ on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+  ~ express or implied. See the License for the specific language governing
+  ~ permissions and limitations under the License.
+  -->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>aws-sdk-java-pom</artifactId>
+        <groupId>software.amazon.awssdk</groupId>
+        <version>2.0.0-preview-13-SNAPSHOT</version>
+        <relativePath>../../pom.xml</relativePath>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>module-path-tests</artifactId>
+    <!-- Setting it to a different version so it can always load dependencies from maven
+    rather than local module-->
+    <version>1.0.0-SNAPSHOT</version>
+
+    <name>AWS Java SDK :: Test :: Module Path Tests</name>
+    <description>A set of tests to run v2 in module path with Java 9+.</description>
+    <url>https://aws.amazon.com/sdkforjava</url>
+
+    <dependencies>
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>s3</artifactId>
+            <version>${project.parent.version}</version>
+            <exclusions>
+                <exclusion>
+                    <!-- TODO remove exclusions after we fix netty module -->
+                    <artifactId>netty-nio-client</artifactId>
+                    <groupId>software.amazon.awssdk</groupId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>protocol-tests</artifactId>
+            <version>${project.parent.version}</version>
+            <exclusions>
+                <exclusion>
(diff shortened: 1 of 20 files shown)


test/module-path-tests/pom.xml
```diff
@@ -0,0 +1,168 @@
...
+        <groupId>software.amazon.awssdk</groupId>
+        <version>2.0.0-preview-13-SNAPSHOT</version>
+        <relativePath>../../pom.xml</relativePath>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>module-path-tests</artifactId>
+    <!-- Setting it to a different version so it can always load dependencies from maven
+    rather than local module-->
+    <version>1.0.0-SNAPSHOT</version>
```
millems: Why do we want to always load from maven?

test/module-path-tests/pom.xml
```diff
@@ -0,0 +1,168 @@
...
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>s3</artifactId>
+            <version>${project.parent.version}</version>
+            <exclusions>
+                <exclusion>
+                    <!-- TODO remove exclusions after we fix netty module -->
+                    <artifactId>netty-nio-client</artifactId>
+                    <groupId>software.amazon.awssdk</groupId>
+                </exclusion>
```
millems: What about SLF4J?

build-tools/src/main/resources/software/amazon/awssdk/checkstyle-suppressions.xml
```diff
@@ -27,6 +27,9 @@
     <suppress checks=".*"
               files=".(flow)[\\/].+\.java$"/>
 
+    <suppress checks=".*"
+              files=".(module-path-tests)[\\/].+\.java$"/>
+
```
millems: Why?

test/module-path-tests/pom.xml
```diff
@@ -0,0 +1,168 @@
...
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>aws-sdk-java-pom</artifactId>
+        <groupId>software.amazon.awssdk</groupId>
+        <version>2.0.0-preview-13-SNAPSHOT</version>
+        <relativePath>../../pom.xml</relativePath>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>module-path-tests</artifactId>
```
millems: Can you exclude this module from the importer, as well?

test/module-path-tests/src/main/java/software/amazon/awssdk/modulepath/tests/integtests/S3ApiCall.java
```diff
@@ -0,0 +1,50 @@
...
+import software.amazon.awssdk.http.urlconnection.UrlConnectionHttpClient;
+import software.amazon.awssdk.services.s3.S3Client;
+import software.amazon.awssdk.testutils.service.AwsTestBase;
+
+public class S3ApiCall extends AwsTestBase {
+    private static final Logger logger = LoggerFactory.getLogger(S3ApiCall.class);
+
+    private S3Client s3Client = S3Client.builder()
+                                        .credentialsProvider(CREDENTIALS_PROVIDER_CHAIN)
+                                        .build();
```
millems: Region?

test/module-path-tests/pom.xml
zoewangg: Because Intellij will always try to load dependencies from local module instead of maven and it shows a bunch of errors saying "module not found" because those modules don't have `Automatic-Module-Name` at compile time.

screen shot 2018-10-26 at 12 51 11 pm

build-tools/src/main/resources/software/amazon/awssdk/checkstyle-suppressions.xml
zoewangg: Forgot to remove it, will fix.

test/module-path-tests/pom.xml
zoewangg: This module is in the test folder though, I thought we don't import test folder

test/module-path-tests/src/main/java/software/amazon/awssdk/modulepath/tests/integtests/S3ApiCall.java
zoewangg: Yeah, will add

test/module-path-tests/pom.xml
zoewangg: We can still use slf4j, with the derived module names.

```
    requires slf4j.api;
    requires slf4j.simple;
```

PR #778, opened 2018-10-25
Fix Publisher.subscribe running check and add module name to poms

zoewangg:
## Description
- Fix `Publisher.subscribe` running check 
- add the metadata information such as name, description to the protocol poms

### DIFF
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/async/SimpleHttpContentPublisher.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/async/SimpleHttpContentPublisher.java
@@ -62,12 +62,14 @@ public final class SimpleHttpContentPublisher implements SdkHttpContentPublisher
 
         @Override
         public void request(long n) {
-            if (n <= 0) {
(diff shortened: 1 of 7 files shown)


core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/async/SimpleHttpContentPublisher.java
```diff
@@ -62,12 +62,14 @@ private SubscriptionImpl(Subscriber<? super ByteBuffer> s) {
 
         @Override
         public void request(long n) {
-            if (n <= 0) {
-                s.onError(new IllegalArgumentException("Demand must be positive"));
-            } else if (running) {
-                s.onNext(ByteBuffer.wrap(content));
-                s.onComplete();
+            if (running) {
```
shorea: Do we need to use an atomic boolean here?

dagnir: I don't think we need to

> A Subscriber MUST ensure that all calls on its Subscription take place from the same thread or provide for respective external synchronization.

PR #779, opened 2018-10-25
Don't notify onError in MakeAsyncHttpRequestStage

dagnir:
## Description

Move the notification back on AsyncRetryable stage so we can decide whether to
call onError based on whether we are retrying or not.

## Motivation and Context
Fix onError behavior.

## Types of changes

## Checklist

## License


core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/AsyncRetryableStage.java
```diff
@@ -128,6 +132,7 @@ private void retryResponseIfNeeded(Response<OutputT> resp, CompletableFuture<Res
                 }
 
                 if (shouldRetry(resp.httpResponse(), resp.exception())) {
+                    responseHandler.onError(err);
```
shorea: Can we add a comment about the last error being delivered in the client? That's a bit tricky and not obvious from the code.

PR #780, opened 2018-10-25
Updating to use all regions defined for each service in endpoints.json

spfink:
The REGIONS array that is in the generated service metadata from each client will now include all regions defined by that service even if those regions aren't explicitly listed as being part of that partition. This means including regions like us-east-1-fips or even "sandbox" for MTurk and "local" for DDB.

## Motivation and Context
This change allows IAM and other global services to accurately resolve to the global endpoint for that service.

## Testing
Updated codegen tests and verified IAM tests now pass.

## Types of changes

## Checklist

## License


zoewangg:
Should we filter out those regions in the codegen?

spfink:
@zoewangg not sure if we want to. They aren't being included in the generated Region class as they aren't actual AWS regions. They are only included in the service metadata and ultimately they do make it easier to use those services with endpoints like FIPS.

The local one for DDB is probably one we can filter out without issue.

PR #781, opened 2018-10-25
Checking in marshaller benchmark code.

shorea:


pom.xml
```diff
@@ -69,6 +69,7 @@
         <module>test/service-test-utils</module>
         <module>test/test-utils</module>
         <module>test/codegen-generated-classes-test</module>
+        <module>test/sdk-benchmarks</module>
```
zoewangg: Hmm, do we want to put it here? This will get released to maven. We can probably just run it locally.

test/sdk-benchmarks/src/main/java/software/amazon/awssdk/benchmark/dynamodb/AbstractItemFactory.java
```diff
@@ -0,0 +1,144 @@
+package software.amazon.awssdk.benchmark.dynamodb;
```
zoewangg: It seems a bit odd to me the groupId is `com.dongieagnir` but the package is under `software.amazon.awssdk` and do we need to put copyright to all the files in this module?

shorea: Oops moved the package but not the group. I'll fix.

pom.xml
shorea: Is there a way to build it but not release it?

zoewangg: There might be a way but I'm not sure. We can put it here for now and figure out how to exclude the test modules later

test/sdk-benchmarks/pom.xml
```diff
@@ -0,0 +1,214 @@
...
+  ~ permissions and limitations under the License.
+  -->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <groupId>software.amazon.awssdk</groupId>
+    <artifactId>sdk-benchmarks</artifactId>
+    <version>1.0-SNAPSHOT</version>
```
zoewangg: Can we update the version to be same as the parent module since we are releasing them together? The release might be an issue if they have different versions.

dagnir:
🎆

PR #784, opened 2018-10-26
Merge 2.0.0-kinesis-event-streaming branch with master

varunnvs92:


http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/NettyNioAsyncHttpClient.java
```diff
@@ -277,6 +277,7 @@ public void close() {
         Builder protocol(Protocol protocol);
 
         /**
+<<<<<<< HEAD
```
zoewangg: This needs to be removed

varunnvs92: will do

varunnvs92:
Updated based on internal/external comments

PR #788, opened 2018-10-29
Add h2 protocol even when service defaults are set

varunnvs92:
**Current Behavior**
H2 protocol is ignored if service http defaults are set

**New Behavior**
H2 protocol will be added to AttributeMap if service defaults is missing the `PROTOCOL` attribute

### DIFF
--- a/codegen/src/main/java/software/amazon/awssdk/codegen/poet/builder/BaseClientBuilderClass.java
+++ b/codegen/src/main/java/software/amazon/awssdk/codegen/poet/builder/BaseClientBuilderClass.java
@@ -80,11 +80,7 @@ public class BaseClientBuilderClass implements ClassSpec {
                    .addMethod(beanStyleSetServiceConfigurationMethod());
         }
 
-        if (model.getCustomizationConfig().getServiceSpecificHttpConfig() != null) {
-            builder.addMethod(serviceSpecificHttpConfigMethod());
-        } else if (model.getMetadata().supportsH2()) {
-            builder.addMethod(enableH2HttpConfigMethod());
-        }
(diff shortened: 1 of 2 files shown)


codegen/src/main/java/software/amazon/awssdk/codegen/poet/builder/BaseClientBuilderClass.java
```diff
@@ -180,27 +176,45 @@ private MethodSpec beanStyleSetServiceConfigurationMethod() {
...
+    private CodeBlock serviceSpecificHttpConfigMethodBody(String serviceDefaultFqcn, boolean supportsH2) {
+        CodeBlock.Builder builder =  CodeBlock.builder()
+                                              .addStatement("$1T result = $1T.builder().build()", AttributeMap.class);
+
+        if (serviceDefaultFqcn != null) {
+            builder.addStatement("result = result.merge($T.defaultHttpConfig())",
+                                 PoetUtils.classNameFromFqcn(model.getCustomizationConfig().getServiceSpecificHttpConfig()));
+        }
+
+        if (supportsH2) {
```
shorea: Minor but can we move this up so we only have two merges instead of three when h2 and service specific config is present? Also would avoid mutation of a local var. I.E. code that looks like this

```java
AttributeMap result = AttributeMap.empty();
return result.merge(serviceConfig);
```

or 

```java
AttributeMap result = AttributeMap.builder()
    .put(PROTOCOL, HTTP2)
    .build();
return result.merge(serviceConfig);
```

varunnvs92: Updated

PR #795, opened 2018-11-04
Add support for endpoint trait

varunnvs92:
**Fixed:** ~~[WIP] APIs with path params are being resolved incorrectly. Will investigate~~

**Ready to merge**


core/aws-core/src/main/java/software/amazon/awssdk/awscore/client/builder/AwsDefaultClientBuilder.java
```diff
@@ -104,7 +105,8 @@ protected final SdkClientConfiguration mergeChildDefaults(SdkClientConfiguration
                                   .option(AwsAdvancedClientOption.ENABLE_DEFAULT_REGION_DETECTION, true)
                                   .option(AwsClientOption.CREDENTIALS_PROVIDER, DefaultCredentialsProvider.create())
                                   .option(SdkClientOption.RETRY_POLICY, AwsRetryPolicy.defaultRetryPolicy())
-                                  .option(AwsClientOption.SERVICE_SIGNING_NAME, signingName()));
+                                  .option(AwsClientOption.SERVICE_SIGNING_NAME, signingName())
+                                  .option(SdkAdvancedClientOption.DISABLE_HOST_PREFIX_INJECTION, false));
```
millems: Why is this an option?

core/sdk-core/src/main/java/software/amazon/awssdk/core/client/handler/BaseClientHandler.java
```diff
@@ -52,7 +55,9 @@ protected BaseClientHandler(SdkClientConfiguration clientConfiguration) {
 
         runBeforeMarshallingInterceptors(executionContext);
         Request<InputT> request = executionParams.getMarshaller().marshall(inputT);
-        request.setEndpoint(clientConfiguration.option(SdkClientOption.ENDPOINT));
+
+
+        request.setEndpoint(resolveEndpoint(clientConfiguration, executionParams.hostPrefixExpression()));
```
shorea: This actually changes a bit with the marshaller refactor. I'm curious to get your opinion on this, right now I'm passing the client configured endpoint to the marshaller and setting the endpoint there. I need to do that because SdkHttpFullRequest can't be built without an endpoint. We could a couple of things I think

1) Relax constraints in SdkHttpFullRequest to not require endpoint on build so we can open it back up later and add endpoint
2) Change marshallers to produce a SdkHttpFullRequest.Builder and fill in the endpoint and build later on in the request lifecycle
3) Somehow get the operation level endpoint to the marshaller and set it there
4) Continue to just set the client level endpoint knowing that it may be overwritten later

Thoughts?

varunnvs92: I am leaning towards 4.
> 2. Change marshallers to produce a SdkHttpFullRequest.Builder and fill in the endpoint and build later on in the request lifecycle

I think its better to pass built objects to avoid any mutability concerns

> 3. Somehow get the operation level endpoint to the marshaller and set it there

Have to do it for each protocol. Don't like it

> 4. Continue to just set the client level endpoint knowing that it may be overwritten later

I like this as we get marshalled request and we can construct a new request (with new endpoint) if host prefix is set. As we don't reuse the marshalled request at any place other than `BaseClientHandler`, it seems safe and modification happen at single place. 
If we are choosing between 4 and 1, relaxing constraints doesn't give any advantage and 4 seems ideal.

What do you think?

core/aws-core/src/main/java/software/amazon/awssdk/awscore/client/builder/AwsDefaultClientBuilder.java
varunnvs92: IMO it is an advanced option and not common use case to set it, so I made it an option instead of directing exposing it in [ClientOverrideConfiguration](https://github.com/aws/aws-sdk-java-v2/blob/master/core/sdk-core/src/main/java/software/amazon/awssdk/core/client/config/ClientOverrideConfiguration.java). Do you think it makes more sense to expose it as top level member in ClientOverrideConfiguration?

millems: Can it not be an option at all?

millems: Talked offline. It can be useful in testing.

core/sdk-core/src/main/java/software/amazon/awssdk/core/client/handler/BaseClientHandler.java
varunnvs92: Discussed offline. Going to keep the logic to update endpoint in `BaseClientHandler` instead of moving to marshallers as we need to pass more information every time we need to support a feature updating endpoint (like endpoint discovery).

varunnvs92:
Addressed all the comments

varunnvs92:
APIs with path params are being resolved incorrectly. Will investigate

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/client/test-async-client-class.java
```diff
@@ -145,6 +146,9 @@ public final String serviceName() {
     @Override
     public CompletableFuture<APostOperationResponse> aPostOperation(APostOperationRequest aPostOperationRequest) {
         try {
+            String hostPrefix = "{StringMember}-foo.";
```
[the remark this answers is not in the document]
varunnvs92: +1
https://github.com/aws/aws-sdk-java-v2/pull/883

PR #799, opened 2018-11-06
Delete codegen config related files

varunnvs92:
We don't use codegen.config file for aws services.
It is referenced in code path related to "api-gateway" protocol. We don't have support for apig service generation in v2. We can revisit and create a config file appropriate when supporting ApiGateway.


codegen/src/main/java/software/amazon/awssdk/codegen/AddMetadata.java
```diff
@@ -50,18 +47,17 @@ public static Metadata constructMetadata(ServiceModel serviceModel,
...
-            // keeping it up to date at this time. Just be aware this has happened when updating the API gateway code.
-            serviceName = codeGenConfig.getInterfaceName();
-            rootPackageName = codeGenConfig.getPackageName();
-
-            metadata.withDefaultEndpoint(codeGenConfig.getEndpoint())
-                    .withDefaultEndpointWithoutHttpProtocol(
-                            Utils.getDefaultEndpointWithoutHttpProtocol(codeGenConfig.getEndpoint()))
-                    .withDefaultRegion(codeGenConfig.getDefaultRegion());
+            // TODO: We don't use codegen config for aws services in v2. Revisit this code when we support
+            // API gateway generation in V2
```
zoewangg: Should we just remove the commented-out lines? Doesn't seem to be beneficial to keep them here. We can use git history to see the old code if we want.

varunnvs92: Done

PR #800, opened 2018-11-06
Warn, instead of fail, when potential simple methods aren't blacklisted or whitelisted.

millems:


codegen/src/main/java/software/amazon/awssdk/codegen/IntermediateModelBuilder.java
```diff
@@ -245,12 +245,14 @@ private void setSimpleMethods(IntermediateModel model) {
                 && !(config.getBlacklistedSimpleMethods().size() == 1 && config.getBlacklistedSimpleMethods().get(0).equals("*"))
                 && !m.getValue().hasStreamingInput()
                 && !m.getValue().hasStreamingOutput()) {
+
                 if (!methodName.matches(Constant.APPROVED_SIMPLE_METHOD_VERBS) &&
                     !config.getVerifiedSimpleMethods().contains(methodName)) {
-                    throw new RuntimeException("Simple method encountered that is not approved or blacklisted: " + methodName);
+                    // TODO: How do we prevent these from being missed before services launch?
```
zoewangg: write into a json file and put it to test module, maybe?

zoewangg: nvm, I thought the TODO is to track the missed methods.

PR #801, opened 2018-11-06
Moving USER_AGENT_SUFFIX value to the end of the user agent

varunnvs92:


core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/ApplyUserAgentStage.java
```diff
@@ -83,6 +79,11 @@ private String getUserAgent(SdkClientConfiguration config, List<ApiName> request
             userAgent.append(SPACE).append(requestUserAgent);
         }
 
+        // Note: This should be at the end of user agent
```
zoewangg: Can we move it to a method, something like `appendUserDefinedSuffix`?  Can we also add a test for it (rather than relying on the note)?

varunnvs92: I moved it to a different method. I can't think of a test that can catch these issues reliably. If we decide to include new config options in user agent, any test we write now can't take the future variables into account.

varunnvs92: The other option is to create an new ApplyUserAgentSuffixStage. I prefer to keep all user agent related changes in single class. But I am okay with splitting too if we decide so.

zoewangg: A new ApplyUserAgentSuffixStage seems a bit overkill. Keeping it as it is looks good to me.

PR #815, opened 2018-11-12
Remove step function high level library for now

zoewangg:
Remove step function high level library for now


shorea:
NVM I see it was deleted.

PR #821, opened 2018-11-13
Fix update-snapshot-version yml

zoewangg:
Fix update-snapsot-version to check master to avoid detached HEAD


test/module-path-tests/pom.xml
```diff
@@ -20,7 +20,7 @@
     <parent>
         <artifactId>aws-sdk-java-pom</artifactId>
         <groupId>software.amazon.awssdk</groupId>
-        <version>2.0.0-preview-13-SNAPSHOT</version>
+        <version>2.0.0-preview-14-SNAPSHOT</version>
```
zoewangg: This is the short term fix. I will figure out a way to update the parent version in this module.

PR #823, opened 2018-11-14
Async interceptor payload + S3 Checksums

spfink:
Exposes the ability to modify the AsyncRequestBody and AsyncResponseTransformer.

Currently this is done by adding two new interceptors.

For AsyncRequestBody:

```java
default AsyncRequestBody modifyAsyncRequestBody(Context.ModifyRequest context,
                                                ExecutionAttributes executionAttributes) {
    return context.requestBody();
}
```

For AsyncResponseTransformer:

```java
default AsyncResponseTransformer modifyAsyncResponseTransformer(Context.ModifyRequest context,
                                                                ExecutionAttributes executionAttributes) {
    return context.responseTransformer();
}
```

Possible other options include making modifyRequest return back a container object instead of an SdkRequest:

```java
default SdkRequestContext modifyRequest(Context.ModifyRequest context, ExecutionAttributes executionAttributes) {
    return context.request();
}
```

Where SdkRequestContext contains SdkRequest, AsyncRequestBody and AsyncResponseTransformer.

Or we could introduce an AsyncExecutionInterceptor interface that again either has two methods as in the PR or one method with a container object for async.

In general, trying to avoid potential confusion of when these interceptors are valid and when the values will be populated. If we use the approach in the PR, users of the Sync client should not use the new interceptors. If we go with the container approach for modifyRequest, then the asyncRequestBody and asyncResponseTransformer will be null.

### DIFF
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/ClientType.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/ClientType.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2010-2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.core;
+
+import software.amazon.awssdk.annotations.SdkPublicApi;
+
+/**
+ * Enum that represents the type of client being used.
+ */
+@SdkPublicApi
+public enum ClientType {
+
+    ASYNC("Async"),
+    SYNC("Sync");
+
+    private final String clientType;
+
+    ClientType(String clientType) {
+        this.clientType = clientType;
+    }
+
+    /* (non-Javadoc)
+     * @see java.lang.Enum#toString()
+     */
+    @Override
+    public String toString() {
+        return clientType;
+    }
+}
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/client/builder/SdkDefaultClientBuilder.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/client/builder/SdkDefaultClientBuilder.java
@@ -15,6 +15,8 @@
 
 package software.amazon.awssdk.core.client.builder;
 
+import static software.amazon.awssdk.core.ClientType.ASYNC;
+import static software.amazon.awssdk.core.ClientType.SYNC;
 import static software.amazon.awssdk.core.client.config.SdkAdvancedAsyncClientOption.FUTURE_COMPLETION_EXECUTOR;
 import static software.amazon.awssdk.core.client.config.SdkAdvancedClientOption.SIGNER;
 import static software.amazon.awssdk.core.client.config.SdkAdvancedClientOption.USER_AGENT_PREFIX;
@@ -55,8 +57,8 @@ import software.amazon.awssdk.core.internal.http.loader.DefaultSdkAsyncHttpClien
 import software.amazon.awssdk.core.internal.http.loader.DefaultSdkHttpClientBuilder;
 import software.amazon.awssdk.core.internal.util.UserAgentUtils;
 import software.amazon.awssdk.core.retry.RetryPolicy;
-import software.amazon.awssdk.http.ExecuteRequest;
-import software.amazon.awssdk.http.InvokeableHttpRequest;
+import software.amazon.awssdk.http.ExecutableHttpRequest;
+import software.amazon.awssdk.http.HttpExecuteRequest;
 import software.amazon.awssdk.http.SdkHttpClient;
 import software.amazon.awssdk.http.async.AsyncExecuteRequest;
 import software.amazon.awssdk.http.async.SdkAsyncHttpClient;
@@ -84,6 +86,7 @@ import software.amazon.awssdk.utils.Validate;
  */
 @SdkProtectedApi
 public abstract class SdkDefaultClientBuilder<B extends SdkClientBuilder<B, C>, C> implements SdkClientBuilder<B, C> {
+
     private static final SdkHttpClient.Builder DEFAULT_HTTP_CLIENT_BUILDER = new DefaultSdkHttpClientBuilder();
     private static final SdkAsyncHttpClient.Builder DEFAULT_ASYNC_HTTP_CLIENT_BUILDER = new DefaultSdkAsyncHttpClientBuilder();
 
@@ -204,6 +207,7 @@ public abstract class SdkDefaultClientBuilder<B extends SdkClientBuilder<B, C>, 
     private SdkClientConfiguration finalizeSyncConfiguration(SdkClientConfiguration config) {
         return config.toBuilder()
                      .option(SdkClientOption.SYNC_HTTP_CLIENT, resolveSyncHttpClient(config))
+                     .option(SdkClientOption.CLIENT_TYPE, SYNC)
                      .build();
     }
 
@@ -214,6 +218,7 @@ public abstract class SdkDefaultClientBuilder<B extends SdkClientBuilder<B, C>, 
         return config.toBuilder()
                      .option(FUTURE_COMPLETION_EXECUTOR, resolveAsyncFutureCompletionExecutor(config))
                      .option(ASYNC_HTTP_CLIENT, resolveAsyncHttpClient(config))
+                     .option(SdkClientOption.CLIENT_TYPE, ASYNC)
                      .build();
     }
 
@@ -365,7 +370,7 @@ public abstract class SdkDefaultClientBuilder<B extends SdkClientBuilder<B, C>, 
         }
 
         @Override
-        public InvokeableHttpRequest prepareRequest(ExecuteRequest request) {
+        public ExecutableHttpRequest prepareRequest(HttpExecuteRequest request) {
             return delegate.prepareRequest(request);
         }
 
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/client/handler/ClientExecutionParams.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/client/handler/ClientExecutionParams.java
@@ -23,6 +23,7 @@ import software.amazon.awssdk.core.async.AsyncRequestBody;
 import software.amazon.awssdk.core.exception.SdkException;
 import software.amazon.awssdk.core.http.HttpResponseHandler;
 import software.amazon.awssdk.core.runtime.transform.Marshaller;
+import software.amazon.awssdk.core.sync.RequestBody;
 
 /**
  * Encapsulates parameters needed for a particular API call. Captures input and output pojo types.
@@ -36,6 +37,7 @@ import software.amazon.awssdk.core.runtime.transform.Marshaller;
 public final class ClientExecutionParams<InputT extends SdkRequest, OutputT> {
 
     private InputT input;
+    private RequestBody requestBody;
     private AsyncRequestBody asyncRequestBody;
     private Marshaller<InputT> marshaller;
     private HttpResponseHandler<OutputT> responseHandler;
@@ -77,6 +79,15 @@ public final class ClientExecutionParams<InputT extends SdkRequest, OutputT> {
     public ClientExecutionParams<InputT, OutputT> withErrorResponseHandler(
             HttpResponseHandler<? extends SdkException> errorResponseHandler) {
         this.errorResponseHandler = errorResponseHandler;
+        return this;
+    }
+
+    public RequestBody getRequestBody() {
+        return requestBody;
+    }
+
+    public ClientExecutionParams<InputT, OutputT> withRequestBody(RequestBody requestBody) {
+        this.requestBody = requestBody;
         return this;
     }
 
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/interceptor/Context.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/interceptor/Context.java
@@ -15,14 +15,21 @@
 
 package software.amazon.awssdk.core.interceptor;
 
+import java.io.InputStream;
+import java.nio.ByteBuffer;
 import java.util.Optional;
+import org.reactivestreams.Publisher;
 import software.amazon.awssdk.annotations.SdkProtectedApi;
 import software.amazon.awssdk.annotations.SdkPublicApi;
 import software.amazon.awssdk.annotations.ThreadSafe;
 import software.amazon.awssdk.core.SdkRequest;
 import software.amazon.awssdk.core.SdkResponse;
+import software.amazon.awssdk.core.async.AsyncRequestBody;
+import software.amazon.awssdk.core.sync.RequestBody;
 import software.amazon.awssdk.http.SdkHttpFullRequest;
 import software.amazon.awssdk.http.SdkHttpFullResponse;
+import software.amazon.awssdk.http.SdkHttpRequest;
+import software.amazon.awssdk.http.SdkHttpResponse;
 
 /**
  * A wrapper for the immutable context objects that are visible to the {@link ExecutionInterceptor}s.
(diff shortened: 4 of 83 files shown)


millems:
The sync payload is currently on the HTTP request, and interceptors modify the HTTP request payload by wrapping the contentStreamProvider in modifyHttpRequest.

Why can't we do the same for async?

zoewangg:
@shorea +1 on your first option. Like the idea of splitting sync and async modifyHttpContents methods.

Would it make sense to further split the `Context`?
`Context.ModifyHttpContent` and `Context.ModifyAsyncHttpContent`

For the second option, I feel it could be a bit confusing because the it can get the inputStream from both `toModify` and `context.httpContent()`, even though `context.httpContent()` would be null at that time, but the person who implements this method  might not know.

```java
InputStream modifyHttpContent(InputStream toModify, Context.ModifyHttpContent context, ExecutionAttributes executionAttributes);
```

shorea:
So I was proposing not to have the content in the context at all until after that modifyHttpContent stage, so it's only in one location (the parameter in modifyHttpContent and the next context object in future callbacks).

millems:
As long as sync and async are consistent, I don't mind too much. 

1. If we're going to have sync payloads on the HTTP request, we should have async payloads there as well. 
2. If we use AsyncRequestBody in the interceptor for async, we should use RequestBody for async.

spfink:
Updated PR to include separate content interceptors.

@millems, not sure I follow your latest comment. Are you saying we should put the response transformer and request body on the SdkHttpFullRequest/Response objects?

core/sdk-core/src/main/java/software/amazon/awssdk/core/client/builder/SdkDefaultClientBuilder.java
```diff
@@ -204,6 +205,7 @@ protected SdkClientConfiguration finalizeChildConfiguration(SdkClientConfigurati
     private SdkClientConfiguration finalizeSyncConfiguration(SdkClientConfiguration config) {
         return config.toBuilder()
                      .option(SdkClientOption.SYNC_HTTP_CLIENT, resolveSyncHttpClient(config))
+                     .option(SdkClientOption.CLIENT_TYPE, "Sync")
```
millems: Enum?

core/sdk-core/src/main/java/software/amazon/awssdk/core/interceptor/Context.java
```diff
@@ -65,19 +69,35 @@ private Context() {}
...
+     */
+    @ThreadSafe
+    @SdkPublicApi
+    public interface ModifyHttpContent extends ModifyHttpRequest {
         /**
          * The {@link SdkHttpFullRequest} that was created as a result of marshalling the {@link #request()}. This is the HTTP
          * request that will be sent to the downstream service.
          */
-        SdkHttpFullRequest httpRequest();
+        SdkHttpRequest httpRequest();
```
millems: Shouldn't this be in AfterMarshalling?

core/sdk-core/src/main/java/software/amazon/awssdk/core/interceptor/Context.java
```diff
@@ -86,6 +106,9 @@ private Context() {}
     @ThreadSafe
     @SdkPublicApi
     public interface BeforeTransmission extends ModifyHttpRequest {
+        RequestBody requestBody();
+
+        AsyncRequestBody asyncRequestBody();
```
millems: Shouldn't this be in AfterMarshalling?

core/sdk-core/src/main/java/software/amazon/awssdk/core/interceptor/Context.java
```diff
@@ -105,15 +144,16 @@ private Context() {}
      */
     @ThreadSafe
     @SdkPublicApi
-    public interface ModifyHttpResponse extends AfterTransmission {
+    public interface ModifyHttpResponseContent extends ModifyAsyncHttpResponseContent {
+        Publisher responsePublisher();
     }
```
millems: No raw generics, please.
Shouldn't this be in AfterTransmision?
Where is the sync payload?

core/sdk-core/src/main/java/software/amazon/awssdk/core/interceptor/ExecutionInterceptor.java
```diff
@@ -15,17 +15,22 @@
...
+import software.amazon.awssdk.core.async.AsyncRequestBody;
 import software.amazon.awssdk.core.retry.RetryPolicy;
+import software.amazon.awssdk.core.sync.RequestBody;
 import software.amazon.awssdk.http.SdkHttpFullRequest;
 import software.amazon.awssdk.http.SdkHttpFullResponse;
+import software.amazon.awssdk.http.SdkHttpRequest;
 
+// Disable CS to avoid "Unused Import" error. If we use the FQCN in the Javadoc, we'll run into line length issues instead.
+// CHECKSTYLE:OFF
+// CHECKSTYLE:ON
```
millems: Remove comments?

core/sdk-core/src/main/java/software/amazon/awssdk/core/interceptor/ExecutionInterceptor.java
```diff
@@ -219,12 +238,49 @@ default void afterTransmission(Context.AfterTransmission context, ExecutionAttri
...
+     * the error code returned by the service is retriable, this will be invoked for each response returned by the service.
+     *
+     * @param context The current state of the execution, including the SDK and HTTP requests and the current HTTP response.
+     * @param executionAttributes A mutable set of attributes scoped to one specific request/response cycle that can be used to
+     * give data to future lifecycle methods.
+     * @return The potentially-modified HTTP response that should be given to the unmarshaller. Must not be null.
+     */
+    default Publisher<ByteBuffer> modifyAsyncHttpResponseContent(Publisher<ByteBuffer> toModify,
+                                                                 Context.ModifyAsyncHttpResponseContent context,
+                                                                 ExecutionAttributes executionAttributes) {
```
millems: The values to be modified aren't usually passed separately. Any reason we're changing that just for these?

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/handlers/GetBucketPolicyInterceptor.java
```diff
@@ -34,18 +34,18 @@
...
-    public SdkHttpFullResponse modifyHttpResponse(Context.ModifyHttpResponse context, ExecutionAttributes executionAttributes) {
+    public RequestBody modifyHttpContent(RequestBody requestBody,
+                                          Context.ModifyHttpContent context,
+                                          ExecutionAttributes executionAttributes) {
         if (context.request() instanceof GetBucketPolicyRequest) {
-            String policy = context.httpResponse().content()
-                                   .map(c -> invokeSafely(() -> IoUtils.toUtf8String(c)))
-                                   .orElse(null);
+
+            String policy = requestBody.contentStreamProvider() == null ? null : invokeSafely(() ->IoUtils.toUtf8String(requestBody.contentStreamProvider().newStream()));
```
millems: run checkstyle

http-client-spi/src/main/java/software/amazon/awssdk/http/ExecuteRequest.java
```diff
@@ -49,20 +56,29 @@ public static Builder builder() {
          * @param request The request.
          * @return This builder for method chaining.
          */
-        Builder request(SdkHttpFullRequest request);
+        Builder request(SdkHttpRequest request);
+
+        Builder contentStreamProvider(Optional<ContentStreamProvider> contentStreamProvider);
```
millems: Don't accept an optional parameter.

http-client-spi/src/main/java/software/amazon/awssdk/http/InvokeableHttpRequest.java
```diff
@@ -26,5 +26,5 @@
 @SdkPublicApi
 public interface InvokeableHttpRequest extends Callable<SdkHttpFullResponse>, Abortable {
     @Override
-    SdkHttpFullResponse call() throws IOException;
+    SdkHttpFullResponse call() throws IOException; // ExecuteResponse - sdkResponse + inputStream
```
millems: TODO

core/sdk-core/src/main/java/software/amazon/awssdk/core/client/builder/SdkDefaultClientBuilder.java
spfink: Will add.

core/sdk-core/src/main/java/software/amazon/awssdk/core/interceptor/Context.java
spfink: See Andrew's comment on composite interceptors.

core/sdk-core/src/main/java/software/amazon/awssdk/core/interceptor/Context.java
spfink: Not if we have them as separate arguments in interceptors.

millems:
@zoewangg @shorea The problem with removing http request from the AfterMarshalling context and http response from the AfterTransmission context, is that it makes afterMarshalling and afterTransmission much less useful. Those methods exist to see those values, so removing them means there's not really any reason to use them (or any way to see the values before they're modified by other interceptors).

If we moved them to the first parameter of afterMarshalling (eg.) it would fix the problem, but three parameters for every method sounds like a lot...

core/sdk-core/src/main/java/software/amazon/awssdk/core/interceptor/Context.java
```diff
@@ -98,6 +124,10 @@ private Context() {}
          * The HTTP response returned by the service with which the SDK is communicating.
          */
         SdkHttpFullResponse httpResponse();
+
+        Publisher<ByteBuffer> responsePublisher();
+
+        InputStream responseBody();
```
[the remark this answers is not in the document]
spfink: We can make these both be Optional but then we'd likely want to make the interceptors that use them return optional as well.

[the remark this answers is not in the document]
millems: That makes sense to me! Good call.

core/sdk-core/src/main/java/software/amazon/awssdk/core/ClientType.java
```diff
@@ -0,0 +1,39 @@
...
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.core;
+
+import software.amazon.awssdk.annotations.SdkPublicApi;
+
+@SdkPublicApi
+public enum ClientType {
```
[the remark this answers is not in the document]
millems: Ping - Javadoc

core/sdk-core/src/main/java/software/amazon/awssdk/core/client/handler/ClientExecutionParams.java
```diff
@@ -80,6 +82,15 @@ public InputT getInput() {
         return this;
     }
 
+    public RequestBody getRequestBody() {
+        return requestBody;
+    }
+
+    public ClientExecutionParams<InputT, OutputT> withRequestBody(RequestBody requestBody) {
```
millems: This is a protected API with with*/get*. Can we fix that really quick?

http-client-spi/src/main/java/software/amazon/awssdk/http/InvokeableHttpRequest.java
millems: Ping TODO

http-client-spi/src/main/java/software/amazon/awssdk/http/SdkHttpFullResponse.java
```diff
@@ -166,5 +167,8 @@ default Builder putHeader(String headerName, String headerValue) {
          * This is usually done by closing the service connection.</p>
          */
         Builder content(AbortableInputStream content);
+
```
millems: Are all the methods on this builder still needed?

http-clients/apache-client/src/main/java/software/amazon/awssdk/http/apache/internal/impl/ApacheHttpRequestFactory.java
```diff
@@ -122,9 +122,9 @@ private HttpRequestBase wrapEntity(SdkHttpFullRequest request,
          * preparation for the retry. Eventually, these wrappers would
          * return incorrect validation result.
          */
-        if (request.contentStreamProvider().isPresent()) {
+        if (request.contentStreamProvider() != null && request.contentStreamProvider().isPresent()) {
```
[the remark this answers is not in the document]
millems: Ping - why is contentStreamProvider null?

http-clients/netty-nio-client/src/test/java/software/amazon/awssdk/http/nio/netty/NettyNioAsyncHttpClientSpiVerificationTest.java
```diff
@@ -136,17 +136,17 @@ private SdkHttpFullRequest createRequest(URI endpoint,
...
-                                 .method(method)
-                                 .encodedPath(resourcePath)
-                                 .applyMutation(b -> params.forEach(b::putRawQueryParameter))
-                                 .applyMutation(b -> {
-                                     b.putHeader("Host", endpoint.getHost());
-                                     if (contentLength != null) {
-                                         b.putHeader("Content-Length", contentLength);
-                                     }
-                                 }).build();
+        return (SdkHttpFullRequest) SdkHttpFullRequest.builder()
```
millems: Why is this (and other similar casts) needed?

services/s3/src/it/java/software/amazon/awssdk/services/s3/GetObjectIntegrationTest.java
```diff
@@ -25,6 +25,7 @@
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.nio.file.Path;
+import org.apache.log4j.BasicConfigurator;
```
millems: Unused import

core/sdk-core/src/main/java/software/amazon/awssdk/core/client/handler/ClientExecutionParams.java
spfink: Will update right before I merge.

http-client-spi/src/main/java/software/amazon/awssdk/http/SdkHttpFullResponse.java
spfink: Yes as discussed.

core/protocols/aws-json-protocol/src/main/java/software/amazon/awssdk/protocols/json/internal/marshall/JsonProtocolMarshaller.java
```diff
@@ -136,7 +136,7 @@ private static JsonMarshallerRegistry createMarshallerRegistry() {
     }
 
     private SdkHttpFullRequest.Builder fillBasicRequestParams(OperationInfo operationInfo) {
-        return ProtocolUtils.createSdkHttpRequest(operationInfo, endpoint)
+        return (SdkHttpFullRequest.Builder) ProtocolUtils.createSdkHttpRequest(operationInfo, endpoint)
```
millems: Needed cast?

http-client-spi/src/main/java/software/amazon/awssdk/http/InvokeableHttpRequest.java
```diff
@@ -26,5 +26,5 @@
 @SdkPublicApi
 public interface InvokeableHttpRequest extends Callable<SdkHttpFullResponse>, Abortable {
     @Override
-    SdkHttpFullResponse call() throws IOException;
+    SdkHttpFullResponse call() throws IOException; // TODO: Change to container instad of SdkHttpFullResponse
```
millems: I meant, can we actually do this?

http-client-spi/src/test/java/software/amazon/awssdk/http/SdkHttpRequestResponseTest.java
```diff
@@ -416,7 +416,7 @@ private SdkHttpFullRequest validRequest() {
     }
 
     private SdkHttpFullRequest.Builder validRequestBuilder() {
-        return SdkHttpFullRequest.builder().protocol("http").host("localhost").method(SdkHttpMethod.GET);
+        return (SdkHttpFullRequest.Builder) SdkHttpFullRequest.builder().protocol("http").host("localhost").method(SdkHttpMethod.GET);
```
millems: Cast still needed?

http-clients/netty-nio-client/src/test/java/software/amazon/awssdk/http/nio/netty/NettyNioAsyncHttpClientWireMockTest.java
```diff
@@ -342,17 +342,17 @@ private SdkHttpFullRequest createRequest(URI uri,
...
-                                 .method(method)
-                                 .encodedPath(resourcePath)
-                                 .applyMutation(b -> params.forEach(b::putRawQueryParameter))
-                                 .applyMutation(b -> {
-                                     b.putHeader("Host", uri.getHost());
-                                     if (contentLength != null) {
-                                         b.putHeader("Content-Length", contentLength);
-                                     }
-                                 }).build();
+        return (SdkHttpFullRequest) SdkHttpFullRequest.builder()
```
millems: Cast still needed?

PR #830, opened 2018-11-16
Translate S3Exception for headObject and headBucket

zoewangg:
- Translate S3Exception for headObject and headBucket via interceptor
Throw `NoSuchKey` exception for headObject and `NoSuchBucket` exception for headBucket for 404 status code.

- Added `modifyException` interceptor method

All S3 integration tests passed

### DIFF
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/AsyncExecutionFailureExceptionReportingStage.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/AsyncExecutionFailureExceptionReportingStage.java
@@ -15,23 +15,20 @@
 
 package software.amazon.awssdk.core.internal.http.pipeline.stages;
 
+import static software.amazon.awssdk.core.internal.http.pipeline.stages.utils.ExceptionReportingUtils.reportFailureToInterceptors;
+
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
-
 import software.amazon.awssdk.annotations.SdkInternalApi;
-import software.amazon.awssdk.core.interceptor.Context;
 import software.amazon.awssdk.core.internal.http.RequestExecutionContext;
 import software.amazon.awssdk.core.internal.http.pipeline.RequestPipeline;
-import software.amazon.awssdk.core.internal.interceptor.DefaultFailedExecutionContext;
 import software.amazon.awssdk.core.internal.util.ThrowableUtils;
 import software.amazon.awssdk.http.SdkHttpFullRequest;
 import software.amazon.awssdk.utils.CompletableFutureUtils;
-import software.amazon.awssdk.utils.Logger;
 
 @SdkInternalApi
-public class AsyncExecutionFailureExceptionReportingStage<OutputT>
+public final class AsyncExecutionFailureExceptionReportingStage<OutputT>
     implements RequestPipeline<SdkHttpFullRequest, CompletableFuture<OutputT>> {
-    private static final Logger log = Logger.loggerFor(AsyncExecutionFailureExceptionReportingStage.class);
 
     private final RequestPipeline<SdkHttpFullRequest, CompletableFuture<OutputT>> wrapped;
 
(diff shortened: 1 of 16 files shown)


core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/AsyncExecutionFailureExceptionReportingStage.java
```diff
@@ -58,19 +59,23 @@ public AsyncExecutionFailureExceptionReportingStage(RequestPipeline<SdkHttpFullR
...
-     * @param failure     The execution failure.
+     * @param failure The execution failure.
      */
-    private static void reportFailureToInterceptors(RequestExecutionContext context, Throwable failure) {
+    private static Throwable reportFailureToInterceptors(RequestExecutionContext context, Throwable failure) {
         try {
             Context.FailedExecution failedContext =
                     new DefaultFailedExecutionContext(context.executionContext().interceptorContext(), failure);
             context.interceptorChain().onExecutionFailure(failedContext, context.executionAttributes());
+        } catch (SdkServiceException e) {
```
millems: Should we go all-out and propagate `SdkException`? All of the exceptions? Do we have similar behavior for the other on* methods, that suppresses exceptions that are thrown?

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/handlers/ExceptionTranslationInterceptor.java
```diff
@@ -0,0 +1,92 @@
...
+import software.amazon.awssdk.core.interceptor.ExecutionInterceptor;
+import software.amazon.awssdk.services.s3.model.HeadBucketRequest;
+import software.amazon.awssdk.services.s3.model.HeadObjectRequest;
+import software.amazon.awssdk.services.s3.model.NoSuchBucketException;
+import software.amazon.awssdk.services.s3.model.NoSuchKeyException;
+import software.amazon.awssdk.services.s3.model.S3Exception;
+
+/**
+ * Translate S3Exception for the API calls that do not contain the detailed error code.
+ */
```
millems: Long-term, should we push these upstream to s3?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/AsyncExecutionFailureExceptionReportingStage.java
dagnir: The swallowing behavior is only `onExecuteFailure` since at this point there's already an `Exception` thrown higher up. The other `on*` methods let the exception propagate

https://github.com/aws/aws-sdk-java-v2/blob/fee061bf19d35435b42df3a53692757e037dc057/core/sdk-core/src/test/java/software/amazon/awssdk/core/client/AsyncClientHandlerInterceptorExceptionTest.java#L142-L160

zoewangg: I'm fine with propagating `SdkException`. @shorea  Do you have any concerns?

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/handlers/ExceptionTranslationInterceptor.java
zoewangg: Yeah, we should.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/AsyncExecutionFailureExceptionReportingStage.java
shorea: Should we add another interceptor method that lets you modify the exception so you don't have to throw?

millems: Andrew coming in with the BIG ideas. I like it!

zoewangg: Added `modifyException` interceptor method.

PR #831, opened 2018-11-16
Allow getting to metadata from a region object. Small method renames for consistency.

millems:
This: 
1. Adds Region.metadata() to get the RegionMetadata for the region.
2. Changes RegionMetadata.partition() to return PartitionMetadata instead of the partition name.
3. Changes RegionMetadata.name() to RegionMetadata.id() to match Region.id().
4. Changes PartitionMetadata.partitionName() to PartitionMetadata.name() to match ParitionMetadata.id().


services/s3/src/it/java/software/amazon/awssdk/services/s3/BlogPost.java
```diff
@@ -0,0 +1,165 @@
...
+import software.amazon.awssdk.core.sync.RequestBody;
+import software.amazon.awssdk.core.sync.ResponseTransformer;
+import software.amazon.awssdk.services.s3.model.CreateBucketRequest;
+import software.amazon.awssdk.services.s3.model.GetObjectRequest;
+import software.amazon.awssdk.services.s3.model.GetObjectResponse;
+import software.amazon.awssdk.services.s3.model.PutObjectRequest;
+import software.amazon.awssdk.services.s3.model.PutObjectResponse;
+import software.amazon.awssdk.utils.StringInputStream;
+
+public class BlogPost {
```
shorea: Did you mean to commit this?

millems: Good catch. Removed!

PR #832, opened 2018-11-16
Add test for abortMultipartUpload

zoewangg:
Add test for abortMultipartUpload
tested locally

### DIFF
--- a/services/s3/src/it/java/software/amazon/awssdk/services/s3/UploadMultiplePartIntegrationTest.java
+++ b/services/s3/src/it/java/software/amazon/awssdk/services/s3/UploadMultiplePartIntegrationTest.java
@@ -18,21 +18,29 @@ package software.amazon.awssdk.services.s3;
 import static org.assertj.core.api.Assertions.assertThat;
 import static software.amazon.awssdk.testutils.service.S3BucketUtils.temporaryBucketName;
 
+import java.util.ArrayList;
+import java.util.List;
+import org.apache.commons.lang3.RandomStringUtils;
 import org.junit.AfterClass;
 import org.junit.BeforeClass;
 import org.junit.Test;
+import software.amazon.awssdk.core.ResponseBytes;
 import software.amazon.awssdk.core.sync.RequestBody;
+import software.amazon.awssdk.core.sync.ResponseTransformer;
+import software.amazon.awssdk.services.s3.model.AbortMultipartUploadResponse;
 import software.amazon.awssdk.services.s3.model.CompleteMultipartUploadResponse;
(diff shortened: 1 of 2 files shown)


services/s3/src/it/java/software/amazon/awssdk/services/s3/UploadMultiplePartIntegrationTest.java
```diff
@@ -45,26 +52,89 @@ public static void tearDownFixture() {
...
+
+        List<MultipartUpload> uploads = listMultipartUploadsResponse.uploads();
+
+        assertThat(uploads).isEmpty();
+    }
+
+    private void verifyMultipartUploadResult(String key, List<String> contentsToUpload) throws Exception {
+        ResponseBytes<GetObjectResponse> objectAsBytes = s3.getObject(b -> b.bucket(BUCKET).key(key),
+                                                                      ResponseTransformer.toBytes());
+        String appendedString = contentsToUpload.stream().collect(StringBuilder::new, StringBuilder::append,
```
dagnir: nit: We can use https://docs.oracle.com/javase/8/docs/api/java/util/stream/Collectors.html#joining--

services/s3/src/it/java/software/amazon/awssdk/services/s3/UploadMultiplePartIntegrationTest.java
```diff
@@ -45,26 +52,89 @@ public static void tearDownFixture() {
...
+                                                                  StringBuilder::append).toString();
+        assertThat(objectAsBytes.asUtf8String()).isEqualTo(appendedString);
     }
 
+    private List<UploadPartResponse> uploadParts(String key, String uploadId, int partCount, List<String> contentsToUpload) {
+        List<UploadPartResponse> uploadPartResponses = new ArrayList<>();
+
+        for (int i = 0; i < partCount; i++) {
+            String content = RandomStringUtils.randomAscii(1000);
```
dagnir: nit: unless the entire string needs to be random, we can just use the same string for each part

services/s3/src/it/java/software/amazon/awssdk/services/s3/UploadMultiplePartIntegrationTest.java
zoewangg: Intellij tells me we can actually just do `String.join("", contentsToUpload)`╮(╯_╰)╭

services/s3/src/it/java/software/amazon/awssdk/services/s3/UploadMultiplePartIntegrationTest.java
zoewangg: +1

Issue #452, opened 2018-04-04
Rapid concurrent S3 uploads sometimes produce a java.io.IOException: Server failed to send complete response


ochrons:
## Expected Behavior
Uploading thousands of files concurrently to S3

## Current Behavior
Previously we suffered from #202 but that seems to have been fixed in SDK preview-9. The new problem is similar, except it happens quite rarely, once per tens of thousands of uploads. We use multipart uploading for all uploads.

In one test case I was uploading batches of 1000 files each, running 32 uploads concurrently. The error happened on the first file of the first batch and also on the fifth file on the first batch. All other batches uploaded without errors.

Internally our service extracts individual files from the batches and uploads them to S3 using the new async API, so the total number of concurrent requests to S3 is probably quite high.

The error originates from https://github.com/aws/aws-sdk-java-v2/blob/1c8ae1457caf9e45a3c17f672136708fa16e2489/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ResponseHandler.java#L124-L133

## Steps to Reproduce (for bugs)
Upload lots of files to S3 concurrently.

## Your Environment
AWS SDK preview-9
JDK8
AWS Linux
i3.xlarge EC2 instance

dagnir:
This was a quirk we noticed when we originally debugged #202. Basically what would happen is that when connecting to S3 over TLS, on very rare occasions, S3 will suddenly close the connection after the request is sent rather than sending a response. Curiously, it does not seem to happen when connected over unencrypted HTTP.

There are a couple issues that look quite similar that have been reported to the PHP SDK: https://github.com/aws/aws-sdk-php/issues/983

dagnir:
I think it makes sense to treat this as a retryable exception.

bollmann:
We're also getting the same exception every once in a while even with AWS SDK "2.0.0-preview-11". We're using the following asynchronous s3 client in scala:

```scala
val amazonS3 = S3AsyncClient.builder()
      .httpClient(NettyNioAsyncHttpClient)
      .credentialsProvider(StaticCredentialsProvider.create(
        AwsBasicCredentials.create(settings.s3AccessId, settings.s3SecretAccessKey)))
      .region(Region.of(settings.s3Region))
      .build()
```

And making the following requests:

```scala
    val request = GetObjectTaggingRequest.builder()
      .bucket(bucketName)
      .key(keyName)
      .build()

[...]

Issue #669, opened 2018-08-23
AttributeKey names should be namespaced to avoid collisions


dagnir:
We use Netty's `AttributeKey` class, but with fairly generic names: https://github.com/aws/aws-sdk-java-v2/blob/b8f833a9f94131583181c93a0aa09fb4870546ac/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ChannelAttributeKey.java#L31

This could cause issues if customers pull in the Netty client but are also using Netty and `AttributeKey` directly in their application because only one `AttributeKey` instance can be created per name: https://netty.io/4.1/api/io/netty/util/AttributeKey.html#newInstance-java.lang.String-

zoewangg:
Fixed via #834

PR #834, opened 2018-11-17
Not fireExceptionCaught if the channel was released back to the pool

zoewangg:
Not fireExceptionCaught if the channel is released back to the pool to avoid warning messages showing up like the following.

```
2018-11-16 15:13:30,347 [aws-java-sdk-NettyEventLoop-1-3] WARN  io.netty.channel.DefaultChannelPipeline - An exceptionCaught() event was fired, and it reached at the tail of the pipeline. It usually means the last handler in the pipeline did not handle the exception.
software.amazon.awssdk.http.nio.netty.internal.FutureCancelledException: java.io.IOException: Server failed to send complete response"
```

see https://github.com/aws/aws-sdk-java-v2/issues/452

Tested with the the example code customer provided in the above issue: sending s3 request every 60s

### DIFF
--- a/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ChannelAttributeKey.java
+++ b/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ChannelAttributeKey.java
@@ -33,33 +33,43 @@ public final class ChannelAttributeKey {
     /**
      * Future that when a protocol (http/1.1 or h2) has been selected.
      */
-    public static final AttributeKey<CompletableFuture<Protocol>> PROTOCOL_FUTURE = AttributeKey.newInstance("protocolFuture");
+    public static final AttributeKey<CompletableFuture<Protocol>> PROTOCOL_FUTURE = AttributeKey.newInstance(
+        "aws.http.nio.netty.async.protocolFuture");
 
     /**
      * Reference to {@link MultiplexedChannelRecord} which stores information about leased streams for a multiplexed connection.
      */
-    public static final AttributeKey<MultiplexedChannelRecord> CHANNEL_POOL_RECORD =
-        AttributeKey.newInstance("channelPoolRecord");
+    public static final AttributeKey<MultiplexedChannelRecord> CHANNEL_POOL_RECORD = AttributeKey.newInstance(
+        "aws.http.nio.netty.async.channelPoolRecord");
 
     /**
      * Value of the MAX_CONCURRENT_STREAMS from the server's SETTING frame.
      */
-    public static final AttributeKey<Long> MAX_CONCURRENT_STREAMS = AttributeKey.newInstance("maxConcurrentStreams");
+    public static final AttributeKey<Long> MAX_CONCURRENT_STREAMS = AttributeKey.newInstance(
+        "aws.http.nio.netty.async.maxConcurrentStreams");
 
     /**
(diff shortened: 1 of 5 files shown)


dagnir:
:shipit: nice!

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ChannelAttributeKey.java
```diff
@@ -61,6 +61,11 @@
     static final AttributeKey<Long> EXECUTION_ID_KEY = AttributeKey.newInstance(
             "aws.http.nio.netty.async.executionId");
 
+    /**
+     * Whether the channel is still in use
+     */
+    static final AttributeKey<Boolean> IN_USE = AttributeKey.newInstance("inUse");
```
dagnir: We should namespace this (#669)

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/NettyRequestExecutor.java
```diff
@@ -130,6 +133,7 @@ private void configureChannel() {
         channel.attr(EXECUTE_FUTURE_KEY).set(executeFuture);
         channel.attr(REQUEST_CONTEXT_KEY).set(context);
         channel.attr(RESPONSE_COMPLETE_KEY).set(false);
+        channel.attr(IN_USE).set(true);
```
dagnir: Why not just set this in `acquire`?

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ChannelAttributeKey.java
zoewangg: +1, will fix. How about `AWS_JAVA_SDK_XXX`, I will fix others as well while we can

dagnir: the one above is prefixed with `aws.http.nio.netty.async` we can just follow the same or similar pattern

shorea: Seems like others are using aws.http.nio.netty.async.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/NettyRequestExecutor.java
dagnir: Okie doke. Same, here works too

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ChannelAttributeKey.java
zoewangg: Can we still fix other attribute keys now? or it would be backwards incompatible change at this point.

dagnir: These are internal so it should be fine.

PR #845, opened 2018-11-17
Cleanup ReviewBeforeRelease

zoewangg:


utils/src/main/java/software/amazon/awssdk/utils/StringUtils.java
```diff
@@ -107,7 +106,7 @@
  * <p>#ThreadSafe#</p>
  * @see java.lang.String
  */
-@ReviewBeforeRelease("Remove the methods we don't end up using (and we've removed software.amazon.awssdk.core.util.StringUtils).")
+//TODO Remove the methods we don't end up using (and we've removed software.amazon.awssdk.core.util.StringUtils).
```
varunnvs92: we can remove it? I don't see any unused methods

utils/src/main/java/software/amazon/awssdk/utils/Validate.java
```diff
@@ -47,7 +46,7 @@
  * <p>#ThreadSafe#</p>
  * @see java.lang.String#format(String, Object...)
  */
-@ReviewBeforeRelease("Remove the methods we don't end up using (and software.amazon.awssdk.core.util.ValidationUtils).")
+//TODO: Remove the methods we don't end up using (and software.amazon.awssdk.core.util.ValidationUtils).
```
varunnvs92: There is one unused method. Can we remove it now.

utils/src/main/java/software/amazon/awssdk/utils/StringUtils.java
zoewangg: Yeah, will do

utils/src/main/java/software/amazon/awssdk/utils/Validate.java
zoewangg: Yeah, will do

PR #846, opened 2018-11-19
Blacklist bad new simple methods

dagnir:
Blacklist bad simple methods for new operations.


services/appstream/src/main/resources/codegen-resources/customization.config
```diff
@@ -1,5 +1,6 @@
 {
      "blacklistedSimpleMethods" : [
-        "updateFleet"
+        "updateFleet",
+         "describeUserStackAssociations"
```
varunnvs92: minor: extra space

PR #853, opened 2018-11-20
Endpoint Discovery for DynamoDb

spfink:
Supports the endpoint discovery spec for DynamoDb.


codegen/src/test/resources/software/amazon/awssdk/codegen/poet/client/test-endpoint-discovery-async.java
```diff
@@ -0,0 +1,257 @@
...
+                    .createResponseHandler(operationMetadata, TestDiscoveryIdentifiersRequiredResponse::builder);
+
+            HttpResponseHandler<AwsServiceException> errorResponseHandler = createErrorResponseHandler(protocolFactory,
+                    operationMetadata);
+
+            String key = clientConfiguration.option(AwsClientOption.CREDENTIALS_PROVIDER).resolveCredentials().accessKeyId();
+            URI cachedEndpoint = endpointDiscoveryCache.get(key,
+                    testDiscoveryIdentifiersRequiredRequest.endpointDiscoveryRequest(),
+                    clientConfiguration.option(SdkClientOption.ENDPOINT));
+            clientConfiguration.copy(o -> o.option(SdkClientOption.ENDPOINT, cachedEndpoint));
```
varunnvs92: This doesn't change the original configuration. Also we don't want to update the client endpoint right as the new computed endpoint is only valid for that operation.

For endpoint trait, I am passing the hostPrefix to CientExecutionParams and modifying the endpoint in [BaseClientHandler](https://github.com/aws/aws-sdk-java-v2/blob/831f9e9c3ba1ca7b8e8898875d7157131545a28e/core/sdk-core/src/main/java/software/amazon/awssdk/core/client/handler/BaseClientHandler.java#L54). But this is causing issues with path as its being modified multiple times. I am thinking of passing the hostprefix to marshallers and resolve the endpoint there before creating the HttpFullRequest.

spfink: Ah, yes, was on a different version when I was doing this. Will update.

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/client/test-endpoint-discovery-async.java
```diff
@@ -0,0 +1,258 @@
...
+            JsonOperationMetadata operationMetadata = JsonOperationMetadata.builder().hasStreamingSuccessResponse(false)
+                    .isPayloadJson(true).build();
+
+            HttpResponseHandler<TestDiscoveryIdentifiersRequiredResponse> responseHandler = protocolFactory
+                    .createResponseHandler(operationMetadata, TestDiscoveryIdentifiersRequiredResponse::builder);
+
+            HttpResponseHandler<AwsServiceException> errorResponseHandler = createErrorResponseHandler(protocolFactory,
+                    operationMetadata);
+
+            String key = clientConfiguration.option(AwsClientOption.CREDENTIALS_PROVIDER).resolveCredentials().accessKeyId();
```
shorea: Would this not take into account request level credentials?

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/client/test-endpoint-discovery-async.java
```diff
@@ -0,0 +1,258 @@
...
+            URI cachedEndpoint = endpointDiscoveryCache.get(key,
```
shorea: This cachedEndpoint doesn't appear to be used. Also for async this seems like it would be kind of blocking, at least for the first call.

core/aws-core/pom.xml
```diff
@@ -33,6 +33,11 @@
             <artifactId>auth</artifactId>
             <version>${awsjavasdk.version}</version>
         </dependency>
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>profiles</artifactId>
```
shorea: Why do we need this dependency?

core/aws-core/src/main/java/software/amazon/awssdk/awscore/endpointdiscovery/EndpointDiscoveryEndpoint.java
```diff
@@ -0,0 +1,90 @@
...
+ */
+
+package software.amazon.awssdk.awscore.endpointdiscovery;
+
+import java.net.URI;
+import java.time.Instant;
+import software.amazon.awssdk.annotations.SdkInternalApi;
+
+@SdkInternalApi
+public class EndpointDiscoveryEndpoint {
```
shorea: Final

core/aws-core/src/main/java/software/amazon/awssdk/awscore/endpointdiscovery/EndpointDiscoveryEndpoint.java
```diff
@@ -0,0 +1,90 @@
...
+
+    public static Builder builder() {
+        return new BuilderImpl();
+    }
+
+    public Builder toBuilder() {
+        return builder().endpoint(endpoint).expirationTime(expirationTime);
+    }
+
+    public interface Builder {
```
shorea: Should these extend any of that SdkBuilder copyable builder stuff?

core/aws-core/src/main/java/software/amazon/awssdk/awscore/endpointdiscovery/EndpointDiscoveryEndpoint.java
```diff
@@ -0,0 +1,90 @@
...
+        }
+
+        @Override
+        public Builder expirationTime(Instant expirationTime) {
+            this.expirationTime = expirationTime;
+            return this;
+        }
+
+        @Override
+        public EndpointDiscoveryEndpoint build() {
```
shorea: NP I like this at the bottom

core/aws-core/src/main/java/software/amazon/awssdk/awscore/endpointdiscovery/EndpointDiscoveryRefreshCache.java
```diff
@@ -0,0 +1,87 @@
...
+            key = key + ":" + request.cacheKey().get();
+        }
+
+        EndpointDiscoveryEndpoint endpoint = cache.get(key);
+
+        if (endpoint == null) {
+            if (request.required()) {
+                return cache.computeIfAbsent(key, k -> discoverEndpoint(request).join()).endpoint();
+            } else {
+                EndpointDiscoveryEndpoint tempEndpoint = EndpointDiscoveryEndpoint.builder()
```
shorea: Don't quite follow this. So if it's not required we return the default endpoint which is valid for sixty seconds. After that time we fetch the discovered endpoint? Would it make more sense to just return the default endpoint and immediately kick off a request to populate the cache with the discovered endpoint?

core/aws-core/src/main/java/software/amazon/awssdk/awscore/endpointdiscovery/EndpointDiscoveryRefreshCache.java
```diff
@@ -0,0 +1,87 @@
...
+                                                                                  .endpoint(defaultEndpoint)
+                                                                                  .expirationTime(Instant.now().plusSeconds(60))
+                                                                                  .build();
+                return cache.computeIfAbsent(key, k -> tempEndpoint).endpoint();
+            }
+        }
+
+        if (endpoint.expirationTime().isBefore(Instant.now())) {
```
shorea: Does this mean it's already expired? If so are we returning the expired endpoint to the customer? Should we attempt to refresh it before it expires?

core/aws-core/src/main/java/software/amazon/awssdk/awscore/endpointdiscovery/EndpointDiscoveryRequest.java
```diff
@@ -0,0 +1,121 @@
...
+        this.identifiers = builder.identifiers;
+        this.cacheKey = builder.cacheKey;
+        this.required = builder.required;
+    }
+
+    public Optional<String> operationName() {
+        return operationName;
+    }
+
+    public Optional<Map<String, String>> identifiers() {
```
shorea: Should we copy and make this  map unmodifiable in constructor?

core/aws-core/src/main/java/software/amazon/awssdk/awscore/endpointdiscovery/providers/EndpointDiscoveryProvider.java
```diff
@@ -0,0 +1,25 @@
...
+
+package software.amazon.awssdk.awscore.endpointdiscovery.providers;
+
+import software.amazon.awssdk.annotations.SdkInternalApi;
+
+@FunctionalInterface
+@SdkInternalApi
+public interface EndpointDiscoveryProvider {
+
+    boolean resolveEndpointDiscovery();
```
shorea: Can you add some docs here? I'm not sure what boolean is supposed to represent, I guess if endpoint discovery is enabled?

codegen/src/main/java/software/amazon/awssdk/codegen/poet/endpointdiscovery/EndpointDiscoveryCacheLoaderGenerator.java
```diff
@@ -0,0 +1,122 @@
...
+import software.amazon.awssdk.awscore.endpointdiscovery.EndpointDiscoveryEndpoint;
+import software.amazon.awssdk.awscore.endpointdiscovery.EndpointDiscoveryRequest;
+import software.amazon.awssdk.codegen.emitters.GeneratorTaskParams;
+import software.amazon.awssdk.codegen.model.intermediate.IntermediateModel;
+import software.amazon.awssdk.codegen.model.intermediate.OperationModel;
+import software.amazon.awssdk.codegen.poet.ClassSpec;
+import software.amazon.awssdk.codegen.poet.PoetExtensions;
+import software.amazon.awssdk.codegen.poet.PoetUtils;
+
+public class EndpointDiscoveryCacheLoaderGenerator implements ClassSpec {
```
shorea: Can you add some code gen tests for this thing?

codegen/src/main/java/software/amazon/awssdk/codegen/poet/model/AwsServiceModel.java
```diff
@@ -366,6 +372,15 @@ private MethodSpec getValueForField() {
         return methodBuilder.build();
     }
 
+    private MethodSpec endpointDiscoveryRequest() {
```
shorea: Do we need to put this on the service model? Seems like an odd place for it.

core/aws-core/src/main/java/software/amazon/awssdk/awscore/endpointdiscovery/EndpointDiscoveryRefreshCache.java
spfink: Expired endpoints can be used until a 421 status code or InvalidEndpointException is thrown.

Once an endpoint is expired, we extend the expiration to 60 seconds and immediately attempt a refresh. We extend the expiration 60 seconds so that we aren't refreshing every time.

shorea: Hmm gotcha. So expired means will become invalid soon?

core/aws-core/src/main/java/software/amazon/awssdk/awscore/endpointdiscovery/EndpointDiscoveryRequest.java
spfink: We certainly could but not sure if we need to given this is an internal use only class.

codegen/src/main/java/software/amazon/awssdk/codegen/poet/model/AwsServiceModel.java
spfink: Any alternative suggestions?

If not on the generated models for the requests, we'd have to either generate a separate class per request or generate a list of which fields need to be pulled out for the identifiers for each request.

core/aws-core/pom.xml
spfink: You can enable endpoint discovery through the profile file so I have a provider for it that requires the profiles module.

core/aws-core/src/main/java/software/amazon/awssdk/awscore/endpointdiscovery/EndpointDiscoveryRefreshCache.java
spfink: This is done for two reasons:

1. For new users, there is going to be a small period of creating their DNS entry.

2. In the case of short-lived functions where the client is not being re-used, this helps to prevent an unnecessary call.

core/sdk-core/src/main/java/software/amazon/awssdk/core/endpointdiscovery/EndpointDiscoveryRefreshCache.java
```diff
@@ -0,0 +1,87 @@
...
+                                                                                  .expirationTime(Instant.now().plusSeconds(60))
+                                                                                  .build();
+                return cache.computeIfAbsent(key, k -> tempEndpoint).endpoint();
+            }
+        }
+
+        if (endpoint.expirationTime().isBefore(Instant.now())) {
+            cache.put(key, endpoint.toBuilder().expirationTime(Instant.now().plusSeconds(60)).build());
+            String finalKey = key;
+            discoverEndpoint(null).thenApply(v -> cache.put(finalKey, v));
```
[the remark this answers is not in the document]
spfink: Changed to the request object.

core/sdk-core/src/main/java/software/amazon/awssdk/core/endpointdiscovery/EndpointDiscoveryRequest.java
```diff
@@ -0,0 +1,121 @@
...
+    public static Builder builder() {
+        return new BuilderImpl();
+    }
+
+    @Override
+    public Builder toBuilder() {
+        return new BuilderImpl(this);
+    }
+
+    public interface Builder extends CopyableBuilder<Builder, EndpointDiscoveryRequest> {
```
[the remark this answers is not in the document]
spfink: Will add.

spfink:
@varunnvs92, can't reply to some of your comments for some reason.

In the async client, it would only be blocking if endpoint discovery is required. Otherwise, it will be non-blocking.

PR #883, opened 2018-11-28
Minor refactoring endpoint trait code generation

varunnvs92:
Remove unused hostPrefix local variable in generated clients

### DIFF
--- a/codegen/src/test/resources/software/amazon/awssdk/codegen/poet/client/test-async-client-class.java
+++ b/codegen/src/test/resources/software/amazon/awssdk/codegen/poet/client/test-async-client-class.java
@@ -146,9 +146,10 @@ final class DefaultJsonAsyncClient implements JsonAsyncClient {
     @Override
     public CompletableFuture<APostOperationResponse> aPostOperation(APostOperationRequest aPostOperationRequest) {
         try {
-            String hostPrefix = "{StringMember}-foo.";
             Validate.paramNotBlank(aPostOperationRequest.stringMember(), "StringMember");
-            String resolvedHostExpression = String.format("%s-foo.", aPostOperationRequest.stringMember());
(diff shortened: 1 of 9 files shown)


codegen/src/test/resources/software/amazon/awssdk/codegen/poet/client/test-async-client-class.java
```diff
@@ -146,9 +146,10 @@ public final String serviceName() {
     @Override
     public CompletableFuture<APostOperationResponse> aPostOperation(APostOperationRequest aPostOperationRequest) {
         try {
-            String hostPrefix = "{StringMember}-foo.";
             Validate.paramNotBlank(aPostOperationRequest.stringMember(), "StringMember");
-            String resolvedHostExpression = String.format("%s-foo.", aPostOperationRequest.stringMember());
+            Validate.paramNotBlank(aPostOperationRequest.hostPrefixMember(), "HostPrefixMember");
+            String resolvedHostExpression = "{StringMember}-{HostPrefixMember}.".replace("{StringMember}",
```
millems: String.replace is pretty inefficient, since it compiles a regular expression each time. Can we do something more efficient? Do we have a StringUtils.replace?

varunnvs92: Actually even [replace](http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/java/lang/String.java#l2225) uses regex.

We don't have dependency on Apache, so no `StringUtils.replace`

shorea: Interesting, looks like they fixed it in 9. https://bugs.openjdk.java.net/browse/JDK-8058779 I have 10 locally so that's why I thought it was a scan and replace.

Issue #459, opened 2018-04-12
GetBucketACL creates a side effect that blocks subsequent PutObject request


yzarubin:
S3 `getBucketAcl()` causes (networking?) side effects that blocks subsequent PUT requests. I have not tried out all possible combinations of which calls block other calls, but I was able to consistently reproduce the problem with `getBucketAcl` and `putObject`

```java
import software.amazon.awssdk.core.async.AsyncRequestProvider;
import software.amazon.awssdk.core.auth.AwsCredentials;
import software.amazon.awssdk.core.auth.StaticCredentialsProvider;
import software.amazon.awssdk.core.regions.Region;
import software.amazon.awssdk.services.s3.S3AsyncClient;
import software.amazon.awssdk.services.s3.model.PutObjectRequest;
import software.amazon.awssdk.services.s3.model.PutObjectResponse;

import java.util.concurrent.CompletableFuture;
import java.util.concurrent.ExecutionException;

public class AsyncClientBug2Main {

    private static final String BUCKET_NAME = "";
    private static final String ACCESS = "";
    private static final String SECRET = "";
    private static final String REGION = "";
    private static final String TEST_KEY = "test-key-1";

    private static final S3AsyncClient s3 = S3AsyncClient.builder()
            .credentialsProvider(StaticCredentialsProvider.create(AwsCredentials.create(ACCESS, SECRET)))
            .region(Region.of(REGION))
            .build();

    public static void main(final String[] args) throws InterruptedException, ExecutionException {
        System.out.println("Start of main thread");

        final int iterations = 10;

        for (int i = 0; i < iterations; i++) {
            System.out.println("----------  iteration: " + i + " ---------");

            System.out.println("Start ACL");
            // This request finishes, but creates a race condition that eventually blocks the subsequent PUT
            s3.getBucketAcl(v -> v.bucket(BUCKET_NAME)).join();
            System.out.println("End ACL");

            final PutObjectRequest putReq = PutObjectRequest.builder()
                    .key(TEST_KEY)
                    .bucket(BUCKET_NAME)
                    .build();

            System.out.println("Start PUT");
            final CompletableFuture<PutObjectResponse> x = s3
                    .putObject(putReq, AsyncRequestProvider.fromString("abc"));
            System.out.println("Sent PUT");

            // This call never returns!
            final PutObjectResponse ans = x.join();
            System.out.println("End PUT");
        }

        System.out.println("End of main thread");
    }
}
```
[...]

PR #889, opened 2018-11-30
Fix the issue where putobject requests sometimes is hanging

zoewangg:
## Description
Related to #459 

For `PutObjectRequest`, we are sending `Expect: 100-continue` header and waiting for the service to return "100 Continue" before streaming request. `ChannelOption.AUTO_READ` is disabled, so sometimes it's possible no channel read is triggered to read that message and the request is forever "blocked" waiting for the "100 Continue" response.

The fix is explicitly call `channel.read()` if the request contains "Expect: 100-continue" header

## Testing
Tested with the sample code provided by customer in #459 and verified it worked after the fix.

## Types of changes

## Checklist

## License

### DIFF
--- a/.changes/next-release/bugfix-NettyNIOAsyncHTTPClient-5b20e7b.json
+++ b/.changes/next-release/bugfix-NettyNIOAsyncHTTPClient-5b20e7b.json
@@ -1,0 +1,5 @@
+{
+    "category": "Netty NIO Async HTTP Client", 
+    "type": "bugfix", 
+    "description": "Fix the issue where streaming requests with `Expect: 100-continue` header sometimes are hanging because 100Continue response message is not being read automatically. See [#459](https://github.com/aws/aws-sdk-java-v2/issues/459)"
+}
--- a/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/NettyRequestExecutor.java
+++ b/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/NettyRequestExecutor.java
@@ -182,12 +182,27 @@ public final class NettyRequestExecutor {
                    }
                });
 
-        // FullDuplex calls need to start reading at the same time we make the request.
-        if (context.executeRequest().fullDuplex()) {
+        if (shouldExplicitlyTriggerRead()) {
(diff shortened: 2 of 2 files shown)


.changes/next-release/bugfix-NettyNIOAsyncHTTPClient-5b20e7b.json
```diff
@@ -0,0 +1,5 @@
+{
+    "category": "Netty NIO Async HTTP Client", 
+    "type": "bugfix", 
+    "description": "Fix the issue where PutObjectRequest sometimes is hanging because 100Continue response message is not being read automatically. See [#459]"
```
dagnir: nit: can we make it more general instead of PutObjectRequest in case it also affects other similar operations?

zoewangg: It's only affecting `PutObjectRequest` because we only add `Expect: 100-continue` header to `PutObjectRequest`. 

https://github.com/aws/aws-sdk-java-v2/blob/d314a0d15fb8f80e3653beca3f8c216bf8767d8d/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/handlers/PutObjectInterceptor.java#L35

But yeah I can make it more general

varunnvs92:
https://github.com/aws/aws-sdk-java-v2/blob/d0289dcbcf86bf5c9acc9a6d4301d4fc95188d4e/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/NettyRequestExecutor.java#L171-L177
After this change, we set the ReadTimeoutHandler and call read() 2 times for APIs with expect continue header. Will that be a problem?
We can change line 172 to `if (!shouldExplicitlyTriggerRead())` to be safe and possibly different method name

zoewangg:
@varunnvs92 No, we need to read twice for it.

Here's the flow,

writeAndFlush -> WRITE "expected: continue" -> first explicitly `channel.read()` -> READ "HTTP/1.1 100 Continue" -> WRITE actual request -> second explicitly `channel.read()` after `writeAndFlush` is completed -> READ "200 OK"

varunnvs92:
But both writes/reads are on the same channel, so isn't one channel.read enough? Like for full duplex, we do multiple writes/reads for events but everything is handled by single channel#read. I might be misunderstanding how channel read works.

Issue #56, opened 2017-07-10
Don't use the context class loader when loading HTTP implementations


shorea:
We are relying on the context class loader (the class loader for the current thread) when loading the HTTP implementations. https://github.com/aws/aws-sdk-java-v2/blob/master/core/src/main/java/software/amazon/awssdk/http/loader/SdkServiceLoader.java#L29

We should instead use the class loader that loaded the SDK to preserve the semantics of ordinary classloading. 

```java
ServiceLoader.load(clzz, SdkServiceLoader.class.getClassLoader()) 
```

PR #891, opened 2018-11-30
Update SdkServiceLoader loading

zoewangg:
## Description
Use the class loader that loaded the SDK to load the HTTP implementations. See #56

## Testing
`mvn clean install`

## Types of changes

## Checklist

## License

### DIFF
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/util/ClassLoaderHelper.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/util/ClassLoaderHelper.java
@@ -24,15 +24,17 @@ public final class ClassLoaderHelper {
     }
 
     private static Class<?> loadClassViaClasses(String fqcn, Class<?>[] classes) {
-        if (classes != null) {
-            for (Class<?> c : classes) {
-                ClassLoader loader = c.getClassLoader();
-                if (loader != null) {
-                    try {
-                        return loader.loadClass(fqcn);
-                    } catch (ClassNotFoundException e) {
-                        // move on to try the next class loader
-                    }
+        if (classes == null) {
+            return null;
+        }
+
+        for (Class<?> clzz: classes) {
+            ClassLoader loader = clzz.getClassLoader();
+            if (loader != null) {
+                try {
+                    return loader.loadClass(fqcn);
+                } catch (ClassNotFoundException e) {
+                    // move on to try the next class loader
                 }
             }
         }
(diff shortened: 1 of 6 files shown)


zoewangg:
Interceptors are loaded using `ClassLoaderHelper`. I will update `SdkServiceLoader` to use it as well. Not sure where classloading is used in sts credentials provider though.

 https://github.com/aws/aws-sdk-java-v2/blob/e00b75797b1f94789e6105913396c0c1bed3e258/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/util/ClassLoaderHelper.java#L117

zoewangg:
Updated the `ClassLoader#classLoader` to load the class via parent class loaders rather than the context loader

varunnvs92:
what is motivation for the change? I mean what is advantage of using sdk class loader vs context class loader - performance?

zoewangg:
@varunnvs92  In some environments, loading classes from the thread context class loader will cause issues.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/util/ClassLoaderHelper.java
```diff
@@ -25,22 +25,20 @@ private ClassLoaderHelper() {
 
     private static Class<?> loadClassViaClasses(String fqcn, Class<?>[] classes) {
         if (classes != null) {
```
varunnvs92: null check is redundant as its now done in classLoader method

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/util/ClassLoaderHelper.java
```diff
@@ -25,22 +25,20 @@ private ClassLoaderHelper() {
...
-            for (Class<?> c : classes) {
```
varunnvs92: The behavior seems different. In old code, we would use class loader of each class if loading class doesn't work. But in new code, we only use the loader for the first class that returns non-null getClassLoader.

zoewangg: Oops.. Good catch, will fix!

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/util/ClassLoaderHelper.java
zoewangg: I'd still prefer leaving it here coz we might add other caller methods w/o the null check in the future.

Issue #852, opened 2018-11-20
Create a ApacheHttpClient from a org.apache.http.impl.client.HttpClient


codefromthecrypt:
Here controls the builder used to make the http client. https://github.com/aws/aws-sdk-java-v2/blob/master/http-clients/apache-client/src/main/java/software/amazon/awssdk/http/apache/ApacheHttpClient.java#L117

Apache Http Client is hard to instrument, it would be easier if one could supply the base builder used.

## Expected Behavior
I'd expect to be able to pass an instance of `org.apache.http.impl.client.HttpClientBuilder`

## Current Behavior
The base builder is hard-coded

## Possible Solution
Add DefaultBuilder.setHttpClientBuilder

millems:
What options are you interested in configuring on the `HttpClientBuilder` that aren't currently exposed by the SDK?

If the SDK took an `HttpClientBuilder`, we have three options: (a) override your settings with our own, or (b) override our own with your settings, (c) ignore all other client builder settings.

The first one isn't safe, because it makes it backwards incompatible for us to add new settings. 

The second one is safer, but harder to implement (how do we know what's already been set?). 

The third is probably the safest, but might be the hardest to discover. We would probably be an entirely different HTTP client builder to communicate that the other settings don't matter, and I'm not sure whether there are any settings that we would *have* to override on the client builder to make a successful call.

adriancole:
unfortunately the only way this would work is overriding the base builder as the only way to change what is broken requires subtyping https://github.com/openzipkin/brave/blob/master/instrumentation/httpclient/src/main/java/brave/httpclient/TracingHttpClientBuilder.java

adriancole:
"only way this would work" meaning "only way this would improve things"

adriancole:
I agree that this feature could be mentioned as a use-at-own risk. FWIW in our tracing projects we often put in a test in our test tree to ensure we don't unknowingly mess up something. Knowing is the important part. IOTW, you could have a feature test in your tree that uses brave interceptor. Doing this is extreme, but you'd get some knowledge.

Other ways could be to make an observable client abstraction, but then again you'd have a double-instrumentation problem possibility which can be a worse problem than the one we are discussing which is config setting liability

millems:
[...]

PR #916, opened 2018-12-05
Adding the ability to control how a HttpURLConnection gets created.

kiiadi:
## Description

Customers require the ability to control how a URL connection gets created - but still want to take advantage of other SDK features (such as setting timeout, building the request parameters etc). See https://github.com/aws/aws-toolkit-jetbrains/issues/685 for an example use case.

## Testing

Ran unit tests

## Types of changes

## Checklist

## License

### DIFF
--- a/.changes/next-release/feature-URLConnectionHTTPClient-777a68c.json
+++ b/.changes/next-release/feature-URLConnectionHTTPClient-777a68c.json
@@ -1,0 +1,5 @@
+{
+    "category": "URLConnection HTTP Client", 
+    "type": "feature", 
+    "description": "Adding a hook to enable custom creation of the initial `HttpURLConnection`. This enables customers to control how a connection is established for a given `URL` including handling any required proxy configuration etc."
+}
--- a/http-clients/url-connection-client/src/main/java/software/amazon/awssdk/http/urlconnection/UrlConnectionHttpClient.java
+++ b/http-clients/url-connection-client/src/main/java/software/amazon/awssdk/http/urlconnection/UrlConnectionHttpClient.java
@@ -25,6 +25,7 @@ import static software.amazon.awssdk.utils.NumericUtils.saturatedCast;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.HttpURLConnection;
+import java.net.URI;
 import java.time.Duration;
 import java.util.List;
 import java.util.Map;
@@ -54,13 +55,40 @@ import software.amazon.awssdk.utils.IoUtils;
 public final class UrlConnectionHttpClient implements SdkHttpClient {
 
     private final AttributeMap options;
+    private final UrlConnectionFactory connectionFactory;
 
-    private UrlConnectionHttpClient(AttributeMap options) {
+    private UrlConnectionHttpClient(AttributeMap options, UrlConnectionFactory connectionFactory) {
         this.options = options;
+        if (connectionFactory != null) {
+            this.connectionFactory = connectionFactory;
+        } else {
+            this.connectionFactory = this::createDefaultConnection;
+        }
+
     }
 
     public static Builder builder() {
         return new DefaultBuilder();
+    }
+
+    /**
+     * Create a {@link HttpURLConnection} client with the default properties
+     *
+     * @return an {@link UrlConnectionHttpClient}
+     */
+    public static SdkHttpClient create() {
+        return new DefaultBuilder().build();
+    }
+
+    /**
+     * Use this method if you want to control the way a {@link HttpURLConnection} is created.
+     * This will ignore SDK defaults like {@link SdkHttpConfigurationOption#CONNECTION_TIMEOUT}
+     * and {@link SdkHttpConfigurationOption#READ_TIMEOUT}
+     * @param connectionFactory a function that, given a {@link URI} will create an {@link HttpURLConnection}
+     * @return an {@link UrlConnectionHttpClient}
+     */
+    public static SdkHttpClient create(UrlConnectionFactory connectionFactory) {
+        return new UrlConnectionHttpClient(AttributeMap.empty(), connectionFactory);
     }
 
     @Override
@@ -75,8 +103,7 @@ public final class UrlConnectionHttpClient implements SdkHttpClient {
     }
 
     private HttpURLConnection createAndConfigureConnection(HttpExecuteRequest request) {
-        HttpURLConnection connection =
-            invokeSafely(() -> (HttpURLConnection) request.httpRequest().getUri().toURL().openConnection());
+        HttpURLConnection connection = connectionFactory.createConnection(request.httpRequest().getUri());
         request.httpRequest()
                .headers()
                .forEach((key, values) -> values.forEach(value -> connection.setRequestProperty(key, value)));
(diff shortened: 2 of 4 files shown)


http-clients/url-connection-client/src/main/java/software/amazon/awssdk/http/urlconnection/UrlConnectionHttpClient.java
```diff
@@ -199,6 +210,31 @@ public void setConnectionTimeout(Duration connectionTimeout) {
...
+         * A function used to create a {@link HttpURLConnection} from a {@link URL}.
+         *
+         * By default {@link URL#openConnection()} is used.
+         *
+         * @param connectionCreator the function to call to create the connection
+         * @return this object for method chaining
+         */
+        @Override
+        public Builder connectionCreator(Function<URL, HttpURLConnection> connectionCreator) {
+            connectionCreator = connectionCreator;
```
dagnir: SpotBugs flagged this dead store. Can we fix and add a test case?

millems:
Sorry to do this, but whatever we do here should be consistent with the rest of the other clients. See my concerns here: https://github.com/aws/aws-sdk-java-v2/issues/852#issuecomment-440471794

Can we have a more in-depth discussion on this?

http-clients/url-connection-client/src/main/java/software/amazon/awssdk/http/urlconnection/UrlConnectionHttpClient.java
kiiadi: Once we agree this is the right approach I'm happy to add a test.

kiiadi:
@millems I guess this is slightly different since any mutations you make to the `HttpUrlConnection` would still apply - we're not (and don't want to) change that logic.

Let me state the requirement and we can go from there as to what you guys feel is the best approach:

The JetBrains toolkit allows users to specify proxy connection settings, there is a built in helper that, given a `location` (String) create an `HttpUrlConnection` for it, taking into account any proxy settings as well as any 'ignore' filters - customers can say ignore the proxy for these hosts / URL's etc. We don't really want to have to reimplement all this logic ourselves - nor do we want to have to re-implement all the SDK's logic around creating the request, putting in params, encoding, headers etc.

So I figure we a few options:
1) This PR, expose a hook to override the initial `HttpURLConnection` creation process through the builder and passing a function.
2) Alternate hook, make the class non-`final`, and create a `protected createUrlConnection(URI)` method - we can then extend the class and provide our own implementation of `createUrlConnection`
3) Expose the ability to configure `java.net.Proxy` settings on the SDK - this would need to be able to be done per-request since we'll need to apply the same filtering logic based on the host.

kiiadi:
PS: This is the logic I'd like to hook into when creating the HttpClient https://github.com/JetBrains/intellij-community/blob/master/platform/platform-api/src/com/intellij/util/net/HttpConfigurable.java#L409

http-clients/url-connection-client/src/main/java/software/amazon/awssdk/http/urlconnection/UrlConnectionHttpClient.java
kiiadi: fixed the actual spot-bugs issue - what a noob mistake - been away from Java too long :P

shorea:
I'd be more in favor of 1. Never a fan of subclassing or hooks. Also with 3 we currently don't have a mechanism for per request HTTP configuration. Looking at the code we'd need to be able to handle a list of proxies and find the first one that works, that behavior is a bit odd and not something I'd expect the SDK to do so I'd rather keep it out of our implementation.

kiiadi:
ping - this is in response to a customer request : https://plugins.jetbrains.com/plugin/11349-aws-toolkit#comment=30003

http-clients/url-connection-client/src/main/java/software/amazon/awssdk/http/urlconnection/UrlConnectionHttpClient.java
```diff
@@ -159,10 +162,18 @@ public void abort() {
          * means infinity, and is not recommended.
          */
         Builder connectionTimeout(Duration connectionTimeout);
+
+        /**
+         * A function used to create a {@link HttpURLConnection} from a {@link URL}.
+         *
+         * By default {@link URL#openConnection()} is used.
+         */
+        Builder connectionCreator(Function<URL, HttpURLConnection> connectionCreator);
```
millems: 1. Should we use our own factory interface, instead of a built-in, so that we can document it better?
2. Do we override customer settings with our own? 
3. If so, how do we handle new, SDK-provided settings in the future? They could potentially override customer settings and break backwards-compatibility.

.changes/next-release/feature-URLConnectionHTTPClient-777a68c.json
```diff
@@ -0,0 +1,5 @@
+{
+    "category": "URLConnection HTTP Client", 
+    "type": "feature", 
+    "description": "Adding a hook to enable custom creation of the initial `HttpURLConnection` that is then further configured by the SDK to make requests. This enables customers to control how a connection is established for a given `URL` including handling any required proxy configuration etc."
```
millems: Update

http-clients/url-connection-client/src/main/java/software/amazon/awssdk/http/urlconnection/UrlConnectionCreator.java
```diff
@@ -0,0 +1,34 @@
...
+
+import java.net.HttpURLConnection;
+import java.net.URI;
+
+/**
+ * An interface that, given a {@link URI} creates a new {@link HttpURLConnection}. This allows customization
+ * of the creation and configuration of the {@link HttpURLConnection}.
+ */
+@FunctionalInterface
+public interface UrlConnectionCreator {
```
millems: Factory?

http-clients/url-connection-client/src/test/java/software/amazon/awssdk/http/urlconnection/UrlConnectionHttpClientWithCustomCreateWireMockTest.java
```diff
@@ -0,0 +1,14 @@
+package software.amazon.awssdk.http.urlconnection;
```
zoewangg: missing copyright

http-clients/url-connection-client/src/main/java/software/amazon/awssdk/http/urlconnection/UrlConnectionCreator.java
```diff
@@ -0,0 +1,34 @@
...
+package software.amazon.awssdk.http.urlconnection;
+
+import java.net.HttpURLConnection;
+import java.net.URI;
+
+/**
+ * An interface that, given a {@link URI} creates a new {@link HttpURLConnection}. This allows customization
+ * of the creation and configuration of the {@link HttpURLConnection}.
+ */
+@FunctionalInterface
```
zoewangg: Checkstyle would fail for missing sdk annotation

zoewangg: Oh our checkstyle does not check for interface apparently. :\ Will fix it.
https://github.com/aws/aws-sdk-java-v2/blob/5005b0f940166e8099feb17b06ed85158a1dab0e/build-tools/src/main/java/software/amazon/awssdk/buildtools/checkstyle/MissingSdkAnnotationCheck.java#L45-L47

kiiadi:
Fixed checkstyle and rebased from master

kiiadi:
Is this ready to merge?

Issue #899, opened 2018-12-04
S3 PutObjectRequest doesn't detect .gz files and set content-type 


spazbob:
## Expected Behavior
In the V1 SDK, PutObjectRequest would set the content-type to application/x-gzip if I uploaded a .gz file.

## Current Behavior
In V2, it sets to the default application/octet-stream

## Possible Solution
It looks like this was more broadly looked at in https://github.com/aws/aws-sdk-java-v2/issues/442 and a mime.types file was created, but this has no mapping for gz

## Steps to Reproduce (for bugs)
client.putObject(PutObjectRequest.builder().bucket(awsBucketName).key(key).build(), RequestBody.fromFile(localFile));

(in the above instance, localFile is a Path to a file ending in .gz)

## Your Environment

* AWS Java SDK version used: 2.1.0
* JDK version used: OpenJDK 11.0.1
* Operating System and version: Docker: openjdk:11-jre-slim

varunnvs92:
As per [iana official list](https://www.iana.org/assignments/media-types/media-types.xhtml), the content-type would be `application/gzip`. It differs from v1. I made a put call and it works fine. Do you have any issues with it?

PR #934, opened 2018-12-07
Update mime types to add gz and gzip extensions

varunnvs92:
Fix https://github.com/aws/aws-sdk-java-v2/issues/899

Official list at [iana](https://www.iana.org/assignments/media-types/media-types.xhtml) doesn't have the .gz extension. So hand editing the file to add .gz and .gzip


varunnvs92:
Added changelog and merged

Issue #941, opened 2018-12-09
UrlConnectionHttpClient never works


caiiiycuk:
I am intrested to use UrlConnectionHttpClient in my project, because it startup time is minimal. But it never works even in test examples.
How to reproduce:
Clone repo with examples:
https://github.com/awsdocs/aws-doc-sdk-examples/tree/master/javav2

Open s3 project, and add dependency:
```
    <dependency>
      <groupId>software.amazon.awssdk</groupId>
      <artifactId>url-connection-client</artifactId>
    </dependency>
```

Then in class S3ObjectOperations, configure S3Client to use UrlConnectionHttpClient:
```
s3 = S3Client.builder().region(region).httpClient(UrlConnectionHttpClient.builder().build()).build();
```

Run example S3ObjectOperations. 

## Expected Behavior
Example should run, and finish without any problem

## Current Behavior
Example stucks forever, each time. It never finish for me. It works for example with Netty client.

Logs:
```
[...]

PR #942, opened 2018-12-10
Fix bug in ChecksumCalculatingInputStream

varunnvs92:
Fix https://github.com/aws/aws-sdk-java-v2/issues/941

**Bug Description**
When request has content stream, [UrlConnectionHttpClient](https://github.com/aws/aws-sdk-java-v2/blob/8128c9922ce1e3516a33e8c29f383bfc852fe456/http-clients/url-connection-client/src/main/java/software/amazon/awssdk/http/urlconnection/UrlConnectionHttpClient.java#L109) copies the content stream to the connection output stream using [IoUtils.copy](https://github.com/aws/aws-sdk-java-v2/blob/8128c9922ce1e3516a33e8c29f383bfc852fe456/utils/src/main/java/software/amazon/awssdk/utils/IoUtils.java#L101-L111) method. The copy methods writes to the outputStream until the endOfStream is reached ie., _in.read method returns -1_. As [ChecksumCalculatingInputStream](https://github.com/aws/aws-sdk-java-v2/blob/8128c9922ce1e3516a33e8c29f383bfc852fe456/services/s3/src/main/java/software/amazon/awssdk/services/s3/checksums/ChecksumCalculatingInputStream.java#L81) returns 0 for eos, the copy method goes into infinite loop. 

**Fix:** Make read method return -1 when eos is reached.

**Testing**
Verified the tests in reported [issue](https://github.com/aws/aws-sdk-java-v2/issues/941) passed after the change.
* Its not possible to write tests in s3 module without adding url client module as dependency and forcing us to update all other tests to explicitly pick http client. Instead of that, I would prefer adding a module in [test directory](https://github.com/aws/aws-sdk-java-v2/tree/master/test) to add integ tests for S3 using UrlConnectionHttpClient. What do you think?
* Ideally we would need tests for all InputStream classes, but its not in the scope of the PR

### DIFF
--- a/.changes/next-release/bugfix-AmazonS3-5df5cc4.json
+++ b/.changes/next-release/bugfix-AmazonS3-5df5cc4.json
@@ -1,0 +1,5 @@
+{
+    "category": "Amazon S3", 
+    "type": "bugfix", 
+    "description": "S3 putObject API using UrlConnectionHttpClient goes into infinite loop. See https://github.com/aws/aws-sdk-java-v2/pull/942 for more details."
+}
--- a/http-clients/url-connection-client/pom.xml
+++ b/http-clients/url-connection-client/pom.xml
@@ -55,6 +55,35 @@
             <artifactId>junit</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>s3</artifactId>
+            <version>${awsjavasdk.version}</version>
+            <scope>test</scope>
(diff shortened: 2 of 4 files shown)


spfink:
What's the background for why this is breaking?

zoewangg:
Can we add some tests?

zoewangg:
Like the idea of putting the tests in `http-client-tests` module.

varunnvs92:
Updated the PR with a new commit to fix dependency issues

PR #950, opened 2018-12-13
Await h2 connection releases before closing pool

dagnir:
## Description

We must wait for the connection release to be completed (it's an asynchronous
operation) before closing the pool, otherwise it could be left dangling and lead
to a socket leak.

## Motivation and Context

Fix socket leak.

## Testing

`mvn clean install`

## Types of changes

## Checklist

## License


http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/Http2MultiplexedChannelPool.java
```diff
@@ -150,6 +164,13 @@ private void releaseParentChannel(Channel parentChannel) {
 
     @Override
     public void close() {
+        connections.forEach(c -> {
```
shorea: Should these awaits eventually timeout? Also it may not be a good idea to dispatch the close to the connection pool as it also awaits so that would block a netty thread. I think we can safely move it to the calling thread. What do you think?

dagnir: Not sure about having a timeout, but what about switching to a normal await() so it's interruptible? +1 moving the close on to the calling thread.

shorea: +1 to the interruptible

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/Http2MultiplexedChannelPool.java
```diff
@@ -77,6 +92,10 @@
     }
 
     private Future<Channel> acquire0(Promise<Channel> promise) {
+        if (closed) {
```
shorea: Future improvement, we can make this a decorator to reuse this functionality for HTTP/1.1

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/Http2MultiplexedChannelPool.java
```diff
@@ -124,6 +143,11 @@ private void releaseParentChannel0(Channel parentChannel, MultiplexedChannelReco
     }
 
     private void release0(Channel channel, Promise<Void> promise) {
+        if (closed) {
```
shorea: Should we still let them release after close?

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/Http2MultiplexedChannelPool.java
```diff
@@ -150,7 +174,27 @@ private void releaseParentChannel(Channel parentChannel) {
...
+                    connectionPool.release(f.getNow()).await();
+                }
+            }
+            connectionPool.close();
+        } catch (InterruptedException ie) {
+            throw new RuntimeException(ie);
+        }
     }
 
+    private Promise<Void> setClosedFlag() {
```
shorea: Is there an advantage to doing it this way versus using an AtomicBoolean or something? Just to let the acquires submitted before close called to get a chance to acquire?

dagnir: Yeah, I wanted to make sure acquire and close serialized so it's not possible to close while in the middle of an acquire.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/Http2MultiplexedChannelPool.java
dagnir: Sounds good

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/Http2MultiplexedChannelPool.java
dagnir: Hmm not sure, I think it's better to error out so the caller can just close the channel if it gets rejected otherwise the other option would be to close it for them here but it seems a little non-obvious

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/Http2MultiplexedChannelPool.java
shorea: +1, that makes sense.

Issue #994, opened 2019-01-03
Async response handler creates a new thread for every request on small instances


kevinkarpenske:
In preview-13 and later, async client request execution [uses the default async pool](https://github.com/aws/aws-sdk-java-v2/blob/e87a1db54877bab16d9d025c094273883acf02ba/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/async/SyncResponseHandlerAdapter.java#L79) during request execution. For instances with 2 or fewer physical cores, the default async pool is backed by a `ThreadPerTaskExecutor` which spawns a new thread for every request.

This issue is further exacerbated by the SDK's `XmlDomParser`, which uses a `ThreadLocal` to retain a reference to its underlying XML factory and therefore must initialize the factory with every request.

## Expected Behavior
The SDK should not spawn a new thread for every request, regardless of instance size.

## Current Behavior
The SDK uses the default async pool, which for small instances has undesirable behavior. See: https://bugs.openjdk.java.net/browse/JDK-8213115

Example of a thread and its call stack per jvisualvm:

![screen shot 2019-01-03 at 3 05 31 pm](https://user-images.githubusercontent.com/627810/50666315-159b6900-0f69-11e9-99ef-d912dc0ef3a2.png)

## Possible Solution
The SDK should use same-thread composition (`thenCompose`) to avoid delegation to the JVM's async pool.

If same-thread composition is deemed undesirable, utilize a new or existing (ideally configurable) executor for async composition. The obvious candidate is the existing configurable `FUTURE_COMPLETION_EXECUTOR` (defaults to the `sdk-async-response` pool) given the [forthcoming fix](https://github.com/aws/aws-sdk-java-v2/issues/968). Given that the initial future was likely executed using that pool, this is further argument for simply using `thenCompose`.

A workaround given the current implementation is to force higher parallelism in the common ForkJoinPool using the `java.util.concurrent.ForkJoinPool.common.parallelism` system property.

## Steps to Reproduce (for bugs)
Execute requests using an async client on an instance with 2 or fewer cores (as reported by `Runtime.getRuntime().availableProcessors()`).

## Context
Encountered while running an application in AWS EC2 using a `c4.large` instance type, which has 2 cores.

## Your Environment
* AWS Java SDK version used: 2.1.1
* JDK version used: 1.8.0_191
* Operating System and version: Ubuntu 14.04

millems:
[...]

PR #996, opened 2019-01-08
Use same thread composition in SyncResponseHandlerAdaptor. See #994

zoewangg:
## Motivation and Context
Related to #994 

## Types of changes

## Checklist

## License

### DIFF
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/async/SyncResponseHandlerAdapter.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/async/SyncResponseHandlerAdapter.java
@@ -76,7 +76,7 @@ public final class SyncResponseHandlerAdapter<T> implements TransformingAsyncRes
     @Override
     public CompletableFuture<T> prepare() {
         streamFuture = new CompletableFuture<>();
-        return streamFuture.thenComposeAsync(baos -> {
+        return streamFuture.thenCompose(baos -> {
             ByteArrayInputStream content = new ByteArrayInputStream(baos.toByteArray());
             // Ignore aborts - we already have all of the content.
             AbortableInputStream abortableContent = AbortableInputStream.create(content);


core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/async/SyncResponseHandlerAdapter.java
```diff
@@ -76,7 +76,7 @@ public void onError(Throwable err) {
     @Override
     public CompletableFuture<T> prepare() {
         streamFuture = new CompletableFuture<>();
-        return streamFuture.thenComposeAsync(baos -> {
+        return streamFuture.thenCompose(baos -> {
```
dagnir: Hmm the stream future is completed by thread calling `onComplete` on the subscriber which will be owned by the async HTTP client. If we complete on the same thread we would block the HTTP client's thread while `responseHandler` runs. Should we use the future completion executor?

zoewangg: The `responseHandler` should run fairly fast. I wonder if it's worth delegating to the future completion executor.

dagnir: That would depend on the size and complexity of the response I suppose. A DDB `BatchGetItem` could return up to 16MB and could take a non-negligible time to un marshall for example (though I've never measured it).

zoewangg: Yeah, ec2 response unmarshalling could also take non-negligible time. I'd be interested to see which is more costly, unmarshalling large payload or transitioning task from one thread to another. Need to run some benchmarks.

Looking at page 63 of the CompletableFuture perf presentation slides https://qconsf.com/sf2017/system/files/presentation-slides/cf.pdf, seems like offloading executor would add large overhead for small quick task and small overhead for large tasks that take 100ms.

dagnir: Cool. I don't think this needs to be a blocker. We can go with a synchronous compose for now and offload if we need to (or provide the option); I think large responses are the exception not the norm.

millems: Something CPU-bound _should_ be fine to happen on the event loop threads. It's a decent bit of overhead to pass to another thread.

dagnir: @millems not sure. Ideally the event loop threads should only be handling I/O IMO. For example, if there was only 1 event loop thread, and multiple concurrent requests it shouldn't be blocked too long doing non I/O things for any one request, especially if data is waiting to be sent/received.

millems: Yeah, it's possible. Benchmarking will tell us the true story.

PR #999, opened 2019-01-09
Fix build for windows

zoewangg:
Fixing `EnvironmentVariableHelperTest` for windows.

### DIFF
--- a/buildspecs/build.yml
+++ b/buildspecs/build.yml
@@ -3,7 +3,7 @@ version: 0.2
 phases:
   build:
     commands:
-      - mvn clean install -Dmaven.wagon.httpconnectionManager.maxPerRoute=2
+      - mvn clean install -D maven.wagon.httpconnectionManager.maxPerRoute=2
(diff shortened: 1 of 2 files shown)


buildspecs/build-windows.yml
```diff
@@ -0,0 +1,6 @@
+version: 0.2
+
+phases:
+  build:
+    commands:
+    - mvn clean install
```
millems: why not just build.yml?

zoewangg: `-Dmaven.wagon.httpconnectionManager.maxPerRoute=2` is not working on Windows somehow. Got `Unknown lifecycle phase ".wagon.httpconnectionManager.maxPerRoute=2".` error

dagnir: I believe it's because there's no space between `-D` and the option. Adding a space should fix it

zoewangg: Yeah, adding the space fixed it.

PR #1002, opened 2019-01-09
Updating to use latest node version

spfink:


services/lambda/src/it/java/software/amazon/awssdk/services/lambda/ServiceIntegrationTest.java
```diff
@@ -75,7 +75,7 @@ public void uploadFunction() throws IOException {
                                                                     .code(FunctionCode.builder().zipFile(functionBits).build())
                                                                     .handler("helloworld.handler")
                                                                     .memorySize(128)
-                                                                    .runtime(Runtime.NODEJS4_3)
+                                                                    .runtime(Runtime.NODEJS8_10)
```
zoewangg: Do we know why it's failing?

spfink: The 4.3 runtime is deprecated.

PR #1004, opened 2019-01-09
Fix #958

### LINKED ISSUE #958: S3: Multi-part upload fails signature check
## Expected Behavior
Creating a new multi-part upload should work ok.

## Current Behavior
Creating a new multi-part upload fails signature check
```
software.amazon.awssdk.services.s3.model.S3Exception: The request signature we calculated does not match the signature you provided. Check your key and signing method.
```

## Possible Solution
It seems that `uploads` query parameter is missing from the HTTP request. This might explain why the signature does not match.
```
Sending Request: DefaultSdkHttpFullRequest(httpMethod=POST, protocol=https, host=bucketname.s3.us-east-2.amazonaws.com, port=443, encodedPath=/2d12b1d999734ab9fcbeba621ccc3efce62c5ef4, headers=[amz-sdk-invocation-id, Cache-Control, Content-Length, Content-MD5, Content-Type, User-Agent], queryParameters=[])
```

## Context
Regular get and put object requests work just fine in the same setup, only the multi-part uploads fail.

## Your Environment

* AWS Java SDK version used: 2.1.4
* JDK version used: JDK8
* Operating System and version:  Win10

zoewangg:
## Description
Fix the bug where the query parameters are incorrectly moved to http body.
See #958 

Root Cause:

In `MoveParametersToBodyStage`, it checks if it should move the query parameters to the http body.
https://github.com/aws/aws-sdk-java-v2/blob/e6aa1fcc7a31269ff97c231dc826d1a01161763f/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/MoveParametersToBodyStage.java#L42-L45

For `CreateMultipartRequest`, we have `CreateMultipartUploadRequestInterceptor` to set the empty request body. For async clients, we never set the `contentStreamProvider` if it's ever changed in the interceptors. So the `input.contentStreamProvider() == null` is true and query parameters are moved to the http request body.

## Testing
Added mock tests as well as integ tests.
All s3 integ tests passed. Will run the whole test suite.

## Types of changes

## Checklist

## License


zoewangg:
All integ tests passed.

PR #1019, opened 2019-01-17
Add better-integ-test buildspec 

zoewangg:
## Description
Add `better-integ-test` buildspec to only trigger integ tests that are related to the changed files so that we can make integ tests as part of PR process.

- Changes in `core`, `codegen` -> minimal required integ tests (`s3`, `dynamodb`, `sqs`). I chose those three services because they can cover `xml`, `json`, `query` protocols respectively. This is open to discussion. 

- Changes in `netty-nio-client` -> `kinesis`, `s3`, `netty-nio-client`
- Changes in `url-connection-client` -> `url-connection-client`
- Changes in `apache-client` -> `apache-client`, `s3`
- Changes in individual service changes -> integ test in that service module

We can add more complex logic in the future as needed. 

## Testing
Tested it using CodeBuild and verified the core change in this PR I made triggered the minimal required integ tests. The job took 8 mins.

## License

### DIFF
--- a/scripts/run-integ-test
+++ b/scripts/run-integ-test
@@ -1,0 +1,72 @@
+#!/usr/bin/env python
+"""Run Integ Tests based on the changed files
+
+"""
+from subprocess import call, check_call, Popen, PIPE
+
+# Minimal modules to tests when core changes are detected.
+# s3 - xml, dynamodb - json, sqs - query
+core_modules_to_test = ["s3", "dynamodb", "sqs"]
+
+# Minimal modules to tests when http client changes are detected.
+# s3 - streaming/non streaming, kinesis - h2
+http_modules_to_test = {
+    "apache-client": ["s3", "apache-client"],
+    "netty-nio-client": ["kinesis", "s3", "netty-nio-client"],
+    "url-connection-client": ["url-connection-client"]
+}
+
+def check_diffs():
+    """
+    Retrieve the changed files
+    """
+    ret = Popen(["git", "diff", "HEAD^", "--name-only"], stdout=PIPE)
+
+    diff, stderr = ret.communicate()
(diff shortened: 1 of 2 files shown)


scripts/run-integ-test
```diff
@@ -0,0 +1,63 @@
...
+    Parse the changed file path and get the respective module names
+    """
+    path = file_path.split('/')
+    top_directory = path[0]
+
+    if top_directory in ["core", "http-clients", "codegen"]:
+        return minimum_modules_to_test
+    elif top_directory== "services":
+        return path[1]
+
```
dagnir: There's no `else` clause so this can return `None`. Is that expected?

scripts/run-integ-test
```diff
@@ -0,0 +1,63 @@
...
+    check_call(["mvn", "verify", "-pl", modules_to_include, "-P", "integration-tests", "-Dfailsafe.rerunFailingTestsCount=1"])
+
+if __name__ == "__main__":
+    diffs = check_diffs()
+    modules = set()
+    for d in diffs:
+        module = get_modules(d)
+        if isinstance(module, list):
+            modules.update(module)
+        elif module: modules.add(module)
```
dagnir: nit: weird formatting here

scripts/run-integ-test
```diff
@@ -0,0 +1,63 @@
...
+
+    if modules:
+        run_tests(modules)
+    else:  print("No modules configured to run. Skipping integ tests")
```
dagnir: nit: weird formatting

scripts/run-integ-test
zoewangg: Yes, this is expected. We don't need to run integ tests for changes in other modules such as `build-tools`

PR #1034, opened 2019-01-24
Add more tests to verify connection closure and small code clean up

zoewangg:
## Description
Reviewed all use of connection closure and verified the resources are being properly released/closed/shutdown after each request as well as after invoke `client#close`.

This PR added more tests to verify the correctness and did some small code cleanup.

### Per Request

- Sync  path:

Sockets are being closed by closing the underlying `InputStream` via [HandleResponseStage#closeInputStreamIfNeeded](https://github.com/aws/aws-sdk-java-v2/blob/7eb85a23ca9e3df49989d605c60b519828ef1ff1/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/HandleResponseStage.java#L62)  unless `needsConnectionLeftOpen` is true (default to false).

For streaming requests, if a customized `ResponseHandler` is provided with `needsConnectionLeftOpen` overriden to true, the underlying connection will not be closed.

- Async path:

Channels are being released after each request in netty module unless `keep alive` is true via [ResponseHandler#finalizeRequest](https://github.com/aws/aws-sdk-java-v2/blob/7eb85a23ca9e3df49989d605c60b519828ef1ff1/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ResponseHandler.java#L107) (I renamed it to `finalizeResponse` in this PR)

### Per Client After invoking `Client#close`

- SDK
   - the thread pools will be shutdown and all `Closable` will be closed via `AmazonSyncHttpClient#close` and `AmazonAsyncHttpClient#close`
   - the underlying http client will be closed unless it's a customer provided client.

- ApacheHttpClient:
`HttpClientConnectionManager` will be shutdown and `IdleConnectionReaper` will be cleaned up.

- NettyNioAyncHttpClient
`EventLoopGroup` will be shutdown and channel pools will be closed

- UrlConnectionHttpClient
No op. Nothing to close

Let me know if I missed anything.

## License

### DIFF
--- a/test/protocol-tests/src/test/java/software/amazon/awssdk/protocol/tests/connection/SyncClientConnectionTest.java
+++ b/test/protocol-tests/src/test/java/software/amazon/awssdk/protocol/tests/connection/SyncClientConnectionTest.java
@@ -1,0 +1,154 @@
+/*
+ * Copyright 2010-2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.protocol.tests.connection;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+import java.io.IOException;
+import java.io.InputStream;
+import org.junit.Before;
+import org.junit.Test;
+import software.amazon.awssdk.auth.credentials.AwsBasicCredentials;
+import software.amazon.awssdk.auth.credentials.StaticCredentialsProvider;
+import software.amazon.awssdk.core.ResponseInputStream;
+import software.amazon.awssdk.core.sync.RequestBody;
+import software.amazon.awssdk.core.sync.ResponseTransformer;
+import software.amazon.awssdk.http.AbortableInputStream;
+import software.amazon.awssdk.http.HttpExecuteResponse;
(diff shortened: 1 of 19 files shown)


test/protocol-tests/src/test/java/software/amazon/awssdk/protocol/tests/connection/SyncClientConnectionTest.java
```diff
@@ -0,0 +1,140 @@
...
+
+    @Test
+    public void streamingOut_connectionLeftOpen_shouldNotCloseStream() {
+        ClosableStringInputStream inputStream = new ClosableStringInputStream("{}");
+        mockHttpClient.stubNextResponse(mockResponse(inputStream, 200));
+
+        client.streamingOutputOperation(b -> b.build(), new ResponseTransformer<StreamingOutputOperationResponse, Object>() {
+
+            @Override
+            public Object transform(StreamingOutputOperationResponse response, AbortableInputStream inputStream) throws Exception {
```
dagnir: What do you think about adding a similar test where we return the `inputStream`, and then call `close()` and then check to see that the wrapped `InputStream` is closed?

We could look into a similar test for async

test/protocol-tests/src/test/java/software/amazon/awssdk/protocol/tests/util/ClosableStringInputStream.java
```diff
@@ -0,0 +1,45 @@
...
+    private boolean isClosed;
+
+    public ClosableStringInputStream(String s) {
+        super(s.getBytes(StandardCharsets.UTF_8));
+        this.string = s;
+    }
+
+    @Override
+    public void close() {
+        isClosed = true;
```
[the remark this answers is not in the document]
zoewangg: Per javadoc, `ByteArrayInputStream#close` has no effect.

>Closing a {@code ByteArrayInputStream} has no effect.

test/protocol-tests/src/test/java/software/amazon/awssdk/protocol/tests/connection/SyncClientConnectionTest.java
zoewangg: +1, will add that test.

I thought about adding a similar test for async but unlike sync where connections are being closed by closing the inputStream, channels are being released in the netty module so there's no easy way to verify that here.

PR #1044, opened 2019-01-29
Fix the deadlock issue in event streaming operations

zoewangg:
## Description
Fix the deadlock issue in event streaming operations triggered in an edge case where `request` and `onComplete` are signalled at the same time.

## Testing
Added unit test. `Kinesis` integ tests passed

## Types of changes

## Checklist

## License

### DIFF
--- a/.changes/next-release/bugfix-AWSSDKforJavav2-6e2fe73.json
+++ b/.changes/next-release/bugfix-AWSSDKforJavav2-6e2fe73.json
@@ -1,0 +1,5 @@
+{
+    "category": "AWS SDK for Java v2", 
+    "type": "bugfix", 
+    "description": "Fix the deadlock issue in `EventStreamAsyncResponseTransformer` for event streaming operations triggered in an edge case where customer subscriber signals `Subscription#request` the same time as `SdkPublisher` signals `Subscriber#onComplete`"
+}
--- a/core/aws-core/src/main/java/software/amazon/awssdk/awscore/eventstream/EventStreamAsyncResponseTransformer.java
+++ b/core/aws-core/src/main/java/software/amazon/awssdk/awscore/eventstream/EventStreamAsyncResponseTransformer.java
@@ -240,6 +240,11 @@ public class EventStreamAsyncResponseTransformer<ResponseT, EventT>
      */
     private void onEventComplete() {
         synchronized (this) {
+            // No op if it's already done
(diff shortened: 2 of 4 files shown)


core/aws-core/src/main/java/software/amazon/awssdk/awscore/eventstream/EventStreamAsyncResponseTransformer.java
```diff
@@ -462,25 +467,51 @@ private void drainEvents() {
...
+
             if (eventsToDeliver.isEmpty() || remainingDemand.get() == 0) {
                 isDelivering.compareAndSet(true, false);
                 // If we still have demand to fulfill then request more if we aren't already requesting
                 if (remainingDemand.get() > 0) {
                     requestDataIfNotAlready();
                 }
             } else {
+                // It's possible a request comes in and onEventComplete is still in progress
+                if (eventsToDeliver.peek() == ON_COMPLETE_EVENT) {
```
dagnir: Not sure I understand why we check again here since we already do this at the top of the method.

dagnir: `ON_COMPLETE_EVENT` is never removed from the list though right?  If a `request` comes in while `onEventComplete` is still executing, wouldn't this just early exit again?

zoewangg: Ahh you are right. Was being too cautious. I will remove it

.changes/next-release/bugfix-AWSSDKforJavav2-6e2fe73.json
```diff
@@ -0,0 +1,5 @@
+{
+    "category": "AWS SDK for Java v2", 
+    "type": "bugfix", 
+    "description": "Fix the deadlock issue in event streaming operations triggered in an edge case where `request` and `onComplete` are signaled at the same time."
```
dagnir: nit: Can we add the interface names to the methods so it's a little clearer? It could be confusing since `EventStreamResponseHandler` has as similarly named `complete()` method

zoewangg: +1 Updated

PR #1055, opened 2019-02-01
Updated connection lifecycle behavior.

### LINKED ISSUE #856: Add more connection TTL options to the Netty HTTP client
Missing feature matrix:

Setting | 1.11.x | 2.0 (Sync, Apache) | 2.0 (Async, Netty)
-- | -- | -- | --
Connection TTL | clientConfig.setConnectionTTL(...)clientConfig.withConnectionTTL(...) | httpClientBuilder.connectionTimeToLive(...) | Not Supported
Connection Max Idle | clientConfig.setConnectionMaxIdleMillis(...)clientConfig.withConnectionMaxIdleMillis(...) | httpClientBuilder.connectionMaxIdleTime(...) | Not Supported
Validate After Inactivity | clientConfig.setValidateAfterInactivityMillis(...)clientConfig.withValidateAfterInactivityMillis(...) | Not Supported | Not Supported
Connection Reaper | clientConfig.setUseReaper(...)clientConfig.withReaper(...) | httpClientBuilder.useIdleConnectionReaper(...) | Not Supported

millems:
1. Added support for `useIdleConnectionReaper`, `connectionTimeToLive` and `connectionMaxIdleTime` to the Netty HTTP client.
2. Enable the idle connection reaper by default for apache and netty clients. Connection time-to-live remains disabled by default.

Fixes #856


http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/OldConnectionReaperHandler.java
```diff
@@ -0,0 +1,97 @@
...
+
+    private void destroy() {
+        if (channelKiller != null) {
+            channelKiller.cancel(false);
+            channelKiller = null;
+        }
+    }
+
+    private void closeChannel(ChannelHandlerContext ctx) {
+        assert ctx.executor().inEventLoop();
```
dagnir: Think we can avoid this by scheduling this within the event loop instead of the executor

```java
ctx.channel().eventLoop()
```

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/HonorCloseOnReleaseChannelPool.java
```diff
@@ -0,0 +1,77 @@
...
+    public Future<Void> release(Channel channel, Promise<Void> promise) {
+        doInEventLoop(channel.eventLoop(), () -> {
+            boolean shouldCloseOnRelease = Boolean.TRUE.equals(channel.attr(ChannelAttributeKey.CLOSE_ON_RELEASE).get());
+
+            if (shouldCloseOnRelease && channel.isOpen() && !channel.eventLoop().isShuttingDown()) {
+                log.debug(() -> "Closing connection (" + channel.id() + "), instead of releasing it.");
+                channel.close();
+            }
+
+            delegatePool.release(channel, promise);
```
zoewangg: If the channel is closed on line 65, do we still do `delegatePool.release(channel, promise)`?

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/HonorCloseOnReleaseChannelPool.java
```diff
@@ -0,0 +1,77 @@
...
+
+/**
+ * Wrap a channel pool so that {@link ChannelAttributeKey#CLOSE_ON_RELEASE} is honored when a channel is released to the
+ * underlying pool.
+ *
+ * When a channel is released and {@link ChannelAttributeKey#CLOSE_ON_RELEASE} is true on the channel, the channel will be closed
+ * before it is released to the underlying pool.
+ */
+@SdkInternalApi
+public class HonorCloseOnReleaseChannelPool implements ChannelPool {
```
zoewangg: Can we add some unit tests for this class? Let's add a test class for every new class :)

http-client-spi/src/main/java/software/amazon/awssdk/http/SdkHttpConfigurationOption.java
```diff
@@ -77,10 +91,19 @@
     public static final SdkHttpConfigurationOption<Boolean> TRUST_ALL_CERTIFICATES =
             new SdkHttpConfigurationOption<>("TrustAllCertificates", Boolean.class);
 
+    /**
+     * Whether idle connection should be removed after the {@link #CONNECTION_MAX_IDLE_TIMEOUT} has passed.
+     */
+    public static final SdkHttpConfigurationOption<Boolean> REAP_IDLE_CONNECTIONS =
```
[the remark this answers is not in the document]
millems: I was following the naming convention of TRUST_ALL_CERTIFICATES, which we can't change now. I'd rather be consistent.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/IdleConnectionReaperHandler.java
```diff
@@ -0,0 +1,50 @@
...
+import io.netty.handler.timeout.IdleStateHandler;
+import java.util.concurrent.TimeUnit;
+import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.utils.Logger;
+
+/**
+ * A handler that closes unused channels that have not had any traffic on them for a configurable amount of time.
+ */
+@SdkInternalApi
+public class IdleConnectionReaperHandler extends IdleStateHandler {
```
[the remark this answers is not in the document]
millems: Its behavior is tested at build time with the functional tests. 

In my opinion, that is better and sufficient for this class. Direct unit tests would be too coupled to the actual class implementation, and would make it impossible to evolve the class over time without having to go back and update the unit tests.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/HonorCloseOnReleaseChannelPool.java
millems: `ChannelPool#acquire()` documents: `<strong>Its important that an acquired is always released to the pool again, even if the {@link Channel} is explicitly closed..</strong>`

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/HonorCloseOnReleaseChannelPool.java
millems: I disagree. Too many class-level tests for simple classes with a large number of dependencies (like this one) makes the code base too rigid and hard to update. Class-level tests written for this class: (a) wouldn't verify it actually *works*, just that it calls the methods we think we should, and (b) would almost certainly have to be updated if we ever touch the implementation.

This class's behavior is better tested (imo) with the functional tests.

zoewangg: Discussed in person. We agree that unit test class for `IdleConnectionReaperHandler.java ` is not needed since the functional tests would cover it. However, the functional test might not cover this class, so unit tests for this class are needed.  (Fun fight! 🍻)

millems: Added tests for the release-on-close behavior.

http-clients/netty-nio-client/src/test/java/software/amazon/awssdk/http/nio/netty/internal/OldConnectionReaperHandlerTest.java
```diff
@@ -0,0 +1,49 @@
+package software.amazon.awssdk.http.nio.netty.internal;
```
zoewangg: Hmm, seems missing copyright headers on the test classes. (We need to add the checkstyles for test files)

millems: Fixed

Issue #1045, opened 2019-01-29
Infinite Pagination in CloudWatchLogsClient.getLogEventsPaginator()


LukeHornibrookHaventec:
When paginating through CloudWatch Logs events via `CloudWatchLogsClient.getLogEventsPaginator()`, the end of the log stream is not properly detected. Instead, the final page of the result set is continually requested in an infinite loop.

## Expected Behavior
The paginator should stop when the last event in the log stream has been consumed.

## Current Behavior
Once the paginator reaches the end of the log stream, the final request is repeated indefinitely.

The problem is that the GetLogEvents API signals the end of the stream by setting `nextForwardToken` or `nextBackwardToken` to the same as the requested value (and not null, like most other paginated APIs). `GetLogEventsResponseFetcher` uses `PaginatorUtils.isOutputTokenAvailable()` to detect the end of the stream, but this only looks for null or empty values.

## Possible Solution
`GetLogEventsResponseFetcher` may need a custom implementation, since `SyncPageFetcher.hasNextPage()` only provides the previous API response, and not the requested token.

As a side note, we should operate on either `nextBackwardToken` or `nextForwardToken` depending on the direction of pagination (ie `GetLogEventsRequest.startFromHead()`).

## Steps to Reproduce (for bugs)
```java
CloudWatchLogsClient.create()
	.getLogEventsPaginator(x -> {
		x.logGroupName("my-log-group");
		x.logStreamName("my-log-stream");
	})
	.events()
	.forEach(event -> {
		// this will be continually called for the final result page
	});
```

## Your Environment
* AWS Java SDK version used: 2.3.9

varunnvs92:
I can repo the behavior. This is tricky to fix without breaking backwards compatibility as the PageFetcher interfaces only take response object. I am still deciding on the solution. I will keep you updated if there is any progress. Let me know if you have any solutions in mind.

>> As a side note, we should operate on either nextBackwardToken or nextForwardToken depending on the direction of pagination

We will not be able to support this as these auto-pagination methods are generated from [paginators](https://github.com/aws/aws-sdk-java-v2/blob/master/services/cloudwatchlogs/src/main/resources/codegen-resources/paginators-1.json#L42-L46) file. we use the `output_token` to determine the field to put the next token. As this file is based on a spec and generated from service model, it would be lot more effort to modify the spec to support backwards pagination. If we observe more methods with similar behavior, we can add this as a feature request.

varunnvs92:
[...]

PR #1057, opened 2019-02-04
Fix cloudwatch logs GetLogEvents auto-pagination bug

varunnvs92:
* Fix https://github.com/aws/aws-sdk-java-v2/issues/1045
* Uses customization and new custom class spec
* See [SameTokenPaginationApiIterable](https://github.com/aws/aws-sdk-java-v2/blob/936b544148f3f525def2bcc95314c5494834ef54/codegen/src/test/resources/software/amazon/awssdk/codegen/poet/paginators/customizations/SameTokenPaginationApiIterable.java) and [SameTokenPaginationApiPublisher](https://github.com/aws/aws-sdk-java-v2/blob/936b544148f3f525def2bcc95314c5494834ef54/codegen/src/test/resources/software/amazon/awssdk/codegen/poet/paginators/customizations/SameTokenPaginationApiPublisher.java) for the modified generated classes
* Stores the lastToken value in the next page fetcher and checks whether its equal to nextToken value in previous page.


varunnvs92:
Even if we introduce a new constructor param, we still need to change the internal `*Fetcher` class to modify to track `lastToken` and how it is used in `hasNextPage` and `nextPage` method. So instead of adding the `lastToken` member to all generated paginators which will go unused, having a separate class structure for this scenario makes sense

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/paginators/customization.config
```diff
@@ -4,5 +4,11 @@
         "nestedContainers",
         "operationWithNoInputOrOutput"
     ],
-    "verifiedSimpleMethods" : ["paginatedOperationWithResultKey"]
+    "verifiedSimpleMethods" : ["paginatedOperationWithResultKey"],
+    "paginationCustomization": {
```
spfink: Can we be more descriptive on naming for all of these values?

Is there ever a case where sync and async won't have the same feature set?

varunnvs92: >> Is there ever a case where sync and async won't have the same feature set?

No, I can use the same value for both. Initially i was going with the fqcn of the class spec file, then decided to keep it as the class name. I can update to use same value for both sync and async

>> Can we be more descriptive on naming for all of these values?

I put the details in the javadoc of the class spec files. Is the following name better?
`ReturnSameTokenGivenInRequestToIndicateLastPage`

spfink: Was more concerned with just the sync / async keys. Having just the one key is nice now. Probably can go with a shorter name...

LastPageHasPreviousToken?

Fine with whatever really since it's just a customization value.

PR #1060, opened 2019-02-05
Fix an issue where an exception could be raised when configuring the idle connection reaper in the apache HTTP client.

millems:
This change is up for debate, but it doesn't make much sense that zero defaults are applied when a client is built, and will be prone to cause issues in the future where customers might assume some type of default would be applied, like is true for every other build() method in the SDK.


millems:
Looks like we already do this, just elsewhere. Hurray! Easier fix updated.

Issue #295, opened 2017-11-17
AWS Java SDK v2 BOM incorrectly locks versions of test dependencies


wjbuys:
## Expected Behavior
When I declare a dependency on [the BOM for the AWS Java SDK v2](https://mvnrepository.com/artifact/software.amazon.awssdk/bom/2.0.0-preview-5), I expect it to only lock the versions of the AWS SDK's *runtime* dependencies, similarly to how the [v1 SDK BOM](https://mvnrepository.com/artifact/com.amazonaws/aws-java-sdk-bom/1.11.231) did it.

## Current Behavior
However, for some reason, the BOM also declares dependency versions for many transitive dependencies, including the following test dependencies of the SDK:

- info.cukes:cucumber-java:1.2.4 
- info.cukes:cucumber-junit:1.2.4
- info.cukes:cucumber-guice:1.2.4
- org.mockito:mockito-core:1.10.19

This forces any project that pulls in the BOM to also depend on these (outdated) dependencies, unless I explicitly override those versions in a dependency management block too.

## Possible Solution
Remove these libraries from the BOM.

## Context
I tried to add the new SDK's BOM to an existing project, and was surprised that all my tests started failing without any code changes. Turns out it had forced Mockito to the 1.x version, and the tests were depending on the 2.x version.

## Your Environment

* AWS Java SDK version used: https://mvnrepository.com/artifact/software.amazon.awssdk/bom/2.0.0-preview-5 
* JDK version used: Oracle JDK 1.8.0_u151
* Operating System and version: Mac OS X 10.12.4

wjbuys:
Looking at it a bit more, I think the problem is that [the SDK v2's BOM's `pom.xml` imports the parent `pom.xml`](https://github.com/aws/aws-sdk-java-v2/blob/master/bom/pom.xml#L25), which contains [a `dependencyManagement` block that depends on all the problematic libraries](https://github.com/aws/aws-sdk-java-v2/blob/master/pom.xml#L127). In SDK v1, the parent pom did not declare all its dependency versions in a `dependencyManagement` block, avoiding this problem.

millems:
This should be fixed in the next release. Thanks!

varunnvs92:
Reopening the issue as the fix has been reverted in #444 for releasing new features. I agree the issue has to be fixed. We can discuss the potential options and go with the right solution. Current options:
1) Remove parent element from BOM's pom file and add all the required elements to satisfy maven publish rules
[...]

PR #1066, opened 2019-02-07
Moved SDK dependencies from the root pom.xml to a bom-internal module.

### LINKED ISSUE #1063: software.amazon.awssdk:core not present in bom
I am in the process of migrating from version 1 to version 2 of the SDK. I have successfully migrated our project to use "com.amazonaws:aws-java-sdk-bom:1.11.490" and I am planning to use it side by side with "software.amazon.awssdk:bom:2.3.9" during the migration (not all features being supported in V2 yet).

One of our modules has an explicit dependency on "com.amazonaws:aws-java-sdk-core" that is used for exception handling. I am unable to add the equivalent dependency "software.amazon.awssdk:core" using the bom mechanism. If I explicitly add a dependency on "software.amazon.awssdk:core:2.3.9" it works fine - although it completely defeats the purpose of using a bom for dependency management if I have to add some of them manually.

Upon further inspection, it turns out that the pom of "software.amazon.awssdk:bom" does not declare a dependency for the core module. I find this really puzzling; I do understand that core is a dependency of all of the other included modules but I think there is a legit use case in having a dependency on just core and nothing else. To me, it would make more sense if it was available in the bom.

millems:
This prevents dependencies in the root pom from polluting the bom module. Also added some missing bom dependencies and cleaned up the pom.xml files across the repository.

Fixes #1063 and #295.


varunnvs92:
Did you verify locally that the issues mentioned in the original issues get fixed by this change?

millems:
Verified!

PR #1069, opened 2019-02-08
Awaits EventLoopGroup Shutdown when closing Netty client

zoewangg:
## Description
Awaits `EventLoopGroup#shutdownGracefully` to complete when closing Netty client.

## Motivation and Context
Fun fact: `EventLoopGroup#shutdownGracefully` is also asynchronous.

## Testing
unit tests and manually triggered smart integ tests.

## Types of changes

## Checklist

## License

### DIFF
--- a/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/NettyNioAsyncHttpClient.java
+++ b/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/NettyNioAsyncHttpClient.java
@@ -24,6 +24,9 @@ import static software.amazon.awssdk.http.SdkHttpConfigurationOption.MAX_PENDING
 import static software.amazon.awssdk.http.SdkHttpConfigurationOption.READ_TIMEOUT;
 import static software.amazon.awssdk.http.SdkHttpConfigurationOption.REAP_IDLE_CONNECTIONS;
 import static software.amazon.awssdk.http.SdkHttpConfigurationOption.WRITE_TIMEOUT;
+import static software.amazon.awssdk.http.nio.netty.internal.NettyConfiguration.EVENTLOOP_SHUTDOWN_FUTURE_TIMEOUT_SECONDS;
+import static software.amazon.awssdk.http.nio.netty.internal.NettyConfiguration.EVENTLOOP_SHUTDOWN_QUIET_PERIOD_SECONDS;
+import static software.amazon.awssdk.http.nio.netty.internal.NettyConfiguration.EVENTLOOP_SHUTDOWN_TIMEOUT_SECONDS;
 import static software.amazon.awssdk.utils.FunctionalUtils.invokeSafely;
 import static software.amazon.awssdk.utils.FunctionalUtils.runAndLogError;
 
@@ -40,6 +43,9 @@ import io.netty.handler.ssl.util.InsecureTrustManagerFactory;
 import java.net.URI;
 import java.time.Duration;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicReference;
 import javax.net.ssl.SSLException;
 import javax.net.ssl.TrustManagerFactory;
@@ -212,7 +218,23 @@ public final class NettyNioAsyncHttpClient implements SdkAsyncHttpClient {
     @Override
     public void close() {
         runAndLogError(log, "Unable to close channel pools", pools::close);
-        runAndLogError(log, "Unable to shutdown event loop", sdkEventLoopGroup.eventLoopGroup()::shutdownGracefully);
+        runAndLogError(log, "Unable to shutdown event loop", () ->
+            closeEventLoopUninterruptibly(sdkEventLoopGroup.eventLoopGroup()));
+    }
+
+    private void closeEventLoopUninterruptibly(EventLoopGroup eventLoopGroup) throws ExecutionException {
+        try {
+            eventLoopGroup.shutdownGracefully(EVENTLOOP_SHUTDOWN_QUIET_PERIOD_SECONDS,
+                                              EVENTLOOP_SHUTDOWN_TIMEOUT_SECONDS,
+                                              TimeUnit.SECONDS)
+                          .get(EVENTLOOP_SHUTDOWN_FUTURE_TIMEOUT_SECONDS, TimeUnit.SECONDS);
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            throw new RuntimeException(e);
+        } catch (TimeoutException e) {
+            throw new RuntimeException(String.format("Shutting down Netty EventLoopGroup did not complete within %s seconds",
+                                                     EVENTLOOP_SHUTDOWN_FUTURE_TIMEOUT_SECONDS));
+        }
     }
 
     /**
--- a/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/SharedSdkEventLoopGroup.java
+++ b/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/SharedSdkEventLoopGroup.java
@@ -15,8 +15,11 @@
 
 package software.amazon.awssdk.http.nio.netty.internal;
 
+import static software.amazon.awssdk.http.nio.netty.internal.utils.NettyUtils.SUCCEEDED_FUTURE;
+
 import io.netty.channel.EventLoopGroup;
 import io.netty.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import software.amazon.awssdk.annotations.SdkInternalApi;
 import software.amazon.awssdk.annotations.SdkTestInternalApi;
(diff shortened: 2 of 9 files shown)


http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/SharedSdkEventLoopGroup.java
```diff
@@ -90,8 +95,35 @@ private ReferenceCountingEventLoopGroup(EventLoopGroup delegate) {
             // Only want to decrement the reference the first time it's closed. Shutdown is idempotent and may be
             // called multiple times.
             if (hasBeenClosed.compareAndSet(false, true)) {
-                decrementReference();
+                return decrementReference();
             }
+            return CompletedCloseFuture.COMPLETED_FUTURE;
```
millems: You can use SucceededFuture with a null executor instead of creating your own future implementation.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/utils/NettyUtils.java
```diff
@@ -131,4 +131,18 @@ public static void doInEventLoop(EventExecutor eventExecutor, Runnable runnable,
             promise.setFailure(e);
         }
     }
+
+    /**
+     * Wait interruptibly for the promise to complete.
+     *
+     * @param promise the promise to complete
+     */
+    public static void syncInterruptibly(Future<?> promise) {
```
millems: This isn't that much different than netty's `awaitUninterruptibly()`. Any reason we care about re-throwing the interrupt exception if we're just going to log it?

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/NettyNioAsyncHttpClient.java
```diff
@@ -212,7 +214,8 @@ private SdkEventLoopGroup nonManagedEventLoopGroup(SdkEventLoopGroup eventLoopGr
     @Override
     public void close() {
         runAndLogError(log, "Unable to close channel pools", pools::close);
-        runAndLogError(log, "Unable to shutdown event loop", sdkEventLoopGroup.eventLoopGroup()::shutdownGracefully);
+        Future<?> closePromise = sdkEventLoopGroup.eventLoopGroup().shutdownGracefully();
+        runAndLogError(log, "Unable to shutdown event loop", () -> syncInterruptibly(closePromise));
```
millems: Should we do the acquiring of the closePromise within the runAndLogError as well so that close() won't throw any exceptions?

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/SharedSdkEventLoopGroup.java
zoewangg: Yeah, I considered using that but was too afraid to pass two nulls to the constructor. I wish they have an overloaded constructor. Will update to use that instead.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/utils/NettyUtils.java
zoewangg: Per Javadoc of `awaitUniterrupttibly`,
> Waits for this future to be completed within the specified time limit without interruption.  This method catches an {@link InterruptedException} and discards it silently.

it doesn't re-interrupt the thread and that's why I created this method to re-interrupt the thread.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/NettyNioAsyncHttpClient.java
zoewangg: +1

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/utils/NettyUtils.java
millems: This does re-interrupt the thread, if you look at the implementation.

zoewangg: Ohh yeah, it does. I will update. I was looking at the implementation of an overloaded method of `awaitUninterruptibly` and from first look, it looks like it doesn't re-interrupt the method.
```
    @Override
    public boolean awaitUninterruptibly(long timeout, TimeUnit unit) {
        try {
            return await0(unit.toNanos(timeout), false);
        } catch (InterruptedException e) {
            // Should not be raised at all.
            throw new InternalError();
        }
    }
```

zoewangg:
`ConnectionReaperTest` seems to be stuck with the changes. Taking a look

zoewangg:
Okay, this time `NettyNioAsyncHttpClientWireMockTest` got stuck, though it passed on my machine and JDK10 CodeBuild job somehow.  :(

zoewangg:
Changing from 

`sdkEventLoopGroup.eventLoopGroup().shutdownGracefully().awaitUninterruptibly()` 

to 

`sdkEventLoopGroup.eventLoopGroup().shutdownGracefully().awaitUninterruptibly(3000)` fixed it.

I compared the source code of `DefaultPromise#awaitUninterruptibly` with `DefaultPromise#awaitUninterruptibly(long)` and they are not exactly the same. There is definitely some race condition in `DefaultPromise#awaitUninterruptibly`

zoewangg:
I updated to use `eventLoopGroup.shutdownGracefully().get(3000, TimeUnit.MILLISECONDS)` instead. It would throw exception if `shutdownGracefully` fails.

`DefaultPromise#get` invokes `DefaultPromise#await`. From the testing, `awaitUninterruptibly` is buggy, but `await` is working fine.

I think timeout is very necessary as it's extremely difficult to debug which process is the culprit when the application gets stuck.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/NettyNioAsyncHttpClient.java
```diff
@@ -212,7 +215,18 @@ private SdkEventLoopGroup nonManagedEventLoopGroup(SdkEventLoopGroup eventLoopGr
...
         runAndLogError(log, "Unable to close channel pools", pools::close);
-        runAndLogError(log, "Unable to shutdown event loop", sdkEventLoopGroup.eventLoopGroup()::shutdownGracefully);
+        runAndLogError(log, "Unable to shutdown event loop in 3000 milliseconds", () ->
+            closeEventLoopUninterruptibly(sdkEventLoopGroup.eventLoopGroup()));
+    }
+
+    private void closeEventLoopUninterruptibly(EventLoopGroup eventLoopGroup) throws
+                                                                              TimeoutException, ExecutionException {
+        try {
+            eventLoopGroup.shutdownGracefully().get(3000, TimeUnit.MILLISECONDS);
```
millems: shutdownGracefully has a timeout passed to it as well, that by default is longer than 3 seconds. Should we align these values in some way? Can we also catch errors we think can happen sometimes (eg. TimeoutException) to log a less scary error, if the customer doesn't need to worry about them?

zoewangg: > Should we align these values in some way?

+1, it took me a while to understand the javadoc of `#shutdownGracefully(long, long, TimeUnit)`.  The default timeout is 15 seconds though and I will update it to 15 seconds.

https://github.com/netty/netty/blob/fa6a8cb09c9679468a6c2d912ddfbbe885ee0c08/common/src/main/java/io/netty/util/concurrent/AbstractEventExecutor.java#L37

> Can we also catch errors we think can happen sometimes (eg. TimeoutException) to log a less scary error, if the customer doesn't need to worry about them? 

Hmm, I do think customers need to be aware of the issue though as unclosed eventloop can cause resource leak. What do you think of changing the log level to warn?

zoewangg: Oh, do you mean just catching `TimeoutException` and changing the error message? Yeah, we can do that. I'll update it

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/NettyNioAsyncHttpClient.java
```diff
@@ -212,7 +217,20 @@ private SdkEventLoopGroup nonManagedEventLoopGroup(SdkEventLoopGroup eventLoopGr
...
     public void close() {
         runAndLogError(log, "Unable to close channel pools", pools::close);
-        runAndLogError(log, "Unable to shutdown event loop", sdkEventLoopGroup.eventLoopGroup()::shutdownGracefully);
+        runAndLogError(log, "Unable to shutdown event loop", () ->
+            closeEventLoopUninterruptibly(sdkEventLoopGroup.eventLoopGroup()));
+    }
+
+    private void closeEventLoopUninterruptibly(EventLoopGroup eventLoopGroup) throws ExecutionException {
+        try {
+            eventLoopGroup.shutdownGracefully().get(EVENTLOOP_SHUTDOWN_TIMEOUT_SECONDS, TimeUnit.SECONDS);
```
millems: Should we pass the same(ish) value to shutdownGracefully, to make it clear why this value was chosen? Worried a future dev might not understand why this particular value is chosen and lessen it.

zoewangg: Yup, updated!

PR #1083, opened 2019-02-18
Updated clock skew adjustment logic.

millems:
1. Generic signature errors only adjust for clock skew and retry when they're also associated with a service/client clock skew of over 4 minutes.
2. Generic 403s and 401s will now also adjust for clock skew and retry when they're associated with a service/client clock skew of over 4 minutes.

This fixes an issue where the SDK could be over-retrying on signature errors and under-retrying on clock skew errors not associated with an error code from the service (eg. 403s from S3 HEAD requests).


core/aws-core/src/main/java/software/amazon/awssdk/awscore/exception/AwsServiceException.java
```diff
@@ -67,9 +73,28 @@ public String getMessage() {
...
+            return true;
+        }
+
+        SdkHttpResponse sdkHttpResponse = awsErrorDetails.sdkHttpResponse();
+
+        if (clockSkew == null || sdkHttpResponse == null) {
+            return false;
+        }
+
+        return (AwsErrorCode.POSSIBLE_CLOCK_SKEW_ERROR_CODES.contains(awsErrorDetails.errorCode()) ||
```
zoewangg: Can we break down the conditions so that it's easier to read?

millems: I was on the border about making `AwsErrorCode.isPossibleClockSkewErrorCode(...)`, so I think you've convinced me with this question.

core/sdk-core/src/main/java/software/amazon/awssdk/core/retry/ClockSkew.java
```diff
@@ -0,0 +1,88 @@
...
+import software.amazon.awssdk.core.internal.http.HttpClientDependencies;
+import software.amazon.awssdk.http.SdkHttpResponse;
+import software.amazon.awssdk.utils.DateUtils;
+import software.amazon.awssdk.utils.Logger;
+
+/**
+ * Utility methods for checking and reacting to the current client-side clock being different from the service-side clock.
+ */
+@ThreadSafe
+@SdkProtectedApi
```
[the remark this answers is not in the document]
millems: Unfortunately it's used from `aws-core`.

millems:
There are already integration tests for clock skew errors (eg. `AutoScalingIntegrationTest.testClockSkewAs`)

zoewangg:
Can we move those tests to mock clients though now that we've actually started to consider removing unnecessary service tests?  It'd also be nice to catch any breaking changes from unit tests rather than integ tests.

Does it cover the scenario that clockskew should be updated when retry is disabled?

millems:
Will verify all of this is covered at build time, if it isn't already.

test/protocol-tests/src/test/java/software/amazon/awssdk/protocol/tests/clockskew/ClockSkewAdjustmentTest.java
```diff
@@ -0,0 +1,205 @@
...
+import software.amazon.awssdk.auth.credentials.StaticCredentialsProvider;
+import software.amazon.awssdk.core.SdkGlobalTime;
+import software.amazon.awssdk.core.exception.SdkException;
+import software.amazon.awssdk.regions.Region;
+import software.amazon.awssdk.services.protocoljsonrpc.ProtocolJsonRpcAsyncClient;
+import software.amazon.awssdk.services.protocoljsonrpc.ProtocolJsonRpcClient;
+import software.amazon.awssdk.services.protocoljsonrpc.model.AllTypesResponse;
+import software.amazon.awssdk.utils.DateUtils;
+
+public class ClockSkewAdjustmentTest {
```
zoewangg: Thanks for adding such thorough tests! 👍

PR #1086, opened 2019-02-20
Filtering out invalid regions

spfink:
Filters out any regions that don't match the partition regex or don't match the partition regex with "-fips" at the end, or don't have global in the name.

## Testing
Updated existing tests to reflect removed regions.

## Types of changes

## Checklist

## License


codegen-lite/src/main/java/software/amazon/awssdk/codegen/lite/regions/ServiceMetadataGenerator.java
```diff
@@ -214,16 +221,34 @@ private MethodSpec signingRegion() {
...
+     *
+     * @param regex - Regex for regions in a given partition.
+     * @param region - Region string being checked.
+     * @return true if the region string should be included as a region.
+     */
+    private boolean validRegion(String regex, String region) {
+        return region.matches(regex) ||
+               region.endsWith("-fips") &&
+               region.matches(regex.replace("$", "(-fips)*$")) ||
+               region.contains("global");
```
millems: Can we add parens, methods or variables to make it more obvious what the precedence is here?

codegen-lite/src/main/java/software/amazon/awssdk/codegen/lite/regions/ServiceMetadataGenerator.java
```diff
@@ -214,16 +221,34 @@ private MethodSpec signingRegion() {
...
+     * and match the partition regex with that included, or include the word "global".
+     *
+     * @param regex - Regex for regions in a given partition.
+     * @param region - Region string being checked.
+     * @return true if the region string should be included as a region.
+     */
+    private boolean validRegion(String regex, String region) {
+        return region.matches(regex) ||
+               region.endsWith("-fips") &&
+               region.matches(regex.replace("$", "(-fips)*$")) ||
```
millems: This line seems more complex than it needs to be. Why not just substring off the -fips in the region?

millems: Can we write some unit tests for this specific logic?

millems: Have we verified how this affects existing regions?

spfink: I've verified with the services that I know had issues that this works. 

It does however introduce some backwards incompatibilies.

Regions like sandbox and local are gone. Those are nice to have but we'd have to one off them.

I don't want to but we may want to have boths -fips as a suffix and fips- as a prefix. S3 has fips- as a suffix and we are removing those...

codegen-lite/src/main/java/software/amazon/awssdk/codegen/lite/regions/ServiceMetadataGenerator.java
spfink: Refactored out to a utility class for clarity and testability.

codegen-lite/src/main/java/software/amazon/awssdk/codegen/lite/regions/ServiceMetadataGenerator.java
spfink: Refactored out to a utility class for clarity and testability.

spfink: Refactored out to a utility class for clarity and testability.

Issue #35, opened 2017-07-03
Dynamo DB Mapper


millems:
Review the inherited state of V1 Dynamo DB mapper support and determine which changes are necessary for V2.

(Feel free to comment on this issue with desired changes).

cjkent:
Please consider supporting immutable classes in the mapper. See https://github.com/aws/aws-sdk-java-v2/issues/48

MikeFHay:
Please remove PaginatedList, and replace uses of it with Stream.

java.util.List implementations are expected to have fast `size()` methods, but as far as I can tell there is no way to implement that for a DynamoDB scan or query. Currently PaginatedList will either load the entire scan result into memory on a `size()` call, or simply fail if configured as ITERATION_ONLY. This is a potentially surprising behaviour which can be made much more intuitive and explicit by returning a Stream and requiring users to `collect(toList())` if they want a List.

millems:
We're looking to make the automatic depagination provided by PaginatedList work for all service's paginated APIs with this story: https://github.com/aws/aws-sdk-java-v2/issues/26. 

+1 to avoiding accidentally blowing up your memory and holding up a thread by using something as simple as `size()`. Using `Stream` is a solid idea. We'll need to think about whether other pagination strategies for different APIs can have an efficient `size()` method. If so, we shouldn't expose it in as innocuous a form for services like Dynamo that make it expensive.

leadVisionary:
One of the things I'd call out here is that I think there's a problem with the com.amazonaws.services.dynamodbv2.datamodeling.StandardBeanProperties method of reflecting over an object. Say I have:

```
@DynamoDBTable(tableName = "Polyps")
public final class Polyp {
private Set<String> endpoints = new ConcurrentSkipListSet<>();
public Set<String> getEndpoints() {
        return endpoints;
    }
    public void setEndpoints(final Collection<String> endpoints) {
        if (endpoints != null && !endpoints.isEmpty()) {
            this.endpoints = new ConcurrentSkipListSet<>(endpoints);
        }
    }
}
```

In theory, even if DynamoDB only respects Sets, I should be able to handle the case where a setter allows a wider Collection type.

It looks like this is because of line 140-146, the `setterOf` method, which does a "clobber get with set and look for the exact type" algorithm. Consequently, this method returns null, which later becomes a problem during line 111 of DynamoDBMapperFieldModel.

I think it should be checking if the return type has parent types, and if so trying to run up the tree.

[...]

PR #1090, opened 2019-02-21
[Discussion] Added initial Dynamo DB project tenets/goals.

millems:


docs/design/services/dynamodb/high-level-library/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+1. The high-level library should be the ideal way to integrate with
+   Dynamo DB.
+2. The nouns and verbs used by the high-level library should match those
+   in Dynamo DB.
+3. The high-level library should use and build on the idiomatic norms in
+   the rest of the 2.x Java SDK.
+4. The high-level library should perform well in cold JVMs.
+5. Using the low-level Dynamo DB client should be a conscious choice by
+   the customer, not an accidental one.
```
dagnir: What does accidental mean in this case?

millems: There is probably a better word for what I'm getting at. Customers that use the low-level client should know there's a high-level client and intentionally have chosen the low-level one. They shouldn't be able to just use the low-level client and not know any better.

dagnir: Okay yeah, I figured that's what you were trying to say. I agree, I think we can find a better way to phrase this. Don't know one off the top of my head though. I'll update when I think of one!

docs/design/services/dynamodb/high-level-library/README.md
```diff
@@ -0,0 +1,38 @@
...
+   
+# Customer-Requested Changes from 1.11.x
```
zoewangg: Can we link the related github issues here for reference?

docs/design/services/dynamodb/high-level-library/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+This project provides a Dynamo DB experience that is easier to use and
+understand than can be provided by the generated Dynamo DB client. It
+provides object-oriented features, like persisting and updating Java
+data objects, as well as expressive data retrieval operations.
+
+# Project Goals
+
+1. The high-level library should be the ideal way to integrate with
+   Dynamo DB.
```
zoewangg: nit, shouldn't it be one word `DynamoDB`?

docs/design/services/dynamodb/high-level-library/README.md
```diff
@@ -0,0 +1,38 @@
...
+2. The nouns and verbs used by the high-level library should match those
+   in Dynamo DB.
+3. The high-level library should use and build on the idiomatic norms in
+   the rest of the 2.x Java SDK.
+4. The high-level library should perform well in cold JVMs.
```
zoewangg: How do we define `well`? Performing better than the low level client?

docs/design/services/dynamodb/high-level-library/README.md
millems: I'll see if I can cobble them together. They're all connected to https://github.com/aws/aws-sdk-java-v2/issues/35

docs/design/services/dynamodb/high-level-library/README.md
millems: Probably? I'm not sure too much on the branding.

docs/design/services/dynamodb/high-level-library/README.md
millems: I'm specifically trying to avoid being specific, so that we have flexibility to make the decision later on. I'm just saying that cold-start performance should be something we worry about.

docs/design/services/dynamodb/high-level-library/README.md
```diff
@@ -0,0 +1,38 @@
...
+3. Discoverability drives usage.
+
+# Introduction
+
+This project provides a Dynamo DB experience that is easier to use and
+understand than can be provided by the generated Dynamo DB client. It
+provides object-oriented features, like persisting and updating Java
+data objects, as well as expressive data retrieval operations.
+
+# Project Goals
```
varunnvs92: Do we have goals on what operations/features from low-level clients that the high level client supports?
I mean: Should all features supported in low-level client be supported in high level client (or) a subset (if so, how do we determine the subset)

millems: "The high-level library should be the ideal way to integrate with Dynamo DB." as a goal seems to indicate to me that we should support all features in the low-level client. i.e. people shouldn't want to use the low level client.

docs/design/services/dynamodb/high-level-library/README.md
```diff
@@ -0,0 +1,38 @@
...
+   the customer, not an accidental one.
+   
+# Customer-Requested Changes from 1.11.x
+
+* All built-in Java types should be supported without type converters.
+* There should be just one method of type conversion that can be
+  extended to support any reasonable, foreseeable customer use-case.
+* The high-level library should support immutable object representations
+  from the customer.
+* The high-level library should support object inheritance.
```
danieladams456: Does this mean I will be able to do a single `query` operation on a `SuperClass` and get back a `List<SuperClass>` which I can iterate over with polymorphism get two different item types?  For example, my partition key might be order ID and a query on that will get back one item of order metadata and `n` items of order line items.

millems: The exact details haven't been shaken out yet, but that's the general idea, yeah. Something like this:

```Java
@DynamoDbSubtype(Dog.class, Cat.class)
class Animal {
    @Attribute 
    @Index(PARTITION_KEY)
    private String id;
}

class Cat extends Animal {
    @Attribute
    private String name;
}

class Dog extends Animal {
    @Attribute
    private String name;
    @Attribute
    private String breed;
}
```

```Java
client.put(new Cat().setId("1").setName("Meowth"));
client.put(new Dog().setId("2").setName("Growlith").setBreed("Awesome"));
client.scan(new Animal()).forEach(animal -> {
    if (animal instanceof Cat) {
        Cat cat = (Cat) animal;
        System.out.println("Cat: " + cat.getName());
    } else if (animal instanceof Dog) {
        Dog dog = (Dog) animal;
        System.out.println("Dog: " + dog.getName() + " (" + dog.getBreed() + ")");
    }
});

danieladams456: Awesome!!! ❤️

millems:
Continued: https://github.com/aws/aws-sdk-java-v2/pull/1159

PR #1096, opened 2019-02-27
Added try catch in various places to prevent uncompleted future

zoewangg:
## Description
- Added try catch in various places to prevent uncompleted future 
- Other small fixes

## Testing
running existing tests.

## Types of changes

## Checklist

## License

### DIFF
--- a/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ResponseHandler.java
+++ b/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ResponseHandler.java
@@ -20,10 +20,13 @@ import static java.util.stream.Collectors.mapping;
 import static software.amazon.awssdk.http.nio.netty.internal.ChannelAttributeKey.EXECUTE_FUTURE_KEY;
 import static software.amazon.awssdk.http.nio.netty.internal.ChannelAttributeKey.REQUEST_CONTEXT_KEY;
 import static software.amazon.awssdk.http.nio.netty.internal.ChannelAttributeKey.RESPONSE_COMPLETE_KEY;
+import static software.amazon.awssdk.http.nio.netty.internal.utils.ExceptionHandlingUtils.tryCatch;
+import static software.amazon.awssdk.http.nio.netty.internal.utils.ExceptionHandlingUtils.tryCatchFinally;
 
 import com.typesafe.netty.http.HttpStreamsClientHandler;
 import com.typesafe.netty.http.StreamedHttpResponse;
 import io.netty.buffer.ByteBuf;
+import io.netty.channel.Channel;
 import io.netty.channel.ChannelHandler.Sharable;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.SimpleChannelInboundHandler;
@@ -40,6 +43,7 @@ import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.stream.Collectors;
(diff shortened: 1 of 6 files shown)


http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/UnusedChannelExceptionHandler.java
```diff
@@ -60,8 +62,7 @@ public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) {
                                 "Java SDK, where a future was not completed while the channel was in use. The channel has " +
                                 "been closed, and the future will be completed to prevent any ongoing issues.", cause);
                 executeFuture.get().completeExceptionally(cause);
-
-            } else if (cause instanceof IOException) {
+            } else if (isNettyIoException(cause)) {
```
zoewangg: I believe NettyIOException is intended to check here, can you confirm? @millems

millems: Adding `ReadTimeoutException` and `WriteTimeoutException` to the check makes sense. I thought we had removed those handlers at this point, though.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ResponseHandler.java
```diff
@@ -246,11 +245,16 @@ public void onNext(HttpContent httpContent) {
...
-                    ByteBuffer b = copyToByteBuffer(httpContent.content());
-                    httpContent.release();
-                    subscriber.onNext(b);
-                    channelContext.read();
+
+                    try {
+                        // Needed to prevent use-after-free bug if the subscriber's onNext is asynchronous
+                        ByteBuffer b = copyToByteBuffer(httpContent.content());
+                        httpContent.release();
+                        subscriber.onNext(b);
```
dagnir: I think we might need to move this out of the try-catch. According to the spec, if `onNext` throws, we need to treat the subscription as cancelled, and we can't call `onError` in that case (see 2.13 https://github.com/reactive-streams/reactive-streams-jvm#2-subscriber-code)

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ResponseHandler.java
```diff
@@ -156,9 +155,9 @@ public static ResponseHandler getInstance() {
      * @param ctx Context for channel
      */
     private static void closeAndRelease(ChannelHandlerContext ctx) {
-        RequestContext requestContext = ctx.channel().attr(REQUEST_CONTEXT_KEY).get();
-        ctx.channel().close();
-        requestContext.channelPool().release(ctx.channel());
+        Channel channel = ctx.channel();
+        RequestContext requestContext = channel.attr(REQUEST_CONTEXT_KEY).get();
+        channel.close().addListener(i -> requestContext.channelPool().release(channel));
```
dagnir: Nice catch!

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ResponseHandler.java
zoewangg: Yeah, fixed

zoewangg:
I agree we should find a way to do catch-all and add more tests to verify. Created backlog item for that.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ResponseHandler.java
```diff
@@ -246,10 +250,20 @@ public void onNext(HttpContent httpContent) {
...
+                        // Needed to prevent use-after-free bug if the subscriber's onNext is asynchronous
+                        byteBuffer = copyToByteBuffer(httpContent.content());
+                    } catch (Throwable throwable) {
+                        onError(throwable);
+                    } finally {
+                        httpContent.release();
+                    }
+
+                    if (byteBuffer != null) {
+                        subscriber.onNext(byteBuffer);
```
dagnir: should we still do a try catch here and complete the future in case we have a bad subscriber?

zoewangg: +1 sounds good

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ResponseHandler.java
```diff
@@ -251,20 +254,21 @@ public void onNext(HttpContent httpContent) {
...
+                    //As per reactive-streams rule 2.13, we should not call subscriber#onError when
+                    //exception is thrown from subscriber#onNext
                     if (byteBuffer != null) {
-                        subscriber.onNext(byteBuffer);
+                        tryCatch(() -> subscriber.onNext(byteBuffer),
+                                 this::notifyError);
                     }
-                    channelContext.read();
+
+                    tryCatch(channelContext::read, this::onError);
```
dagnir: Does this need to be moved into an `else` block? Seems like we wouldn't want to keep reading if `byteBuffer` is null since that means we failed to copy the `ByteBuf`.

zoewangg: You mean moving it to `if` block? yeah, we should otherwise it will get executed if exception is thrown.

zoewangg:
Integ tests passed. Merging.

PR #1101, opened 2019-02-28
[Discussion] "Option 1" prototype for the DynamoDB high-level library: DynamoDB-specific client

millems:
This contains non-functioning interfaces for one design direction that could be taken for the high-level library. 

Another "option 2" we wish to explore is what it would look like to make a non-DynamoDB-specific client that acts as an general abstraction over document databases. "Option 2" would enable support for other backing data stores, like MongoDB. A PR for this option will come at a later date.

Prototyped features in this "option 1" PR:
1. A single "DynamoDb" class that is meant to act as the entry-point for all DynamoDB functionality across AWS. The modularization strategy behind this class is discussed in the PR.
2. A single "document" client that intends to replace both the document and mapper clients of 1.11.x and the DynamoDbClient of 2.x.
3. Interfaces for "putting" and "getting" items from DynamoDB. This is done via Java POJOs/beans (similar to 1.11.x's DynamoDBMapper) or an "Item" abstraction (similar to 1.11.x's DynamoDBDocumentClient).
4. Interfaces for custom type conversion between Java-specific types and DyamoDB-specific types.

**Some of the code samples from the PR (please comment on them in the PR directly)**

*Create Table:*
```Java
try (DynamoDbDocumentClient client = DynamoDb.documentClient()) {
    ProvisionedCapacity tableCapacity = ProvisionedCapacity.builder()
                                                           .readCapacity(5)
                                                           .writeCapacity(5)
                                                           .build();
     
    KeySchema tableKeys = KeySchema.builder()
                                   .putKey("partition-key", ItemAttributeIndexType.PARTITION_KEY)
                                   .build();
     
    client.createTable(CreateTableRequest.builder()
                                         .tableName("my-table")
                                         .provisionedCapacity(tableCapacity)
                                         .keySchema(tableKeys)
                                         .build());
     
    System.out.println("Table created successfully.");
} catch (TableAlreadyExistsException e) {
    System.out.println("Table creation failed.");
}
```

*List Tables:*
```Java
try (DynamoDbDocumentClient client = DynamoDb.documentClient()) {
    String tables = client.listTables().tables().stream()
                          .map(Table::name)
                          .collect(Collectors.joining(","));
    System.out.println("Current Tables: " + tables);
}
```

*Put Item:*
```Java
try (DynamoDbDocumentClient client = DynamoDb.documentClient()) {
    Table table = client.getTable("my-table");
    table.putItem(Item.builder()
                      .putAttribute("partition-key", UUID.randomUUID())
                      .putAttribute("creation-time", Instant.now())
                      .build());
}
```
Another PutItem API is provided with additional options, like consumed capacity. See code for an example.

*Get Item*:
```Java
try (DynamoDbDocumentClient client = DynamoDb.documentClient()) {
    Table table = client.getTable("my-table");
    UUID id = UUID.randomUUID();
    table.putItem(Item.builder()
                      .putAttribute("partition-key", id)
                      .putAttribute("creation-time", Instant.now())
                      .build());

    // Wait a little bit, because getItem is eventually consistent by default.
    Thread.sleep(5_000);

    Item item = table.getItem(Item.builder()
                                  .putAttribute("partition-key", id)
                                  .build());

    // Times are stored as numbers, by default, so they can also be used as sort keys.
    assert item.attribute("creation-time").isNumber();
    assert item.attribute("creation-time").as(Instant.class).isBetween(Instant.now().minus(1, MINUTE),
                                                                       Instant.now());
} catch (NoSuchItemException e) {
    System.out.println("Item could not be found. Maybe we didn't wait long enough for consistency?");
    throw e;
}
```
Another GetItem API is provided with additional options, like consistent reads. See code for an example.

Put Object:
```Java
public class MyItem {
    @Attribute("partition-key")
    @Index(AttributeIndexType.PARTITION_KEY)
    private String partitionKey;

    @Attribute("creation-time")
    private Instant creationTime;

    public String getPartitionKey() { return this.partitionKey; }
    public Instant getCreationTime() { return this.creationTime; }
    public void setPartitionKey(String partitionKey) { this.partitionKey = partitionKey; }
    public void setCreationTime(Instant creationTime) { this.creationTime = creationTime; }
}

try (DynamoDbDocumentClient client = DynamoDb.documentClient()) {
    Table table = client.getTable("my-table");

    MyItem itemToCreate = new MyItem();
    itemToCreate.setPartitionKey(UUID.randomUUID());
    itemToCreate.setCreationTime(Instant.now());

    table.putObject(itemToCreate);
}
```
Another PutObject API is provided with additional options, like consumed capacity. See code for an example.

Get Object:
```Java
public class MyItem {
    @Attribute("partition-key")
    @Index(AttributeIndexType.PARTITION_KEY)
    private String partitionKey;

    @Attribute("creation-time")
    private Instant creationTime;

    public String getPartitionKey() { return this.partitionKey; }
    public Instant getCreationTime() { return this.creationTime; }
    public void setPartitionKey(String partitionKey) { this.partitionKey = partitionKey; }
    public void setCreationTime(Instant creationTime) { this.creationTime = creationTime; }
}

try (DynamoDbDocumentClient client = DynamoDb.documentClient()) {
    Table table = client.getTable("my-table");

    UUID id = UUID.randomUUID();

    MyItem itemToCreate = new MyItem();
    itemToCreate.setPartitionKey(id);
    itemToCreate.setCreationTime(Instant.now());

    table.putObject(itemToCreate);

    // Wait a little bit, because getObject is eventually consistent by default.
    Thread.sleep(5_000);

    MyItem itemToRetrieve = new MyItem();
    itemToRetrieve.setPartitionKey(id);

    MyItem retrievedItem = table.getObject(itemToRetrieve);
    assert retrievedItem.getCreationTime().isBetween(Instant.now().minus(1, MINUTE),
                                                     Instant.now());
} catch (NoSuchItemException e) {
    System.out.println("Item could not be found. Maybe we didn't wait long enough for consistency?");
    throw e;
}
```
Another GetObject API is provided with additional options, like consistent reads. See code for an example.


docs/design/services/dynamodb/high-level-library/prototype/option-1/sync/Prototype.java
```diff
@@ -0,0 +1,1025 @@
+/**
```
zoewangg: what would be the package name?

docs/design/services/dynamodb/high-level-library/prototype/option-1/sync/Prototype.java
```diff
@@ -0,0 +1,1025 @@
...
+ *
+ * <p>
+ *     <b>Maven Module Location</b>
+ *     This would be in a separate maven module (software.amazon.awssdk:dynamodb-all) that depends on all other DynamoDB modules
+ *     (software.amazon.awssdk:dynamodb, software.amazon.awssdk:dynamodb-document). Customers that only want one specific client
+ *     could instead depend directly on the module that contains it.
+ * </p>
+ */
+@ThreadSafe
+public interface DynamoDb {
```
zoewangg: Are we going to have an `AsyncDynamoDb` which contains async clients or the low level async clients will be in this interface?

docs/design/services/dynamodb/high-level-library/prototype/option-1/sync/Prototype.java
```diff
@@ -0,0 +1,1025 @@
...
+         * Default: {@code DynamoDbClient.create()}
+         */
+        DynamoDbDocumentClient.Builder dynamoDbClient(DynamoDbClient client);
+
+        /**
+         * Configure the DynamoDB document client with a specific set of configuration values that override the defaults.
+         *
+         * Default: {@code DocumentClientConfiguration.create()}
+         */
+        DynamoDbDocumentClient.Builder documentClientConfiguration(DocumentClientConfiguration configuration);
```
zoewangg: I wonder if we should unify this configuration builder method name for all HLLs to make it more discoverable,  overrideConfiguraion?

docs/design/services/dynamodb/high-level-library/prototype/option-1/sync/Prototype.java
```diff
@@ -0,0 +1,1025 @@
...
+     *
+     *         table.createGlobalSecondaryIndex(CreateGlobalSecondaryIndexRequest.builder()
+     *                                                                           .indexName("my-new-index")
+     *                                                                           .provisionedCapacity(tableCapacity)
+     *                                                                           .keySchema(tableKeys)
+     *                                                                           .build());
+     *     }
+     * </code>
+     */
+    CreateGlobalSecondaryIndexResponse createGlobalSecondaryIndex(CreateGlobalSecondaryIndexRequest createRequest);
```
zoewangg: Can we have builder consumer overload :)

docs/design/services/dynamodb/high-level-library/prototype/option-1/sync/Prototype.java
```diff
@@ -0,0 +1,1025 @@
...
+     * Already GA in module software.amazon.awssdk:dynamodb.
+     */
+    DynamoDbClientBuilder clientBuilder();
+
+    /**
+     * Create a high-level "document" DynamoDB client with default configuration.
+     *
+     * Usage Example:
+     * <code>
+     *     try (DynamoDbDocumentClient client = DynamoDb.documentClient()) {
```
zoewangg: nit: not related, but something to think about, we might not want to use try-with-resources in the code snippet  in the Javadoc because that might encourage people to create new client per request

docs/design/services/dynamodb/high-level-library/prototype/option-1/sync/Prototype.java
millems: It's up to debate, but I was thinking a group ID and package name of software.amazon.aws. The target libraries aren't necessarily always going to be under the AWS SDK umbrella (eg. KCL could be one example), so a general ".aws" felt appropriate.

docs/design/services/dynamodb/high-level-library/prototype/option-1/sync/Prototype.java
millems: I was thinking this would also contain the async clients, so that there's only one specific entry point.

docs/design/services/dynamodb/high-level-library/prototype/option-1/sync/Prototype.java
millems: The cons of that is that we will deviate from the type-matches-the-setter-name approach we have elsewhere.

We could call it DocumentOverrideConfiguration or something similarly long so that we can have our cake and eat it too.

docs/design/services/dynamodb/high-level-library/prototype/option-1/sync/Prototype.java
millems: We'd add consumer-builders everywhere in the real solution. I left them out to keep the design focused on the methods themselves.

docs/design/services/dynamodb/high-level-library/prototype/option-1/sync/Prototype.java
millems: Other hand: not using try-with-resources in code snippets might encourage people not to close their clients.

Point taken, though.

docs/design/services/dynamodb/high-level-library/prototype/option-1/sync/Prototype.java
```diff
@@ -0,0 +1,1026 @@
...
+         *         // InstantsAsStringsConverter.
+         *         assert item.attribute("creation-time").isString();
+         *         assert item.attribute("creation-time").as(Instant.class).isBetween(Instant.now().minus(1, MINUTE),
+         *                                                                            Instant.now());
+         *     }
+         * </code>
+         */
+        DocumentClientConfiguration.Builder converters(List<ItemAttributeValueConverter<?>> converters);
+        DocumentClientConfiguration.Builder addConverter(ItemAttributeValueConverter<?> converter);
+        DocumentClientConfiguration.Builder clearConverters();
```
varunnvs92: What is the difference between clearConverters and passing null/empty map to converters(..) method?

docs/design/services/dynamodb/high-level-library/prototype/option-1/sync/Prototype.java
```diff
@@ -0,0 +1,1026 @@
...
+     * </code>
+     *
+     * @see DynamoDbDocumentClient.Builder
+     */
+    DynamoDbDocumentClient.Builder documentClientBuilder();
+}
+
+/**
+ * A synchronous client for interacting with DynamoDB. While the low-level {@link DynamoDbClient} is generated from a service
+ * model, this client is hand-written and provides a richer client experience for DynamoDB.
```
varunnvs92: Not a design question: As the goals indicate HLL should be the preferred way, we need to hand-code every single API. I feel this might become like v1 s3 client - where we need to be up-to-date whenever new features are added in DynamoDB. Are there any parts which we can generate? (its an implementation detail, so we can discuss later too)

docs/design/services/dynamodb/high-level-library/prototype/option-1/sync/Prototype.java
```diff
@@ -0,0 +1,1026 @@
...
+     *     }
+     * </code>
+     */
+    <T> PutObjectResponse<T> putObject(PutObjectRequest<T> putRequest)
+            throws ConditionFailedException;
+
+    /**
+     * Invoke DynamoDB to retrieve an Item in this table, based on its partition key (and sort key, if the table has one).
+     *
+     * This method is optimized for performance, and provides no additional response data. For additional options
```
varunnvs92: Can you explain why this will have more performance? Even though we might return part of response, our marshallers stills need to marshall the entire response right

docs/design/services/dynamodb/high-level-library/prototype/option-1/sync/Prototype.java
millems: No difference. This is just matching the standard we have for other setters in the SDK.

docs/design/services/dynamodb/high-level-library/prototype/option-1/sync/Prototype.java
millems: It would be nice to generate certain parts or reuse the low-level abstractions, where it makes sense. I do this in a few places (or I did - not sure if I removed it), where we expose the low-level put response on the high-level putObject/putItem results. Hopefully we can find more opportunities like that throughout the implementation.

docs/design/services/dynamodb/high-level-library/prototype/option-1/sync/Prototype.java
millems: I picture the marshalling chain to roughly work like this:
1. Create a context for the current item.
2. Find a converter for this context:
   a. Iterate over each `contextSatisfies` converter to see if it can marshal this context. If so, use that converter.
   b. Look up the context.attributeSchema().javaType() in a `typeConverterCache`. If a converter is found, use that converter.
   c. Repeat (b) for the parent class (eventually, the base Object converter will be used), caching the result as the converter for that child type.
3. Use the convert to transform the item.

2(b) will converge on O(1) 
2(a) will always be O(# of contextSatisfies marshallers)

docs/design/services/dynamodb/high-level-library/prototype/option-1/sync/Prototype.java
```diff
@@ -0,0 +1,1026 @@
...
+         *         // Items are usually stored as a number, but it was stored as an ISO-8601 string now because of the
+         *         // InstantsAsStringsConverter.
+         *         assert item.attribute("creation-time").isString();
+         *         assert item.attribute("creation-time").as(Instant.class).isBetween(Instant.now().minus(1, MINUTE),
+         *                                                                            Instant.now());
+         *     }
+         * </code>
+         */
+        DocumentClientConfiguration.Builder converters(List<ItemAttributeValueConverter<?>> converters);
+        DocumentClientConfiguration.Builder addConverter(ItemAttributeValueConverter<?> converter);
```
dagnir: How are converters chosen if more than one is able to convert the item/object?

docs/design/services/dynamodb/high-level-library/prototype/option-1/sync/Prototype.java
```diff
@@ -0,0 +1,1026 @@
...
+     *
+     * Even if this condition is not satisfied, it can still be invoked directly via
+     * {@link ItemAttributeValue#convert(ItemAttributeValueConverter)}.
+     */
+    ConversionCondition defaultConversionCondition();
+
+    /**
+     * Convert the provided Java type into an {@link ItemAttributeValue}.
+     */
+    ItemAttributeValue toAttributeValue(T input, ConversionContext context);
```
dagnir: minor: would `? super T` be appropriate here?

docs/design/services/dynamodb/high-level-library/prototype/option-1/sync/Prototype.java
```diff
@@ -0,0 +1,1026 @@
...
+}
+
+/**
+ * The schema for a specific item. This describes the item's structure and which attributes it contains.
+ *
+ * This is mostly an implementation detail, and can be ignored except by developers interested in creating
+ * {@link ItemAttributeValueConverter}.
+ */
+@ThreadSafe
+public interface ItemSchema {
```
dagnir: How will polymorphism be supported? Will item schemas nest?

docs/design/services/dynamodb/high-level-library/prototype/option-1/sync/Prototype.java
```diff
@@ -0,0 +1,1026 @@
...
+         */
+        ItemSchema.Builder attributeSchemas(Map<String, ItemAttributeSchema> attributeSchemas);
+        ItemSchema.Builder putAttributeSchema(String attributeName, ItemAttributeSchema attributeSchema);
+        ItemSchema.Builder removeAttributeSchema(String attributeName);
+        ItemSchema.Builder clearAttributeSchemas();
+
+        /**
+         * The converter that should be used for converting all items that conform to this schema.
+         */
+        ItemSchema.Builder converter(ItemAttributeValueConverter<?> converter);
```
dagnir: Was this meant to be a different type? Seems like `ItemAttributeValueConverter` is for individual AttributeValues

docs/design/services/dynamodb/high-level-library/prototype/option-1/sync/Prototype.java
```diff
@@ -0,0 +1,1026 @@
...
+     *
+     *         table.putObject(itemToCreate);
+     *
+     *         // Wait a little bit, because getObject is eventually consistent by default.
+     *         Thread.sleep(5_000);
+     *
+     *         MyItem itemToRetrieve = new MyItem();
+     *         itemToRetrieve.setPartitionKey(id);
+     *
+     *         MyItem retrievedItem = table.getObject(itemToRetrieve);
```
dagnir: Why take in the whole object instead of just the keys?

docs/design/services/dynamodb/high-level-library/prototype/option-1/sync/Prototype.java
millems: Resolution Path:
1. Iterate over all `contextSatisfies` converters to try to find a match.
2. Iterate over all `isInstanceOf` converts to try to find a match on the class.
3. Iterate over all `isInstanceOf` converters to try to find a match on the parent class.
4. Iterate over all `isInstanceOf` converters to try to find a match on the parent's parent class. (Etc. up to Object which has an SDK-provided converter.)

The result of (2+) can be cached for a given Java type, to prevent having to crawl the chain each time. (1) must occur each time.

When "iterating over" the following precedence applies:
1. Attribute-Level (customer-provided) 
2. Item-Level (customer-provided) 
3. Client-Level (customer provided) 
4. SDK Converters

docs/design/services/dynamodb/high-level-library/prototype/option-1/sync/Prototype.java
millems: I don't think so. A converter of `Instant` probably can't convert `Object`s.

docs/design/services/dynamodb/high-level-library/prototype/option-1/sync/Prototype.java
millems: Inheritance isn't implemented in this prototype, but if it were... in my mind, an `ItemSchema` is the runtime equivalent of a class definition.

Class Definitions:
```Java
@ChildTypes(Dog.class)
public class Animal {
    @Index(PARTITION_KEY)
    @Attribute
    private UUID id;

    @Attribute
    private Instant birthDate;

    // getters/setters
}

@Converter(MyDogConverter.class)
public class Dog extends Animal {
    @Attribute
    @Converter(BreedConverter.class)
    private String breed;

    // getters/setters
}
```

Item Schemas:
```
{
    javaType: Animal.class,
    childTypes: [ Dog.class ],
    attributeSchemas: {
        id: {
            indexType: PARTITION_KEY,
            javaType: UUID.class
        },
        birthDate: {
            javaType: Instant.class
        }
    }
}

{
    javaType: Dog.class,
    converter: MyDogConverter.class,
    attributeSchemas: {
        breed: {
            javaType: String.class
            converter: BreedConverter.class
        }
    }
}
```

docs/design/services/dynamodb/high-level-library/prototype/option-1/sync/Prototype.java
millems: I had that originally. What syntax would you propose?

docs/design/services/dynamodb/high-level-library/prototype/option-1/sync/Prototype.java
millems: The syntax may not be perfect. I hand-waved it since I figured we could figure out the exact details at implementation time.

This is the converter to use for converting this specific Item into an ItemAttributeValue, assuming it's not overridden at the attribute level. 

Consider the following class structure:
```Java
class Car {
    @Attribute
    private Engine engine;
}

@Converter(EngineConverter.class)
class Engine {
    @Attribute
    private int cylinderCount;
}
```

We'd have the following ItemSchemas:
```
{
    javaType: Car.class,
    attributeSchemas: {
        engine: {
            javaType: Engine.class
        }
    }
}

{
    javaType: Engine.class,
    converter: EngineConverter.class,
    attributeSchemas: {
        cylinderCount: {
            javaType: int.class
        }
    }
}
```

This would instruct the SDK to marshal all usages of this specific `Engine` type using `EngineConverter`.

docs/design/services/dynamodb/high-level-library/prototype/option-1/sync/Prototype.java
```diff
@@ -0,0 +1,1026 @@
...
+     * <code>
+     *     try (DynamoDbDocumentClient client = DynamoDb.documentClient()) {
+     *         Table table = client.getTable("my-table");
+     *         table.putItem(PutItemRequest.builder()
+     *                                     .item(Item.builder()
+     *                                               .putAttribute("partition-key", "key")
+     *                                               .putAttribute("version", 2)
+     *                                               .build())
+     *                                     .condition("version = :expected_version")
+     *                                     .putConditionAttribute(":expected_version", 1)
```
spfink: If I'm reading this correctly, it's not super-intuitive to me. Can we link these together better than just strings?

docs/design/services/dynamodb/high-level-library/prototype/option-1/sync/Prototype.java
```diff
@@ -0,0 +1,1026 @@
...
+     *
+     * This will convert the provided object into an {@link Item} automatically using the default Object-to-Item
+     * {@link ItemAttributeValueConverter}, unless an alternate converter has been overridden for the provided type.
+     *
+     * This method is optimized for performance, and provides no additional response data. For additional options
+     * like conditions or consumed capacity, see {@link #putObject(PutObjectRequest)}.
+     *
+     * Usage Example:
+     * <code>
+     *     public class MyItem {
```
spfink: Can we infer without the attribute annotation?

docs/design/services/dynamodb/high-level-library/prototype/option-1/sync/Prototype.java
```diff
@@ -0,0 +1,1026 @@
...
+     * Usage Example:
+     * <code>
+     *     try (DynamoDbDocumentClient client = DynamoDb.documentClient()) {
+     *         client.listTables().tables().forEach(System.out::println);
+     *     }
+     * </code>
+     *
+     * @see DynamoDbDocumentClient
+     */
+    DynamoDbDocumentClient documentClient();
```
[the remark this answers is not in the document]
millems: I thought this was literally the way we agreed to expose these.

docs/design/services/dynamodb/high-level-library/prototype/option-1/sync/Prototype.java
```diff
@@ -0,0 +1,1026 @@
...
+     * Usage Example:
+     * <code>
+     *     try (DynamoDbDocumentClient client = DynamoDb.documentClient()) {
+     *         ProvisionedCapacity tableCapacity = ProvisionedCapacity.builder()
+     *                                                                .readCapacity(5)
+     *                                                                .writeCapacity(5)
+     *                                                                .build();
+     *
+     *         KeySchema tableKeys = KeySchema.builder()
+     *                                        .putKey("partition-key", ItemAttributeIndexType.PARTITION_KEY)
```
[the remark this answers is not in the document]
millems: Sure!

docs/design/services/dynamodb/high-level-library/prototype/option-1/sync/Prototype.java
```diff
@@ -0,0 +1,1026 @@
...
+ * by a client builder.
+ *
+ * @see DynamoDbDocumentClient.Builder#documentClientConfiguration(DocumentClientConfiguration)
+ */
+@ThreadSafe
+public interface DocumentClientConfiguration {
+    /**
+     * Create document client configuration with default values.
+     */
+    static DocumentClientConfiguration create();
```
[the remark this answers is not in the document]
millems: Yeah, there will be a `toBuilder`, and a lot more stuff.

docs/design/services/dynamodb/high-level-library/prototype/option-1/sync/Prototype.java
```diff
@@ -0,0 +1,1026 @@
...
+         *                                        .build();
+         *
+         *     try (DynamoDbDocumentClient client = DynamoDb.documentClientBuilder()
+         *                                                  .documentClientConfiguration(clientConfiguration)
+         *                                                  .build()) {
+         *
+         *         Table table = client.getTable("my-table");
+         *         UUID id = UUID.randomUUID();
+         *         table.putItem(Item.builder()
+         *                           .putAttribute("partition-key", id)
```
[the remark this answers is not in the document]
millems: Maybe. Partition key / sort key is mostly important when creating a table, and there's not currently an API to create a table from an item. It is theoretically useful in some converters, but that might be an add-when-it-is-needed thing.

docs/design/services/dynamodb/high-level-library/prototype/option-1/sync/Prototype.java
millems: This is currently just a passthrough for the Dynamo DB conditions. There will hopefully also be a higher-level abstraction for conditions. I didn't want to design that in the prototype.

docs/design/services/dynamodb/high-level-library/prototype/option-1/sync/Prototype.java
millems: We could either have it be opt-in with @Attribute annotations or opt-out with @NotAttribute annotations. I don't have a strong opinion.

docs/design/services/dynamodb/high-level-library/prototype/option-1/sync/Prototype.java
dagnir: I don't really have a specific one in mind. I just thought might be a little confusing with passing in the item type since it's not super clear that we just want the keys. I don't feel that strongly about it though, and do like that this is concise.

docs/design/services/dynamodb/high-level-library/prototype/option-1/sync/Prototype.java
dagnir: Okay, and for attributes within the same level, whichever is first?

docs/design/services/dynamodb/high-level-library/prototype/option-1/sync/Prototype.java
millems: It's tricky because retrieval needs the attribute value names for the keys, and I wanted to avoid needing to use the POJO APIs for writing and the Item APIs for retrieving.

I agree that it's not super clear, but this felt the simplest way of fixing the above problem. If there's another way, I'm open to it!

docs/design/services/dynamodb/high-level-library/prototype/option-1/sync/Prototype.java
millems: Yeah, if they do add(converter1) and then add(converter2), convert1 would be checked first.

millems:
Option 2 prototype is up: https://github.com/aws/aws-sdk-java-v2/pull/1116/files

millems:
Continued: https://github.com/aws/aws-sdk-java-v2/pull/1159

PR #1102, opened 2019-02-28
Fix possible NPE in HttpOrHttp2ChannelPool

dagnir:
## Description
The access to protocolImpl#close() could happen before it's assigned to, leading
to a NullPointerException.

Also, `protocolImpl` is not volatile or accessed in a threadafe manner, so
it's possible the thread calling `HttpOrHttp2ChannelPool#close` has a stale value.

## Motivation and Context
Bug fix.

## Types of changes

## Checklist

## License

### DIFF
--- a/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/HttpOrHttp2ChannelPool.java
+++ b/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/HttpOrHttp2ChannelPool.java
@@ -19,11 +19,10 @@ import static software.amazon.awssdk.http.nio.netty.internal.ChannelAttributeKey
 import static software.amazon.awssdk.http.nio.netty.internal.ChannelAttributeKey.PROTOCOL_FUTURE;
 import static software.amazon.awssdk.http.nio.netty.internal.utils.NettyUtils.doInEventLoop;
 
-import io.netty.bootstrap.Bootstrap;
 import io.netty.channel.Channel;
 import io.netty.channel.EventLoop;
+import io.netty.channel.EventLoopGroup;
 import io.netty.channel.pool.ChannelPool;
-import io.netty.util.concurrent.DefaultPromise;
 import io.netty.util.concurrent.Future;
 import io.netty.util.concurrent.GenericFutureListener;
 import io.netty.util.concurrent.Promise;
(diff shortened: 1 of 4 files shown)


http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/HttpOrHttp2ChannelPool.java
```diff
@@ -171,7 +171,22 @@ private void release0(Channel channel, Promise<Void> promise) {
...
     }
 
+    private void close0() {
+        if (protocolImplPromise == null) {
+            return;
+        }
+
+        protocolImplPromise.addListener((Future<ChannelPool> f) -> {
+            if (f.isSuccess()) {
+                f.getNow().close();
```
zoewangg: I wonder if we should wait for close to finish.

dagnir: I had the same though but https://netty.io/4.0/api/io/netty/channel/pool/ChannelPool.html#close--

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/HttpOrHttp2ChannelPool.java
```diff
@@ -171,7 +171,22 @@ private void release0(Channel channel, Promise<Void> promise) {
 
     @Override
     public void close() {
-        doInEventLoop(eventLoop, protocolImpl::close);
+        doInEventLoop(eventLoop, this::close0);
     }
 
+    private void close0() {
+        if (protocolImplPromise == null) {
```
zoewangg: Seems like `failProtocolImplPromise` would null `protocolImplPromise` out, do we need to close the channel pool there?

dagnir: Hmm I don't think so; if `protocolImplPromise` is null then we didn't create `protocolImpl` so closing `delegatePool` directly would be the right thing to do I think. Also the pool is still technically open at that point; the caller could call `acquire` again, and it could succeed this time.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/HttpOrHttp2ChannelPool.java
```diff
@@ -46,20 +45,21 @@
 
     private Promise<ChannelPool> protocolImplPromise;
     private ChannelPool protocolImpl;
+    private boolean closed;
```
zoewangg: Does it need to be `volatile`?

dagnir: I don't think so, the only thread that reads/writes this is the event loop

PR #1106, opened 2019-03-01
Configuring Codecov

zoewangg:
## Description
- Ignore test files for code coverage

## License

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,4 @@
+ignore:
+  - "codegen-lite/test/resources/**/*"
+  - "codegen/src/test/resources/**/*"
+  - "test/**/*"


.codecov.yml
```diff
@@ -0,0 +1,4 @@
+ignore:
+  - "codegen-lite/test/resources/**/*"
+  - "codegen/src/test/resources/**/*"
+  - "test/**/*"
```
varunnvs92: Extra character at the end

zoewangg: Hmm, not sure whey it showed up like that, but there's no extra character in this file though

varunnvs92: Any reason why we are just ignoring these folders? Are there others that need to be excluded?

zoewangg: This is what we discussed this morning, excluding the test folders for the test coverage. Double checked and those are the only folders we want to ignore.

zoewangg: Oh, I misunderstood your question. Yeah, those folders are the ones showing on codecov page that should not be counted.

screen shot 2019-03-01 at 3 36 31 pm

PR #1110, opened 2019-03-04
Add missing non-service modules to bom

### LINKED ISSUE #1063: software.amazon.awssdk:core not present in bom

zoewangg:
## Description
Add missing non-service modules to bom

## Motivation and Context
Fixes #1063 

## Types of changes

## Checklist

## License

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -17,7 +17,115 @@
         <sonar.skip>true</sonar.skip>
     </properties>
     <dependencyManagement>
+        <!-- Non-services -->
         <dependencies>
+            <dependency>
+                <artifactId>annotations</artifactId>
+                <groupId>software.amazon.awssdk</groupId>
+                <version>${awsjavasdk.version}</version>
+            </dependency>
+            <dependency>
+                <artifactId>auth</artifactId>
+                <groupId>software.amazon.awssdk</groupId>
+                <version>${awsjavasdk.version}</version>
+            </dependency>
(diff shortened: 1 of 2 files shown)


bom/pom.xml
```diff
@@ -17,7 +17,115 @@
         <sonar.skip>true</sonar.skip>
     </properties>
     <dependencyManagement>
+        <!-- Non-services -->
```
varunnvs92: Nit: Non-service moudles

Issue #1076, opened 2019-02-14
Head bucket operation doesn't throw correct exception


mguillet:
## Expected Behavior
When querying for `S3AsyncClient.headBucket` without specifying the region for a bucket that relies somewhere else than us-east-1, I'm expecting to get the proper 400 exception (was working up to 2.3.9)

## Current Behavior
We're getting an `SdkClientException` that is not the result of the http 400 failed response but rather:
```Caused by: java.io.IOException: Server failed to send complete response```

## Steps to Reproduce (for bugs)
```java
HeadBucketResponse response = S3AsyncClient.builder()
    .credentialsProvider(credentialsProvider)
    .build()
    .headBucket(HeadBucketRequest.builder().bucket("sentinel-inventory").build())
    .get();
```
Should throw a `S3Exception` and not an `SdkClientException`

<details><summary>Complete Stacktrace</summary>

```
java.util.concurrent.ExecutionException: software.amazon.awssdk.core.exception.SdkClientException

	at java.base/java.util.concurrent.CompletableFuture.reportGet(CompletableFuture.java:395)
	at java.base/java.util.concurrent.CompletableFuture.get(CompletableFuture.java:1999)
	at com.dd.logs.archive_reloader.planner.client.TestS3Head.test(TestS3Head.java:41)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
[...]

PR #1115, opened 2019-03-06
Update `SslCloseComletionEventHandler`

zoewangg:
## Description
Problem: `java.io.IOException: Server failed to send complete response` was thrown incorrectly for error response from the service. See #1076 

The root cause is that there is a race condition between when `ResponseHandler#channelnactive` is triggered and when the response is completed.

The fix is to update `SslCloseComletionEventHandler` to close channel for `SslCloseCompletionEvent` only if the channel is unused. The channel will be closed in `ReponseHandler` afterwards.

See the following stacktrace, by the time the channel was inactive, the response had not been completed yet, so `java.io.IOException: Server failed to send complete response` was thrown.
```
2019-03-05 17:31:02,202 [aws-java-sdk-NettyEventLoop-2-3] READ COMPLETE
2019-03-05 17:31:02,202 [aws-java-sdk-NettyEventLoop-2-3] INFO  software.amazon.awssdk.http.nio.netty.internal.ResponseHandler - onSubscribe being invoked
2019-03-05 17:31:02,205 [aws-java-sdk-NettyEventLoop-2-3] DEBUG INACTIVE
2019-03-05 17:31:02,205 [aws-java-sdk-NettyEventLoop-2-3] INFO  software.amazon.awssdk.http.nio.netty.internal.ResponseHandler - channelInactive being invoked
2019-03-05 17:31:02,205 [aws-java-sdk-NettyEventLoop-2-3] ERROR software.amazon.awssdk.http.nio.netty.internal.ResponseHandler - RESPONSE HAS NOT COMPLETED YET!!!
2019-03-05 17:31:02,215 [aws-java-sdk-NettyEventLoop-2-3] DEBUG CLOSE
2019-03-05 17:31:02,216 [aws-java-sdk-NettyEventLoop-2-3] DEBUG UNREGISTERED
2019-03-05 17:31:02,216 [aws-java-sdk-NettyEventLoop-2-3] INFO  PublisherAdapter.onComplete, Completing the response
2019-03-05 17:31:02,218 [aws-java-sdk-NettyEventLoop-2-3] INFO  software.amazon.awssdk.core.internal.http.async.AsyncResponseHandler - BaosSubscriber.onComplete
``` 

## Testing
Added unit tests and s3 Integ test passed.

## Types of changes

## Checklist

## License


zoewangg:
The channel will be closed in `ResponseHandler` after the response is completed or any exception occurs.

https://github.com/aws/aws-sdk-java-v2/blob/9912da700b70512bd46316cc9bbd9f23f11d0b22/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ResponseHandler.java#L126-L132

zoewangg:
I guess there might be still some edge cases where the inactive channel is unclosed. We should probably add `channel.close` in `ResponseHandler#channelInactive` to reduce the chance. The requests would fail at the first time but should succeed in retry.

spfink:
Let's definitely test this by using S3 and making requests at low TPS. I don't think this will make the problem as bad as when we didn't have this but I think this increases the likelihood of reusing a bad channel. 

If the testing shows the results are poor, we can always modify the SslCloseCompletionEventHandler to tag the thread or something so that we can make sure not to re-use it later.

zoewangg:
Did some tests with low TPS of s3.getObject and result didn't get worse compared with before.

However, I can still consistently replicate ` java.io.IOException: Server failed to send complete response` issue when the channel sometimes becomes `INACTIVE` right after `READ COMPLETE` without user event `SslCompletionEvent`, which means the service closes the connection without close_notify. There is a race condition between response completing and `ResponseHandler#channelInactive`. I'll address it in another PR

zoewangg:
Found out this does not correctly fix the issue after doing more tests, will create a new PR.

zoewangg:
This change would not make the `fail to complete response` exception worse and will fix the incorrect exception for headBucket. Re-opening.

PR #1120, opened 2019-03-08
[Discussion] Add initial design tenets/goals for S3 TransferManager

dagnir:
Initial goals/tenets document for S3 TransferManager. Feedback requested!

### DIFF
--- a/docs/design/services/s3/transfermanager/README.md
+++ b/docs/design/services/s3/transfermanager/README.md
@@ -1,0 +1,66 @@
+# Project Tenets (unless you know better ones)
+
+1. Meeting customers in their problem space allows them to deliver value
+   quickly.
+2. Meeting customer expectations drives usability.
+3. Discoverability drives usage.
+
+# Introduction
+
+This project provides a much improved experience for S3 customers needing to
+easily perform uploads and downloads of objects to and from S3 by providing the
+S3 `S3TransferManager`, a high level library built on the S3 client.
+
+# Project Goals
+
+1. For the use cases it addresses, i.e. the transfer of objects to and from S3,
+   S3TransferManager is the preferred solution. It is easier and more intuitive
+   than using the S3 client. In the majority of situations, it is more
+   performant.
+1. S3TransferManager provides a truly asynchronous, non-blocking API that
+   conforms to the norms present in the rest of the SDK.
+1. S3TransferManager makes efficient use of system resources.
+1. S3TransferManager supplements rather than replaces the lower level S3 client.
+
+# Non Project Goals
+
+1. Ability to use the blocking, synchronous client.
+
+   Using a blocking client would severely impede the ability to deliver on goals
+   #2 and #3.
+
+# Customer-Requested Changes from 1.11.x
+
+* S3TransferManager supports progress listeners that are easier to use.
+
+  Ref: https://github.com/aws/aws-sdk-java-v2/issues/37#issuecomment-316218667
+
+* S3TransferManager provides bandwidth limiting of uploads and downloads.
+
+  Ref: https://github.com/aws/aws-sdk-java/issues/1103
+
+* The size of resources used by Transfermanager and configured by the user
+  should not affect its stability.
+
+  For example, the configured size of a threadpool should be irellevant to its
+  ability to successfuly perform an operation.
+
+  Ref: https://github.com/aws/aws-sdk-java/issues/939
+
+* S3TransferManager supports parallel downloads of any object.
+
+  Any object stored in S3 should be downloadable in multiple parts
+  simultaneously, not just those uploaded using the Multipart API.
+
+* S3TransferManager has the ability to upload to and download from a pre-signed
+  URL.
+
+* S3TransferManager allows uploads and downloads from and to memory.
+
+  Ref: https://github.com/aws/aws-sdk-java/issues/474
+
+* Ability to easily use canned ACL policies with all transfers to S3.
+
+  Ref: https://github.com/aws/aws-sdk-java/issues/1207
+
+* Trailing checksums for parallel uploads and downloads.


docs/design/services/s3/transfermanager/README.md
```diff
@@ -0,0 +1,66 @@
...
+the S3 `TransferManager`, a high level library built on the S3 client.
+
+# Project Goals
+
+1. For the use cases it addresses, i.e. the transfer of objects to and from S3,
+   TransferManager is the preferred solution. It is easier and more intuitive
+   than using the S3 client. In the majority of situations, it is more
+   performant.
+1. TransferManager provides a truly asynchronous API that conforms to the norms
+   present in the rest of the SDK.
```
millems: Non-blocking, I assume?

docs/design/services/s3/transfermanager/README.md
```diff
@@ -0,0 +1,66 @@
...
+1. Ability to use the blocking, synchronous client.
+
+   Using a blocking client would severely impede the ability to deliver on goals
+   #2 and #3.
+
+# Customer-Requested Changes from 1.11.x
+
+* TransferManager supports progress listeners that are easier to use.
+
+  Ref: https://github.com/aws/aws-sdk-java-v2/issues/37#issuecomment-316218667
```
millems: More of a meta-question: do we know yet whether these will be S3-specific, or built into the core?

docs/design/services/s3/transfermanager/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+  Ref: https://github.com/aws/aws-sdk-java/issues/939
+
+* TransferManager supports parallel downloads of any object.
+
+  Any object stored in S3 should be downloadable in multiple parts
+  simultaneously, not just those uploaded using the Multipart API.
+
+* TransferManager has the ability to upload to and download from a pre-signed
+  URL.
```
millems: Does this include parallel downloads (see previous goal)?

docs/design/services/s3/transfermanager/README.md
dagnir: Yes, good catch. I'll update the wording

docs/design/services/s3/transfermanager/README.md
dagnir: I think it makes sense for them to be same thing. Or at least, TransferManager supports both the core and an "enhanced" TransferManager specific version if it makes sense to have one.

docs/design/services/s3/transfermanager/README.md
dagnir: For downloads definitely.  I wasn't sure if it's possible for uploads so I didn't want to specify parallel in case it's not. I'll confirm and then update the wording if necessary

docs/design/services/s3/transfermanager/README.md
```diff
@@ -0,0 +1,66 @@
...
+   performant.
+1. TransferManager provides a truly asynchronous API that conforms to the norms
+   present in the rest of the SDK.
+1. TransferManager makes efficient use of system resources.
+1. TransferManager performs well in cold JVMs.
+1. TransferManager supplements rather than replaces the lower level S3 client.
+
+# Non Project Goals
+
+1. Ability to use the blocking, synchronous client.
```
spfink: Not sure where you separate client vs http client but I think it should be a primary goal to support sync http clients like Apache. Each HTTP client has different levels of configuration and allowing customers to use their preferred one should be a requirement.

dagnir: I guess both as in the blocking service client and HTTP clients. I can't see supporting a sync HTTP client as a goal can coexist with the async/non-blocking, efficiency, and performance goals. We already saw a lot of this when we explored whether we could expose sync HTTP clients as async and vice versa.

To me, a big part of the original TransferManager was to provide the async/non-blocking experience so by focusing on using the async HTTP clients, we further enhance that experience.

If it's a matter of configuration I would think supporting a config present in Apache but not Netty would be easier to achieve than supporting Apache/a non blocking HTTP client in TransferManager.

millems: I think just limiting it to non-blocking clients is fine, as long as we have support for HTTP proxies in Netty. Sync clients and async abstractions like TransferManager don't mix well.

docs/design/services/s3/transfermanager/README.md
```diff
@@ -0,0 +1,66 @@
...
+1. Meeting customers in their problem space allows them to deliver value
+   quickly.
+2. Meeting customer expectations drives usability.
+3. Discoverability drives usage.
+
+# Introduction
+
+This project provides a much improved experience for S3 customers needing to
+easily perform uploads and downloads of objects to and from S3 by providing
+the S3 `TransferManager`, a high level library built on the S3 client.
```
zoewangg: `S3TransferManager`?  😉

dagnir: +1 sounds good

docs/design/services/s3/transfermanager/README.md
spfink: So no shared connection pool if customers are using the sync s3 client? Which is made worse given that you have to use both the low-level client and transfer manager.

No lightweight JDK client?

No GAE support?

A worse debugging experience.

docs/design/services/s3/transfermanager/README.md
varunnvs92: * Parallel downloads are possible, but there might be edge cases where we can't compute object length and hence can't do parallel downloads
* Multi-part uploads are possible but its complex and requires customers to provide multiple presigned urls.

docs/design/services/s3/transfermanager/README.md
dagnir: In my view, TransferManager is an enhancement of the async client, where this is all true right now as well.

I think it's important that customers who can't use the async clients for the reasons given above are not left without an option, but I think that means we provide a similar, but separate library for non-async clients.

docs/design/services/s3/transfermanager/README.md
dagnir: In that case I'd probably just leave it as is for now as is for now, or add a caveat that parallel parallel downloads *where possible*.

docs/design/services/s3/transfermanager/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+* TransferManager allows uploads and downloads from and to memory.
+
+  Ref: https://github.com/aws/aws-sdk-java/issues/474
+
+* Ability to easily use canned ACL policies with all transfers to S3.
+
+  Ref: https://github.com/aws/aws-sdk-java/issues/1207
+
+* Trailing checksums for parallel uploads using the Multipart API.
```
spfink: And downloads?

docs/design/services/s3/transfermanager/README.md
```diff
@@ -0,0 +1,67 @@
...
+# Project Goals
+
+1. For the use cases it addresses, i.e. the transfer of objects to and from S3,
+   TransferManager is the preferred solution. It is easier and more intuitive
+   than using the S3 client. In the majority of situations, it is more
+   performant.
+1. TransferManager provides a truly asynchronous, non-blocking API that conforms
+   to the norms present in the rest of the SDK.
+1. TransferManager makes efficient use of system resources.
+1. TransferManager performs well in cold JVMs.
```
spfink: I'm not sure if this actually needs to be a project goal. Once you get to the point where using Transfer Manager is more performant, a slightly increased cold start won't likely be a big deal.

docs/design/services/s3/transfermanager/README.md
dagnir: +1

docs/design/services/s3/transfermanager/README.md
spfink: And to further clarify, I don't think this is actually going to be an issue in Transfer Manager but if for some weird reason we needed to use something like ObjectMapper I don't think that'd be a huge deal for this case.

dagnir: Yeah that sounds fine to me. As you said, I don't foresee the implementation doing anything that would increase cold start over the baseline.

dagnir:
@spfink The rendered view is numbered correctly :). Saves having to fix the numbers as you add/remove things https://github.com/aws/aws-sdk-java-v2/blob/5b26383246987213a35dd3fc459d3f336d15db5a/docs/design/services/s3/transfermanager/README.md

PR #1125, opened 2019-03-11
S3: Add getUrl method

varunnvs92:
* Add getUrl method similar to [v1 method](https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/AmazonS3.java#L5558)
* Introduces a new internal hand-written (enhancement) client with the getUrl method
* Low-level client exposes the APIs and delegates them to the enhancement client
* Added unit tests

### DIFF
--- a/.changes/next-release/feature-AmazonS3-f2cd9dd.json
+++ b/.changes/next-release/feature-AmazonS3-f2cd9dd.json
@@ -1,0 +1,5 @@
+{
+    "category": "Amazon S3", 
+    "type": "feature", 
+    "description": "Add support for getUrl operation. The API can be used to generate a URL that represents an object in Amazon S3. The url can be used to download the object content if the object has public read permissions."
+}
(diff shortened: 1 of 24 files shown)


varunnvs92:
> Why are they private? What if just want to view the utilities available?

As the methods are exposed through the low-level client, the handwritten client doesn't need to be exposed to customers.

> Why are we putting the methods in the low-level client? 

It would be better customer experience to have the methods directly in the low-level client rather than access them via utilities method (which I guess no one would be aware unless they read docs).
I was initially opposed to this idea due to the complex poet generation but its not as bad as we are just delegating to the hand-written client and custom code is only needed to create the hand-written client.

If we don't want to go with this approach, then I prefer the utilities() method be in the aggregate class (that exposes TransferManager) instead of being hidden in the low-level client. But I still feel adding the methods in low-level client is better customer experience.

spfink:
Why are these methods included in the low-level client but not methods like upload and download like we'll eventually have in Transfer Manager?

varunnvs92:
TransferManager is focused on improving object transfers (uploads, downloads, copy). getUrl doesn't make sense in the TransferManager. So it should either go in low-level client or a separate utility class. I feel using low-level client is more discoverable.

varunnvs92:
Closing in favor of https://github.com/aws/aws-sdk-java-v2/pull/1134

Issue #1078, opened 2019-02-14
UrlConnectionHttpClient leads to signature validation failures for some services


abrooksv:
## Expected Behavior
Both calls in example code should succeed

## Current Behavior
CloudFormation works, Lambda does not

```
StackSummary(StackId=arn:aws:cloudformation:us-west-2:.......)

The request signature we calculated does not match the signature you provided. Check your AWS Secret Access Key and signing method. Consult the service documentation for details. (Service: Lambda, Status Code: 403, Request ID: f9801917-3096-11e9-910f-47052ed185ec)
software.amazon.awssdk.services.lambda.model.LambdaException: The request signature we calculated does not match the signature you provided. Check your AWS Secret Access Key and signing method. Consult the service documentation for details. (Service: Lambda, Status Code: 403, Request ID: f9801917-3096-11e9-910f-47052ed185ec)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.HandleResponseStage.handleErrorResponse(HandleResponseStage.java:115)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.HandleResponseStage.handleResponse(HandleResponseStage.java:73)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.HandleResponseStage.execute(HandleResponseStage.java:58)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.HandleResponseStage.execute(HandleResponseStage.java:41)
	at software.amazon.awssdk.core.internal.http.pipeline.RequestPipelineBuilder$ComposingRequestPipelineStage.execute(RequestPipelineBuilder.java:205)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.ApiCallAttemptTimeoutTrackingStage.execute(ApiCallAttemptTimeoutTrackingStage.java:63)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.ApiCallAttemptTimeoutTrackingStage.execute(ApiCallAttemptTimeoutTrackingStage.java:36)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.TimeoutExceptionHandlingStage.execute(TimeoutExceptionHandlingStage.java:77)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.TimeoutExceptionHandlingStage.execute(TimeoutExceptionHandlingStage.java:39)
[...]

PR #1130, opened 2019-03-13
Removing body on GET requests

spfink:
This PR is intended for discussion on #1078 and more generally the issue of bodies on GET requests. Both our Apache client and our URL connection client don't support GET requests with bodies. 

Apache will simply silently drop the body. We can easily fix this.

The URL connection client will automatically switch any request with a body to a POST. It would not be easy to fix this without significantly forking.

Need to look into Netty.

Even without looking into Netty, the problem is clear. While GETs with bodies arent strictly disallowed, our clients don't support it by default. We probably could support it in all of our clients but this means specific code in each HTTP client. We can't support GETs with bodies generically for all clients.

This PR simply modifies the DefaultSdkHttpRequest to drop (silently) any body on a GET request. This is one option that allows us to make an easy change that would apply to all clients and any client that gets plugged in.

We could move this logic to the async and sync code paths rather than modifying the request pojo. It would likely allow us to log a little better.

We might be able to add support for this on a per client basis. Even if we could in all 3 of our clients, its not something people expect and since we are pluggable, its an edge case developers won't be aware of.

### DIFF
--- a/http-client-spi/src/main/java/software/amazon/awssdk/http/DefaultSdkHttpFullRequest.java
+++ b/http-client-spi/src/main/java/software/amazon/awssdk/http/DefaultSdkHttpFullRequest.java
@@ -56,7 +56,7 @@ final class DefaultSdkHttpFullRequest implements SdkHttpFullRequest {
         this.queryParameters = deepUnmodifiableMap(builder.queryParameters, () -> new LinkedHashMap<>());
         this.httpMethod = Validate.paramNotNull(builder.httpMethod, "method");
         this.headers = deepUnmodifiableMap(builder.headers, () -> new TreeMap<>(String.CASE_INSENSITIVE_ORDER));
-        this.contentStreamProvider = builder.contentStreamProvider;
+        this.contentStreamProvider = standardizeBody(builder.contentStreamProvider);
     }
 
     private String standardizeProtocol(String protocol) {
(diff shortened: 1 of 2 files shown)


dagnir:
This is tangential, but when we begin to support generating clients for API Gateway, this strategy will mean we don't support GET w/ bodies for those API's as well. This is true for v1 today so that's fine, but I just wanted to point that out to make sure we're okay with for v2 as well.

http-client-spi/src/main/java/software/amazon/awssdk/http/DefaultSdkHttpFullRequest.java
```diff
@@ -97,6 +97,14 @@ private Integer standardizePort(Integer port) {
         return port;
     }
 
+    private ContentStreamProvider standardizeBody(ContentStreamProvider contentStreamProvider) {
+        if (httpMethod.equals(SdkHttpMethod.GET)) {
```
millems: Should this also apply to DELETE and HEAD?

http-client-spi/src/main/java/software/amazon/awssdk/http/DefaultSdkHttpFullRequest.java
```diff
@@ -97,6 +97,14 @@ private Integer standardizePort(Integer port) {
...
+            return null;
```
millems: Should we warn or throw an exception if there was some body that got ignored because of the method?

millems:
Does this break any existing APIs?

millems:
Changelog?

spfink:
Closing in favor of #1187

PR #1131, opened 2019-03-13
Improved integration testing.

millems:
1. Increase waiter timeouts and added more waiters to reduce integration test failures in us-east-1.
2. Delete integration tests that are validating service (not SDK) behavior.
3. Run integ-test.yml buildspec with -T1C.


millems:
I preserved the SendEmptyListIntegrationTest tests, since it seems to be related to SDK behavior.

PR #1134, opened 2019-03-15
Add getUrl method

varunnvs92:
* Add `S3Utilities` class with `getUrl` method
* Add `utilites()` method in low-level client that returns an instance of `S3Utilities` with the configuration set on the client
* Added unit tests for `getUrl` method

### DIFF
--- a/.changes/next-release/feature-AmazonS3-f2cd9dd.json
+++ b/.changes/next-release/feature-AmazonS3-f2cd9dd.json
@@ -1,0 +1,5 @@
+{
+    "category": "Amazon S3", 
+    "type": "feature", 
+    "description": "Add support for getUrl operation. The API can be used to generate a URL that represents an object in Amazon S3. The url can only be used to download the object content if the object has public read permissions. Original issue: https://github.com/aws/aws-sdk-java-v2/issues/860"
+}
--- a/codegen/src/main/java/software/amazon/awssdk/codegen/poet/client/AsyncClientClass.java
+++ b/codegen/src/main/java/software/amazon/awssdk/codegen/poet/client/AsyncClientClass.java
@@ -44,6 +44,7 @@ import software.amazon.awssdk.awscore.client.handler.AwsAsyncClientHandler;
 import software.amazon.awssdk.awscore.client.handler.AwsClientHandlerUtils;
 import software.amazon.awssdk.awscore.eventstream.EventStreamTaggedUnionJsonMarshaller;
 import software.amazon.awssdk.codegen.emitters.GeneratorTaskParams;
+import software.amazon.awssdk.codegen.model.config.customization.UtilitiesMethod;
 import software.amazon.awssdk.codegen.model.intermediate.IntermediateModel;
 import software.amazon.awssdk.codegen.model.intermediate.MemberModel;
 import software.amazon.awssdk.codegen.model.intermediate.OperationModel;
@@ -113,6 +114,10 @@ public final class AsyncClientClass extends AsyncClientInterface {
 
         if (model.containsRequestSigners() || model.containsRequestEventStreams()) {
             classBuilder.addMethod(applySignerOverrideMethod(poetExtensions, model));
+        }
+
+        if (model.getCustomizationConfig().getUtilitiesMethod() != null) {
+            classBuilder.addMethod(utilitiesMethod());
         }
 
         model.getEndpointOperation().ifPresent(
@@ -281,5 +286,19 @@ public final class AsyncClientClass extends AsyncClientInterface {
 
     private TypeName eventStreamType(ShapeModel shapeModel) {
         return poetExtensions.getModelClass(shapeModel.getShapeName());
+    }
+
+    private MethodSpec utilitiesMethod() {
+        UtilitiesMethod config = model.getCustomizationConfig().getUtilitiesMethod();
+        ClassName returnType = PoetUtils.classNameFromFqcn(config.getReturnType());
+
+        return MethodSpec.methodBuilder(UtilitiesMethod.METHOD_NAME)
+                         .returns(returnType)
+                         .addModifiers(Modifier.PUBLIC)
+                         .addAnnotation(Override.class)
+                         .addStatement("return $T.create($L)",
+                                       returnType,
+                                       config.getCreateMethodParams().stream().collect(Collectors.joining(",")))
+                         .build();
     }
 }
--- a/codegen/src/test/resources/software/amazon/awssdk/codegen/poet/client/c2j/json/customization.config
+++ b/codegen/src/test/resources/software/amazon/awssdk/codegen/poet/client/c2j/json/customization.config
@@ -8,5 +8,9 @@
     "verifiedSimpleMethods" : ["paginatedOperationWithResultKey"],
     "blacklistedSimpleMethods" : [
         "eventStreamOperation"
-    ]
+    ],
+  "utilitiesMethod": {
+    "returnType": "software.amazon.awssdk.services.json.JsonUtilities",
+    "createMethodParams": ["param1", "param2", "param3"]
+  }
 }
--- a/services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Utilities.java
+++ b/services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Utilities.java
@@ -1,0 +1,242 @@
+/*
+ * Copyright 2010-2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.services.s3;
+
+import java.net.MalformedURLException;
+import java.net.URI;
+import java.net.URL;
+import java.util.function.Consumer;
+import software.amazon.awssdk.annotations.Immutable;
+import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.annotations.SdkPublicApi;
+import software.amazon.awssdk.awscore.client.config.AwsClientOption;
+import software.amazon.awssdk.awscore.internal.EndpointUtils;
+import software.amazon.awssdk.core.client.config.SdkClientConfiguration;
+import software.amazon.awssdk.core.client.config.SdkClientOption;
+import software.amazon.awssdk.core.exception.SdkException;
+import software.amazon.awssdk.http.SdkHttpFullRequest;
+import software.amazon.awssdk.http.SdkHttpMethod;
+import software.amazon.awssdk.http.SdkHttpRequest;
+import software.amazon.awssdk.protocols.core.OperationInfo;
+import software.amazon.awssdk.protocols.core.PathMarshaller;
+import software.amazon.awssdk.protocols.core.ProtocolUtils;
+import software.amazon.awssdk.regions.Region;
+import software.amazon.awssdk.services.s3.internal.S3EndpointUtils;
+import software.amazon.awssdk.services.s3.model.GetUrlRequest;
+import software.amazon.awssdk.utils.Validate;
+
+/**
+ * Utilities for working with Amazon S3 objects. An instance of this class can be created by:
+ * <p>
+ * 1) Directly using the {@link #builder()} method. You have to manually specify the configuration params like region,
+ * s3Configuration on the builder.
+ *
+ * <pre>
+ * S3Utilities utilities = S3Utilities.builder().region(Region.US_WEST_2).build()
+ * GetUrlRequest request = GetUrlRequest.builder().bucket("foo-bucket").key("key-without-spaces").build()
+ * URL url = pathStyleUtilities.getUrl(request);
+ * </pre>
+ * </p>
+ *
+ * <p>
+ * 2) Using the low-level client {@link S3Client#utilities()} method. This is recommended as SDK will use the same
+ * configuration from the {@link S3Client} object to create the {@link S3Utilities} object.
+ *
+ * <pre>
+ * S3Client s3client = S3Client.create();
+ * S3Utilities utilities = s3client.utilities();
+ * GetUrlRequest request = GetUrlRequest.builder().bucket("foo-bucket").key("key-without-spaces").build()
+ * URL url = pathStyleUtilities.getUrl(request);
+ * </pre>
+ * </p>
+ *
+ * Note: This class does not make network calls.
+ */
+@Immutable
+@SdkPublicApi
+public final class S3Utilities {
+
+    private final Region region;
+
+    private final S3Configuration s3Configuration;
+
+    /**
+     * SDK currently validates that region is present while constructing {@link S3Utilities} object.
+     * This can be relaxed in the future when more methods are added that don't use region.
+     */
+    private S3Utilities(Builder builder) {
+        this.region = Validate.paramNotNull(builder.region, "Region");
+        this.s3Configuration = builder.s3Configuration;
+    }
+
+    /**
+     * Creates a builder for {@link S3Utilities}.
+     */
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    // Used by low-level client
+    @SdkInternalApi
+    static S3Utilities create(SdkClientConfiguration clientConfiguration) {
+        return S3Utilities.builder()
+                          .region(clientConfiguration.option(AwsClientOption.AWS_REGION))
+                          .s3Configuration((S3Configuration) clientConfiguration.option(SdkClientOption.SERVICE_CONFIGURATION))
+                          .build();
+    }
+
+    /**
+     * Returns the URL for an object stored in Amazon S3.
+     *
+     * If the object identified by the given bucket and key has public read permissions,
+     * then this URL can be directly accessed to retrieve the object's data.
+     *
+     * <p>
+     *     If same configuration options are set on both #GetUrlRequest and #S3Utilities objects (for example: region),
+     *     the configuration set on the #GetUrlRequest takes precedence.
+     * </p>
+     *
+     * <p>
+     *     This is a convenience which creates an instance of the {@link GetUrlRequest.Builder} avoiding the need to
+     *     create one manually via {@link GetUrlRequest#builder()}
+     * </p>
+     *
+     * @param getUrlRequest A {@link Consumer} that will call methods on {@link GetUrlRequest.Builder} to create a request.
+     * @return A URL for an object stored in Amazon S3.
+     * @throws MalformedURLException Generated Url is malformed
+     */
+    public URL getUrl(Consumer<GetUrlRequest.Builder> getUrlRequest) {
+        return getUrl(GetUrlRequest.builder().applyMutation(getUrlRequest).build());
+    }
+
+    /**
+     * Returns the URL for an object stored in Amazon S3.
+     *
+     * If the object identified by the given bucket and key has public read permissions,
+     * then this URL can be directly accessed to retrieve the object's data.
+     *
+     * <p>
+     *     If same configuration options are set on both #GetUrlRequest and #S3Utilities objects (for example: region),
+     *     the configuration set on the #GetUrlRequest takes precedence.
+     * </p>
+     *
+     * @param getUrlRequest request to construct url
+     * @return A URL for an object stored in Amazon S3.
+     * @throws MalformedURLException Generated Url is malformed
+     */
+    public URL getUrl(GetUrlRequest getUrlRequest) {
+        Region resolvedRegion = resolveRegionForGetUrl(getUrlRequest);
+        URI resolvedEndpoint = resolveEndpoint(getUrlRequest.endpoint(), resolvedRegion);
+
+        SdkHttpFullRequest marshalledRequest = createMarshalledRequest(getUrlRequest, resolvedEndpoint);
+
+        SdkHttpRequest httpRequest = S3EndpointUtils.applyEndpointConfiguration(marshalledRequest,
+                                                                                getUrlRequest,
+                                                                                resolvedRegion,
+                                                                                s3Configuration,
+                                                                                getUrlRequest.bucket());
+
+        try {
+            return httpRequest.getUri().toURL();
+        } catch (MalformedURLException exception) {
+            throw SdkException.create(String.format("Generated URI is malformed: " + httpRequest.getUri()),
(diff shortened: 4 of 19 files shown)


services/s3/src/main/java/software/amazon/awssdk/services/s3/model/GetUrlRequest.java
```diff
@@ -0,0 +1,240 @@
...
+
+/**
+ * Request to generate a URL representing an object in Amazon S3.
+ *
+ * If the object identified by the given bucket and key has public read permissions,
+ * then this URL can be directly accessed to retrieve the object's data.
+ */
+@SdkPublicApi
+public final class GetUrlRequest extends SdkRequest implements ToCopyableBuilder<GetUrlRequest.Builder, GetUrlRequest> {
+    private static final SdkField<String> BUCKET_FIELD = SdkField
```
varunnvs92: Need these fields as this class extend `SdkRequest` which is required to create `ModifyHttpRequest` instance. Also `BUCKET_FIELD` is used by `EndpointAddressInterceptor`

services/s3/src/main/resources/codegen-resources/customization.config
```diff
@@ -51,5 +51,12 @@
     "GetBucketNotification",
     "PutBucketLifecycle",
     "PutBucketNotification"
-  ]
+  ],
+  "utilitiesMethod": {
+    "name": "utilities",
```
dagnir: Should we remove this param so there's no chance of them becoming inconsistent across services? Will we ever not name these `utilities`?

varunnvs92: Added incase there is a need to use a new name. I can remove it for now though to ensure consistency.

spfink: +1

services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Utilities.java
```diff
@@ -0,0 +1,231 @@
...
+     * <p>
+     *     If same configuration options are set on both #GetUrlRequest and #S3Utilities objects (for example: region),
+     *     the configuration set on the #GetUrlRequest takes precedence.
+     * </p>
+     *
+     * @param getUrlRequest request to construct url
+     * @return A URL for an object stored in Amazon S3.
+     * @throws MalformedURLException Generated Url is malformed
+     */
+    public URL getUrl(GetUrlRequest getUrlRequest) throws MalformedURLException {
```
spfink: Consumer?

services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Utilities.java
```diff
@@ -0,0 +1,231 @@
...
+        URI resolvedEndpoint = resolveEndpoint(getUrlRequest.endpoint(), resolvedRegion);
+
+        SdkHttpRequest httpRequest = applyEndpointInterceptor(createMarshalledRequest(getUrlRequest, resolvedEndpoint),
+                                                              getUrlRequest,
+                                                              resolvedRegion);
+
+        return httpRequest.getUri().toURL();
+    }
+
+    private Region resolveRegionForGetUrl(GetUrlRequest getUrlRequest) {
```
spfink: Eventually we'll probably want to move some of this logic out so this class doesn't get huge but fine for now since it's all private anyways.

services/s3/src/main/java/software/amazon/awssdk/services/s3/model/GetUrlRequest.java
millems: I don't really agree with this extending `SdkRequest`. Ignoring implementation details, this allows the customer to override api call timeouts, headers, query strings, etc. That doesn't make sense.

codegen/src/main/java/software/amazon/awssdk/codegen/poet/client/AsyncClientClass.java
```diff
@@ -282,4 +287,18 @@ private CodeBlock createEventStreamTaggedUnionJsonMarshaller(ShapeModel eventStr
     private TypeName eventStreamType(ShapeModel shapeModel) {
         return poetExtensions.getModelClass(shapeModel.getShapeName());
     }
+
+    private MethodSpec utilitiesMethod() {
+        UtilitiesMethod config = model.getCustomizationConfig().getUtilitiesMethod();
+        ClassName returnType = PoetUtils.classNameFromFqcn(config.getReturnType());
+
+        return MethodSpec.methodBuilder(config.getName())
```
millems: Does the name need to be configurable, or should it always be utilities() for consistency?

services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Utilities.java
```diff
@@ -0,0 +1,231 @@
...
+
+    private final Region region;
+
+    private final S3Configuration s3Configuration;
+
+    /**
+     * SDK currently validates that region is present while constructing {@link S3Utilities} object.
+     * This can be relaxed in future when more methods are added that doesn't use region.
+     */
+    public S3Utilities(Builder builder) {
```
millems: private?

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/client/c2j/json/customization.config
```diff
@@ -8,5 +8,10 @@
     "verifiedSimpleMethods" : ["paginatedOperationWithResultKey"],
     "blacklistedSimpleMethods" : [
         "eventStreamOperation"
-    ]
+    ],
+  "utilitiesMethod": {
+    "name": "utilities",
+    "returnType": "software.amazon.awssdk.services.json.JsonUtilities",
+    "createMethodParams": ["param1", "param2", "param3"]
```
millems: Should the params just be hard-coded to be the sdk configuration?

services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Utilities.java
```diff
@@ -0,0 +1,231 @@
...
+import software.amazon.awssdk.utils.Validate;
+
+/**
+ * Utilities for working with Amazon S3 objects. An instance of this class can be created by:
+ * 1) Directly using the {@link #builder()} method. You have to manually specify the configuration params like region,
+ * s3Configuration on the builder.
+ * 2) Using the low-level client {@link S3Client#utilities()} method. This is recommended as SDK will use the same
+ * configuration from the {@link S3Client} object to create the {@link S3Utilities} object.
+ *
+ * Note: This class should not have methods that make network calls.
```
millems: Should we add some code examples here too?

Also, nit: Probably better to comment it for the user, not the implementer. "This class does not make network calls."

services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Utilities.java
```diff
@@ -0,0 +1,231 @@
...
+@SdkPublicApi
+public final class S3Utilities {
+
+    private final Region region;
+
+    private final S3Configuration s3Configuration;
+
+    /**
+     * SDK currently validates that region is present while constructing {@link S3Utilities} object.
+     * This can be relaxed in future when more methods are added that doesn't use region.
```
millems: nit:
the future*
don't*

services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Utilities.java
millems: Any reason not to have `GetUrlResponse` for symmetry and extensibility?

services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Utilities.java
```diff
@@ -0,0 +1,231 @@
...
+    public static final class Builder {
+        private Region region;
+
+        private S3Configuration s3Configuration;
+
+        private Builder() {
+        }
+
+        /**
+         * Configure the region with which the SDK should communicate.
```
millems: This comment probably needs to be updated. The SDK won't be doing any communication.

services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Utilities.java
```diff
@@ -0,0 +1,231 @@
...
+         * @return This object for method chaining
+         */
+        public Builder region(Region region) {
+            this.region = region;
+            return this;
+        }
+
+        /**
+         * Sets the S3 configuration options.
```
millems: Can we clarify some examples of why the user might want to set these options without them needing to dig down into the S3Configuration?

services/s3/src/main/java/software/amazon/awssdk/services/s3/model/GetUrlRequest.java
```diff
@@ -0,0 +1,240 @@
...
+        Builder region(Region region);
+
+        /**
+         * Sets the endpoint to use for constructing the URL.
+         *
+         * @param endpoint
+         *        The endpoint to use for constructing the URL.
+         * @return Returns a reference to this object so that method calls can be chained together.
+         */
+        Builder endpoint(URI endpoint);
```
millems: Should this be endpointOverride to match the semantics of the client builder's builder.endpointOverride?

.changes/next-release/feature-AmazonS3-f2cd9dd.json
```diff
@@ -0,0 +1,5 @@
+{
+    "category": "Amazon S3", 
+    "type": "feature", 
+    "description": "Add support for getUrl operation. The API can be used to generate a URL that represents an object in Amazon S3. The url can only be used to download the object content if the object has public read permissions."
```
zoewangg: nit, can we link the issue here?

services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Utilities.java
```diff
@@ -0,0 +1,231 @@
...
+    /**
+     * Creates a builder for {@link S3Utilities}.
+     */
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    // Used by low-level client
+    @SdkInternalApi
+    static S3Utilities create(SdkClientConfiguration clientConfiguration) {
```
millems: Should we have a zero-arg create() method that uses the default client configuration? We could even cheat a bit and do...

```Java
return S3Client.builder().httpClient(NoopHttpClient.create()).build().utilities();
```

services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Utilities.java
varunnvs92: Will add

services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Utilities.java
zoewangg: nit, missing `<p>`. Can we use html tags like `<ul>` to make it more readable in javadocs?

codegen/src/main/java/software/amazon/awssdk/codegen/poet/client/AsyncClientClass.java
varunnvs92: Removed the "name" from configuration for now to maintain consistency.

services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Utilities.java
varunnvs92: Good catch

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/client/c2j/json/customization.config
varunnvs92: I want to keep it configurable as we can't predict that future utilities just take sdk configuration.

services/s3/src/test/java/software/amazon/awssdk/services/s3/S3UtilitiesTest.java
```diff
@@ -0,0 +1,139 @@
...
+                                                                                     .accelerateModeEnabled(true)
+                                                                                     .dualstackEnabled(true)
+                                                                                     .checksumValidationEnabled(true)
+                                                                                     .build();
+    private static final S3Configuration PATH_STYLE_CONFIG = S3Configuration.builder()
+                                                                   .pathStyleAccessEnabled(true)
+                                                                   .build();
+
+    private static final S3Client defaultClient = S3Client.create();
+    private static final S3Utilities defaultUtilities = defaultClient.utilities();
```
zoewangg: Can we add the assertions in the same tests here to cover utilities for async clients?

services/s3/src/main/java/software/amazon/awssdk/services/s3/model/GetUrlRequest.java
```diff
@@ -0,0 +1,240 @@
...
+
+        @Override
+        public Builder endpoint(URI endpoint) {
+            this.endpoint = endpoint;
+            return this;
+        }
+
+        @Override
+        public RequestOverrideConfiguration overrideConfiguration() {
+            return null;
```
zoewangg: This is odd.. We should probably not make it extending SdkRequest

services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Utilities.java
varunnvs92: Updated.

services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Utilities.java
varunnvs92: I think there was some pushback in the design meeting (my initial pr had `GetUrlResponse`) as the only use case for this api was to generate a url.

services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Utilities.java
```diff
@@ -0,0 +1,256 @@
...
+     * <p>
+     *     This is a convenience which creates an instance of the {@link GetUrlRequest.Builder} avoiding the need to
+     *     create one manually via {@link GetUrlRequest#builder()}
+     * </p>
+     *
+     * @param getUrlRequest A {@link Consumer} that will call methods on {@link GetUrlRequest.Builder} to create a request.
+     * @return A URL for an object stored in Amazon S3.
+     * @throws MalformedURLException Generated Url is malformed
+     */
+    public URL getUrl(Consumer<GetUrlRequest.Builder> getUrlRequest) throws MalformedURLException {
```
zoewangg: If we are creating the URL, should we make sure `MalformedURLException` is not thrown by validating the input?

services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Utilities.java
```diff
@@ -0,0 +1,256 @@
...
+        ExecutionAttributes executionAttributes = new ExecutionAttributes();
+        executionAttributes.putAttribute(AwsSignerExecutionAttribute.SERVICE_CONFIG, s3Configuration);
+        executionAttributes.putAttribute(AwsExecutionAttribute.AWS_REGION, resolvedRegion);
+
+        EndpointAddressInterceptor interceptor = new EndpointAddressInterceptor();
+        return interceptor.modifyHttpRequest(constructModifyHttpRequest(sdkHttpFullRequest, getUrlRequest),
+                                             executionAttributes);
+    }
+
+    // Used as input for EndpointAddressInterceptor#modifyHttpRequest method
```
zoewangg: nit, let's use /**/ to add comments on method.

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/client/c2j/json/customization.config
millems: Since this is internal, can we just YAGNI it and add it if we need in the future?

services/s3/src/main/java/software/amazon/awssdk/services/s3/model/GetUrlRequest.java
varunnvs92: I thought of that but didn't understand why we went with `endpointOverride`. For me, `endpoint` seems better and non-confusing.

services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Utilities.java
varunnvs92: +1

services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Utilities.java
varunnvs92: Done

services/s3/src/main/java/software/amazon/awssdk/services/s3/model/GetUrlRequest.java
millems: `endpointOverride` was used because it overrides the endpoint that would be generated from the region with a custom endpoint.

services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Utilities.java
varunnvs92: `create` method is for internal use. I prefer to avoid direct defaults in the S3Utilities class and let customer configure the fields they want or use the S3Client utilities method.

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/client/c2j/json/customization.config
varunnvs92: why to remove a supported feature? As its internal, changes to it shouldn't affect customers.

services/s3/src/main/java/software/amazon/awssdk/services/s3/model/GetUrlRequest.java
varunnvs92: Refactored

services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Utilities.java
varunnvs92: It is thrown when converting URI to URL `httpRequest.getUri().toURL()`. There is nothing to avoid this unless returning URI as output instead of URL.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/S3EndpointUtils.java
```diff
@@ -0,0 +1,152 @@
...
+import software.amazon.awssdk.services.s3.model.CreateBucketRequest;
+import software.amazon.awssdk.services.s3.model.DeleteBucketRequest;
+import software.amazon.awssdk.services.s3.model.ListBucketsRequest;
+
+/**
+ * Utilities for working with Amazon S3 bucket names, such as validation and
+ * checked to see if they are compatible with DNS addressing.
+ */
+@SdkInternalApi
+public final class S3EndpointUtils {
```
varunnvs92: Majority of the code is moved from `EndpointAddressInterceptor` class. Tests are covered by `EndpointAddressInterceptorTest` class

services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Utilities.java
zoewangg: In reality, this exception would not happen since we are the one that are creating the URL for them, right? I just feel we should not put `throws MalformatedURLException` on the signature because if I was the customer, I would not expect `MalformedURLException` to be thrown from this method, can we try catch it and wrap it with SdkClientException?

varunnvs92: I don't want to hide it in the  SdkClientException. If there is malformed url (for whatever reason), its good for customer to handle it.

services/s3/src/test/java/software/amazon/awssdk/services/s3/S3UtilitiesTest.java
varunnvs92: Added a test using async client

services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Utilities.java
zoewangg: Well, we are not exactly hiding it and are just wrapping it with `SdkClientException` so that customers can use the same exception handling logic like other SDK operations.

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/client/c2j/json/customization.config
dagnir: I think another option would be to provide a params object (`ClientUtilityParams`?) instead so we don't need to expand the params.

services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Utilities.java
millems: +1, we shouldn't throw checked exceptions.

services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Utilities.java
millems: I like response objects, since changing return types is impossible, and adding a new method with a different return type will always be confusing, but I don't want to reopen old wounds.

services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Utilities.java
varunnvs92: In addition, we decided that S3Utilities is not a client and so `SdkClientException` doesn't make sense.

millems: SdkException ;)?

millems: Alternatively, IllegalArgumentException or IllegalStateException. They're not passing a URL, so getting back an IllegalURLException seems very weird.

zoewangg: Hmm, I'd imagine `SdkClientException` applies to all **SDK-side** exceptions compared with `SdkServiceException` and doesn't have to be only limited for the actual service client. We are throwing `SdkClientException` in some of our utils methods as well, technically, customers can use the utils by itself and not using an actual service client. 

https://github.com/aws/aws-sdk-java-v2/blob/d17d35f52bd197da33572e5b7854ed54750fd2dc/core/sdk-core/src/main/java/software/amazon/awssdk/core/util/json/JacksonUtils.java#L72

I'm also fine with `SdkException`.

services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Utilities.java
varunnvs92: I don't see us adding a new method with just `GetUrlRequest` type as input.

PR #1136, opened 2019-03-15
Use Unpooled in StreamedRequest#onNext

### LINKED ISSUE #1133: Avoid allocating ByteBuf's on non-SDK threads
## Expected Behavior
We use the default `PooledByteBufAllocator` in Netty which creates per-thread caches stored in a `ThreadLocal`. To avoid polluting threads we don't own, we should ensure`ByteBuf`s allocated by the Netty client are allocated only on SDK threads.

## Current Behavior
It's possible for the Netty client to allocate within a customer's thread, for example, in the `onNext` of an SDK `Subscriber` with a subscription to a customer provided `Publisher`:

https://github.com/aws/aws-sdk-java-v2/blob/1fee753c79558efd8f547fb3c36b773ce57880f3/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/NettyRequestExecutor.java#L436-L446

## Possible Solution
The easiest solution would be to allocate using [`io.netty.buffer.Unpooled`](https://netty.io/4.1/api/io/netty/buffer/Unpooled.html). Otherwise, we should defer the allocation and run it in the `EventLoop`.

## Your Environment

* AWS Java SDK version used: 2.5.10
* JDK version used:

dagnir:
## Description
the onNext method can be called from a non SDK owned thread, and if we allocate
the ByteBuf using the pooling allocator, then we pollute the customer's thread
with a ThreadLocal.

Using Unpooled#wrappedBuffer avoids actually copying the arrays so this should
be more performant as well.

## Motivation and Context
Fixes #1133

## Types of changes

## Checklist

## License


dagnir:
@spfink  Can't think of a good one. This avoids creating the ThreadLocal(s) used by the memory pooling mechanism in Netty so ideally we would check that they aren't created but not sure how besides using reflection to to get the TreadLocal instance and calling `get()`...

PR #1147, opened 2019-03-19
Add sslProvider configuration in NettyNioAsyncHttpClient

zoewangg:
## Description
Add sslProvider configuration in `NettyNioAsyncHttpClient` so that customer can configure it.

Currently, we are using `SslContext.defaultClientProvider()` which will always prefer `openSsl` if available. However, there are cases where we still want to use other ssl providers when openssl is in the classpath such as perf testing.

Another benefit of this PR is that the integ tests and unit tests for async path will run faster with tcnative openssl test dependency. 😉 

## Testing
Added unit tests

## Types of changes

## Checklist

## License


http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/NettyNioAsyncHttpClient.java
```diff
@@ -399,6 +411,19 @@ private void closeEventLoopUninterruptibly(EventLoopGroup eventLoopGroup) throws
...
+         *
+         * <p>If not configured, {@link SslContext#defaultClientProvider()} will be used to determine the SslProvider.
+         *
+         * <p>Note that you might need to add other dependencies if not using JDK's default Ssl Provider.
+         * See https://netty.io/wiki/requirements-for-4.x.html#transport-security-tls
+         *
+         * @param sslProvider the SslProvider
+         * @return the builder of the method chaining.
+         */
+        Builder sslProvider(SslProvider sslProvider);
```
dagnir: What happens if the provider is not available (missing dependency)?

zoewangg: Exception would be thrown when it's sending out the request, for example, `java.lang.UnsatisfiedLinkError: failed to load the required native library` if missing openssl dependency

PR #1152, opened 2019-03-20
Use consistent protocol literal field except for kinesis event stream…

varunnvs92:
* Add serviceId to IntermediateModel metadata
*  Use `jsonProtocolFactory` variable name only for kinesis. Before this fix, any service that doesn't support cbor will generate non-compilable classes if they have event stream APIs


dagnir:
> Before this fix, any service that doesn't support cbor will generate non-compilable classes if they have event stream APIs

Just for my understanding, this only applies to any JSON service right?

varunnvs92:
@dagnir That's correct. We have this customization only in JsonProtocolSpec, so only json (json-rpc, rest-json) services are affected.

PR #1161, opened 2019-03-22
Add benchmarks for async operations

zoewangg:
## Description
Add benchmarks for async operations

- NettyClientH1NonTlsBenchmark: testing against local http server

- NettyHttpClientH1Benchmark: testing against local https server using `SslProvider.OPENSSL` and `SslProvider.JDK`
 
- NettyHttpClientH2Benchmark: testing against local h2 server using `SslProvider.OPENSSL` and `SslProvider.JDK`

There are two benchmark methods in each class:
- concurrentApiCall: making 50 concurrent api calls at the same time.
- sequentialApiCall: making one api call at one time.
 
## Testing
Tested locally and on codebuild

## License


test/sdk-benchmarks/src/main/java/software/amazon/awssdk/benchmark/utils/AlwaysSuccessServlet.java
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * Always succeeds with with a 200 response.
+ */
+public class AlwaysSuccessServlet extends HttpServlet {
+
+    @Override
+    public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+        response.setStatus(HttpStatus.OK_200);
+        response.setContentType("application/json");
+        response.setContentLength(JSON_BODY.getBytes().length);
```
dagnir: nit: Use a specific charset

test/sdk-benchmarks/src/main/java/software/amazon/awssdk/benchmark/utils/BenchmarkUtil.java
```diff
@@ -170,7 +184,46 @@
...
+            countDownLatch.countDown();
+        });
+    }
+
+    public static SslProvider getSslProvider(String sslProviderValue) {
+        SslProvider sslProvider;
+        switch (sslProviderValue) {
+            case DEFAULT_JDK_SSL_PROVIDER:
+                sslProvider = SslProvider.JDK;
+                break;
```
dagnir: nit: can we just return instead of breaking?

test/sdk-benchmarks/src/main/java/software/amazon/awssdk/benchmark/utils/BenchmarkUtil.java
```diff
@@ -15,21 +15,35 @@
...
  */
 @SuppressWarnings("unchecked")
 public final class BenchmarkUtil {
 
-    public static final int PORT_NUMBER = 8089;
-    public static final URI LOCAL_URI = URI.create(String.format("http://localhost:%s", PORT_NUMBER));
+    public static final String DEFAULT_JDK_SSL_PROVIDER = "jdk";
+    public static final String OPEN_SSL_PROVIDER = "openssl";
+
+    public static final int HTTP_PORT_NUMBER = 8089;
```
dagnir: nit: can we move these to a "BenchmarkConstants" class? They're not really utils

test/sdk-benchmarks/src/main/java/software/amazon/awssdk/benchmark/utils/BenchmarkUtil.java
```diff
@@ -15,21 +15,35 @@
...
+    public static final int HTTPS_PORT_NUMBER = 9999;
+    public static final int CONCURRENT_CALLS = 50;
```
dagnir: Any reason why 50?

test/sdk-benchmarks/README.md
```diff
@@ -1,10 +1,29 @@
...
+  
+- From Maven command
 ```bash
+   mvn install -P quick -pl :sdk-benchmarks --am
+   mvn install -pl :bom-internal
+   cd test/sdk-benchmarks
+   mvn exec:exec
+```   
 
+- From command line
```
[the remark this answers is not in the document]
zoewangg: Yup. Will update

test/sdk-benchmarks/src/main/java/software/amazon/awssdk/benchmark/utils/BenchmarkUtil.java
zoewangg: Sounds good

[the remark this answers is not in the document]
zoewangg: Nope, will update

test/sdk-benchmarks/src/main/java/software/amazon/awssdk/benchmark/utils/BenchmarkUtil.java
zoewangg: Keeping it 50 because the default max connection is 50. Don't think we need to use a larger value for regression tests. What do you think?

test/sdk-benchmarks/README.md
```diff
@@ -3,27 +3,31 @@
...
 
+- Using`mvn exec:exec` commands to invoke `BenchmarkRunner` main method
+```bash
+   mvn clean install -P quick -pl :sdk-benchmarks --am
+   mvn clean install -pl :bom-internal
+   cd test/sdk-benchmarks
+   mvn exec:exec
+```   
+
+- From IDE
```
dagnir: Since running from the IDE may be slower, let's add a note about that as well.

zoewangg: Yeah, added more notes in README

PR #1170, opened 2019-03-28
Fix javadoc command by running install first

zoewangg:
## Description
Fix javadoc command by running install first to avoid the following error:
```
Could not resolve dependencies for project software.amazon.awssdk:profiles:jar:2.5.18: Could not find artifact software.amazon.awssdk:test-utils:jar:2.5.18 
``` 

## License


dagnir:
LGTM! Has this been tested? For example disabling the network connection between the two `mvn` commands?

zoewangg:
Tested with snapshot version, which is not relying on network and it worked.

PR #1175, opened 2019-03-28
Update UnusedChannelExceptionHandler to check the cause of the exception

zoewangg:
## Description
Update `UnusedChannelExceptionHandler` to check the cause of the exception so that it does not emit warn logs if the cause is netty io exception. See #1171 

## Testing
Added unit tests

## Types of changes

## Checklist

## License


http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/UnusedChannelExceptionHandler.java
```diff
@@ -78,6 +77,10 @@ public static UnusedChannelExceptionHandler getInstance() {
     }
 
     private boolean isNettyIoException(Throwable cause) {
-        return cause instanceof IOException || cause instanceof ReadTimeoutException || cause instanceof WriteTimeoutException;
+        return cause instanceof IOException || cause instanceof TimeoutException;
+    }
+
+    private boolean hasNettyIoExceptionCause(Throwable cause) {
+        return cause.getCause() != null && isNettyIoException(cause.getCause());
```
millems: Should this be done recursively, checking all causes?

zoewangg: I thought about that, but followed the same pattern with our retryPolicy, which is just checking one level of cause.

https://github.com/aws/aws-sdk-java-v2/blob/master/core/sdk-core/src/main/java/software/amazon/awssdk/core/retry/conditions/RetryOnExceptionsCondition.java#L49

millems: Is the retry policy doing it wrong, as well?

PR #1176, opened 2019-03-29
Add best practices doc

zoewangg:
## Description
Add best practices doc.

~~It might be controversial to add `Prefer high level libraries` section because there is no HLL in v2 yet. ┑(￣Д ￣)┍~~ (Removed)

## License

### DIFF
--- a/docs/BestPractices.md
+++ b/docs/BestPractices.md
@@ -1,0 +1,61 @@
+#  AWS Java SDK 2.x Best Practices
+
+Here are the best practices of using AWS Java SDK 2.x.
+
+### Reuse SDK client if possible
+
+Each SDK client maintains its own HTTP connection pool so that connections can be reused by a new request 
+to cut down the time to establish a new connection. It is recommended to share a single instance of the client to 
+avoid the overhead of having too many connection pools that are not being utilized effectively. All SDK clients are thread safe.
+
+If it is not desirable to share a client instance, invoke `client.close()` to release the resources once the client is not needed.
+
+### Close input streams from client operations
+
+For streaming operations such as `S3Client#getObject`,  if you are working with `ResponseInputStream` directly, we have the following recommendations:
+
+- Read all the data from the input stream as soon as possible
+- Close the input stream as soon as possible
+
+This is because the input stream is a direct stream of data from the HTTP connection and the underlying HTTP connection cannot be
+reused until all data from the stream has been read and the stream is closed. If these rules are not followed, the client can 
+run out of resources by allocating too many open, but unused, HTTP connections.
+
+### Tune HTTP configurations based on performance tests
+
+The SDK provides a set of [default http configurations] that apply to general use cases. Customers are recommended to tune the configurations for their applications based on their use cases.
+
+### Use OpenSSL for Netty async client
+
+By default, `NettyNioAsyncHttpClient` uses JDK as the SslProvider. In our local tests, we found that using `OpenSSL`
+is more performant than JDK. Using OpenSSL is also the recommended approach by Netty community, see [Netty TLS with OpenSSL].
+Note that you need to add `netty-tcnative` to your dependency in order to use OpenSSL, see [Configuring netty-tcnative]
+
+Once you have `netty-tcnative` configured correctly, `NettyAsyncHttpClient` will select OpenSSL automatically, or you can set the
+`SslProvider` explicitly on the NettyNioAsyncHttpClient builder.
+
+```
+   NettyNioAsyncHttpClient.builder()
+                          .sslProvider(SslProvider.OPENSSL)
+                          .build();
+```
+
+### Utilize timeout configurations
+
+The SDK provides timeout configurations on requests out of box and they can be configured via `ClientOverrideConfiguration#apiCallAttemptTimeout` and `ClientOverrideConfiguration#ApiCallTimeout`.
+
+```java
+  S3Client.builder()
+          .overrideConfiguration(b -> b.apiCallTimeout(Duration.ofMillis(API_CALL_TIMEOUT))
+                                       .apiCallAttemptTimeout(Duration.ofMillis(API_CALL_ATTEMPT_TIMEOUT))
+          .build();
+```
+
+- `ApiCallAttemptTimeout` tracks the amount of time for a single http attempt and the request can be retried if timed out on api call attempt. 
+- `ApiCallTimeout` configures the amount of time for the entire execution including all retry attempts. 
+
+By default, timeouts are disabled. Using them together is helpful to set a hard limit on total time spent on all attempts across retries and each individual HTTP request to fail fast on one slow request.
+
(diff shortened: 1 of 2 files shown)


docs/BestPractices.md
```diff
@@ -0,0 +1,44 @@
+#  AWS Java SDK 2.x Best Practices
+
+Here are the best practices of using AWS Java SDK 2.x.
+
+### Reuse SDK client if possible
+
+Each SDK client maintains many resources such as connections and threads, so it is recommended to share a single instance of the client to avoid the overhead of resources management. All SDK clients are thread safe.
```
dagnir: I think we can make this a little more clear. The reason for resource pooling is because creating them is expensive. As it's written, it sounds like the managing them is the expensive part 😛

docs/BestPractices.md
```diff
@@ -0,0 +1,44 @@
...
+- `ApiCallAttemptTimeout` tracks the amount of time for a single http attempt and the request can be retried if timed out on api call attempt. 
+- `ApiCallTimeout` configures the amount of time for the entire execution including all retry attempts. 
+
+Using them together is helpful to set a hard limit on total time spent on all attempts across retries and each individual HTTP request to fail fast on one slow request.
+
+### Turn on wire logging for debugging
+
+Wire logs are helpful to determine the root cause when there is an error. See [how to enable logging](https://docs.aws.amazon.com/sdk-for-java/v2/developer-guide/java-dg-logging.html)
+
+### Prefer high level libraries
```
dagnir: I think this is a little weird to have when we're still developing them. I think we should remove this for now until we have available.

docs/BestPractices.md
```diff
@@ -0,0 +1,44 @@
...
+SDK provides timeout configurations on requests out of box and they can be configured via `ClientOverrideConfiguration#apiCallAttemptTimeout` and `ClientOverrideConfiguration#ApiCallTimeout`. See [example]
+
+- `ApiCallAttemptTimeout` tracks the amount of time for a single http attempt and the request can be retried if timed out on api call attempt. 
+- `ApiCallTimeout` configures the amount of time for the entire execution including all retry attempts. 
+
+Using them together is helpful to set a hard limit on total time spent on all attempts across retries and each individual HTTP request to fail fast on one slow request.
+
+### Turn on wire logging for debugging
+
+Wire logs are helpful to determine the root cause when there is an error. See [how to enable logging](https://docs.aws.amazon.com/sdk-for-java/v2/developer-guide/java-dg-logging.html)
```
dagnir: Not sure if we should have this as part of a best practices doc. Might be better to add it to the "Bug Reports" section of the contributing doc: https://github.com/aws/aws-sdk-java-v2/blob/master/CONTRIBUTING.md#bug-reports

docs/BestPractices.md
```diff
@@ -0,0 +1,44 @@
...
+### Tune HTTP configurations based on performance tests
+
+SDK provides a set of [default http configurations] that apply to general use cases. Customers are recommended to tune the configurations for their applications based on their use cases.
+
+### Use OpenSSL for `NettyAsynHttpClient`
+
+By default, `NettyAsynHttpClient` uses JDK as the SslProvider. In our local tests, we found that using `OpenSSL`
+is more performant than JDK. Using OpenSSL is also the recommended approach by Netty community, see [Netty TLS with OpenSSL].
+
+### Utilize `ApiCallAttemptTimeout` and `ApiCallTimeout` configurations
```
dagnir: We could reword this to just 'Utilize timeout configurations'

docs/BestPractices.md
```diff
@@ -0,0 +1,44 @@
...
+
+### Close input stream after read
+
+For streaming operations such as `S3Client#getObject`,  if you are working with `ResponseInputStream` directly, we recommend closing the input stream as soon as possible so that the underlying HTTP connection can be reused.
+
+### Tune HTTP configurations based on performance tests
+
+SDK provides a set of [default http configurations] that apply to general use cases. Customers are recommended to tune the configurations for their applications based on their use cases.
+
+### Use OpenSSL for `NettyAsynHttpClient`
```
dagnir: Should we have a code sample?

docs/BestPractices.md
```diff
@@ -0,0 +1,44 @@
...
+
+Here are the best practices of using AWS Java SDK 2.x.
+
+### Reuse SDK client if possible
+
+Each SDK client maintains many resources such as connections and threads, so it is recommended to share a single instance of the client to avoid the overhead of resources management. All SDK clients are thread safe. 
+
+If it is not desirable to share a client instance, call `client.close()` to release the resources once the client is not needed.
+
+### Close input stream after read
```
dagnir: This sounds a little ambiguous. How about "Close InputStreams from client operations".

We might also what to add a comment about reading the stream completely, otherwise the connection can't be reused

docs/BestPractices.md
zoewangg: Yeah, the reason I put it here is to make it more discoverable. I guess we can move it to bug-reports.

docs/BestPractices.md
zoewangg: Yeah, will update.

docs/BestPractices.md
```diff
@@ -0,0 +1,56 @@
...
+avoid the overhead of having too many connection pools that are not being utilized effectively. All SDK clients are thread safe.
+
+If it is not desirable to share a client instance, invoke `client.close()` to release the resources once the client is not needed.
+
+### Close input stream from client operations
+
+For streaming operations such as `S3Client#getObject`,  if you are working with `ResponseInputStream` directly, we have the following recommendations:
+
+- Use the data from the input stream as soon as possible
+- Read all data from the stream
```
dagnir: minor: can we combine the first item and this together into something like 'Read all the data from the stream as soon as possible'?

docs/BestPractices.md
```diff
@@ -0,0 +1,56 @@
...
+
+- `ApiCallAttemptTimeout` tracks the amount of time for a single http attempt and the request can be retried if timed out on api call attempt. 
+- `ApiCallTimeout` configures the amount of time for the entire execution including all retry attempts. 
+
+By default, timeouts are disabled. Using them together is helpful to set a hard limit on total time spent on all attempts across retries and each individual HTTP request to fail fast on one slow request.
+
+[default http configurations]: https://github.com/aws/aws-sdk-java-v2/blob/master/http-client-spi/src/main/java/software/amazon/awssdk/http/SdkHttpConfigurationOption.java
+[Netty TLS with OpenSSL]: https://netty.io/wiki/requirements-for-4.x.html#tls-with-openssl
+[Configuring netty-tcnative]: https://netty.io/wiki/forked-tomcat-native.html
+[example]: https://github.com/aws/aws-sdk-java-v2/blob/master/test/protocol-tests/src/test/java/software/amazon/awssdk/protocol/tests/timeout/sync/SyncApiCallTimeoutTest.java#L62
```
debora-ito: Links to specific lines in master branch tend to get out-of-date, maybe link to a tag instead?
Or maybe put the example in the body of the text.

zoewangg: Yeah, good point. Will update

PR #1185, opened 2019-04-02
Add async calls in SimpleMethodsIntegrationTests

zoewangg:
## Description
Add async calls in generated `SimpleMethodsIntegrationTests` to cover async path.

## Testing
Running simple methods integ tests now.

## License


zoewangg:
Okay, many simple methods are broken and need to be blacklisted. Will close for now.

PR #1186, opened 2019-04-02
[Discussion] Add initial S3TransferManager prototype

dagnir:
This commit adds the initial design draft of `S3TransferManager` intended to tease out the surface level API and basic user experience.

Project tenets and goals doc can be found here: https://github.com/aws/aws-sdk-java-v2/tree/master/docs/design/services/s3/transfermanager


docs/design/services/s3/transfermanager/prototype.java
```diff
@@ -0,0 +1,622 @@
...
+     *                 .build())
+     *         .build(),
+     *         Paths.get("/tmp/myFile.txt"));
+     * // Wait for the transfer to complete
+     * myDownload.completionFuture().join();
+     * }
+     * </pre>
+     */
+    default DownloadObject downloadObject(DownloadObjectRequest request, Path file) {
+        return downloadObject(request, ObjectPartTransformerCreator.forFile(file));
```
varunnvs92: Are we going to use parts or ranged Gets for downloads? I vote for ranged Gets as it can support all objects even if they are not uploaded in parts.

dagnir: Yes, the current plan is to use range gets for all downloads with a possible setting to allow parallel downloads only for multipart objects (like in V1) if there's a benefit to that.

docs/design/services/s3/transfermanager/prototype.java
```diff
@@ -0,0 +1,622 @@
...
+/**
+ * Configuration object for {@link S3TransferManager}.
+ */
+public interface TransferManagerConfiguration {
+    /**
+     * The aggregate max upload rate in bytes per second over all active upload
+     * transfers. The default value is {@link Long#MAX_VALUE}, or effectively
+     * unlimited.
+     */
+    long maximumUploadBytesPersSecond();
```
zoewangg: I wonder if it's worth separating download and upload configurations. Currently, there are just too many configurations and we will potentially add more in the future.

docs/design/services/s3/transfermanager/prototype.java
```diff
@@ -0,0 +1,622 @@
...
+    /**
+     * The minimum size for each part of a multipart download.
+     */
+    long minimumDownloadPartSize();
+}
+
+/**
+ * Override configuration for a single transfer.
+ */
+public interface TransferOverrideConfiguration {
```
zoewangg: How do we pass this to a single transfer?

docs/design/services/s3/transfermanager/prototype.java
```diff
@@ -0,0 +1,622 @@
...
+}
+
+/**
+ * A factory capable of creating the streams for individual parts of a given
+ * object to be uploaded to S3.
+ * <p>
+ * There is no ordering guaranatee for when {@link
+ * #streamForPart(PartUploadContext)} is called.
+ */
+public interface ObjectPartStreamCreator {
```
zoewangg: The name seems a bit confusing to me. If I understand it correctly, this is essentially a factory for `AsyncRequestBody`?

docs/design/services/s3/transfermanager/prototype.java
```diff
@@ -0,0 +1,622 @@
...
+    /**
+     * Return the stream for the object part described by given {@link
+     * PartUploadContext}.
+     *
+     * @param context The context describing the part to be uploaded.
+     * @return The part stream.
+     */
+    Publisher<ByteBuffer> streamForPart(PartUploadContext context);
```
zoewangg: Can we make it return `AsyncRequestBody` and rename it to something like `fromXXX`?

docs/design/services/s3/transfermanager/prototype.java
dagnir: Yes +1 to this, I was thinking the same thing yesterday after going through this again.

docs/design/services/s3/transfermanager/prototype.java
dagnir: Yeah the issue is `AsyncRequestBody` is just a stream with an added `contentLength` property, but technically the `TransferManager` dictates that via the multipart upload configurations.

docs/design/services/s3/transfermanager/prototype.java
spfink: Even if using ranged gets we should still download on the part boundaries and S3 should have an API for retrieving the ranges. This should provide significant benefit both to the client and S3 itself.

dagnir: Yup, I think we would prefer to download on part boundaries if the object was uploaded multipart.

docs/design/services/s3/transfermanager/prototype.java
```diff
@@ -0,0 +1,721 @@
...
+ * <pre>
+ * {@code
+ * // Create using all default configuration values
+ * S3TransferManager tm = S3TranferManager.create();
+ *
+ * // Using custom configuration values to set max upload speed to avoid
+ * // saturating the network interface.
+ * S3TransferManager tm = S3TransferManager.builder()
+ *         .configuration(TransferManagerConfiguration.builder()
+ *             .maximumUploadBytesSecond(32 * 1024 * 1024) // 32 MiB
```
spfink: Can we expose a minimum as well?

Additionally, can we remove "Upload" and make it generic for both?

docs/design/services/s3/transfermanager/prototype.java
```diff
@@ -0,0 +1,721 @@
...
+     * <pre>
+     * {@code
+     * // Initiate the transfer
+     * DownloadObject myDownload = tm.downloadObject(DownloadObjectRequest.builder()
+     *         .getObjectRequest(GetObjectRequest.builder()
+     *                 .bucket(BUCKET)
+     *                 .key(KEY)
+     *                 .build())
+     *          // Set the known length of the object to avoid a HeadObject call
+     *         .hints(DownloadHints.builder()
```
spfink: Hint sounds like it might not be accurate to me...

docs/design/services/s3/transfermanager/prototype.java
```diff
@@ -0,0 +1,721 @@
...
+    }
+
+    interface Initiated extends EventContext {
+        /**
+         * The transfer this listener associated with.
+         */
+        Transfer<?> transfer();
+    }
+
+    interface BytesTransferred extends Initiated {
```
spfink: Can we expose an average transfer rate?

docs/design/services/s3/transfermanager/prototype.java
dagnir: Yeah it's a little funky but it is used in some places like SQL queries https://en.wikipedia.org/wiki/Hint_%28SQL%29.

Maybe "KnownProperties" or something along those lines?

docs/design/services/s3/transfermanager/prototype.java
dagnir: Hmm how would minimum work at the TransferManager level? Seems like it might more sense at the individual transfer level.

I'm not sure about a single config for both upload and download. Seems to me like there might be cases where you don't want them to be symmetric; maybe the server is serving other traffic so can't dedicate all outbound bandwidth to uploads for example.

docs/design/services/s3/transfermanager/prototype.java
dagnir: Sounds good

docs/design/services/s3/transfermanager/prototype.java
```diff
@@ -0,0 +1,753 @@
...
+                       .build(),
+                       ObjectPartTransformerCreator.forDirectory(destinationDirectory));
+    }
+
+    /**
+     * Download a set of files with the given prefix from S3.
+     * <p>
+     * This is intended for advanced use not covered by other overloads of
+     * {@code downloadDirectory} as implementing an {@code
+     * AsyncResponseTransformer} can be difficult and error prone. When
```
millems: `AsyncResponseTransformer` -> `ObjectPartTransformerCreator`?

docs/design/services/s3/transfermanager/prototype.java
```diff
@@ -0,0 +1,753 @@
...
+/**
+ * Configuration object for {@link S3TransferManager}.
+ */
+public interface TransferManagerConfiguration {
+    /**
+     * The max number of requests the Transfer Manager will have at any point in
+     * time. This must be less than or equal to the max concurrent setting on
+     * the S3 client.
+     */
+    public int maximumConcurrency();
```
millems: Does this method name match the one in the netty client?

docs/design/services/s3/transfermanager/prototype.java
```diff
@@ -0,0 +1,753 @@
...
+
+    /**
+     * The aggregate max upload rate in bytes per second over all active upload
+     * transfers. The default value is {@link Long#MAX_VALUE}, or effectively
+     * unlimited.
+     */
+    public long maximumUploadBytesPerSecond();
```
millems: Our configuration objects in core use boxed types for configuration values so that we can represent "not configured". Should we do that here?

docs/design/services/s3/transfermanager/prototype.java
```diff
@@ -0,0 +1,753 @@
...
+ * <p>
+ * There is no ordering guarantee for when {@link
+ * #transformerForPart(PartDownloadContext)} invocations. It is invoked when the
+ * response from S3 is received for the given part.
+ */
+public interface ObjectPartTransformerCreator {
+    /**
+     * Return a transformer for downloading a single part of an object.
+     */
+    AsyncResponseTransformer<GetObjectResponse, ?> transformerForPart(MultipartDownloadContext context);
```
millems: Does the creator need to be parameterized with the output type?

docs/design/services/s3/transfermanager/prototype.java
```diff
@@ -0,0 +1,753 @@
...
+    Optional<GetObjectRequest> objectRequest();
+}
+
+/**
+ * Progress listener for a Transfer.
+ * <p>
+ * The SDK guarantees that calls to {@link #transferProgressEvent(EventContext)}
+ * are externally synchronized.
+ */
+public interface TransferProgressListener {
```
millems: When are we going to figure out how this relates to the progress listeners in the SDK core?

docs/design/services/s3/transfermanager/prototype.java
```diff
@@ -0,0 +1,753 @@
...
+         * The transfer this listener associated with.
+         */
+        Transfer<?> transfer();
+    }
+
+    interface BytesTransferred extends Initiated {
+        /**
+         * Average number of bytes transferred per second for this transfer.
+         */
+        long averageTransferBytesPerSecond();
```
millems: To cause additional churn, average feels weird as a value we expose. Couldn't we just expose the duration and total bytes transfered, and they can calculate the average?

docs/design/services/s3/transfermanager/prototype.java
```diff
@@ -0,0 +1,753 @@
...
+    void transferProgressEvent(EventContext ctx);
+
+    interface EventContext {
+    }
+
+    interface Initiated extends EventContext {
+        /**
+         * The transfer this listener associated with.
+         */
+        Transfer<?> transfer();
```
millems: Any reason this isn't just in the base event context?

docs/design/services/s3/transfermanager/prototype.java
```diff
@@ -0,0 +1,753 @@
...
+     */
+    long objectLength();
+}
+
+/**
+ * A download transfer of a single object from S3.
+ */
+public interface DownloadObject extends Transfer<DownloadObject> {
+    @Override
+    CompletableFuture<DownloadObject> completionFuture();
```
millems: Could we do `CompletableFuture<Void>` or something like `CompletableFuture<CompletedDownloadObject>`?

docs/design/services/s3/transfermanager/prototype.java
```diff
@@ -0,0 +1,753 @@
...
+    @Override
+    UploadObjectState pause();
+}
+
+/**
+ * Represents the transfer of one or more objects to or from S3.
+ */
+public interface Transfer<T extends Transfer> {
+
+    CompletableFuture<T> completionFuture();
```
millems: Should we do `CompletableFuture<?>` to avoid needing the parameterization of the transfer? It feels like customers would always use `Transfer<?>` anyway.

docs/design/services/s3/transfermanager/prototype.java
```diff
@@ -0,0 +1,753 @@
...
+
+    /**
+     * Pause this transfer, cancelling any requests in progress.
+     * <p>
+     * The returned state object can be used to resume this transfer at a later
+     * time.
```
millems: Can it? Wouldn't the customer need to know the subtype to actually do that?

docs/design/services/s3/transfermanager/prototype.java
```diff
@@ -0,0 +1,753 @@
...
+}
+
+/**
+ * Encapsulation of the state of the object.
+ */
+public interface TransferState {
+    /**
+     * Persist this state so it can later be resumed.
+     */
+    void persistTo(OutputStream os);
```
millems: How do they "de-persist" it?

docs/design/services/s3/transfermanager/prototype.java
```diff
@@ -0,0 +1,753 @@
...
+ * A request to download an object. The object to download may be specified as
+ * an S3 {@link GetObjectRequest} or as a presigned URL.
+ */
+public class DownloadObjectRequest extends AbstractTransferRequest {
+    /**
+     * The S3 GetObjectRequest representing the object to download.
+     */
+    public Optional<GetObjectRequest> getObjectRequest() {
+        return Optional.ofNullable(getObjectRequest);
+    }
```
millems: Instead of optional for these "either" fields, should we have a legit union type with is*, as* and visitor methods?

docs/design/services/s3/transfermanager/prototype.java
```diff
@@ -0,0 +1,721 @@
...
+     * {@code
+     * // Initiate transfer
+     * DownloadObject myFileDownload = tm.downloadObject(BUCKET, KEY, Paths.get("/tmp/myFile.txt");
+     * // Wait for transfer to complete
+     * myFileDownload().completionFuture().join();
+     * }
+     * </pre>
+     *
+     */
+    default DownloadObject downloadObject(String bucket, String key, Path file) {
```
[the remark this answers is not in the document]
dagnir: Yup, I'm thinking we'll add the overloads, e.g. `downloadObject(String bucket, String key, Outputstream)` as we have them (i.e. matching implementations for `ObjectPartTransformerCreator`).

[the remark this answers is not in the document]
millems: Do we think customers could ever write a good `ObjectPartTransformerCreator`? If not, we could consider hiding it...

docs/design/services/s3/transfermanager/prototype.java
```diff
@@ -0,0 +1,721 @@
...
+     * <pre>
+     * {@code
+     * // Initiate transfer
+     * DownloadObject myFileDownload = tm.downloadObject(myPresignedUrl, Paths.get("/tmp/myFile.txt");
+     * // Wait for transfer to complete
+     * myFileDownload()completionFuture().join();
+     * }
+     * </pre>
+     */
+    default DownloadObject downloadObject(URL presignedUrl, Path file) {
```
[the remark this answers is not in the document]
dagnir: I didn't want to use the same names as the S3 client operations, so threw out `getObject`. In v1 we have `download` and `downloadDirectory`, but I felt `downloadObject` was more symmetric. I think this may be something we discuss with the other SDKs so we get consistency with any transfermanager-like libraries.

docs/design/services/s3/transfermanager/prototype.java
```diff
@@ -0,0 +1,721 @@
...
+     *         .presignedUrl(url)
+     *         .build(),
+     *         ObjectPartStreamFactory.forInputStream(myStream));
+     * myUpload.completionFuture().join();
+     * </pre>
+     *
+     * @param request The upload request.
+     * @param partStreamFactory The part stream factory.
+     */
+    UploadObject uploadObject(UploadObjectRequest request, ObjectPartStreamCreator partStreamFactory);
```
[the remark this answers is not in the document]
dagnir: Yeah it's definitely scary, but I'm not sure how I can be simplified further while leaving it open to support the use cases we want.

docs/design/services/s3/transfermanager/prototype.java
```diff
@@ -0,0 +1,753 @@
...
+     * @param bucket The bucket.
+     * @param prefix The prefix.
+     * @param destinationDirectory The directory where the objects will be
+     * downloaded to.
+     */
+    default DownloadDirectory downloadDirectory(String bucket, String prefix, Path destinationDirectory) {
+        return downloadDirectory(DownloadDirectoryRequest.builder()
+                       .bucket(bucket)
+                       .prefix(prefix)
+                       .build(),
```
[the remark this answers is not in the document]
dagnir: Sounds good

docs/design/services/s3/transfermanager/prototype.java
dagnir: Yup https://github.com/aws/aws-sdk-java-v2/blob/fbec1bf7ba835d7f98d073828d8954c87c2e30bc/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/NettyNioAsyncHttpClient.java#L448

docs/design/services/s3/transfermanager/prototype.java
dagnir: Not in the way it's currently used since the `Transfer` object (or subclasses) doesn't expose the output from the transformer.

docs/design/services/s3/transfermanager/prototype.java
dagnir: I wanted to make sure we could support progress events in the way we want to for TransferManager first, then ideally we could generalize that for the core listeners.

docs/design/services/s3/transfermanager/prototype.java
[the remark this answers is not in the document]
millems: I use `getItem` and `putItem` in the mapper, which matches the low-level client. I can understand the benefit in being different here, though.

I'm not too hung up on the verb, as long as we're consistent. I'm not a fan of 'Object', though. In the Java world it literally could mean anything.

docs/design/services/s3/transfermanager/prototype.java
dagnir: Yeah that would work too by adding a timestamp member to `Initiated`

docs/design/services/s3/transfermanager/prototype.java
dagnir: Hmm I like the `CompletableFuture<CompletedDownloadObject>` idea :)

docs/design/services/s3/transfermanager/prototype.java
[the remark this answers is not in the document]
millems: Maybe calling it `TransferManagerRequestBody` or `TransferRequestBody` or something that better matches the low-level client's name?

docs/design/services/s3/transfermanager/prototype.java
millems: Nit: `maxConcurrency` vs `maximumConcurrency`

docs/design/services/s3/transfermanager/prototype.java
millems: Then are we using the wrong abstraction? If the output of the transformation doesn't matter, does it need to be a transformation, or is it just a `Consumer`?

docs/design/services/s3/transfermanager/prototype.java
millems: Will we do that before we GA transfer-manager?

docs/design/services/s3/transfermanager/prototype.java
dagnir: It's possible, but I was hesitant to simplify it since the `GetObjectResponse` given to the `AsyncResponseTransformer` may be important to the implementation, as well as the retry behavior you get by completing the future returned from `prepare()` appropriately.

docs/design/services/s3/transfermanager/prototype.java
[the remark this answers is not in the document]
dagnir: I'm fine with hiding it for now and exposing later if we want to.

docs/design/services/s3/transfermanager/prototype.java
```diff
@@ -0,0 +1,721 @@
...
+        /**
+         * The custom S3AsyncClient this transfer manager will use to make calls
+         * to S3.
+         */
+        Builder s3client(S3AsyncClient s3Client);
+
+        /**
+         * The configuration for this transfer manager.
+         */
+        Builder configuration(TransferManagerConfiguration configuration);
```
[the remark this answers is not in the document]
dagnir: Sounds good, I'll take a look at how the enhanced client. FWIW, I see `TransferManagerConfiguration` properties as those that directly affect the behavior `TransferManager` itself such as bandwidth limits, which is different from `s3Client` or the progress listeners.

docs/design/services/s3/transfermanager/prototype.java
```diff
@@ -0,0 +1,721 @@
...
+     * effectively unlimited.
+     */
+    public long maximumDownloadBytesSecond();
+
+    /**
+     * The configuration for multipart uploads.
+     */
+    public Optional<MultipartUploadConfiguration> multipartUploadConfiguration() {
+        return Optional.ofNullable(multipartUploadConfiguration);
+    }
```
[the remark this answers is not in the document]
dagnir: Returning default sounds good to me

docs/design/services/s3/transfermanager/prototype.java
dagnir: Woops good catch

docs/design/services/s3/transfermanager/prototype.java
dagnir: Yes definitely

docs/design/services/s3/transfermanager/prototype.java
dagnir: Nope, I think it can be moved up

docs/design/services/s3/transfermanager/prototype.java
dagnir: The subclasses of `Transfer` return the correct type, e.g. `DownloadObjectState`

docs/design/services/s3/transfermanager/prototype.java
dagnir: I'll add a `loadFrom` method :)

docs/design/services/s3/transfermanager/prototype.java
dagnir: +1 for consistency!

docs/design/services/s3/transfermanager/prototype.java
dagnir: I'm in favor of `CompletableFuture<CompletedDownloadObject>` given above instead. I think it'd be nice to be able to expose information about the transfer after it has completed.

millems: Since this is the base class, maybe `CompletableFuture<? extends CompletedTransfer>`?

docs/design/services/s3/transfermanager/prototype.java
millems: Would it be useful to use this base type method?

dagnir: Nope. I think we can remove it and just add it to the concrete implementations

docs/design/services/s3/transfermanager/prototype.java
dagnir: +1

docs/design/services/s3/transfermanager/prototype.java
dagnir: Done. I'm not happy about the name `UploadObjectSpecification`/`DownloadObjectSpecification` but don't have a better one atm.

docs/design/services/s3/transfermanager/prototype.java
dagnir: It's added on the download/upload request object

docs/design/services/s3/transfermanager/prototype.java
```diff
@@ -0,0 +1,864 @@
...
+     * <pre>
+     * {@code
+     * // Initiate transfer
+     * Download myFileDownload = tm.downloadObject(myPresignedUrl, Paths.get("/tmp/myFile.txt");
+     * // Wait for transfer to complete
+     * myFileDownload()completionFuture().join();
+     * }
+     * </pre>
+     */
+    default Download downloadObject(URL presignedUrl, Path file) {
```
millems: Nit: S3 noun Object still feels weird next to the file noun Directory of the other methods.

docs/design/services/s3/transfermanager/prototype.java
```diff
@@ -0,0 +1,864 @@
...
+     * // Initiate the transfer
+     * Download myDownload = tm.downloadObject(DownloadObjectRequest.builder()
+     *         .getObjectRequest(GetObjectRequest.builder()
+     *                 .bucket(BUCKET)
+     *                 .key(KEY)
+     *                 .build())
+     *          // Set the known length of the object to avoid a HeadObject call
+     *         .knownProperties(KnownObjectProperties.builder()
+     *                 .size(1024 * 1024 * 5)
+     *                 .build())
```
millems: Does this need to be a separate configuration? Can't it just be `DownloadObjectRequest.builder().size(...).build()`? We could always use an interface for it if customers want to share code.

docs/design/services/s3/transfermanager/prototype.java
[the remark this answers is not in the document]
millems: Bump

docs/design/services/s3/transfermanager/prototype.java
```diff
@@ -0,0 +1,864 @@
...
+    /**
+     * The minimum size for an object to be uploaded in multipe parts.
+     */
+    public Long multipartUploadThreshold();
+
+    /**
+     * The minimum size for each uploaded part.
+     */
+    public Long minUploadPartSize();
+}
```
millems: No Integer maxUploadPartCount?

docs/design/services/s3/transfermanager/prototype.java
```diff
@@ -0,0 +1,864 @@
...
+public interface TransferOverrideConfiguration {
+    /**
+     * The maximum rate for this transfer in bytes per second.
+     */
+    OptionalLong maxTransferBytesPerSecond();
+
+    /**
+     * Override configuration for multipart downloads.
+     */
+    Optional<MultipartDownloadConfiguration> multipartDownloadConfiguration();
```
millems: Does this need to be optional?

docs/design/services/s3/transfermanager/prototype.java
millems: Cool, we can always work it out later.

docs/design/services/s3/transfermanager/prototype.java
```diff
@@ -0,0 +1,864 @@
...
+}
+
+/**
+ * Metrics for a completed transfer.
+ */
+public interface TransferMetrics {
+    /**
+     * The number of milliseconds that elapsed before this transfer completed.
+     */
+    long elapsedMillis();
```
millems: `Duration`?

docs/design/services/s3/transfermanager/prototype.java
```diff
@@ -0,0 +1,864 @@
...
+public class DownloadObjectRequest extends AbstractTransferRequest {
+    /**
+     * The specification for how to download the object.
+     */
+    DownloadObjectSpecification downloadSpecification();
+
+    /**
+     * Known properties to help optimize an object download.
+     */
+    public Optional<KnownObjectProperties> knownProperties();
```
millems: -1 Optional

docs/design/services/s3/transfermanager/prototype.java
```diff
@@ -0,0 +1,864 @@
...
+
+    /**
+     * Returns {@code true} if this is an API request, {@code false} otherwise.
+     */
+    boolean isApiRequest();
+
+    /**
+     * Visit this object.
+     */
+    void visit(DownloadObjectSpecificationVisitor visitor);
```
millems: Nit: If we end up not using this, we can probably exclude it. I don't know if a customer would want it.

docs/design/services/s3/transfermanager/prototype.java
dagnir: Yeah I think it can just be made part of the request.

docs/design/services/s3/transfermanager/prototype.java
dagnir: Yep, missed this! It should be `download` as well.

docs/design/services/s3/transfermanager/prototype.java
dagnir: +1

docs/design/services/s3/transfermanager/prototype.java
dagnir: Nope, removing.

docs/design/services/s3/transfermanager/prototype.java
dagnir: Removed

docs/design/services/s3/transfermanager/prototype.java
dagnir: Sounds good. Lifted the settings initially from v1, and it doesn't have that config!

docs/design/services/s3/transfermanager/prototype.java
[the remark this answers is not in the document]
dagnir: Fine with moving them to the builder level as in the enhanced DDB client.

PR #1187, opened 2019-04-03
Fail to build for services with unsupported payloads

spfink:
With this change the code generator will no longer allow models to be build that have bodies for GET, DELETE or HEAD requests.

If a body is detected with one of these methods, the core will log a warning. I can switch this to also throwing an error but figured to throw a warning in case someone wanted to use our core and plugabble HTTP to support these.

## Testing
Added test for code generation failure.

## Types of changes

## Checklist

## License


spfink:
To note, this is what we could do post Pinpoint Email being updated to not use GET/DELETE with body. I guess I can make the code generator just log a warning until we Pinpoint Email is changed.

codegen/src/main/java/software/amazon/awssdk/codegen/IntermediateModelBuilder.java
```diff
@@ -157,9 +161,28 @@ public IntermediateModel build() {
...
+
         return trimmedModel;
     }
 
+    /**
+     * Validates that operations that use GET, DELETE or HEAD do not have an HTTP body.
+     */
+    private void validateOperations(IntermediateModel model) {
+        List<String> methods = model.getOperations()
+                                    .entrySet()
```
millems: values().stream()? You're not using the keys below.

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/client/c2j/unsupportedpayloads/service-2.json
```diff
@@ -0,0 +1,2664 @@
...
+      "type":"structure",
+      "members":{
+        "SendingPoolName":{
+          "shape":"PoolName",
+          "documentation":"<p>The name of the dedicated IP pool that you want to associate with the configuration set.</p>"
+        }
+      },
+      "documentation":"<p>Used to associate a configuration set with a dedicated IP pool.</p>"
+    },
+    "Destination":{
```
millems: Do we need all this, or can it be shortened?

core/sdk-core/src/main/java/software/amazon/awssdk/core/client/handler/BaseClientHandler.java
```diff
@@ -108,6 +120,13 @@ private static SdkHttpFullRequest modifyEndpointHostIfNeeded(SdkHttpFullRequest
                               .build();
     }
 
+    private static void warnOnUnsupportedRequests(SdkHttpFullRequest request) {
+        if (NO_BODY_METHODS.contains(request.method()) && request.contentStreamProvider().isPresent()) {
+            SdkStandardLogger.REQUEST_LOGGER.warn(() -> NO_BODY_METHODS +
+                                                        " requests with bodies are not supported by the default SDK clients.");
+        }
```
millems: Is this needed, if we have generator validation?

spfink: I added this in case someone was using our core with their own plugged in client that supported GET requests with a body. Very unlikely to occur and won't occur for any AWS service. Can just throw or remove here if preferred though.

millems: I'd favor throwing or removing. Feels like complexity we're managing for not a lot of gain.

codegen/src/main/java/software/amazon/awssdk/codegen/IntermediateModelBuilder.java
spfink: Updated

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/client/c2j/unsupportedpayloads/service-2.json
spfink: Shortened.

stevenmhood:
Is there a way to know which APIs are impacted by this?  I have a Lambda API call to GetLayerVersionByArn that's failing, and this seems possibly related.  I think it's similar to [1078](https://github.com/aws/aws-sdk-java-v2/issues/1078), but it's unclear.

PR #1188, opened 2019-04-04
Cast and return the result of Validate.isAssignableFrom.

millems:


utils/src/main/java/software/amazon/awssdk/utils/Validate.java
```diff
@@ -564,11 +564,13 @@ public static double exclusiveBetween(final double start, final double end, fina
      * @param values  the optional values for the formatted exception message, null array not recommended
      * @throws IllegalArgumentException if argument can not be converted to the specified class
      */
-    public static void isAssignableFrom(final Class<?> superType, final Class<?> type,
+    public static <T> Class<? extends T> isAssignableFrom(final Class<T> superType, final Class<?> type,
```
varunnvs92: Does this break binary compatibility?

millems: Yes, but:
1. `Validate` is marked as a protected API, so it shouldn't be used outside of our code.
2. `Validate.isAssignableFrom` is only used from `AttributeMap`, which is also in the `utils` module, so binary compatibility shouldn't come into play.

varunnvs92: SdkProtectedApi [docs](https://github.com/aws/aws-sdk-java-v2/blob/f6e6f7d22b51ce8cd85494882d72d4f9d0a99daf/core/annotations/src/main/java/software/amazon/awssdk/annotations/SdkProtectedApi.java#L23) say we should not make breaking changes. But as it is used only from `AttributeMap`, I am okay with the change

millems: Yeah, this one method is technically internal for now. Now that it's used from the mapper, it shouldn't be broken.

PR #1192, opened 2019-04-04
Update buildspec to disable connection pooling for Java9+ build

zoewangg:
## Description
Update buildspec to disable connection pooling for Java9+ build

## License

### DIFF
--- a/buildspecs/build.yml
+++ b/buildspecs/build.yml
@@ -3,12 +3,13 @@ version: 0.2
 phases:
   build:
     commands:
-      - mvn clean install -D maven.wagon.httpconnectionManager.maxPerRoute=2
       - JAVA_VERSION=$(java -version 2>&1 | grep -i version | cut -d'"' -f2 | cut -d'.' -f1-1)
-      - echo $JAVA_VERSION
       - |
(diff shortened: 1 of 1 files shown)


buildspecs/build.yml
```diff
@@ -3,12 +3,13 @@ version: 0.2
 phases:
   build:
     commands:
-      - mvn clean install -D maven.wagon.httpconnectionManager.maxPerRoute=2
       - JAVA_VERSION=$(java -version 2>&1 | grep -i version | cut -d'"' -f2 | cut -d'.' -f1-1)
-      - echo $JAVA_VERSION
       - |
         if [ "$JAVA_VERSION" -ge "9" ]; then
+          mvn clean install -Dmaven.wagon.http.pool=false -T1C
```
millems: Does -Dmaven.wagon.http.pool=false not work for Java 8? I'd like to use this command in all of our buildspecs, since we get connection resets on pushing to maven, etc. as well.

zoewangg: It works for Java 8 but it's slower because there's only one connection around. Let me check how slower it would be with this commad for Java 8

PR #1199, opened 2019-04-09
Moved dynamodb-enhanced to services-custom directory. Improved Javadoc.

millems:
Moved dynamodb-enhanced to services-custom directory, to eventually include all hand-written service-specific libraries that aren't in the service's module. Improved JavaDoc.


pom.xml
```diff
@@ -53,7 +53,7 @@
         <module>aws-sdk-java</module>
         <module>core</module>
         <module>services</module>
-        <module>services-hll/dynamodb-enhanced</module>
+        <module>services-custom/dynamodb-enhanced</module>
```
zoewangg: How about `services-enhanced`?

millems: services-enhanced feels like it'd just be for `ServiceNameEnhancedClient`s. If we decide that `TransferManager` should go in a separate module, it wouldn't make sense here, imo.

services-custom/dynamodb-enhanced/it/main/java/software/amazon/awssdk/enhanced/dynamodb/EnhancedClientIntegrationTest.java
```diff
@@ -36,10 +34,13 @@
...
 import software.amazon.awssdk.testutils.Waiter;
+import software.amazon.awssdk.testutils.service.AwsTestBase;
 
-public class EnhancedClientIntegrationTest {
+public class EnhancedClientIntegrationTest extends AwsTestBase {
     private static final String TABLE = "books-" + UUID.randomUUID();
-    private static final DynamoDbClient dynamo = DynamoDbClient.create();
+    private static final DynamoDbClient dynamo = DynamoDbClient.builder()
+                                                               .credentialsProvider(CREDENTIALS_PROVIDER_CHAIN)
+                                                               .build();
```
zoewangg: I think we need to pass dynamodb to line 78 and create an async dynamodb with credentialsProvider and pass it to line 97

pom.xml
zoewangg: Discussed offline. Yeah, I'm fine with this since this is not customer facing and we can change it later if we want.

services-custom/dynamodb-enhanced/it/main/java/software/amazon/awssdk/enhanced/dynamodb/EnhancedClientIntegrationTest.java
millems: Oops. Thanks.

PR #1210, opened 2019-04-17
Reduced the frequency of 'server failed to send complete response' exceptions when using S3AsyncClient.

millems:
Should help reduce #1122 with S3.

### DIFF
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/FileAsyncResponseTransformer.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/FileAsyncResponseTransformer.java
@@ -121,13 +121,21 @@ public final class FileAsyncResponseTransformer<ResponseT> implements AsyncRespo
             if (byteBuffer == null) {
                 throw new NullPointerException("Element must not be null");
             }
-            
+
+            performWrite(byteBuffer);
+        }
+
+        private void performWrite(ByteBuffer byteBuffer) {
             writeInProgress = true;
+
             fileChannel.write(byteBuffer, position.get(), byteBuffer, new CompletionHandler<Integer, ByteBuffer>() {
                 @Override
                 public void completed(Integer result, ByteBuffer attachment) {
-                    if (result > 0) {
-                        position.addAndGet(result);
+                    position.addAndGet(result);
+
+                    if (byteBuffer.hasRemaining()) {
+                        performWrite(byteBuffer);
+                    } else {
                         synchronized (FileSubscriber.this) {
                             if (closeOnLastWrite) {
                                 close();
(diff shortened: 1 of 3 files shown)


core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/FileAsyncResponseTransformer.java
```diff
@@ -121,21 +121,26 @@ public void onNext(ByteBuffer byteBuffer) {
...
-                            } else {
-                                subscription.request(1);
-                            }
-                            writeInProgress = false;
+                    position.addAndGet(result);
+                    synchronized (FileSubscriber.this) {
+                        if (closeOnLastWrite) {
+                            close();
+                        } else {
+                            subscription.request(1);
```
dagnir: We don't seem to check that the buffer was written entirely before requesting another one. Seems like an edge case (fine with addressing in a different PR).

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/FileAsyncResponseTransformer.java
```diff
@@ -121,21 +121,26 @@ public void onNext(ByteBuffer byteBuffer) {
             if (byteBuffer == null) {
                 throw new NullPointerException("Element must not be null");
             }
-            
+
+            if (byteBuffer.remaining() == 0) {
```
dagnir: Is this just an optimization? The docs say the write finishes immediately if the buffer is empty

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/FileAsyncResponseTransformer.java
millems: Do you know what the expectation is when the buffer was not written entirely?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/FileAsyncResponseTransformer.java
millems: I can update it, then. Missed that part of the docs.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/FileAsyncResponseTransformer.java
millems: Fixed.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/FileAsyncResponseTransformer.java
millems: Fixed.

PR #1213, opened 2019-04-17
Make `BytesWrapper`, parent of `SdkBytes` and `ResponseBytes`, public.

### LINKED ISSUE #1208: Can't call SdkBytes.asByteArray via reflection
## Expected Behavior
To be able to call `SdkBytes.asByteArray` via reflection, necessary for the method to be called from Clojure via interop. See also https://dev.clojure.org/jira/browse/CLJ-1243 and https://bugs.java.com/bugdatabase/view_bug.do?bug_id=4283544.

## Current Behavior
Via clojure reflection:
```
java.lang.IllegalArgumentException
   Can't call public method of non-public class: public final byte[]
   software.amazon.awssdk.core.BytesWrapper.asByteArray()
```

In reproduction case below:
```
Exception in thread "main" java.lang.IllegalAccessException: Class ReflectionFail can not access a member of class software.amazon.awssdk.core.BytesWrapper with modifiers "public final"
	at sun.reflect.Reflection.ensureMemberAccess(Reflection.java:102)
	at java.lang.reflect.AccessibleObject.slowCheckMemberAccess(AccessibleObject.java:296)
	at java.lang.reflect.AccessibleObject.checkAccess(AccessibleObject.java:288)
	at java.lang.reflect.Method.invoke(Method.java:491)
	at ReflectionFail.main(ReflectionFail.java:10)
```

## Possible Solution
Make `BytesWrapper` public, similarly to https://github.com/netty/netty/issues/1780

```java
import software.amazon.awssdk.core.SdkBytes;

import java.lang.reflect.Method;

public class ReflectionFail {
    public static void main(String[] args) throws Exception {
        SdkBytes bytes = SdkBytes.fromUtf8String("test");
       
        // Works
        bytes.asByteArray();

        // Doesn't work:
        Class klass = bytes.getClass();
        Method method = klass.getMethod("asByteArray");
        method.invoke(bytes);
    }
}
```

## Context
This issue came up in trying to convert an `SdkBytes` inside a DynamoDB `AttributeValue` to a byte array using `asByteArray`.

Currently we're working around this by using `method.setAccessible(true);` on the reflected `Method`, which allows us to call `asByteArray` on `SdkBytes` in Clojure via reflection.

## Your Environment

* AWS Java SDK version used: 2.5.25
* JDK version used: 1.8.0_202
* Operating System and version: macOS 10.14.4

millems:
Fixes #1208.

### DIFF
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/BytesWrapper.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/BytesWrapper.java
@@ -25,18 +25,27 @@ import java.nio.charset.CharacterCodingException;
 import java.nio.charset.Charset;
 import java.util.Arrays;
 import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.annotations.SdkPublicApi;
 import software.amazon.awssdk.utils.StringUtils;
 import software.amazon.awssdk.utils.Validate;
 
-@SdkInternalApi
-abstract class BytesWrapper {
+/**
+ * A base class for {@link SdkBytes} and {@link ResponseBytes} that enables retrieving an underlying byte array as multiple
+ * different types, like a byte buffer (via {@link #asByteBuffer()}, or a string (via {@link #asUtf8String()}.
+ */
+@SdkPublicApi
+public abstract class BytesWrapper {
+    private static final byte[] EMPTY_BYTES = new byte[0];
+
     private final byte[] bytes;
 
     // Needed for serialization
+    @SdkInternalApi
     BytesWrapper() {
-        this(null);
+        this(EMPTY_BYTES);
     }
 
+    @SdkInternalApi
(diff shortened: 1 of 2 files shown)


core/sdk-core/src/main/java/software/amazon/awssdk/core/BytesWrapper.java
```diff
@@ -25,18 +25,25 @@
...
+ */
+@SdkPublicApi
+public abstract class BytesWrapper {
     private final byte[] bytes;
 
     // Needed for serialization
+    @SdkInternalApi
     BytesWrapper() {
-        this(null);
+        this(new byte[0]);
```
johnou: Since `new byte[0]` is immutable it can be moved to a constant.

millems: Fixed.

PR #1222, opened 2019-04-23
Add more options for FileAsyncResponseTransformer

dagnir:
## Description
Add the following things to control the behavior of the transformer:

 - position parameter to specify where in the file to write the stream
 - isNewFile parameter to specify whether this is a new file
 - deleteOnFailure to specify whether the file should be deleted on failure

## Motivation and Context
File opening flexibility.

## Testing
Added new tests.

## Types of changes

## Checklist

## License

### DIFF
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/async/AsyncResponseTransformer.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/async/AsyncResponseTransformer.java
@@ -119,6 +119,26 @@ public interface AsyncResponseTransformer<ResponseT, ResultT> {
     }
 
     /**
+     * Creates an {@link AsyncResponseTransformer} that writes all the content to the given file.
+     *
+     * @param path            Path to file to write to.
+     * @param position        The value for the data between the current end of the file and the starting position is
+     *                        undefined.
+     * @param isNewFile       Whether this is a new file. If this is {@code true} and the file already exists, the
+     *                        transformer will complete with an exception.
+     * @param deleteOnFailure Whether the file on disk should be deleted in the event of a failure when writing the
+     *                        stream.
+     * @param <ResponseT> Pojo Response type.
+     * @return AsyncResponseTransformer instance.
+     */
+    static <ResponseT> AsyncResponseTransformer<ResponseT, ResponseT> toFile(Path path,
+                                                                             long position,
+                                                                             boolean isNewFile,
+                                                                             boolean deleteOnFailure) {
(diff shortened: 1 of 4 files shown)


dagnir:
Missed the default delete-on-error behavior. I'll need to fix that.

core/sdk-core/src/main/java/software/amazon/awssdk/core/async/AsyncResponseTransformer.java
```diff
@@ -118,6 +118,26 @@
...
+     *                        undefined.
+     * @param isNewFile       Whether this is a new file. If this is {@code true} and the file already exists, the
+     *                        transformer will complete with an exception.
+     * @param deleteOnFailure Whether the file on disk should be deleted in the event of a failure when writing the
+     *                        stream.
+     * @param <ResponseT> Pojo Response type.
+     * @return AsyncResponseTransformer instance.
+     */
+    static <ResponseT> AsyncResponseTransformer<ResponseT, ResponseT> toFile(Path path,
+                                                                             long position,
```
zoewangg: I wonder if we should use a request wrapper for this. Something like `toFile(Path path, FileOption, option)`? My concern is that we might need to add other parameters in the future.

dagnir: I think in that case case a `Builder` might be a better option, something like

```java
AsyncResponseTransformer.toFileTransformerBuilder()
                        .position(...)
                        .deleteOnFailure(...)
                        .build();
```
We have one for `FileAsyncRequestBody` https://github.com/aws/aws-sdk-java-v2/blob/a7519245f57c1ac0e57b79f094a0274c715726ee/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/FileAsyncRequestBody.java#L103

zoewangg: Like the idea of builder but I'm not sure if we want to put `toFileTransformerBuilder` api to `AsyncResponseTransformer`.

What would `toFile` look like? `toFile(FileAsyncResponseTransformer.Builder builder)`?

dagnir: We can probably just keep `toFile(Path)` as-is, since it's behavior is probably what most customers want. This extra stuff is for less common/more specialized cases I would think (such as use in the transfer manager).

dagnir:
Closing this for now.

PR #1233, opened 2019-04-27
Fix bugs in BufferingSubscriber and FlatteningSubscriber

varunnvs92:
**Bugs**
* BufferingSubscriber doesn't request data when current size is less than buffer size. So the application sits idle until ReadTimeoutException is thrown
* FlatteningSubscriber decrements the demand even when buffer is empty
* FlatteningSubscriber was bit complex to add unit test, so updated an integ test to use the response handler variant that was reported in the original issue (internal)

### DIFF
--- a/services/transcribestreaming/src/it/java/software/amazon/awssdk/services/transcribestreaming/TestResponseHandlers.java
+++ b/services/transcribestreaming/src/it/java/software/amazon/awssdk/services/transcribestreaming/TestResponseHandlers.java
@@ -1,0 +1,85 @@
+package software.amazon.awssdk.services.transcribestreaming;
+
+import static software.amazon.awssdk.core.http.HttpResponseHandler.X_AMZN_REQUEST_ID_HEADER;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import software.amazon.awssdk.core.async.SdkPublisher;
+import software.amazon.awssdk.services.transcribestreaming.model.StartStreamTranscriptionResponse;
+import software.amazon.awssdk.services.transcribestreaming.model.StartStreamTranscriptionResponseHandler;
+import software.amazon.awssdk.services.transcribestreaming.model.TranscriptEvent;
+import software.amazon.awssdk.services.transcribestreaming.model.TranscriptResultStream;
+
+public class TestResponseHandlers {
+
+    private static final Logger log = LoggerFactory.getLogger(TestResponseHandlers.class);
+
+    /**
+     * A simple consumer of events to subscribe
+     */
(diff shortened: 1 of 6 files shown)


services/transcribestreaming/src/it/java/software/amazon/awssdk/services/transcribestreaming/TestResponseHandlers.java
```diff
@@ -0,0 +1,81 @@
...
+    /**
+     * A simple consumer of events to subscribe
+     */
+    public static StartStreamTranscriptionResponseHandler responseHandlerBuilder_Consumer() {
+        return StartStreamTranscriptionResponseHandler.builder()
+                                                      .onResponse(r -> {
+                                                          String idFromHeader = r.sdkHttpResponse()
+                                                                                 .firstMatchingHeader(X_AMZN_REQUEST_ID_HEADER)
+                                                                                 .orElse(null);
+                                                          System.out.println("Received Initial response: " + idFromHeader);
```
spfink: log instead of print?

utils/src/main/java/software/amazon/awssdk/utils/async/FlatteningSubscriber.java
```diff
@@ -48,7 +48,9 @@ public void onSubscribe(Subscription subscription) {
             public void request(long l) {
                 synchronized (lock) {
                     demand.addAndGet(l);
-                    if (currentBatch.isEmpty() && !requestedNextBatch) {
+                    // Execution goes into `if` block only once for the initial request
```
spfink: Not sure I fully follow here, why do we no longer check if currentBatch.isEmpty()

varunnvs92: 1) When the request is called for first time, `requestedNextBatch` is false and `currentBatch` is empty. This is the only time when the code execution goes into the `if` block. As `currentBatch` is always empty for first request, there is no explicit check is needed
2) After first time, requestedNextBatch is always true and so the condition `!requestedNextBatch` will always be evaluated to `false`. So there is no need to again check for `currentBatch.isEmpty()` as its evaluation no longer matter because other condition always evaluates to false.

services/transcribestreaming/src/it/java/software/amazon/awssdk/services/transcribestreaming/TestResponseHandlers.java
varunnvs92: Fixed

varunnvs92: Accidentally updated the old commit, sorry about that. The only changes are replacing System.out with log statements

PR #1235, opened 2019-05-01
Fix OneTimeReadTimeoutHandler behavior

dagnir:
## Description
 - Extend ReadTimeoutHandler for timeout behavior
 - Remove self from pipeline once a message is read

## Motivation and Context
Bug fix.

## Types of changes

## Checklist

## License


http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/OneTimeReadTimeoutHandler.java
```diff
@@ -16,34 +16,24 @@
...
     @Override
-    public void channelRead0(ChannelHandlerContext ctx, Object msg) {
-        ReferenceCountUtil.retain(msg);
-        ctx.pipeline().addFirst(READ_TIMEOUT_HANDLER_NAME, new ReadTimeoutHandler(readTimeoutMillis, timeUnit));
-        ctx.fireChannelRead(msg);
-
-        ctx.pipeline().remove(READ_TIMEOUT_HANDLER_NAME);
+    public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception {
         ctx.pipeline().remove(this);
+        super.channelRead(ctx, msg);
```
zoewangg: Shouldn't it be `channelRead` first and then `remove`?

dagnir: I don't think so. If a downstream handler removes us first (for example if the channel is released and `HandlerRemovingChannelPool` removes it), then `remove` will fail; so we remove the read timeout handler first, then fire the read down the rest of the pipeline.

zoewangg: Oh I see, yeah.

Is there any reason why `super.channelRead` is being used? Should we do `ctx.fireChannelRead(msg)` ?

dagnir: Yeah that would work, but technically this is just a specialized `ReadTimeoutHandler` so I think it makes more sense to just defer to it's `channelRead` impl.

zoewangg: Looking at the source code of `IdleStateHandler`, there's some extra logic there. Probably doesn't matter for now, but I'd lean towards to using `ctx.fireChannelRead` directly just in case there's some broken change introduced in the future(somehow).

```java
    @Override
    public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception {
        if (readerIdleTimeNanos > 0 || allIdleTimeNanos > 0) {
            reading = true;
            firstReaderIdleEvent = firstAllIdleEvent = true;
        }
        ctx.fireChannelRead(msg);
    }

```

dagnir: I think that's more my point though. Not calling `super.channelRead` since there's extra logic there could also break behavior no?

zoewangg: Hmm, yeah, I guess that makes sense.

PR #1236, opened 2019-05-01
Add design for automatic event streaming reconnects.

millems:


docs/design/core/event-streaming/reconnect/README.md
```diff
@@ -0,0 +1,81 @@
...
+```Java
+try (TranscribeStreamingAsyncClient client = TranscribeStreamingAsyncClient.create()) {
+    // ...
+    // Current method (behavior is unchanged)
+    client.startStreamTranscription(audioMetadata,
+                                    audioStream,
+                                    responseHandler);
+    
+    // New method that transparently reconnects on network errors (name to be bikeshed)
+    client.startStreamTranscriptionWithReconnects(audioMetadata,
```
spfink: Not that it matters too much as name is TBD but the name you use here is different than in the Option 1 code.

millems: Good catch.

PR #1239, opened 2019-05-02
Added design for an alternate event stream operation syntax that doesn't require reactive streams.

millems:


docs/design/core/event-streaming/alternate-syntax/README.md
```diff
@@ -0,0 +1,165 @@
...
+
+```Java
+try (TranscribeStreamingAsyncClient client = TranscribeStreamingAsyncClient.create();
+    // Create the connection to transcribe and send the initial request message
+    RunningStartStreamTranscription transcription =
+            client.startStreamTranscription(r -> r.languageCode(LanguageCode.EN_US)
+                                                  .mediaEncoding(MediaEncoding.PCM)
+                                                  .mediaSampleRateHertz(16_000))) {
+    // Asynchronously log response transcription events, as we receive them
+    transcription.readAll(TranscriptEvent.class, e -> System.out.println(e.transcript().results()));
```
spfink: Can we see if we can drop the future compatibility here and drop the need to cast?

millems: That'd be worth talking about for Transcribe and Kinesis. 

It would need to be modeled, since it's something the service needs to opt into, and probably something that should be validated in coralcop - that they can't add anything new.

Let's save that discussion for when we get to actually modeling it?

PR #1242, opened 2019-05-02
Made various documentation improvements.

millems:
1. Added a dedicated 'design' README
2. Updated documentation README to link to the new design README
3. Started some cleanup of the design documentation (standard language, standard heading sizes, etc.)


vinoddin:
Wording lgtm. Would suggest we use tags vs updating the design README for state changes, so it doesnt require manual updates to a doc. Just add links to the filtered list of design PRs in a given state.

Also: each doc should link to this doc to re-confirm the state its in. Something along the lines of "This document is in proposed state. See [link] for what that means". Obvious wordsmithing required, but trust you get what i mean.

millems:
I don't follow the tags suggestion, sorry. I updated the docs to link back to the main README.

docs/README.md
```diff
@@ -1,16 +1,18 @@
-# AWS SDK for Java 2.0 Docs
+# AWS SDK for Java 2.x Documentation
 
-## 1.11 to 2.0 Changelog
-- [Launch Changelog](LaunchChangelog.md)
+## SDK Usage
 
-## New Contributions
-- [Getting Started](GettingStarted.md)
```
spfink: No link anywhere to this anymore?

millems: The GettingStarted.md is linked to by Contributing.md, which is already linked. It was weird to direct people to "getting started" when "contributing" seemed like the right channel.

vinoddin:
I meant: instead of the section in the README.md that summarizes docs that are in stages like "Published", "In progress", etc, there should be Github links that map to files that have been tagged as being in those states. That way we don't need to keep the readme.md file up-to-date manually. 

Tagging is one suggestion, any mechanism that only requires changing metadata of the design doc in question and does not require also updating readme.md would do.

millems:
@vinoddin How does that work?

PR #1243, opened 2019-05-03
Not Invoke onError from publisher if subscrber cancels subscription

zoewangg:
## Description
When subscription is cancelled from the subscriber, we should not invoke `SdkAsyncHttpResponseHandler#onError` from the publisher and we should let the subscriber to handle the cancellation.

## Motivation and Context
When subscription is cancelled in the subscriber, we are invoking `SdkAsyncHttpResponseHandler#onError` with SdkCancellationException. This would cause the original exception to be swallowed.
```
                        SdkCancellationException e = new SdkCancellationException(
                                "Subscriber cancelled before all events were published");
                        requestContext.handler().onError(e);
                        executeFuture.completeExceptionally(e);
```
Stacktrace:
```
Caused by: software.amazon.awssdk.core.exception.SdkClientException
    at software.amazon.awssdk.core.exception.SdkClientException$BuilderImpl.build(SdkClientException.java:97)
    at software.amazon.awssdk.core.internal.util.ThrowableUtils.asSdkException(ThrowableUtils.java:98)
    at software.amazon.awssdk.core.internal.http.pipeline.stages.AsyncRetryableStage$RetryExecutor.retryIfNeeded(AsyncRetryableStage.java:123)
    ... 15 more
Caused by: software.amazon.awssdk.http.SdkCancellationException: Subscriber cancelled before all events were published
    at software.amazon.awssdk.http.nio.netty.internal.ResponseHandler$PublisherAdapter$1.onCancel(ResponseHandler.java:241)
    ... 8 more
```

## Testing
Added new test and tested with the repro case locally.

Will run the whole integ test suites.

## Types of changes

## Checklist

## License


zoewangg:
This integ test got stuck with the change
https://github.com/aws/aws-sdk-java-v2/blob/master/services/kinesis/src/it/java/software/amazon/awssdk/services/kinesis/SubscribeToShardIntegrationTest.java#L117

services/kinesis/src/it/java/software/amazon/awssdk/services/kinesis/SubscribeToShardIntegrationTest.java
```diff
@@ -114,61 +115,59 @@ public void subscribeToShard_ReceivesAllData() {
...
+                                             errorOccurred.set(true);
+                                         }
+
+                                         @Override
+                                         public void complete() {
+                                             countDownLatch.countDown();
+                                         }
+                                     }).join();
+
+        assertThat(countDownLatch.getCount()).isEqualTo(0);
```
dagnir: I don't think this is the correct behavior, count should still be 2. Cancelling the subscription should not call the `complete` since that should only be signaled when the full stream is read.

core/aws-core/src/main/java/software/amazon/awssdk/awscore/eventstream/EventStreamAsyncResponseTransformer.java
```diff
@@ -431,6 +431,11 @@ public void request(long l) {
                     @Override
                     public void cancel() {
                         dataSubscription.cancel();
+
+                        // Need to complete the futures, otherwise the downstream subscriber will never
```
dagnir: Hmm not sure about this behavior. Presumably the user called `cancel` so why do they need to be notified via `future`?

zoewangg: Yeah, I think we'd still want to complete the future because if customer subscriber fails to complete the future when cancelling the subscription, it will hang forever.

dagnir: My concern is that it would be inconsistent with streaming (not event streaming) API's; do we complete the future for an operation like `GetObject` if the subscription is cancelled?

zoewangg: It would depend on the implementation of the `AsyncRresponseTransformer`. In `FileAsyncResponseTransformer`, we do complete the future exceptionally after cancelling but we don't in `ByteArrayAsyncResponseTransformer`. 

I don't think we can expect all custom `AsyncResponseTransformer` to complete the future and we can at least make sure the future doesn't hang if they don't complete the future by themselves.

https://github.com/aws/aws-sdk-java-v2/blob/e95033056ac5daf2da2556b55aee6778b261b45f/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/FileAsyncResponseTransformer.java#L153-L155

dagnir: Right, but I don't think we defacto always complete the future returned by the operation:

```
CompletableFuture cf = s3.getObject(...);
```

`cf` is not the same future returned by the transformer, and we don't complete it when the subscription is cancelled I believe.

dagnir: Spoke offline. After thinking about this more, I think this is fine given that the customer doesn't have direct access to the future returned from an even streaming operation.

zoewangg:
Stability tests and integration tests passed.

PR #1247, opened 2019-05-08
Support requiresLength trait for streaming payloads. 

varunnvs92:
~~**This is a draft. Ignore any missing tests, formatting issues**~~
PR is ready for review.

* Support requiresLength trait for streaming payloads. 
* Set request headers based on requiresLength and authType trait
* Core logic is in `StreamingRequestMarshaller` and `AsyncStreamingRequestMarshaller`

**Header resolution logic:** 

-  If content-length is present, always use Content-Length header
- If content-length is not present:
    - If `requiresLength` is present, throw an exception to fail-fast and let user know Content-Length is required. This can only happen for async requests when custom publisher is used. Based on our team discussion, we didn't want to compute the Content-Length in this case to avoid complexity of wrapping publisher and computing content-length
    - If `transferEncoding` (determined at generation based on `v4-unsigned-body`Auth type) is present and not `http2`, set `transfer-encoding: chunked` header
    - For all other cases, keep the behavior same as now. Underlying http clients can determine whether to send `Content-Length` or `Transfer-Encoding` and this is fine as services accept both headers. This is better than computing Content-Length as there will be performance impact if SDK has to read data to find content length. (More information on this in internal doc)

### DIFF
--- a/codegen/src/main/java/software/amazon/awssdk/codegen/poet/client/specs/ProtocolSpec.java
+++ b/codegen/src/main/java/software/amazon/awssdk/codegen/poet/client/specs/ProtocolSpec.java
@@ -15,6 +15,7 @@
 
 package software.amazon.awssdk.codegen.poet.client.specs;
 
+import com.squareup.javapoet.ClassName;
 import com.squareup.javapoet.CodeBlock;
 import com.squareup.javapoet.FieldSpec;
 import com.squareup.javapoet.MethodSpec;
@@ -27,8 +28,11 @@ import software.amazon.awssdk.codegen.model.intermediate.IntermediateModel;
 import software.amazon.awssdk.codegen.model.intermediate.OperationModel;
 import software.amazon.awssdk.codegen.model.intermediate.ShapeModel;
 import software.amazon.awssdk.codegen.model.intermediate.ShapeType;
+import software.amazon.awssdk.codegen.model.service.AuthType;
 import software.amazon.awssdk.codegen.poet.PoetExtensions;
 import software.amazon.awssdk.core.client.handler.SyncClientHandler;
+import software.amazon.awssdk.core.runtime.transform.AsyncStreamingRequestMarshaller;
+import software.amazon.awssdk.core.runtime.transform.StreamingRequestMarshaller;
 import software.amazon.awssdk.protocols.core.ExceptionMetadata;
 import software.amazon.awssdk.utils.StringUtils;
 
@@ -90,5 +94,49 @@ public interface ProtocolSpec {
         return opModel.getEndpointDiscovery() != null
                ? ".discoveredEndpoint(cachedEndpoint)\n"
                : "";
+    }
+
+
+    /**
+     * For sync streaming operations, wrap request marshaller in {@link StreamingRequestMarshaller} class.
+     */
+    default CodeBlock syncStreamingMarshaller(IntermediateModel model, OperationModel opModel, ClassName marshaller) {
+        return streamingMarshallerCode(model, opModel, marshaller, "protocolFactory", false);
+    }
+
+    default CodeBlock asyncMarshaller(IntermediateModel model, OperationModel opModel, ClassName marshaller,
+                                     String protocolFactory) {
+        if (opModel.hasStreamingInput()) {
+            return streamingMarshallerCode(model, opModel, marshaller, protocolFactory, true);
+        } else {
+            return CodeBlock.builder().add("new $T($L)", marshaller, protocolFactory).build();
+        }
+    }
+
+    default CodeBlock streamingMarshallerCode(IntermediateModel model, OperationModel opModel, ClassName marshaller,
+                                              String protocolFactory, boolean isAsync) {
+        CodeBlock.Builder builder = CodeBlock
+            .builder()
+            .add("$T.builder().delegateMarshaller(new $T($L))",
+                 isAsync ? AsyncStreamingRequestMarshaller.class : StreamingRequestMarshaller.class,
+                 marshaller,
+                 protocolFactory)
+            .add(".$L(requestBody)", isAsync ? "asyncRequestBody" : "requestBody");
+
+        if (opModel.hasRequiresLengthInInput()) {
+            builder.add(".requiresLength(true)");
+        }
+
+        if (AuthType.V4_UNSIGNED_BODY.equals(opModel.getAuthType())) {
+            builder.add(".transferEncoding(true)");
+        }
+
+        if (model.getMetadata().supportsH2()) {
+            builder.add(".useHttp2(true)");
+        }
+
+        builder.add(".build()");
+
+        return builder.build();
     }
 }
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/interceptor/ExecutionInterceptorChain.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/interceptor/ExecutionInterceptorChain.java
@@ -86,7 +86,7 @@ public class ExecutionInterceptorChain {
 
             SdkHttpFullRequest sdkHttpFullRequest = (SdkHttpFullRequest) context.httpRequest();
             if (!result.requestBody().isPresent() && sdkHttpFullRequest.contentStreamProvider().isPresent()) {
-                int contentLength = Integer.parseInt(sdkHttpFullRequest.firstMatchingHeader("Content-Length").orElse("0"));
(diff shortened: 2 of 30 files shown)


codegen/src/main/java/software/amazon/awssdk/codegen/poet/client/specs/ProtocolSpec.java
```diff
@@ -91,4 +95,67 @@ default String discoveredEndpoint(OperationModel opModel) {
...
+    default CodeBlock asyncStreamingMarshaller(IntermediateModel model, OperationModel opModel,
+                                                      ClassName marshaller, String protocolFactory) {
+        CodeBlock.Builder builder = CodeBlock
+            .builder()
+            .add("$T.builder().delegateMarshaller(new $T($L))", AsyncStreamingRequestMarshaller.class,
+                 marshaller, protocolFactory)
+            .add(".asyncRequestBody(requestBody)");
+
+        // Set requires length
+        if (opModel.hasRequiresLengthInInput()) {
```
zoewangg: There are some duplicates between this method and `syncStreamingMarshaller`, can we extract them out? Otherwise it's really easy to miss either one of them in the future if we add more header logic.

core/sdk-core/src/main/java/software/amazon/awssdk/core/runtime/transform/StreamingMarshallerUtil.java
```diff
@@ -0,0 +1,60 @@
...
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.core.runtime.transform;
```
zoewangg: I think all internal apis should be put in internal package

core/sdk-core/src/main/java/software/amazon/awssdk/core/runtime/transform/StreamingRequestMarshaller.java
```diff
@@ -32,29 +32,116 @@
...
     private final RequestBody requestBody;
+    private final boolean requiresLength;
+    private final boolean transferEncoding;
+    private final boolean useHttp2;
 
     /**
      * @param delegate    POJO marshaller (for path/query/header members).
      * @param requestBody {@link RequestBody} representing HTTP contents.
      */
+    @Deprecated
```
zoewangg: Can we remove this constructor?

http-client-spi/src/main/java/software/amazon/awssdk/http/HeaderValue.java
```diff
@@ -0,0 +1,30 @@
...
+
+import software.amazon.awssdk.annotations.SdkProtectedApi;
+
+/**
+ * Constants for HTTP header values
+ */
+@SdkProtectedApi
+public class HeaderValue {
+
+    public static final String CHUNKED = "chunked";
```
zoewangg: Can we put the header value in `Header` class above? It seems unnecessary to create another class for it to me.

core/sdk-core/src/main/java/software/amazon/awssdk/core/runtime/transform/AsyncStreamingRequestMarshaller.java
```diff
@@ -0,0 +1,120 @@
...
+import software.amazon.awssdk.core.async.AsyncRequestBody;
+import software.amazon.awssdk.http.SdkHttpFullRequest;
+
+/**
+ * Augments a {@link Marshaller} to add contents for an async streamed request.
+ *
+ * @param <T> Type of POJO being marshalled.
+ */
+@SdkProtectedApi
+public final class AsyncStreamingRequestMarshaller<T> implements Marshaller<T> {
```
zoewangg: Looks like this class is almost same as `StreamingRequestMarshaller` except the `requestBody`, can we create an abstract class to hold those common fields? It would be easier to maintain in the future if we want to add another field. We can also get rid of `StreamingMarshallerUtil` if we have the parent class

core/sdk-core/src/main/java/software/amazon/awssdk/core/interceptor/ExecutionInterceptorChain.java
```diff
@@ -86,7 +86,7 @@ public InterceptorContext modifyHttpRequestAndHttpContent(InterceptorContext con
 
             SdkHttpFullRequest sdkHttpFullRequest = (SdkHttpFullRequest) context.httpRequest();
             if (!result.requestBody().isPresent() && sdkHttpFullRequest.contentStreamProvider().isPresent()) {
-                int contentLength = Integer.parseInt(sdkHttpFullRequest.firstMatchingHeader("Content-Length").orElse("0"));
+                long contentLength = Long.parseLong(sdkHttpFullRequest.firstMatchingHeader("Content-Length").orElse("0"));
```
zoewangg: nice

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/transform/AbstractStreamingRequestMarshaller.java
```diff
@@ -0,0 +1,119 @@
...
+                              boolean useHttp2) {
+        if (contentLength.isPresent()) {
+            marshalled.putHeader(CONTENT_LENGTH, Long.toString(contentLength.get()));
+            return;
+        }
+
+        if (requiresLength) {
+            throw SdkClientException.create("This API requires Content-Length header to be set. "
+                                            + "Please set the content length on the RequestBody.");
+        } else if (transferEncoding && !useHttp2) {
```
zoewangg: nit, `else` is not needed.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/transform/AbstractStreamingRequestMarshaller.java
```diff
@@ -0,0 +1,119 @@
...
+     * If requiresLength and transferEncoding is not set to true and Content Length is missing,
+     * SDK is not required to calculate the Content-Length and delegate that behavior to the underlying http client.
+     *
+     * @param marshalled A mutable builder for {@link SdkHttpFullRequest} representing a HTTP request.
+     * @param contentLength Optional of content length
+     * @param requiresLength True if Content-Length header is required on the request
+     * @param transferEncoding True if "Transfer-Encoding: chunked" header should be set on request
+     * @param useHttp2 True if the operation uses http2
+     */
+    protected void addHeaders(SdkHttpFullRequest.Builder marshalled,
```
zoewangg: nit, `final`

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/transform/AbstractStreamingRequestMarshaller.java
varunnvs92: It is needed. There is another case where Content Length is missing and  both requiresLength and transferEncoding is false, where we don't set any header and let underlying http client decide the behavior.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/transform/AbstractStreamingRequestMarshaller.java
varunnvs92: I don't see the need for it. If the subclasses want to override the behavior, they should be able to do it. It doesn't matter in this case as these are all protected APIs and not used by customers.

zoewangg: It's generally good practice to either document it if it's expected to be extended by subclass or mark it `final` if it's not intended to be overridden. I'd lean in towards to prohibiting it if it's not needed for now and we can loose it up in the future if needed.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/transform/AbstractStreamingRequestMarshaller.java
zoewangg: gotcha

varunnvs92:
Added the integ test for mediastore

PR #1248, opened 2019-05-09
Sign events with current signing Clock time

dagnir:
## Description
Events should be signed with the current time (as provided by the Clock), not
the initial request time.

## Testing
Updated unit test, ran `TranscribeStreamingIntegrationTest` to verify against service.

## Types of changes

## Checklist

## License

### DIFF
--- a/core/auth/src/main/java/software/amazon/awssdk/auth/signer/internal/Aws4SignerRequestParams.java
+++ b/core/auth/src/main/java/software/amazon/awssdk/auth/signer/internal/Aws4SignerRequestParams.java
@@ -16,6 +16,7 @@
 package software.amazon.awssdk.auth.signer.internal;
 
 import java.time.Clock;
+import java.time.Duration;
 import software.amazon.awssdk.annotations.SdkInternalApi;
 import software.amazon.awssdk.auth.signer.params.Aws4SignerParams;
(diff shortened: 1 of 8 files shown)


core/auth/src/main/java/software/amazon/awssdk/auth/signer/internal/Aws4SignerRequestParams.java
```diff
@@ -141,4 +129,25 @@ public String getFormattedSigningDate() {
     public String getSigningAlgorithm() {
         return SignerConstant.AWS4_SIGNING_ALGORITHM;
     }
+
+    private Clock resolveSigningClock(Aws4SignerParams signerParams) {
+        if (signerParams.signingClockOverride().isPresent()) {
+            return signerParams.signingClockOverride().get();
+        }
+        Clock baseClock = Clock.systemUTC();
```
varunnvs92: For clarification, do we always use time in UTC?

dagnir: Yeah, the previous code used `System.currentTimeMillis()` to calculate the request date if no override was given, which is in UTC.

PR #1249, opened 2019-05-09
Updated dynamodb-enhanced documentation to specify which method calls…

millems:
… and conditions could cause exceptions.

### DIFF
--- a/services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/AsyncTable.java
+++ b/services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/AsyncTable.java
@@ -25,6 +25,7 @@ import software.amazon.awssdk.enhanced.dynamodb.model.ResponseItem;
 import software.amazon.awssdk.services.dynamodb.DynamoDbAsyncClient;
 import software.amazon.awssdk.services.dynamodb.model.GetItemRequest;
 import software.amazon.awssdk.services.dynamodb.model.PutItemRequest;
+import software.amazon.awssdk.utils.CompletableFutureUtils;
 
 /**
  * An asynchronous way of interacting with a DynamoDB table. This is usually created via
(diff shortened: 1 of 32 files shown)


services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/AsyncTable.java
```diff
@@ -52,6 +53,9 @@
     /**
      * Retrieve the name of the DynamoDB table. This does not make a service call, and does not guarantee that the DynamoDB table
      * actually exists.
+     *
+     * <p>
+     * This call should never fail with a {@link RuntimeException}.
```
zoewangg: This seems a bit ambiguous. Does it mean it might fail with a checked exception?

millems: I can fix this one pretty easily by specifying it will never throw an `Exception`, but what about elsewhere where it does sometimes throw `RuntimeException`s? Is that fine to leave as-is?

zoewangg: You mean there are some places where checked exception might be thrown?

millems: Checked exceptions will never be thrown.

zoewangg: Yeah, that's where it's a bit confusing, can we just say `This call should never throw a exception` so that people know right away without checking if it has `throws` to see if it throws a checked exception.

millems: Fixed!

PR #1191, opened 2019-04-04
Lazily initializing exceptions

zoewangg:
## Description
Delay the creation of `ApiCallTimeoutException` and `ApiCallAttemptTimeoutException` as they are not necessary and costly to create. This would improve performance in async path.

## Testing
Benchmark comparison (throughput ops/s)

Benchmark  |  before | after |     
------------ | -------------  | -----------                              
NettyClientH1NonTlsBenchmark.concurrentApiCall  |  7119 | 8453
NettyClientH1NonTlsBenchmark.sequentialApiCall |  2303 | 2393
NettyHttpClientH1Benchmark.concurrentApiCall (jdk) |  6390 | 7260
NettyHttpClientH1Benchmark.sequentialApiCall (jdk) |  1775 |  1827
NettyHttpClientH2Benchmark.concurrentApiCall (jdk) | 4513 | 4655
NettyHttpClientH2Benchmark.sequentialApiCall (jdk) | 1687 | 1723

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)

## Checklist

- [x] I have read the **CONTRIBUTING** document
- [x] Local run of `mvn install` succeeds
- [x] My code follows the code style of this project
- [ ] My change requires a change to the Javadoc documentation
- [ ] I have updated the Javadoc documentation accordingly
- [x] I have read the **README** document
- [ ] I have added tests to cover my changes
- [x] All new and existing tests passed
- [x] A short description of the change has been added to the **CHANGELOG**
- [ ] My change is to implement 1.11 parity feature and I have updated [LaunchChangelog](https://github.com/aws/aws-sdk-java-v2/blob/master/docs/LaunchChangelog.md)

## License

- [x] I confirm that this pull request can be released under the Apache 2 license

PR #1250, opened 2019-05-10
Update baseline data

zoewangg:
## Description
- Update baseline result to the latest data
- Update `BenchmarkResultProcessor` to check difference percentage if the current data and baseline data are statistically different. 

## License


dagnir:
Can we be a little more restrictive? Or is anything higher too strict as well?

zoewangg:
> Can we be a little more restrictive? Or is anything higher too strict as well?

Going to test it out

zoewangg:
The `compare` method turns out not very reliable. It's always showing statistically different even though the difference is only `21`.  Going to try to add a 10% margin

```
Baseline: apicall.httpclient.async.NettyHttpClientH1Benchmark.sequentialApiCall 1720.1013604261893

Result: apicall.httpclient.async.NettyHttpClientH1Benchmark.sequentialApiCall 1699.547
```

zoewangg:
Benchmark  |  w/  HealthCheckedChannelPool | w/o HealthCheckedChannelPool |     
------------ | -------------  | -----------                              
NettyClientH1NonTlsBenchmark.concurrentApiCall  |  7313 | 8238
NettyClientH1NonTlsBenchmark.sequentialApiCall |  2192 | 2380
NettyHttpClientH1Benchmark.concurrentApiCall (jdk) |  6279 | 7113
NettyHttpClientH1Benchmark.sequentialApiCall (jdk) |  1699 |  1813
NettyHttpClientH2Benchmark.concurrentApiCall (jdk) | 4272 | 4545
NettyHttpClientH2Benchmark.sequentialApiCall (jdk) | 1605 | 1712

dagnir:
Are these baseline numbers run with `HealthCheckedChannelPool` removed?

zoewangg:
No, it's with current code base.  The change in #1191 has improved the performance (we didn't update the baseline after the change), and then the latency introduced in `HealthCheckedChannelPool` "reverted" the performance back to before.

PR #1264, opened 2019-05-22
Allowing keys with leading slashes for S3

spfink:
S3 allows keys to begin with slashes for unknown reasons. This change allows those keys to work with the SDK and our HTTP clients.

## Testing
Added integration test

## Types of changes

## Checklist

## License


millems:
Changelog?

core/protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/internal/marshall/SimpleTypePathMarshaller.java
```diff
@@ -37,7 +37,7 @@
      * so that it preserves the path structure.
      */
     public static final XmlMarshaller<String> GREEDY_STRING =
-        new SimplePathMarshaller<>(ValueToStringConverter.FROM_STRING, PathMarshaller.GREEDY);
+        new SimplePathMarshaller<>(ValueToStringConverter.FROM_STRING, PathMarshaller.GREEDY_WITH_SLASHES);
```
millems: What services are affected by this change? How certain are we that they wouldn't be broken by this?

services/s3/src/it/java/software/amazon/awssdk/services/s3/KeysWithLeadingSlashIntegrationTest.java
```diff
@@ -0,0 +1,43 @@
...
+
+import java.nio.charset.StandardCharsets;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import software.amazon.awssdk.core.sync.RequestBody;
+
+public class KeysWithLeadingSlashIntegrationTest extends S3IntegrationTestBase {
+
+    private static final String BUCKET = temporaryBucketName(KeysWithLeadingSlashIntegrationTest.class);
+    private static final String KEY = "/stupidkeywithillegalleadingslashthatsucks";
```
millems: No passive aggression here. Nope.

core/protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/internal/marshall/SimpleTypePathMarshaller.java
spfink: This only effects XML services of which none of them currently use the greedy path marshaller.

services/s3/src/it/java/software/amazon/awssdk/services/s3/KeysWithLeadingSlashIntegrationTest.java
spfink: Just a random key name I swear.

core/protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/internal/marshall/SimpleTypePathMarshaller.java
spfink: Other than S3 that is.

spfink:
I don't like to change the protocol tests as I've done here but I think in this case it's fine. 

I've confirmed no other XML service uses the greedy path besides S3 and this is the correct behavior for S3. We shouldn't see any new services use XML. It could effect existing services if they add operations with a greedy path in the future. At that point we could switch this to be an S3 only customization but that would introduce a significant amount of additional handwritten S3 code and a bunch of new codegen customizations.

PR #1300, opened 2019-06-20
Converter refactor (part 1 of ?) - Updated interfaces

millems:
The existing interfaces required the same API for type converters that
worked on subtypes (e.g. converting collections) and the ones that
didn't work on subtypes (e.g. converting an Integer). The internals
differentiated the two based on the conversion condition.

This commit splits the subtype-based converters from the
non-subtype-based converters and removes the concept of an explicit
conversion condition. That's now based on the converter type.

This allows adding type safety to the converter interfaces, which will
pave the way to type-safe conversion of attributes (if your bean has a
String field, you can't try to use an Integer converter for it). It also
allowed for simplification of the non-subtype-based converter interface.


services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/converter/attribute/AttributeConverter.java
```diff
@@ -0,0 +1,107 @@
...
+ * @param <T> The Java type supported by this converter.
+ *
+ * @see AttributeConverterAware
+ * @see SubtypeAttributeConverter
+ */
+@SdkPublicApi
+@ThreadSafe
+public interface AttributeConverter<T> extends Converter<T> {
+    /**
+     * Convert the provided Java object into an {@link ItemAttributeValue}. This will raise a {@link RuntimeException} if the
```
dagnir: Why `RuntimeException` instead of `SdkClientException`?

millems: It's similar to our behavior elsewhere we use the `Validate` class to validate inputs - we raise a `RuntimeException`.

PR #1311, opened 2019-06-27
Fix equals and hashcode in service response and request classes

zoewangg:
## Description
- Fix `equals` and `hashcode` in service response and request classes to call `super`
- Fix `AwsSessionCredentials#equals` to not compare super because the super is an interface.
-  Introduce a new method `equalsBySdkFields` to compare only modeled fields in pojos.
## Testing
Added unit tests

## Types of changes

## Checklist

## License

### DIFF
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/SdkPojo.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/SdkPojo.java
@@ -30,4 +30,20 @@ public interface SdkPojo {
      * @return List of {@link SdkField} in this POJO. May be empty list but should never be null.
      */
     List<SdkField<?>> sdkFields();
+
+    /**
+     * Indicates whether some other object is "equal to" this one by SDK fields.
+     * An SDK field is a modeled, non-inherited field in an {@link SdkPojo} class,
+     * and is generated based on a service model.
+     *
+     * <p>
+     * If an {@link SdkPojo} class does not have any inherited fields, {@code equalsBySdkFields}
+     * and {@code equals} are essentially the same.
+     *
+     * @param other the object to be compared with
+     * @return true if the other object equals to this object by sdk fields, false otherwise.
+     */
+    default boolean equalsBySdkFields(Object other) {
+        throw new UnsupportedOperationException();
+    }
 }
--- a/.changes/next-release/bugfix-AWSSDKforJavav2-0fd78bc.json
+++ b/.changes/next-release/bugfix-AWSSDKforJavav2-0fd78bc.json
@@ -1,0 +1,5 @@
+{
+    "category": "AWS SDK for Java v2", 
+    "type": "bugfix", 
+    "description": "Fix `AwsSessionCredentials#equals` to not compare super because the super is an interface."
+}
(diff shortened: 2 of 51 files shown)


zoewangg:
I still need to fix the codegen tests, but wanted to get feedback before I fix it because there are a  lot of test files to fix.

`equalsBySdkFields` is being added to **ALL** model classes: request, response and non-request, non-response shapes because I think we should keep the pojo classes consistent, even though for `equals` and `equalsBySdkFileds` are essentially same for non-request and non-response shapes. Let me know if you think we should only add it to request and response classes.

Note that I didn't make `equals` in the `AwsRequest` to just delegate to `equalsBySdkFields` because it would be backwards incompatible with internal models.

Generated request class.
```java
   @Override
    public boolean equals(Object obj) {
        return super.equals(obj) && equalsBySdkFields(obj);
    }

    @Override
    public boolean equalsBySdkFields(Object obj) {
        if (this == obj) {
            return true;
        }
        if (obj == null) {
            return false;
        }
        if (!(obj instanceof SubscribeToShardRequest)) {
            return false;
        }
        SubscribeToShardRequest other = (SubscribeToShardRequest) obj;
        return Objects.equals(consumerARN(), other.consumerARN()) && Objects.equals(shardId(), other.shardId())
                && Objects.equals(startingPosition(), other.startingPosition());
    }
```

Generated Response class:
```java
    @Override
    public boolean equals(Object obj) {
        return super.equals(obj) && equalsBySdkFields(obj);
    }

    @Override
    public boolean equalsBySdkFields(Object obj) {
        if (this == obj) {
            return true;
        }
        if (obj == null) {
            return false;
        }
        if (!(obj instanceof SubscribeToShardResponse)) {
            return false;
        }
        return true;
    }
```
Generated shape classes:

```java
    @Override
    public boolean equals(Object obj) {
        return equalsBySdkFields(obj);
    }

    @Override
    public boolean equalsBySdkFields(Object obj) {
        if (this == obj) {
            return true;
        }
        if (obj == null) {
            return false;
        }
        if (!(obj instanceof StartingPosition)) {
            return false;
        }
        StartingPosition other = (StartingPosition) obj;
        return Objects.equals(typeAsString(), other.typeAsString()) && Objects.equals(sequenceNumber(), other.sequenceNumber())
                && Objects.equals(timestamp(), other.timestamp());
    }

```

core/sdk-core/src/main/java/software/amazon/awssdk/core/SdkPojo.java
```diff
@@ -30,4 +30,18 @@
...
+     * Indicates whether some other object is "equal to" this one by sdk fields.
+     * This method only compares the non-inherited fields within this class.
+     *
+     * <p>
+     * If this class does not have a super class, this is essential same as {@link #equals(Object)}.
+     *
+     * @param other the object to be compared with
+     * @return true if the other object equals to this object by sdk fields, false otherwise.
+     */
+    default boolean equalsBySdkFields(Object other) {
```
millems: Should we clarify what is meant by "sdk fields". I know we use that phrasing elsewhere here, but it's not really made clear.

zoewangg: Updated to 
```
Indicates whether some other object is "equal to" this one by SDK fields. 

An SDK field is a modeled, non-inherited field in an SdkPojo class, and is generated based on a service model.

If an SdkPojo class does not have any inherited fields, equalsBySdkFields and equals are essentially the same
```

PR #1312, opened 2019-06-27
Converter refactor (part 3 of ?) - Add remaining attribute converters

millems:


services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/converter/attribute/bundled/InstantAsIntegerAttributeConverter.java
```diff
@@ -0,0 +1,84 @@
...
+ * {@code ItemAttributeValue.fromString("1970-01-01T00:00:00+01:00")} will be converted to
+ * {@code Instant.EPOCH.minus(1, ChronoUnit.HOURS)}).
+ *
+ * <p>
+ * This can be created via {@link #create()}.
+ */
+@SdkPublicApi
+@ThreadSafe
+@Immutable
+public final class InstantAsIntegerAttributeConverter implements AttributeConverter<Instant> {
```
dagnir: This class says "integer" but it describes conversion to/from a fractional value

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/converter/attribute/bundled/InstantAsIntegerAttributeConverter.java
```diff
@@ -0,0 +1,84 @@
...
+ *     <li>{@code Instant.EPOCH.minusMillis(1)} is stored as {@code ItemAttributeValue.fromNumber("-0.001")}</li>
+ *     <li>{@code Instant.EPOCH.plusNanos(1)} is stored as {@code ItemAttributeValue.fromNumber("0.000000001")}</li>
+ *     <li>{@code Instant.EPOCH.minusNanos(1)} is stored as {@code ItemAttributeValue.fromNumber("-0.000000001")}</li>
+ * </ul>
+ *
+ * <p>
+ * This converter can read any values written by itself, {@link InstantAsStringAttributeConverter},
+ * {@link OffsetDateTimeAsStringAttributeConverter} or {@link ZonedDateTimeAsStringAttributeConverter}. Offset and zoned times
+ * will be automatically converted to the equivalent {@code Instant} based on the time zone information in the record (e.g.
+ * {@code ItemAttributeValue.fromString("1970-01-01T00:00:00+01:00")} will be converted to
```
dagnir: This example doesn't seem to match the others; seems like it should show `ItemAttributeValue.fromNumber("...")`

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/converter/ConverterUtils.java
```diff
@@ -0,0 +1,165 @@
...
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.enhanced.dynamodb.internal.converter;
+
+import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.utils.Validate;
+
+@SdkInternalApi
+public class ConverterUtils {
```
dagnir: It's a little hard to figure what these methods do. Can we add some docs for them?

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/converter/attribute/ChainAttributeConverter.java
```diff
@@ -0,0 +1,265 @@
...
+    private ChainAttributeConverter(Builder<T> builder) {
+        this.type = builder.type;
+        this.parent = builder.parent;
+
+        // Subtype converters are used in the REVERSE order of how they were added to the builder.
+        for (int i = builder.subtypeConverters.size() - 1; i >= 0; i--) {
+            this.subtypeConverters.add(builder.subtypeConverters.get(i));
+        }
+
+        // Converters are used in the REVERSE order of how they were added to the builder.
```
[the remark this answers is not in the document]
millems: I can clarify in the comments. The implementation iterates the list in order, so we need to reverse it to match the contract defined in `AttributeConverterAware`:

> Converters added last always have a higher precedence than those added first. For example, the last {@code AttributeConverter<String>)} converter added to a {@link RequestItem.Builder} will have higher precedence than any other {@code AttributeConverter<String>} converters that were added before it.

It doesn't affect the VAST majority of customers, but the reasoning for that is because the code closest to the call site should have the ability to override the behavior of earlier code, because it's more aware of the context around the request. E.g. If they have global logic that pre-configures the client, they could want to override that behavior later.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/converter/attribute/bundled/InstantAsIntegerAttributeConverter.java
millems: Good call. I will change these all to `AsNumber`

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/converter/attribute/bundled/InstantAsIntegerAttributeConverter.java
millems: It actually supports *reading* string-formatted values, and this was meant to illustrate that it is compatible reading values written by `OffsetDateTimeAsStringAttributeConverter`. I'll clarify.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/converter/ConverterUtils.java
millems: +1

PR #1320, opened 2019-07-03
Adding support for resolving web identity tokens from disk

spfink:
Adds support for resolving JWT tokens from disk for identifying with STS

## Testing
Existing tests pass. Added basic tests for JWT functionality.

## Types of changes

## Checklist

## License


core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/ProfileCredentialsUtils.java
```diff
@@ -95,10 +96,15 @@ public ProfileCredentialsUtils(Profile profile, Function<String, Optional<Profil
         if (properties.containsKey(ProfileProperty.ROLE_ARN)) {
             boolean hasSourceProfile = properties.containsKey(ProfileProperty.SOURCE_PROFILE);
             boolean hasCredentialSource = properties.containsKey(ProfileProperty.CREDENTIAL_SOURCE);
+            boolean hasWebIdentityTokenFile = properties.containsKey(ProfileProperty.WEB_IDENTITY_TOKEN_FILE);
             Validate.validState(!(hasSourceProfile && hasCredentialSource),
                                 "Invalid profile file: profile has both %s and %s.",
                                 ProfileProperty.SOURCE_PROFILE, ProfileProperty.CREDENTIAL_SOURCE);
 
+            if (hasWebIdentityTokenFile) {
```
dagnir: further down we have `requireProperties(ProfileProperty.ROLE_ARN, ProfileProperty.WEB_IDENTITY_TOKEN_FILE)` so we should be checking of the role ARN is defined as well.

dagnir:
Seems this change has the effect putting the web identity credentials at the top of the chain before static creds in environment variables and in the shared config file. Is that the behavior we want?

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/ProfileCredentialsUtils.java
spfink: Updated.

spfink:
This does have the effect of making web identity tokens always take precedence. I'm assuming that this is acceptable because they have to explicitly set this environment.

I imagine there will be scenarios where customers want to switch to web identity tokens so they add this env variable not realizing that they are continuing to set env creds. This would not be an issue in this case as the SDK would default to the web tokens.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/WebIdentityCredentialsUtil.java
```diff
@@ -0,0 +1,57 @@
...
+@SdkInternalApi
+public final class WebIdentityCredentialsUtil {
+
+    private static final String STS_WEB_IDENTITY_CREDENTIALS_PROVIDER_FACTORY =
+        "software.amazon.awssdk.services.sts.internal.StsWebIdentityCredentialsProviderFactory";
+
+    private WebIdentityCredentialsUtil() {}
+
+    public static String resolveWebIdentityToken(Path filePath) {
+        InputStream webIdentityTokenStream = FunctionalUtils.invokeSafely(() -> Files.newInputStream(filePath));
```
dagnir: Forgot to check this earlier, but we should be checking that the path absolute. We aren't supporting relative paths

spfink: Added check and tests on this class.

services/sts/src/main/java/software/amazon/awssdk/services/sts/internal/StsWebIdentityCredentialsProviderFactory.java
```diff
@@ -0,0 +1,132 @@
...
+            }
+        }
+    }
+
+
+    private static final class StsRetryCondition implements RetryCondition {
+
+        @Override
+        public boolean shouldRetry(RetryPolicyContext context) {
+            if (context.exception() instanceof IdpCommunicationErrorException) {
```
dagnir: nit: can be shortened to `return context.exception() instanceof IdpCommunicationErrorException`

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/ProfileCredentialsUtils.java
```diff
@@ -83,7 +83,8 @@ public ProfileCredentialsUtils(Profile profile, Function<String, Optional<Profil
      * providers will be created.
      */
     public Optional<AwsCredentialsProvider> credentialsProvider() {
-        return credentialsProvider(new HashSet<>());
+        Optional<AwsCredentialsProvider> credsProvider = credentialsProvider(new HashSet<>());
+        return credsProvider;
```
millems: Why the change?

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/SystemSettingsCredentialsProvider.java
```diff
@@ -67,8 +86,16 @@ public AwsCredentials resolveCredentials() {
                                     .build();
         }
 
-        return sessionToken == null ? AwsBasicCredentials.create(accessKey, secretKey)
-                                    : AwsSessionCredentials.create(accessKey, secretKey, sessionToken);
+        if (webIdentityTokenFile != null && roleArn != null) {
```
millems: Should we warn if one is specified without the other?

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/SystemSettingsCredentialsProvider.java
```diff
@@ -43,12 +46,28 @@
...
+
+        return credentialsProvider.resolveCredentials();
+    }
+
+    private void resolveProvider() {
         String accessKey = trim(loadSetting(SdkSystemSetting.AWS_ACCESS_KEY_ID).orElse(null));
         String secretKey = trim(loadSetting(SdkSystemSetting.AWS_SECRET_ACCESS_KEY).orElse(null));
         String sessionToken = trim(loadSetting(SdkSystemSetting.AWS_SESSION_TOKEN).orElse(null));
 
+        String roleArn = trim(loadSetting(SdkSystemSetting.AWS_ROLE_ARN).orElse(null));
```
millems: Should this also be used to assume a role using the assume-role credentials provider?

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/WebIdentityCredentialsUtil.java
```diff
@@ -0,0 +1,62 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.auth.credentials.WebIdentityTokenCredentialsProviderFactory;
+import software.amazon.awssdk.core.exception.SdkClientException;
+import software.amazon.awssdk.utils.FunctionalUtils;
+import software.amazon.awssdk.utils.IoUtils;
+
+@SdkInternalApi
+public final class WebIdentityCredentialsUtil {
```
millems: Utils*

millems: Javadoc?

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/SystemSettingsCredentialsProvider.java
```diff
@@ -67,8 +86,16 @@ public AwsCredentials resolveCredentials() {
                                     .build();
         }
 
-        return sessionToken == null ? AwsBasicCredentials.create(accessKey, secretKey)
-                                    : AwsSessionCredentials.create(accessKey, secretKey, sessionToken);
+        if (webIdentityTokenFile != null && roleArn != null) {
+            String webIdentityToken = WebIdentityCredentialsUtil.resolveWebIdentityToken(Paths.get(webIdentityTokenFile));
+            WebIdentityTokenCredentialsProviderFactory factory = WebIdentityCredentialsUtil.factory();
+            credentialsProvider = factory.create(roleArn, roleSessionName, webIdentityToken);
```
millems: Is this a closeable credentials provider? Do we need to make this provider closeable as well?

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/SystemSettingsCredentialsProvider.java
```diff
@@ -43,12 +46,28 @@
...
 public abstract class SystemSettingsCredentialsProvider implements AwsCredentialsProvider {
+
+    private AwsCredentialsProvider credentialsProvider;
+
     @Override
     public AwsCredentials resolveCredentials() {
+
+        if (credentialsProvider == null) {
+            resolveProvider();
+        }
```
millems: Any reason not to just do this on creation, instead of first use?

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/WebIdentityCredentialsUtil.java
```diff
@@ -0,0 +1,62 @@
...
+
+    private static final String STS_WEB_IDENTITY_CREDENTIALS_PROVIDER_FACTORY =
+        "software.amazon.awssdk.services.sts.internal.StsWebIdentityCredentialsProviderFactory";
+
+    private WebIdentityCredentialsUtil() {}
+
+    public static String resolveWebIdentityToken(Path filePath) {
+        if (!filePath.isAbsolute()) {
+            throw SdkClientException.create("Web identity token file path must be an absolute file path");
+        }
```
millems: How do absolute paths work with home-based paths? Is ~/file an absolute path?

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/WebIdentityCredentialsUtil.java
```diff
@@ -0,0 +1,62 @@
...
+
+        InputStream webIdentityTokenStream = FunctionalUtils.invokeSafely(() -> Files.newInputStream(filePath));
```
millems: Try with resources?

core/sdk-core/src/main/java/software/amazon/awssdk/core/SdkSystemSetting.java
```diff
@@ -42,6 +42,15 @@
...
     AWS_SESSION_TOKEN("aws.sessionToken", null),
 
+    /**
+     * Configure the AWS web identity token file path.
+     */
+    AWS_WEB_IDENTITY_TOKEN_FILE("AWS_WEB_IDENTITY_TOKEN_FILE", null),
+
+    AWS_ROLE_ARN("AWS_ROLE_ARN", null),
+
+    AWS_ROLE_SESSION_NAME("AWS_ROLE_SESSION_NAME", null),
```
millems: Javadoc

core/sdk-core/src/main/java/software/amazon/awssdk/core/SdkSystemSetting.java
```diff
@@ -42,6 +42,15 @@
      */
     AWS_SESSION_TOKEN("aws.sessionToken", null),
 
+    /**
+     * Configure the AWS web identity token file path.
+     */
+    AWS_WEB_IDENTITY_TOKEN_FILE("AWS_WEB_IDENTITY_TOKEN_FILE", null),
```
millems: The first parameter should be a system property.

services/sts/src/main/java/software/amazon/awssdk/services/sts/auth/StsAssumeRoleWithWebIdentityCredentialsProvider.java
```diff
@@ -92,6 +95,11 @@ private Builder() {
          * @return This object for chained calls.
          */
         public Builder refreshRequest(AssumeRoleWithWebIdentityRequest assumeRoleWithWebIdentityRequest) {
+            this.assumeRoleWithWebIdentityRequest = () -> assumeRoleWithWebIdentityRequest;
+            return this;
+        }
+
+        public Builder refreshRequest(Supplier<AssumeRoleWithWebIdentityRequest> assumeRoleWithWebIdentityRequest) {
             this.assumeRoleWithWebIdentityRequest = assumeRoleWithWebIdentityRequest;
```
millems: Is this the same pattern as the other providers?
+ Javadoc.

services/sts/src/main/java/software/amazon/awssdk/services/sts/internal/StsWebIdentityCredentialsProviderFactory.java
```diff
@@ -0,0 +1,128 @@
...
+import software.amazon.awssdk.services.sts.StsClient;
+import software.amazon.awssdk.services.sts.auth.StsAssumeRoleWithWebIdentityCredentialsProvider;
+import software.amazon.awssdk.services.sts.model.AssumeRoleWithWebIdentityRequest;
+import software.amazon.awssdk.services.sts.model.IdpCommunicationErrorException;
+import software.amazon.awssdk.utils.FunctionalUtils;
+import software.amazon.awssdk.utils.IoUtils;
+import software.amazon.awssdk.utils.SdkAutoCloseable;
+
+@SdkInternalApi
+public final class StsWebIdentityCredentialsProviderFactory implements WebIdentityTokenCredentialsProviderFactory {
```
millems: Javadoc

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/WebIdentityTokenCredentialsProviderFactory.java
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * Currently this is used to allow a {@link Profile} or environment variable configured with a role that should be assumed with
+ * a web identity token to create a credentials provider via the
+ * 'software.amazon.awssdk.services.sts.internal.StsWebIdentityCredentialsProviderFactory', assuming STS is on the classpath.
+ */
+@FunctionalInterface
+@SdkProtectedApi
+public interface WebIdentityTokenCredentialsProviderFactory {
+    AwsCredentialsProvider create(String roleArn, String roleSessionName, String webIdentityToken);
+}
```
millems: Can we reuse the existing profile-based provider for this, and create one for environment variables? I don't like the idea of adding a new one of these for every credential source we support, which is why the other one is generic to profiles.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/ProfileCredentialsUtils.java
spfink: Not sure, might have been me testing and missed reverting.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/SystemSettingsCredentialsProvider.java
spfink: I'd be fine warning if web identity was specified and role arn wasn't but not the other way around.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/SystemSettingsCredentialsProvider.java
spfink: This was brought up on the Go PR as well. I think it is definitely something that we should support but might do it as a separate PR.

services/sts/src/main/java/software/amazon/awssdk/services/sts/auth/StsAssumeRoleWithWebIdentityCredentialsProvider.java
spfink: Mostly. Other provider passes the non-supplier to the supplier method instead of directly setting. Will update.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/WebIdentityTokenCredentialsProviderFactory.java
spfink: We certainly could re-use I guess but the environment/system properties don't really have a container which would mean you are just passing strings which is then just generic and why not re-use for the profile file?

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/WebIdentityCredentialsUtil.java
dagnir: It shouldn't be since the resolved path depends on the user that owns the process.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/WebIdentityTokenCredentialsProviderFactory.java
dagnir: Don't have any strong feelings either way, but slightly leaning towards reuse as well.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/SystemSettingsCredentialsProvider.java
spfink: We could do this but it would force us to remove some tests we have that instantiate an environment variable credential provider.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/WebIdentityCredentialsUtil.java
spfink: Does that really provide much benefit here?

millems: Yes, if IoUtils.toUtf8String throws, then we leak file descriptors.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/SystemSettingsCredentialsProvider.java
millems: Why would we have to remove the tests? This is a lot of complexity, and I don't see the benefit.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/SystemSettingsCredentialsProvider.java
millems: It may not be backwards-compatible to add it later, so should we figure it out now?

spfink: It should be fine to add it backwards compatibly as it's presence without a web identity token would mean use regular assume role. I wouldn't want to add it before working through the process to add to all SDKs since none support it today AFAIK.

millems: If they specify just role_arn with this change, it'll use regular credentials. Later, it would start using assume role credentials. That seems backwards-incompatible to me.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/EnvironmentVariableCredentialsProvider.java
```diff
@@ -29,6 +29,7 @@
 public final class EnvironmentVariableCredentialsProvider extends SystemSettingsCredentialsProvider {
 
     private EnvironmentVariableCredentialsProvider() {
+        super();
```
millems: This is redundant.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/SystemPropertyCredentialsProvider.java
```diff
@@ -29,6 +29,7 @@
 public final class SystemPropertyCredentialsProvider extends SystemSettingsCredentialsProvider {
 
     private SystemPropertyCredentialsProvider() {
+        super();
```
millems: This is redundant.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/ProfileCredentialsUtils.java
```diff
@@ -155,6 +161,17 @@ private AwsCredentialsProvider credentialProcessCredentialsProvider() {
...
 
+    private AwsCredentialsProvider roleAndWebIdentityTokenProfileCredentialsProvider() {
+        requireProperties(ProfileProperty.ROLE_ARN, ProfileProperty.WEB_IDENTITY_TOKEN_FILE);
+
+        String roleArn = properties.get(ProfileProperty.ROLE_ARN);
+        String roleSessionName = properties.get(ProfileProperty.ROLE_SESSION_NAME);
+        String webIdentityTokenPath = properties.get(ProfileProperty.WEB_IDENTITY_TOKEN_FILE);
+
+        return WebIdentityCredentialsUtils.factory()
+                                          .create(roleArn, roleSessionName, webIdentityTokenPath);
```
millems: Can we use a parameter object, since this is cross-module?

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/SystemSettingsCredentialsProvider.java
```diff
@@ -42,33 +47,84 @@
...
+                                              .build();
+        } else if (StringUtils.isEmpty(accessKey) || StringUtils.isEmpty(secretKey)) {
+            loadException = SdkClientException.builder()
+                                              .message(String.format("An access key and secret key must both be specified via " +
+                                                                     "their environment variables (%s) (%s) or system " +
+                                                                     "properties (%s) (%s)",
+                                                                     SdkSystemSetting.AWS_ACCESS_KEY_ID.environmentVariable(),
+                                                                     SdkSystemSetting.AWS_ACCESS_KEY_ID.property(),
+                                                                     SdkSystemSetting.AWS_SECRET_ACCESS_KEY.environmentVariable(),
+                                                                     SdkSystemSetting.AWS_SECRET_ACCESS_KEY.property()))
```
millems: Incorrect parameter order.

services/sts/src/main/java/software/amazon/awssdk/services/sts/internal/StsWebIdentityCredentialsProviderFactory.java
```diff
@@ -0,0 +1,132 @@
...
+import software.amazon.awssdk.utils.FunctionalUtils;
+import software.amazon.awssdk.utils.IoUtils;
+import software.amazon.awssdk.utils.SdkAutoCloseable;
+
+/**
+ * An implementation of {@link WebIdentityTokenCredentialsProviderFactory} that allows users to assume a role
+ * using a web identity token file specified in either a {@link Profile} or environment variables.
+ */
+@SdkInternalApi
+public final class StsWebIdentityCredentialsProviderFactory implements WebIdentityTokenCredentialsProviderFactory {
```
millems: Is this used cross-module?

spfink: Yes

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/ProfileCredentialsUtils.java
spfink: What's the benefit here? Why wrap all of these if they are already unwrapped?

millems: It makes forward-compatibility a lot easier. It's a cheap price to pay for an annoying experience later if we ever need to add something else.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/WebIdentityCredentialsUtils.java
```diff
@@ -0,0 +1,77 @@
...
+
+    private WebIdentityCredentialsUtils() {}
+
+    /**
+     * Resolves a String token from a JWT web identity token file.
+     *
+     * @param filePath The absolute path to the JWT token file.
+     * @return The JWT token as a string.
+     */
+    public static String resolveWebIdentityToken(Path filePath) {
```
millems: I don't see this being used?

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/WebIdentityCredentialsUtils.java
```diff
@@ -0,0 +1,77 @@
...
+        if (!filePath.isAbsolute()) {
+            throw SdkClientException.create("Web identity token file path must be an absolute file path");
+        }
+
+        InputStream webIdentityTokenStream = FunctionalUtils.invokeSafely(() -> Files.newInputStream(filePath));
+        String webIdentityToken = FunctionalUtils.invokeSafely(() -> IoUtils.toUtf8String(webIdentityTokenStream));
```
millems: If `IoUtils.toUtf8String` throws an exception, the `webIdentityTokenStream` won't be closed.

services/sts/src/main/java/software/amazon/awssdk/services/sts/internal/StsWebIdentityCredentialsProviderFactory.java
millems: Should it be an @SdkProtectedApi?

services/sts/src/main/java/software/amazon/awssdk/services/sts/internal/StsWebIdentityCredentialsProviderFactory.java
```diff
@@ -0,0 +1,135 @@
...
+                     FunctionalUtils.invokeSafely(() -> Files.newInputStream(Paths.get(filePath)))) {
+                return FunctionalUtils.invokeSafely(() -> IoUtils.toUtf8String(webIdentityTokenStream));
+            } catch (IOException e) {
+                throw new RuntimeException(e);
+            }
+        }
+    }
+
+
+    private static final class StsRetryCondition implements RetryCondition {
```
millems: Should this be included in the other STS credential providers? Maybe the entire STS retry condition?

services/sts/src/main/java/software/amazon/awssdk/services/sts/internal/StsWebIdentityCredentialsProviderFactory.java
```diff
@@ -0,0 +1,135 @@
...
+        @Override
+        public Object get() {
+            return request.toBuilder().webIdentityToken(getToken(webIdentityTokenFilePath)).build();
+        }
+
+        private String getToken(String filePath) {
+            try (InputStream webIdentityTokenStream =
+                     FunctionalUtils.invokeSafely(() -> Files.newInputStream(Paths.get(filePath)))) {
+                return FunctionalUtils.invokeSafely(() -> IoUtils.toUtf8String(webIdentityTokenStream));
+            } catch (IOException e) {
```
millems: You're already catching `IOException`, do you need invokeSafely?

services/sts/src/main/java/software/amazon/awssdk/services/sts/internal/StsWebIdentityCredentialsProviderFactory.java
```diff
@@ -0,0 +1,135 @@
...
+                throw new RuntimeException(e);
```
millems: Maybe `UncheckedIOException`?

services/sts/src/main/java/software/amazon/awssdk/services/sts/internal/StsWebIdentityCredentialsProviderFactory.java
spfink: I can rename but this is specific to this provider.

services/sts/src/main/java/software/amazon/awssdk/services/sts/internal/StsWebIdentityCredentialsProviderFactory.java
spfink: Sure

services/sts/src/main/java/software/amazon/awssdk/services/sts/internal/StsWebIdentityCredentialsProviderFactory.java
spfink: Removed

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/WebIdentityCredentialsUtils.java
spfink: Removed this

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/WebIdentityCredentialsUtils.java
spfink: Removed

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/WebIdentityTokenCredentialsProvider.java
```diff
@@ -0,0 +1,64 @@
...
+import software.amazon.awssdk.auth.credentials.internal.CredentialProperties;
+import software.amazon.awssdk.auth.credentials.internal.WebIdentityCredentialsUtils;
+import software.amazon.awssdk.core.SdkSystemSetting;
+import software.amazon.awssdk.utils.ToString;
+
+/**
+ * A credential provider that will read web identity token file path, aws role arn
+ * and aws session name from system properties or environment variables for using
+ * web identity token credentials with STS.
+ */
```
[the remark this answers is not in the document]
spfink: Added note.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/CredentialProperties.java
```diff
@@ -0,0 +1,101 @@
...
+
+package software.amazon.awssdk.auth.credentials.internal;
+
+import software.amazon.awssdk.annotations.SdkInternalApi;
+
+/**
+ * A container for credential properties.
+ */
+@SdkInternalApi
+public class CredentialProperties {
```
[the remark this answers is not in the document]
spfink: Switched to WebIdentityTokenCredentialProperties

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/WebIdentityTokenCredentialsProvider.java
```diff
@@ -0,0 +1,64 @@
...
+        String roleArn = trim(SdkSystemSetting.AWS_ROLE_ARN.getStringValueOrThrow());
+        Optional<String> roleSessionName = SdkSystemSetting.AWS_ROLE_SESSION_NAME.getStringValue();
+
+        CredentialProperties credentialProperties = CredentialProperties.builder()
+                                                                        .roleArn(roleArn)
+                                                                        .roleSessionName(roleSessionName.orElse(null))
+                                                                        .webIdentityTokenFilePath(webIdentityTokenFlePath)
+                                                                        .build();
+
+        credentialsProvider = WebIdentityCredentialsUtils.factory().create(credentialProperties);
```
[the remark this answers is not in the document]
spfink: Added

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/WebIdentityTokenCredentialsProvider.java
```diff
@@ -0,0 +1,173 @@
...
+            loadException = e;
+        }
+
+        if (loadException != null) {
+            this.loadException = loadException;
+            this.credentialsProvider = null;
+        } else {
+            this.credentialsProvider = credentialsProvider;
+            this.loadException = null;
+        }
```
millems: Both branches of this if statement will end up doing the same thing, so can we just make it...
```java
this.loadException = loadException;
this.credentialsProvider = credentialsProvider;
```
...?

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/WebIdentityTokenCredentialsProvider.java
```diff
@@ -0,0 +1,173 @@
...
+
+        /**
+         * Define the role session name that should be used by this credentials provider.
+         */
+        Builder roleSessionName(String roleSessionName);
+
+        /**
+         * Define the absolute path to the web identity token file that should be used by this credentials provider.
+         */
+        Builder webIdentityTokenFilePath(String webIdentityTokenFilePath);
```
millems: Should this be a `Path`? If so, can we name it `webIdentityTokenFile`?

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/WebIdentityTokenCredentialsProvider.java
```diff
@@ -0,0 +1,173 @@
...
+
+        /**
+         * Create a {@link WebIdentityTokenCredentialsProvider} using the configuration applied to this builder.
+         */
+        WebIdentityTokenCredentialsProvider build();
+    }
+
+    static final class BuilderImpl implements Builder {
```
millems: private?

services/sts/src/main/java/software/amazon/awssdk/services/sts/internal/StsWebIdentityCredentialsProviderFactory.java
millems: Ping

services/sts/src/main/java/software/amazon/awssdk/services/sts/internal/StsWebIdentityCredentialsProviderFactory.java
millems: I more meant, should we do this for all STS usages - retry on this exception.

spfink: AFAIK, this is specific to this use case (was called out in SEP). Can see what other SDKs are doing here though.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/WebIdentityTokenCredentialsProvider.java
spfink: Yep yep

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/WebIdentityTokenCredentialsProvider.java
spfink: Updated

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/WebIdentityTokenCredentialsProvider.java
```diff
@@ -0,0 +1,170 @@
...
+        AwsCredentialsProvider credentialsProvider = null;
+        RuntimeException loadException = null;
+
+        try {
+            Path webIdentityTokenFile;
+            String roleArn;
+            Optional<String> roleSessionName;
+
+            // Don't allow a mix of environment variables and builder configured values
+            if (builder.webIdentityTokenFile != null) {
```
millems: I'm a little worried that builder.webIdentityTokenFile = null means "use environment variables" even if they specified the roleArn and roleSessionName.

Do we have a precedent for this?

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/WebIdentityTokenFileCredentialsProvider.java
```diff
@@ -0,0 +1,168 @@
...
+    public AwsCredentials resolveCredentials() {
+        if (loadException != null) {
+            throw loadException;
+        }
+        return credentialsProvider.resolveCredentials();
+    }
+
+    public static Builder builder() {
+        return new BuilderImpl();
+    }
```
millems: Wrong method order

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/WebIdentityTokenFileCredentialsProvider.java
```diff
@@ -0,0 +1,168 @@
...
+
+    @Override
+    public String toString() {
+        return ToString.create("WebIdentityTokenCredentialsProvider");
+    }
```
millems: Should we include whichever fields they've configured?

PR #1336, opened 2019-07-10
Generate code to notify error for streaming operations

zoewangg:
## Description
Generating code to notify error for streaming operations.

```
            executeFuture.whenComplete((r, e) -> {
                if (e != null) {
                    runAndLogError(log, "Exception thrown in exceptionOccurred callback, ignoring",
                            () -> asyncResponseTransformer.exceptionOccurred(e));
                }
            });
```
Per docs,
https://github.com/aws/aws-sdk-java-v2/blob/5a799393ce9f34b07625ca5d614e5f6367ec7780/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/AsyncRetryableStage.java#L167-L170

we should invoke `AsyncResponseTransformer#exceptionOccurred` in the generated client code; we were only generating the code in json services previously, so the fix is to add the generated code for xml services.
 

## Testing
added codegen tests, unit tests, integ tests.

## Types of changes

## Checklist

## License

### DIFF
--- a/codegen/src/test/resources/software/amazon/awssdk/codegen/poet/client/test-query-async-client-class.java
+++ b/codegen/src/test/resources/software/amazon/awssdk/codegen/poet/client/test-query-async-client-class.java
@@ -1,0 +1,272 @@
+package software.amazon.awssdk.services.query;
+
+import static software.amazon.awssdk.utils.FunctionalUtils.runAndLogError;
+
+import java.util.concurrent.CompletableFuture;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import software.amazon.awssdk.annotations.Generated;
+import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.awscore.client.handler.AwsAsyncClientHandler;
+import software.amazon.awssdk.awscore.exception.AwsServiceException;
+import software.amazon.awssdk.core.async.AsyncRequestBody;
+import software.amazon.awssdk.core.async.AsyncResponseTransformer;
+import software.amazon.awssdk.core.client.config.SdkClientConfiguration;
+import software.amazon.awssdk.core.client.handler.AsyncClientHandler;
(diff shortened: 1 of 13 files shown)


codegen/src/test/resources/software/amazon/awssdk/codegen/poet/client/test-query-async-client-class.java
```diff
@@ -0,0 +1,271 @@
...
+                new ClientExecutionParams<StreamingOutputOperationRequest, StreamingOutputOperationResponse>()
+                    .withOperationName("StreamingOutputOperation")
+                    .withMarshaller(new StreamingOutputOperationRequestMarshaller(protocolFactory))
+                    .withResponseHandler(responseHandler).withErrorResponseHandler(errorResponseHandler)
+                    .withInput(streamingOutputOperationRequest), asyncResponseTransformer);
+            executeFuture.whenComplete((r, e) -> {
+                if (e != null) {
+                    asyncResponseTransformer.exceptionOccurred(e);
+                }
+            });
```
millems: should we return the result of executeFuture.whenComplete? I think the only difference is that if exceptionOccurred throws an exception, the exception is propagated to the result future.

zoewangg: Maybe we should do `runAndLogError` like line 252. Customers probably don't care if exception occurs in `exceptionOccurred`.

millems: Sure! I don't think exceptionOccurred is supposed to raise an exception anyway, right?

zoewangg: Right, at least not in our built-in `AsyncResponseTransformer`s. I'll update the javadoc for `exceptionOccurred` to make it more clearly.

PR #1339, opened 2019-07-10
Add new profile to build tests jar in stability-tests module

zoewangg:
## Description
- Add new profile to build tests jar in stability-tests module to make it easier to run tests on a remote instance.
- Kinesis tests: Only verify data if all events have been received.
-  S3 tests: create bucket and object instead of reusing existing one.

## Motivation and Context
Some issue can only be reproduced when running on a larger ec2 instance, and this will be handy to generate the tests jar and run the tests there.

## License

### DIFF
--- a/test/stability-tests/src/it/java/software/amazon/awssdk/stability/tests/s3/S3AsyncStabilityTest.java
+++ b/test/stability-tests/src/it/java/software/amazon/awssdk/stability/tests/s3/S3AsyncStabilityTest.java
@@ -25,7 +25,6 @@ import java.util.function.IntFunction;
 import org.apache.commons.lang3.RandomStringUtils;
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.BeforeAll;
-import org.junit.jupiter.api.Disabled;
 import software.amazon.awssdk.core.async.AsyncRequestBody;
 import software.amazon.awssdk.core.async.AsyncResponseTransformer;
 import software.amazon.awssdk.stability.tests.exceptions.StabilityTestsRetryableException;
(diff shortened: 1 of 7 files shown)


zoewangg:
Stability tests succeeded 7 times in a row

test/stability-tests/src/it/java/software/amazon/awssdk/stability/tests/s3/S3AsyncStabilityTest.java
```diff
@@ -56,16 +55,29 @@ public void putObject_getObject() {
     }
 
     @RetryableTest(maxRetries = 3, retryableException = StabilityTestsRetryableException.class)
-    @Disabled // Test needs to be fixed so that it doesn't rely on static bucket/key
     public void getLargeObject() throws IOException {
         LOGGER.info(() -> "Starting to test getLargeObject");
-        verifyObjectExist(LARGE_KEY_BUCKET_NAME, LARGE_KEY_NAME, (long) 2e+9);
+        RandomTempFile file = null;
+        try {
+            file = new RandomTempFile((long) 2e+9);
```
dagnir: Would be more efficient just to stream random data on the fly, maybe using something like [`Flowable.repeat()`](http://reactivex.io/RxJava/2.x/javadoc/io/reactivex/Flowable.html#repeat-long-)?

zoewangg: Yeah, but I figured we could test using `FileAsyncRequestBody` to upload large object this way.

dagnir: Hmm I guess that fine. I don't like conflating testing the two components this way, especially since the test is called **get**LargeObject though.  I'd prefer two distinct test cases if we wanted to explicitly test file provider.

zoewangg: Sure, I'll rename the test to make it more clear. We'd need to run get and then put in the same test though, cause the latter depends on the former.

PR #1349, opened 2019-07-15
Work around to make KEEP_ALIVE not private and update name space

zoewangg:
## Description
Work around to make KEEP_ALIVE private and update name space

## License


http-clients/netty-nio-client/src/test/java/software/amazon/awssdk/http/nio/netty/internal/PublisherAdapterTest.java
```diff
@@ -73,22 +77,31 @@
     private RequestContext requestContext;
 
     private CompletableFuture<Void> executeFuture;
+    private ResponseHandler nettyResponseHandler;
```
bmaizels: Should this be inlined?

bmaizels: Should this be inline?

zoewangg: Yeah, this can be inline for now, but I figured it would be handy to keep it there because we might want to add assertions for it in the future. Let me know if you have concerns.

PR #1350, opened 2019-07-15
Increase sdk benchmarks tolerance level

zoewangg:
## Description
Increase sdk benchmarks tolerance level

Some tests occasionally fail and the difference is around 5.1%  , thus increasing tolerance level to 7%

## License


dagnir:
Closing as stale.

Issue #1343, opened 2019-07-12
S3AsyncClient NullPointerException


zoewangg:
I'm still getting the error with 2.7.4 here is the code.

    `final Path tempFile = Files.createTempFile(workDirectory, tempFilePrefix, "tmp");
        AwsCredentialsProvider creds = StaticCredentialsProvider.create(AwsBasicCredentials.create(awsKey, 
    awsSecret));
        S3AsyncClient client = S3AsyncClient.builder().credentialsProvider(creds).build();
        
        final CompletableFuture<GetObjectResponse> futureGet = client.getObject(
                GetObjectRequest.builder()
                        .bucket(s3Bucket)
                        .key(s3Path)
                        .build(),
                AsyncResponseTransformer.toFile(tempFile.toFile()));
    
        futureGet.whenComplete((resp, err) -> {
            try {
                if (resp != null) {
                    System.out.println(resp);
                } else {
                    err.printStackTrace();
                }
            } finally {
                client.close();
            }
        });
    
        futureGet.join();`

Here is the stack trace.
`16:32:23,432 ERROR [stderr] (aws-java-sdk-NettyEventLoop-1-4) java.util.concurrent.CompletionException: software.amazon.awssdk.core.exception.SdkClientException
16:32:23,432 ERROR [stderr] (aws-java-sdk-NettyEventLoop-1-4)   at software.amazon.awssdk.utils.CompletableFutureUtils.errorAsCompletionException(CompletableFutureUtils.java:61)
16:32:23,432 ERROR [stderr] (aws-java-sdk-NettyEventLoop-1-4)   at software.amazon.awssdk.core.internal.http.pipeline.stages.AsyncExecutionFailureExceptionReportingStage.lambda$execute$0(AsyncExecutionFailureExceptionReportingStage.java:51)
16:32:23,433 ERROR [stderr] (aws-java-sdk-NettyEventLoop-1-4)   at software.amazon.awssdk.core.internal.http.pipeline.stages.AsyncExecutionFailureExceptionReportingStage$$Lambda$260/1366089113.apply(Unknown Source)
16:32:23,433 ERROR [stderr] (aws-java-sdk-NettyEventLoop-1-4)   at java.util.concurrent.CompletableFuture$HandleCompletion.run(CompletableFuture.java:1403)
[...]

PR #1360, opened 2019-07-23
Fixed the issue where AsyncStreamingResponseHandler#prepare was not i…

zoewangg:
…nvoked

## Description
The issue is that `AsyncStreamingResponseHandler#prepare` is never invoked before `#onHeaders` in the first attempt as described on javadoc, so `transformFuture` will be null and when there is an exception thrown from `onHeaders` , NPE will be thrown when it is trying to complete the `transformFuture`

https://github.com/aws/aws-sdk-java-v2/blob/393687dc177a3c94814490a9847b1abaaf2fed9d/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/async/AsyncStreamingResponseHandler.java#L55-L56

## Motivation and Context
Related to #1343 

## Testing
Added unit tests.

## Types of changes

## Checklist

## License

### DIFF
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/client/handler/BaseAsyncClientHandler.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/client/handler/BaseAsyncClientHandler.java
@@ -16,6 +16,7 @@
 package software.amazon.awssdk.core.client.handler;
 
 import static software.amazon.awssdk.core.interceptor.SdkInternalExecutionAttribute.ASYNC_RESPONSE_TRANSFORMER_FUTURE;
+import static software.amazon.awssdk.utils.FunctionalUtils.runAndLogError;
 
 import java.util.concurrent.CompletableFuture;
 import java.util.function.Function;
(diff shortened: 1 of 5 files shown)


core/sdk-core/src/main/java/software/amazon/awssdk/core/client/handler/BaseAsyncClientHandler.java
```diff
@@ -77,16 +80,16 @@ protected BaseAsyncClientHandler(SdkClientConfiguration clientConfiguration,
...
 
         ExecutionContext context = createExecutionContext(executionParams);
         context.executionAttributes().putAttribute(ASYNC_RESPONSE_TRANSFORMER_FUTURE, asyncTransformerFuture);
 
         HttpResponseHandler<OutputT> decoratedResponseHandlers =
             decorateResponseHandlers(executionParams.getResponseHandler(), context);
-
-        AsyncStreamingResponseHandler<OutputT, ReturnT> asyncStreamingResponseHandler =
-            new AsyncStreamingResponseHandler<>(asyncResponseTransformer, decoratedResponseHandlers);
+        asyncStreamingResponseHandler.responseHandler(decoratedResponseHandlers);
```
zoewangg: Injecting a responseHandler after creation of asyncStreamingResponseHanlder is not ideal, but let me know if you have any other ideas. Basically, we need `asyncStreamingResponseHandler.prepare()` to be invoked upfront.

millems: Sounds like a challenge! It's not ideal at all...

millems: As discussed, we probably shouldn't capture the future in the `AsyncStreamingResponseHandler`, but I guess I'm okay with the mutation here. It should be on a volatile variable, though...

I still think this is more of an architectural issue than an unfixable problem.

PR #1361, opened 2019-07-24
Update HandlerRemovingChannelPool to only remove per request handlers…

zoewangg:
## Description
Update `HandlerRemovingChannelPool` to only remove per request handlers if the channel is open or registered to avoid the race condition when the DefaultChannelPipeline is trying to removing the handler at the same time, causing `NoSuchElementException`.

## Testing
Added unit tests.  Will run integ tests and stability tests.

## Types of changes

## Checklist

## License

### DIFF
--- a/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/HandlerRemovingChannelPool.java
+++ b/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/HandlerRemovingChannelPool.java
@@ -70,10 +70,17 @@ public class HandlerRemovingChannelPool implements ChannelPool {
 
     private void removePerRequestHandlers(Channel channel) {
         channel.attr(IN_USE).set(false);
-        removeIfExists(channel.pipeline(),
-                       HttpStreamsClientHandler.class,
-                       ResponseHandler.class,
-                       ReadTimeoutHandler.class,
-                       WriteTimeoutHandler.class);
+
+        // Only remove per request handler if the channel is registered
+        // or open since DefaultChannelPipeline would remove handlers if
+        // channel is closed and unregistered
+        // See DefaultChannelPipeline.java#L1403
+        if (channel.isOpen() || channel.isRegistered()) {
+            removeIfExists(channel.pipeline(),
+                           HttpStreamsClientHandler.class,
+                           ResponseHandler.class,
+                           ReadTimeoutHandler.class,
+                           WriteTimeoutHandler.class);
+        }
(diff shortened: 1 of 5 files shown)


http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/HandlerRemovingChannelPool.java
```diff
@@ -70,10 +70,17 @@ public void close() {
...
-        removeIfExists(channel.pipeline(),
-                       HttpStreamsClientHandler.class,
-                       ResponseHandler.class,
-                       ReadTimeoutHandler.class,
-                       WriteTimeoutHandler.class);
+
+        // Only remove per request handler if the channel is registered
+        // or open since DefaultChannelPipeline would remove handlers if
+        // channel is closed and unregistered
+        // See DefaultChannelPipeline.java#L1403
```
millems: Why doesn't removeIfExists prevent the issue? Is the name on the method wrong?

zoewangg: The name is not technically wrong and what it does is to first check if the handler is null and removing it if not null.

```
            if (pipeline.get(handler) != null) {
                pipeline.remove(handler);
            }
```

The issue is that when a channel is being closed and unregistered, another thread will remove handlers on its pipeline, so there is a race condition when two threads trying to remove the handlers at the same time. Basically, right after a thread goes into the if block,  another thread has removed the handler.

millems: With the current catch added, is this line 78 change still needed?

zoewangg: It'd be nice to have line 18 to reduce the chances of the race condition because filling stacktrace for an exception is expensive.

spfink: Are these two calls cheap?

zoewangg: Yeah, should be. They are basically just getters.

zoewangg:
Integration tests and stability tests have passed.

PR #1373, opened 2019-08-05
Added test for bug-fix and updated change log

Quanzzzz:
A new change file was added for the bug fix of redundant requests issue.

## Description

A new change file was added.

## Motivation and Context

Added the change file to indicate that redundant requests issue has been fixed in last update.

## Testing

Used two threads to send request concurrently to test whether the change really works.

## Types of changes

## Checklist

## License

### DIFF
--- a/.changes/next-release/bugfix-AWSSDKforJavav2-1b88003.json
+++ b/.changes/next-release/bugfix-AWSSDKforJavav2-1b88003.json
@@ -1,0 +1,5 @@
+{
+    "category": "AWS SDK for Java v2", 
+    "type": "bugfix", 
+    "description": "Fixed the issue where ByteArrayAsyncRequestBody can send duplicate requests when another request comes in at the same time the subscription completes."
+}
--- a/core/sdk-core/src/test/java/software/amazon/awssdk/core/internal/async/ByteArrayAsyncRequestBodyTest.java
+++ b/core/sdk-core/src/test/java/software/amazon/awssdk/core/internal/async/ByteArrayAsyncRequestBodyTest.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright 2010-2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.core.internal.async;
+
+import static org.junit.Assert.assertTrue;
+import java.nio.ByteBuffer;
+import java.util.concurrent.atomic.AtomicBoolean;
+import org.junit.Test;
+import org.reactivestreams.Subscriber;
+import org.reactivestreams.Subscription;
+
+public class ByteArrayAsyncRequestBodyTest {
+    private class testSubscriber implements Subscriber<ByteBuffer> {
+        private Subscription subscription;
+        protected AtomicBoolean onCompleteCalled = new AtomicBoolean(false);
+
+        @Override
+        public void onSubscribe(Subscription s) {
+            this.subscription = s;
+            s.request(1);
+        }
+
+        @Override
+        public void onNext(ByteBuffer byteBuffer) {
+            
+        }
+
+        @Override
+        public void onError(Throwable throwable) {
+
+        }
+
+        @Override
+        public void onComplete() {
+            subscription.request(1);
+            onCompleteCalled.set(true);
+        }
+    }
+
+    testSubscriber subscriber = new testSubscriber();
+
+    @Test
+    public void concurrentRequests_shouldCompleteNormally() {
+        ByteArrayAsyncRequestBody byteArrayReq = new ByteArrayAsyncRequestBody("Hello World!".getBytes());
+        byteArrayReq.subscribe(subscriber);
+        assertTrue(subscriber.onCompleteCalled.get());
+    }
+
+}


.changes/next-release/bugfix-AWSSDKforJavav2-1b88003.json
```diff
@@ -0,0 +1,5 @@
+{
+    "category": "AWS SDK for Java v2", 
+    "type": "bugfix", 
+    "description": "Fixed a minor error in ByteArrayAsyncRequestBody.java and resolved the issue of redundant requests."
```
zoewangg: Can we update it to something like 

Fixed the issue where  `ByteArrayAsyncRequestBody` can send duplicate requests when another request comes in at the same time the subscription completes.

core/sdk-core/src/test/java/software/amazon/awssdk/core/internal/async/ByteArrayAsyncRequestTest.java
```diff
@@ -0,0 +1,67 @@
...
+        public void onComplete() {
+            onCompleteCalled.set(true);
+        }
+    }
+
+    @Test
+    public void byteArrayAsyncRequestTest() {
+        ByteArrayAsyncRequestBody byteArrayReq = new ByteArrayAsyncRequestBody("Hello World!".getBytes());
+        byteArrayReq.subscribe(subscriber);
+        assertThat(subscriber.onCompleteCalled);
```
zoewangg: It should be `assertThat(subscriber.onCompleteCalled).isTrue()`; otherwise it's not asserting anything.

core/sdk-core/src/test/java/software/amazon/awssdk/core/internal/async/ByteArrayAsyncRequestTest.java
```diff
@@ -0,0 +1,67 @@
...
+import java.nio.ByteBuffer;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.Executor;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.atomic.AtomicBoolean;
+import org.junit.Test;
+import org.reactivestreams.Subscriber;
+import org.reactivestreams.Subscription;
+
+public class ByteArrayAsyncRequestTest {
```
zoewangg: `ByteArrayAsyncRequestBodyTest`?

core/sdk-core/src/test/java/software/amazon/awssdk/core/internal/async/ByteArrayAsyncRequestTest.java
```diff
@@ -0,0 +1,67 @@
...
+    testSubscriber subscriber = new testSubscriber();
+    private class testSubscriber implements Subscriber<ByteBuffer> {
```
zoewangg: nit, can you move the private class below the unit test? It would be easier to read this way.

core/sdk-core/src/test/java/software/amazon/awssdk/core/internal/async/ByteArrayAsyncRequestTest.java
```diff
@@ -0,0 +1,67 @@
...
+        }
+
+        @Override
+        public void onComplete() {
+            onCompleteCalled.set(true);
+        }
+    }
+
+    @Test
+    public void byteArrayAsyncRequestTest() {
```
zoewangg: nit, ideally the unit test name should reveal the intention. How about `concurrentRequets_shouldCompleteNormally`?

dagnir:
Please change to title to reflect that we're actually adding tests here as well, not just a changelog entry.

core/sdk-core/src/test/java/software/amazon/awssdk/core/internal/async/ByteArrayAsyncRequestBodyTest.java
```diff
@@ -0,0 +1,65 @@
...
+
+public class ByteArrayAsyncRequestBodyTest {
+    private class testSubscriber implements Subscriber<ByteBuffer> {
+        private Subscription subscription;
+        protected AtomicBoolean onCompleteCalled = new AtomicBoolean(false);
+
+        @Override
+        public void onSubscribe(Subscription s) {
+            this.subscription = s;
+            CompletableFuture<Void> completableFuture1 = CompletableFuture.runAsync(() -> s.request(1));
```
dagnir: I don't get the purpose of this test. [2.7 of the Reactive Streams spec](https://github.com/reactive-streams/reactive-streams-jvm/blob/v1.0.2/README.md) requires external synchronization on all calls to a `Subscription`:

> A Subscriber MUST ensure that all calls on its Subscription take place from the same thread or provide for respective external synchronization.

So this is not something we need to account for.

cc: @zoewangg

dagnir: Reading @zoewangg's previous review comments, I think she had this in mind

When `onComplete()` is called, try and call `subscription.request(1)` inside of `onComplete()` and see if `onNext` is called again. Is that right Zoe?

zoewangg: I didn't know this was against the spec. I don't remember how exactly http client subscriber ended up sending out concurrent demands. I thought this was the sequence:
`subscriber.request -> subscriber.request -> subscriber.onNext -> subscriber.onComplete`, but I might be wrong.
We can probably try invoke `request` in `onComplete` as you mentioned and see if it throws statckoverflowException without the fix.

Quanzzzz: I tried calling `subscription.request` asynchronously in both `onSubscribe()` and `onComplete()`, and this would throw out a StackOverflowError without the change. But it works after the change.

core/sdk-core/src/test/java/software/amazon/awssdk/core/internal/async/ByteArrayAsyncRequestBodyTest.java
```diff
@@ -0,0 +1,65 @@
...
+
+public class ByteArrayAsyncRequestBodyTest {
+    private class testSubscriber implements Subscriber<ByteBuffer> {
+        private Subscription subscription;
+        protected AtomicBoolean onCompleteCalled = new AtomicBoolean(false);
+
+        @Override
+        public void onSubscribe(Subscription s) {
+            this.subscription = s;
+            CompletableFuture<Void> completableFuture = CompletableFuture.runAsync(() -> s.request(1));
```
dagnir: You don't need to run these async anymore, just call `request()` directly

PR #1354, opened 2019-07-18
Implement CloudWatch MetricPublisher

varunnvs92:

PR #1375, opened 2019-08-06
Metric publisher implementation for CloudWatch

spfink:
Continuing from https://github.com/aws/aws-sdk-java-v2/pull/1354

## Description
Adds a metric publisher implementation for publishing collected metrics to CloudWatch.

## Testing
Tests included.

## Types of changes

## Checklist

## License


metrics-publishers/cloudwatch-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/cloudwatch/internal/Producer.java
```diff
@@ -0,0 +1,101 @@
...
+
+            results.addAll(metricTransformer.transform(metricRegistry));
+
+            for (MetricRegistry mr : metricRegistry.apiCallAttemptMetrics()) {
+                results.addAll(metricTransformer.transform(mr));
+            }
+
+            queue.addAll(results);
+            log.debug(() -> "Number of metrics added to queue: " + results.size());
+        } catch (IllegalStateException exception) {
```
zoewangg: Do we need to catch `IllegalStateException` just for extra logging? 
I feel the logging at line 65 is sufficient

metrics-publishers/cloudwatch-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/cloudwatch/internal/Producer.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    /**
+     * Builder class to create instances of {@link Producer}
+     */
+    public static final class Builder {
+
+        private BlockingQueue<MetricDatum> queue;
+        private MetricTransformer metricTransformer;
+
```
zoewangg: missing private ctor

metrics-publishers/cloudwatch-publisher/src/test/java/software/amazon/awssdk/metrics/publishers/cloudwatch/CloudWatchPublisherIntegrationTest.java
```diff
@@ -0,0 +1,110 @@
...
+import software.amazon.awssdk.metrics.meter.LongCounter;
+import software.amazon.awssdk.metrics.registry.DefaultMetricRegistry;
+import software.amazon.awssdk.metrics.registry.MetricRegistry;
+import software.amazon.awssdk.services.cloudwatch.CloudWatchAsyncClient;
+import software.amazon.awssdk.services.cloudwatch.model.MetricDatum;
+import software.amazon.awssdk.services.cloudwatch.model.PutMetricDataRequest;
+import software.amazon.awssdk.testutils.service.AwsIntegrationTestBase;
+
+@RunWith(MockitoJUnitRunner.class)
+public class CloudWatchPublisherIntegrationTest extends AwsIntegrationTestBase {
```
zoewangg: Hmm, this is not integration test though. Can we add some integration tests?

metrics-publishers/cloudwatch-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/cloudwatch/CloudWatchPublisher.java
```diff
@@ -0,0 +1,262 @@
...
+
+        try {
+            while (queue.peek() != null && ++count <= MAX_SERVICE_CALLS_PER_PUBLISH) {
+                futures.add(consumer.call());
+            }
+        } catch (Throwable throwable) {
+            log.debug(() -> "An error occurred when uploading metrics to CloudWatch.", throwable);
+        }
+
+        CompletableFuture<Void> finalFuture = CompletableFuture.allOf(futures.toArray(new CompletableFuture[futures.size()]));
```
zoewangg: Do we need to cancel other futures if one of future fails of non-recoverable errors like 400? because it's likely all the futures will fail of the same reason. We are doing it for transfermanager.

https://github.com/aws/aws-sdk-java-v2/blob/s3-transfermanager/utils/src/main/java/software/amazon/awssdk/utils/CompletableFutureUtils.java#L113

metrics-publishers/cloudwatch-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/cloudwatch/CloudWatchPublisher.java
```diff
@@ -0,0 +1,262 @@
...
+        publishFutures.add(finalFuture);
```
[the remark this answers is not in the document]
spfink: Removed for now. Not sure what Varun's plan for this was...

metrics-publishers/cloudwatch-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/cloudwatch/internal/Consumer.java
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    /**
+     * If metrics are present in {@link #queue}, uploads them to Amazon CloudWatch.
+     */
+    @Override
+    public CompletableFuture<PutMetricDataResponse> call() throws Exception {
+        List<MetricDatum> metricDatums = metricDatums();
+        if (metricDatums.isEmpty()) {
+            return CompletableFuture.completedFuture(null);
```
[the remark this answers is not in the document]
spfink: I think returning just an empty PutMetricDataResponse object would be better here.

metrics-publishers/cloudwatch-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/cloudwatch/CloudWatchPublisher.java
spfink: Why do we think that one 400 will make all 400?

zoewangg: It might not necessary be 400, maybe something like permission issue and all requests would fail if the credential is not correct. We don't have to do it right now, just something to think about.

metrics-publishers/cloudwatch-metrics-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/cloudwatch/CloudWatchMetricsPublisher.java
```diff
@@ -0,0 +1,258 @@
...
+    }
+
+    @Override
+    public void close() {
+        try {
+            if (producerExecutorService != null) {
+                producerExecutorService.shutdown();
+            }
+
+            if (client != null) {
```
zoewangg: Should we do try-catch for individual resources here so that the failure to close the first resource would not affect the closing of the subsequent resources?

metrics-publishers/cloudwatch-metrics-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/cloudwatch/internal/MetricConsumer.java
```diff
@@ -0,0 +1,130 @@
...
+     * CloudWatch limits number of {@link MetricDatum} to 20 per {@link PutMetricDataRequest}.
+     */
+    private static final int MAX_METRIC_PER_REQUEST = 20;
+
+    private final CloudWatchAsyncClient client;
+    private final BlockingQueue<MetricDatum> queue;
+    private final String namespace;
+
+    private MetricConsumer(Builder builder) {
+        this.client = Validate.notNull(builder.cloudWatchClient, "CloudWatch client cannot be null");
```
zoewangg: nit, we can use `Validate.paramNotNull(builder.cloudWatchClient, "cloudWatchClient")`

metrics-publishers/cloudwatch-metrics-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/cloudwatch/internal/MetricConsumer.java
```diff
@@ -0,0 +1,130 @@
...
+        this.queue = Validate.notNull(builder.queue, "Queue cannot be null");
+        this.namespace = Validate.notNull(builder.namespace, "Namespace cannot be null or empty.");
```
zoewangg: This seems only checking null and not checking empty

metrics-publishers/cloudwatch-metrics-publisher/src/test/java/software/amazon/awssdk/metrics/publishers/cloudwatch/CloudWatchPublisherIntegrationTest.java
```diff
@@ -0,0 +1,110 @@
...
+import software.amazon.awssdk.metrics.meter.LongCounter;
+import software.amazon.awssdk.metrics.registry.DefaultMetricRegistry;
+import software.amazon.awssdk.metrics.registry.MetricRegistry;
+import software.amazon.awssdk.services.cloudwatch.CloudWatchAsyncClient;
+import software.amazon.awssdk.services.cloudwatch.model.MetricDatum;
+import software.amazon.awssdk.services.cloudwatch.model.PutMetricDataRequest;
+import software.amazon.awssdk.testutils.service.AwsIntegrationTestBase;
+
+@RunWith(MockitoJUnitRunner.class)
+public class CloudWatchPublisherIntegrationTest extends AwsIntegrationTestBase {
```
zoewangg: Can we rename the test since it's not an integration test? Can we add an integration test?

metrics-publishers/cloudwatch-metrics-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/cloudwatch/internal/MetricProducer.java
```diff
@@ -0,0 +1,100 @@
...
+@SdkInternalApi
+public final class MetricProducer {
+    private static final Logger log = Logger.loggerFor(MetricProducer.class);
+
+    private final BlockingQueue<MetricDatum> queue;
+    private final MetricTransformer metricTransformer;
+
+    private MetricProducer(Builder builder) {
+        this.queue = Validate.notNull(builder.queue, "Queue cannot be null");
+        this.metricTransformer = builder.metricTransformer != null ? builder.metricTransformer
```
zoewangg: Can customer provide `metricTransformer`? It doesn't look like so from the `CloudWatchMetricsPublisher` apis. 
Should we remove it from the builder?

metrics-publishers/cloudwatch-metrics-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/cloudwatch/CloudWatchMetricsPublisher.java
```diff
@@ -0,0 +1,258 @@
...
+        }
+    }
+
+    @Override
+    public CompletableFuture<Void> publish() {
+        List<CompletableFuture<PutMetricDataResponse>> futures = new ArrayList<>();
+        int count = 0;
+
+        try {
+            while (queue.peek() != null && ++count <= MAX_SERVICE_CALLS_PER_PUBLISH) {
```
zoewangg: What do we think about moving the queue in the `MetricConsumer` class? This class doesn't need to have knowledge of the queue at all and two classes having access to the same queue might be problematic in the future.

metrics-publishers/cloudwatch-metrics-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/cloudwatch/internal/MetricConsumer.java
```diff
@@ -0,0 +1,130 @@
...
+        PutMetricDataRequest request = PutMetricDataRequest.builder()
+                                                           .namespace(namespace)
+                                                           .metricData(metricDatums)
+                                                           .build();
+
+        log.debug(() -> "Call putMetricData API to upload metrics to CloudWatch");
+        return client.putMetricData(request);
+    }
+
+    private List<MetricDatum> metricDatums() {
```
zoewangg: Do we need this list? Can we poll from the queue directly?

metrics-publishers/cloudwatch-metrics-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/cloudwatch/internal/MetricProducer.java
```diff
@@ -0,0 +1,100 @@
...
+                                                                   : MetricTransformer.getInstance();
+    }
+
+    /**
+     * Add the metrics (both top-level and perAttempt metrics) to the {@link #queue}.
+     * @param metricRegistry
+     */
+    public void addMetrics(MetricRegistry metricRegistry) {
+        try {
+            List<MetricDatum> results = new ArrayList<>();
```
zoewangg: Same question, do we need this list? can we off the item to the queue directly?

metrics-publishers/cloudwatch-metrics-publisher/src/test/java/software/amazon/awssdk/metrics/publishers/cloudwatch/CloudWatchPublisherIntegrationTest.java
spfink: Renamed. Will add integration test once I combine this and the other PR.

metrics-publishers/cloudwatch-metrics-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/cloudwatch/internal/MetricProducer.java
spfink: Removed and added note to sync with Varun on this. Not sure why it was ever exposed.

metrics-publishers/cloudwatch-metrics-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/cloudwatch/CloudWatchMetricsPublisher.java
spfink: As in move this logic to the MetricConsumer class? Not sure I follow otherwise.

metrics-publishers/cloudwatch-metrics-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/cloudwatch/internal/MetricConsumer.java
spfink: We have to have a separate list to guarantee the size of the data being published to cloudwatch.

metrics-publishers/cloudwatch-metrics-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/cloudwatch/internal/MetricProducer.java
spfink: In this case, it helps maintain a bit of grouping. Rather than adding to the transforming and adding to the queue X times, where other threads could also be touching the queue, this will add the entire list all together at once.

metrics-publishers/cloudwatch-metrics-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/cloudwatch/CloudWatchMetricsPublisher.java
spfink: Done

metrics-publishers/cloudwatch-metrics-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/cloudwatch/internal/MetricConsumer.java
spfink: Updated

metrics-publishers/cloudwatch-metrics-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/cloudwatch/internal/MetricConsumer.java
spfink: Fixed

metrics-publishers/cloudwatch-metrics-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/cloudwatch/CloudWatchMetricsPublisher.java
zoewangg: Yeah, I meant moving the creation of the queue and this logic into `MetricConsumer` class.

spfink: The MetricProducer also uses the queue.

zoewangg: Oh I missed that part. Do we need to check `queue.peek` here? can we move this logic to consumer?

dagnir:
What's the status of this PR? @spfink

PR #1402, opened 2019-09-03
ApacheUtils: Only warn once on class-load if setNormalizeUri() is not…

bmaizels:
… found in the Apache HttpClient library rather than on every request

## Description
If customers were using a version of apache httpclient <4.5.8 they would get a warning on every request recommending they upgrade to >=4.5.9. After this change, customers will only be warned once per SDK class-load instead of on every request.

## Motivation and Context

The repeated warnings were creating unnecessary spam in logs.

## Testing

Unit tests ran against Apache  httpcomponents 4.5.9 and 4.5.2 for comparison.

## Types of changes

## Checklist

## License


utils/src/main/java/software/amazon/awssdk/utils/ReflectionMethodInvoker.java
```diff
@@ -89,21 +91,32 @@ public R invoke(T obj, Object... args) throws NoSuchMethodException {
...
-        }
+    /**
+     * Initializes the method invoker by finding and caching the target method from the target class that will be used
+     * for subsequent invoke operations. If the invoker is already initialized this call will do nothing.
+     * @throws NoSuchMethodException if the JVM could not find a method matching the signature specified in the
+     * initialization of this proxy.
+     */
+    public void initialize() throws NoSuchMethodException {
+        getTargetMethod();
+    }
```
millems: Is this backwards-compatible? We usually remain backwards-compatible with SdkProtectedApi classes like these.

bmaizels: Mostly. Invoke() will initialize the object even if initialize() is not explicitly called, and so will behave as the older version did. What has changed, however, is that invoke() used to cache a thrown NoSuchMethodException and it no longer does that which will result in a small efficiency drop in cases where it is repeatedly called and the method does not exist. Functionally it behaves the same, just with additional features/controls.

Issue #1182, opened 2019-04-02
S3Client.putObjectAcl() Causes MalformedACLError


LukeHornibrookHaventec:
Calling `S3Client.putObjectAcl()` with an `AccessControlPolicy` triggers a `MalformedACLError` with error message "The XML you provided was not well-formed or did not validate against our published schema".

## Expected Behavior
PutObjectACL requests should complete successfully.

## Current Behavior
The request fails with a `MalformedACLError` code and message "The XML you provided was not well-formed or did not validate against our published schema". An example request body is:

```xml
<?xml version="1.0" encoding="UTF-8"?><AccessControlPolicy xmlns="http://s3.amazonaws.com/doc/2006-03-01/ "><AccessControlList><Grant><Grantee><ID>***</ID></Grantee><Permission>FULL_CONTROL</Permission></Grant><Grant><Grantee><ID>***</ID></Grantee><Permission>READ</Permission></Grant></AccessControlList><Owner><ID>***</ID></Owner></AccessControlPolicy>
```

The error may be caused by incorrect ordering of XML elements. The S3 schema at http://doc.s3.amazonaws.com/2006-03-01/AmazonS3.xsd  defines the `AccessControlPolicy` type as:

```xml
<xsd:complexType name="AccessControlPolicy">
    <xsd:sequence>
      <xsd:element name="Owner" type="tns:CanonicalUser"/>
      <xsd:element name="AccessControlList" type="tns:AccessControlList"/>
    </xsd:sequence>
  </xsd:complexType>
```

In particular, the `Owner` element occurs before `AccessControlList` (whereas the XML generated by the Java 2 SDK has these elements in the opposite order).

This issue probably affects PutBucketACL too.

## Possible Solution
As a workaround, the `grant*` methods on `PutObjectAclRequest` can be used instead of setting the `AccessControlPolicy`; these use request headers instead of XML in the body. 

## Steps to Reproduce
```java
S3Client.create().putObjectAcl(x -> {
	x.bucket("<bucket>");
	x.key("<key>");
	x.accessControlPolicy(y -> {
		y.owner(z -> {
			z.id ("<owner canonical id>");
		});
		y.grants(Grant.builder()
			.permission(Permission.READ )
			.grantee(z -> {
				z.id ("<grantee canonical id>");
			})
			.build());
	});
});
```

## Your Environment
* AWS Java SDK version used: 2.5.17

zoewangg:
[...]

PR #1404, opened 2019-09-05
Add marshalling support for models with xml attribute. 

zoewangg:
## Description
Add marshalling support for models with xml attribute. See #1182 

The root cause is that the marshalled request was missing xml namespace prefix  `xmlns:xsi `and element type `xsi:type` was put in the wrong place.

expected:
```
<AccessControlList>
    <Grant>
      <Grantee
               xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
               xsi:type="CanonicalUser">
            <ID>xxx</ID>
        </Grantee>
        <Permission>READ</Permission>
    </Grant>
</AccessControlList>
```

actual:
```
<AccessControlList>
    <Grant>
        <Grantee>
            <ID>xxx</ID>
            <xsi:type>CanonicalUser</xsi:type>
        </Grantee>
        <Permission>READ</Permission>
    </Grant>
</AccessControlList>
```

Note that Unmarshalling is still broken (`Grantee#type` is not populated) and is out of scope of this PR.

## Testing
- Added codegen tests.
- Added s3 integration tests and function tests.

## Types of changes

## Checklist

## License


core/sdk-core/src/main/java/software/amazon/awssdk/core/traits/XmlNamespaceTrait.java
```diff
@@ -0,0 +1,87 @@
...
+package software.amazon.awssdk.core.traits;
+
+import java.util.function.Function;
+import software.amazon.awssdk.annotations.SdkProtectedApi;
+
+/**
+ * Trait to include the xml namespace attributes
+ */
+@SdkProtectedApi
+public class XmlNamespaceTrait implements Trait {
```
dagnir: Hmm after thinking about this a bit, can we just make a more generic `XmlAttributeTrait`?

Then when generating the `SdkField` for `Grantee`, we just add two `XmlAttributeTrait`s:

```
.traits(new XmlAttribute("xmlns:xsi", (grantee) -> "http://www.w3.org/2001/XMLSchema-instance")
        new XmlAttribute("xsi:type", (grantee) -> grantee.type()))
```

zoewangg: Yeah, like that idea. Let me try it

zoewangg: How about this?

```java
    private static final Map<String, AttributeAccessors> ATTRIBUTES_MAP = new LinkedHashMap<>();
    static {
        ATTRIBUTES_MAP.put("xmlns:xsi", AttributeAccessors.builder()
                                                          .attributeGetter((ignore) -> "http://www.w3.org/2001/XMLSchema-instance")
                                                          .build());
        ATTRIBUTES_MAP.put("xsi:type", AttributeAccessors.builder()
                                                         .attributeGetter(t -> ((Grantee) t).typeAsString())
                                                         .build());
    }

    private static final SdkField<Grantee> GRANTEE_FIELD = SdkField
            .<Grantee> builder(MarshallingType.SDK_POJO)
        .getter(getter(Grant::grantee))
        .setter(setter(Builder::grantee))
        .constructor(Grantee::builder)
        .traits(LocationTrait.builder().location(MarshallLocation.PAYLOAD).locationName("Grantee")
                    .unmarshallLocationName("Grantee").build(),
                XmlAttributesTrait.create(ATTRIBUTES_MAP)
                ).build();



```

dagnir: Nice, I like it.

PR #1415, opened 2019-09-12
Expose instance signature through EC2MetadataUtils

notdryft:
A method was added in aws-sdk-java to help retrieve instance identify signature. Here is a pull request to port it to v2.

See https://github.com/aws/aws-sdk-java/pull/1653

### DIFF
--- a/core/regions/src/test/java/software/amazon/awssdk/regions/internal/util/EC2MetadataUtilsServer.java
+++ b/core/regions/src/test/java/software/amazon/awssdk/regions/internal/util/EC2MetadataUtilsServer.java
@@ -110,6 +110,10 @@ public class EC2MetadataUtilsServer {
 
         } else if (path.equals("/latest/dynamic/instance-identity/document")) {
             outputInstanceInfo(output);
+
+        } else if (path.equals("/latest/dynamic/instance-identity/signature")) {
+            outputInstanceSignature(output);
+
         } else {
             throw new RuntimeException("Unknown path: " + path);
         }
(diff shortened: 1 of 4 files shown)


notdryft:
Hi, just added a test for `EC2MetadataUtils.getInstanceSignature()`

> Can you confirm that this pull request can be released under the Apache 2 license?

I do confirm that this pull request can be released under the Apache 2 license.

notdryft:
About the changelog... should I modify it directly or is there an external process to do it? Not sure which version I should put it under.

zoewangg:
You can add a changelog entry by running the `scripts/new-change` script and follow the instructions, then commit the new file created by the script in `.changes/next-release` with your changes.

Once you add the changelog entry, could you squash your commits into one?

notdryft:
Neat script !

All done, squashed and all :)

core/regions/src/test/java/software/amazon/awssdk/regions/internal/util/EC2MetadataUtilsServer.java
```diff
@@ -110,6 +110,10 @@ private void handleConnection(BufferedReader input,
 
         } else if (path.equals("/latest/dynamic/instance-identity/document")) {
             outputInstanceInfo(output);
+
+        } else if (path.equals("instance-identity/signature")) {
```
zoewangg: Hmm, integ tests failed 

```
RROR] software.amazon.awssdk.regions.util.EC2MetadataUtilsIntegrationTest.testInstanceSignature(software.amazon.awssdk.regions.util.EC2MetadataUtilsIntegrationTest) 
[ERROR]   Run 1: EC2MetadataUtilsIntegrationTest.testInstanceSignature:117 expected:<foobar> but was:<null> 
[ERROR]   Run 2: EC2MetadataUtilsIntegrationTest.testInstanceSignature:117 expected:<foobar> but was:<null> 
[ERROR]   Run 3: EC2MetadataUtilsIntegrationTest.testInstanceSignature:117 expected:<foobar> but was:<null> 
```

I think this should be ` else if (path.equals("/latest/dynamic/instance-identity/signature"))`

notdryft: Fixed

notdryft:
Updated

PR #1444, opened 2019-09-27
Update benchmark baseline because of performance improvement from #1436 

zoewangg:
## Description
Update benchmark baseline because of performance improvement from #1436 

## License


dagnir:
:shipit:

PR #1456, opened 2019-10-04
Add initial design for URL presigners.

millems:


docs/design/core/presigners/README.md
```diff
@@ -0,0 +1,525 @@
...
+     * This may differ from the local clock, based on the skew between the local and AWS service clocks.
+     */
+    Instant expiration();
+
+    /**
+     * Returns true if the url returned by the url method can be executed in a browser.
+     *
+     * This is true when the HTTP request method is GET, and hasSignedHeaders and hasSignedPayload are false.
+     */
+    boolean isBrowserCompatible();
```
dagnir: Hmm thoughts about renaming this? I would consider client-side JavaScript to be technically "executed in a browser" as well, in which case some of these conditions don't need to be true.

Maybe `isBrowserNavigable`?

millems: I agree this name isn't ideal. "Navigable" It makes it sound like the browser is a boat, though.

We could consider removing it, but I think it's important for a customer to be able to determine why their presigned URL isn't working when they hit it with a browser.

Maybe `isLinkable`?

dagnir: I like having some form of "navigation" in there since the address bar is also often referred to as the "navigation bar".

`navigableByBrowser`? :P. I think I would be fine with "linkable" as well.

millems: We can see what the others think.

docs/design/core/presigners/README.md
```diff
@@ -0,0 +1,525 @@
...
+
+The proposal above does not distinguish between blocking or non-blocking request presigning. This is because the
+SDK currently only distinguishes between blocking and non-blocking, when it comes to an HTTP client implementation.
+
+The generated presigned request can be executed by a blocking OR non-blocking HTTP client.
+
+In the future, the SDK could implement non-blocking region providers and non-blocking credential providers, at which
+time it could become relevant to distinguish between blocking or non-blocking URL presigners.
+
+For this reason, the `presigner()` method to get to a pre-configured URL presigner will ONLY be included on the
```
zoewangg: We have `#utilities` on s3 async client though. Do we want to make presigner consistent with it?

millems: Updated the wording to specify that we're not sure and we'll decide when we actually implement the `presigner()` method.

docs/design/core/presigners/README.md
millems: Updated to note that this isn't universally agreed upon as the method name and we can iterate on it in CR.

PR #1462, opened 2019-10-08
Implement arn parser functions

zoewangg:
## Description
Implement arn parser functions. Basically same as v1. 
The only differences are:
- Add more javadocs
- Make it follow v2 pattern
- Make `ArnResource` return `Optional<String> resourceType()` and ` Optional<String> qualifier()` since they are optional.

## Testing
Same tests copied from v1. changed assertions to use assertj

## Types of changes

## Checklist

## License

### DIFF
--- a/core/arns/src/main/java/software/amazon/awssdk/arns/Arn.java
+++ b/core/arns/src/main/java/software/amazon/awssdk/arns/Arn.java
@@ -1,0 +1,373 @@
+/*
+ * Copyright 2010-2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.arns;
+
+import java.util.Objects;
+import java.util.Optional;
+import software.amazon.awssdk.annotations.SdkPublicApi;
+import software.amazon.awssdk.utils.Validate;
+import software.amazon.awssdk.utils.builder.CopyableBuilder;
+import software.amazon.awssdk.utils.builder.ToCopyableBuilder;
+
+/**
+ * The Arns generated and recognized by this code are the Arns described here:
+ *
+ * https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html
+ *
+ * <p>
+ * The primary supported Arn format is:
+ *
+ * <code>
+ * arn::&#60;partition&#62;::&#60;service&#62;::&#60;region&#62;::&#60;account&#62;::&#60;resource&#62;
+ * </code>
+ *
+ * <p>
+ * {@link #resourceAsString()} returns everything after the account section of the Arn
+ * as a single string.
+ *
+ * <p>
+ * However, the following Arn formats are supported where the values are present and well
+ * formatted through {@link #resource()}:
+ *
+ * <pre>
+ * arn:&#60;partition&#62;:&#60;service&#62;:&#60;region&#62;:&#60;account&#62;:&#60;resourcetype&#62;/resource
+ * arn:&#60;partition&#62;:&#60;service&#62;:&#60;region&#62;:&#60;account&#62;:&#60;resourcetype&#62;/resource/qualifier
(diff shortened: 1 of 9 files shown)


core/arns/src/main/java/software/amazon/awssdk/arns/Arn.java
```diff
@@ -0,0 +1,372 @@
...
+    private final String region;
+    private final String accountId;
+    private final String resource;
+    private final ArnResource arnResource;
+
+    private Arn(DefaultBuilder builder) {
+        this.partition = Validate.paramNotBlank(builder.partition, "partition");
+        this.service = Validate.paramNotBlank(builder.service, "service");
+        this.region = Validate.paramNotBlank(builder.region, "region");
+        this.accountId = Validate.paramNotBlank(builder.accountId, "accountId");
```
bmaizels: Why did we change the contract from the v1 Arn here where these are optional fields?

core/arns/src/main/java/software/amazon/awssdk/arns/ArnResource.java
```diff
@@ -0,0 +1,237 @@
...
+@SdkPublicApi
+public final class ArnResource implements ToCopyableBuilder<ArnResource.Builder, ArnResource> {
+
+    private final String resourceType;
+    private final String resource;
+    private final String qualifier;
+
+    private ArnResource(DefaultBuilder builder) {
+        this.resourceType = builder.resourceType;
+        this.resource = builder.resource;
```
bmaizels: If we are asserting that resource is not optional, should we validate that here? Either that or change the getter to Optional. Will it ever make sense to have ArnResource.fromString("")?

core/arns/src/main/java/software/amazon/awssdk/arns/ArnResource.java
```diff
@@ -0,0 +1,237 @@
...
+        return result;
+    }
+
+    @Override
+    public Builder toBuilder() {
+        return builder()
+            .resource(resource)
+            .resourceType(resourceType)
+            .qualifier(qualifier);
+    }
```
bmaizels: There is no test coverage for this code.

core/arns/src/main/java/software/amazon/awssdk/arns/Arn.java
```diff
@@ -0,0 +1,372 @@
...
+
+    @Override
+    public Builder toBuilder() {
+        return builder().accountId(accountId)
+                        .partition(partition)
+                        .region(region)
+                        .resource(resource)
+                        .service(service)
+            ;
+    }
```
bmaizels: There is no test coverage for this code.

core/arns/src/main/java/software/amazon/awssdk/arns/Arn.java
```diff
@@ -0,0 +1,372 @@
...
+
+    @Override
+    public int hashCode() {
+        int hash = partition.hashCode();
+        hash = 31 * hash + service.hashCode();
+        hash = 31 * hash + region.hashCode();
+        hash = 31 * hash + accountId.hashCode();
+        hash = 31 * hash + resource.hashCode();
+        return hash;
+    }
```
bmaizels: I don't see test coverage for this code.

core/arns/src/main/java/software/amazon/awssdk/arns/ArnResource.java
```diff
@@ -0,0 +1,237 @@
...
+        return Objects.equals(qualifier, that.qualifier);
+    }
+
+    @Override
+    public int hashCode() {
+        int result = resourceType != null ? resourceType.hashCode() : 0;
+        result = 31 * result + (resource != null ? resource.hashCode() : 0);
+        result = 31 * result + (qualifier != null ? qualifier.hashCode() : 0);
+        return result;
+    }
```
bmaizels: I don't see test coverage for this code.

core/arns/src/main/java/software/amazon/awssdk/arns/Arn.java
zoewangg: Didn't know we need test `hashCode` because it's generated, but I can see the value of adding one, so added.

core/arns/src/main/java/software/amazon/awssdk/arns/Arn.java
zoewangg: added tests

core/arns/src/main/java/software/amazon/awssdk/arns/ArnResource.java
zoewangg: added tests

core/arns/src/main/java/software/amazon/awssdk/arns/Arn.java
zoewangg: This is my fault. Updated.

core/arns/src/main/java/software/amazon/awssdk/arns/ArnResource.java
zoewangg: Yeah, added validation because the resource is validated in `Arn` as well.

PR #1463, opened 2019-10-08
Fixes for SdkAutoConstructList/Map

### LINKED ISSUE #1445: DefaultSdkAutoConstructList.equals() does not adhere to the List.equals() contract
I'm using the AWS Java SDK 2 version 2.8.3 on Windows 10 with OpenJDK 11.

The Java `java.util.List.equals()` contract requires all lists to implement `equals()` so that they compare equal to any other implementation of `List` with the same number of elements in the same order:

```java
    /**
     * Compares the specified object with this list for equality.  Returns
     * {@code true} if and only if the specified object is also a list, both
     * lists have the same size, and all corresponding pairs of elements in
     * the two lists are <i>equal</i>.  (Two elements {@code e1} and
     * {@code e2} are <i>equal</i> if {@code Objects.equals(e1, e2)}.)
     * In other words, two lists are defined to be
     * equal if they contain the same elements in the same order.  This
     * definition ensures that the equals method works properly across
     * different implementations of the {@code List} interface.
     *
     * @param o the object to be compared for equality with this list
     * @return {@code true} if the specified object is equal to this list
     */
    boolean equals(Object o);
```

Unless I'm missing it somewhere, the `software.amazon.awssdk.core.util.DefaultSdkAutoConstructList` does not implement `equals()` and therefore does not adhere to the `List.equals()` contract. The same goes for `List.hashCode()`.

dagnir:
## Description
Fix the implementations of equals() and hashCode() to follow the general
contract of equals and hashCode.

Also ensure that toString returns readable results.

Fixes #1445

## Motivation and Context
Bug fix.

## Testing
New unit tests.

## Types of changes

## Checklist

## License


core/sdk-core/src/main/java/software/amazon/awssdk/core/util/DefaultSdkAutoConstructList.java
```diff
@@ -33,7 +33,7 @@
 public final class DefaultSdkAutoConstructList<T> implements SdkAutoConstructList<T> {
     private static final DefaultSdkAutoConstructList INSTANCE = new DefaultSdkAutoConstructList();
 
-    private final List impl = Collections.unmodifiableList(Collections.emptyList());
+    private final List impl = Collections.emptyList();
```
dagnir: Changed this because `emptyMap`/`emptyList` are already immutable so wrapping in `unmodifiable` is redundant.

Issue #1466, opened 2019-10-09
SqsAsyncClient is failing unpredictably during streaming


jaylynstoesz:
I'm reading from SQS via [Alpakka](https://doc.akka.io/docs/alpakka/current/sqs.html) using the async client. It appears to randomly stop receiving messages from the queue, anywhere from a few minutes to a few hours apart. It appears to happen more frequently when volume is higher, though not necessarily during peak times.

## Expected Behavior
My service continuously polls messages from SQS, processes them, and deletes them from the queue. Invalid messages are sent to a dead letter queue. In the attached screenshot of the queue monitoring dashboard, I would expect the **NumberOfMessagesReceived** graph to match the **NumberOfMessagesSent** graph.

## Current Behavior
The SQS client randomly throws `java.io.IOException: Server failed to send complete response` and stops polling messages.
![Screen Shot 2019-10-09 at 1 29 35 PM](https://user-images.githubusercontent.com/11892230/66514369-31084700-ea9a-11e9-9e51-ae1d917a3422.png)

## Steps to Reproduce (for bugs)
```
import akka.actor.ActorSystem
import akka.stream.ActorMaterializer
import akka.stream.alpakka.sqs.javadsl.SqsSource
import akka.stream.alpakka.sqs.SqsSourceSettings
import akka.stream.javadsl.Sink
import software.amazon.awssdk.http.nio.netty.NettyNioAsyncHttpClient
import software.amazon.awssdk.regions.Region
import software.amazon.awssdk.services.sqs.model.Message
import software.amazon.awssdk.services.sqs.SqsAsyncClient

//...

fun <T> subscribe() {
    /* Initialize Akka */
    val actorSystem = ActorSystem.create()

    /* Provision async SQS client */
    val sqsAsyncClient = SqsAsyncClient
        .builder()
        .httpClientBuilder(NettyNioAsyncHttpClient.builder())
        .region(Region.US_EAST_1)
        .build()

    /* Provision Alpakka SQS module (Akka Streams) */
    val source = SqsSource.create(
        "https://sqs.us-east-1.amazonaws.com/000000000000/queue-name",
        SqsSourceSettings.Defaults(),
        sqsAsyncClient
    )

    /* Akka receives between 1k - 1.5k messages per minute from SQS */
    source
        .map { message: Message -> message.messageId() }
        .runWith(Sink.ignore(), ActorMaterializer.create(actorSystem))
        .handle { _, error ->
[...]

Issue #1380, opened 2019-08-11
CloudWatch putMetricData sometimes produce a java.io.IOException: Server failed to send complete response 


oripwk:
This issue is exactly as #452, but instead of S3 it happens in CloudWatch.

## Expected Behavior
`putMetricData(…)` should not throw an exception.

## Current Behavior
`java.io.IOException: Server failed to send complete response` is thrown several times a minute. Most of the requests seem to succeed.
```
java.util.concurrent.CompletionException: software.amazon.awssdk.core.exception.SdkClientException
	at software.amazon.awssdk.utils.CompletableFutureUtils.errorAsCompletionException(CompletableFutureUtils.java:61)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.AsyncExecutionFailureExceptionReportingStage.lambda$execute$0(AsyncExecutionFailureExceptionReportingStage.java:51)
	at java.util.concurrent.CompletableFuture.uniHandle(CompletableFuture.java:822)
	at java.util.concurrent.CompletableFuture$UniHandle.tryFire(CompletableFuture.java:797)
	at java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:474)
	at java.util.concurrent.CompletableFuture.completeExceptionally(CompletableFuture.java:1977)
	at software.amazon.awssdk.utils.CompletableFutureUtils.lambda$forwardExceptionTo$0(CompletableFutureUtils.java:75)
	at java.util.concurrent.CompletableFuture.uniWhenComplete(CompletableFuture.java:760)
	at java.util.concurrent.CompletableFuture$UniWhenComplete.tryFire(CompletableFuture.java:736)
	at java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:474)
	at java.util.concurrent.CompletableFuture.completeExceptionally(CompletableFuture.java:1977)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.AsyncRetryableStage$RetryExecutor.retryErrorIfNeeded(AsyncRetryableStage.java:175)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.AsyncRetryableStage$RetryExecutor.retryIfNeeded(AsyncRetryableStage.java:126)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.AsyncRetryableStage$RetryExecutor.lambda$execute$0(AsyncRetryableStage.java:107)
[...]

PR #1476, opened 2019-10-17
Update HealthCheckedChannelPool to check KEEP_ALIVE attribute

zoewangg:
# Description
Update `HealthCheckedChannelPool` to check `KEEP_ALIVE` when acquiring a channel from the pool to avoid soon-to-be inactive channels being picked up by a new request. This should reduce the frequency of `IOException: Server failed to complete response` errors. See #1380 #1466

## Testing
- Added unit tests.
- Manual tests 

Disabled retry and set maxIdleConnectionTimeout = 1 mins on purpose to expose the errors:

Before

```
CloudWatchAsyncStabilityTest: totalRequestCount: 4000, ioExceptionCount: 34

Based on the logs, 28 of ioExceptions were because the connections that were not reusable got picked up
```

After 

```

CloudWatchAsyncStabilityTest:, totalRequestCount: 4000, ioExceptionCount: 0

```

## Types of changes

## Checklist

## License

### DIFF
--- a/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/HealthCheckedChannelPool.java
+++ b/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/HealthCheckedChannelPool.java
@@ -15,6 +15,8 @@
 
 package software.amazon.awssdk.http.nio.netty.internal;
 
+import static software.amazon.awssdk.http.nio.netty.internal.ChannelAttributeKey.KEEP_ALIVE;
+
 import io.netty.channel.Channel;
 import io.netty.channel.EventLoopGroup;
 import io.netty.channel.pool.ChannelPool;
@@ -159,6 +161,12 @@ public class HealthCheckedChannelPool implements ChannelPool {
      * Determine whether the provided channel is 'healthy' enough to use.
      */
     private boolean isHealthy(Channel channel) {
+        // There might be cases where the channel is not reusable but still active at the moment
+        // See https://github.com/aws/aws-sdk-java-v2/issues/1380
+        if (channel.attr(KEEP_ALIVE).get() != null && !channel.attr(KEEP_ALIVE).get()) {
+            return false;
+        }
+
         return channel.isActive();
     }
 }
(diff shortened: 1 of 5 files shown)


zoewangg:
`KEEP_ALIVE` was set from the previous response based on the response headers. `KEEP_ALIVE=false` indicates the connection might be closed at any moment and should not be reused. 

Basically, there could be a race condition when a channel is being closed at the same moment it's being acquired, and a non-reusable connection might gets picked up by a new request.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/HealthCheckedChannelPool.java
```diff
@@ -159,6 +161,12 @@ private void closeIfUnhealthy(Channel channel) {
      * Determine whether the provided channel is 'healthy' enough to use.
      */
     private boolean isHealthy(Channel channel) {
+        // There might be cases where the channel is not reusable but still active at the moment
+        // See https://github.com/aws/aws-sdk-java-v2/issues/1380
+        if (channel.attr(KEEP_ALIVE).get() != null && !channel.attr(KEEP_ALIVE).get()) {
+            return false;
+        }
+
```
dagnir: I think this is fine, but any reason we don't close the channel after the response ins consumed instead?

zoewangg: We do invoke `ChannelHandlerContext#close` in `ResponseHandler` for this case but the channel might not be actually closed and is still active at this point.  It's not guaranteed the channel is closed before it's released into the pool.

Also, because of the fact that the underlying [`HttpStreamsClientHandler`](https://github.com/playframework/netty-reactive-streams/blob/master/netty-reactive-streams-http/src/main/java/com/typesafe/netty/http/HttpStreamsClientHandler.java#L66-L73) from netty-reactive-streams does not close the channel if there are still inflight messages, `channel.close` might not be initiated at all. (This is what we discussed in the morning and we might need to figure out why there is still inflight message when the response is complete)

PR #1480, opened 2019-10-21
Updating Jackson and Netty versions

spfink:


dagnir:
LGTM; can we run a perf test (if you haven't already) to make sure there are no regressions?

spfink:
I ran the integ tests and stability tests locally and they passed.

PR #1481, opened 2019-10-22
Add LastHttpContentHandler

zoewangg:
## Description
Add `LastHttpContentHandler` to flag if the LastHttpContent has been received and don't throw "server failed to complete response" error if the content is buffered.

## Testing
Added unit tests and function tests
Tested with long running canaries tests
Integ tests and stability tests passed

## Types of changes

## Checklist

## License

### DIFF
--- a/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/LastHttpContentHandler.java
+++ b/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/LastHttpContentHandler.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2010-2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.http.nio.netty.internal;
+
+import static software.amazon.awssdk.http.nio.netty.internal.ChannelAttributeKey.LAST_HTTP_CONTENT_RECEIVED_KEY;
+
+import io.netty.channel.ChannelHandler;
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.ChannelInboundHandlerAdapter;
+import io.netty.handler.codec.http.LastHttpContent;
+import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.utils.Logger;
+
+/**
+ * Marks {@code ChannelAttributeKey.LAST_HTTP_CONTENT_RECEIVED_KEY} if {@link LastHttpContent} is received.
+ */
+@SdkInternalApi
+@ChannelHandler.Sharable
+public final class LastHttpContentHandler extends ChannelInboundHandlerAdapter {
+    private static final LastHttpContentHandler INSTANCE = new LastHttpContentHandler();
+    private static final Logger logger = Logger.loggerFor(LastHttpContent.class);
+
+    @Override
+    public void channelRead(ChannelHandlerContext ctx, Object msg) {
+        if (msg instanceof LastHttpContent) {
+            logger.debug(() -> "Received LastHttpContent " + ctx.channel());
+            ctx.channel().attr(LAST_HTTP_CONTENT_RECEIVED_KEY).set(true);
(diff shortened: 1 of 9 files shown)


http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/LastHttpContentHandler.java
```diff
@@ -0,0 +1,49 @@
...
+    public void channelRead(ChannelHandlerContext ctx, Object msg) {
+        if (msg instanceof LastHttpContent) {
+            logger.debug(() -> "Received LastHttpContent " + ctx.channel());
+            ctx.channel().attr(LAST_HTTP_CONTENT_RECEIVED_KEY).set(true);
+        }
+
+        ctx.fireChannelRead(msg);
+    }
+
+    public static LastHttpContentHandler create() {
```
dagnir: nit: `getInstance()`

zoewangg: Oh I thought our naming convention for static factory method is `create`.

https://github.com/aws/aws-sdk-java-v2/blob/master/docs/design/FavorStaticFactoryMethods.md#naming-conventions

dagnir: Believe that only applies when a new instance is returned for each call.

zoewangg: I think we use that for new instance as well as singleton instance in other places.

https://github.com/aws/aws-sdk-java-v2/blob/4c0e42ed66975de42d68df9d508ffd9584d4428f/core/auth/src/main/java/software/amazon/awssdk/auth/credentials/DefaultCredentialsProvider.java#L58-L59

dagnir: Seems misleading to me but it's fine if there's existing classes doing this I guess.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/NettyRequestExecutor.java
```diff
@@ -162,6 +165,7 @@ private boolean tryConfigurePipeline() {
             return false;
         }
 
+        pipeline.addLast(LastHttpContentHandler.create());
```
millems: Did we verify this works for HTTP/2?

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ResponseHandler.java
```diff
@@ -386,7 +383,10 @@ private Throwable wrapException(Throwable originalCause) {
     private void notifyIfResponseNotCompleted(ChannelHandlerContext handlerCtx) {
         RequestContext requestCtx = handlerCtx.channel().attr(REQUEST_CONTEXT_KEY).get();
         boolean responseCompleted = handlerCtx.channel().attr(RESPONSE_COMPLETE_KEY).get();
-        if (!responseCompleted) {
+        boolean lastHttpContentReceived = handlerCtx.channel().attr(LAST_HTTP_CONTENT_RECEIVED_KEY).get();
+        handlerCtx.channel().attr(KEEP_ALIVE).set(false);
```
millems: Wouldn't this get set when we close the channel (change above)?

zoewangg: Yes for most of the cases but there are certain edge cases it would not get set. For example, a response header does not contain `connection: close` (so we set the `KEEP_ALIVE` to true) but `channelInactive` is invoked after the content is buffered, and in this case, the "bad" channel might get picked up again in a new request once it's released in the pool before it's actually closed.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/NettyRequestExecutor.java
zoewangg: Verified with h2 test service. I'll double check with the actual h2 service.

zoewangg: Verified it works. 

Also checked the code and we do convert h2 endstream to a `DefaultLastHttpContent`.

https://github.com/aws/aws-sdk-java-v2/blob/4c0e42ed66975de42d68df9d508ffd9584d4428f/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/Http2ToHttpInboundAdapter.java#L63-L69

millems:
LGTM. @dagnir ?

PR #1498, opened 2019-11-04
Updated crt http client to use latest crt api, which uses byte[] for …

JonathanHenson:
…response body.


http-clients/aws-crt-client/src/main/java/software/amazon/awssdk/http/crt/internal/AwsCrtAsyncHttpStreamAdapter.java
```diff
@@ -99,10 +99,10 @@ public int onResponseBody(HttpStream stream, ByteBuffer bodyBytesIn) {
 
         // Queue a Deep Copy since bodyBytesIn is only guaranteed to contain valid memory for the lifetime of this
```
justinboswell: This comment is no longer accurate

dagnir:
Believe this is superseded by https://github.com/aws/aws-sdk-java-v2/pull/2021 :). Closing this.

PR #1509, opened 2019-11-14
When creating an SdkException without a message, inherit the "cause" exception message.

millems:


core/sdk-core/src/test/java/software/amazon/awssdk/core/exception/SdkExceptionTest.java
```diff
@@ -0,0 +1,19 @@
...
+    @Test
+    public void sdkExceptionMessagesDefaultToChildMessages() {
+        Exception barCause = new Exception("bar");
+        Exception nullCause = new Exception();
+
+        assertThat(SdkException.builder().build().getMessage()).isEqualTo(null);
+        assertThat(SdkException.builder().cause(nullCause).build().getMessage()).isEqualTo(null);
+        assertThat(SdkException.builder().message("foo").build().getMessage()).isEqualTo("foo");
+        assertThat(SdkException.builder().message("foo").cause(barCause).build().getMessage()).isEqualTo("foo");
+        assertThat(SdkException.builder().cause(barCause).build().getMessage()).isEqualTo("bar");
```
bmaizels: Ideally these would be separate tests so as to follow the 'unit test should test just one behavior' pattern/constraint.

millems: Improved

spfink:
Are you submitting this under the Apache 2 License?

core/sdk-core/src/test/java/software/amazon/awssdk/core/exception/SdkExceptionTest.java
```diff
@@ -0,0 +1,33 @@
+package software.amazon.awssdk.core.exception;
```
zoewangg: copyright header?

spfink: Clearly skipping checkstyle. If only we had a checklist for this...

millems: Fixed.

PR #1515, opened 2019-11-20
Implement PING health check for HTTP2 connections 

zoewangg:
## Description
Implement PING health check for HTTP2 connections that encounter Read/WrtieTimeoutException

## Testing
Added functional tests, unit tests.
~~will run~~ integ tests, stability tests and long-running canaries tests passed

## Types of changes

## Checklist

## License

### DIFF
--- a/.changes/next-release/feature-NettyNIOHTTPClient-dcedcab.json
+++ b/.changes/next-release/feature-NettyNIOHTTPClient-dcedcab.json
@@ -1,0 +1,5 @@
+{
+    "category": "Netty NIO HTTP Client", 
+    "type": "feature", 
+    "description": "Detect unhealthy http2 connections when read or write times out by sending PING frames"
+}
--- a/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/Http2FrameExceptionHandler.java
+++ b/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/Http2FrameExceptionHandler.java
@@ -1,0 +1,113 @@
+/*
+ * Copyright 2010-2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.http.nio.netty.internal.http2;
+
+import static software.amazon.awssdk.http.nio.netty.internal.ChannelAttributeKey.CHANNEL_POOL_RECORD;
+import static software.amazon.awssdk.http.nio.netty.internal.ChannelAttributeKey.PING_TRACKER;
+import static software.amazon.awssdk.http.nio.netty.internal.NettyConfiguration.HTTP2_CONNECTION_PING_TIMEOUT_SECONDS;
+import static software.amazon.awssdk.http.nio.netty.internal.utils.NettyUtils.doInEventLoop;
+
+import io.netty.channel.Channel;
+import io.netty.channel.ChannelHandler;
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.ChannelInboundHandlerAdapter;
+import io.netty.handler.codec.http2.DefaultHttp2PingFrame;
+import io.netty.handler.codec.http2.Http2PingFrame;
+import io.netty.handler.timeout.ReadTimeoutException;
+import io.netty.handler.timeout.WriteTimeoutException;
+import io.netty.util.concurrent.ScheduledFuture;
+import java.io.IOException;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
+import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.utils.Logger;
+
+@ChannelHandler.Sharable
+@SdkInternalApi
+public final class Http2FrameExceptionHandler extends ChannelInboundHandlerAdapter {
+    private static final Http2FrameExceptionHandler INSTANCE = new Http2FrameExceptionHandler();
+    private static final Http2PingFrame DEFAULT_PING_FRAME = new DefaultHttp2PingFrame(0);
+    private static final Logger log = Logger.loggerFor(Http2FrameExceptionHandler.class);
+
+    private Http2FrameExceptionHandler() {
+    }
+
+    public static Http2FrameExceptionHandler create() {
+        return INSTANCE;
+    }
+
+    @Override
+    public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) {
+        if ((cause instanceof ReadTimeoutException || cause instanceof WriteTimeoutException) && ctx.channel().parent() != null) {
+            Channel parent = ctx.channel().parent();
+            doInEventLoop(parent.eventLoop(), () -> healthCheckIfNotAlready(parent));
+        }
+
+        ctx.fireExceptionCaught(cause);
+    }
+
+    private void healthCheckIfNotAlready(Channel parent) {
+
+        // No need to do another health check if there is a PING in flight
+        if (parent.attr(PING_TRACKER).get() != null) {
+            return;
+        }
+
+        log.debug(() -> "Read/WriteTimeoutException occurred, sending PING frame to " + parent);
+        Supplier<ScheduledFuture<?>> pingTimerFutureSupplier =  () ->
+            parent.eventLoop().schedule(() -> timeoutHealthCheck(parent),
+                                        HTTP2_CONNECTION_PING_TIMEOUT_SECONDS,
+                                        TimeUnit.SECONDS);
+
+        PingTracker pingTracker = new PingTracker(pingTimerFutureSupplier);
+        parent.attr(PING_TRACKER).set(pingTracker);
+
+        parent.writeAndFlush(DEFAULT_PING_FRAME).addListener(res -> {
+            if (!res.isSuccess()) {
+                log.debug(() -> "Failed to write and flush PING frame to connection", res.cause());
+
+                pingTracker.cancel();
+                closeH2Connection(parent, PingFailedException.PING_WRITE_FAILED_INSTANCE);
+            }
+        });
+        pingTracker.start();
+    }
+
+    private void timeoutHealthCheck(Channel parentChannel) {
+        log.debug(() -> String.format("Has not received PING ACK within %s seconds, closing the connection %s",
+                                      HTTP2_CONNECTION_PING_TIMEOUT_SECONDS, parentChannel));
+        closeH2Connection(parentChannel, PingFailedException.PING_NOT_ACK_INSTANCE);
+    }
+
+    private void closeH2Connection(Channel parent, PingFailedException exception) {
+        MultiplexedChannelRecord channelRecord = parent.attr(CHANNEL_POOL_RECORD).get();
+        channelRecord.shutdownChildChannels(exception);
+        parent.close();
+    }
+
+    static final class PingFailedException extends IOException {
+        static final PingFailedException PING_WRITE_FAILED_INSTANCE =
+            new PingFailedException("Failed to send PING to the service");
+
+        static final PingFailedException PING_NOT_ACK_INSTANCE =
+            new PingFailedException(String.format("Failed to receive PING ACK from the service within %s seconds",
+                                                  HTTP2_CONNECTION_PING_TIMEOUT_SECONDS));
+
+        private PingFailedException(String msg) {
+            super(msg);
+        }
+    }
+}
--- a/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/Http2PingFrameHandler.java
+++ b/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/Http2PingFrameHandler.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright 2010-2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.http.nio.netty.internal.http2;
+
+import static software.amazon.awssdk.http.nio.netty.internal.ChannelAttributeKey.PING_TRACKER;
+
+import io.netty.channel.Channel;
+import io.netty.channel.ChannelHandler;
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.SimpleChannelInboundHandler;
+import io.netty.handler.codec.http2.Http2PingFrame;
+import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.http.nio.netty.internal.utils.ChannelUtils;
+import software.amazon.awssdk.utils.Logger;
+
+@SdkInternalApi
+@ChannelHandler.Sharable
+public final class Http2PingFrameHandler extends SimpleChannelInboundHandler<Http2PingFrame> {
+    private static final Logger log = Logger.loggerFor(Http2PingFrameHandler.class);
+    private static final Http2PingFrameHandler INSTANCE = new Http2PingFrameHandler();
+
+    private Http2PingFrameHandler() {
+    }
+
+    public static Http2PingFrameHandler create() {
+        return INSTANCE;
+    }
+
+    @Override
+    protected void channelRead0(ChannelHandlerContext ctx, Http2PingFrame frame) {
+        if (frame.ack()) {
+            Channel channel = ctx.channel();
+            log.debug(() -> "Received PING ACK from channel " + channel);
+            ChannelUtils.getAttribute(channel, PING_TRACKER).ifPresent(tracker -> {
+                tracker.cancel();
+                channel.attr(PING_TRACKER).set(null);
+            });
+        } else {
+            ctx.fireChannelRead(frame);
+        }
+    }
+
+}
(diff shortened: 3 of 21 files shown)


http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/Http2PingFrameHandler.java
```diff
@@ -0,0 +1,52 @@
...
+            ctx.channel().attr(KEEP_ALIVE).set(true);
+        } else {
+            ctx.fireChannelRead(frame);
+        }
+    }
+
+    private static final class SingletonHolder {
+        private static final Http2PingFrameHandler INSTANCE = new Http2PingFrameHandler();
+    }
+}
```
dagnir: Why isn't `INSTANCE` just a member of `Http2PingFrameHandler`?

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/Http2PingFrameHandler.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    public static Http2PingFrameHandler create() {
+        return SingletonHolder.INSTANCE;
+    }
+
+    @Override
+    protected void channelRead0(ChannelHandlerContext ctx, Http2PingFrame frame) {
+        if (frame.ack()) {
+            log.debug(() -> "Received PING ACK from channel " + ctx.channel());
```
dagnir: I think logging the remote address+port should be more helpful

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/Http2ToHttpInboundAdapter.java
```diff
@@ -56,6 +63,30 @@ protected void channelRead0(ChannelHandlerContext ctx, Http2Frame frame) throws
...
+    }
+
+    private void healthCheck(ChannelHandlerContext ctx) {
+        Channel parent = ctx.channel().parent();
+        log.debug(() -> "ReadTimeoutException occurred, sending PING frame to " + parent);
+        parent.attr(KEEP_ALIVE).set(false);
+
+        parent.writeAndFlush(PingFrame.DEFAULT_PING_FRAME).addListener(res -> {
+            if (!res.isSuccess()) {
+                log.debug(() -> "Failed to write and flush PING frame to connection ", res.cause());
```
dagnir: nit: extra space

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/Http2ToHttpInboundAdapter.java
```diff
@@ -56,6 +63,30 @@ protected void channelRead0(ChannelHandlerContext ctx, Http2Frame frame) throws
...
+            healthCheck(ctx);
+        }
+
+        ctx.fireExceptionCaught(cause);
+    }
+
+    private void healthCheck(ChannelHandlerContext ctx) {
+        Channel parent = ctx.channel().parent();
+        log.debug(() -> "ReadTimeoutException occurred, sending PING frame to " + parent);
+        parent.attr(KEEP_ALIVE).set(false);
```
dagnir: Is there a race between this and getting a stream from the `MultiplexChannelRecord#acquire`? For example, is it possible for `acquire()` to think the channel is still live when it might not be?

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/Http2ToHttpInboundAdapter.java
```diff
@@ -56,6 +63,30 @@ protected void channelRead0(ChannelHandlerContext ctx, Http2Frame frame) throws
...
+    @Override
+    public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) {
+        if ((cause instanceof ReadTimeoutException || cause instanceof WriteTimeoutException) && ctx.channel().parent() != null) {
+            healthCheck(ctx);
+        }
+
+        ctx.fireExceptionCaught(cause);
+    }
+
+    private void healthCheck(ChannelHandlerContext ctx) {
```
dagnir: This is just a thought, but it might make it cleaner to do the health checking and management of the keep alive flag as a separate handler that gets added/removed from the parent pipeline as necessary

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/Http2ToHttpInboundAdapter.java
zoewangg: Yes, it's possible another request comes in right before `KEEP_ALIVE` is set to false and picks up this channel, but since we are shutting down the child channel very soon, that request should fail fast.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/Http2PingFrameHandler.java
zoewangg: `ctx.channel().toString` actually contains that info.
example log:
```
Received PING ACK from channel [id: 0xfd7b29f9, L:/127.0.0.1:49571 - R:localhost/127.0.0.1:49570]
```

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/Http2PingFrameHandler.java
zoewangg: Lazy initialization coz h1 doesn't need this handler, but initialization of this class is not expensive at all though, so this is not necessary I guess.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/Http2ToHttpInboundAdapter.java
dagnir: Wouldn't it be possible that it can take up to the read timeout in some cases?

zoewangg: Oh actually yes, I was only thinking for the case where it would fail to write the PING frame and then would get shut down. I'll think it a bit more to how to better address it.

(I forgot to mention that we will reduce the read timeout for certain service clients. I'll make that change.)

.changes/next-release/feature-NettyNIOHTTPClient-dcedcab.json
```diff
@@ -0,0 +1,5 @@
+{
+    "category": "Netty NIO HTTP Client", 
+    "type": "feature", 
+    "description": "Detect unhealthy http2 connections when read or write times out by sending PING frames"
```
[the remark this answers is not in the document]
zoewangg: Yes, verified it.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/Http2ToHttpInboundAdapter.java
zoewangg: @dagnir Are you saying moving the keep alive flag management into a separate parent handler?

The exception here happens on a child channel and we have to set `KEEP_ALIVE  = false` for the parent channel in this child channel handler.

I can move this healthcheck logic into a separate handler, say `Http2FrameExceptionHandler`.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/Http2ToHttpInboundAdapter.java
zoewangg: We can add another keep alive flag check for parent channel in `HealthCheckedChannelPool` and that should reduce the chances. 

`channel.parent() != null && !channel.parent().attr(KEEP_ALIVE).get()`. 

What do you think?

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/Http2ToHttpInboundAdapter.java
dagnir: Yes, basically what I had in mind was that if we determine that we need to do a health check, then instead of taking care of that in this handler, add a new handler to the pipeline before this one that takes care of the health checking like constructing/writing the frame, setting the flag, checking the ACK etc.

zoewangg: I see. But we only know we need to do health check when there's a predefined exception thrown from a child channel and even though we create a parent handler for that, I'm not sure how we can trigger the parent handler to do the health check from the child handler.

I guess we can create a specific UserEvent to trigger the parent channel to do health check, but I feel like that's overly complicated.

zoewangg: Discussed offline. To sum up, we can add a parent handler to handle all PING frames when this happens, something like  `ctx.channel().parent().pipeline().addLast(Http2PingFrameHandler.create()); ` 
and use `handlerAdded` callback to send PING frame.

There is a concern that adding a handler can add delays and we need to run perf tests to be sure. For now, we are just going to send the PING frame from a separate child handler.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/Http2FrameExceptionHandler.java
```diff
@@ -0,0 +1,101 @@
...
+                log.debug(() -> "Failed to write and flush PING frame to connection", res.cause());
+                closeH2Connection(parent);
+            } else {
+                ctx.executor().schedule(() -> timeoutHealthCheck(parent), HTTP2_CONNECTION_PING_TIMEOUT_SECONDS,
+                                        TimeUnit.SECONDS);
+            }
+        });
+    }
+
+    private void timeoutHealthCheck(Channel parentChannel) {
```
dagnir: AFAICT this runs from the child channel's event loop, whereas the `ACK_RECEIVED` gets set from the parent's loop. Should we be doing the check from the same loop? (probably the parent's).

zoewangg: Yeah, updated

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/Http2FrameExceptionHandler.java
```diff
@@ -0,0 +1,100 @@
...
+    private void healthCheckIfNotAlready(ChannelHandlerContext ctx) {
+        Channel parent = ctx.channel().parent();
+
+        // No need to do another health check if there is a PING in flight
+        if (PingStatus.IN_FLIGHT.equals(parent.attr(PING_STATUS).get())) {
+            return;
+        }
+
+        log.debug(() -> "Read/WriteTimeoutException occurred, sending PING frame to " + parent);
+        parent.attr(PING_STATUS).set(PingStatus.IN_FLIGHT);
```
millems: Will this always be on the same event loop?

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/Http2FrameExceptionHandler.java
```diff
@@ -0,0 +1,100 @@
...
+
+        parent.writeAndFlush(DEFAULT_PING_FRAME).addListener(res -> {
+            if (!res.isSuccess()) {
+                log.debug(() -> "Failed to write and flush PING frame to connection", res.cause());
+                closeH2Connection(parent);
+            }
+        });
+        parent.eventLoop().schedule(() -> timeoutHealthCheck(parent), HTTP2_CONNECTION_PING_TIMEOUT_SECONDS,
+                                TimeUnit.SECONDS);
```
millems: Should this only be done when the send-ping succeeds?

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/Http2FrameExceptionHandler.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    private void closeH2Connection(Channel parent) {
+        MultiplexedChannelRecord channelRecord = parent.attr(CHANNEL_POOL_RECORD).get();
+        channelRecord.shutdownChildChannels(Http2FrameExceptionHandler.PingFailedException.INSTANCE);
+        parent.close();
+    }
+
+    static final class PingFailedException extends IOException {
+        static final PingFailedException INSTANCE = new PingFailedException("Failed to send PING to the service");
```
millems: Should our errors distinguish between failing to send, and failing to get a response?

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/Http2PingFrameHandler.java
```diff
@@ -0,0 +1,50 @@
...
+
+    public static Http2PingFrameHandler create() {
+        return INSTANCE;
+    }
+
+    @Override
+    protected void channelRead0(ChannelHandlerContext ctx, Http2PingFrame frame) {
+        if (frame.ack()) {
+            log.debug(() -> "Received PING ACK from channel " + ctx.channel());
+            ctx.channel().attr(PING_STATUS).set(PingStatus.ACK_RECEIVED);
```
millems: What happens if:
[0:00] We get a ReadTimeout
[0:00] We send a PING request (state = IN_FLIGHT)
[0:00] We set a timer to check for the PING response after 5 seconds
[0:01] We get the PING response (state = ACK_RECEIVED)
[0:04] We get a ReadTimeout
[0:04] We send a PING request (state = IN_FLIGHT)
[0:05] The timer triggers and sees that the PING is IN_FLIGHT

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/MultiplexedChannelRecord.java
```diff
@@ -175,12 +176,20 @@ void release(Channel channel) {
         childChannels.remove(channel.id());
     }
 
-    public Future<Channel> getConnectionFuture() {
-        return connectionFuture;
+    boolean reusable() {
+        return !isPingInflight() && availableStreams.get() > 0 && !goAway;
```
millems: Why skip using it while a ping is in-flight?

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/Http2FrameExceptionHandler.java
zoewangg: It was actually done within the listener in the previous revision and I moved it out because write itself might take a long time as well. 

I tried adding a Write timeout handler there and it did not work somehow.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/MultiplexedChannelRecord.java
zoewangg: Because in the worst case, ping can take another 5 seconds to fail and the new requests picking up the same channel would end up getting "stuck" for 5 seconds. Establishing a new connection can prevent it from happening.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/Http2FrameExceptionHandler.java
zoewangg: No, it could be on different event loops depending on which event loop the child channel is running on.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/Http2FrameExceptionHandler.java
zoewangg: Sure

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/Http2PingFrameHandler.java
zoewangg: Good call.. we need to cancel the timer once it's received

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/Http2FrameExceptionHandler.java
```diff
@@ -0,0 +1,78 @@
...
+            healthCheck(ctx);
+        }
+
+        ctx.fireExceptionCaught(cause);
+    }
+
+    private void healthCheck(ChannelHandlerContext ctx) {
+        Channel parent = ctx.channel().parent();
+        log.debug(() -> "Read/WriteTimeoutException occurred, sending PING frame to " + parent);
+        parent.attr(KEEP_ALIVE).set(false);
```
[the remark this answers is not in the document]
zoewangg: What do you mean? That shouldn't be an issue.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/Http2FrameExceptionHandler.java
```diff
@@ -0,0 +1,115 @@
...
+            doInEventLoop(parent.eventLoop(), () -> healthCheckIfNotAlready(parent));
+        }
+
+        ctx.fireExceptionCaught(cause);
+    }
+
+    private void healthCheckIfNotAlready(Channel parent) {
+
+        // No need to do another health check if there is a PING in flight
+        if (PingStatus.IN_FLIGHT.equals(parent.attr(PING_STATUS).get())) {
```
millems: Can we consolidate PING_STATUS and PING_TIMER_FUTURE?

If we get a read timeout:
1. We schedule a "shut down after 5 seconds" task.
2. We send a ping to the service

If we get a ping response:
1. We cancel the "shut down after 5 seconds" task.

PR #1522, opened 2019-11-22
Add support for S3 us-east-1 regional

dagnir:
## Description
Add support for S3 us-east-1 regional flag

## Motivation and Context
New feature.

## Testing
Unit tests, integration tests, stability tests.

## Types of changes

## Checklist

## License


dagnir:
Will push a fix for that Travis failure after the integ test run completes.

codegen-lite/src/main/java/software/amazon/awssdk/codegen/lite/regions/ServiceMetadataProviderGenerator.java
```diff
@@ -101,6 +101,10 @@ private CodeBlock regions(Partitions partitions) {
     }
 
     private ClassName serviceMetadataClass(String service) {
+        if ("s3".equals(service)) {
```
spfink: Might have been better to plumb this through as a customization but we can always do that later.

core/regions/src/main/java/software/amazon/awssdk/regions/servicemetadata/EnhancedS3ServiceMetadata.java
```diff
@@ -0,0 +1,100 @@
...
+            synchronized (this) {
+                if (!profileSettingChecked) {
+                    try {
+                        String profileName = ProfileFileSystemSetting.AWS_PROFILE.getStringValueOrThrow();
+                        ProfileFile profileFile = ProfileFile.defaultProfileFile();
+                        profileSetting = profileFile.profile(profileName)
+                                .flatMap(p -> p.property(ProfileProperty.S3_US_EAST_1_REGIONAL_ENDPOINT))
+                                .orElse(null);
+                    } catch (Throwable t) {
+                        log.warn(() -> "Unable to load config file", t);
```
zoewangg: Should we use `debug` instead? Would this be logged for every request?

dagnir: No, it's only checked once https://github.com/aws/aws-sdk-java-v2/blob/442ec96fe12b5398a034bd693760c4fe853da8e6/core/regions/src/main/java/software/amazon/awssdk/regions/servicemetadata/EnhancedS3ServiceMetadata.java#L95

dagnir: It wouldn't log on each request AFAIK since the generated `GeneratedServiceMetadataProvider` class that instantiates this is a is a singleton instance.

zoewangg: My concern is that customers not using the profiles will get concerned when they see the warn message. I recall we've had customers reporting similar issue with other profile config before.

dagnir: There shouldn't be an exception if they don't have any profile files:

https://github.com/aws/aws-sdk-java-v2/blob/4e4bebd8e17ba4651b5d0ee6d667a80aa5fc2601/core/profiles/src/main/java/software/amazon/awssdk/profiles/ProfileFile.java#L134-L148

The created `ProfileFile` will essentially be empty.

zoewangg: Gotcha, then we're good

PR #1539, opened 2019-12-05
Reap unhealthy and idle HTTP/2 connections.

millems:
1. Periodically PING HTTP/2 connections and close them if the service doesn't respond within a certain period of time.
2. Close HTTP/2 connections that have had no streams checked out for a certain period of time.


http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/Http2ChannelHealthChecker.java
```diff
@@ -0,0 +1,118 @@
...
+
+    Http2ChannelHealthChecker(Channel channel, Duration pingTimeout) {
+        this.channel = channel;
+        this.pingTimeoutMillis = pingTimeout.toMillis();
+    }
+
+    public void start() {
+        warnIfNotInEventLoop(channel.eventLoop());
+
+        channel.pipeline().addLast(new PingResponseListener());
```
dagnir: Does it make sense to guard against adding more than 1?

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/Http2ChannelHealthChecker.java
```diff
@@ -0,0 +1,118 @@
...
+import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.utils.Logger;
+
+/**
+ * Used by {@link Http2PingedChannelPool} to periodically check the health of HTTP2 connections via PING frames.
+ *
+ * If a channel is found to be unhealthy, this will invoke {@link ChannelPipeline#fireExceptionCaught(Throwable)}.
+ */
+@SdkInternalApi
+class Http2ChannelHealthChecker {
```
dagnir: Hmm did you consider making this a channel handler? It seems like that might be easier as an implementation since you don't need to really worry about being out of the loop.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/Http2ChannelHealthChecker.java
millems: +1

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/Http2ChannelHealthChecker.java
millems: I'll try it and see how pretty it is!

millems: I like it. Updated!

PR #1540, opened 2019-12-07
Add list/map existence check (#867)

### LINKED ISSUE #867: Modeled list / map existence check
See https://github.com/aws/aws-sdk-java-v2/issues/865

V2 currently populates empty maps/lists if the services does not return anything. We should provide a way to check if the service actually returned something (other than `instanceof AutoConstructList`)

jeffalder:
Add an existence check method for Lists and Maps. 

## Description
Adds a getter method, `hasX` for every member that is a list or map type. Adds a NamingStrategy for this type of method.

## Motivation and Context
Fixes #867 

There is no single way for SDK consumers to identify if a collection might be an empty list or null without exposing the AutoConstruct types. This fix provides the existence check.

## Testing
protocol tests caught an issue in the naming.
other fixes were caught by the parameterized tests.

## Types of changes

## Checklist

## License

### DIFF
--- a/codegen/src/main/java/software/amazon/awssdk/codegen/internal/DocumentationUtils.java
+++ b/codegen/src/main/java/software/amazon/awssdk/codegen/internal/DocumentationUtils.java
@@ -36,6 +36,11 @@ public final class DocumentationUtils {
 
     private static final String DEFAULT_GETTER_PARAM = "The value of the %s property for this object.";
 
+    private static final String DEFAULT_EXISTENCE_CHECK = "Returns true if the %s property was specified by the sender "
+                                                          + "(it may be empty), or false if the sender did not specify "
+                                                          + "the value (it will be empty). For responses returned by the SDK, "
+                                                          + "the sender is the AWS service.";
+
     private static final String DEFAULT_FLUENT_RETURN =
             "Returns a reference to this object so that method calls can be chained together.";
 
@@ -163,5 +168,9 @@ public final class DocumentationUtils {
 
     public static String defaultFluentReturn() {
         return DEFAULT_FLUENT_RETURN;
+    }
+
+    public static String defaultExistenceCheck() {
+        return DEFAULT_EXISTENCE_CHECK;
     }
 }
--- a/codegen/src/main/java/software/amazon/awssdk/codegen/model/intermediate/MemberModel.java
+++ b/codegen/src/main/java/software/amazon/awssdk/codegen/model/intermediate/MemberModel.java
@@ -16,6 +16,7 @@
 package software.amazon.awssdk.codegen.model.intermediate;
 
 import static software.amazon.awssdk.codegen.internal.Constant.LF;
+import static software.amazon.awssdk.codegen.internal.DocumentationUtils.defaultExistenceCheck;
 import static software.amazon.awssdk.codegen.internal.DocumentationUtils.defaultFluentReturn;
 import static software.amazon.awssdk.codegen.internal.DocumentationUtils.defaultGetter;
 import static software.amazon.awssdk.codegen.internal.DocumentationUtils.defaultGetterParam;
@@ -24,11 +25,15 @@ import static software.amazon.awssdk.codegen.internal.DocumentationUtils.default
 import static software.amazon.awssdk.codegen.internal.DocumentationUtils.stripHtmlTags;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.squareup.javapoet.ClassName;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import software.amazon.awssdk.codegen.internal.TypeUtils;
 import software.amazon.awssdk.core.SdkBytes;
 import software.amazon.awssdk.core.protocol.MarshallingType;
+import software.amazon.awssdk.core.util.SdkAutoConstructList;
+import software.amazon.awssdk.core.util.SdkAutoConstructMap;
 import software.amazon.awssdk.protocols.core.PathMarshaller;
 import software.amazon.awssdk.utils.StringUtils;
 
@@ -69,6 +74,8 @@ public class MemberModel extends DocumentationModel {
     private String fluentSetterMethodName;
 
     private String fluentEnumSetterMethodName;
+
+    private String existenceCheckMethodName;
 
     private String beanStyleGetterName;
 
@@ -247,6 +254,19 @@ public class MemberModel extends DocumentationModel {
         return this;
     }
 
+    public String getExistenceCheckMethodName() {
+        return existenceCheckMethodName;
+    }
+
+    public void setExistenceCheckMethodName(String existenceCheckMethodName) {
+        this.existenceCheckMethodName = existenceCheckMethodName;
+    }
+
+    public MemberModel withExistenceCheckMethodName(String existenceCheckMethodName) {
+        setExistenceCheckMethodName(existenceCheckMethodName);
+        return this;
+    }
+
     public ReturnTypeModel getGetterModel() {
         return getterModel;
     }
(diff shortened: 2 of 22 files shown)


jeffalder:
For reference, here is the decompiled `AttributeValue` with these changes.

```
 public String s() {
        return this.s;
    }

    public String n() {
        return this.n;
    }

    public SdkBytes b() {
        return this.b;
    }

    public boolean hasSS() {
        return this.ss != null && !(this.ss instanceof SdkAutoConstructList);
    }

    public List<String> ss() {
        return this.ss;
    }

    public boolean hasNS() {
        return this.ns != null && !(this.ns instanceof SdkAutoConstructList);
    }

    public List<String> ns() {
        return this.ns;
    }

    public boolean hasBS() {
        return this.bs != null && !(this.bs instanceof SdkAutoConstructList);
    }

    public List<SdkBytes> bs() {
        return this.bs;
    }

    public boolean hasM() {
        return this.m != null && !(this.m instanceof SdkAutoConstructMap);
    }

    public Map<String, AttributeValue> m() {
        return this.m;
    }

    public boolean hasL() {
        return this.l != null && !(this.l instanceof SdkAutoConstructList);
    }

    public List<AttributeValue> l() {
        return this.l;
    }
```

codegen/src/main/java/software/amazon/awssdk/codegen/internal/DocumentationUtils.java
```diff
@@ -36,6 +36,8 @@
 
     private static final String DEFAULT_GETTER_PARAM = "The value of the %s property for this object.";
 
+    private static final String DEFAULT_EXISTENCE_CHECK = "Returns true if the %s property has a meaningful value.";
```
millems: Hah, this might be a little too honest. How about something like "Returns true if the %s property was specified by the sender (it may be empty), or false if the sender did not specify the value (it will be empty). For responses returned by the SDK, the sender is the AWS service."

codegen/src/main/java/software/amazon/awssdk/codegen/poet/model/AwsServiceModel.java
```diff
@@ -381,14 +383,28 @@ private MethodSpec getValueForField() {
...
+        if (autoConstructClass != null) {
+            result.add(existenceCheckGetter(member, autoConstructClass));
+        }
+
         result.add(memberGetter(member));
 
         return result.stream();
     }
 
+    private ClassName getAutoConstructClassIfExists(MemberModel member) {
```
millems: Nit: `Optional<ClassName>`?

codegen/src/main/java/software/amazon/awssdk/codegen/naming/DefaultNamingStrategy.java
```diff
@@ -292,6 +292,11 @@ public String getFluentEnumGetterMethodName(String memberName, Shape parentShape
         return getterMethodName;
     }
 
+    @Override
+    public String getExistenceCheckMethodName(String memberName) {
+        return String.format("has%s", Utils.capitalize(memberName));
```
millems: Can we write a test for a reserved member name? What I mean is a list and map named "build" or something else reserved. I think to be consistent with the setter/getter we'll need to use `rewriteInvalidMemberName`.

millems:
Can we add a changelog entry for this?

The best way is via the python script at `scripts/new-change`.

codegen/src/main/java/software/amazon/awssdk/codegen/internal/DocumentationUtils.java
jeffalder: Perfect, will change.

codegen/src/main/java/software/amazon/awssdk/codegen/poet/model/AwsServiceModel.java
jeffalder: Will do; I wasn't aware of the convention you were using.

codegen/src/main/java/software/amazon/awssdk/codegen/internal/DocumentationUtils.java
millems: Thanks! We might also add a note to the getter that the caller can check the "has" method to see whether a value was sent in this field.

codegen/src/main/java/software/amazon/awssdk/codegen/naming/DefaultNamingStrategy.java
jeffalder: Sure, can do.

jeffalder:
@millems Thanks, I'll get to these tonight.

codegen/src/main/java/software/amazon/awssdk/codegen/naming/DefaultNamingStrategy.java
jeffalder:
```
    "ExistenceCheckNamingResponse": {
      "type": "structure",
      "members": {
        "Build":{"shape": "ListOfStrings"},
```
yields:
```
    public interface Builder extends JsonProtocolTestsResponse.Builder, SdkPojo,
                                     CopyableBuilder<Builder, ExistenceCheckNamingResponse> {
        /**
         * Sets the value of the Build property for this object.
         *
         * @param build
         *        The new value for the Build property for this object.
         */
        Builder build(Collection<String> build);

        /**
         * Sets the value of the Build property for this object.
         *
         * @param build
         *        The new value for the Build property for this object.
         */
        Builder build(String... build);
```
😬 
Seems like those should render to `buildValue`. But I'm not going to fix this in this PR.

codegen/src/main/java/software/amazon/awssdk/codegen/poet/model/AwsServiceModel.java
jeffalder: Fixed in 1d86076

codegen/src/main/java/software/amazon/awssdk/codegen/internal/DocumentationUtils.java
jeffalder: Both documentation changes made in 48b3879

codegen/src/main/java/software/amazon/awssdk/codegen/naming/DefaultNamingStrategy.java
jeffalder: Fixed in 50de534

jeffalder:
Changefile added in b169225

codegen/src/main/java/software/amazon/awssdk/codegen/model/intermediate/MemberModel.java
```diff
@@ -429,6 +455,14 @@ public String getFluentSetterDocumentation() {
                + getEnumDoc();
     }
 
+    public String getExistenceCheckDocumentation() {
+        StringBuilder docBuilder = new StringBuilder();
+        docBuilder.append(StringUtils.isNotBlank(documentation) ? documentation : defaultExistenceCheck().replace("%s", name))
+                .append(LF);
+
+        return docBuilder.toString();
```
millems: Should we just use the default existence check text for all methods? I don't think it's particularly useful to have the documentation for the field on the has* method.

Current:
```Java
    /**
     * <p>
     * An attribute of type Binary Set. For example:
     * </p>
     * <p>
     * <code>"BS": ["U3Vubnk=", "UmFpbnk=", "U25vd3k="]</code>
     * </p>
     */
    public boolean hasBs() {
        return bs != null && !(bs instanceof SdkAutoConstructList);
    }
```

jeffalder: 🤦‍♂  Yes; will change.

jeffalder: Fixed:
```
    /**
     * Returns true if the BS property was specified by the sender (it may be empty), or false if the sender did not
     * specify the value (it will be empty). For responses returned by the SDK, the sender is the AWS service.
     */
    public boolean hasBs() {
        return bs != null && !(bs instanceof SdkAutoConstructList);
    }
```

jeffalder:
Fixed and squashed; should be ready to merge.

millems:
Thanks for another great contribution!

PR #1543, opened 2019-12-10
Add apache-async-client

rschmitt:
This commit adds a new `SdkAsyncHttpClient` implementation,
`ApacheAsyncHttpClient`, as well as a service provider for that
implementation. These live in a new module, apache-async-client, which
is part of the http-clients module.

This client is intended to be roughly equivalent in power to the
netty-nio-client, supporting features like HTTP/2, full-duplex HTTP
requests, and reactive streams.

The client supports the following options:

| Option                         | Default value | Defined by...       | Implemented via...                    |
| ------------------------------ | ------------- | ------------------- | ------------------------------------- |
| `socketTimeout`                | 30 sec        | http-client-spi     | `IOReactorConfig`                     |
| `connectionTimeout`            | 2 sec         | http-client-spi     | `RequestConfig`                       |
| `connectionAcquisitionTimeout` | 20 sec        | http-client-spi     | `RequestConfig`                       |
| `maxConnections`               | 50            | http-client-spi     | `PoolingAsyncClientConnectionManager` |
| `proxyConfiguration`           | empty object  | apache-async-client | `HttpAsyncClient`                     |
| `localAddress`                 | `null`        | apache-async-client | `HttpAsyncClient`                     |
| `expectContinueEnabled`        | `true`        | apache-async-client | `RequestConfig`                       |
| `connectionTimeToLive`         | unlimited     | http-client-spi     | `PoolingAsyncClientConnectionManager` |
| `connectionMaxIdleTime`        | 60 sec        | http-client-spi     | `HttpAsyncClient`                     |
| `protocol`                     | `null`        | http-client-spi     | `HttpVersionPolicy`                   |
| `maxHttp2Streams`              | `null`        | apache-async-client | `H2Config`                            |
| `httpRoutePlanner`             | `null`        | apache-async-client | `HttpAsyncClient`                     |
| `credentialsProvider`          | `null`        | apache-async-client | `HttpAsyncClient`                     |
| `tlsKeyManagersProvider`       | JVM defaults  | http-client-spi     | `SSLContext`                          |

The following standard options are not supported:

* `SdkHttpConfigurationOption.WRITE_TIMEOUT`: `socketTimeout` maps to
  `READ_TIMEOUT`, which controls both read and write behavior on every
  client except netty-nio-client
* `SdkHttpConfigurationOption.MAX_PENDING_CONNECTION_ACQUIRES`: Apache
  supports an acquisition timeout, not a limit on the number of pending
  connection acquires


http-clients/apache-async-client/src/main/java/software/amazon/awssdk/http/apache/async/ApacheAsyncRequestFactory.java
```diff
@@ -0,0 +1,117 @@
...
+/**
+ * Responsible for creating Apache HttpClient 5 request objects.
+ */
+@SdkInternalApi
+class ApacheAsyncRequestFactory {
+    private static final List<String> IGNORE_HEADERS = Collections.singletonList(HttpHeaders.HOST);
+
+    public AsyncRequestProducer create(AsyncExecuteRequest request, boolean isHttp2) {
+        URI uri = request.request().getUri();
+        AsyncRequestBuilder base = createApacheRequest(request, uri.toString());
```
michael-o: Never use `URI#toString()` it will not properly percent-encode parts of the URI. Use `toASCIIString()` only.

http-clients/apache-async-client/src/main/java/software/amazon/awssdk/http/apache/async/ApacheAsyncRequestFactory.java
```diff
@@ -0,0 +1,117 @@
...
+            case OPTIONS:
+                return AsyncRequestBuilder.options(uri);
+            case PATCH:
+                return AsyncRequestBuilder.patch(uri).setEntity(wrapEntity(request));
+            case POST:
+                return AsyncRequestBuilder.post(uri).setEntity(wrapEntity(request));
+            case PUT:
+                return AsyncRequestBuilder.put(uri).setEntity(wrapEntity(request));
+            default:
+                throw new RuntimeException("Unknown HTTP method name: " + request.request().method());
```
michael-o: Why not IllegalArgument or IllegalState?

pom.xml
```diff
@@ -109,6 +109,8 @@
         <rxjava.version>2.1.9</rxjava.version>
         <commons-codec.verion>1.10</commons-codec.verion>
         <jmh.version>1.21</jmh.version>
+        <httpcore5.version>5.0-beta10</httpcore5.version>
+        <httpclient5.version>5.0-beta7-SNAPSHOT</httpclient5.version>
```
michael-o: One should not rely on snapshots. There is no guarantee that someone can really pull them.

http-clients/apache-async-client/src/main/java/software/amazon/awssdk/http/apache/async/ApacheAsyncRequestFactory.java
rschmitt: I inherited this choice of exception from `ApacheHttpRequestFactory`, part of the `apache-client` (synchronous) integration that I forked. I'm generally not too picky about exception types, so I didn't change it.

http-clients/apache-async-client/src/main/java/software/amazon/awssdk/http/apache/async/ApacheAsyncRequestFactory.java
rschmitt: Great catch, thank you.

pom.xml
rschmitt: There's a show-stopping bug in beta6, so I had to use a local snapshot temporarily.

http-clients/apache-async-client/src/main/java/software/amazon/awssdk/http/apache/async/ApacheAsyncRequestFactory.java
rschmitt: Come to think of it, it would probably be a better idea to just remain in the `URI` domain, because otherwise Apache is just going to run the String through `URI#create` again.

http-clients/apache-async-client/src/main/java/software/amazon/awssdk/http/apache/async/ProxyConfiguration.java
```diff
@@ -0,0 +1,404 @@
...
+    /**
+     * Returns the Java system property for nonProxyHosts as set of Strings.
+     * See http://docs.oracle.com/javase/7/docs/api/java/net/doc-files/net-properties.html.
+     */
+    private Set<String> parseNonProxyHostsProperty() {
+        String nonProxyHosts = ProxySystemSetting.NON_PROXY_HOSTS.getStringValue().orElse(null);
+
+        if (!StringUtils.isEmpty(nonProxyHosts)) {
+            return Arrays.stream(nonProxyHosts.split("\\|"))
+                         .map(String::toLowerCase)
```
michael-o: The  `toLowerCase()` is locale-specific and might introduce subtile bugs. Use `ROOT` or `ENGLISH`.

pom.xml
ok2c: @rschmitt @michael-o We will get another BETA (the last one) or GA soon.

rschmitt:
I've updated this PR to depend on the latest beta. It no longer requires snapshot builds and is in pretty good shape to merge to a feature branch or something.

dagnir:
Hey Ryan, I'm going to go ahead and close this as the branch has gone stale. I think we as a team should also discuss how want to handle new HTTP client implementations going forward.

Issue #1313, opened 2019-06-28
Simpler S3 CopyObjectRequest#copySource


worpet:
In the original SDK, the source bucket and key vs the destination bucket and key was clear:
```
new CopyObjectRequest(sourceBucket, sourceKey, destinationBucket, destinationKey)
```

But in SDK 2, the way the `Builder` methods are named makes this overly confusing.

## Expected Behavior
`CopyObjectRequest.Builder` should have methods like `sourceBucket(bucket)` and `sourceKey(key)` to clearly set the source bucket and source key.

The `bucket(bucket)` and `key(key)` methods should be documented as being for the **destination** values.

## Current Behavior
`CopyObjectRequest.Builder` has only the method `copySource(String copySource)` which accepts the source bucket and source key, separated by a slash. More importantly, you are **required to URL-encode** this value, which is easy to overlook. For example, even the [AWS documentation example](https://github.com/awsdocs/aws-doc-sdk-examples/blob/master/javav2/example_code/s3/src/main/java/com/example/s3/CopyObject.java) makes the mistake of not encoding. A casual test will appear to work, but may fail later when you happen to have a key that contains an unsafe URL character.

The `bucket(bucket)` and `key(key)` methods are not documented as being for the **destination** values so, when coding, there is no clear way to simply set a copy source and destination without searching for more information.

millems:
I agree the naming is less clear. This change is due to the migration from the hand-written S3 client to the generated S3 client. The SDK now matches the naming and semantics that the S3 team applied to the fields, instead of what the SDK team decided to apply to the fields. 

The SDK should consider adding an S3 customization for this, because having to encode the `copySource` field is not a good experience. Having to specify the bucket and source key separated by a slash, in general, is not a good experience.

Marking as a feature request.

jeffalder:
@millems There's no clear customization route that I can see.

1. The existing `bucket` and `key` methods on the builder need to be deprecated. They can't be simply renamed, and no existing customization route provides this.
2. There's no requirement that the property be renamed everywhere; the real need is simply a renamed builder method. Although -- it would be useful to rename it consistently everywhere.

I _think_ what makes sense is to add an `emitPropertyName` modification with the _new_ name, then inject a deprecated member with the _old_ name and existing location (if possible). Later, the injected member can be removed. Is there something I'm missing? Should a new modification path be added, instead?

jeffalder:
[...]

PR #1547, opened 2019-12-12
Support ability to rename fields and deprecate existing names; does so for S3

jeffalder:
## Description
When renaming properties via customization.config, the existing names can be deprecated. 
On S3 CopyObjectRequest, deprecates `bucket` in favor of `destinationBucket` and `key` in favor of `destinationKey`.

## Motivation and Context
This covers half of the requests in #1313 .

## Testing
Added tests for deprecation
Validated resulting CopyObjectResult class

n/a

## Types of changes

## Checklist

## License

### DIFF
--- a/codegen/src/test/resources/software/amazon/awssdk/codegen/poet/model/deprecatedrenamerequest.java
+++ b/codegen/src/test/resources/software/amazon/awssdk/codegen/poet/model/deprecatedrenamerequest.java
@@ -1,0 +1,272 @@
+package software.amazon.awssdk.services.jsonprotocoltests.model;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.function.BiConsumer;
+import java.util.function.Consumer;
+import java.util.function.Function;
+import software.amazon.awssdk.annotations.Generated;
+import software.amazon.awssdk.awscore.AwsRequestOverrideConfiguration;
+import software.amazon.awssdk.core.SdkField;
+import software.amazon.awssdk.core.SdkPojo;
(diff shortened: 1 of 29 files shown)


jeffalder:
@millems Here ya go! Have at it. Thanks for your lightning-fast turnaround on feedback, too!

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/model/deprecatedrenamerequest.java
```diff
@@ -0,0 +1,264 @@
...
+        public final Builder newName(String newName) {
+            this.newName = newName;
+            return this;
+        }
+
+        @Override
+        public final Builder originalNameDeprecated(String newName) {
+            this.newName = newName;
+            return this;
+        }
```
millems: Minor Nit: Should probably have bean-based methods for the deprecated field. These are around for people using things like spring-xml, so it's likely not going to break anyone to exclude them, but it'd be nice to have for completeness.

jeffalder: I'll try to have this for you by Monday.

jeffalder:
I added the bean-style setters. Then I realized that there was a gap in the logic for lists and maps; I updated the checks to make sure that lists and maps can't be deprecated (yet), along with enum types.

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/model/deprecatedrenamerequest.java
jeffalder: I think this is good now.

Issue #953, opened 2018-12-14
S3: Data read has a different checksum than expected


ben-manes:
This might be fixed by 472f5bfbce08038653b641b070e3ac09ae846313, but not sure. I also am seeing some errors where the `response.contentLength()` differs from the actual size again, so there might be a race condition again. This is using `2.1.4` + netty.

```
Caused by: software.amazon.awssdk.core.exception.SdkClientException: Data read has a different checksum than expected.
        at software.amazon.awssdk.core.exception.SdkClientException$BuilderImpl.build(SdkClientException.java:97)
        at software.amazon.awssdk.core.exception.SdkClientException.create(SdkClientException.java:39)
        at software.amazon.awssdk.services.s3.internal.handlers.AsyncChecksumValidationInterceptor.validatePutObjectChecksum(AsyncChecksumValidationInterceptor.java:105)
        at software.amazon.awssdk.services.s3.internal.handlers.AsyncChecksumValidationInterceptor.afterUnmarshalling(AsyncChecksumValidationInterceptor.java:91)
        at software.amazon.awssdk.core.interceptor.ExecutionInterceptorChain.lambda$afterUnmarshalling$9(ExecutionInterceptorChain.java:152)
        at software.amazon.awssdk.core.interceptor.ExecutionInterceptorChain.reverseForEach(ExecutionInterceptorChain.java:210)
        at software.amazon.awssdk.core.interceptor.ExecutionInterceptorChain.afterUnmarshalling(ExecutionInterceptorChain.java:152)
        at software.amazon.awssdk.core.client.handler.BaseClientHandler.runAfterUnmarshallingInterceptors(BaseClientHandler.java:120)
        at software.amazon.awssdk.core.client.handler.BaseClientHandler.lambda$interceptorCalling$2(BaseClientHandler.java:133)
        at software.amazon.awssdk.core.client.handler.AttachHttpMetadataResponseHandler.handle(AttachHttpMetadataResponseHandler.java:40)
        at software.amazon.awssdk.core.client.handler.AttachHttpMetadataResponseHandler.handle(AttachHttpMetadataResponseHandler.java:28)
        at software.amazon.awssdk.core.internal.http.async.SyncResponseHandlerAdapter.lambda$prepare$0(SyncResponseHandlerAdapter.java:85)
```

dylanmroweFS:
[...]

PR #1550, opened 2019-12-17
Reset checksum calculation on retry

mar-kolya:
## Description
Reset checksum calculation on retry
Otherwise previous calculations of the checksum affect checksum calculated on retry.

## Motivation and Context
This fixes at least one possible cause of #953. After this is applied we no longer see errors for this bug in our system.

## Testing
We've run this fix with our service and observed that there were no errors anymore.

## Types of changes

## Checklist

## License

Signed-off-by: Nikolay Martynov <mar.kolya@gmail.com>

### DIFF
--- a/services/s3/src/main/java/software/amazon/awssdk/services/s3/checksums/ChecksumCalculatingAsyncRequestBody.java
+++ b/services/s3/src/main/java/software/amazon/awssdk/services/s3/checksums/ChecksumCalculatingAsyncRequestBody.java
@@ -54,6 +54,9 @@ public class ChecksumCalculatingAsyncRequestBody implements AsyncRequestBody {
                                       SdkChecksum sdkChecksum) {
             this.wrapped = wrapped;
             this.checksum = sdkChecksum;
+            // Note: there may be multiple places that hold references to this checksum object, but assumption is that in cse of
+            // retry all 'writes' to this will be done before new attempt to upload is run.
+            checksum.reset();
         }
 
         @Override


millems:
Great catch! Will review...

services/s3/src/main/java/software/amazon/awssdk/services/s3/checksums/ChecksumCalculatingAsyncRequestBody.java
```diff
@@ -54,6 +54,9 @@ public void subscribe(Subscriber<? super ByteBuffer> s) {
                                       SdkChecksum sdkChecksum) {
             this.wrapped = wrapped;
             this.checksum = sdkChecksum;
+            // Note: there may be multiple places that hold references to this checksum object, but assumption is that in cse of
+            // retry all 'writes' to this will be done before new attempt to upload is run.
+            checksum.reset();
```
millems: It's a bit weird that `SdkChecksum` is injected, if we'll be resetting it here, don't you think? Should we create a new SdkChecksum for each `subscribe` call? I know that involves a bit of refactoring, but it seems cleaner overall.

LMK if you don't want to invest this time and we can take a crack at it when we get a chance.

mar-kolya: I would agree that this design is weird.
I do not think I'll have to for this at the moment, unfortunately.
However it would be really appreciate the fix being released sooner rather than later - the bug has been opened for quite a while and is quite serious in a sense that it may lead to data being lost.

millems: I understand. I'll pick up this change and ferry it through.

millems: Yeah, it's definitely too much work to refactor: https://github.com/aws/aws-sdk-java-v2/pull/1552

millems:
Closing to continue in https://github.com/aws/aws-sdk-java-v2/pull/1552.

Thanks for your contribution!

PR #1552, opened 2019-12-18
Do not raise a "checksum mismatch" when an async PutObject is retried.

millems:


services/s3/src/main/java/software/amazon/awssdk/services/s3/checksums/ChecksumsEnabledValidator.java
```diff
@@ -131,7 +131,9 @@ public static void validatePutObjectChecksum(PutObjectResponse response, Executi
             byte[] ssHash = Base16Lower.decode(response.eTag().replace("\"", ""));
 
             if (!Arrays.equals(digest, ssHash)) {
-                throw SdkClientException.create("Data read has a different checksum than expected.");
+                throw SdkClientException.create(
+                    String.format("Data read has a different checksum than expected. Was 0x%s, but expected 0x%s",
```
spfink: Not sure how much the actual values is useful...

millems: It was already logging them on the "getobject" case, so I was just making it consistent...

PR #1553, opened 2019-12-19
Do not validate checksums when the service reports it is using customer-managed keys with server-side encryption.

millems:
Previously this was only done when the customer reported this in their request.


services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/handlers/AsyncChecksumValidationInterceptor.java
```diff
@@ -76,10 +79,10 @@
 
     @Override
     public void afterUnmarshalling(Context.AfterUnmarshalling context, ExecutionAttributes executionAttributes) {
-        boolean putObjectChecksumsEnabled =
-            putObjectChecksumEnabled(context.request(), ASYNC, executionAttributes, context.httpRequest());
+        boolean recordingChecksum = executionAttributes.getAttribute(ASYNC_RECORDING_CHECKSUM) == Boolean.TRUE;
```
spfink: Should we flip the comparison here?

PR #1554, opened 2019-12-20
AWS-XML: Support for handling error in response body with success HTT…

bmaizels:
…P code for all non-streaming requests.

S3 uses AWS XML protocol. Some S3 operations will very occasionally return a serialized error in the XML body of an HTTP response that has a status code of 200. This change refactors the core response pipeline to delegate responsibility for determining whether a response is an error or not upstream to the protocol handler so special logic could be inserted for S3 to deal with this specific and unusual use-case.

TODO checklist (future PRs):
- Wrap the inputStream with a capturing input stream so that the raw response can be inserted into the exception.

## Motivation and Context
Occasionally S3 will respond with an HTTP status 200 but with a serialized error in the XML payload. Before this change, the SDK would incorrectly report such a response as a success.

Initially we considered adding this as an interceptor, however that would have required us to proactively parse the XML to check for an error, somehow reset the input stream and then re-parse the response body for every operation that might have an error in the body and we didn't want to take the performance hit of doing that.

## Testing
Added unit test for the problem case. Ran all existing unit tests, and integration tests for S3.

## Types of changes

## Checklist

## License


core/protocols/aws-query-protocol/src/main/java/software/amazon/awssdk/protocols/query/internal/unmarshall/AwsXmlErrorUnmarshaller.java
```diff
@@ -0,0 +1,233 @@
...
+import software.amazon.awssdk.protocols.core.ExceptionMetadata;
+import software.amazon.awssdk.protocols.query.unmarshall.XmlElement;
+import software.amazon.awssdk.protocols.query.unmarshall.XmlErrorUnmarshaller;
+
+/**
+ * Unmarshalls an AWS XML exception from parsed XML.
+ */
+@SdkInternalApi
+public final class AwsXmlErrorUnmarshaller {
+    private static final String X_AMZN_REQUEST_ID_HEADER = "x-amzn-RequestId";
```
spfink: Try and unmarshall id-2 as well?

core/protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/AwsS3ProtocolFactory.java
```diff
@@ -41,6 +51,12 @@ private AwsS3ProtocolFactory(Builder builder) {
         return Optional.of(document);
     }
 
+    private Optional<XmlElement> getErrorRootFromSuccessBody(XmlElement document) {
+        return ERROR_IN_SUCCESS_BODY_ELEMENT_NAME.equals(document.elementName())
+            ? Optional.of(document)
```
spfink: Does this follow our checkstyle?

core/protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/AwsXmlProtocolFactory.java
```diff
@@ -100,10 +107,32 @@
             staxOperationMetadata.isHasStreamingSuccessResponse());
     }
 
+    protected <T extends AwsResponse> Function<AwsXmlUnmarshallingContext, T> createResponseTransformer(
+        Supplier<SdkPojo> pojoSupplier) {
+
+        return new AwsXmlResponseTransformer<>(
+            XmlProtocolUnmarshaller.builder().build(), r -> pojoSupplier.get());
```
spfink: Add a create() to XmlProtocolUnmarshaller?

core/protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/internal/unmarshall/AwsXmlPredicatedResponseHandler.java
```diff
@@ -0,0 +1,183 @@
...
+ */
+@SdkInternalApi
+public class AwsXmlPredicatedResponseHandler<OutputT> implements HttpResponseHandler<Response<OutputT>> {
+    private static final Logger log = LoggerFactory.getLogger(AwsXmlPredicatedResponseHandler.class);
+
+    private final Function<SdkHttpFullResponse, SdkPojo> pojoSupplier;
+    private final Function<AwsXmlUnmarshallingContext, OutputT> successResponseTransformer;
+    private final Function<AwsXmlUnmarshallingContext, ? extends SdkException> errorResponseTransformer;
+    private final Function<AwsXmlUnmarshallingContext, AwsXmlUnmarshallingContext> decorateContextWithError;
+    private final boolean needsConnectionLeftOpen;
```
spfink: This seems out of place for this class...

core/protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/internal/unmarshall/AwsXmlResponseTransformer.java
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * A transformer function that takes a parsed XML response and converts it into an {@link AwsResponse}. Used
+ * as a component in the {@link AwsXmlPredicatedResponseHandler}.
+ */
+@SdkInternalApi
+public final class AwsXmlResponseTransformer<T extends AwsResponse>
+        implements Function<AwsXmlUnmarshallingContext, T> {
+
+    private static final String X_AMZN_REQUEST_ID_HEADER = "x-amzn-RequestId";
```
spfink: id-2 since they are common in xml?

core/sdk-core/src/it/java/software/amazon/awssdk/core/http/AmazonHttpClientSslHandshakeTimeoutIntegrationTest.java
```diff
@@ -61,8 +64,7 @@ public void testSslHandshakeTimeout() {
                       .request(request)
                       .originalRequest(NoopTestRequest.builder().build())
                       .executionContext(executionContext(request))
-                      .errorResponseHandler(new NullErrorResponseHandler())
-                      .execute();
+                      .execute(combinedSyncResponseHandler(null, new NullErrorResponseHandler()));
```
spfink: Is this common in any client?

core/sdk-core/src/it/java/software/amazon/awssdk/core/http/ConnectionPoolMaxConnectionsIntegrationTest.java
```diff
@@ -79,7 +82,7 @@ public void leasing_a_new_connection_fails_with_connection_pool_timeout() {
                       .request(request)
                       .originalRequest(NoopTestRequest.builder().build())
                       .executionContext(executionContext(request))
-                      .execute();
+                      .execute(combinedSyncResponseHandler(null, null));
```
spfink: Same as previous question

core/sdk-core/src/main/java/software/amazon/awssdk/core/client/handler/SdkAsyncClientHandler.java
```diff
@@ -20,6 +20,7 @@
 import software.amazon.awssdk.annotations.ThreadSafe;
 import software.amazon.awssdk.core.client.config.SdkClientConfiguration;
 import software.amazon.awssdk.core.client.config.SdkClientOptionValidation;
+import software.amazon.awssdk.core.internal.handler.BaseAsyncClientHandler;
```
spfink: ?

core/sdk-core/src/main/java/software/amazon/awssdk/core/interceptor/SdkInternalExecutionAttribute.java
```diff
@@ -34,11 +35,23 @@
...
     /**
      * The key to store the {@link CompletableFuture} returned by {@link AsyncResponseTransformer#prepare()} method
-     * in the first attempt of a request. This is used only for async streaming requests
+     * in the first attempt of a request. This used to be used only for async streaming requests.
+     *
+     * The purpose this attribute used to service has been replaced with
+     * {@link SdkInternalExecutionAttribute#EXECUTION_ATTEMPT} which now tracks the execution attempt number and is
+     * incremented with successive retries.
      */
+    @Deprecated
```
spfink: Sucks we marked SdkInternalExecutionAttribute as protected

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/handler/BaseAsyncClientHandler.java
```diff
@@ -120,22 +171,20 @@ protected BaseAsyncClientHandler(SdkClientConfiguration clientConfiguration,
                 finalizeSdkHttpRequestContext.requestBody().isPresent()) {
                 marshalled = marshalled.toBuilder()
                                        .contentStreamProvider(
-                                           finalizeSdkHttpRequestContext.requestBody().get().contentStreamProvider())
+                                           finalizeSdkHttpRequestContext.requestBody()
+                                                                        // To appease static-code analysis
+                                                                        .orElse(null)
+                                                                        .contentStreamProvider())
```
spfink: This seems hacky and easy to get an exception

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/handler/BaseClientHandler.java
```diff
@@ -182,7 +197,67 @@ protected boolean isCalculateCrc32FromCompressedData() {
...
+
+    private static <T extends SdkResponse> BiFunction<T, SdkHttpFullResponse, T>
+        responseTransformations(ExecutionContext executionContext) {
+
+        return composeResponseFunctions(runAfterUnmarshallingInterceptors(executionContext),
+                                        attachHttpResponseToResult());
+    }
+
+    private static <T, R> BiFunction<T, R, T> composeResponseFunctions(BiFunction<T, R, T> function1,
+                                                               BiFunction<T, R, T> function2) {
```
spfink: Wouldn't think this passes checkstyle either...

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/IdempotentAsyncResponseHandler.java
```diff
@@ -0,0 +1,117 @@
...
+
+        return new IdempotentAsyncResponseHandler<>(wrappedHandler,
+                                                    preparedScopeSupplier,
+                                                    scopeInRangePredicate);
+    }
+
+    @Override
+    public CompletableFuture<T> prepare() {
+        // Test if the new scope is in range of the cached scope; if not delegate and cache result and new scope
+        if (cachedPreparedFuture.get() == null ||
```
spfink: This is some heavy new code added along a common path...

codegen/src/main/java/software/amazon/awssdk/codegen/poet/client/specs/XmlProtocolSpec.java
```diff
@@ -49,6 +53,31 @@ public XmlProtocolSpec(IntermediateModel model,
...
                                      OperationModel opModel) {
+
+        if (opModel.hasStreamingOutput()) {
+            return streamingResponseHandler(opModel);
+        }
+
+        ClassName responseType = poetExtensions.getModelClass(opModel.getReturnType().getReturnType());
+
+        TypeName handlerType =
+            com.squareup.javapoet.ParameterizedTypeName.get(
```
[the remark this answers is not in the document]
bmaizels: That's a mistake (I had bad IDE settings for about 20 minutes before I noticed, probably missed this in the cleanup). Will fix.

core/protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/internal/unmarshall/AwsXmlPredicatedResponseHandler.java
```diff
@@ -0,0 +1,183 @@
...
+                    + parsedResponse.sdkHttpFullResponse().statusCode() + ", Response Text: "
+                    + parsedResponse.sdkHttpFullResponse().statusText().orElse(null);
+            throw SdkClientException.builder().message(errorMessage).cause(e).build();
+        }
+    }
+
+    /**
+     * Responsible for handling an error response, including unmarshalling the error response
+     * into the most specific exception type possible, and throwing the exception.
+     *
```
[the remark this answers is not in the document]
bmaizels: Will fix.

core/protocols/aws-query-protocol/src/main/java/software/amazon/awssdk/protocols/query/internal/unmarshall/AwsXmlErrorUnmarshaller.java
bmaizels: I don't understand the comment... can you paraphrase/explain?

core/protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/AwsS3ProtocolFactory.java
bmaizels: Yes...? Are you concerned that I'm calling a method (equals) on a String constant? I consider that good practice because it mitigates the NPE that could be thrown if document.elementName() happens to be null.

core/protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/AwsXmlProtocolFactory.java
bmaizels: Will fix.

core/protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/internal/unmarshall/AwsXmlPredicatedResponseHandler.java
bmaizels: What seems out of place? The 'needsConnectionLeftOpen'? That is required because it's the responsibility of this class to close the connection.

core/protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/internal/unmarshall/AwsXmlResponseTransformer.java
bmaizels: Still confused. Can you explain?

core/sdk-core/src/it/java/software/amazon/awssdk/core/http/AmazonHttpClientSslHandshakeTimeoutIntegrationTest.java
bmaizels: The execute() method on the client was internal and was only ever called by tests. I consider that an antipattern so I removed it, and made the test do a bit of extra heavy lifting instead.

core/sdk-core/src/it/java/software/amazon/awssdk/core/http/ConnectionPoolMaxConnectionsIntegrationTest.java
bmaizels: The execute() method on the client was internal and was only ever called by tests. I consider that an antipattern so I removed it, and made the test do a bit of extra heavy lifting instead.

core/sdk-core/src/main/java/software/amazon/awssdk/core/client/handler/SdkAsyncClientHandler.java
bmaizels: I moved BaseAsyncClientHandler to internal namespace, that required an import where none was required before.

core/sdk-core/src/main/java/software/amazon/awssdk/core/interceptor/SdkInternalExecutionAttribute.java
bmaizels: I agree. Resolving this conversation because there's nothing you or I can do about it at this point.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/handler/BaseAsyncClientHandler.java
bmaizels: Yes I was trying to appease the static analysis gods and made it worse. I'm going to revert.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/handler/BaseClientHandler.java
bmaizels: Why not?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/IdempotentAsyncResponseHandler.java
bmaizels: It's replacing some code that instead persists the CompletableFuture object created by prepare() in ExecutionAttributes and then reads it out later when deciding whether to call prepare a second time. Every retry also writes explicitly writes null into ExecutionAttributes to effectively 'reset' the state. I think this replacement wrapper is not heavier than the code it replaces.

bmaizels: For reference the implementation of .get() on an AtomicReference just reads and returns the value of a volatile member variable.

core/protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/internal/unmarshall/AwsXmlResponseTransformer.java
zoewangg: id-2 and other extended response metadata is stored in `generateResponseMetadata` and accessible from service specific response metadata class, like `S3ResponseMetadata`

core/protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/AwsS3ProtocolFactory.java
spfink: Nope, was on the location of the ? on the new line. Thought we were more indented usually

core/sdk-core/src/it/java/software/amazon/awssdk/core/http/AmazonHttpClientSslHandshakeTimeoutIntegrationTest.java
spfink: I meant having null on either both or one of the response handlers

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/handler/BaseClientHandler.java
spfink: The second BiFunction should line up with the first.

core/protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/internal/unmarshall/AwsXmlPredicatedResponseHandler.java
spfink: Why is that the responsibility of this class?

core/protocols/aws-query-protocol/src/main/java/software/amazon/awssdk/protocols/query/internal/unmarshall/AwsXmlErrorUnmarshaller.java
bmaizels: As discussed we'll leave this alone as it's existing/refactored code.

core/protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/internal/unmarshall/AwsXmlPredicatedResponseHandler.java
bmaizels: Before I made this change it was already established as the responsibility of the 'success handler' to determine whether the underlying connection could/should be closed after handling the response, this is just a refactoring of that responsibility. In practice this flag should never be set to true because this class should not be used with streaming operations, but that's due to circumstance and I don't think that constraint should be codified.

core/protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/internal/unmarshall/AwsXmlResponseTransformer.java
bmaizels: As discussed with @spfink we're going to leave this as-is as it's existing/refactored code and there's no reason to touch it.

core/sdk-core/src/it/java/software/amazon/awssdk/core/http/AmazonHttpClientSslHandshakeTimeoutIntegrationTest.java
bmaizels: The null is handled by combinedSyncResponseHandler() which is a test helper method. There's nothing in the actual code that expects or handles a null for a handler (although there used to be through that test-only path).

core/sdk-core/src/it/java/software/amazon/awssdk/core/http/ConnectionPoolMaxConnectionsIntegrationTest.java
bmaizels: The null is handled by combinedSyncResponseHandler() which is a test helper method. There's nothing in the actual code that expects or handles a null for a handler (although there used to be through that test-only path).

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/handler/BaseClientHandler.java
bmaizels: Will fix.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/IdempotentAsyncResponseHandler.java
bmaizels: As discussed leaving as-is for now, but considering as a candidate for attention when tuning for performance. An option we have is to remove the thread-safety, which will give us a small gain on prepare()

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/handler/BaseClientHandler.java
```diff
@@ -182,7 +197,67 @@ protected boolean isCalculateCrc32FromCompressedData() {
...
+            }
+
+            if (executionParams.getErrorResponseHandler() != null) {
+                throw new IllegalArgumentException("Only one of 'combinedResponseHandler' and 'errorResponseHandler' "
+                                                   + "may be specified in a ClientExecutionParams object");
+            }
+        }
+    }
+
+    private static <T extends SdkResponse> BiFunction<T, SdkHttpFullResponse, T>
```
zoewangg: can we add some javadocs on those private methods while we are at it?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/handler/BaseAsyncClientHandler.java
```diff
@@ -149,6 +198,8 @@ protected BaseAsyncClientHandler(SdkClientConfiguration clientConfiguration,
             runAndLogError(
                 log.logger(),
                 "Error thrown from TransformingAsyncResponseHandler#onError, ignoring.",
+                // TODO: Have the StandardAsyncHttpResponseHandler forward errors and create a new wrapper in the
+                // MakeAsyncHttpRequestStage to stop errors from retries escaping. Then remove this hack.
```
zoewangg: Just curious, why is this a hack?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/handler/BaseAsyncClientHandler.java
```diff
@@ -60,44 +67,88 @@ protected BaseAsyncClientHandler(SdkClientConfiguration clientConfiguration,
...
     @Override
     public <InputT extends SdkRequest, OutputT extends SdkResponse, ReturnT> CompletableFuture<ReturnT> execute(
         ClientExecutionParams<InputT, OutputT> executionParams,
         AsyncResponseTransformer<OutputT, ReturnT> asyncResponseTransformer) {
 
-        // For streaming requests, prepare() should be called as early as possible to avoid NPE in client
-        // See https://github.com/aws/aws-sdk-java-v2/issues/1268
+        validateExecutionParams(executionParams);
+
+        if (executionParams.getCombinedResponseHandler() != null) {
```
zoewangg: should this also be moved to `validateExecutionParams`?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/handler/BaseAsyncClientHandler.java
```diff
@@ -60,44 +67,88 @@ protected BaseAsyncClientHandler(SdkClientConfiguration clientConfiguration,
...
         ClientExecutionParams<InputT, OutputT> executionParams) {
-        ExecutionContext executionContext = createExecutionContext(executionParams);
 
-        HttpResponseHandler<OutputT> decoratedResponseHandlers =
-            decorateResponseHandlers(executionParams.getResponseHandler(), executionContext);
+        validateExecutionParams(executionParams);
+        ExecutionContext executionContext = createExecutionContext(executionParams, createInitialExecutionAttributes());
+        TransformingAsyncResponseHandler<Response<OutputT>> combinedResponseHandler;
+
+        if (executionParams.getCombinedResponseHandler() == null) {
```
zoewangg: Can we use some other predicates here to imply `All other protocols take this path` and ` // XML non-streaming protocol takes this path` in the `else` block?

It's not very clear to me that `executionParams.getCombinedResponseHandler() != null` means `XML non-streaming protocol`

core/sdk-core/src/main/java/software/amazon/awssdk/core/interceptor/SdkInternalExecutionAttribute.java
zoewangg: I'd vote for removing it. It's a protected class and I'd be surprised if it's being used by any customers, and it's a pain to maintain it in the future.

core/sdk-core/src/main/java/software/amazon/awssdk/core/Response.java
```diff
@@ -0,0 +1,172 @@
...
+import software.amazon.awssdk.http.SdkHttpFullResponse;
+
+/**
+ * Response wrapper that indicates success or failure with the associated unmarshalled response object or exception
+ * object. This object is used by the core request/response pipeline to pass response metadata alongside the actual
+ * deserialized response object through different stages of the pipeline.
+ *
+ * @param <T> the modelled SDK response type.
+ */
+@SdkProtectedApi
```
zoewangg: internal api?

core/sdk-core/src/main/java/software/amazon/awssdk/core/Response.java
```diff
@@ -0,0 +1,172 @@
...
+        result = 31 * result + (httpResponse != null ? httpResponse.hashCode() : 0);
+        return result;
+    }
+
+    public static final class Builder<T> {
+        private Boolean isSuccess;
+        private T response;
+        private SdkException exception;
+        private SdkHttpFullResponse httpResponse;
+
```
zoewangg: missing private ctor

core/protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/internal/unmarshall/XmlResponseParserUtils.java
```diff
@@ -0,0 +1,61 @@
...
+     * Parse an XML response if one is expected and available. If we are not expecting a payload, but the HTTP response
+     * code shows an error then we will parse it anyway, as it should contain a serialized error.
+     * @param sdkPojo the SDK builder object associated with the final response
+     * @param response the HTTP response
+     * @return A parsed XML document or an empty XML document if no payload/contents were found in the response.
+     */
+    public static XmlElement parse(SdkPojo sdkPojo, SdkHttpFullResponse response) {
+
+        try {
+            if ((!response.isSuccessful() || hasPayloadMembers(sdkPojo)) && response.content().isPresent()) {
```
zoewangg: Can we extract the check statement to a well-named, well-documented separate method? It's a bit unclear to me that why just checking `response.content().isPresent` is not sufficient.

core/protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/internal/unmarshall/AwsXmlUnmarshallingContext.java
```diff
@@ -0,0 +1,165 @@
...
+        return result;
+    }
+
+    public static final class Builder {
+        private SdkHttpFullResponse sdkHttpFullResponse;
+        private XmlElement parsedXml;
+        private ExecutionAttributes executionAttributes;
+        private Boolean responseIsSuccess;
+        private XmlElement parsedErrorXml;
+
```
zoewangg: missing private ctor

core/protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/internal/unmarshall/AwsXmlUnmarshallingContext.java
```diff
@@ -0,0 +1,165 @@
...
+     * The {@link ExecutionAttributes} associated with this request.
+     */
+    public ExecutionAttributes executionAttributes() {
+        return executionAttributes;
+    }
+
+    /**
+     * true if the response indicates success; false if not; null if that has not been determined yet
+     */
+    public Boolean responseIsSuccess() {
```
zoewangg: `isResponseSuccess` or `isSuccessfulResponse`?

core/protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/internal/unmarshall/AwsXmlPredicatedResponseHandler.java
```diff
@@ -0,0 +1,183 @@
...
+    /**
+     * Standard constructor
+     * @param pojoSupplier A method that supplies an empty builder of the correct type
+     * @param successResponseTransformer A function that can unmarshall a response object from parsed XML
+     * @param errorResponseTransformer A function that can unmarshall an exception object from parsed XML
+     * @param decorateContextWithError A function that determines if the response was an error or not
+     * @param needsConnectionLeftOpen true if the underlying connection should not be closed once parsed
+     */
+    public AwsXmlPredicatedResponseHandler(
+        Function<SdkHttpFullResponse, SdkPojo> pojoSupplier,
```
zoewangg: Not sure about the constructor with 5 parameters and 4 of them are functions. My concern is that we might add more fields in the future and it's going to be super difficult to maintain different combinations of ctors.

core/protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/AwsS3ProtocolFactory.java
```diff
@@ -57,4 +73,48 @@ public AwsS3ProtocolFactory build() {
...
+        Supplier<SdkPojo> pojoSupplier, XmlOperationMetadata staxOperationMetadata) {
+
+        return new AwsXmlPredicatedResponseHandler<>(r -> pojoSupplier.get(),
+                                                     createResponseTransformer(pojoSupplier),
+                                                     createErrorTransformer(),
+                                                     findErrorInS3Response(),
+                                                     staxOperationMetadata.isHasStreamingSuccessResponse());
+    }
+
+    private Function<AwsXmlUnmarshallingContext, AwsXmlUnmarshallingContext> findErrorInS3Response() {
```
zoewangg: Can we make it an inner class instead? Because it's difficult to debug issues in a large function and a well-named inner class is more readable and testable, to me.

core/protocols/aws-xml-protocol/pom.xml
```diff
@@ -61,6 +61,25 @@
...
-
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-jar-plugin</artifactId>
+                <configuration>
+                    <archive>
+                        <manifestEntries>
+                            <Automatic-Module-Name>software.amazon.awssdk.protocols.xml</Automatic-Module-Name>
```
zoewangg: Oops, thanks for adding it!

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/CombinedResponseHandler.java
```diff
@@ -0,0 +1,150 @@
...
+    private SdkException handleErrorResponse(SdkHttpFullResponse httpResponse,
+                                             ExecutionAttributes executionAttributes)
+            throws IOException, InterruptedException {
+        try {
+            SdkException exception = errorResponseHandler.handle(httpResponse, executionAttributes);
+            exception.fillInStackTrace();
+            SdkStandardLogger.REQUEST_LOGGER.debug(() -> "Received error response: " + exception);
+            return exception;
+        } catch (InterruptedException | IOException e) {
+            throw e;
```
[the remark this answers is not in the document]
bmaizels: This code was refactored from the original HttpRequestStage handler, and retains all the same nuances. It wasn't already doing this, therefore if we think that's a bug we should probably address it separately and not as part of this change.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/IdempotentAsyncResponseHandler.java
```diff
@@ -0,0 +1,117 @@
...
+
+import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.annotations.ThreadSafe;
+import software.amazon.awssdk.http.SdkHttpResponse;
+
+/**
+ * Wrapper for a {@link TransformingAsyncResponseHandler} that allows attachment to an external scope and given a way
+ * of evaluating whether that scope has changed or not will only allow prepare() to be called on its delegate once
+ * per state change and will cache and reserve the future that is returned by the delegate the rest of the time.
+ *
```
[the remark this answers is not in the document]
bmaizels: Will fix.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/IdempotentAsyncResponseHandler.java
```diff
@@ -0,0 +1,117 @@
...
+                                                    scopeInRangePredicate);
+    }
+
+    @Override
+    public CompletableFuture<T> prepare() {
+        // Test if the new scope is in range of the cached scope; if not delegate and cache result and new scope
+        if (cachedPreparedFuture.get() == null ||
+            !newScopeInRangePredicate.test(newScopeSupplier.get(), cachedScope.get())) {
+
+            synchronized (this) {
```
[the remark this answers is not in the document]
bmaizels: I was having the same conversation with Sam. The thread-safety was force of habit to increase the re-usability of the class. It's unclear how much penalty we're actually paying for it since prepare() only gets called twice for most requests. I'm keeping it in our back pocket if we need to trim any millis on the perf tests once we're done with everything else.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/handler/BaseClientHandler.java
bmaizels: OK

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/IdempotentAsyncResponseHandler.java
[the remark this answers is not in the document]
zoewangg: My concern it that this is going to affect any async client. We can run some benchmarks to get an idea of the penalty and it'd be nice to understand how much it's going to affect the performance w/ the change. 

Yeah, agreed it's not a blocker for the PR, but let's get the benchmark numbers diff and create a backlog item to track it.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/handler/BaseAsyncClientHandler.java
bmaizels: Rereading it, I don't think this code is doing what I thought it was doing. I'm removing the TODO.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/handler/BaseAsyncClientHandler.java
bmaizels: That would break the non-streaming code-path as they both use validateExecutionParams and this additional check only applies to the streaming case.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/handler/BaseAsyncClientHandler.java
bmaizels: I don't think the comment is adding value. I'll try and break this code up into better named methods.

core/sdk-core/src/main/java/software/amazon/awssdk/core/interceptor/SdkInternalExecutionAttribute.java
bmaizels: Yes already done actually. I refactored the new execution attribute into a truly internal class, and left the rest of this one alone as it was being referenced from the genned code.

core/sdk-core/src/main/java/software/amazon/awssdk/core/Response.java
bmaizels: It's now protected because the codegen has responsibility to pass a combined handler that produces a Response object based on protocol. That means there's now a direct reference to this class from generated code.

core/sdk-core/src/main/java/software/amazon/awssdk/core/Response.java
bmaizels: Fixed.

core/protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/internal/unmarshall/AwsXmlUnmarshallingContext.java
bmaizels: Will fix.

core/protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/internal/unmarshall/XmlResponseParserUtils.java
bmaizels: The reason is that sometimes the response content is present but it's like an empty stream and the XML parser gets called anyway and then throws an exception because it's not valid XML so I assume this logic was added to avoid unecessary parsing of an empty stream. This code was refactored from another place, it is not new. I'll add a comment.

core/protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/internal/unmarshall/AwsXmlUnmarshallingContext.java
bmaizels: I like isResponseSuccess more. Changing.

core/protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/internal/unmarshall/AwsXmlPredicatedResponseHandler.java
bmaizels: It's internal and only called from one place. I propose we keep it as-is for now and refactor in future if necessary as we're free to do so.

core/protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/AwsS3ProtocolFactory.java
bmaizels: I agree (especially the testable part)!! I'll do that. And write tests.

core/protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/AwsS3ProtocolFactory.java
bmaizels: It did pass checkstyle but I changed it anyway, see if you like the new version more.

core/protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/internal/unmarshall/AwsXmlUnmarshallingContext.java
spfink: Changed the name of the variable but not the method?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/handler/BaseSyncClientHandler.java
```diff
@@ -51,28 +57,51 @@ protected BaseSyncClientHandler(SdkClientConfiguration clientConfiguration,
...
+
+        ExecutionContext executionContext = createExecutionContext(executionParams, createInitialExecutionAttributes());
 
         HttpResponseHandler<OutputT> decoratedResponseHandlers =
             decorateResponseHandlers(executionParams.getResponseHandler(), executionContext);
 
         HttpResponseHandler<ReturnT> httpResponseHandler =
             new HttpResponseHandlerAdapter<>(decoratedResponseHandlers, responseTransformer);
-        return execute(executionParams, executionContext, httpResponseHandler);
+
```
spfink: Extra new line here

bmaizels:
Benchmark results:

All deltas appear to be within margin of error.

Benchmark | (sslProviderValue) | (testItem) | Mode | Cnt | Score | Error | Units | Control Score | Control Error | Diff
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
s.a.a.b.apicall.httpclient.async.NettyClientH1NonTlsBenchmark.concurrentApiCall | N/A | N/A | thrpt | 3 | 9666.772 | 308.372 | ops/s | 7160.275 | 22157.44 | 2506.497
s.a.a.b.apicall.httpclient.async.NettyClientH1NonTlsBenchmark.sequentialApiCall | N/A | N/A | thrpt | 3 | 2860.288 | 108.906 | ops/s | 2612.264 | 379.512 | 248.024
s.a.a.b.apicall.httpclient.async.NettyHttpClientH1Benchmark.concurrentApiCall | jdk | N/A | thrpt | 3 | 8337.921 | 2893.23 | ops/s | 8590.875 | 1192.64 | -252.954
s.a.a.b.apicall.httpclient.async.NettyHttpClientH1Benchmark.concurrentApiCall | openssl | N/A | thrpt | 3 | 8696.178 | 3637.256 | ops/s | 8864.626 | 1788.643 | -168.448
s.a.a.b.apicall.httpclient.async.NettyHttpClientH1Benchmark.sequentialApiCall | jdk | N/A | thrpt | 3 | 2131.37 | 234.347 | ops/s | 2108.877 | 932.332 | 22.493
s.a.a.b.apicall.httpclient.async.NettyHttpClientH1Benchmark.sequentialApiCall | openssl | N/A | thrpt | 3 | 2440.526 | 809.596 | ops/s | 2289.892 | 105.093 | 150.634
s.a.a.b.apicall.httpclient.async.NettyHttpClientH2Benchmark.concurrentApiCall | jdk | N/A | thrpt | 3 | 4362.438 | 853.174 | ops/s | 4422.493 | 505.414 | -60.055
s.a.a.b.apicall.httpclient.async.NettyHttpClientH2Benchmark.concurrentApiCall | openssl | N/A | thrpt | 3 | 4966.218 | 355.114 | ops/s | 4851.642 | 3690.06 | 114.576
s.a.a.b.apicall.httpclient.async.NettyHttpClientH2Benchmark.sequentialApiCall | jdk | N/A | thrpt | 3 | 1773.576 | 556.887 | ops/s | 1781.88 | 154.8 | -8.304
s.a.a.b.apicall.httpclient.async.NettyHttpClientH2Benchmark.sequentialApiCall | openssl | N/A | thrpt | 3 | 1804.14 | 525.387 | ops/s | 1832.483 | 481.078 | -28.343
s.a.a.b.apicall.httpclient.sync.ApacheHttpClientBenchmark.concurrentApiCall | N/A | N/A | thrpt | 3 | 16145.81 | 1177.119 | ops/s | 15631.611 | 13821.98 | 514.195
s.a.a.b.apicall.httpclient.sync.ApacheHttpClientBenchmark.sequentialApiCall | N/A | N/A | thrpt | 3 | 3533.113 | 2419.538 | ops/s | 3625.046 | 925.672 | -91.933
s.a.a.b.apicall.httpclient.sync.UrlConnectionHttpClientBenchmark.sequentialApiCall | N/A | N/A | thrpt | 3 | 152.431 | 71.83 | ops/s | 159.357 | 30.611 | -6.926
s.a.a.b.apicall.protocol.Ec2ProtocolBenchmark.successfulResponse | N/A | N/A | thrpt | 3 | 7833.179 | 314.797 | ops/s | 7986.578 | 162.732 | -153.399
s.a.a.b.apicall.protocol.JsonProtocolBenchmark.successfulResponse | N/A | N/A | thrpt | 3 | 13694.1 | 674.068 | ops/s | 14058.659 | 1965.719 | -364.555
s.a.a.b.apicall.protocol.QueryProtocolBenchmark.successfulResponse | N/A | N/A | thrpt | 3 | 8505.287 | 3262.158 | ops/s | 8650.727 | 165.145 | -145.44
s.a.a.b.apicall.protocol.XmlProtocolBenchmark.successfulResponse | N/A | N/A | thrpt | 3 | 7830.517 | 562.464 | ops/s | 7737.278 | 370.641 | 93.239
s.a.a.b.marshaller.dynamodb.V1DynamoDbAttributeValue.getItem | N/A | TINY | thrpt | 3 | 1237671 | 145159.1 | ops/s | 1245949.442 | 44618.4 | -8278.61
s.a.a.b.marshaller.dynamodb.V1DynamoDbAttributeValue.getItem | N/A | SMALL | thrpt | 3 | 315646.2 | 48585.78 | ops/s | 319598.389 | 10923.59 | -3952.17
s.a.a.b.marshaller.dynamodb.V1DynamoDbAttributeValue.getItem | N/A | HUGE | thrpt | 3 | 43062.89 | 11088.48 | ops/s | 43063.494 | 777.565 | -0.601
s.a.a.b.marshaller.dynamodb.V1DynamoDbAttributeValue.putItem | N/A | TINY | thrpt | 3 | 920191.5 | 582007.2 | ops/s | 969062.581 | 50573 | -48871.1
s.a.a.b.marshaller.dynamodb.V1DynamoDbAttributeValue.putItem | N/A | SMALL | thrpt | 3 | 351612.1 | 132471.4 | ops/s | 408102.544 | 8886.067 | -56490.5
s.a.a.b.marshaller.dynamodb.V1DynamoDbAttributeValue.putItem | N/A | HUGE | thrpt | 3 | 56229.56 | 1973.811 | ops/s | 56899.852 | 811.853 | -670.292
s.a.a.b.marshaller.ec2.V1Ec2MarshallerBenchmark.marshall | N/A | TINY | thrpt | 3 | 3557001 | 474911.9 | ops/s | 3683916.041 | 107841.2 | -126915
s.a.a.b.marshaller.ec2.V1Ec2MarshallerBenchmark.marshall | N/A | SMALL | thrpt | 3 | 466854.7 | 28155.38 | ops/s | 463407.805 | 22409.44 | 3446.922
s.a.a.b.marshaller.ec2.V1Ec2MarshallerBenchmark.marshall | N/A | HUGE | thrpt | 3 | 464272.6 | 171156.2 | ops/s | 464969.916 | 12936.93 | -697.298

bmaizels:
Xml protocol benchmark (20 iterations): 

Control:

Result Benchmark                                 Mode  Cnt     Score    Error  Units
XmlProtocolBenchmark.successfulResponse  thrpt   20  7633.925 ± 93.102  ops/s

Experiment:
Benchmark                                 Mode  Cnt     Score    Error  Units
XmlProtocolBenchmark.successfulResponse  thrpt   20  7694.503 ± 88.972  ops/s

PR #1557, opened 2019-12-23
Flush parent before reading on H2 connection

dagnir:
## Description
It's possible that WINDOW_UPDATE in the outbound buffer is not be written in a
timely manner to the socket, causing subsequent read()'s to hang until the
remote endpoint sees the local window increase.

## Testing
New unit and integ tests. Running full integ and benchmark suites.

## Types of changes

## Checklist

## License

### DIFF
--- a/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/NettyRequestExecutor.java
+++ b/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/NettyRequestExecutor.java
@@ -62,6 +62,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import software.amazon.awssdk.annotations.SdkInternalApi;
 import software.amazon.awssdk.http.Protocol;
+import software.amazon.awssdk.http.nio.netty.internal.http2.FlushOnReadHandler;
 import software.amazon.awssdk.http.nio.netty.internal.http2.Http2ToHttpInboundAdapter;
 import software.amazon.awssdk.http.nio.netty.internal.http2.HttpToHttp2OutboundAdapter;
 import software.amazon.awssdk.http.nio.netty.internal.utils.ChannelUtils;
@@ -172,6 +173,9 @@ public final class NettyRequestExecutor {
         }
 
         pipeline.addLast(LastHttpContentHandler.create());
+        if (Protocol.HTTP2.equals(protocol)) {
+            pipeline.addLast(FlushOnReadHandler.getInstance());
+        }
         pipeline.addLast(new HttpStreamsClientHandler());
         pipeline.addLast(ResponseHandler.getInstance());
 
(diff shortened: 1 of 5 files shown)


http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/NettyRequestExecutor.java
```diff
@@ -172,6 +173,9 @@ private boolean tryConfigurePipeline() {
         }
 
         pipeline.addLast(LastHttpContentHandler.create());
+        if (Protocol.HTTP2.equals(protocol)) {
+            pipeline.addLast(FlushOnReadHandler.getInstance());
```
zoewangg: Do we need to remove `FlushOnReadHandler` from the pipeline when we release the channel?

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/FlushOnReadHandler.java
```diff
@@ -0,0 +1,47 @@
...
+
+    @Override
+    public void read(ChannelHandlerContext ctx) {
+        //Note: order is important, we need to fire the read() event first
+        // since it's what triggers the WINDOW_UPDATE frame write
+        ctx.read();
+        ctx.channel().parent().flush();
+    }
+
+    public static FlushOnReadHandler getInstance() {
```
zoewangg: `create` to be consistent with other static factory method?

dagnir: We have both :) https://github.com/aws/aws-sdk-java-v2/blob/8077c499ce6e887256079851230b67554fbc0762/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/SslCloseCompletionEventHandler.java#L58 

My preference is for `getInstance`

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/NettyRequestExecutor.java
dagnir: Yep! good catch.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/FlushOnReadHandler.java
zoewangg: Right, but we are using `create` in core consistently and I'd prefer using `create` going forward since this naming convention is written in our design doc. :)

https://github.com/aws/aws-sdk-java-v2/blob/master/docs/design/FavorStaticFactoryMethods.md#naming-conventions

dagnir: The key phrase is is "new instance" though. `getInstance()` makes it explicit that it's a singleton instance

PR #1569, opened 2020-01-02
Add RequestBody.fromRemainingByteBuffer()

### LINKED ISSUE #1534: RequestBody.fromByteBuffer does not respect ByteBuffer's position
The RequestBody.fromByteBuffer() (and AsyncRequsetBody.fromByteBuffer()) method reads the contents of the ByteBuffer starting from position 0 to the ByteBuffer's limit. This is incorrect usage of ByteBuffers.

## Expected Behavior
When passing RequestBody.fromByteBuffer() a ByteBuffer with a non-zero position, the contents of the request body should be the contents of the ByteBuffer from it's position to it's limit.

## Current Behavior
When passing RequestBody.fromByteBuffer() a ByteBuffer with a non-zero position, the contents of the request body are the contents of the ByteBuffer from position zero to it's limit.

## Possible Solution
RequestBody.fromByteBuffer() is currently defined as:
```
public static RequestBody fromByteBuffer(ByteBuffer byteBuffer) {
    return fromBytesDirect(BinaryUtils.copyAllBytesFrom(byteBuffer));
}
```
The bug can be fixed by changing it to:
```
public static RequestBody fromByteBuffer(ByteBuffer byteBuffer) {
    return fromBytesDirect(BinaryUtils.copyBytesFrom(byteBuffer));
}
```
Unfortunately, this is not a backward compatible change, if people are relying on this behavior, changing it would break them.

There are a couple of easy workarounds.
```
// fromByteBuffer makes a copy anyways, just read the contents yourself and use fromBytes
byte[] bytes = new byte[byteBuffer.remaining()];
byteBuffer.get(bytes);
RequestBody requsetBody = RequsetBody.fromBytes(bytes);

// slice() creates a new ByteBuffer with position = 0
RequestBody requestBody = RequestBody.fromByteBuffer(byteBuffer.slice());
```

JUnit 5 unit test:
```
@Test
void requestBody_fromByteBuffer_positionAtOne() throws IOException {
    byte[] expected = "this is a string".getBytes();

    // create a byte buffer with expected as it's contents, starting at position 1
    ByteBuffer byteBuffer = ByteBuffer.allocate(expected.length + 1);
    byteBuffer.position(1);
    byteBuffer.duplicate().put(expected);

    // double check it's correct
    assert byteBuffer.equals(ByteBuffer.wrap(expected));

    RequestBody body = RequestBody.fromByteBuffer(byteBuffer);
    try (InputStream is = body.contentStreamProvider().newStream()) {
        byte[] allBytes = is.readAllBytes();
        assert Arrays.equals(allBytes, expected);
    }
}
```

## Context
I ran into this when trying to do a multi-part upload to S3 from sections of the same ByteBuffer. Since the position is non-zero for all but the first part, each of the parts contained all the previous parts, making the uploaded object incorrect.

## Your Environment
* AWS Java SDK version used: 2.10.22, I think it affects all 2.x versions
* JDK version used: 13, but it is relevant to all versions.
* Operating System and version: Windows 10, should affect all operating systems.

dagnir:
Fixes #1534

## Testing
New unit tests.

## Types of changes

## Checklist

## License

### DIFF
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/sync/RequestBody.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/sync/RequestBody.java
@@ -161,12 +161,27 @@ public final class RequestBody {
     /**
      * Creates a {@link RequestBody} from a {@link ByteBuffer}. Buffer contents are copied so any modifications
      * made to the original {@link ByteBuffer} are not reflected in the {@link RequestBody}.
+     * <p>
+     * <b>NOTE:</b> This method always copies the entire contents of the buffer, ignoring the current read position. Use
+     * {@link #fromRemainingByteBuffer(ByteBuffer)} if you need it to copy only the remaining readable bytes.
      *
      * @param byteBuffer ByteBuffer to send to the service.
      * @return RequestBody instance.
      */
     public static RequestBody fromByteBuffer(ByteBuffer byteBuffer) {
         return fromBytesDirect(BinaryUtils.copyAllBytesFrom(byteBuffer));
+    }
+
+    /**
+     * Creates a {@link RequestBody} from the remaining readable bytes from a {@link ByteBuffer}. Unlike
+     * {@link #fromByteBuffer(ByteBuffer)}, this method respects the current read position of the buffer and reads only
+     * the remaining bytes. The buffer is copied before reading so no changes are made to original buffer.
+     *
+     * @param byteBuffer ByteBuffer to send to the service.
+     * @return RequestBody instance.
+     */
+    public static RequestBody fromRemainingByteBuffer(ByteBuffer byteBuffer) {
+        return fromBytesDirect(BinaryUtils.copyRemainingBytesFrom(byteBuffer));
     }
 
     /**
--- a/core/sdk-core/src/test/java/software/amazon/awssdk/core/sync/RequestBodyTest.java
+++ b/core/sdk-core/src/test/java/software/amazon/awssdk/core/sync/RequestBodyTest.java
@@ -20,7 +20,6 @@ import static org.assertj.core.api.Assertions.assertThat;
 import com.google.common.jimfs.Configuration;
 import com.google.common.jimfs.Jimfs;
 import java.io.File;
-import java.io.FileInputStream;
 import java.io.FileWriter;
 import java.io.IOException;
 import java.io.InputStream;
@@ -103,5 +102,24 @@ public class RequestBodyTest {
     public void emptyBytesConstructorHasCorrectContentType() {
         RequestBody requestBody = RequestBody.empty();
         assertThat(requestBody.contentType()).isEqualTo(Mimetype.MIMETYPE_OCTET_STREAM);
+    }
+
+    @Test
+    public void remainingByteBufferConstructorOnlyRemainingBytesCopied() throws IOException {
+        ByteBuffer bb = ByteBuffer.allocate(4);
+        bb.put(new byte[]{1, 2, 3, 4});
+        bb.flip();
+
+        bb.get();
+        bb.get();
+
+        int originalRemaining = bb.remaining();
+
+        RequestBody requestBody = RequestBody.fromRemainingByteBuffer(bb);
+
+        assertThat(requestBody.contentLength()).isEqualTo(originalRemaining);
+
+        byte[] requestBodyBytes = IoUtils.toByteArray(requestBody.contentStreamProvider().newStream());
+        assertThat(ByteBuffer.wrap(requestBodyBytes)).isEqualTo(bb);
     }
 }
--- a/utils/src/test/java/software/amazon/awssdk/utils/BinaryUtilsTest.java
+++ b/utils/src/test/java/software/amazon/awssdk/utils/BinaryUtilsTest.java
@@ -15,6 +15,7 @@
 
 package software.amazon.awssdk.utils;
 
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNull;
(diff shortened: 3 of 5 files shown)


core/sdk-core/src/main/java/software/amazon/awssdk/core/sync/RequestBody.java
```diff
@@ -169,6 +169,18 @@ public static RequestBody fromByteBuffer(ByteBuffer byteBuffer) {
...
     }
 
+    /**
+     * Creates a {@link RequestBody} from the remaining readable bytes from a {@link ByteBuffer}. Unlike
+     * {@link #fromByteBuffer(ByteBuffer)}, this method respects the current read position of the buffer and reads only
+     * the remaining bytes. The buffer is copied before reading so no changes are made to original buffer.
+     *
+     * @param byteBuffer ByteBuffer to send to the service.
+     * @return RequestBody instance.
+     */
```
bmaizels: Can we update the javadoc in the original fromByteBuffer(...) to make it clear that this one does -not- respect the current read position.

core/sdk-core/src/test/java/software/amazon/awssdk/core/sync/RequestBodyTest.java
```diff
@@ -104,4 +103,19 @@ public void emptyBytesConstructorHasCorrectContentType() {
...
+    @Test
+    public void remainingByteBufferConstructorOnlyRemainingBytesCopied() throws IOException {
+        ByteBuffer bb = ByteBuffer.allocate(4);
+        bb.putInt(42);
+        bb.flip();
+
+        bb.get();
+
+        RequestBody requestBody = RequestBody.fromRemainingByteBuffer(bb);
+        assertThat(requestBody.contentLength()).isEqualTo(bb.remaining());
```
bmaizels: Capture bb.remaining() before calling fromRemainingByteBuffer just to ensure that it's not being mutated.

bmaizels: Or have a separate test to assert such...

core/sdk-core/src/test/java/software/amazon/awssdk/core/sync/RequestBodyTest.java
```diff
@@ -104,4 +103,19 @@ public void emptyBytesConstructorHasCorrectContentType() {
...
+
+        byte[] requestBodyBytes = IoUtils.toByteArray(requestBody.contentStreamProvider().newStream());
+        assertThat(bb).isEqualTo(ByteBuffer.wrap(requestBodyBytes));
```
bmaizels: Not keen on the way this test is written, it's vulnerable to bb being mutated, also the assertion reads kinda back to front usually we assert the thing that's under test rather than the thing being under test as the thing we compare to.

utils/src/test/java/software/amazon/awssdk/utils/BinaryUtilsTest.java
```diff
@@ -159,4 +159,46 @@ public void testCopyBytesFrom_DirectByteBuffer_Idempotent() {
         assertTrue(partial1.length == 3);
         assertTrue(Arrays.equals(new byte[] {2, 3, 4}, partial1));
     }
+
+    @Test
+    public void testCopyRemainingBytesFrom_nullBuffer() {
+        assertNull(BinaryUtils.copyRemainingBytesFrom(null));
```
bmaizels: Can we use assertJ for all our new tests? I get that you don't want to refactor all the existing tests, but at least for new tests.

utils/src/test/java/software/amazon/awssdk/utils/BinaryUtilsTest.java
```diff
@@ -159,4 +159,46 @@ public void testCopyBytesFrom_DirectByteBuffer_Idempotent() {
...
+        bb.putInt(42);
+        bb.putInt(42);
+        bb.flip();
+
+        bb.getInt();
+
+        byte[] copy = BinaryUtils.copyRemainingBytesFrom(bb);
+        assertEquals(bb, ByteBuffer.wrap(copy));
+        assertEquals(4, copy.length);
+    }
```
bmaizels: These tests might be clearer if we work with bytes instead of ints. Also, we should assert that the sequence is read correctly, eg: given a sequence of [ 1, 2, 3, 4 ] of which we read two bytes we can assert that we are left with a buffer that is equivalent to [ 3, 4 ] rather than just checking length.

core/sdk-core/src/main/java/software/amazon/awssdk/core/sync/RequestBody.java
dagnir: +1

core/sdk-core/src/test/java/software/amazon/awssdk/core/sync/RequestBodyTest.java
dagnir: Will do

utils/src/test/java/software/amazon/awssdk/utils/BinaryUtilsTest.java
dagnir: +1 bytes instead of ints. The `assertEquals()` should take care of the ordering check since `ByteBuffer#equals()` enforces that.

utils/src/test/java/software/amazon/awssdk/utils/BinaryUtilsTest.java
dagnir: Yeah, I didn't want to start mixing them within the same test class, but I'm fine either way.

utils/src/test/java/software/amazon/awssdk/utils/BinaryUtilsTest.java
bmaizels: Yes, sorry my comment wasn't clear. My concern about that was you were writing the same int (42) twice, therefore you don't know if what you are left with is the start or the end of the original buffer.

core/sdk-core/src/test/java/software/amazon/awssdk/core/sync/RequestBodyTest.java
dagnir: Gotcha, I'll flip it.

utils/src/test/java/software/amazon/awssdk/utils/BinaryUtilsTest.java
dagnir: Ah gotcha, yep that makes sense.

Issue #1577, opened 2020-01-06
GetBucketPolicy with async S3 client doesn't work


avgdorom:
I'm trying to use the `S3AsyncClient` in order to get bucket policies from my AWS account.
I created a client, giving it the correct credentials and region, but I keep getting `null` in the content of the response.

Just to be sure that I initialized the client correctly, I tried other requests like, `getBucketPolicyStatus`, `listBuckets` and more. Every other request I tried returned the expected data, except for `getBucketPolicy`... In that case the response body is always null, which then causes an XML parsing error (I debugged until I came to `GetBucketPolicyInterceptor.modifyHttpResponseContent()`).

**NOTE**: For Exactly the same request, the S3 **sync** client works like a charm.

debora-ito:
Transferred the issue to Java V2 repo.

debora-ito:
@avgdorom thank you for reporting this, we'll investigate.

avgdorom:
@bmaizels Thanks! This is great. One question though: how do I get the fix? Do I need to wait for the next Java SDK release? If so, when is that due?

bmaizels:
This should be included with the 2.10.47 release on 1/10.

avgdorom:
@bmaizels Thanks, but in that case, I don't think it works... Tried it again twice, with version 2.10.47, and it just doesn't work. I get exactly what I reported in the bug in the first place.
Can you please check?

debora-ito:
@avgdorom actually it is available in today's release, version `2.10.48` - https://github.com/aws/aws-sdk-java-v2/blob/master/CHANGELOG.md#21048-2020-01-13

Apologies for the confusion.

bmaizels:
@avgdorom That was my fault, sorry about that confusion. I've retested it to be sure and I see broken in 2.10.47 and fixed in 2.10.48. Please let me know if you have any more problems with it.

avgdorom:
@bmaizels It's fine, it happens 😄 Works like a charm now! thank you.

PR #1586, opened 2020-01-10
Bugfix: s3 asynchronous client GetBucketPolicy now works and does not…

bmaizels:
… always through an XML parsing error

Fixes: https://github.com/aws/aws-sdk-java-v2/issues/1577

## Description

An interceptor to wrap the S3 output in XML so it could be parsed by the SDK had only been written for the synchronous client. I filled in the missing functionality for the async client.

## Motivation and Context

S3 GetBucketPolicy is totally broken in the async client.

## Testing

Functional + unit tests.

## Types of changes

## Checklist

## License


core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/EnvelopeWrappedSdkPublisher.java
```diff
@@ -0,0 +1,151 @@
...
+            }
+        }
+
+        @Override
+        public void onError(Throwable throwable) {
+            wrappedSubscriber.onError(throwable);
+        }
+
+        @Override
+        public void onComplete() {
```
dagnir: I'm not sure, but I think this behavior might violate the spec. A `Publisher` is allowed to signal `onComplete` without first getting a `request()` signal, so it's possible that this calls `onNext` before the downstream subscriber signals any demand which is not allowed

.changes/next-release/bugfix-AmazonS3-36e9d26.json
```diff
@@ -0,0 +1,5 @@
+{
+    "type": "bugfix",
+    "category": "Amazon S3",
```
zoewangg: minor: we probably should add another entry with category `AWS Java SDK for v2` since we've made changes in core as well.

core/sdk-core/src/main/java/software/amazon/awssdk/core/async/SdkPublishers.java
```diff
@@ -0,0 +1,64 @@
...
+                                                                    String envelopePrefix,
+                                                                    String envelopeSuffix) {
+        return EnvelopeWrappedSdkPublisher.of(publisher,
+                                              wrap(envelopePrefix),
+                                              wrap(envelopeSuffix),
+                                              SdkPublishers::concat);
+    }
+
+    private static ByteBuffer wrap(String s) {
+        return ByteBuffer.wrap(s.getBytes(StandardCharsets.UTF_8));
```
zoewangg: Should we add another parameter to specify Charset instead of assuming it's `UTF_8`? we can make it default to `UTF_8` if not specified

core/sdk-core/src/main/java/software/amazon/awssdk/core/async/SdkPublishers.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    private static ByteBuffer concat(ByteBuffer b1, ByteBuffer b2) {
+        ByteBuffer result = ByteBuffer.allocate(b1.limit() + b2.limit());
+        result.put(b1);
+        result.put(b2);
+        result.rewind();
```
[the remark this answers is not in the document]
bmaizels: If the input buffers are not already rewound then I think we'd only want to copy the remaining bytes, so I favor just using .remaining() instead. Do you agree?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/EnvelopeWrappedSdkPublisher.java
```diff
@@ -0,0 +1,151 @@
...
+ * header) and suffix (or footer). The prefix content will be prepended to the first published object from the
+ * wrapped publisher, and the suffix content will be published when the wrapped publisher signals completion.
+ * <p>
+ * The envelope prefix will not be published until the wrapped publisher publishes something or is completed.
+ * The envelope suffix will not be published until the wrapped publisher is completed.
+ * <p>
+ * This class can be used in an asynchronous interceptor in the AWS SDK to wrap content around the incoming
+ * bytestream from a response.
+ * <p>
+ * A function must be supplied that can be used to concatinate the envelope content to the content being published by
```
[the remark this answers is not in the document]
bmaizels: Will fix the spelling.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/EnvelopeWrappedSdkPublisher.java
```diff
@@ -0,0 +1,151 @@
...
+    @Override
+    public void subscribe(Subscriber<? super T> subscriber) {
+        wrappedPublisher.subscribe(new ContentWrappedSubscriber(subscriber));
+    }
+
+    private class ContentWrappedSubscriber implements Subscriber<T> {
+        private final Subscriber<? super T> wrappedSubscriber;
+        private final AtomicBoolean prefixApplied = new AtomicBoolean(false);
+        private final AtomicBoolean suffixApplied = new AtomicBoolean(false);
+
```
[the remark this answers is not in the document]
bmaizels: Great, I'll tear that out then.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/EnvelopeWrappedSdkPublisher.java
```diff
@@ -0,0 +1,151 @@
...
+ * If publisher publishes a single string "1", wrappedPublisher will publish "S1" (prepending the envelop prefix). If
+ * publisher then publishes a second string "2", wrappedPublisher will then publish "2" (no added content). If
+ * publisher then completes, wrappedPublisher will then publish "E" and then complete.
+ *
+ * @param <T> The type of objects being published
+ */
+
+@ThreadSafe
+@SdkInternalApi
+public class EnvelopeWrappedSdkPublisher<T> implements SdkPublisher<T> {
```
zoewangg: We probably should add reactive-stream tcl test for it as well.

https://github.com/aws/aws-sdk-java-v2/blob/master/CONTRIBUTING.md#reactive-streams

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/EnvelopeWrappedSdkPublisher.java
```diff
@@ -0,0 +1,151 @@
...
+
+        @Override
+        public void onError(Throwable throwable) {
+            wrappedSubscriber.onError(throwable);
+        }
+
+        @Override
+        public void onComplete() {
+            try {
+                // In the event onComplete() is called multiple times, only transmit the envelope once
```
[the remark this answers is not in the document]
bmaizels: Okay, will make that change.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/EnvelopeWrappedSdkPublisher.java
bmaizels: As discussed we'll change this to not send the envelope at all on an empty stream which is the only scenario we think this will crop up in.

core/sdk-core/src/main/java/software/amazon/awssdk/core/async/SdkPublishers.java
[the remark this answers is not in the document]
dagnir: Yeah I think either way works fine.

.changes/next-release/bugfix-AmazonS3-36e9d26.json
bmaizels: The core change is just a new class to support the S3 change (and possibly other similar changes in the future in other clients if it ever becomes necessary). Do you think it's still necessary to add a release note for a new utility class in this content?

core/sdk-core/src/main/java/software/amazon/awssdk/core/async/SdkPublishers.java
bmaizels: Maybe, seems like overkill in this case. I could change the name to wrapUtf8() instead?

.changes/next-release/bugfix-AmazonS3-36e9d26.json
zoewangg: Oh okay, I thought you made other changes as well. Then no need to create one for core.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/handlers/GetBucketPolicyInterceptor.java
```diff
@@ -33,24 +39,43 @@
...
+        if (INTERCEPTOR_CONTEXT_PREDICATE.test(context)) {
+            return context.responsePublisher().map(
+                body -> SdkPublishers.envelopeWrappedPublisher(body, XML_ENVELOPE_PREFIX, XML_ENVELOPE_SUFFIX));
+        }
+
+        return context.responsePublisher();
+    }
+
+    @Override
+    public void beforeUnmarshalling(Context.BeforeUnmarshalling context, ExecutionAttributes executionAttributes) {
```
zoewangg: Do we need this?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/EnvelopeWrappedSdkPublisher.java
bmaizels: As discussed offline, this implementation will never be truly compliant as crudely acts as a man-in-the-middle between an external publisher and subscriber. In this case we control the publisher side, so we're able to hide/mitigate the way we are inserting data where there should be none giving the outward appearance of a compliant publisher to any subscribers. I've marked it internal-only and clearly documented this gotcha/limitation.

bmaizels: Also I did run the TCL tests and fixed all the issues that were flagged that were not working as intended.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/handlers/GetBucketPolicyInterceptor.java
bmaizels: I don't think so... removing it.

PR #1588, opened 2020-01-10
CRT 0.4.13 API Update

rccarper:
## Description
* This branch contains changes from the following branches.
  * crt-api-update (https://github.com/aws/aws-sdk-java-v2/pull/1498)
  * aws-crt-dev-preview
  * master

This was done by rebasing master onto crt-api-update, then rebasing crt-api-update onto aws-crt-dev-preview, making some additional changes to account for API changes, and then creating this branch.  I'm open to merging this into aws-crt-dev-preview first if that's preferable to merging everything into master.

As such, there are a lot of changes here not done by myself.  Only the changes done in commit 2723640 are, in that sense, actually new.  More detail about those changes:
* This points the SDK to the latest build of the Java CRT available in Maven, ie, 0.4.19.
* This required updating a handful of places where the API has changed.
* I'm currently relying on the existing unit tests invoked by mvn clean install, and would be open to any other means of validating these changes if someone knows better.
* Possible point for API discussion: the AwsCrtAsyncHttpClient now takes an EventLoopGroup and HostResolver via its builder.  It would be possible to have the client create these objects itself, making a simpler experience for the user.  However, because this does add the cost of an EventLoopGroup and HostResolver to each created AwsCrtAsyncHttpClient, and handling it for them potentially hides that cost, it seemed better to have the user explicitly specify their own.  For example, a customer may use a pattern that creates a large number of clients, not realizing each client is spinning up an EventLoopGroup and HostResolver.

## Testing
mvn clean install

[x ] I confirm that this pull request can be released under the Apache 2 license

### DIFF
--- a/http-clients/aws-crt-client/src/main/java/software/amazon/awssdk/http/crt/AwsCrtAsyncHttpClient.java
+++ b/http-clients/aws-crt-client/src/main/java/software/amazon/awssdk/http/crt/AwsCrtAsyncHttpClient.java
@@ -45,6 +45,7 @@ import software.amazon.awssdk.http.async.AsyncExecuteRequest;
 import software.amazon.awssdk.http.async.SdkAsyncHttpClient;
 import software.amazon.awssdk.http.crt.internal.AwsCrtAsyncHttpStreamAdapter;
 import software.amazon.awssdk.utils.AttributeMap;
+import software.amazon.awssdk.utils.IoUtils;
 import software.amazon.awssdk.utils.Logger;
 import software.amazon.awssdk.utils.Validate;
 import software.amazon.awssdk.utils.http.SdkHttpUtils;
@@ -250,12 +251,12 @@ public class AwsCrtAsyncHttpClient implements SdkAsyncHttpClient {
     public void close() {
         isClosed.set(true);
         for (HttpConnectionPoolManager connPool : connectionPools.values()) {
-            connPool.close();
+            IoUtils.closeQuietly(connPool, log.logger());
         }
 
         while (ownedSubResources.size() > 0) {
             CrtResource r = ownedSubResources.pop();
-            r.close();
+            IoUtils.closeQuietly(r, log.logger());
         }
     }
 
--- a/http-clients/aws-crt-client/src/main/java/software/amazon/awssdk/http/crt/AwsCrtAsyncHttpClient.java
+++ b/http-clients/aws-crt-client/src/main/java/software/amazon/awssdk/http/crt/AwsCrtAsyncHttpClient.java
@@ -26,6 +26,7 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicBoolean;
 import software.amazon.awssdk.annotations.SdkPublicApi;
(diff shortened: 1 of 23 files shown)


http-clients/aws-crt-client/src/main/java/software/amazon/awssdk/http/crt/AwsCrtAsyncHttpClient.java
```diff
@@ -0,0 +1,370 @@
...
+    private final AtomicBoolean isClosed = new AtomicBoolean(false);
+    private final ClientBootstrap bootstrap;
+    private final SocketOptions socketOptions;
+    private final TlsContextOptions tlsContextOptions;
+    private final TlsContext tlsContext;
+    private final int windowSize;
+    private final int maxConnectionsPerEndpoint;
+
+
+    public AwsCrtAsyncHttpClient(DefaultBuilder builder, AttributeMap config) throws CrtRuntimeException {
```
graebm: Are you sure we want to declare a checked exception here?
caveat: I know very little about standards/style with Java exceptions

http-clients/aws-crt-client/src/main/java/software/amazon/awssdk/http/crt/AwsCrtAsyncHttpClient.java
```diff
@@ -0,0 +1,370 @@
...
+         * Resources are closed in the wrong order (Eg closing the Bootstrap/Threadpool when there are still open
+         * connections). By creating and owning our own Native CrtResources we can guarantee that things are shutdown
+         * in the correct order.
+         */
+
+        try (EventLoopGroup elg = new EventLoopGroup(builder.eventLoopSize);
+            HostResolver hr = new HostResolver(elg)) {
+
+            bootstrap = own(new ClientBootstrap(elg, hr));
+            socketOptions = own(new SocketOptions());
```
graebm: why are the `elg` and `hr` "try-with-resources", while the rest of the CrtResources are "owned"?

http-clients/aws-crt-client/src/main/java/software/amazon/awssdk/http/crt/AwsCrtAsyncHttpClient.java
rccarper: I originally pulled that over as part of inlining the ClientBootstrap constructor that's since been removed.  (It created the EventLoopGroup and HostResolver like I had also previously inlined.)  Because they are now passed as arguments, and nothing is being added, I don't think anything is being added here that warrants in.  (This is not to say that it *had* to be brought over in the first place.). So, I have taken it out.  However, I would be open to adding it back in if there is a good argument for it!

http-clients/aws-crt-client/src/main/java/software/amazon/awssdk/http/crt/AwsCrtAsyncHttpClient.java
rccarper: Shouldn't have to worry about this now that we aren't creating the EventLoopGroup or HostResolver inside of this constructor.  Previously, it was because I was brining over a chunk of code from a now removed ClientBootStrap constructor that created these objects internally.  Because of that, and because they were being passed to the ClientBootstrap, it didn't seem necessary to track them with "own".  (But I might have been wrong!)

PR #1603, opened 2020-01-21
Additional logging for GOAWAY

dagnir:
## Description
Should help during debugging and troubleshooting.

## Motivation and Context
Should help during debugging and troubleshooting.

## Types of changes

## Checklist

## License


http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/Http2MultiplexedChannelPool.java
```diff
@@ -316,7 +316,7 @@ private boolean acquireStreamOnInitializedConnection(MultiplexedChannelRecord ch
     }
 
     public void handleGoAway(Channel parentChannel, int lastStreamId, GoAwayException exception) {
-        log.debug(() -> "Received GOAWAY on " + parentChannel + " with lastStreamId of " + lastStreamId);
+        log.warn(() -> "Received GOAWAY on " + parentChannel + " with lastStreamId of " + lastStreamId);
```
zoewangg: Still not sure if we should set it to `warn` as `GOAWAY` is not uncommon and retry normally should work. Using `warn` might raise unnecessary concerns. 

Can we add the message in the exception instead?

dagnir: Yeah looking at this again, I don't think it's necessary to increase the level here. The changes `MultiplexedChannelRecord` should be all we need

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/MultiplexedChannelRecord.java
```diff
@@ -81,7 +82,15 @@ boolean acquireStream(Promise<Channel> promise) {
...
-                String message = "Connection received GOAWAY or was closed while acquiring new stream.";
+                String message;
+                // GOAWAY
+                if (state == RecordState.CLOSED_TO_NEW) {
+                    message = String.format("Connection %s received GOAWAY with Last Stream ID %d. Unable to open new "
+                                            + "streams on this connection.", connection, lastStreamId);
+                } else {
+                    message = String.format("Connection %s was closed while acquiring new stream.", connection);
+                }
+                log.warn(() -> message);
```
zoewangg: Same here. Can we add the message in the exception instead?

dagnir: I would rather keep this here. This is the specific case we care about right, trying to create a stream on one that has received a `GOAWAY`. The message is added to the exception as well

PR #1614, opened 2020-01-24
Throws IOException for an h2 connection acquiring race condition

zoewangg:
…s picked up at the same time it gets closed

## Description
- ~~Reduce the chances of the race condition where an HTTP2 connection gets reused at the same time it gets inactive~~

- Throw `IOException` instead of `IllegalStatementException` so the failed requests can be retried.

## Testing
Added unit tests

## Types of changes

## Checklist

## License

### DIFF
--- a/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/MultiplexedChannelRecord.java
+++ b/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/MultiplexedChannelRecord.java
@@ -28,6 +28,7 @@ import io.netty.util.concurrent.Future;
 import io.netty.util.concurrent.GenericFutureListener;
 import io.netty.util.concurrent.Promise;
 import io.netty.util.concurrent.ScheduledFuture;
+import java.io.IOException;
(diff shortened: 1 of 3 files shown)


http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/MultiplexedChannelRecord.java
```diff
@@ -248,9 +249,13 @@ public Channel getConnection() {
...
     }
 
-    public boolean claimStream() {
+    private boolean claimStream() {
         lastReserveAttemptTimeMillis = System.currentTimeMillis();
         for (int attempt = 0; attempt < 5; ++attempt) {
+            if (!connection.isActive()) {
+                return false;
+            }
+
```
millems: Do we know that this actually helps, or are we just guessing?

zoewangg: As discussed offline, the check does not cover for all cases and we need to figure out a better way to reduce the race condition. Removed it for now.

PR #1622, opened 2020-01-29
h2 connection exception handling 

zoewangg:
## Description
- Fix the issue where publisher signals onComplete if the stream is closed as a result of outbound GOAWAY. The root cause is that when `PingFailedException` is fired, `UnusedChannelExceptionHandler` is handling it instead of `ReleaseOnExceptionHandler`. `UnusedChannelExceptionHandler` only closes h2 connections. As a result, the exception would not be delivered to child channels as expected.

- ~~Change `UnusedChannelExceptionHandler` to a per-request handler because it doesn't really make sense to attach it to connection channel pipeline since we never set `IN_USE` attribute key on it.~~ Only attach `UnusedChannelExceptionHandler` to h2 stream channels and h1 channels.
- When delivering a connection exception to a child channel, wrap it with a more descriptive exception.

Previously, if a channel gets closed because of `PingFailedException`:
It could complete successfully or throw the following exception depending on the operation.
```
java.util.concurrent.CompletionException: java.io.IOException: Server failed to send complete response
```
Now it would throw the following exception:
```
java.util.concurrent.CompletionException: java.lang.IOException: An exception occurred on the connection: Server did not respond to PING after 487ms (limit: 200ms)
Caused by: software.amazon.awssdk.http.nio.netty.internal.http2.PingFailedException: Server did not respond to PING after 487ms (limit: 200ms)
```

## Testing
Added functional tests and unit tests.

## Types of changes

## Checklist

## License

### DIFF
--- a/.changes/next-release/bugfix-NettyNIOHTTPClient-dba3e80.json
+++ b/.changes/next-release/bugfix-NettyNIOHTTPClient-dba3e80.json
@@ -1,0 +1,5 @@
+{
+    "category": "Netty NIO HTTP Client", 
+    "type": "bugfix", 
+    "description": "Deliver exceptions to stream channels correctly if there's an exception thrown on connection. This also fixes a bug where publisher signals onComplete if the stream is closed as a result of outbound GOAWAY."
+}
--- a/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ChannelPipelineInitializer.java
+++ b/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ChannelPipelineInitializer.java
@@ -109,7 +109,14 @@ public final class ChannelPipelineInitializer extends AbstractChannelPoolHandler
         }
 
         pipeline.addLast(FutureCancelHandler.getInstance());
-        pipeline.addLast(UnusedChannelExceptionHandler.getInstance());
+
+        // Only add it for h1 channel because it does not apply to
+        // h2 connection channel. It will be attached
+        // to stream channels when they are created.
+        if (protocol == Protocol.HTTP1_1) {
+            pipeline.addLast(UnusedChannelExceptionHandler.getInstance());
+        }
+
         pipeline.addLast(new LoggingHandler(LogLevel.DEBUG));
     }
 
--- a/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/MultiplexedChannelRecord.java
+++ b/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/MultiplexedChannelRecord.java
@@ -38,6 +38,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.Consumer;
 import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.http.nio.netty.internal.UnusedChannelExceptionHandler;
 import software.amazon.awssdk.utils.Logger;
 
 /**
@@ -106,6 +107,7 @@ public class MultiplexedChannelRecord {
                 }
 
                 Http2StreamChannel channel = future.getNow();
+                channel.pipeline().addLast(UnusedChannelExceptionHandler.getInstance());
                 childChannels.put(channel.id(), channel);
                 promise.setSuccess(channel);
 
@@ -184,7 +186,16 @@ public class MultiplexedChannelRecord {
      * Delivers the exception to all registered child channels, and prohibits new streams being created on this connection.
      */
     void closeChildChannels(Throwable t) {
-        closeAndExecuteOnChildChannels(ch -> ch.pipeline().fireExceptionCaught(t));
+        closeAndExecuteOnChildChannels(ch -> ch.pipeline().fireExceptionCaught(decorateConnectionException(t)));
+    }
+
+    private Throwable decorateConnectionException(Throwable t) {
+        String message = "An error occurred on the connection: " + t.getMessage();
+        if (t instanceof IOException) {
+            return new IOException(message, t);
+        }
+
+        return new Throwable(message, t);
     }
 
     private void closeAndExecuteOnChildChannels(Consumer<Channel> childChannelConsumer) {
--- a/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/GoAwayException.java
+++ b/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/GoAwayException.java
@@ -24,7 +24,7 @@ import software.amazon.awssdk.annotations.SdkInternalApi;
  * Exception thrown when a GOAWAY frame is sent by the service.
  */
 @SdkInternalApi
-class GoAwayException extends IOException {
+public class GoAwayException extends IOException {
     private final String message;
 
     GoAwayException(long errorCode, ByteBuf debugData) {
--- a/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/PingFailedException.java
+++ b/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/PingFailedException.java
@@ -19,7 +19,7 @@ import java.io.IOException;
 import software.amazon.awssdk.annotations.SdkInternalApi;
 
 @SdkInternalApi
-final class PingFailedException extends IOException {
+public final class PingFailedException extends IOException {
     PingFailedException(String msg) {
         super(msg);
     }
--- a/http-clients/netty-nio-client/src/test/java/software/amazon/awssdk/http/nio/netty/fault/GoAwayTest.java
+++ b/http-clients/netty-nio-client/src/test/java/software/amazon/awssdk/http/nio/netty/fault/GoAwayTest.java
@@ -13,9 +13,10 @@
  * permissions and limitations under the License.
  */
 
-package software.amazon.awssdk.http.nio.netty.internal.http2;
+package software.amazon.awssdk.http.nio.netty.fault;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
 import io.netty.bootstrap.ServerBootstrap;
 import io.netty.buffer.ByteBuf;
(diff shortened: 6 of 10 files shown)


http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/MultiplexedChannelRecord.java
```diff
@@ -184,7 +185,19 @@ void closeChildChannels() {
      * Delivers the exception to all registered child channels, and prohibits new streams being created on this connection.
      */
     void closeChildChannels(Throwable t) {
-        closeAndExecuteOnChildChannels(ch -> ch.pipeline().fireExceptionCaught(t));
+        closeAndExecuteOnChildChannels(ch -> {
+            RuntimeException ioException = new RuntimeException("An exception occurred on the connection ", t);
```
zoewangg: `RuntimeException` might not be the best exception to be wrapped with. The reason I didn't use `IOException` is because I figured not all exceptions thrown on connection are retryable, but is that true?

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/HandlerRemovingChannelPool.java
```diff
@@ -83,7 +83,8 @@ private void removePerRequestHandlers(Channel channel) {
                            FlushOnReadHandler.class,
                            ResponseHandler.class,
                            ReadTimeoutHandler.class,
-                           WriteTimeoutHandler.class);
+                           WriteTimeoutHandler.class,
+                           UnusedChannelExceptionHandler.class);
```
millems: Do we still need the IN_USE checks in the handler if we are removing it?

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/MultiplexedChannelRecord.java
```diff
@@ -184,7 +185,19 @@ void closeChildChannels() {
...
+
+            // Deliver the exception only if the channel knows to handle the exception
+            if (ch.pipeline().get(ResponseHandler.class) != null) {
+                log.debug(() -> "Delivering exception to child channel " + ch + ioException);
+                ch.pipeline().fireExceptionCaught(ioException);
+            } else {
+                log.debug(() -> "Closing child channel " + ch + ioException);
+                ch.close();
```
millems: This seems super fragile. Can we just have a handler at the end of every pipeline that closes on exception?

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/MultiplexedChannelRecord.java
millems: Wouldn't we want IOExceptions to stay IOExceptions? Any reason we need to re-wrap here?

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/HandlerRemovingChannelPool.java
zoewangg: Hmm, good point. I'll change it to only add it once when the channel is created and not removing it.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/MultiplexedChannelRecord.java
zoewangg: The connection could get closed before any handler is attached to stream channels and there would be nothing catching the exception. I found out this issue in my testing where I set a very low `pingPeriod`.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/MultiplexedChannelRecord.java
zoewangg: The reason I re-wrap it here is to provide more descriptive message on the exception. Currently, it's not very clear why the request fails.

millems: Wouldn't this make currently-retried exceptions not be retried? Should we make it an IOException if the wrapped is an IOException?

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/MultiplexedChannelRecord.java
millems: Can we have a handler attached to the end of every channel as soon as it is created, before it is even connected, that calls `close()` when an exception is fired on the channel, assuming nothing else closed the channel already?

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/MultiplexedChannelRecord.java
zoewangg: I thought our default retry policy would check if the root cause of the exception is assignable from `IOException`, but turns out it would check one level down. 

https://github.com/aws/aws-sdk-java-v2/blob/138def267fc8814418e5b7c8ab80f038b9d40641/core/sdk-core/src/main/java/software/amazon/awssdk/core/retry/conditions/RetryOnExceptionsCondition.java#L59-L62

I'll make it an IOException if the wrapped is an IOException

millems: Or we could update the retry policy to check down the entire chain? That would probably be better.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/MultiplexedChannelRecord.java
zoewangg: Updated to attach `UnusedChannelExceptionHandler` to the stream channels once they are created and that should take care of it.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/HandlerRemovingChannelPool.java
zoewangg: Reverted this change. Now we only attach `UnusedChannelExceptionHandler` to h1 channels once they are created

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/MultiplexedChannelRecord.java
zoewangg: +1. I'll create a backlog item for it. I think it'd still be nice to wrap IOException with an IOException here to make it clear this is a transient retryable I/O related error.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ChannelPipelineInitializer.java
```diff
@@ -109,7 +109,14 @@ public void channelCreated(Channel ch) {
         }
 
         pipeline.addLast(FutureCancelHandler.getInstance());
-        pipeline.addLast(UnusedChannelExceptionHandler.getInstance());
+
+        // Only add it for h1 channel because it does not apply to
+        // h2 connection channel. It will be attached
+        // to stream channels when they are created.
+        if (protocol == Protocol.HTTP1_1) {
+            pipeline.addLast(UnusedChannelExceptionHandler.getInstance());
```
millems: What happens when an exception is fired on a parent H2 channel after this change?

zoewangg: `ReleaseOnExceptionHandler` will take care of it.

https://github.com/aws/aws-sdk-java-v2/blob/138def267fc8814418e5b7c8ab80f038b9d40641/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/Http2MultiplexedChannelPool.java#L380-L396

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/MultiplexedChannelRecord.java
```diff
@@ -184,7 +186,16 @@ void closeChildChannels() {
      * Delivers the exception to all registered child channels, and prohibits new streams being created on this connection.
      */
     void closeChildChannels(Throwable t) {
-        closeAndExecuteOnChildChannels(ch -> ch.pipeline().fireExceptionCaught(t));
+        closeAndExecuteOnChildChannels(ch -> ch.pipeline().fireExceptionCaught(decorateConnectionException(t)));
+    }
+
+    private Throwable decorateConnectionException(Throwable t) {
+        String message = "An error occurred on the connection: " + t.getMessage();
+        if (IOException.class.isAssignableFrom(t.getClass())) {
```
dagnir: minor: `instanceof` here is more concise

.changes/next-release/bugfix-NettyNIOHTTPClient-dba3e80.json
```diff
@@ -0,0 +1,5 @@
+{
+    "category": "Netty NIO HTTP Client", 
+    "type": "bugfix", 
+    "description": "Deliver exceptions to stream channels correctly if there's an exception thrown on connection.This also fixes a bug where publisher signals onComplete if the stream is closed as a result of outbound GOAWAY."
```
dagnir: nit: space in "connection.This"

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/MultiplexedChannelRecord.java
zoewangg: Fixed

.changes/next-release/bugfix-NettyNIOHTTPClient-dba3e80.json
zoewangg: Fixed

PR #1639, opened 2020-02-06
Chckyn/arn bugfix

chckyn:
## Description

Minor change to fix bug in how the Arn.fromString() method assigns values to the region and accountId fields when they are not specified.

## Motivation and Context

Arn.fromString() behaves differently than Arn.Builder when the region and accountId fields are not specified. This causes problems for consumers when the same behavior is expected from both methods.

## Testing

Unit tests pass.

## Types of changes

- [ x ] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [ x ] I have read the **CONTRIBUTING** document
- [ x ] Local run of `mvn install` succeeds
- [ x ] My code follows the code style of this project
- [ x ] I have read the **README** document
- [ x ] I have added tests to cover my changes
- [ x ] All new and existing tests passed
- [ x ] A short description of the change has been added to the **CHANGELOG**

## License

- [ x ] I confirm that this pull request can be released under the Apache 2 license


.changes/next-release/bugfix-AWSSDKforJavav2-67abe11.json
```diff
@@ -0,0 +1,5 @@
+{
+    "category": "AWS SDK for Java v2", 
+    "type": "bugfix", 
+    "description": "Change the Arn.fromString() method to store empty Optionals in the accountId and region fields when those fields on not specified in an resource ARN. This matches the behavior of the Arn.Builder class when those fields are not specified."
```
millems: Nit: fields on not -> fields are not

chckyn:
> Are there any other fields that should get the same treatment?

I don't believe so. It has been a minute since I was looking at this section of the code. Perhaps that is a good question for one of the maintainers?

millems:
> > Are there any other fields that should get the same treatment?
> 
> I don't believe so. It has been a minute since I was looking at this section of the code. Perhaps that is a good question for one of the maintainers?

Hah, I'm a maintainer, but I'd have to dive into the code to see as well.

chckyn:
> > > Are there any other fields that should get the same treatment?
> > 
> > 
> > I don't believe so. It has been a minute since I was looking at this section of the code. Perhaps that is a good question for one of the maintainers?
> 
> Hah, I'm a maintainer, but I'd have to dive into the code to see as well.

Of course. I suppose I meant the person most familiar with this section of the code base. Perhaps that is you as well =]. From looking at it myself, I don't believe the other fields require a similar change. I assume the `Validate.paramNotBlank` is taking care of the check on all the other fields.

chckyn:
I have no intention of seeing this pull request merged anymore.

PR #728, opened 2018-09-24
Update README

zoewangg:
Update README to reflect production readiness.

PR #1644, opened 2020-02-11
PR #728 attempt 2: Fixed buggy netty-reactive-streams test and re-added changes from PR

millems:
#728 was reverted because of an intermittent test failure, and we didn't want to risk releasing a bug. I've fixed that test failure after discovering it's a bug in the netty-reactive-streams test, not our implementation.

This change re-introduces the changes from #728 in one commit, and the fixes for the intermittent test failure in a second commit, for easier reviewing.

Before the fix, `HandlerPublisherVerificationTest#stochastic_spec103_mustSignalOnMethodsSequentially` would fail once every few hundred iterations. After the fix, `HandlerPublisherVerificationTest#stochastic_spec103_mustSignalOnMethodsSequentially` did not fail after a few thousand iterations.

### DIFF
--- a/http-clients/netty-nio-client/src/test/java/software/amazon/awssdk/http/nio/netty/internal/nrs/util/BatchedProducer.java
+++ b/http-clients/netty-nio-client/src/test/java/software/amazon/awssdk/http/nio/netty/internal/nrs/util/BatchedProducer.java
@@ -1,0 +1,80 @@
+/*
+ * Copyright 2010-2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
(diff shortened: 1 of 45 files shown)


http-clients/netty-nio-client/src/test/java/software/amazon/awssdk/http/nio/netty/internal/nrs/util/BatchedProducer.java
```diff
@@ -58,10 +58,18 @@ public void read(final ChannelHandlerContext ctx) throws Exception {
...
+                long valueToProduce = sequence.get();
+                for (int i = 0; i < batchSize && valueToProduce != eofOn;) {
+                    if (!sequence.compareAndSet(valueToProduce, valueToProduce + 1)) {
+                        // Another thread already produced this value, try again with the next value.
+                        valueToProduce = sequence.get();
+                        continue;
+                    }
+
+                    ctx.fireChannelRead(valueToProduce);
+                    ++i;
```
bmaizels: As discussed, let's look into simplifying this either with a single threaded test or using .getAndIncrement() to cache the value instead of get() that then requires a compareAndSet()

millems: Re-wrote it to use a single-threaded executor, which is what I *think* the original author was trying to do by scheduling it on the parent event-loop executor. That way there's no concurrency concerns at all.

PR #1651, opened 2020-02-17
String converters for Java types - 1 of 4 for Attribute Converters

spfink:
String converters that are used for converting from Java types to string. This is 1 of 4 PRs for adding attribute converters to DDB.

1. String Converters
2. Attribute Converters without collections
3. Collection Attribute Converters
4. Wiring

## Testing
String converters are tested as part of attribute converters in PR: https://github.com/aws/aws-sdk-java-v2/pull/1652

### DIFF
--- a/services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/converter/string/bundled/InstantStringConverter.java
+++ b/services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/converter/string/bundled/InstantStringConverter.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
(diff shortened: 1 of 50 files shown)


services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/converter/string/bundled/InstantStringConverter.java
```diff
@@ -0,0 +1,47 @@
...
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.enhanced.dynamodb.converter.string.bundled;
```
bmaizels: Comment applies to all converters with 'bundled' in the package name: not sure of the value of having that there, i'd vote to remove it.

PR #1652, opened 2020-02-17
Attribute converters for Java types - 2 of 4 for Attribute Converters

spfink:
Attribute converters that are used for converting from Java types to DDB AttributeValues. This is 2 of 4 PRs for adding attribute converters to DDB.

1. String Converters
2. Attribute Converters without collections
3. Collection Attribute Converters
4. Wiring


services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/converter/attribute/DefaultAttributeConverterProvider.java
```diff
@@ -0,0 +1,209 @@
...
+/**
+ *
+ * <p>
+ * Given an input, this will identify a converter that can convert the specific Java type and invoke it. If a converter cannot
+ * be found, it will invoke a "parent" converter, which would be expected to be able to convert the value (or throw an exception).
+ */
+@SdkPublicApi
+@ThreadSafe
+@Immutable
+public final class DefaultAttributeConverterProvider<T> implements AttributeConverterProvider {
```
bmaizels: Why does this class need a parameterized type?

PR #1663, opened 2020-02-24
Moving NamingUtils to internal as CodegenNamingUtils

spfink:


millems:
^ Pretend that was me

PR #1674, opened 2020-02-27
DynamoDb Enhanced Client : changed the way extensions are loaded, loa…

bmaizels:
…ds VersionedRecordExtension by default

## Motivation and Context
Many customers are going to want to use versioned records, and for those that don't if they don't tag any attributes on their table as 'version attributes' the extension will not get in the way so it made sense to load it by default.

We are leaving the door open in the future to provide meta-extensions that can do intelligent ordering of multiple extensions if such a thing is possible, but for now the order must be specified manually.

Relates to https://github.com/aws/aws-sdk-java-v2/issues/35

## Testing
Unit tests

## License

### DIFF
--- a/services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/DynamoDbEnhancedAsyncClient.java
+++ b/services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/DynamoDbEnhancedAsyncClient.java
@@ -35,7 +35,7 @@ import software.amazon.awssdk.services.dynamodb.DynamoDbAsyncClient;
  * Asynchronous interface for running commands against a DynamoDb database.
  */
 @SdkPublicApi
-public interface DynamoDbEnhancedAsyncClient {
+public interface DynamoDbEnhancedAsyncClient extends DynamoDbEnhancedResource {
 
     /**
      * Returns a mapped table that can be used to execute commands that work with mapped items against that table.
@@ -90,11 +90,23 @@ public interface DynamoDbEnhancedAsyncClient {
     /**
      * The builder definition for a {@link DynamoDbEnhancedAsyncClient}.
      */
-    interface Builder {
-        Builder dynamoDbClient(DynamoDbAsyncClient dynamoDbAsyncClient);
+    interface Builder extends DynamoDbEnhancedResource.Builder {
+        /**
+         * The regular low-level SDK client to use with the enhanced client.
+         * @param dynamoDbClient an initialized {@link DynamoDbAsyncClient}
+         */
+        Builder dynamoDbClient(DynamoDbAsyncClient dynamoDbClient);
 
-        Builder extendWith(DynamoDbEnhancedClientExtension dynamoDbEnhancedClientExtension);
+        @Override
+        Builder extensions(DynamoDbEnhancedClientExtension... dynamoDbEnhancedClientExtensions);
 
+        @Override
+        Builder extensions(List<DynamoDbEnhancedClientExtension> dynamoDbEnhancedClientExtensions);
+
+        /**
+         * Builds an enhanced client based on the settings supplied to this builder
+         * @return An initialized {@link DynamoDbEnhancedAsyncClient}
+         */
         DynamoDbEnhancedAsyncClient build();
     }
 }
--- a/services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/DynamoDbEnhancedClient.java
+++ b/services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/DynamoDbEnhancedClient.java
@@ -34,7 +34,7 @@ import software.amazon.awssdk.services.dynamodb.DynamoDbClient;
  * Synchronous interface for running commands against a DynamoDb database.
  */
 @SdkPublicApi
-public interface DynamoDbEnhancedClient {
+public interface DynamoDbEnhancedClient extends DynamoDbEnhancedResource {
 
     /**
      * Returns a mapped table that can be used to execute commands that work with mapped items against that table.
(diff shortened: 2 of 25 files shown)


services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/client/ExtensionResolver.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    /**
+     * Static provider for the default extensions that are bundled with the DynamoDB Enhanced Client. Currently this is
+     * just the {@link software.amazon.awssdk.enhanced.dynamodb.extensions.VersionedRecordExtension}.
+     *
+     * These extensions will be used by default unless overridden in the enhanced client builder.
+     */
+    public static List<DynamoDbEnhancedClientExtension> defaultExtensions() {
+        return DEFAULT_EXTENSIONS;
```
zoewangg: unmodifiable list?

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/client/ExtensionResolver.java
```diff
@@ -0,0 +1,69 @@
...
+     * Resolves a list of extensions into a single extension. If the list is a singleton, will just return that extension
+     * otherwise it will combine them with the {@link software.amazon.awssdk.enhanced.dynamodb.internal.extensions.ChainExtension}
+     * meta-extension using the order provided in the list.
+     *
+     * @param extensions A list of extensions to be combined in strict order
+     * @return A single extension that combines all the supplied extensions or null if no extensions were provided
+     */
+    public static DynamoDbEnhancedClientExtension resolveExtensions(List<DynamoDbEnhancedClientExtension> extensions) {
+        if (extensions == null || extensions.isEmpty()) {
+            return null;
```
zoewangg: Where do we do null check for `DynamoDbEnhancedClientExtension`? 

Should we use Optional?

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/DynamoDbEnhancedAsyncClient.java
```diff
@@ -91,10 +91,38 @@
...
     interface Builder {
-        Builder dynamoDbClient(DynamoDbAsyncClient dynamoDbAsyncClient);
-
-        Builder extendWith(DynamoDbEnhancedClientExtension dynamoDbEnhancedClientExtension);
-
+        /**
+         * The regular low-level SDK client to use with the enhanced client.
+         * @param dynamoDbClient an initialized {@link DynamoDbAsyncClient}
+         */
+        Builder dynamoDbClient(DynamoDbAsyncClient dynamoDbClient);
```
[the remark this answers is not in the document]
bmaizels: Yeah ok. I was thinking about it as I was copy and pasting all of it, it's right on the threshold for me. Excessive hierarchy can interfere with readability, but I'll try it out and see how it looks.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/client/ExtensionResolver.java
bmaizels: It's already immutable.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/client/ExtensionResolver.java
bmaizels: Null is valid all through the code and is treated as 'no extension'. I could change this interface to optional though but i'd be unwrapping it to null on the client.

zoewangg: I see. I'm not against `return null` as long as we make sure null-check is in place somewhere.

bmaizels: There are multiple tests that exercise this code path. There's no immediate validation that can take place because null is valid. The requirement is that all operations can handle it correctly, and there are tests that verify this.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/client/ExtensionResolver.java
bmaizels: As discussed I'll add a test to assert this and leave it at that.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/DynamoDbEnhancedClient.java
```diff
@@ -88,11 +88,29 @@ static Builder builder() {
...
+        /**
+         * The regular low-level SDK client to use with the enhanced client.
+         * @param dynamoDbClient an initialized {@link DynamoDbClient}
+         */
         Builder dynamoDbClient(DynamoDbClient dynamoDbClient);
 
-        Builder extendWith(DynamoDbEnhancedClientExtension dynamoDbEnhancedClientExtension);
-
+        /**
+         * {@inheritDoc}
```
zoewangg: nit: `{@inheritDoc}` is not necessay. `@Override` automatically renders the inherited doc.

bmaizels: oh! I'll whip that out then. Thanks.

PR #1678, opened 2020-02-28
Add support for maven lambda archetype

zoewangg:
## Description
Add support for maven lambda archetype

## Testing
Added unit tests

## Types of changes

## Checklist

## License


archetypes/archetype-lambda/src/main/resources/archetype-resources/pom.xml
```diff
@@ -0,0 +1,110 @@
...
+    <version>${version}</version>
+    <packaging>jar</packaging>
+    <properties>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <maven.compiler.source>1.8</maven.compiler.source>
+        <maven.compiler.target>1.8</maven.compiler.target>
+        <maven.shade.plugin.version>3.1.1</maven.shade.plugin.version>
+        <maven.compiler.plugin.version>3.6.1</maven.compiler.plugin.version>
+        <exec-maven-plugin.version>1.6.0</exec-maven-plugin.version>
+        <aws.java.sdk.version>2.10.66</aws.java.sdk.version>
```
bmaizels: How do we pin this to the most recent version?

archetypes/archetype-lambda/src/main/resources/archetype-resources/template.yaml
```diff
@@ -0,0 +1,11 @@
+AWSTemplateFormatVersion: '2010-09-09'
+Transform: AWS::Serverless-2016-10-31
+Resources:
+  ${handlerClassName}Function:
+    Type: AWS::Serverless::Function
+    Properties:
+      Runtime: java8
+      Handler: ${package}.${handlerClassName}::handleRequest
+      Timeout: 60
+      MemorySize: 512
```
bmaizels: Maybe we want comments here so people know that they could/should tweak based on their needs (assuming we never make this parameters of the archtype)

archetypes/archetype-lambda/src/main/resources/global.vm
```diff
@@ -0,0 +1,26 @@
+## global variables used by the project
+#set( $symbol_pound = '#' )
+#set( $symbol_dollar = '$' )
+#set( $symbol_escape = '\' )
+## TODO: find a better way to handle this
+#if( $service == 'dynamodb')
+    #set ( $service = 'dynamoDb')
+#end
```
bmaizels: Ouch. Is this the only one?

archetypes/archetype-lambda/src/main/resources/META-INF/maven/archetype-metadata.xml
```diff
@@ -0,0 +1,40 @@
...
+        <include>template.yaml</include>
+      </includes>
+    </fileSet>
+  </fileSets>
+  <requiredProperties>
+    <requiredProperty key="handlerClassName">
+      <defaultValue>App</defaultValue>
+    </requiredProperty>
+    <requiredProperty key="service">
+    </requiredProperty>
```
bmaizels: Ideally we'd have validation here, what if they typo the service?

archetypes/archetype-lambda/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Parameters
+      
+Parameter Name | Default Value | Description
+---|---|---
+`service` (required) | n/a | Specifies service client to be used in the lambda function. You can find the  eg: s3, dynamodb
+`region` (required) | n/a | Specifies region to be set for the SDK client in the application
+`groupId`(required) | n/a | Specifies the group ID of the project
+`artifactId`(required) | n/a | Specifies the artifact ID of the project
+`httpClient` | url-connection-client | Specifies which http client to be used by the SDK client, available options are `url-connection-client`, `apache-client`, `netty-nio-client`
```
bmaizels: Might be worth mentioning sync vs async here

archetypes/archetype-lambda/src/main/resources/archetype-resources/pom.xml
zoewangg: Yeah, I should use `${awsjavasdk.version}`. I hardcoded the version for testing and forgot to change it back.

archetypes/archetype-lambda/src/main/resources/archetype-resources/template.yaml
zoewangg: +1

archetypes/archetype-lambda/src/main/resources/global.vm
zoewangg: No, this would not be the only one. I'll think about a better way to map serviceId -> serviceClientName

archetypes/archetype-lambda/src/main/resources/META-INF/maven/archetype-metadata.xml
zoewangg: That's a limitation for archetype; they only have regex validation, which is not sufficient. We can probably make it more clear on the documentation that this should be the serviceId.

archetypes/archetype-lambda/README.md
zoewangg: Sure

archetypes/archetype-lambda/src/main/resources/global.vm
zoewangg: Can't think of an easy way to do it. Currently, the service module name is lower case derived from serviceId, and the client class name is camel case derived from service name.

We can probably write some scripts to search service-2.json files to map serviceId to serviceName, but that seems a bit overkill. I'm leaning towards just fixing the popular services here.

archetypes/archetype-lambda/README.md
```diff
@@ -0,0 +1,65 @@
...
+    -Dservice=s3  \
+    -Dregion=us-west-2 \
+    -DinteractiveMode=false \
+```
+
+### Parameters
+      
+Parameter Name | Default Value | Description
+---|---|---
+`service` (required) | n/a | Specifies the service client to be used in the lambda function, eg: s3, dynamodb. You can find available services [here][java-sdk-v2-services].
```
millems: Can they just specify the module name (s3) *and* client name (S3Client) or something, so that we don't need to maintain a mapping?

Can we also generate instructions for adding more services? Maybe in a comment or something?

archetypes/archetype-lambda/src/main/resources/archetype-resources/src/main/java/DependencyFactory.java
```diff
@@ -0,0 +1,27 @@
...
+ * The module containing all dependencies required by the {@link ${handlerClassName}}.
+ */
+public class DependencyFactory {
+
+    private DependencyFactory() {}
+
+    /**
+     * @return an instance of ${serviceClientClassName}
+     */
+    public static ${serviceClientClassName} ${service}Client() {
```
millems: In the future, we should consider showing them how to do true DI with dagger.

archetypes/archetype-lambda/src/main/resources/global.vm
```diff
@@ -0,0 +1,29 @@
+## global variables used by the project
+#set( $symbol_pound = '#' )
+#set( $symbol_dollar = '$' )
+#set( $symbol_escape = '\' )
+## TODO: there is no easy way to map service module artifactId to the client name derived the service name
+## As a workaround: mapping the common service here
+#if( $service == 'dynamodb')
+    #set ( $service = 'dynamoDb')
+#end
```
millems: If we want to invest time, we could generate these mappings and upload them to maven for the archetype to use.

archetypes/archetype-lambda/src/test/resources/projects/nettyclient/reference/src/main/java/software/amazonaws/test/DependencyFactory.java
```diff
@@ -0,0 +1,26 @@
...
+    private DependencyFactory() {}
+
+    /**
+     * @return an instance of KinesisAsyncClient
+     */
+    public static KinesisAsyncClient kinesisClient() {
+        return KinesisAsyncClient.builder()
+                       .credentialsProvider(EnvironmentVariableCredentialsProvider.create())
+                       .region(Region.US_EAST_1)
+                       .httpClientBuilder(NettyNioAsyncHttpClient.builder())
```
millems: Do we include open SSL or other tweaks for netty?

archetypes/archetype-lambda/src/main/resources/archetype-resources/src/main/java/DependencyFactory.java
zoewangg: Yup. I thought about adding another parameter, something like `di`, which allows customers to switch between relatively lightweight DI such as `dagger` or `guice` or no DI at all.

archetypes/archetype-lambda/src/test/resources/projects/nettyclient/reference/src/main/java/software/amazonaws/test/DependencyFactory.java
zoewangg: That would require another openssl dependency `netty-tcnative` though, and not all platforms support it. We can probably add a comment here to recommend using openssl if possible.

millems: But since it's on Lambda, we know the platform, right?

archetypes/archetype-lambda/README.md
zoewangg: They'd need to know the exact sync/async client names and I feel like that's kind of against the purpose of quickly bootstrapping a project from a template. 

Yeah, I can add some comments to explain how to add more services in the code.

archetypes/archetype-lambda/src/test/resources/projects/nettyclient/reference/src/main/java/software/amazonaws/test/DependencyFactory.java
zoewangg: lol, good point. I'll give it a try

archetypes/archetype-lambda/README.md
millems: Sure, we can always write a build-tool to generate them pretty easily in the future.

Could we do a quick bash script to generate them for the currently-launched services? That shouldn't be too hard, if we're just hunting for things that extend SdkClient in each service module.

archetypes/archetype-lambda/src/test/resources/projects/nettyclient/reference/src/main/java/software/amazonaws/test/DependencyFactory.java
zoewangg: Wow, the `netty-tcnative-boringssl-static` binding jar itself is 3MB...Certainly wouldn't help with cold start issue, but I guess it would improve overall latency a bit.

archetypes/archetype-lambda/README.md
zoewangg: Yeah, generated `serviceMapping.vm `  file that contains currently-launched services. I'll add the script in another PR to make it update the file automatically after every release.

millems: What if we added a build script to do it *during* the release, instead of having to do it after the release?

zoewangg: Yeah, I misspoke. We can update it during the release like how we update the changelog.

millems: I misspoke too, I meant: can we write a build script, that as part of a mvn clean install generates the serviceMapping.vm used in this archetype?

millems: Like, a maven plugin or build tool that does it as part of the build.

zoewangg: Yeah, I thought about it, but it'll take a really long time to build this specific module because it technically depends on all services. Release is probably fine because it builds all modules anyway but for development, if you are making changes in this module,  you'll have to wait for a long time for it to build.

zoewangg: On second thought, it'll probably be easier to maintain if we keep the script within this module and build the file during build time.

archetypes/archetype-lambda/src/test/resources/projects/nettyclient/reference/src/main/java/software/amazonaws/test/DependencyFactory.java
zoewangg: Ended up just adding `netty-tcnative-boringssl-static` dependency so that it'd use openssl if supported and fallback to jdk ssl provider otherwise.

archetypes/archetype-lambda/src/main/resources/map-service-to-client-prefix
```diff
@@ -0,0 +1,50 @@
...
+    if not os.path.exists(target):
+        os.mkdir(target)
+
+    filename = os.path.join(target, MAPPING_FILE_NAME)
+
+    with open(filename, 'w') as f:
+        f.write('#set ( $serviceMapping =')
+        f.write(json.dumps(service_mapping))
+        f.write(')')
+    return filename
```
millems: How does this work for modules with multiple clients, e.g. dynamodb?

zoewangg: It only works with the top level client in those kind of modules, i.e., `dynamodb` and `waf`. :( 

I think it should be fine since there are only two broken clients `dynamodbstream` and `wafregional` and we can easily customize it.

We did specify in the `README` that the available service option can be found in https://github.com/aws/aws-sdk-java-v2/tree/master/services, so technically, `wafregional` and `dynamodbstream` are just not supported.

millems: Sounds fine, as long as dynamodb doesn't generate a client for dynamodbstreams.

zoewangg: Added customization for `dynamodbstreams` and `wafregional` and they should work now.

PR #1685, opened 2020-03-05
Fix merging of single value headers

dagnir:
## Description
Use the last seen HTTP/1.1 header value for headers defined to only appear once in an HTTP message instead of merging them all into a list. The order in which header values are inspected is: headers set on the request, overridden headers set on the client, then finally overridden headers set on an SDK request. See https://tools.ietf.org/html/rfc2616#section-4.2 for more information.

## Motivation and Context
Bug fix.

## Testing
`mvn clean install`. Will run integ tests.

## Types of changes

## Checklist

## License

### DIFF
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/MergeCustomHeadersStage.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/MergeCustomHeadersStage.java
@@ -27,6 +27,7 @@ import software.amazon.awssdk.core.internal.http.HttpClientDependencies;
 import software.amazon.awssdk.core.internal.http.RequestExecutionContext;
 import software.amazon.awssdk.core.internal.http.pipeline.MutableRequestToRequestPipeline;
 import software.amazon.awssdk.http.SdkHttpFullRequest;
+import software.amazon.awssdk.utils.http.SdkHttpUtils;
 
 /**
  * Merge customer supplied headers into the marshalled request.
@@ -52,7 +53,13 @@ public class MergeCustomHeadersStage implements MutableRequestToRequestPipeline 
     private final Map<String, List<String>> mergeHeaders(Map<String, List<String>>... headers) {
         Map<String, List<String>> result = new LinkedHashMap<>();
         for (Map<String, List<String>> header : headers) {
-            header.forEach((k, v) -> result.computeIfAbsent(k, ignored -> new ArrayList<>()).addAll(v));
+            header.forEach((headerName, headerValues) -> {
+                List<String> resultHeaderValues = result.computeIfAbsent(headerName, ignored -> new ArrayList<>());
+                if (SdkHttpUtils.isSingleHeader(headerName)) {
+                    resultHeaderValues.clear();
+                }
+                resultHeaderValues.addAll(headerValues);
+            });
         }
         return result;
     }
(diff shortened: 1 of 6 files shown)


utils/src/main/java/software/amazon/awssdk/utils/http/SdkHttpUtils.java
```diff
@@ -297,4 +307,12 @@ public static String appendUri(String baseUri, String path) {
     public static Optional<String> firstMatchingHeader(Map<String, List<String>> headers, String header) {
         return allMatchingHeaders(headers, header).findFirst();
     }
+
+    public static Set<String> singleHeaders() {
```
spfink: Is this needed?

dagnir: Not really needed, I just needed to able access the list from the tests, but I can just copy them there.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/MergeCustomHeadersStage.java
```diff
@@ -52,7 +54,15 @@ public MergeCustomHeadersStage(HttpClientDependencies dependencies) {
     private final Map<String, List<String>> mergeHeaders(Map<String, List<String>>... headers) {
         Map<String, List<String>> result = new LinkedHashMap<>();
         for (Map<String, List<String>> header : headers) {
-            header.forEach((k, v) -> result.computeIfAbsent(k, ignored -> new ArrayList<>()).addAll(v));
```
millems: We're now introducing an implicit 'priority' in the sense that the last header 'wins'. Can we add a comment about that?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/MergeCustomHeadersStage.java
```diff
@@ -52,7 +54,15 @@ public MergeCustomHeadersStage(HttpClientDependencies dependencies) {
...
+            header.forEach((headerName, headerValues) -> {
+                List<String> resultHeaderValues = result.computeIfAbsent(headerName, ignored -> new ArrayList<>());
+                if (SdkHttpUtils.isSingleHeader(headerName)) {
+                    resultHeaderValues.clear();
+                    resultHeaderValues.add(CollectionUtils.lastElement(headerValues));
```
millems: Should we disallow (or log) when there are multiple values specified in the same headerValues for a single-header?

It's a bit weird if I add multiple content-types at the request level, but it just takes the last one.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/MergeCustomHeadersStage.java
dagnir: Sure, i'll add a note on the setters in the Client/Request override configuration.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/MergeCustomHeadersStage.java
dagnir: I went back and forth with this. Alternatively what do you think about keeping all the values in the current list, but not merging them with the previously seen list?

e.g.

"FooHeader" set by marshaller -> "a"
"FooHeader" override on client -> ["b", "c"]
"FooHeader" override on request -> ["d", "e", "f"]

"FooHeader" on the wire -> ["d", "e", "f"]

dagnir: After discussing offline, went with the above solution.

PR #1695, opened 2020-03-09
DynamoDB Enhanced: Added static constructors to TableSchema interface

bmaizels:
Making it easier for customers to discover the different implementations of TableSchema.

## License


bmaizels:
> 
> 
> If fromXXX is consistent, LGTM

I don't think we have an established pattern for this. The alternative that was a contender for me was 'createFromBean' but I felt that was unnecessarily verbose. 'of' is a pattern we've used before, but bean needs to be in there somewhere so 'ofBean' could have worked but I liked it less than 'fromBean'

PR #1697, opened 2020-03-10
Create wrapper class for ddb enhanced extension input parameters

zoewangg:
## Description
Create wrapper class for ddb enhanced extension input parameters

## Testing
Updated tests

## Types of changes

## Checklist

## License

### DIFF
--- a/services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/DynamoDbExtensionContext.java
+++ b/services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/DynamoDbExtensionContext.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.enhanced.dynamodb;
+
+import java.util.Map;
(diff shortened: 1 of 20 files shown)


services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/DynamoDbExtensionContext.java
```diff
@@ -0,0 +1,63 @@
...
+import java.util.Map;
+import software.amazon.awssdk.annotations.SdkProtectedApi;
+import software.amazon.awssdk.annotations.SdkPublicApi;
+import software.amazon.awssdk.enhanced.dynamodb.internal.operations.OperationContext;
+import software.amazon.awssdk.services.dynamodb.model.AttributeValue;
+
+/**
+ * A wrapper for the immutable context objects that are visible to the {@link DynamoDbEnhancedClientExtension}s.
+ */
+@SdkProtectedApi
```
bmaizels: I've been avoiding using 'protected' in this library because I thought it doesn't apply since none of it is code-genned. I think this should just be public?

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/EnhancedClientUtils.java
```diff
@@ -68,10 +69,12 @@ public static String cleanAttributeName(String key) {
         }
 
         if (dynamoDbEnhancedClientExtension != null) {
-            ReadModification readModification = dynamoDbEnhancedClientExtension.afterRead(itemMap,
-                                                                                          operationContext,
-                                                                                          tableSchema.tableMetadata());
-
+            ReadModification readModification = dynamoDbEnhancedClientExtension.afterRead(
+                DefaultDynamoDbExtensionContext.builder()
```
bmaizels: I think if we're going to go public on this interface we should either implement it in public (make it a class and not an interface that includes a builder), or give them a static builder, in which case the Builder needs to go on the public interface too.

bmaizels: as discussed people dont need to create these, just consume them, so it's fine I think

PR #1716, opened 2020-03-13
DynamoDB Enhanced: Remove the need to pass in model class for operati…

bmaizels:
…ons that take typed consumer builders

## License


bmaizels:
> 
> 
> Got to update the javadoc for the interfaces.

They looked correct to me, meaning they were actually incorrect before I made this change but are now correct. Do you see any specifically that aren't?

PR #1717, opened 2020-03-14
Use signing date when deriving signing key

dagnir:
## Description
Fix an issue where the signing key is created only once at the start of the
request for event streaming requests. This causes requests that span two more
more days to have signing errors once the date changes because the signing key
was derived only once using the date at the beginning of the request.

## Motivation and Context
Bug fix.

## Testing
 - `mvn clean install`
 - ~~Running integ tests~~ integ test

## Types of changes

## Checklist

## License

### DIFF
--- a/core/auth/src/test/java/software/amazon/awssdk/auth/signer/internal/SignerKeyTest.java
+++ b/core/auth/src/test/java/software/amazon/awssdk/auth/signer/internal/SignerKeyTest.java
@@ -1,0 +1,50 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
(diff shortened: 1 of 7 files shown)


core/auth/src/test/java/software/amazon/awssdk/auth/signer/internal/SignerKeyTest.java
```diff
@@ -0,0 +1,50 @@
...
+package software.amazon.awssdk.auth.signer.internal;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import java.time.Instant;
+import org.junit.Test;
+
+public class SignerKeyTest {
+
+    @Test
+    public void isValidForDate_dayBefore() {
```
cenedhryn: Shouldn't it be isNotValid? and for the after method too

dagnir: I normally like to name the tests to include the name of the method I'm testing, which is `isValidForDate()` in this case :)

cenedhryn: I like to include the expected behavior to be sure I understand the test's intention.

dagnir: That's reasonable. Will update

PR #1719, opened 2020-03-16
DDB Enhanced: Add support for a 'key item' in operations that take a key

bmaizels:
Adding the new KeyItem/Item shortcut form to TransactGetItems and TransactWriteItems enhanced requests caused ambiguity resolving the difference between that overloaded method and the method that takes a consumer builder. We decided that keyItem was the more convenient of the two convenience methods so chopped the consumer builder version just from those operations.

## License

### DIFF
--- a/services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/DynamoDbAsyncTable.java
+++ b/services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/DynamoDbAsyncTable.java
@@ -152,12 +152,12 @@ public interface DynamoDbAsyncTable<T> extends MappedTableResource<T> {
      * <pre>
      * {@code
      *
-     * mappedTable.delete(DeleteItemEnhancedRequest.builder().key(key).build()).join();
+     * MyItem previouslyPersistedItem = mappedTable.delete(DeleteItemEnhancedRequest.builder().key(key).build()).join();
      * }
      * </pre>
      *
      * @param request A {@link DeleteItemEnhancedRequest} with key and optional directives for deleting an item from the table.
-     * @return a {@link CompletableFuture} of the deleted item
+     * @return a {@link CompletableFuture} of the item that was persisted in the database before it was deleted.
      */
     default CompletableFuture<T> deleteItem(DeleteItemEnhancedRequest request) {
         throw new UnsupportedOperationException();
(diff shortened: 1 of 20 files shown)


services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/DynamoDbAsyncTable.java
```diff
@@ -212,6 +212,28 @@
...
+     * Deletes a single item from the mapped table using just the key of a supplied modelled 'key item' object.
+     * <p>
+     * This operation calls the low-level DynamoDB API DeleteItem operation. Consult the DeleteItem documentation for
+     * further details and constraints.
+     * <p>
+     * Example:
+     * <pre>
+     * {@code
+     *
+     * mappedTable.deleteItem(keyItem);
```
cenedhryn: This example is missing  `.join()`

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/DynamoDbAsyncTable.java
```diff
@@ -212,6 +212,28 @@
         throw new UnsupportedOperationException();
     }
 
+    /**
+     * Deletes a single item from the mapped table using just the key of a supplied modelled 'key item' object.
```
cenedhryn: Should we emphasize that the item you're returning isn't necessarily the one you send in?

services-custom/dynamodb-enhanced/src/test/java/software/amazon/awssdk/enhanced/dynamodb/internal/operations/BatchWriteItemOperationTest.java
```diff
@@ -179,6 +179,39 @@ public void generateRequest_multipleTables_mixedCommands_usingShortcutForm() {
...
     }
 
+    @Test
+    public void generateRequest_multipleTables_mixedCommands_usingKeyItemForm() {
+        BatchWriteItemEnhancedRequest batchWriteItemEnhancedRequest =
+            BatchWriteItemEnhancedRequest.builder()
+                                         .writeBatches(
+                                             WriteBatch.builder(FakeItem.class)
+                                                       .mappedTableResource(fakeItemMappedTable)
+                                                       .addPutItem(FAKE_ITEMS.get(0))
```
cenedhryn: The overall effect when you mix-n-match using the new format is one of pleasing uniformity, I think it enhances the usability.

PR #1721, opened 2020-03-17
Log signed event for easier debugging

dagnir:
## Description
In case of future bugs related to event stream signing, this should help in
debuggging to see the contents of the message in an easier to read format.

## Motivation and Context
Nice to have for future debugging.

## Testing
`mvn clean install`

## Types of changes

## Checklist

## License

### DIFF
--- a/core/auth/src/main/java/software/amazon/awssdk/auth/signer/internal/BaseEventStreamAsyncAws4Signer.java
+++ b/core/auth/src/main/java/software/amazon/awssdk/auth/signer/internal/BaseEventStreamAsyncAws4Signer.java
@@ -20,7 +20,9 @@ import static software.amazon.awssdk.auth.signer.internal.SignerConstant.X_AMZ_C
 import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
 import java.time.Instant;
+import java.util.Arrays;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.Map;
 import java.util.Optional;
 import java.util.TreeMap;
@@ -143,7 +145,6 @@ public abstract class BaseEventStreamAsyncAws4Signer extends BaseAsyncAws4Signer
                 /**
                  * Calculate rolling signature
                  */
-
                 byte[] payload = byteBuffer.array();
                 byte[] signatureBytes = signEventStream(priorSignature, signingKey, signingInstant, requestParams,
                                                         nonSignatureHeaders, payload);
@@ -160,6 +161,13 @@ public abstract class BaseEventStreamAsyncAws4Signer extends BaseAsyncAws4Signer
                  * Encode signed event to byte
                  */
                 Message signedMessage = new Message(sortHeaders(headers), payload);
+
+                if (LOG.isLoggingLevelEnabled("trace")) {
+                    LOG.trace(() -> "Signed message: " + toDebugString(signedMessage, false));
+                } else {
+                    LOG.debug(() -> "Signed message: " + toDebugString(signedMessage, true));
+                }
+
                 return signedMessage.toByteBuffer();
             }
         };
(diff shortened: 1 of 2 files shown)


core/auth/src/main/java/software/amazon/awssdk/auth/signer/internal/BaseEventStreamAsyncAws4Signer.java
```diff
@@ -160,6 +161,13 @@ public ByteBuffer apply(ByteBuffer byteBuffer) {
                  * Encode signed event to byte
                  */
                 Message signedMessage = new Message(sortHeaders(headers), payload);
+
+                if (LOG.isLoggingLevelEnabled("trace")) {
```
bmaizels: Log4j2 docs suggest we could use logger.isTraceEnabled() here? Might be slightly better than passing a string

bmaizels: another approach is to use LOG.printf(.... which allows you to dynamically set the log level which means that theoretically you could do all this in a single line without having an if statement.

core/auth/src/main/java/software/amazon/awssdk/auth/signer/internal/BaseEventStreamAsyncAws4Signer.java
```diff
@@ -259,4 +267,35 @@ public void subscribe(Subscriber<? super ByteBuffer> s) {
...
         }
     }
+
+    private static String toDebugString(Message m, boolean truncatePayload) {
+        StringBuilder sb = new StringBuilder("Message = {headers={");
+        Map<String, HeaderValue> headers = m.getHeaders();
+
+        Iterator<Map.Entry<String, HeaderValue>> headersIter = headers.entrySet().iterator();
+
+        while (headersIter.hasNext()) {
```
bmaizels: why not do this with some kind of stream thing?

core/auth/src/main/java/software/amazon/awssdk/auth/signer/internal/BaseEventStreamAsyncAws4Signer.java
```diff
@@ -259,4 +267,35 @@ public void subscribe(Subscriber<? super ByteBuffer> s) {
...
+        }
+
+        sb.append("}, payload=");
+
+        byte[] payload = m.getPayload();
+
+        byte[] payloadToLog;
+        if (truncatePayload) {
+            // Would be nice if BinaryUtils.toHex() could take an array index range instead so we don't need to copy
+            payloadToLog = Arrays.copyOf(payload, Math.min(32, payload.length));
```
bmaizels: Could payload.length be 0? what happens then?

core/auth/src/main/java/software/amazon/awssdk/auth/signer/internal/BaseEventStreamAsyncAws4Signer.java
```diff
@@ -259,4 +267,35 @@ public void subscribe(Subscriber<? super ByteBuffer> s) {
             return transformedRequestBody.contentLength();
         }
     }
+
+    private static String toDebugString(Message m, boolean truncatePayload) {
```
bmaizels: I know we don't usually write unit tests for logs but maybe this method is complex enough to justify one?

core/auth/src/main/java/software/amazon/awssdk/auth/signer/internal/BaseEventStreamAsyncAws4Signer.java
dagnir: This is our [`software.amazon.awssdk.utils.Logger`](https://github.com/aws/aws-sdk-java-v2/blob/b349c1ba93584c6489f717ef2ba7c2998a6936d4/utils/src/main/java/software/amazon/awssdk/utils/Logger.java#L25) not Log4j's

core/auth/src/main/java/software/amazon/awssdk/auth/signer/internal/BaseEventStreamAsyncAws4Signer.java
dagnir: I found it easier to use this while working with the string builder. I can change it if it's a blocker for you

core/auth/src/main/java/software/amazon/awssdk/auth/signer/internal/BaseEventStreamAsyncAws4Signer.java
dagnir: Sure will add one

core/auth/src/main/java/software/amazon/awssdk/auth/signer/internal/BaseEventStreamAsyncAws4Signer.java
bmaizels: Ahh ok nevermind then. I assume we don't have a similar method then?

core/auth/src/main/java/software/amazon/awssdk/auth/signer/internal/BaseEventStreamAsyncAws4Signer.java
bmaizels: No not a blocker, just a nit. You can ignore.

core/auth/src/main/java/software/amazon/awssdk/auth/signer/internal/BaseEventStreamAsyncAws4Signer.java
dagnir: This should be fine, it will generate an empty string

core/auth/src/main/java/software/amazon/awssdk/auth/signer/internal/BaseEventStreamAsyncAws4Signer.java
dagnir: Not currently

PR #1727, opened 2020-03-19
DDB Enhanced client: flattened batch get items

zoewangg:
## Description
- sync batchGetItem returns`BatchGetItemPageIterable` that allows customers to iterate through flattened results via `BatchGetItemIterable#resultsForTable`

- async batchGetItem returns `BatchGetItemPagePublisher` that allows customers to subscribe to flattened results via `BatchGetItemPagePublisher#resultsForTable`

- renamed `getResutlsForTable` -> `resultsForTable`

~~- add `unprocessedKeys` in `BatchGetItemPage` so that it's consistent with `Page`~~

## License


services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/model/BatchGetResultPage.java
```diff
@@ -74,6 +75,17 @@ public static Builder builder() {
...
     }
 
+    /**
+     * Returns a map of tables and their respective keys that were not processed with the current response. The UnprocessedKeys
+     * value is in the same form as RequestItems, so the value can be provided directly to a subsequent BatchGetItem operation.
+     *
+     * @return unprocessed keys
+     * @see BatchGetItemResponse#unprocessedKeys()
+     */
+    public Map<String, KeysAndAttributes> unprocessedKeys() {
```
bmaizels: As we discussed we should remove this as there's no use-case for needing it with pagination and flatmapped iteration support.

PR #1731, opened 2020-03-20
DDB Enhanced: updates README with bean examples

cenedhryn:
## Description
Adds sections for the StaticTableSchema and BeanTableSchema, including usage of
- AttributeConverterProviders
- AttributeConverters
- Flattning and extending beans
- Ignoring and including bean attributes

### DIFF
--- a/services-custom/dynamodb-enhanced/README.md
+++ b/services-custom/dynamodb-enhanced/README.md
@@ -159,7 +159,6 @@ index. Here's an example of how to do this:
    PageIterable<Customer> customersWithName = 
        customersByName.query(r -> r.queryConditional(equalTo(k -> k.partitionValue("Smith"))));
    ```
-
 ### Non-blocking asynchronous operations
 If your application requires non-blocking asynchronous calls to
 DynamoDb, then you can use the asynchronous implementation of the
@@ -195,8 +194,7 @@ key differences:
    // Perform other work and let the processor handle the results asynchronously
    ```
 
-
-### Using extensions
+## Using extensions
 The mapper supports plugin extensions to provide enhanced functionality
 beyond the simple primitive mapped operations. Extensions have two hooks, beforeWrite() and
 afterRead(); the former can modify a write operation before it happens,
@@ -221,7 +219,7 @@ DynamoDbEnhancedClient enhancedClient =
                           .build();
 ```
 
-#### VersionedRecordExtension
+### VersionedRecordExtension
 
 This extension is loaded by default and will increment and track a record version number as
 records are written to the database. A condition will be added to every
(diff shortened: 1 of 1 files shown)


services-custom/dynamodb-enhanced/README.md
```diff
@@ -291,32 +296,205 @@ Using composition:
...
+
+    public String getCreatedDate() { return this.createdDate; }
+    public void setCreatedDate(String createdDate) { this.createdDate = createdDate;}
+}
+```
+You can flatten as many different eligible classes as you like using the flatten annotation.
+The only constraints are that attributes must not have the same name when they are being rolled
+together, and there must never be more than one partition key, sort key or table name.
+
+### Control how attributes are converted
```
bmaizels: Same comment as before, I think we should organize by use-case and then a description of how to do it in the BeanTableSchema and then the StaticTableSchema

services-custom/dynamodb-enhanced/README.md
```diff
@@ -291,32 +296,205 @@ Using composition:
...
+    .attributeConverterProviders(converterProvider1, converterProvider2)
+    .build();
+```
+Note that if you supply your own chain of attribute converter providers, you will override
+the default converter provider and must therefore include it in the chain if you wish to
+use its attribute converters.
+
+#### Override the mapping of a single attribute
+Supply an AttributeConverter when creating the attribute to directly override any
+converters provided by the table schema AttributeConverterProviders:
```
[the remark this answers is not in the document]
cenedhryn: ok

services-custom/dynamodb-enhanced/README.md
```diff
@@ -291,32 +296,205 @@ Using composition:
...
+using the @DynamoDbAttribute annotation:
+```java
+private String internalKey;
+
+@DynamoDbAttribute("renamedInternalKey")
+public String getInternalKey() { return this.internalKey; }
+public void setInternalKey(String internalKey) { return this.internalKey = internalKey;}
+```
+
+### Flat map attributes from another class
```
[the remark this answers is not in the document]
cenedhryn: Yes, I rearranged it and its clearer, consistent and less repetitive.

services-custom/dynamodb-enhanced/README.md
```diff
@@ -248,14 +246,143 @@ Or using a StaticTableSchema:
...
+public class Customer {
+
+}
+```
+
+Chain of converter providers ending with the default (least priority):
+```java
+@DynamoDbBean(converterProviders = {
+   converterProvider1.class, 
+   converterProvider2.class,
```
bmaizels: Capital C?

services-custom/dynamodb-enhanced/README.md
```diff
@@ -270,53 +397,96 @@ public abstract class GenericRecord {
...
+    public String getId() { return this.id; }
+    public void setId(String id) { this.id = id;}
+
+    public String getCreatedDate() { return this.createdDate; }
+    public void setCreatedDate(String createdDate) { this.createdDate = createdDate;}
+}
+```
+You can flatten as many different eligible classes as you like using the flatten annotation.
+The only constraints are that attributes must not have the same name when they are being rolled
+together, and there must never be more than one partition key, sort key or table name.
```
bmaizels: I think we should move this text underneath the StaticTableSchema example.

services-custom/dynamodb-enhanced/README.md
cenedhryn: :-)

PR #1733, opened 2020-03-23
Allow DefaultS3Presigner.Builder to take a custom S3Configuration

scheerer:
## Description
Adds an additional builder field to `DefaultS3Presigner` to allow overriding the default `S3Configuration` being used.

## Motivation and Context
Currently it is impossible to leverage S3 accelerated endpoints with the `S3Presigner`. This is due to the fact that there is no way to override the S3Configuration being used by the `DefaultS3Presigner`.

Normally this would be accomplished by setting `.accelerateModeEnabled(true)` on an `S3Configuration` object and in turn using that configuration when building S3 endpoints.

In order to accomplish this today one must completely duplicate the `DefaultS3Presigner` themselves in their own project and customize their copy like I've done in this pull request.

This change allows the developer to easily provide a custom `S3Configuration` using the provided `Builder` already in place.

## Testing

I added a unit test to ensure the intended usage works as desired.

## Types of changes

## Checklist

## License


services/s3/src/main/java/software/amazon/awssdk/services/s3/presigner/S3Presigner.java
```diff
@@ -512,6 +513,8 @@ default PresignedAbortMultipartUploadRequest presignAbortMultipartUpload(
     @SdkPublicApi
     @NotThreadSafe
     interface Builder extends SdkPresigner.Builder {
+        Builder s3Configuration(S3Configuration s3Configuration);
```
millems: Can we call this `serviceConfiguration` to match `S3BaseClientBuilder#serviceConfiguration`?

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/presigner/DefaultS3Presigner.java
```diff
@@ -98,9 +98,13 @@
 @SdkInternalApi
 public final class DefaultS3Presigner extends DefaultSdkPresigner implements S3Presigner {
     private static final AwsS3V4Signer DEFAULT_SIGNER = AwsS3V4Signer.create();
+    private static final S3Configuration DEFAULT_S3_CONFIGURATION = S3Configuration.builder()
+            .checksumValidationEnabled(false)
+            .build();
```
millems: Why is disabling checksum validation required?

services/s3/src/test/java/software/amazon/awssdk/services/s3/S3PresignerTest.java
```diff
@@ -325,4 +328,23 @@ public void putObject_Sigv4PresignerHonorsSignatureDuration() {
...
+        PresignedGetObjectRequest presignedRequest =
+                presigner.presignGetObject(r -> r.signatureDuration(Duration.ofMinutes(5))
+                        .getObjectRequest(go -> go.bucket("foo34343434")
+                                .key("bar")));
+
+
+        System.out.println(presignedRequest.url());
+
+        assertThat(presignedRequest.httpRequest().host()).contains(".s3-accelerate.");
+    }
```
millems: Can we test all possible flags of S3Configuration?

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/presigner/DefaultS3Presigner.java
scheerer: This is just carrying forward the existing default this class was using on line 292. To be honest, I'm not real sure why it is set that way to begin with.

services/s3/src/main/java/software/amazon/awssdk/services/s3/presigner/S3Presigner.java
scheerer: Yes, I'll update this

services/s3/src/test/java/software/amazon/awssdk/services/s3/S3PresignerTest.java
```diff
@@ -325,4 +330,94 @@ public void putObject_Sigv4PresignerHonorsSignatureDuration() {
...
+                .accelerateModeEnabled(true)
+                .build())
+                .build();
+
+        PresignedGetObjectRequest presignedRequest =
+                presigner.presignGetObject(r -> r.signatureDuration(Duration.ofMinutes(5))
+                        .getObjectRequest(go -> go.bucket(BUCKET)
+                                .key("bar")));
+
+        assertThat(presignedRequest.httpRequest().host()).isEqualTo(String.format("%s.s3-accelerate.dualstack.amazonaws.com", BUCKET));
```
millems: Nice! Thanks for testing this. 

For the other features:
1. pathStyleAccessEnabled, useArnRegionEnabled both affect the endpoint. If they're working, can we add a test for them? If they're not working, maybe we should throw an exception if they're configured.
2. chunkedEncodingEnabled and checksumValidationEnabled don't make sense for presigned requests right now. Can we add some javadoc to mention that they currently aren't used for presigning?

services/s3/src/main/java/software/amazon/awssdk/services/s3/presigner/S3Presigner.java
```diff
@@ -512,6 +513,8 @@ default PresignedAbortMultipartUploadRequest presignAbortMultipartUpload(
     @SdkPublicApi
     @NotThreadSafe
     interface Builder extends SdkPresigner.Builder {
+        Builder serviceConfiguration(S3Configuration serviceConfiguration);
+
```
millems, line 517: Can we include the javadoc here as well? I see you added it to the concrete implementation already.

services/s3/src/test/java/software/amazon/awssdk/services/s3/S3PresignerTest.java
```diff
@@ -325,4 +329,128 @@ public void putObject_Sigv4PresignerHonorsSignatureDuration() {
...
+
+        PresignedGetObjectRequest presignedRequest =
+                presigner.presignGetObject(r -> r.signatureDuration(Duration.ofMinutes(5))
+                        .getObjectRequest(go -> go.bucket(accessPointArn)
+                                .key("bar")));
+
+        assertThat(presignedRequest.url().toString()).startsWith(customEndpoint);
+    }
+
+    @Test //(expected = IllegalArgumentException.class)
```
millems, line 442: Comment not needed.

scheerer: oops 👍

services/s3/src/main/java/software/amazon/awssdk/services/s3/presigner/S3Presigner.java
scheerer: sure thing!

dagnir:
Looks like this has been merged; closing this.

PR #1756, opened 2020-04-01
DDB Enhanced: Added support for projection expression

musketyr:
## Description
Added missing support for projection expression into `QueryEnhancedRequest` and `ScanEnhancedRequest` classes.

## Motivation and Context
At the moment, there is no way how to obtain just a portion of the data from DynamoDB in OOP manner.

## Testing
The tests have been updated to reflect the change.

## Types of changes

## Checklist

## License

### DIFF
--- a/services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/operations/ScanOperation.java
+++ b/services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/operations/ScanOperation.java
@@ -15,11 +15,19 @@
 
 package software.amazon.awssdk.enhanced.dynamodb.internal.operations;
 
+import static software.amazon.awssdk.enhanced.dynamodb.internal.EnhancedClientUtils.cleanAttributeName;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 import java.util.function.Function;
+import java.util.function.UnaryOperator;
 import software.amazon.awssdk.annotations.SdkInternalApi;
 import software.amazon.awssdk.core.async.SdkPublisher;
 import software.amazon.awssdk.core.pagination.sync.SdkIterable;
 import software.amazon.awssdk.enhanced.dynamodb.DynamoDbEnhancedClientExtension;
+import software.amazon.awssdk.enhanced.dynamodb.Expression;
 import software.amazon.awssdk.enhanced.dynamodb.TableMetadata;
 import software.amazon.awssdk.enhanced.dynamodb.TableSchema;
 import software.amazon.awssdk.enhanced.dynamodb.internal.EnhancedClientUtils;
@@ -27,12 +35,15 @@ import software.amazon.awssdk.enhanced.dynamodb.model.Page;
 import software.amazon.awssdk.enhanced.dynamodb.model.ScanEnhancedRequest;
 import software.amazon.awssdk.services.dynamodb.DynamoDbAsyncClient;
 import software.amazon.awssdk.services.dynamodb.DynamoDbClient;
+import software.amazon.awssdk.services.dynamodb.model.AttributeValue;
 import software.amazon.awssdk.services.dynamodb.model.ScanRequest;
 import software.amazon.awssdk.services.dynamodb.model.ScanResponse;
 
 @SdkInternalApi
 public class ScanOperation<T> implements PaginatedTableOperation<T, ScanRequest, ScanResponse>,
                                          PaginatedIndexOperation<T, ScanRequest, ScanResponse> {
+
+    private static final UnaryOperator<String> PROJECTION_EXPRESSION_KEY_MAPPER = k -> "#AMZN_MAPPED_" + cleanAttributeName(k);
 
     private final ScanEnhancedRequest request;
 
@@ -48,20 +59,42 @@ public class ScanOperation<T> implements PaginatedTableOperation<T, ScanRequest,
     public ScanRequest generateRequest(TableSchema<T> tableSchema,
                                        OperationContext operationContext,
                                        DynamoDbEnhancedClientExtension extension) {
+        Map<String, AttributeValue> expressionValues = null;
+        Map<String, String> expressionNames = null;
+
+        if (this.request.filterExpression() != null) {
+            expressionValues = this.request.filterExpression().expressionValues();
+            expressionNames = this.request.filterExpression().expressionNames();
+        }
+
+        String projectionExpression = null;
+        if (this.request.attributesToProject() != null) {
+            List<String> placeholders = new ArrayList<>();
+            Map<String, String> projectionPlaceholders = new HashMap<>();
+            this.request.attributesToProject().forEach(attr -> {
+                String placeholder = PROJECTION_EXPRESSION_KEY_MAPPER.apply(attr);
+                placeholders.add(placeholder);
+                projectionPlaceholders.put(placeholder, attr);
+            });
+            projectionExpression = String.join(",", placeholders);
+            expressionNames = Expression.joinNames(expressionNames, projectionPlaceholders);
+        }
+
         ScanRequest.Builder scanRequest = ScanRequest.builder()
             .tableName(operationContext.tableName())
             .limit(this.request.limit())
             .exclusiveStartKey(this.request.exclusiveStartKey())
-            .consistentRead(this.request.consistentRead());
+            .consistentRead(this.request.consistentRead())
+            .expressionAttributeValues(expressionValues)
+            .expressionAttributeNames(expressionNames)
+            .projectionExpression(projectionExpression);
 
         if (!TableMetadata.primaryIndexName().equals(operationContext.indexName())) {
             scanRequest = scanRequest.indexName(operationContext.indexName());
         }
 
         if (this.request.filterExpression() != null) {
-            scanRequest = scanRequest.filterExpression(this.request.filterExpression().expression())
-                                     .expressionAttributeValues(this.request.filterExpression().expressionValues())
-                                     .expressionAttributeNames(this.request.filterExpression().expressionNames());
+            scanRequest = scanRequest.filterExpression(this.request.filterExpression().expression());
         }
 
         return scanRequest.build();
--- a/services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/model/QueryEnhancedRequest.java
+++ b/services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/model/QueryEnhancedRequest.java
@@ -15,7 +15,12 @@
 
 package software.amazon.awssdk.enhanced.dynamodb.model;
 
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import software.amazon.awssdk.annotations.SdkPublicApi;
 import software.amazon.awssdk.enhanced.dynamodb.DynamoDbAsyncIndex;
@@ -41,6 +46,7 @@ public final class QueryEnhancedRequest {
     private final Integer limit;
     private final Boolean consistentRead;
     private final Expression filterExpression;
+    private final List<String> attributesToProject;
 
     private QueryEnhancedRequest(Builder builder) {
         this.queryConditional = builder.queryConditional;
@@ -49,6 +55,9 @@ public final class QueryEnhancedRequest {
         this.limit = builder.limit;
         this.consistentRead = builder.consistentRead;
         this.filterExpression = builder.filterExpression;
+        this.attributesToProject = builder.attributesToProject != null
+                ? Collections.unmodifiableList(builder.attributesToProject)
+                : null;
     }
 
     /**
@@ -67,7 +76,8 @@ public final class QueryEnhancedRequest {
                         .scanIndexForward(scanIndexForward)
                         .limit(limit)
                         .consistentRead(consistentRead)
-                        .filterExpression(filterExpression);
+                        .filterExpression(filterExpression)
+                        .attributesToProject(attributesToProject);
     }
 
     /**
@@ -113,6 +123,13 @@ public final class QueryEnhancedRequest {
         return filterExpression;
     }
 
+    /**
+     * Returns the list of projected attributes on this request object, or an null if no projection is specified.
+     */
+    public List<String> attributesToProject() {
+        return attributesToProject;
+    }
+
     @Override
     public boolean equals(Object o) {
         if (this == o) {
@@ -142,6 +159,12 @@ public final class QueryEnhancedRequest {
         if (consistentRead != null ? ! consistentRead.equals(query.consistentRead) : query.consistentRead != null) {
             return false;
         }
+        if (attributesToProject != null
+                ? ! attributesToProject.equals(query.attributesToProject)
+                : query.attributesToProject != null
+        ) {
+            return false;
+        }
         return filterExpression != null ? filterExpression.equals(query.filterExpression) : query.filterExpression == null;
     }
 
@@ -153,6 +176,7 @@ public final class QueryEnhancedRequest {
         result = 31 * result + (limit != null ? limit.hashCode() : 0);
         result = 31 * result + (consistentRead != null ? consistentRead.hashCode() : 0);
         result = 31 * result + (filterExpression != null ? filterExpression.hashCode() : 0);
+        result = 31 * result + (attributesToProject != null ? attributesToProject.hashCode() : 0);
         return result;
     }
 
@@ -168,6 +192,7 @@ public final class QueryEnhancedRequest {
         private Integer limit;
         private Boolean consistentRead;
         private Expression filterExpression;
+        private List<String> attributesToProject;
 
         private Builder() {
         }
@@ -252,6 +277,73 @@ public final class QueryEnhancedRequest {
          */
         public Builder filterExpression(Expression filterExpression) {
             this.filterExpression = filterExpression;
+            return this;
+        }
+
+        /**
+         * <p>
+         * Sets a collection of the attribute names to be retrieved from the database. These attributes can include
+         * scalars, sets, or elements of a JSON document.
+         * </p>
+         * <p>
+         * If no attribute names are specified, then all attributes will be returned. If any of the requested attributes
+         * are not found, they will not appear in the result.
+         * </p>
+         * <p>
+         * For more information, see <a href=
+         * "https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Expressions.AccessingItemAttributes.html"
+         * >Accessing Item Attributes</a> in the <i>Amazon DynamoDB Developer Guide</i>.
+         * </p>
+         * @param attributesToProject
+         *        A collection of the attributes names to be retrieved from the database.
+         * @return Returns a reference to this object so that method calls can be chained together.
+         */
+        public Builder attributesToProject(Collection<String> attributesToProject) {
+            this.attributesToProject = attributesToProject != null ? new ArrayList<>(attributesToProject) : null;
+            return this;
+        }
+
+        /**
+         * <p>
+         * Sets one or more attribute names to be retrieved from the database. These attributes can include
+         * scalars, sets, or elements of a JSON document.
+         * </p>
+         * <p>
+         * If no attribute names are specified, then all attributes will be returned. If any of the requested attributes
+         * are not found, they will not appear in the result.
+         * </p>
+         * <p>
+         * For more information, see <a href=
+         * "https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Expressions.AccessingItemAttributes.html"
+         * >Accessing Item Attributes</a> in the <i>Amazon DynamoDB Developer Guide</i>.
+         * </p>
+         * @param attributesToProject
+         *        One or more  attributes names to be retrieved from the database.
+         * @return Returns a reference to this object so that method calls can be chained together.
+         */
+        public Builder attributesToProject(String... attributesToProject) {
+            return attributesToProject(Arrays.asList(attributesToProject));
+        }
+
+        /**
+         * <p>
+         * Adds a single attribute name to be retrieved from the database. This attribute can include
+         * scalars, sets, or elements of a JSON document.
+         * </p>
+         * <p>
+         * For more information, see <a href=
+         * "https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Expressions.AccessingItemAttributes.html"
+         * >Accessing Item Attributes</a> in the <i>Amazon DynamoDB Developer Guide</i>.
+         * </p>
+         * @param attributeToProject
+         *        An additional single attribute name to be retrieved from the database.
+         * @return Returns a reference to this object so that method calls can be chained together.
+         */
+        public Builder addAttributeToProject(String attributeToProject) {
+            if (attributesToProject == null) {
+                attributesToProject = new ArrayList<>();
+            }
+            attributesToProject.add(attributeToProject);
             return this;
         }
 
--- a/services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/model/ScanEnhancedRequest.java
+++ b/services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/model/ScanEnhancedRequest.java
@@ -15,7 +15,12 @@
 
 package software.amazon.awssdk.enhanced.dynamodb.model;
 
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import software.amazon.awssdk.annotations.SdkPublicApi;
 import software.amazon.awssdk.enhanced.dynamodb.DynamoDbTable;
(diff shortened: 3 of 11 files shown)


musketyr:
I'm not sure how can I fix the Quality Gate - the duplication and missing coverages origins in `hashCode` and `equals`.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/model/QueryEnhancedRequest.java
```diff
@@ -255,6 +273,40 @@ public Builder filterExpression(Expression filterExpression) {
...
+         *        If no attribute names are specified, then all attributes will be returned. If any of the requested
+         *        attributes are not found, they will not appear in the result.
+         *        </p>
+         *        <p>
+         *        For more information, see <a href=
+         *        "https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Expressions.AccessingItemAttributes.html"
+         *        >Accessing Item Attributes</a> in the <i>Amazon DynamoDB Developer Guide</i>.
+         * @return Returns a reference to this object so that method calls can be chained together.
+         */
+        public Builder projectionExpression(String projectionExpression) {
```
bmaizels, line 305: I'm thinking that since this is a high-level client, we should model this as '''List<String> attributesToProject''' or something like that? What do you think?

bmaizels:
> 
> 
> I'm not sure how can I fix the Quality Gate - the duplication and missing coverages origins in `hashCode` and `equals`.

Let's see what happens when we add the unit tests and functional tests that I mentioned in the review, might fix it, if not we can take another look then.

bmaizels:
@musketyr This is a legit missing feature that we planned to follow up with in the future. Thanks for taking it on, we welcome these kinds of contributions.

musketyr:
Thanks @bmaizels for your feedback, I will change the method signature to `attributesToProject` and will add the tests you have mentioned.

musketyr:
hi, @bmaizels. I've added the required tests as you wanted. regarding the duplications - it would make sense and make the life of the toolsmiths easier when there is a common interface/ancestor to `Scan*` and `Query*`  classes but it goes beyond this PR.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/operations/ScanOperation.java
```diff
@@ -48,21 +59,42 @@ private ScanOperation(ScanEnhancedRequest request) {
...
+        Map<String, String> expressionNames = null;
+
+        if (this.request.filterExpression() != null) {
+            expressionValues = this.request.filterExpression().expressionValues();
+            expressionNames = this.request.filterExpression().expressionNames();
+        }
+
+        String projectionExpression = null;
+        if (!this.request.attributesToProject().isEmpty()) {
+            List<String> placeholders = new LinkedList<>();
```
bmaizels, line 72: Can we use an ArrayList here as we're only adding and iterating?

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/operations/ScanOperation.java
```diff
@@ -15,25 +15,36 @@
...
 import software.amazon.awssdk.services.dynamodb.DynamoDbClient;
+import software.amazon.awssdk.services.dynamodb.model.AttributeValue;
 import software.amazon.awssdk.services.dynamodb.model.ScanRequest;
 import software.amazon.awssdk.services.dynamodb.model.ScanResponse;
 
 @SdkInternalApi
 public class ScanOperation<T> implements PaginatedTableOperation<T, ScanRequest, ScanResponse>,
                                          PaginatedIndexOperation<T, ScanRequest, ScanResponse> {
 
+    private static final UnaryOperator<String> PROJECTION_EXPRESSION_KEY_MAPPER = k -> "#AMZN_PROJ" + cleanAttributeName(k);
```
bmaizels, line 46: Do you think we should follow the same pattern established in other operations where we map attribute names? eg:- https://github.com/aws/aws-sdk-java-v2/blob/d4ea84ff7f601f489b6bfb396157bc3149138c88/services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/operations/UpdateItemOperation.java#L55-L56

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/operations/ScanOperation.java
```diff
@@ -48,21 +59,42 @@ private ScanOperation(ScanEnhancedRequest request) {
...
+
+        String projectionExpression = null;
+        if (!this.request.attributesToProject().isEmpty()) {
+            List<String> placeholders = new LinkedList<>();
+            Map<String, String> projectionPlaceholders = new HashMap<>();
+            for (String attr : this.request.attributesToProject()) {
+                String placeholder = PROJECTION_EXPRESSION_KEY_MAPPER.apply(attr);
+                placeholders.add(placeholder);
+                projectionPlaceholders.put(placeholder, attr);
+            }
```
bmaizels, line 78: Nit: Generally speaking we've been preferring Java 8 style (stream.forEach) where it's possible to use. I don't see any reason not to use that here.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/model/QueryEnhancedRequest.java
```diff
@@ -117,10 +120,10 @@ public Expression filterExpression() {
     }
 
     /**
-     * Returns the projection expression set on this request object, or null if it doesn't exist.
+     * Returns the list of projected properties on this request object, or an empty list if no projection is specified.
```
bmaizels, line 123: Can we reword this to say 'projected attributes' instead of 'projected properties' which I think will be more consistent with how we refer to them everywhere else.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/model/QueryEnhancedRequest.java
```diff
@@ -274,39 +277,28 @@ public Builder filterExpression(Expression filterExpression) {
...
-         *        "https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Expressions.AccessingItemAttributes.html"
-         *        >Accessing Item Attributes</a> in the <i>Amazon DynamoDB Developer Guide</i>.
+         * @param attributesToProject
+         *        The list of the attributes's paths to be retrieved from the database.
          * @return Returns a reference to this object so that method calls can be chained together.
          */
-        public Builder projectionExpression(String projectionExpression) {
-            this.projectionExpression = projectionExpression;
+        public Builder attributesToProject(List<String> attributesToProject) {
+            this.attributesToProject.addAll(attributesToProject);
```
bmaizels, line 287: We have a particular style around collections in builders that we should try and consistently follow. For an optional list like this:
1. If no values are given to the builder then the builder should pass 'null' to the constructor to indicate 'no optional values supplied'.
2. On instantiation, the object should shallow copy all the values out of the builder array and store it in an immutable List implementation (we typically use Collections.unmodifiableList of an ArrayList).
3. On the builder we provide a method to add a single element to the existing builder list (or create one if it is still null), these are named 'addXXX(...)' where XXX is the singular form of the collection noun. (eg: addAttributeToProject(...)).
4. Also on the builder we provide 2 overloaded methods to replace the entire list with one that is supplied by taking a shallow copy of it (this will overwrite any values already stored in the builder list). These are named 'XXX(...)' where XXX is the plural form of the collection noun. Eg: (attributesToProject(...)). The overload provides a varargs form and a List<...> form.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/model/ScanEnhancedRequest.java
```diff
@@ -92,10 +95,10 @@ public Expression filterExpression() {
     }
 
     /**
-     * Returns the projection expression set on this request object, or null if it doesn't exist.
+     * Returns the list of projected properties on this request object, or an empty list if no projection is specified.
```
bmaizels, line 98: Can we say 'projected attributes' here instead of 'projected properties'.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/model/ScanEnhancedRequest.java
```diff
@@ -35,14 +38,14 @@
...
-    private final String projectionExpression;
+    private final List<String> attributesToProject;
 
     private ScanEnhancedRequest(Builder builder) {
         this.exclusiveStartKey = builder.exclusiveStartKey;
         this.limit = builder.limit;
         this.consistentRead = builder.consistentRead;
         this.filterExpression = builder.filterExpression;
-        this.projectionExpression = builder.projectionExpression;
+        this.attributesToProject = builder.attributesToProject;
```
bmaizels, line 48: We should ensure this list cannot be modified, especially as we're vending it in a getter. I suggest ```this.attributesToProject = Collections.unmodifiableList(builder.attributesToProject);```

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/model/QueryEnhancedRequest.java
```diff
@@ -50,7 +53,7 @@ private QueryEnhancedRequest(Builder builder) {
         this.limit = builder.limit;
         this.consistentRead = builder.consistentRead;
         this.filterExpression = builder.filterExpression;
-        this.projectionExpression = builder.projectionExpression;
+        this.attributesToProject = builder.attributesToProject;
```
bmaizels, line 56: We should ensure this list cannot be modified, especially as we're vending it in a getter. I suggest ```this.attributesToProject = Collections.unmodifiableList(builder.attributesToProject);```

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/model/ScanEnhancedRequest.java
```diff
@@ -211,39 +214,28 @@ public Builder filterExpression(Expression filterExpression) {
...
+         * The list of the attributes's paths to be retrieved from the database.
+         *
+         * @param attributesToProject
+         *        The list of the attributes's paths to be retrieved from the database.
+         * @return Returns a reference to this object so that method calls can be chained together.
+         */
+        public Builder attributesToProject(String... attributesToProject) {
+            return attributesToProject(Arrays.asList(attributesToProject));
+        }
+
```
bmaizels: Please see comments I made in QueryEnhancedRequest around handling optional lists in builders.

services-custom/dynamodb-enhanced/src/test/java/software/amazon/awssdk/enhanced/dynamodb/functionaltests/BasicQueryTest.java
```diff
@@ -157,6 +157,28 @@ public void queryAllRecordsDefaultSettings_shortcutForm() {
         assertThat(page.lastEvaluatedKey(), is(nullValue()));
     }
 
+    @Test
+    public void queryAllRecordsDefaultSettings_withProjection() {
```
bmaizels, line 161: Can we add a test for multiple projected attributes? Also given what we know about the implementation I think I would sleep better if we also had a test with projected attributes AND a filter expression together, perhaps in a configuration that is likely to require them to merge correctly.

services-custom/dynamodb-enhanced/src/test/java/software/amazon/awssdk/enhanced/dynamodb/functionaltests/BasicScanTest.java
```diff
@@ -140,6 +141,24 @@ public void scanAllRecordsDefaultSettings() {
         assertThat(page.lastEvaluatedKey(), is(nullValue()));
     }
 
+    @Test
+    public void queryAllRecordsDefaultSettings_withProjection() {
```
bmaizels, line 145: Can we add a test for multiple projected attributes? Also given what we know about the implementation I think I would sleep better if we also had a test with projected attributes AND a filter expression together, perhaps in a configuration that is likely to require them to merge correctly.

services-custom/dynamodb-enhanced/src/test/java/software/amazon/awssdk/enhanced/dynamodb/internal/operations/QueryOperationTest.java
```diff
@@ -272,6 +272,22 @@ public void generateRequest_consistentRead() {
         assertThat(queryRequest.consistentRead(), is(true));
     }
 
+    @Test
+    public void generateRequest_projectionExpression() {
```
bmaizels, line 276: Could we test a filter expression together with a projection expression and verify the merging logic is working correctly?

services-custom/dynamodb-enhanced/src/test/java/software/amazon/awssdk/enhanced/dynamodb/internal/operations/ScanOperationTest.java
```diff
@@ -188,6 +189,25 @@ public void generateRequest_consistentRead() {
         assertThat(request, is(expectedRequest));
     }
 
+    @Test
```
bmaizels, line 192: Could we test a filter expression together with a projection expression and verify the merging logic is working correctly?

services-custom/dynamodb-enhanced/src/test/java/software/amazon/awssdk/enhanced/dynamodb/model/QueryEnhancedRequestTest.java
```diff
@@ -44,7 +47,7 @@ public void builder_minimal() {
         assertThat(builtObject.limit(), is(nullValue()));
         assertThat(builtObject.queryConditional(), is(nullValue()));
         assertThat(builtObject.scanIndexForward(), is(nullValue()));
-        assertThat(builtObject.projectionExpression(), is(nullValue()));
+        assertThat(builtObject.attributesToProject(), is(emptyIterable()));
```
bmaizels, line 50: A default optional list should be 'null' so we can differentiate between 'they didn't mention it' and 'they explicitly gave us an empty list as a value' if we ever need to.

services-custom/dynamodb-enhanced/src/test/java/software/amazon/awssdk/enhanced/dynamodb/model/QueryEnhancedRequestTest.java
```diff
@@ -69,7 +74,7 @@ public void builder_maximal() {
                                                                .limit(3)
                                                                .queryConditional(queryConditional)
                                                                .scanIndexForward(true)
-                                                               .projectionExpression(projectionExpression)
+                                                               .attributesToProject(attributesToProjectArray)
```
bmaizels, line 77: When we have the overloaded versions of this, hopefully in the implementation one calls the other, in which case we should make sure we are testing the one that will give us coverage over both. Usually that's the varargs form because that should call the other one with Array.asList(...). We will also need to add a separate test for the .addAttributeToProject(...) method that we are hopefully adding.

services-custom/dynamodb-enhanced/src/test/java/software/amazon/awssdk/enhanced/dynamodb/model/ScanEnhancedRequestTest.java
```diff
@@ -40,7 +43,7 @@ public void builder_minimal() {
         assertThat(builtObject.exclusiveStartKey(), is(nullValue()));
         assertThat(builtObject.consistentRead(), is(nullValue()));
         assertThat(builtObject.filterExpression(), is(nullValue()));
-        assertThat(builtObject.projectionExpression(), is(nullValue()));
+        assertThat(builtObject.attributesToProject(), is(emptyIterable()));
```
bmaizels, line 46: A default optional list should be 'null' so we can differentiate between 'they didn't mention it' and 'they explicitly gave us an empty list as a value' if we ever need to.

services-custom/dynamodb-enhanced/src/test/java/software/amazon/awssdk/enhanced/dynamodb/model/ScanEnhancedRequestTest.java
```diff
@@ -56,20 +59,21 @@ public void builder_maximal() {
...
-        String projectionExpression = "one,two";
+        String[] attributesToProjectArray = {"one", "two"};
+        List<String> attributesToProject = Arrays.asList(attributesToProjectArray);
 
         ScanEnhancedRequest builtObject = ScanEnhancedRequest.builder()
                                                              .exclusiveStartKey(exclusiveStartKey)
                                                              .consistentRead(false)
                                                              .filterExpression(filterExpression)
-                                                             .projectionExpression(projectionExpression)
+                                                             .attributesToProject(attributesToProjectArray)
```
bmaizels, line 69: When we have the overloaded versions of this, hopefully in the implementation one calls the other, in which case we should make sure we are testing the one that will give us coverage over both. Usually that's the varargs form because that should call the other one with Array.asList(...). We will also need to add a separate test for the .addAttributeToProject(...) method that we are hopefully adding.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/model/QueryEnhancedRequest.java
```diff
@@ -274,39 +277,28 @@ public Builder filterExpression(Expression filterExpression) {
...
          */
-        public Builder projectionExpression(String projectionExpression) {
-            this.projectionExpression = projectionExpression;
+        public Builder attributesToProject(List<String> attributesToProject) {
+            this.attributesToProject.addAll(attributesToProject);
             return this;
         }
 
+        /**
+         * The list of the attributes's paths to be retrieved from the database.
```
bmaizels, line 292: 'path' is a new term we have not used anywhere else in the library and may cause confusion. Can we simply refer to this as 'attribute names'? I think a little more description or at least a link to the official DynamoDB documentation would also help people understand what to expect from this feature. (Please apply this comment to other places it's applicable too)

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/operations/ScanOperation.java
musketyr: Sincerely, I barely use `LinkedList` (mostly because I live in a Groovy world where every new list `[]` is an `ArrayList`) but for the operations, you have also mentioned it looks like the best option to choose. As far as I can remember `ArrayList` is designed for access by index and `LinkedList` for adding and interacting.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/operations/ScanOperation.java
musketyr: I'm not sure if we need to distinct the placeholder by a different prefix. Probably not. Is there a reason why these mappers themselves are not part of `EnhancedClientUtils`? They appear in at least 4 classes now.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/operations/ScanOperation.java
musketyr: Me, too. There was a reference to a non-final variable in a previous version of the code. I'm switching to `forEach`.

services-custom/dynamodb-enhanced/src/test/java/software/amazon/awssdk/enhanced/dynamodb/model/QueryEnhancedRequestTest.java
musketyr: This is already testing the varargs variant. I will add the one for the `add...` method.

services-custom/dynamodb-enhanced/src/test/java/software/amazon/awssdk/enhanced/dynamodb/model/ScanEnhancedRequestTest.java
musketyr: This is already testing the varargs variant. I will add the one for the add... method.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/operations/ScanOperation.java
musketyr: Actually, you are right, we should use the same pattern to deduplicate the keys and names.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/operations/ScanOperation.java
bmaizels: My best understanding is that LinkedList is beneficial if you need to insert records into the middle or start of a list, or if you wish to join two lists (and any other operations that require splicing and displacing the index of other records), whereas for this use case ArrayList is simpler, faster and uses less memory. It's even more efficient if you initialize the ArrayList with an accurate predicted capacity, but most people don't bother with this optimization.

musketyr: right, haven't thought about copying, just inserting.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/model/QueryEnhancedRequest.java
```diff
@@ -255,6 +280,73 @@ public Builder filterExpression(Expression filterExpression) {
...
+         * <p>
+         * For more information, see <a href=
+         * "https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Expressions.AccessingItemAttributes.html"
+         * >Accessing Item Attributes</a> in the <i>Amazon DynamoDB Developer Guide</i>.
+         * </p>
+         * @param attributeToProject
+         *        An additional single attribute name to be retrieved from the database.
+         * @return Returns a reference to this object so that method calls can be chained together.
+         */
+        public Builder addAttributesToProject(String attributeToProject) {
```
bmaizels, line 342: Can we use the singular noun-form here: ```addAttributeToProject(...)```

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/model/ScanEnhancedRequest.java
```diff
@@ -192,6 +217,73 @@ public Builder filterExpression(Expression filterExpression) {
...
+         * <p>
+         * For more information, see <a href=
+         * "https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Expressions.AccessingItemAttributes.html"
+         * >Accessing Item Attributes</a> in the <i>Amazon DynamoDB Developer Guide</i>.
+         * </p>
+         * @param attributeToProject
+         *        An additional single attribute name to be retrieved from the database.
+         * @return Returns a reference to this object so that method calls can be chained together.
+         */
+        public Builder addAttributesToProject(String attributeToProject) {
```
bmaizels, line 279: Can we use the singular noun form here: ```addAttributeToProject(...)```

bmaizels:
Ok I think we're in great shape. Just a few final procedural things. The first you can probably help me with: can we squash this into a single commit with a single concise descriptive commit message for the change? You'll need to squash it first in your local repo then force push to your branch.

bmaizels:
> Ok I think we're in great shape. Just a few final procedural things. The first you can probably help me with: can we squash this into a single commit with a single concise descriptive commit message for the change? You'll need to squash it first in your local repo then force push to your branch.

Actually let's wait and make sure checks pass first so you only have to do this once.

musketyr:
I can try so but in just wonder, isn't it the same as using "Squash and merge" option here on GitHub?

bmaizels:
> 
> 
> I can try so but in just wonder, isn't it the same as using "Squash and merge" option here on GitHub?

We disabled that for this repo unfortunately.

bmaizels:
Don't worry about the quality gate I've examined the warnings and they are false positives as far as I'm concerned, so go ahead and squash whenever you are ready.

musketyr:
@bmaizels done, luckily it was a quite trivial thing to do.

bmaizels:
@musketyr Hey, we release daily at 11am, and because we missed today's release the Changelog needs rebasing/redoing. If you can fix that within the next few hours I should be able to merge this and it will go out in Monday's release. Thanks!

bmaizels:
Merged. Thank you very much for your contribution!

musketyr:
Thanks you as well!

metinkilic:
There is a bug on this PR and I mentioned on the https://github.com/aws/aws-sdk-java-v2/commit/8b85f34e829b5774aae758352818ab570cfc45f5?diff=unified

PR #1757, opened 2020-04-01
DDB Enhanced:  Expose attribute extractors and converters

musketyr:
## Description
This PR exposes the `Attribute` and `AttributeType` API from the `TableSchema` to allow
easier generalization of the construction of queries.

## Motivation and Context

In version 1, there were API to convert the Java value to the `AttributeValue` (`DynamoDBMapperFieldModel`) which is not present in version 2.  

See [RangeConditionCollector](https://github.com/agorapulse/micronaut-aws-sdk/blob/master/subprojects/micronaut-aws-sdk-dynamodb/src/main/groovy/com/agorapulse/micronaut/aws/dynamodb/builder/RangeConditionCollector.java) for an example usage with version 1. Exposing the already present information about the attributes enables reimplementing similar behaviour for version 2.

## Testing

The main problem with this PR might be that `AttributeType` has been moved to the public API so the package has been changed. This might break some existing applications which ignored that `AttributeType` was a internal API so far.

## Types of changes

NOTE: this change is making an internal interface into public one which may break some existing 3rd party code which relies on the internal API. As the interface was marked as internal before then it should be considered non-breaking change.

## Checklist

## License


bmaizels:
Hi @musketyr, thank you for taking a look at the enhanced client preview and also submitting your contribution to improve the experience.

Before really drilling on the change itself I first need to make sure I understand the motivation and what it is you are trying to do because generally speaking we are very resistant to making internal interfaces public unless absolutely necessary as that restricts what we can do with those interfaces in the future, so the cost is high.

Based on your description of your use-case it sounds like you are trying to apply a TableSchema to extract and convert one or more object modeled attributes to their AttributeValue equivalents that would be written to the DynamoDB table if you were doing, say, a PutItem?

If that is true, is this existing method in TableSchema not sufficient to accomplish that?

```
    /**
     * Takes a modelled object and extracts a specific set of attributes which are then returned as a map of
     * {@link AttributeValue} that the DynamoDb low-level SDK can work with. This method is typically used to extract
     * just the key attributes of a modelled item and will not ignore nulls on the modelled object.
     *
     * @param item The modelled Java object to extract the map of attributes from.
     * @param attributes A collection of attribute names to extract into the output map.
     * @return A map of String to {@link AttributeValue} representing the requested modelled attributes in the model
     * object.
     */
    Map<String, AttributeValue> itemToMap(T item, Collection<String> attributes);
```

bmaizels:
Also, to convert and retrieve a single AttributeValue you can use:

```
    /**
     * Returns a single attribute value from the modelled object.
     *
     * @param item The modelled Java object to extract the attribute from.
     * @param key The attribute name describing which attribute to extract.
     * @return A single {@link AttributeValue} representing the requested modelled attribute in the model object or
     * null if the attribute has not been set with a value in the modelled object.
     */
    AttributeValue attributeValue(T item, String key);
```

musketyr:
Hi @bmaizels,

thank you for a quick answer! The motivation is to reimplement advanced queries and updates using the declarative clients in Micronaut. There is a working version using v1.

https://agorapulse.github.io/micronaut-aws-sdk/#_declarative_services_with_service

If you take a look at the `@Query` example

```
 @Query({
        query(DynamoDBEntity) {
            hash hashKey
            range { between DynamoDBEntity.DATE_INDEX, after, before }
        }
    })
```

I would like to convert the values `after` and `before` to the `AttributeValue` but the only information I have is the name of the attribute (in this case an index, but can be an attribute in general as a similar DSL exists for scanning. To use the existing method I would have to populate a new object with and assign the filter values (probably using reflection or similar way)

```
public <T> AttributeValue convert(TableSchema<T> schema, String attrName, Object value) {
    T instance = schema.itemType().rawClass().newInstance();
    setAttrOnInstance(instance, attrName, value);
    return schema.attrributeValue(instance, attrName);
}
```

With the new exposed API, it would be much easier:

```
public <T> AttributeValue convert(TableSchema<T> schema, String attrName, Object value) {
    return schema.attribute(attrName).attributeType(). objectToAttributeValue(value);
}
```

This approach is much cleaner, it uses the information already present in the metamodel, it is reflection-free and it avoids unnecessary object creation.

bmaizels:
@musketyr Thanks for explaining better, I understand what you are trying to do now and why you are trying to do it. Having said that I'm still hesitant to approve making these interfaces public for the aforementioned reasons. Please give me some time to think on it and discuss with my team and see if we can come up with any other ideas about how to solve this for you.

bmaizels:
>> Please give me some time to think on it and discuss with my team and see if we can come up with any other ideas about how to solve this for you.
My initial thoughts is this use-case might be better served by the Document API we are planning to release as part of the DDB Enhanced Client. I'll get back to you.

musketyr:
Thank you! I'm just wondering, would it be more acceptable for you to add the `convert` method mentioned above to the `TableSchema` and keeping the rest of the API changes internal?

musketyr:
@bmaizels  see #1765 for a minimal version of this PR which only adds a single method into `TableSchema` class.

PR #1768, opened 2020-04-06
Mark a connection as unreusable if there was a 5xx error so that new …

zoewangg:
## Description
Mark a connection as unreusable if there was a 5xx error so that new request will execute on new connection
- h1: closing connection after the request is complete
- h2: stopping creating streams on that connection and closing the connection once there are no more streams on it.

## Testing
Add functional tests
integ tests and stability tests pass
Baking the changes in canaries now.

## Types of changes

## Checklist

## License


http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/Http2ToHttpInboundAdapter.java
```diff
@@ -54,7 +59,21 @@ protected void channelRead0(ChannelHandlerContext ctx, Http2Frame frame) throws
...
+
+        if (HttpStatusFamily.of(httpResponse.status().code()) == HttpStatusFamily.SERVER_ERROR) {
+            fireConnectionExceptionForServerError(ctx);
+        }
+    }
+
+    private void fireConnectionExceptionForServerError(ChannelHandlerContext ctx) {
+        if (ctx.channel().parent() != null) {
+            Channel parent = ctx.channel().parent();
+            log.debug(() -> "An 5xx server error occurred on an Http2 stream, notifying the connection channel " + parent);
```
dagnir, line 74: Should we log the stream ID?

zoewangg: yup, updated

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/Http2ToHttpInboundAdapter.java
```diff
@@ -54,7 +59,21 @@ protected void channelRead0(ChannelHandlerContext ctx, Http2Frame frame) throws
...
+        if (HttpStatusFamily.of(httpResponse.status().code()) == HttpStatusFamily.SERVER_ERROR) {
+            fireConnectionExceptionForServerError(ctx);
+        }
+    }
+
+    private void fireConnectionExceptionForServerError(ChannelHandlerContext ctx) {
+        if (ctx.channel().parent() != null) {
+            Channel parent = ctx.channel().parent();
+            log.debug(() -> "A 5xx server error occurred on an Http2 stream, notifying the connection channel " + ctx.channel());
+            parent.pipeline().fireExceptionCaught(Http2StreamServerErrorException.INSTANCE);
```
millems, line 75: Do we not support modeled 5xx exceptions for HTTP2 connections? It feels like this would break them, if we did.

millems: Actually, looks like this may not close the connection immediately, just closes it to new requests. I'm not sure if I like the instanceof checks to disambiguate this IOException from other IOExceptions. Should we have a different root exception type?

zoewangg: We support modeled 5xx exceptions in core, and http client modules do not have knowledge of it.

Yeah, it doesn't close the connection immediately; the `RecordState` will be changed to `CLOSE_TO_NEW`

I can change it to `RuntimeException`. WDUT?

millems: So, all RuntimeExceptions on the channel will put it into no-new-connections state? I don't think that's a bad thing, but wanted to make sure my proposal was clear.

zoewangg: So,  `Http2StreamServerErrorException` (extends RuntimeException) or `Http2StreamIoException`(extends IOException) on the stream channel will put the connection channel into close_to_new state. 

All other exceptions on the stream channels will not have any effect on connection channel.

zoewangg: Discussed offline. We decided to have a single exception class for both server errors and io error on streams so that we don't need to update the exception handling logic if there are more cases in the future

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/Http2ConnectionTerminatingException.java
```diff
@@ -0,0 +1,29 @@
...
+
+package software.amazon.awssdk.http.nio.netty.internal.http2;
+
+import software.amazon.awssdk.annotations.SdkInternalApi;
+
+/**
+ * Exception indicating a connection is terminating
+ */
+@SdkInternalApi
+final class Http2ConnectionTerminatingException extends RuntimeException {
```
millems, line 24: Just a sanity check: we aren't relying on the existing exceptions we're replacing with this being an IOException (e.g. for retries) are we?

zoewangg: Nope, this exception is only used internally to change the connection state to `CLOSE_TO_NEW` and is not returned to upstream.

PR #1798, opened 2020-04-29
Expose the extendedRequestId from SdkServiceException

Sarev0k:
## Description
Adds a new `extendedRequestId` property to the `SdkServiceException` class and includes it in the contents of AwsServiceException messages.

## Motivation and Context
Currently the only way to access this information from the exception is via the headers supplied by the `SdkHttpResponse` in the exception's `errorDetails`.  However when the AWS client is automatically retrying and causing my Lambda function to timeout due to the added time it takes to retry, I have no way to access this information.  AWS support requires this information in order to deep dive issues for some services such as S3.

## Testing
Ran mvn clean install and the build succeeded.

## Types of changes

## Checklist

## License

### DIFF
--- a/core/aws-core/src/main/java/software/amazon/awssdk/awscore/exception/AwsServiceException.java
+++ b/core/aws-core/src/main/java/software/amazon/awssdk/awscore/exception/AwsServiceException.java
@@ -65,7 +65,8 @@ public class AwsServiceException extends SdkServiceException {
             return awsErrorDetails().errorMessage() +
                     " (Service: " + awsErrorDetails().serviceName() +
                     ", Status Code: " + statusCode() +
-                    ", Request ID: " + requestId() + ")";
+                    ", Request ID: " + requestId() +
+                    ", Extended Request ID: " + extendedRequestId() + ")";
         }
 
         return super.getMessage();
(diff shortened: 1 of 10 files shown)


Sarev0k:
Hey @millems, could you please take a look at this?

Thanks.

core/aws-core/src/main/java/software/amazon/awssdk/awscore/exception/AwsServiceException.java
```diff
@@ -62,10 +62,13 @@ public AwsErrorDetails awsErrorDetails() {
...
     public String getMessage() {
         if (awsErrorDetails != null) {
+            String extendedRequestId =
+                awsErrorDetails().sdkHttpResponse().firstMatchingHeader("x-amz-id-2").orElse("not available");
             return awsErrorDetails().errorMessage() +
                     " (Service: " + awsErrorDetails().serviceName() +
                     ", Status Code: " + statusCode() +
-                    ", Request ID: " + requestId() + ")";
+                    ", Request ID: " + requestId() +
+                    ", Extended Request ID: " + extendedRequestId + ")";
```
millems, line 71: I realize it's actually DRYer to do this here, but for consistency can we set the extendedRequestId the same way we set the requestId: when the exception is created? (And maybe just print null instead of not available)

We can add a getter for retrieving it (that sometimes returns null).

Sarev0k: Yep, that's actually how I initially implemented this, but I was trying to make this simpler because I was worried the change would be too complicated.  I'll put out the original version today.

PR #1802, opened 2020-04-30
Update design docs based on meetings with team

dagnir:
## Types of changes

## Checklist

## License


docs/design/core/metrics/prototype/MetricEvents.java
```diff
@@ -0,0 +1,49 @@
...
+     * @throws IllegalStateException If the given event is already present.
+     * @throws IllegalArgumentException If {@code eventData} is {@code null}.
+     */
+    <T> void putMetricEvent(MetricEvent<T> event, T eventData);
+
+    /**
+     * Return the metric data associated with the given event. Returns {@code
+     * null} if no event is found.
+     */
+    <T> T getMetric(MetricEvent<T> event);
```
bmaizels, line 43: Should we name this getMetricData?

docs/design/core/metrics/prototype/MetricEvents.java
```diff
@@ -0,0 +1,49 @@
...
+import software.amazon.awssdk.metrics.meter.Timer;
+
+/**
+ * Object used to store metric events collected by the SDK.
+ */
+@SdkPublicApi
+public interface MetricEvents {
+    /**
+     * Add the given metric with associated data.
+     * @throws IllegalStateException If the given event is already present.
```
bmaizels, line 34: I don't think this is a good fit for IllegalStateException. Usually this is used when we try and do something at the wrong time like read from a stream that has been closed or whatever. In this case I think it could be covered by IllegalArgumentException. What do you think?

docs/design/core/metrics/prototype/MetricEvents.java
```diff
@@ -0,0 +1,49 @@
...
+    /**
+     * Return the metric data associated with the given event. Returns {@code
+     * null} if no event is found.
+     */
+    <T> T getMetric(MetricEvent<T> event);
+
+    /**
+     * Return the contained metric events.
+     */
+    Map<MetricEvent<?>, ?> getMetricEvents();
```
bmaizels, line 48: I'm not a fan of returning maps on public interfaces. I would prefer we have specific accessors for specific use-cases, for instance if they need a specific MetricEvent they can use getMetric/getMetricData, if they require an iterator/iterable we could provide that too.

docs/design/core/metrics/prototype/MetricPublisher.java
```diff
@@ -64,5 +54,5 @@
      * Implementations can also call publish method for every reported metric. But this can be expensive and
      * is not recommended.
      */
-    CompletableFuture<Void> publish();
+    CompletableFuture<Void> publish(MetricEvents metricEvents);
```
bmaizels, line 57: Although the events themselves are immutable, the MetricEvents registry is not. This brings up questions in my mind like what happens if a metric is added during the publish cycle, is that undefined behavior? What are they meant to do with the MetricEvents object once it's been published, discard it? I think we need either better documentation explaining all that or a small rethink around this aspect of it.

docs/design/core/metrics/prototype/MetricEvents.java
dagnir: +1

docs/design/core/metrics/prototype/MetricEvents.java
dagnir: Yeah I think that makes more sense

docs/design/core/metrics/prototype/MetricEvents.java
dagnir: Sure I think that makes sense. Mostly I wanted to provide the ability to get at a metric data without the casting step but that can be solved with a `getMetricData` overload that doesn't take a `Class<T>`.

docs/design/core/metrics/prototype/MetricPublisher.java
dagnir: Yeah the mutability is an issue. I think that can be solved by passing around a `MetricEvents.Builder` instead, and `MetricEvents` becomes immutable and publish can take consumer builder. WDYT?

docs/design/core/metrics/prototype/MetricEvents.java
bmaizels: The publisher is going to need some kind of iterator or list though isn't it?

docs/design/core/metrics/prototype/MetricPublisher.java
bmaizels: Well two different approaches come to my mind:
1) Make the MetricEvents a closeable object, and once it is closed no new metrics can be added and only closed MetricEvents can be published. This requires orchestration upstream though to manage these multiple MetricEvents objects.
2) Make the whole thing a stream that has publishers attached to it with markers so they know the last event successfully published.

Your idea with passing around the builder really conforms more to my first approach.

docs/design/core/metrics/prototype/MetricEvents.java
dagnir: Yeah an iterator makes sense in this case.

docs/design/core/metrics/prototype/MetricPublisher.java
dagnir: I think the builder and suggestion 1 approach are the cleanest IMO. requiring a `MetricEvents` rather than a `MetricEvents.Builder` makes the "closing" explicit so I'm sort of leaning toward this approach

bmaizels: Is it not the job of the aggregators to aggregate and bundle metrics to then be passed as a batch to publisher? Maybe we're missing a whole layer here : Metrics flow as a stream to aggregators, aggregators create metric bundles and send those to publishers, publishers publish the bundle. It's then up to the aggregator to handle the stream to batch orchestration.

dagnir: Discussed this offline. The interface doesn't require implementations of `MetricPublisher` to immediately publish the metrics given in the `publish` call and allows for batching/aggregation on the back end; the aggregator is collapsed into the publisher. This falls in line with the concept of streams (in this case a streams of `MetricEvents`) mentioned above.

docs/design/core/metrics/prototype/MetricEvents.java
```diff
@@ -25,25 +25,49 @@
...
+     * Return an iterator of the contained metric events and their data.
      */
-    <T> T getMetric(MetricEvent<T> event);
+    Iterator<MetricEventRecord<?>> getMetricEventsIterator();
+
+
+   /**
+    * A container associating an event with its data.
+    */
+   interface MetricEventRecord<T> {
```
bmaizels, line 48: Should we move this to the top level of the package?

docs/design/core/metrics/prototype/MetricEvents.java
```diff
@@ -25,25 +25,49 @@
...
+    <T> T getMetricEventData(MetricEvent<T> event);
+
 
     /**
-     * Return the metric data associated with the given event. Returns {@code
-     * null} if no event is found.
+     * Return an iterator of the contained metric events and their data.
      */
-    <T> T getMetric(MetricEvent<T> event);
+    Iterator<MetricEventRecord<?>> getMetricEventsIterator();
```
bmaizels, line 42: Do we have any concerns around type-safety here? How is the publisher going to check the type of the events?

docs/design/core/metrics/prototype/MetricPublisher.java
```diff
@@ -20,39 +20,72 @@
...
+     * <p>
+     * With the exception of a {@code null} {@code metricEvents}, all
+     * invocations of this method must return normally. The only legal way to
+     * report an error is by completing the returned future exceptionally. This
+     * is to ensure that callers of the publisher can safely assume that even
+     * in situations where an error happens during publishing that it will not
+     * interrupt the calling thread.
+     * <p>
+     * The future is completed when the metrics calculated or otherwise derived
+     * from the given {@code metricEvents} have been published.
```
bmaizels, line 75: That sounds like an ambitious contract, are we going to keep track of each completable future that acts as a component for each aggregated metric and then make sure we keep track of it all when we publish? What if the publisher is complex and publishes to other publishers or something like that? Do we need to return anything at all? What is the caller going to do with the information?

dagnir: I don't really think it's necessary to return anything. One use case I can think if is preventing JVM exit by waiting on the futures returned but we also enforce flushing before returning in the `close()` method, so I think it's actually fine for this to be `void`.

docs/design/core/metrics/prototype/MetricEvents.java
dagnir: I think in practice not really. The component aggregating the metric events necessarily needs to have knowledge of the types of events that are it can expect and their payloads

docs/design/core/metrics/prototype/MetricEvents.java
dagnir: Sure

dagnir:
Mostly superseded by #1844. Will move the docs update there as well

PR #1817, opened 2020-05-06
Fix generation for services that contain operations with the same name as the service

adamthom-amzn:
## Description
If a 'FooService' contains an operation named 'Foo', the generator will attempt to generate a 'FooRequest' that extends an abstract 'FooRequest'.  This adds 'Default' to the concrete class names only in cases where there is an operation with a clashing name.

## Motivation and Context
Some simpler microservices do not adhere to the standard naming guidelines.  This allows clients to be generated for them.

## Testing
This should only impact models for which non-compiling code would be generated today.  This was tested with a simple unit test, and out-of-band against a similar service model end-to-end.

## Types of changes

## Checklist

## License

### DIFF
--- a/codegen/src/test/java/software/amazon/awssdk/codegen/IntermediateModelBuilderTest.java
+++ b/codegen/src/test/java/software/amazon/awssdk/codegen/IntermediateModelBuilderTest.java
@@ -15,16 +15,34 @@
 
 package software.amazon.awssdk.codegen;
 
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.Assert.assertEquals;
 
 import java.io.File;
 import org.junit.Test;
 import software.amazon.awssdk.codegen.model.config.customization.CustomizationConfig;
 import software.amazon.awssdk.codegen.model.intermediate.IntermediateModel;
+import software.amazon.awssdk.codegen.model.intermediate.ShapeModel;
 import software.amazon.awssdk.codegen.model.service.ServiceModel;
 import software.amazon.awssdk.codegen.utils.ModelLoaderUtils;
 
 public class IntermediateModelBuilderTest {
+
+    @Test
+    public void testServiceAndShapeNameCollisions() throws Exception {
+        final File modelFile = new File(IntermediateModelBuilderTest.class
+                                            .getResource("poet/client/c2j/collision/service-2.json").getFile());
+        IntermediateModel testModel = new IntermediateModelBuilder(
+            C2jModels.builder()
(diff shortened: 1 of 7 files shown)


millems:
Wouldn't this change make it backwards-incompatible to add an operation with the same name as the service, since it would rename an existing class?

Would it be better to rename the operation?

adamthom-amzn:
> Wouldn't this change make it backwards-incompatible to add an operation with the same name as the service, since it would rename an existing class?
> 
> Would it be better to rename the operation?

Yeah, though I think it's unlikely someone would do that, it's a fair point.  I will see if I can figure out what a decent heuristic for renaming the operation would be.

adamthom-amzn:
Changed it to change the generation of request/response/exception implementations rather than the base class.  As discussed offline, I think this is a bit more understandable than changing the name of the operation (as I think customers will expect the operation name to align with the model and its documentation more than the request type name), though I can also explore renaming the operation if the naming convention change for the request types seems too invasive upon review.

I also don't know why AddEmptyInputShape/AddEmptyOutputShape were not delegating naming of the shapes to the naming strategy; if that was not an oversight, I can revisit those changes as well.

codegen/src/test/java/software/amazon/awssdk/codegen/IntermediateModelBuilderTest.java
```diff
@@ -0,0 +1,47 @@
...
+                                            .getResource("poet/client/c2j/collision/service-2.json").getFile());
+        IntermediateModel testModel = new IntermediateModelBuilder(
+            C2jModels.builder()
+                     .serviceModel(ModelLoaderUtils.loadModel(ServiceModel.class, modelFile))
+                     .customizationConfig(CustomizationConfig.create())
+                     .build())
+            .build();
+
+        assertEquals(new HashSet<>(Arrays.asList("DefaultCollisionException", "DefaultCollisionRequest", "DefaultCollisionResponse")),
+                     testModel.getShapes().values().stream().map(s -> s.getShapeName()).collect(Collectors.toSet()));
```
millems, line 44:
```java
assertThat(testModel.getShapes().values())
    .extracting(ShapeModel::getShapeName)
    .containsExactlyInAnyOrder("DefaultCollisionException", "DefaultCollisionRequest", "DefaultCollisionResponse");
```
?

adamthom-amzn: Changed.

PR #1821, opened 2020-05-07
Fix binding of SdkPojoSupplier to event stream type header.

adamthom-amzn:
## Description
Event stream type header values are generated from the member name in the event stream structure, not the structure name of the event itself.  When the member name and the event type mismatch, then unmarshalling of events sent from the service fails.

From the spec: `:event-type: This header’s value, a UTF-8 string, is derived from the event’s member name in the eventstream structure`

## Motivation and Context
This brings the implementation of event stream unmarshalling into sync with the spec.

## Testing
I modified the existing generation tests to generate a mismatch between output event type names and member names.  I have also tested this offline end-to-end with a service model that has this mismatch.

## Types of changes

## Checklist

## License


millems:
Do these two values match in Kinesis and Transcribe, and that's why they work? I want to make sure this wouldn't break one of those services.

adamthom-amzn:
transcribe-streaming has [AudioEvent](https://github.com/aws/aws-sdk-java-v2/blob/master/services/transcribestreaming/src/main/resources/codegen-resources/service-2.json#L68) that refers to the AudioEvent structure, and [TranscriptEvent](https://github.com/aws/aws-sdk-java-v2/blob/master/services/transcribestreaming/src/main/resources/codegen-resources/service-2.json#L329) that points at a shape of the same name, as well as a bunch of exceptions that follow this pattern.

Kinesis [follows this pattern](https://github.com/aws/aws-sdk-java-v2/blob/b577439747192d8af298d43d039c2d131a28b178/services/kinesis/src/main/resources/codegen-resources/service-2.json#L1878) as well.

PR #1823, opened 2020-05-07
Support event streams that are shared between two operations.

adamthom-amzn:
## Description
Event streams do not have to be 1:1 with operations, but the generation of event stream model types selects the first operation linked with an event stream when generating.  This change adds support for all associated operations to event stream interfaces and event types.

## Motivation and Context
This brings event stream client generation closer to the models allowed by the event stream specification.

## Testing
I added a generation test for a model that shares an event stream between two operations, and performed an end-to-end test offline with a test service that reuses event stream structures.

## Types of changes

## Checklist

## License

### DIFF
--- a/codegen/src/test/resources/software/amazon/awssdk/codegen/poet/model/sharedstream/person.java
+++ b/codegen/src/test/resources/software/amazon/awssdk/codegen/poet/model/sharedstream/person.java
@@ -1,0 +1,250 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.services.sharedeventstream.model;
+
+import java.io.Serializable;
+import java.time.Instant;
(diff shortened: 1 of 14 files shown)


codegen/src/test/resources/software/amazon/awssdk/codegen/poet/model/sharedstream/person.java
```diff
@@ -0,0 +1,234 @@
...
+  private Person(BuilderImpl builder) {
+    this.name = builder.name;
+    this.birthday = builder.birthday;
+  }
+
+  /**
+   * Returns the value of the Name property for this object.
+   * @return The value of the Name property for this object. */
+  public String name() {
+    return name;}
```
millems, line 74: Formatting seems off. Did the generator do this? Does this compile?

adamthom-amzn: Yeah, I just copy and pasted the output from the generator and reviewed it to make sure it generated the methods I wanted.  IDEA only complains about missing types when I move it into a source root.

millems: Can we run a quick formatter on it? The comparison doesn't care about whitespace, I don't think.

adamthom-amzn: Ran IDEA's formatter on all of the newly added generated classes.

adamthom-amzn: And broke the tests.  Don't merge this yet.

adamthom-amzn: IDEA changed

```
/*
 */
```
to

```
/*
 *
 */
```

which, of course, is not whitespace-only.  I pushed an update, should be good as soon as the automated tests don't call me a liar.

PR #1825, opened 2020-05-08
Audited the javadoc in dynamodb-enhanced and corrected some error.

Quanzzzz:
Audit the javadoc in dynamodb-enhanced, corrected some bad formatted examples and some javadoc errors.
## Testing

Tested all the Javadocs by generator in IntelliJ.
## License


services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/converter/attribute/ListAttributeConverter.java
```diff
@@ -45,28 +45,36 @@
...
+ * {@literal AttributeConverter<Set<Integer>> setConverter =
+ * CollectionAttributeConverter.setConverter(IntegerAttributeConverter.create());}
+ * </code>
  * <p>
- * AttributeConverter<SortedSet<Integer>> sortedSetConverter =
- * CollectionAttributeConverter.sortedSetConverter(IntegerAttributeConverter.create());
+ * <code>
+ * {@literal AttributeConverter<SortedSet<Integer>> sortedSetConverter =
+ * CollectionAttributeConverter.sortedSetConverter(IntegerAttributeConverter.create());}
  * </code>
```
millems, line 78: What's the difference between <code> and <pre>?

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/converter/attribute/MapAttributeConverter.java
```diff
@@ -47,25 +47,28 @@
...
+ * <p>
  * <code>
- * AttributeConverter<Map<MonthDay, String>> mapConverter =
+ * {@literal AttributeConverter<Map<MonthDay, String>> mapConverter =
  * MapAttributeConverter.builder(EnhancedType.mapOf(Integer.class, String.class))
  * .mapConstructor(HashMap::new)
  * .keyConverter(MonthDayStringConverter.create())
  * .valueConverter(StringAttributeConverter.create())
- * .build();
+ * .build();}
```
millems, line 57: Why @literal and not @code?

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/mapper/annotations/DynamoDbBean.java
```diff
@@ -46,9 +46,11 @@
  * </ul>
  *
  * Example using attribute converter providers with one custom provider and the default provider:
+ * <pre>
  * {@code
  * (converterProviders = {CustomAttributeConverter.class, DefaultAttributeConverterProvider.class});
  * }
```
millems, line 52: Does @code work when the content has a } in it?

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/converter/attribute/ListAttributeConverter.java
Quanzzzz: ```<code>``` makes the font different, but it cannot guarantee the formats, and ```<pre>``` can let the format of the code be showed as what they are in the original code file, i.e showing the indents and spaces.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/converter/attribute/MapAttributeConverter.java
Quanzzzz: The ```@literal``` is for the ```'<'``` and``` '>'``` signs, they are special signs in html, thus they cannot be recognize as the other signs, adding ```@literal``` guarantees that all the strings contained are just what they should be looked like. Usually this annotation is for special signs i.e ```'<', '>', '@'```

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/mapper/annotations/DynamoDbBean.java
Quanzzzz: Only if all the brackets are matched, if not, there will be an error.

PR #1833, opened 2020-05-13
Add SourceRegion to CopyDBClusterSnapshot and CreateDBCluster operations.

millems:
As with CopyDBSnapshot and CreateDBInstanceReadReplica, specifying this field will automatically populate the PresignedURL field with a valid value.

### DIFF
--- a/services/rds/src/test/java/software/amazon/awssdk/services/rds/internal/PresignRequestWireMockTest.java
+++ b/services/rds/src/test/java/software/amazon/awssdk/services/rds/internal/PresignRequestWireMockTest.java
@@ -1,0 +1,101 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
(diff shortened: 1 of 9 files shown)


services/rds/src/test/java/software/amazon/awssdk/services/rds/internal/PresignRequestWireMockTest.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    public void verifyMethodCallSendsPresignedUrl(Runnable methodCall, String actionName) {
+        stubFor(any(anyUrl()).willReturn(aResponse().withStatus(200).withBody("<body/>")));
+
+        methodCall.run();
+
+        List<LoggedRequest> requests = findAll(anyRequestedFor(anyUrl()));
+
+        assertThat(requests.isEmpty()).isFalse();
```
bmaizels, line 94: Nit: can we do ```assertThat(requests).isNotEmpty();``` ?

PR #1840, opened 2020-05-15
Enable endpoint discovery by default for services that support it

bmaizels:
## Description
Endpoint discovery allows the SDK to discover an endpoint based on the resource being accessed. This change makes this feature opt-out instead of opt-in for clients of services that support endpoint discovery. An additional call to the endpoint discovery service will be made when an endpoint needs to be discovered, that value is then cached and re-used.

## Motivation and Context
There may be services launched in the future that only support endpoint discovery instead of traditional static endpoints. We believe endpoint discovery will provide a better experience for customers, and therefore we want to give them that better experience out of the box without expecting them to know to enable it. Any customers that do not wish to use endpoint discovery may opt-out by setting 'endpointDiscoveryEnabled(false)' on the client builder.

## Testing
Ran all unit/functional and mock integration tests. Verified through manual integration testing that endpoint discovery calls were being made.

## Types of changes

## Checklist

## License

### DIFF
--- a/.changes/next-release/feature-AWSSDKforJavav2-534a8f5.json
+++ b/.changes/next-release/feature-AWSSDKforJavav2-534a8f5.json
@@ -1,0 +1,5 @@
+{
+    "type": "feature",
+    "category": "AWS SDK for Java v2",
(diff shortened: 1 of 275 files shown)


bmaizels:
We've decided not to do this. I'm going to adjust the logic and goals to only enable by default when endpoint discovery is required by the service and post a new PR.

PR #1843, opened 2020-05-15
Full POC for metrics system

dagnir:
POC of the metrics system for V2 incorporating the feedback from @bmaizels and @millems. Includes the metrics SPI, default implementations of these classes, and wiring up the metrics into an API call (for XML protocols).


dagnir:
Oops should be a draft.

PR #1844, opened 2020-05-15
Full POC for metrics system

dagnir:
POC of the metrics system for V2 incorporating the feedback from @bmaizels and @millems. Includes the metrics SPI, default implementations of these classes, and wiring up the metrics into an API call (for XML protocols).

### DIFF
--- a/core/metrics-spi/src/main/java/software/amazon/awssdk/metrics/MetricCategory.java
+++ b/core/metrics-spi/src/main/java/software/amazon/awssdk/metrics/MetricCategory.java
@@ -1,0 +1,82 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.metrics;
+
+import software.amazon.awssdk.annotations.SdkPublicApi;
+
+/**
+ * A enum class representing the different types of metric categories in the SDK.
+ * <p>
+ * A metric can be tagged with multiple categories. Clients can enable/disable metric collection
+ * at a {@link MetricCategory} level.
+ */
+@SdkPublicApi
+public enum MetricCategory {
+
+    /**
+     * All metrics defined by the SDK are classified under this category at a minimum. If the metrics feature is enabled
+     * but the category to collect is not, only metrics that are classified under this category are collected by the SDK
+     */
+    DEFAULT("default"),
+
+    /**
+     * Metrics collected at the http client level are classified under this category.
+     */
+    HTTP_CLIENT("httpclient"),
+
+    /**
+     * Metrics specific to streaming, eventStream APIs are classified under this category.
+     */
+    STREAMING("streaming"),
+
+    /**
+     * This is an umbrella category (provided for convenience) that records metrics belonging to every category
+     * defined in this enum. Clients who wish to collect lot of SDK metrics data should use this.
+     * <p>
+     * Note: Enabling this option is verbose and can be expensive based on the platform the metrics are uploaded to.
+     * Please make sure you need all this data before using this category.
+     */
+    ALL("all")
+
+    ;
+
+    private final String value;
+
+    MetricCategory(String value) {
+        this.value = value;
+    }
+
+    public String getValue() {
+        return value;
+    }
+
+    /**
+     * Create a {@link MetricCategory} from the given String value. This method is case insensitive.
+     *
+     * @param value the value to create the {@link MetricCategory} from
+     * @return A {@link MetricCategory} if the given {@link #value} matches one of the enum values.
+     *         Otherwise throws {@link IllegalArgumentException}
+     */
+    public static MetricCategory fromString(String value) {
+        for (MetricCategory mc : MetricCategory.values()) {
+            if (mc.value.equalsIgnoreCase(value)) {
+                return mc;
+            }
+        }
+
+        throw new IllegalArgumentException("MetricCategory cannot be created from value: " + value);
+    }
+}
--- a/core/metrics-spi/src/main/java/software/amazon/awssdk/metrics/MetricCollection.java
+++ b/core/metrics-spi/src/main/java/software/amazon/awssdk/metrics/MetricCollection.java
@@ -1,0 +1,38 @@
+package software.amazon.awssdk.metrics;
+
+import java.util.Collection;
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * An immutable collection of metrics.
+ */
+public interface MetricCollection extends Iterable<MetricRecord<?>> {
+    /**
+     * @return The name of this metric collection.
+     */
+    String name();
+
+    /**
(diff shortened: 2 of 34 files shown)


core/metrics-spi/src/main/java/software/amazon/awssdk/metrics/MetricCollection.java
```diff
@@ -0,0 +1,38 @@
...
+    <T> List<T> getAllMetricValues(SdkMetric<T> metric);
+
+    /**
+     * Return the values of the given metric.
+     *
+     * @param metric The metric.
+     * @param <T> The type of the value.
+     * @return The value of this metric.
+     */
+    <T> Optional<T> getMetricValue(SdkMetric<T> metric);
```
millems, line 32: How do these differ?

core/metrics-spi/src/main/java/software/amazon/awssdk/metrics/MetricCollection.java
```diff
@@ -0,0 +1,38 @@
+package software.amazon.awssdk.metrics;
+
+import java.util.Collection;
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * An immutable collection of metrics.
+ */
+public interface MetricCollection extends Iterable<MetricRecord<?>> {
```
millems, line 10: SdkIterable?

core/metrics-spi/src/main/java/software/amazon/awssdk/metrics/SdkMetric.java
```diff
@@ -0,0 +1,129 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import software.amazon.awssdk.annotations.SdkTestInternalApi;
+import software.amazon.awssdk.utils.Validate;
+
+/**
+ * A specific SDK metric
+ * @param <T>
+ */
+public final class SdkMetric<T> {
```
millems, line 31: Any way we can reuse some of the logic from AttributeMap.Key for this?

core/metrics-spi/src/main/java/software/amazon/awssdk/metrics/SdkMetric.java
```diff
@@ -0,0 +1,129 @@
...
+    private static final Set<SdkMetric<?>> SDK_METRICS = new HashSet<>();
```
millems, line 32: Why do we need this?

core/metrics-spi/src/main/java/software/amazon/awssdk/metrics/internal/DefaultMetricCollector.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    @Override
+    public void close() {
+        if (closed) return;
+
+        children.forEach(MetricCollector::close);
+
+        this.closed = true;
+    }
```
millems, line 94: This doesn't really do anything, except prevent publish from getting called, right? Do we really need this?

dagnir: Yeah, mostly premature optimization. I wanted to prevent cases where the whole tree is traversed unnecessarily.

core/metrics-spi/src/main/java/software/amazon/awssdk/metrics/SdkMetric.java
```diff
@@ -0,0 +1,129 @@
...
+         */
+    public static <T> SdkMetric<T> of(String name, Class<T> clzz, Set<MetricCategory> categories) {
+        Validate.noNullElements(categories, "categories must not contain null elements");
+        SdkMetric<T> event = new SdkMetric<>(name, clzz, categories);
+        synchronized (SDK_METRICS) {
+            if (SDK_METRICS.contains(event)) {
+                throw new IllegalArgumentException("Metric with name " + name + " has already been created");
+            }
+            SDK_METRICS.add(event);
+        }
```
bmaizels, line 116: Have we considered pros and cons of using ConcurrentSkipListSet here instead of synchronizing?

millems: Or `ConcurrentHashMap.newKeySet()`

core/metrics-spi/src/main/java/software/amazon/awssdk/metrics/internal/DefaultMetricCollection.java
```diff
@@ -0,0 +1,68 @@
...
+    private final String name;
+    private final Map<SdkMetric<?>, List<MetricRecord<?>>> metrics;
+    private final Collection<MetricCollection> children;
+
+
+    public DefaultMetricCollection(String name, Map<SdkMetric<?>, List<MetricRecord<?>>> metrics, Collection<MetricCollection> children) {
+        this.name = name;
+        this.metrics = metrics;
+        this.children = children != null ? Collections.unmodifiableCollection(children) : Collections.emptyList();
+    }
```
millems, line 25: We should be clear in the Javadoc that we're not doing copies of these inputs, since in other places we do that.

core/metrics-spi/src/main/java/software/amazon/awssdk/metrics/internal/DefaultMetricCollection.java
```diff
@@ -0,0 +1,68 @@
...
+                    .next()
+                    .value());
+        }
+        return Optional.empty();
+    }
+
+    @Override
+    public Collection<MetricCollection> children() {
+        return children;
+    }
```
millems, line 60: Can we do a List, so that it's ordered?

core/metrics-spi/src/main/java/software/amazon/awssdk/metrics/MetricCategory.java
```diff
@@ -0,0 +1,82 @@
...
+    STREAMING("streaming"),
+
+    /**
+     * This is an umbrella category (provided for convenience) that records metrics belonging to every category
+     * defined in this enum. Clients who wish to collect lot of SDK metrics data should use this.
+     * <p>
+     * Note: Enabling this option is verbose and can be expensive based on the platform the metrics are uploaded to.
+     * Please make sure you need all this data before using this category.
+     */
+    ALL("all")
```
millems, line 52: It feels like we have to treat "all" magically. Can't we just do MetricCategory.values() to get all of them?

core/metrics-spi/src/main/java/software/amazon/awssdk/metrics/SdkMetric.java
dagnir: Will take a look into alternative solutions that can take advantage of compare-and-set.

core/metrics-spi/src/main/java/software/amazon/awssdk/metrics/MetricCollection.java
dagnir: Discussed this in the meeting, we can get rid of the single value getter.

core/metrics-spi/src/main/java/software/amazon/awssdk/metrics/SdkMetric.java
dagnir: +1

core/metrics-spi/src/main/java/software/amazon/awssdk/metrics/SdkMetric.java
dagnir: Yep, these are modeled after AttributeMap.Key, so it should be possible

core/metrics-spi/src/main/java/software/amazon/awssdk/metrics/internal/DefaultMetricCollection.java
dagnir: Sure, I think that could be valuable

PR #1845, opened 2020-05-15
Default endpointDiscovery to true for services that require it

bmaizels:
1) Services that have at least one operation that requires endpoint discovery will have their client builders default endpoint discovery to true.
2) The method 'enableEndpointDiscovery()' has been deprecated in client builders that support endpoint discovery.
3) The method 'endpointDiscoveryEnabled(boolean)' has been added to client builders that support endpoint discovery.

This behavior should now match the v1 SDK.

## License


codegen/src/main/java/software/amazon/awssdk/codegen/IntermediateModelBuilder.java
```diff
@@ -107,12 +107,20 @@ public IntermediateModel build() {
             new HashMap<>(new AddCustomAuthorizers(this.service, getNamingStrategy()).constructAuthorizers());
 
         OperationModel endpointOperation = null;
+        boolean endpointCacheRequired = false;
 
         for (OperationModel o : operations.values()) {
             if (o.isEndpointOperation()) {
                 endpointOperation = o;
-                break;
             }
```
millems, line 115: We just use the last endpoint operation? Is that the right behavior?

codegen/src/main/java/software/amazon/awssdk/codegen/IntermediateModelBuilder.java
```diff
@@ -107,12 +107,20 @@ public IntermediateModel build() {
...
+
+            if (o.getEndpointDiscovery() != null && o.getEndpointDiscovery().isRequired()) {
+                endpointCacheRequired = true;
```
millems, line 118: So the "new" behavior is: if any operation requires endpoint discovery, enable it by default?

codegen/src/main/java/software/amazon/awssdk/codegen/poet/builder/AsyncClientBuilderClass.java
```diff
@@ -54,12 +54,24 @@ public TypeSpec poetSpec() {
...
+                         .addModifiers(Modifier.PUBLIC)
+                         .returns(builderClassName)
+                         .addParameter(boolean.class, "endpointDiscoveryEnabled")
+                         .addStatement("this.endpointDiscoveryEnabled = endpointDiscoveryEnabled")
+                         .addStatement("return this")
+                         .build();
+    }
+
     private MethodSpec enableEndpointDiscovery() {
         return MethodSpec.methodBuilder("enableEndpointDiscovery")
```
millems, line 76: Deprecate here as well? Can we also remove it for new services?

codegen/src/main/java/software/amazon/awssdk/codegen/IntermediateModelBuilder.java
bmaizels: Yes, exactly. I also stated that in the PR description.

codegen/src/main/java/software/amazon/awssdk/codegen/IntermediateModelBuilder.java
bmaizels: I'll admit this is not how I would have written it if I had done it myself, but it is actually a copy/paste from the v1 SDK. The logic is correct because it only needs to know there is at least one endpoint operation (and the last one suffices for that purpose), and the flag as to whether an endpoint is required is cumulative (once it gets set it doesn't get unset) so ends up effectively meaning 'at least one endpoint is required'.

bmaizels: I added a couple of tests to assert this logic was working correctly as well.

codegen/src/main/java/software/amazon/awssdk/codegen/poet/builder/AsyncClientBuilderClass.java
bmaizels: I will deprecate here as well. Do you have a strategy for removing it for new services? How do we identify 'new services'? One strategy is just to say 'service != DDB' which seems yuck.

codegen/src/main/java/software/amazon/awssdk/codegen/IntermediateModelBuilder.java
millems: Can we clarify it a bit and make it a boolean or something?

codegen/src/main/java/software/amazon/awssdk/codegen/poet/builder/AsyncClientBuilderClass.java
millems: We could also add a customization: "enableEndpointDiscoveryMethodRequired" and just enable it for DynamoDB.

codegen/src/main/java/software/amazon/awssdk/codegen/IntermediateModelBuilder.java
bmaizels: Changing it will require more refactoring than the change is worth in my opinion. At least it's in sync with v1 in this state. I'll add comments to explain what it's doing.

PR #1854, opened 2020-05-19
Update docs based on latest discussions

dagnir:
Update of the design docs to reflect the the changes discussed with the rest of team. See #1844 for full context on what we looked at during the discussion as well as comments from @millems and @bmaizels that are reflected in the changes to the SPI here.


docs/design/core/metrics/prototype/MetricPublisher.java
```diff
@@ -1,68 +1,48 @@
...
+ * use.  As long as a publisher is not closed, then it can receive {@code
+ * MetricCollection} objects at any time. In addition, as the SDK makes use of
+ * multithreading, it's possible that the publisher is shared concurrently by
+ * multiple threads, and necessitates that all implementations are threadsafe.
  * <p>
- *     In certain situations (high throttling errors, metrics are reported faster than publishing etc), storing all the metrics
- *     might take up lot of memory and can crash the application. In these cases, it is recommended to have a max limit on
- *     number of metrics stored or memory used for metrics and drop the metrics when the limit is breached.
- * </p>
+ * The SDK may invoke methods on the interface from multiple threads
```
cenedhryn, line 14: Seems to be a repeat of the multithreading info

PR #1855, opened 2020-05-20
Add extra note about limit_key in javadocs and change the test files to match the modification

Quanzzzz:
Add extra note about limit_key in javadocs and change the test files to match the modification
## Description

Introduced ```PaginatorDefinition``` into ```PaginatorDocs``` and used the limit_key in the ```PaginatorDefinition``` as part of the newly added note in the javadocs;
Changed the related test files to match this modification;

Example of the newly added note (with operation name "PaginatedOperationWithResultKey"):
```Please notice that the configuration of MaxResults won't limit the number of results you get with the paginator, instead, it only limits the number of results in each page.```

## Testing

```mvn clean install -pl :codegen --am``` passed

## License

### CODE REVIEW

File: codegen/src/main/java/software/amazon/awssdk/codegen/docs/PaginationDocs.java [hunk 1]
```diff
@@ -238,6 +244,18 @@ private ClassName asyncPaginatedResponseType() {
...
     }
 
+    private String getPaginatorLimitKeyName() {
+        return paginatorDefinition != null ? paginatorDefinition.getLimitKey() : "";
+    }
+
+    private CodeBlock noteAboutLimitConfigurationMethod() {
+        return CodeBlock.builder()
+                        .add("\n<p><b>Please notice that the configuration of $L won't limit the number of results "
+                             + "you get with the paginator, instead, it only limits the number of results in each page.</b></p>",
```
millems, line 254: Editorial Nit:
Please note that the configuration of $L won't limit the number of results you get with the paginator. It only limits the number of results in each page.

PR #1884, opened 2020-06-09
Async metrics Part1: Add metrics codegen changes for async clients

zoewangg:
## Description
Add metrics codegen changes for async clients

## License


codegen/src/test/resources/software/amazon/awssdk/codegen/poet/client/test-endpoint-discovery-async.java
```diff
@@ -264,6 +288,10 @@ public final String serviceName() {
             return executeFuture;
         } catch (Throwable t) {
             return CompletableFutureUtils.failedFuture(t);
+        } finally {
+            Optional<MetricPublisher> metricPublisher = MetricUtils.resolvePublisher(clientConfiguration,
+                    testDiscoveryRequiredRequest);
+            metricPublisher.ifPresent(p -> p.publish(apiCallMetricCollector.collect()));
```
dagnir, line 294: Publishing needs to happen after execution so we would need to add a completion stage to the future instead of publishing here

zoewangg: oh yeah, I guess I just blindly copied the sync change 😅

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/util/MetricUtils.java
```diff
@@ -54,6 +56,19 @@ private MetricUtils() {
...
     }
 
+    /**
+     * Resolve the correct metric publisher to use. The publisher set on the request always takes precedence.
+     *
+     * @param clientConfig The client configuration.
+     * @param requestConfig The request override configuration.
+     * @return The metric publisher to use.
+     */
+    public static Optional<MetricPublisher> resolvePublisher(SdkClientConfiguration clientConfig,
```
zoewangg, line 66: I created this overload because for some operations, the input request is modified (eg: applySignerOverride), so the request variable is not final. As a result, we can't use the input request parameter in the `whenComplete` block because variables used in lambda should be final or effectively final

dagnir: Not a fan of using `Optional` as a method parameter. Can we change to so it takes an override config, and change the operation code:

```java
MetricUtils.resolvePublisher(clientConfig, overrideConfig.orElse(null));
```
```

Comment (zoewangg) [reply core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/util/MetricUtils.java:66]: In most cases, I'd agree on the use of optional as a method parameter, but I can't find a better way to handle it in this case.

`MetricUtils.resolvePublisher(clientConfig, overrideConfig.orElse(null));` seems a bit code smell to me. 

Ideally, we should create another overload which doesn't take a requestOverrideConfig, but it seems overkill here.
```java
MetricUtils.resolvePublisher(SdkClientConfiguration clientConfig, AwsRequestOverrideConfiguration overrideConfig)
MetricUtils.resolvePublisher(SdkClientConfiguration clientConfig)
```

PR #1885, opened 2020-06-09
Added support for HTTP/2 metrics to Netty.

millems:
Updated existing HTTP metrics to specify that they are "concurrency", not "connections". For HTTP/2, we can't calculate the maximum number of connections, and concurrency is more useful to customers anyway.

Fixed the AVAILABLE_CONCURRENCY in Netty to actually be the number of established but idle concurrency, not the difference between max concurrency and leased concurrency.

### DIFF
--- a/http-client-spi/src/main/java/software/amazon/awssdk/http/HttpMetric.java
+++ b/http-client-spi/src/main/java/software/amazon/awssdk/http/HttpMetric.java
@@ -20,7 +20,8 @@ import software.amazon.awssdk.metrics.MetricCategory;
 import software.amazon.awssdk.metrics.SdkMetric;
 
 /**
- * Metrics collected by HTTP clients.
+ * Metrics collected by HTTP clients for HTTP/1 and HTTP/2 operations. See {@link Http2Metric} for metrics that are only available
+ * on HTTP/2 operations.
  */
 @SdkPublicApi
 public final class HttpMetric {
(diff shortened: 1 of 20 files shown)


http-client-spi/src/main/java/software/amazon/awssdk/http/HttpMetric.java
```diff
@@ -30,24 +31,64 @@
...
+     * For HTTP/2 operations, this is equal to the maximum number of streams that can be pooled by the HTTP client.
+     *
+     * <p>Note: Depending on the HTTP client, this is either a value for all endpoints served by the HTTP client, or a value
+     * that applies only to the specific endpoint/host used in the request. For 'apache-http-client', this value is
+     * for the entire HTTP client. For 'netty-nio-client', this value is per-endpoint. In all cases, this value is scoped to an
+     * individual HTTP client instance, and does not include concurrency that may be available in other HTTP clients running
+     * within the same JVM.
      */
-    public static final SdkMetric<Integer> MAX_CONNECTIONS = metric("MaxConnections", Integer.class);
+    public static final SdkMetric<Integer> MAX_CONCURRENCY = metric("MaxConcurrency", Integer.class);
```
zoewangg, line 45: Seems like for h2 operations,  all metrics here are for streams, should we add metrics for connections?

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/IdleConnectionCountingChannelPool.java
```diff
@@ -0,0 +1,234 @@
...
+ */
+@SdkInternalApi
+public class IdleConnectionCountingChannelPool implements SdkChannelPool {
+    private static final Logger log = Logger.loggerFor(IdleConnectionCountingChannelPool.class);
+
+    /**
+     * The idle channel state for a specific channel. This should only be accessed from the {@link #executor}.
+     */
+    private static final AttributeKey<ChannelIdleState> CHANNEL_STATE =
+        AttributeKey.newInstance("IdleConnectionCountingChannelPool.CHANNEL_STATE");
```
zoewangg, line 46: should we namespace the key?

millems: What do you mean?

http-client-spi/src/main/java/software/amazon/awssdk/http/HttpMetric.java
millems: I'm not entirely confident that it's particularly useful to do so, but it wouldn't be difficult. Maybe we do it if we find it's needed? I'm worried about confusing people if we expose both.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/IdleConnectionCountingChannelPool.java
zoewangg: I meant prefixing it with sdk namespace so it doesn't conflict with other keys. It seems we are using `NettyUtils.getOrCreateAttributeKey` now introduced in #1887

http-client-spi/src/main/java/software/amazon/awssdk/http/HttpMetric.java
zoewangg: I think leasedConnection and availableConnection will provide helpful insights when we troubleshoot async issues. For example, in the past, we spent a lot of time searching logs and found out the streams are not balanced allocated across all connections. It'd save us time if we have such metrics available.

PR #1890, opened 2020-06-11
Stricter implementation of ARN rules when parsing resource strings. R…

bmaizels:
…equired to support paths properly in S3 operations that take ARNs

## License


cenedhryn (approved):

giacatno68 (approved):

millems:
Request is stale and was never merged. We'll re-open if this is needed.

PR #1895, opened 2020-06-12
Async core metrics part3: add tests to verify async core metrics for event streaming apis

zoewangg:
Add functional and integ tests to verify async core metrics for event streaming operation

## License

### DIFF
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/util/MetricUtils.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/util/MetricUtils.java
@@ -16,6 +16,7 @@
 package software.amazon.awssdk.core.internal.util;
 
 import static software.amazon.awssdk.core.client.config.SdkClientOption.METRIC_PUBLISHER;
+import static software.amazon.awssdk.core.http.HttpResponseHandler.X_AMZN_REQUEST_ID_HEADER;
 
 import java.time.Duration;
 import java.util.Optional;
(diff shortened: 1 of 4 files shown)


core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/util/MetricUtils.java
```diff
@@ -106,6 +107,8 @@ public static void collectHttpMetrics(MetricCollector metricCollector, SdkHttpFu
         metricCollector.reportMetric(CoreMetric.HTTP_STATUS_CODE, httpResponse.statusCode());
         httpResponse.firstMatchingHeader("x-amz-request-id")
                     .ifPresent(v -> metricCollector.reportMetric(CoreMetric.AWS_REQUEST_ID, v));
+        httpResponse.firstMatchingHeader(X_AMZN_REQUEST_ID_HEADER)
+                    .ifPresent(v -> metricCollector.reportMetric(CoreMetric.AWS_REQUEST_ID, v));
```
cenedhryn, line 111: Is this meant to be request id header, not request id?

zoewangg: It's the header name for request Ids with different format.

Currently, we have two formats I think
```
x-amzn-RequestId
x-amz-request-id
```

cenedhryn: ok!

Issue #1811, opened 2020-05-03
Mimetype.getMimeType() incorrectly returns `null`


garretwilson:
The documentation for `software.amazon.awssdk.core.internal.util.getMimetype(Path path)` says that the method will return `application/octet-stream` if the MIME type cannot be determined from the filename:

```java
    /**
     * Determines the mimetype of a file by looking up the file's extension in an internal listing
     * to find the corresponding mime type. If the file has no extension, or the extension is not
     * available in the listing contained in this class, the default mimetype
     * <code>application/octet-stream</code> is returned.
     * <p>
     * A file extension is one or more characters that occur after the last period (.) in the file's name.
     * If a file has no extension,
     * Guesses the mimetype of file data based on the file's extension.
     *
     * @param path the file whose extension may match a known mimetype.
     * @return the file's mimetype based on its extension, or a default value of
     * <code>application/octet-stream</code> if a mime type value cannot be found.
     */
    public String getMimetype(Path path) {
```

Most of the work is performed by `getMimetype(String fileName)`, which _does_ return `application/octet-stream` if the the MIME type could not be determined from the given:

```
    public String getMimetype(Path path) {
        Validate.notNull(path, "path");
        Path file = path.getFileName();

        if (file != null) {
            return getMimetype(file.toString());
        }
        return null;
    }
```

The problem is if the original `Path` doesn't have a filename at all. In that case you return `null`, breaking your API contract.

It is highly unlikely that a given `Path` would _not_ have a filename, and indeed this is probably why no one has ran into this bug before. Nevertheless the fact that it is rare doesn't change the fact that it breaks the API. The API contract says nothing of returning `null`, and indeed the code that uses this, `RequestBody.fromFile(Path path)`, would throw a `NullPointerException` if this method did return `null`.

You need to decide what you intend to do if the given path doesn't have a filename. If you don't intend to support paths without filenames, you should state that in the API contract and deal with it:

```
    public String getMimetype(Path path) {
        Validate.notNull(path, "path");
        Path file = path.getFileName();

        if (file != null) {
            return getMimetype(file.toString());
        }
        throw new IllegalArgumentException("Path "+ path + " has no filename.");
    }
```

Or if you merely intend for a missing filename to also result in a `application/octet-stream` (which better matches what the current API implies), you could do that:

```
    public String getMimetype(Path path) {
        Validate.notNull(path, "path");
        Path file = path.getFileName();

        if (file != null) {
[...]

PR #1901, opened 2020-06-15
Fixed wrong Mimetype result for null fileName and deleted wrongly pasted javadocs

Quanzzzz:
Fixed wrong Mimetype result for null fileName to keep consistency with our API contract;
Deleted wrongly pasted javadocs;

## Motivation and Context

#1811 
## Testing

There is no newly added unit test for this change because it's impossible to set the fileName as null here in the path.

## Types of changes

## License


core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/util/Mimetype.java
```diff
@@ -132,18 +128,14 @@ public String getMimetype(Path path) {
         if (file != null) {
             return getMimetype(file.toString());
         }
-        return null;
+        return MIMETYPE_OCTET_STREAM;
```
bmaizels, line 131: Looks like this behavior is missing a test?

PR #1882, opened 2020-06-08
Support core metrics for sync clients

dagnir:
## Description
This commits adds collection of [core metrics](https://github.com/dagnir/aws-sdk-java-v2/blob/core-metrics-pt2/core/sdk-core/src/main/java/software/amazon/awssdk/core/metrics/CoreMetric.java#L24) for sync clients.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)

## Checklist

- [ ] I have read the **CONTRIBUTING** document
- [ ] Local run of `mvn install` succeeds
- [ ] My code follows the code style of this project
- [ ] My change requires a change to the Javadoc documentation
- [ ] I have updated the Javadoc documentation accordingly
- [ ] I have read the **README** document
- [ ] I have added tests to cover my changes
- [ ] All new and existing tests passed
- [ ] A short description of the change has been added to the **CHANGELOG**
- [ ] My change is to implement 1.11 parity feature and I have updated [LaunchChangelog](https://github.com/aws/aws-sdk-java-v2/blob/master/docs/LaunchChangelog.md)

## License

- [x] I confirm that this pull request can be released under the Apache 2 license

PR #1906, opened 2020-06-17
Metric utils test

dagnir:
## Description
Follow-up on review comment on #1882

## Types of changes

## Checklist

## License


core/sdk-core/src/test/java/software/amazon/awssdk/core/internal/util/MetricUtilsTest.java
```diff
@@ -0,0 +1,144 @@
...
+                .putHeader("x-amz-request-id", requestId)
+                .putHeader(HttpResponseHandler.X_AMZN_REQUEST_ID_HEADER, amznRequestId)
+                .putHeader(HttpResponseHandler.X_AMZ_ID_2_HEADER, requestId2)
+                .build();
+
+        MetricUtils.collectHttpMetrics(mockCollector, response);
+
+        verify(mockCollector).reportMetric(CoreMetric.HTTP_STATUS_CODE, statusCode);
+        verify(mockCollector).reportMetric(CoreMetric.AWS_REQUEST_ID, requestId);
+        verify(mockCollector).reportMetric(CoreMetric.AWS_REQUEST_ID, amznRequestId);
```
cenedhryn, line 120: Why do we report the same metric with two different values?

dagnir: Depending on the service, the Header used for the request ID can be differ between these two values. This is just to ensure we're catching both cases, I don't think a service will use both.

cenedhryn: ok!

PR #1907, opened 2020-06-18
Add CloudWatchMetricPublisher, a MetricPublisher implementation that uploads metric data to CloudWatch.

millems:


metric-publishers/cloudwatch-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/cloudwatch/internal/transform/MetricAggregator.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    /**
+     * Execute the provided consumer if this {@code MetricAggregator} is a {@link DetailedMetricAggregator}.
+     */
+    default void ifDetailed(Consumer<DetailedMetricAggregator> detailsConsumer) {
+        if (this instanceof DetailedMetricAggregator) {
+            detailsConsumer.accept((DetailedMetricAggregator) this);
+        }
+    }
```
bmaizels, line 77: Not sure about this, seems like an OO smell that we have to put implementation specific logic (reflection at that!) inside an abstract interface. Can we rethink this do you have any other ideas how we can do this without breaking encapsulation/layering?

metric-publishers/cloudwatch-metric-publisher/src/test/java/software/amazon/awssdk/metrics/publishers/cloudwatch/CloudWatchMetricPublisherTest.java
```diff
@@ -0,0 +1,219 @@
...
+import software.amazon.awssdk.metrics.MetricCollector;
+import software.amazon.awssdk.metrics.SdkMetric;
+import software.amazon.awssdk.metrics.publishers.cloudwatch.internal.transform.MetricCollectionAggregator;
+import software.amazon.awssdk.services.cloudwatch.CloudWatchAsyncClient;
+import software.amazon.awssdk.services.cloudwatch.model.Dimension;
+import software.amazon.awssdk.services.cloudwatch.model.MetricDatum;
+import software.amazon.awssdk.services.cloudwatch.model.PutMetricDataRequest;
+import software.amazon.awssdk.services.cloudwatch.model.PutMetricDataResponse;
+
+public class CloudWatchMetricPublisherTest {
```
bmaizels, line 42: Do you think it's possible to test the shutdown behavior? That's the part of this class that actually worries me the most.

metric-publishers/cloudwatch-metric-publisher/src/test/java/software/amazon/awssdk/metrics/publishers/cloudwatch/internal/transform/MetricCollectionAggregatorTest.java
```diff
@@ -0,0 +1,403 @@
...
+                                                                               .maximum(4.0)
+                                                                               .sum(14.0)
+                                                                               .sampleCount(5.0)
+                                                                               .build());
+            });
+        });
+    }
+
+    @Test
+    public void detailedMetricsAreCorrect() {
```
bmaizels, line 227: Can we have tests that probe the boundary conditions to do with upload limits, like what happens if the publish is on the boundary, just over and just under kinda thing?

metric-publishers/cloudwatch-metric-publisher/src/test/resources/log4j.properties
```diff
@@ -0,0 +1,35 @@
...
+#
+#  http://aws.amazon.com/apache2.0
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+#
+
+log4j.rootLogger=INFO, A1
```
bmaizels, line 16: What are these ^M artifacts?

metric-publishers/cloudwatch-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/cloudwatch/CloudWatchMetricPublisher.java
```diff
@@ -0,0 +1,507 @@
...
+         *
+         * <p>All {@link SdkMetric}s are associated with at least one {@code MetricCategory}. This setting determines which
+         * category of metrics uploaded to CloudWatch. Any metrics {@link #publish(MetricCollection)}ed that do not fall under
+         * these configured categories are ignored.
+         *
+         * <p>Note: If there are {@link #dimensions(Collection)} configured that do not fall under these {@code MetricCategory}
+         * values, the dimensions will NOT be ignored. In other words, the metric category configuration only affects which
+         * metrics are uploaded to CloudWatch, not which values can be used for {@code dimensions}.
+         */
+        public Builder metricCategories(Collection<MetricCategory> metricCategories) {
```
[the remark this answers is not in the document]
millems: I don't have an opinion either way. Save it for the surface area review? I can't think of something that isn't too long.

metric-publishers/cloudwatch-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/cloudwatch/CloudWatchMetricPublisher.java
```diff
@@ -0,0 +1,507 @@
...
+         * is recommended to minimize the number of {@code detailedMetrics} configured on the publisher, or to upload metric data
+         * more frequently. As with all performance and resource concerns, profiling in a production-like environment is
+         * encouraged.
+         *
+         * <p>In addition to additional heap memory usage, detailed metrics can result in more requests being sent to CloudWatch,
+         * which can also introduce additional usage cost. The {@link #maximumCallsPerUpload(Integer)} acts as a safeguard against
+         * too many calls being made, but if you configure multiple {@code detailedMetrics}, you may need to increase the
+         * {@code maximumCallsPerUpload} limit.
+         */
+        public Builder detailedMetrics(Collection<SdkMetric<?>> detailedMetrics) {
```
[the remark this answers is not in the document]
millems: I mentioned on the call that I tried this and it felt like I was being too cute. I can try again, though.

metric-publishers/cloudwatch-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/cloudwatch/internal/task/UploadMetricsTasks.java
```diff
@@ -0,0 +1,65 @@
...
+        List<PutMetricDataRequest> requests = allRequests;
+        if (requests.size() > maximumRequestsPerFlush) {
+            METRIC_LOGGER.warn(() -> "Maximum AWS SDK client-side metric call count exceeded: " + allRequests.size() +
+                                     " > " + maximumRequestsPerFlush + ". Some metric requests will be dropped. This occurs when "
+                                     + "the caller has configured too many metrics or too unique of dimensions without an "
+                                     + "associated increase in the maximum-calls-per-upload configured on the publisher.");
+
+            // Randomly pick which requests we intend to keep/drop so that we aren't always dropping the same ones if this happens
+            // a lot of times in a row.
+            Collections.shuffle(requests);
```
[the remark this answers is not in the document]
millems: I don't think that's what we concluded, but to avoid arguing, I'll just remove it.

metric-publishers/cloudwatch-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/cloudwatch/internal/transform/MetricAggregator.java
millems: We talked about this already on the call. Let's discuss offline.

metric-publishers/cloudwatch-metric-publisher/src/test/java/software/amazon/awssdk/metrics/publishers/cloudwatch/CloudWatchMetricPublisherTest.java
millems: What specific part worries you? All of the tests currently rely on the close behavior flushing out the executor pool. Almost every test in this class would fail if it didn't work. The only thing we don't test is the timeout behavior, which would require a non-public constructor.

metric-publishers/cloudwatch-metric-publisher/src/test/java/software/amazon/awssdk/metrics/publishers/cloudwatch/internal/transform/MetricCollectionAggregatorTest.java
millems: How is that different than `MetricCollectionAggregatorTest#maximumRequestsIsHonored` and `MetricCollectionAggregatorTest#maximumMetricValuesIsHonored`?

metric-publishers/cloudwatch-metric-publisher/src/test/resources/log4j.properties
millems: No idea, I copied this file from another module. Maybe a different kind of line ending?

metric-publishers/cloudwatch-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/cloudwatch/internal/transform/MetricAggregator.java
millems: Summarizing here for posterity:

```
The alternatives are: polymorphism, a visitor and what I did. 

The first was the worst in terms of 'separation of concerns', because it leaked the cloudwatch concerns of limits and request building to something that was on the 'aggregate' path, which to this point is cloudwatch agnostic.

Visitor is the 'standard' solution for the double-dispatch needed here, but I didn't like it because creating an anonymous class too greatly impacted the readability of the callsite with a lot of boilerplate and distraction.

What I did was a compromise on the visitor solution, because it allowed me to use lambdas, which avoided the boilerplate issue and avoided needing to define a 'visitor' class.

File: metric-publishers/cloudwatch-metric-publisher/src/test/java/software/amazon/awssdk/metrics/publishers/cloudwatch/internal/transform/MetricCollectionAggregatorTest.java (hunk 3 above)
Comment (bmaizels) [reply metric-publishers/cloudwatch-metric-publisher/src/test/java/software/amazon/awssdk/metrics/publishers/cloudwatch/internal/transform/MetricCollectionAggregatorTest.java:227]: Oh yeah true, do you think we should be testing the aggregation/values published in those tests as well as the counts?

File: metric-publishers/cloudwatch-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/cloudwatch/CloudWatchMetricPublisher.java (hunk 5 above)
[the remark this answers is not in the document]
Comment (bmaizels) [reply metric-publishers/cloudwatch-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/cloudwatch/CloudWatchMetricPublisher.java:454]: ok

File: metric-publishers/cloudwatch-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/cloudwatch/CloudWatchMetricPublisher.java (hunk 6 above)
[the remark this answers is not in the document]
Comment (bmaizels) [reply metric-publishers/cloudwatch-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/cloudwatch/CloudWatchMetricPublisher.java:488]: We can defer to the surface review? I think it'd be a good thing for future extensibility.

File: metric-publishers/cloudwatch-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/cloudwatch/CloudWatchMetricPublisher.java [hunk 8]
```diff
@@ -0,0 +1,507 @@
...
+    }
+
+    private int resolveMaximumCallsPerUpload(Builder builder) {
+        return builder.maximumCallsPerUpload == null ? DEFAULT_MAXIMUM_CALLS_PER_UPLOAD : builder.maximumCallsPerUpload;
+    }
+
+    @Override
+    public void publish(MetricCollection metricCollection) {
+        try {
+            executor.submit(new AggregateMetricsTask(metricAggregator, metricCollection));
```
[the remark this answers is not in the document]
bmaizels: As discussed I don't consider this feedback to be blocking.

metric-publishers/cloudwatch-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/cloudwatch/internal/transform/MetricAggregator.java
bmaizels: Okay fair enough looks like you gave a few alternatives there some serious consideration! If I had done it I might have gone with the visitor pattern, but I don't think we should block on this concern given these are internal classes.

metric-publishers/cloudwatch-metric-publisher/src/test/java/software/amazon/awssdk/metrics/publishers/cloudwatch/CloudWatchMetricPublisherTest.java
bmaizels: What happens if a thread gets stuck? That'd be the timeout case I guess. What happens if the thread is interrupted during shutdown?

metric-publishers/cloudwatch-metric-publisher/src/test/resources/log4j.properties
bmaizels: Hopefully easy to 'fix'?

metric-publishers/cloudwatch-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/cloudwatch/CloudWatchMetricPublisher.java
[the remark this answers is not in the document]
millems: Let's merge and then add a task to prototype alternatives, to make sure we don't forget during the surface review.

metric-publishers/cloudwatch-metric-publisher/src/test/java/software/amazon/awssdk/metrics/publishers/cloudwatch/CloudWatchMetricPublisherTest.java
millems: +1, the interrupted case we can test. Timeout case is definitely harder.

metric-publishers/cloudwatch-metric-publisher/src/test/java/software/amazon/awssdk/metrics/publishers/cloudwatch/internal/transform/MetricCollectionAggregatorTest.java
millems: I'll try it out. I don't think that's covered elsewhere.

metric-publishers/cloudwatch-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/cloudwatch/internal/task/UploadMetricsTasks.java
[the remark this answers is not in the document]
bmaizels: I think our options are 1) send nothing  2) send as much as we can but undetermined order  3) send as much as we can but determined order   4) send as much as we can but actively shuffle. I think no matter what option we pick the settings need to be changed as this isn't a good situation for the customer. I'm almost leaning towards 1) under the assumption that no data might be better than incomplete/misleading data especially where aggregates are concerned but 2 might be more appropriate because it wouldn't fail all the time necessarily so there may be partial data anyway. Doing 3 or 4 seems like we're saying this is another kind of happy path and I don't think I'm ready to agree with that at this point.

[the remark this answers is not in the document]
bmaizels: Crazy idea but can we buffer and try and send them in subsequent requests? Obviously the buffer can fill up at which point we really have to drop stuff on the ground, but it would give the system a bit of elasticity to deal with spikes.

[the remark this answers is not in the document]
millems: (3) is also quite difficult to implement. (1) could still result in inconsistent values, since previous and future flushes could include data from the 'time bucket' of the dropped values. I think (2) or (4) are both probably fine, from the customer perspective. In that case (2)  might win since it's less code.

PR #1914, opened 2020-06-22
Fixed a cloudwatch-metric-publisher issue where very small values would result in a failure calling cloudwatch.

millems:


metric-publishers/cloudwatch-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/cloudwatch/internal/transform/MetricValueNormalizer.java
```diff
@@ -0,0 +1,45 @@
...
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.metrics.publishers.cloudwatch.internal.transform;
+
+import software.amazon.awssdk.annotations.SdkInternalApi;
+
+@SdkInternalApi
+class MetricValueNormalizer {
```
dagnir, line 21: nit final

PR #1921, opened 2020-06-22
Add HttpClient collector to HTTP request

dagnir:
## Types of changes

## Checklist

## License

### DIFF
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/MakeHttpRequestStage.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/MakeHttpRequestStage.java
@@ -59,16 +59,19 @@ public class MakeHttpRequestStage
     }
 
     private HttpExecuteResponse executeHttpRequest(SdkHttpFullRequest request, RequestExecutionContext context) throws Exception {
+        MetricCollector metricCollector = context.metricCollector();
+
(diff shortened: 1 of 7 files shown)


core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/MakeHttpRequestStage.java
```diff
@@ -59,17 +59,20 @@ public MakeHttpRequestStage(HttpClientDependencies dependencies) {
     }
 
     private HttpExecuteResponse executeHttpRequest(SdkHttpFullRequest request, RequestExecutionContext context) throws Exception {
+        MetricCollector metricCollector = context.metricCollector();
```
Quanzzzz, line 62: Not a comment, just a small question. What's the difference between these two collectors?

dagnir: `metricCollector` here is the collector for all API call attempt metrics, the name is `ApiCallAttempt`. The code is creating a child collector of this collector, `HttpClient` and passing that as a collector for the HTTP client so that any metrics the HTTP client reports are added to this collector instead so it's clear from the node path later, `ApiCallAttempt > HttpClient` that the metrics were reported by the HTTP client.

Quanzzzz: It's clear now. Thanks.

PR #1926, opened 2020-06-24
Updated the metrics design to include details on how metrics can be enabled at the request level, client level and global level.

millems:


docs/design/core/metrics/Design.md
```diff
@@ -60,107 +60,174 @@ standard metrics collected by the SDK.
...
+
+This option is enabled using an environment variable or system property. If both are specified, the system property 
+will be used. If metrics are enabled at the client level using **Option 2** above, this option is ignored. Overriding 
+the metric publisher at request time using **Option 1** overrides any publishers that have been enabled globally.
+
+**System Property:** `aws.metricPublishingEnabled=true` 
+
+**Environment Variable:** `AWS_METRIC_PUBLISHING_ENABLED=true`
+
+The value specified must be one of `"true"` or `"false"`. Specifying any other string values will result in
```
zoewangg, line 132: Should we make it case insensitive?

millems: +1, as we talked about, though, we should use what other booleans do, first and foremost.

PR #1930, opened 2020-06-29
Mutiple metrics fixes

dagnir:
## Description
 - Allow lists of publishers to be set.

   Allow setting of metric pubishers to be set on the client of request override
   rather than a single publisher, making using multiple MetricPublishers per
   client or request easier.

 - Remove any non-core usage of MetricUtils.

 - Fix metrics test in TranscribeStreamIntegrationTest.

## Types of changes

## Checklist

## License

### DIFF
--- a/codegen/src/test/resources/software/amazon/awssdk/codegen/poet/client/test-async-client-class.java
+++ b/codegen/src/test/resources/software/amazon/awssdk/codegen/poet/client/test-async-client-class.java
@@ -3,7 +3,8 @@ package software.amazon.awssdk.services.json;
 import static software.amazon.awssdk.utils.FunctionalUtils.runAndLogError;
 
 import java.nio.ByteBuffer;
-import java.util.Optional;
+import java.util.Collections;
+import java.util.List;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.Executor;
 import java.util.function.Consumer;
@@ -23,6 +24,7 @@ import software.amazon.awssdk.awscore.eventstream.EventStreamTaggedUnionPojoSupp
 import software.amazon.awssdk.awscore.eventstream.RestEventStreamAsyncResponseTransformer;
 import software.amazon.awssdk.awscore.exception.AwsServiceException;
 import software.amazon.awssdk.core.ApiName;
+import software.amazon.awssdk.core.RequestOverrideConfiguration;
 import software.amazon.awssdk.core.SdkPojoBuilder;
 import software.amazon.awssdk.core.SdkResponse;
 import software.amazon.awssdk.core.async.AsyncRequestBody;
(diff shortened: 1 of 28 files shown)


codegen/src/test/resources/software/amazon/awssdk/codegen/poet/client/test-async-client-class.java
```diff
@@ -403,15 +402,14 @@ public final String serviceName() {
...
-                Optional<MetricPublisher> metricPublisher = MetricUtils.resolvePublisher(clientConfiguration,
-                        requestOverrideConfig);
-                metricPublisher.ifPresent(p -> p.publish(apiCallMetricCollector.collect()));
+                List<MetricPublisher> metricPublishers = resolveMetricPublishers(clientConfiguration, requestOverrideConfig);
+                metricPublishers.forEach(p -> p.publish(apiCallMetricCollector.collect()));
             });
             return executeFuture;
         } catch (Throwable t) {
-            Optional<MetricPublisher> metricPublisher = MetricUtils.resolvePublisher(clientConfiguration,
+            List<MetricPublisher> metricPublisher = resolveMetricPublishers(clientConfiguration,
```
zoewangg, line 410: `metricPublishers`?

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/client/test-async-client-class.java
```diff
@@ -1070,6 +1061,21 @@ public void close() {
                                 .exceptionBuilderSupplier(InvalidInputException::builder).httpStatusCode(400).build());
     }
 
+    private static List<MetricPublisher> resolveMetricPublishers(SdkClientConfiguration clientConfiguration,
```
zoewangg, line 1064: Out of curiosity, why do we move this method out of  `MetricUtils`?

test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/metrics/async/AsyncClientMetricPublisherResolutionTest.java
```diff
@@ -0,0 +1,160 @@
...
+
+    @Test
+    public void testApiCall_publishersSetOnClient_clientPublishersInvoked() throws IOException {
+        MetricPublisher publisher1 = mock(MetricPublisher.class);
+        MetricPublisher publisher2 = mock(MetricPublisher.class);
+
+        client = clientWithPublishers(publisher1, publisher2);
+
+        try {
+            client.allTypes().join();
```
zoewangg, line 97: minor: missing `fail("no exception thrown")`, same as other tests.

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/client/test-async-client-class.java
dagnir: Discussed this with @millems. `MetricUtils` is `@SdkInternal`, and we'd rather keep it that way than expose it as a protected API.

test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/metrics/async/AsyncClientMetricPublisherResolutionTest.java
dagnir: Is this necessary? It doesn't really matter if an exception is thrown or not; the test is just just checking that the right publishers are being used based on configuration. It just happens to throw here because the mock isn't fully set up

zoewangg: oh okay, I misread the test. We don't need it in this case

PR #1931, opened 2020-06-29
Updated metric publishing to address internal review comments.

millems:
Also fixed an issue where multiple parts of the SDK were ignoring the "alternate" request ID.


dagnir:
Ran the integ tests and they are passing as well.

PR #1933, opened 2020-06-30
Added decision log for metrics.

millems:


docs/design/core/metrics/DecisionLog.md
```diff
@@ -0,0 +1,41 @@
...
+
+**Closed Decisions:**
+
+1. Question? Decision. Justification.
+
+**Open Decisions:**
+
+1. (Old/Reopened/New) Question?
+
+## 7/30/20
```
Quanzzzz, line 20: Should this be 6/30/20? Or it's not a date?

millems: +1 good catch! I guess I subconsciously really want 2020 to be over faster.

PR #1957, opened 2020-07-17
Implement generic Waiter functionality

zoewangg:
## Description
Implement generic Waiter functionality

### DIFF
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/waiters/DefaultWaiter.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/waiters/DefaultWaiter.java
@@ -1,0 +1,85 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.core.internal.waiters;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.function.Supplier;
+import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.annotations.ThreadSafe;
+import software.amazon.awssdk.core.waiters.PollingStrategy;
+import software.amazon.awssdk.core.waiters.Waiter;
+import software.amazon.awssdk.core.waiters.WaiterAcceptor;
+import software.amazon.awssdk.core.waiters.WaiterResponse;
+import software.amazon.awssdk.utils.Validate;
+
+/**
+ * Default implementation of the generic {@link Waiter}.
+ * @param <T> the type of the response expected to return from the polling function
+ */
+@SdkInternalApi
+@ThreadSafe
+public final class DefaultWaiter<T> implements Waiter<T> {
+    private final PollingStrategy pollingStrategy;
+    private final List<WaiterAcceptor<T>> waiterAcceptors;
+
+    private DefaultWaiter(DefaultBuilder<T> builder) {
+        this.pollingStrategy = Validate.paramNotNull(builder.pollingStrategy, "pollingStrategy");
+        this.waiterAcceptors = Collections.unmodifiableList(builder.waiterAcceptors);
+    }
+
+    @Override
+    public WaiterResponse<T> run(Supplier<T> pollingFunction) {
+        WaiterExecutor<T> handler = new WaiterExecutor<>(pollingStrategy, waiterAcceptors);
+        return handler.execute(pollingFunction);
+    }
+
+    public static <T> Builder<T> builder() {
+        return new DefaultBuilder<>();
+    }
+
+    public static final class DefaultBuilder<T> implements Builder<T> {
+        private List<WaiterAcceptor<T>> waiterAcceptors = new ArrayList<>();
+        private PollingStrategy pollingStrategy;
+
+        private DefaultBuilder() {
+        }
+
+        @Override
+        public Builder<T> acceptors(List<WaiterAcceptor<T>> waiterAcceptors) {
+            this.waiterAcceptors = new ArrayList<>(waiterAcceptors);
+            return this;
+        }
+
+        @Override
+        public Builder<T> pollingStrategy(PollingStrategy pollingStrategy) {
+            this.pollingStrategy = pollingStrategy;
+            return this;
+        }
+
+        @Override
+        public Builder<T> addAcceptor(WaiterAcceptor<T> waiterAcceptor) {
+            waiterAcceptors.add(waiterAcceptor);
+            return this;
+        }
+
+        @Override
+        public Waiter<T> build() {
+            return new DefaultWaiter<>(this);
+        }
+    }
+}
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/waiters/PollingStrategy.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/waiters/PollingStrategy.java
@@ -1,0 +1,116 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.core.waiters;
+
+import java.time.Duration;
+import software.amazon.awssdk.annotations.SdkPublicApi;
+import software.amazon.awssdk.core.retry.backoff.BackoffStrategy;
+import software.amazon.awssdk.utils.Validate;
+
+/**
+ * Define the polling strategy for a {@link Waiter} to poll the resource
+ */
+@SdkPublicApi
+public final class PollingStrategy {
+
+    private final int maxAttempts;
+    private final BackoffStrategy backoffStrategy;
+    private final Duration maxWaitTime;
+
+    public PollingStrategy(Builder builder) {
+        this.maxAttempts = Validate.paramNotNull(builder.maxAttempts, "maxAttempts");
+        this.backoffStrategy = Validate.paramNotNull(builder.backoffStrategy, "backoffStrategy");
+        this.maxWaitTime = Validate.isPositiveOrNull(builder.maxWaitTime, "maxWaitTime");
+    }
+
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    /**
+     * Define the maximum number of attempts to try before transitioning the waiter to a failure state.
+     * @return a reference to this object so that method calls can be chained together.
+     */
+    public int maxAttempts() {
+        return maxAttempts;
+    }
+
+    /**
+     * Define the {@link BackoffStrategy} that computes the delay before the next retry request.
+     *
+     * @return a reference to this object so that method calls can be chained together.
+     */
+    public BackoffStrategy backoffStrategy() {
+        return backoffStrategy;
+    }
+
+    /**
+     * Define the amount of time to wait for the resource to transition to the desired state before
+     * giving up. This wait time doesn't have strict guarantees on how quickly a request is aborted
+     * when the timeout is breached. The request can timeout early if it is determined that the next
+     * retry will breach the max wait time.
+     *
+     * @return a reference to this object so that method calls can be chained together.
+     */
+    public Duration maxWaitTime() {
+        return maxWaitTime;
+    }
+
+    public static final class Builder {
+        private BackoffStrategy backoffStrategy;
+        private Integer maxAttempts;
+        private Duration maxWaitTime;
+
+        private Builder() {
+        }
+
+        /**
+         * Define the {@link BackoffStrategy} that computes the delay before the next retry request.
+         *
+         * @param backoffStrategy The new backoffStrategy value.
+         * @return This object for method chaining.
+         */
+        public Builder backoffStrategy(BackoffStrategy backoffStrategy) {
+            this.backoffStrategy = backoffStrategy;
+            return this;
+        }
+
+        /**
+         * Define the maximum number of attempts to try before transitioning the waiter to a failure state.
+         *
+         * @param maxAttempts The new maxAttempts value.
+         * @return This object for method chaining.
+         */
+        public Builder maxAttempts(Integer maxAttempts) {
+            this.maxAttempts = maxAttempts;
+            return this;
+        }
+
+        /**
+         * Define the maximum time to try before transitioning the waiter to a failure state.
+         * @param maxWaitTime The new maxWaitTime value.
+         * @return This object for method chaining.
+         */
+        public Builder maxWaitTime(Duration maxWaitTime) {
+            this.maxWaitTime = maxWaitTime;
+            return this;
+        }
+
+        public PollingStrategy build() {
+            return new PollingStrategy(this);
+        }
+    }
+}
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/waiters/Waiter.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/waiters/Waiter.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.core.waiters;
+
+import java.util.function.Supplier;
+import software.amazon.awssdk.annotations.SdkPublicApi;
+import software.amazon.awssdk.core.internal.waiters.DefaultWaiter;
+
+/**
+ * Waiter utility class that waits for a resource to transition to the desired state.
+ *
+ * @param <T> the type of the resource returned from the polling function
+ */
+@SdkPublicApi
+public interface Waiter<T> {
+
+    /**
+     * It returns when the resource enters into a desired state or
+     * it is determined that the resource will never enter into the desired state.
+     *
+     * @param pollingFunction Represents the input of a <code>DescribeTable</code> operation.
+     * @return the response
+     */
+    WaiterResponse<T> run(Supplier<T> pollingFunction);
+
+    /**
+     * Creates a newly initialized builder for the waiter object.
+     *
+     * @param responseClass the response class
+     * @param <T> the type of the response
+     * @return a Waiter builder
+     */
+    static <T> Builder<T> builder(Class<? extends T> responseClass) {
+        return DefaultWaiter.builder();
+    }
+
+    /**
+     * The Waiter Builder
+     * @param <T> the type of the resource
+     */
+    interface Builder<T> extends WaiterBuilder<T, Builder<T>> {
+
+        /**
+         * An immutable object that is created from the properties that have been set on the builder.
+         * @return a reference to this object so that method calls can be chained together.
+         */
+        Waiter<T> build();
+    }
+}
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/waiters/WaiterAcceptor.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/waiters/WaiterAcceptor.java
@@ -1,0 +1,223 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.core.waiters;
+
+import java.util.Optional;
+import java.util.function.Predicate;
+import software.amazon.awssdk.annotations.SdkPublicApi;
+import software.amazon.awssdk.utils.Validate;
+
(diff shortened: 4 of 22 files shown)


core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/waiters/AsyncWaiterHandler.java
```diff
@@ -0,0 +1,118 @@
...
+import software.amazon.awssdk.core.waiters.PollingStrategy;
+import software.amazon.awssdk.core.waiters.WaiterAcceptor;
+import software.amazon.awssdk.core.waiters.WaiterResponse;
+import software.amazon.awssdk.core.waiters.WaiterState;
+import software.amazon.awssdk.utils.Logger;
+import software.amazon.awssdk.utils.Validate;
+
+/**
+ * Handler for async waiter operations
+ * @param <T> the type of the response
```
millems, line 36: Can you add more here? Not sure what a 'handler' does other than 'handle', and I don't know what it means to 'handle' an async waiter operation.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/waiters/AsyncWaiterHandler.java
```diff
@@ -0,0 +1,118 @@
...
+        Validate.paramNotNull(waiterAcceptors, "waiterAcceptors");
+        this.executorService = Validate.paramNotNull(executorService, "executorService");
+        this.handlerHelper = new WaiterHandlerHelper<>(waiterAcceptors, pollingStrategy);
+    }
+
+    /**
+     * Execute the provided async polling function
+     */
+    CompletableFuture<WaiterResponse<T>> execute(Supplier<CompletableFuture<T>> asyncPollingFunction) {
+        log.info(() -> "starting to execute");
```
millems, line 60: Leftover logging?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/waiters/AsyncWaiterHandler.java
```diff
@@ -0,0 +1,118 @@
...
+                switch (state) {
+                    case SUCCESS:
+                        future.complete(handlerHelper.createWaiterResponse(response, exception, attemptNumber));
+                        break;
+                    case RETRY:
+                        maybeRetry(asyncPollingFunction, future);
+                        break;
+                    case FAILURE:
+                        future.completeExceptionally(SdkClientException.create("A waiter acceptor was matched and transitioned "
+                                                                               + "the waiter to failure state"));
```
millems, line 89: Any way that we could let an acceptor can provide a failure reason to include in the exception?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/waiters/AsyncWaiterHandler.java
```diff
@@ -0,0 +1,118 @@
...
+                        break;
+                    default:
+                        future.completeExceptionally(new UnsupportedOperationException());
+                }
+            } else {
+                if (exception != null) {
+                    future.completeExceptionally(SdkClientException.create("An exception was thrown and did not match any "
+                                                                           + "waiter acceptors"));
```
millems, line 97: Should we preserve "exception" as the cause?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/waiters/AsyncWaiterHandler.java
```diff
@@ -0,0 +1,118 @@
...
+        });
+    }
+
+    private void maybeRetry(Supplier<CompletableFuture<T>> asyncPollingFunction,
+                            CompletableFuture<WaiterResponse<T>> future) {
+        if (attemptNumber >= pollingStrategy.maxAttempts()) {
+            future.completeExceptionally(SdkClientException.create("The waiter has exceeded max retry attempts: " +
+                                                                   pollingStrategy.maxAttempts()));
+            return;
+        }
```
millems, line 112: Could we make this class thread-safe and reusable by just passing through the attemptNumber on each `doExecute` call? This seems like unnecessary mutability.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/waiters/AsyncWaiterHandler.java
```diff
@@ -0,0 +1,118 @@
...
+    private void doExecute(Supplier<CompletableFuture<T>> asyncPollingFunction, CompletableFuture<WaiterResponse<T>> future) {
+        ++attemptNumber;
+        log.info(() -> "runOnce" + attemptNumber);
+        runAsyncPollingFunction(asyncPollingFunction, future);
+    }
+
+    private void runAsyncPollingFunction(Supplier<CompletableFuture<T>> asyncPollingFunction,
+                                         CompletableFuture<WaiterResponse<T>> future) {
+        asyncPollingFunction.get().whenComplete((response, exception) -> {
+            Optional<WaiterState> waiterState = handlerHelper.nextWaiterStateIfMatched(response, exception);
```
millems, line 75: Any reason we don't have the "max attempts" be an acceptor, instead of special-casing it within this handler?  I realize that means we have to pass something through to allow the acceptor to track "execution context" for a waiter invocation (so that it can keep count), but that seems like a useful feature that better separates the responsibilities of this 'handler'.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/waiters/DefaultWaiter.java
```diff
@@ -0,0 +1,101 @@
...
+import software.amazon.awssdk.core.waiters.WaiterResponse;
+import software.amazon.awssdk.utils.Validate;
+
+/**
+ * Default implementation of the generic {@link Waiter}.
+ * @param <T> the type of the response expected to return from the polling function
+ */
+@SdkInternalApi
+@ThreadSafe
+public final class DefaultWaiter<T> implements Waiter<T> {
```
millems, line 38: Should we differentiate between a `Waiter` and an `AsyncWaiter`? It seems weird that someone that only wants the synchronous run(...) method can (should?) configure an executor service.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/waiters/DefaultWaiter.java
```diff
@@ -0,0 +1,101 @@
...
+    private final PollingStrategy pollingStrategy;
+    private final ScheduledExecutorService executorService;
+    private final List<WaiterAcceptor<T>> waiterAcceptors;
+
+    private DefaultWaiter(DefaultBuilder<T> builder) {
+        this.executorService = builder.scheduledExecutorService;
```
millems, line 44: What handles the "scheduleExecutorService is null" case? When runAsync is invoked?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/waiters/DefaultWaiter.java
```diff
@@ -0,0 +1,101 @@
...
+
+        @Override
+        public Builder<T> scheduledExecutorService(ScheduledExecutorService scheduledExecutorService) {
+            this.scheduledExecutorService = scheduledExecutorService;
+            return this;
+        }
+
+        @Override
+        public Builder<T> acceptors(List<WaiterAcceptor<T>> waiterAcceptors) {
+            this.waiterAcceptors = waiterAcceptors;
```
millems, line 81: Copy?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/waiters/WaiterHandlerHelper.java
```diff
@@ -0,0 +1,96 @@
...
+import software.amazon.awssdk.core.waiters.PollingStrategy;
+import software.amazon.awssdk.core.waiters.WaiterAcceptor;
+import software.amazon.awssdk.core.waiters.WaiterResponse;
+import software.amazon.awssdk.core.waiters.WaiterState;
+
+/**
+ * The waiter handler helper class
+ */
+@SdkInternalApi
+public final class WaiterHandlerHelper<T> {
```
millems, line 31: "handler helper". Helps the handler? Seems very generic. Is this just for code sharing between the async and sync code paths?

If so, is there any way we could use a strategy pattern here, where this is a parent class that handles the core logic, but it delegates to the children to do it synchronously or asynchronously?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/waiters/WaiterHandlerHelper.java
```diff
@@ -0,0 +1,96 @@
...
+     * @return the optional {@link WaiterState}
+     */
+    public Optional<WaiterState> nextWaiterStateIfMatched(T response, Throwable exception) {
+        Optional<WaiterState> waiterState;
+        if (exception != null) {
+            waiterState = exceptionMatches(exception);
+        } else {
+            waiterState = responseMatches(response);
+        }
+        return waiterState;
```
millems, line 72: ternary might read simpler.

core/sdk-core/src/main/java/software/amazon/awssdk/core/waiters/PollingStrategy.java
```diff
@@ -0,0 +1,90 @@
...
+import software.amazon.awssdk.utils.Validate;
+
+/**
+ * Define the polling strategy for a {@link Waiter} to poll the resource
+ */
+@SdkPublicApi
+public final class PollingStrategy {
+
+    private final int maxAttempts;
+    private final BackoffStrategy backoffStrategy;
```
millems, line 29: I feel like a lot of customers would want "maxWaitTime" as a configuration more than maxAttempts. I realize the max wait time can be indirectly managed via the maxAttempts, backoffStrategy and client configuration, but it's not easy to do. It would be nice if we did it for them, and we can use the request-level max-execution-time and some checks before sleeping/delegating to the scheduled executor service to do so fairly easily.

core/sdk-core/src/main/java/software/amazon/awssdk/core/waiters/Waiter.java
```diff
@@ -0,0 +1,134 @@
...
+         */
+        default Builder<T> pollingStrategy(Consumer<PollingStrategy.Builder> pollingStrategy) {
+            PollingStrategy.Builder builder = PollingStrategy.builder();
+            pollingStrategy.accept(builder);
+            return pollingStrategy(builder.build());
+        }
+
+        /**
+         * Defines the {@link ScheduledExecutorService} used to schedule async polling attempts
+         * Only required if you are calling {@link Waiter#runAsync(Supplier)}
```
millems, line 121: Again, "only required during creation, if you invoke this method during usage" feels like a gross invariant. Wouldn't it be better to encode that into the type? (`AsyncWaiter` versus `Waiter`)

core/sdk-core/src/main/java/software/amazon/awssdk/core/waiters/WaiterAcceptor.java
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    /**
+     * Creates a retry on exception waiter acceptor which determines if the exception should transition the waiter to retry state
+     *
+     * @param errorPredicate the {@link Throwable} predicate
+     * @param <T> the response type
+     * @return a {@link WaiterAcceptor}
+     */
+    static <T> WaiterAcceptor<T> retryOnExceptionAcceptor(Predicate<Throwable> errorPredicate) {
```
millems, line 144: What about `retryOnResponseAcceptor`?

core/sdk-core/src/main/java/software/amazon/awssdk/core/waiters/WaiterResponse.java
```diff
@@ -0,0 +1,43 @@
...
+
+    /**
+     * @return the optional response received that has matched with the waiter success condition
+     */
+    Optional<T> response();
+
+    /**
+     * @return the optional exception thrown from the waiter operation that has matched with the waiter success condition
+     */
+    Optional<Throwable> exception();
```
millems, line 36: Should we use an `Either` type here? Or something that is more explicitly a union type?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/waiters/AsyncWaiterHandler.java
zoewangg: To me, a `WatierAcceptor` is really an inspector that checks the response or exception returned from an operation and it shouldn't really care about the execution context. IMO, It's the handler's responsibility to handle it, rather than the WaiterAcceptor.

It'd probably be something like the following. Not sure if I like the fact that `WaiterExecutionContext`  is added to be only used by this acceptor and it feels the acceptor is more like a WaiterIntercetor now. 

```
    static <T> WaiterAcceptor<T> retryAttempsAcceptor(int maxAttempts) {
        return new WaiterAcceptor<T>() {
            @Override
            public WaiterState waiterState() {
                return WaiterState.FAILURE;
            }

            @Override
            public boolean matches(T response, WaiterExecutionContext context) {
                if (context.attempts() > maxAttemtps) {
                    return true;
                }
                return false;
            }

            @Override
            public boolean matches(Throwable throwable, WaiterExecutionContext context) {
                if (context.attempts() > maxAttemtps) {
                    return true;
                }
                return false;
            }
        }
    }
```

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/waiters/AsyncWaiterHandler.java
zoewangg: Yeah, let me try it

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/waiters/AsyncWaiterHandler.java
millems: I think I might be trying too hard. You're probably right that we should keep it simple, since I don't see us needing really fancy acceptors. 

My original comment was thinking a little more generic, like:

```java
    static <T> WaiterAcceptor<T> retryAttempsAcceptor(int maxAttempts) {
        return new WaiterAcceptor<T>() {
            private static AttributeKey<Integer> ATTEMPTS = AttributeKey.create("Attempts");

            @Override
            public WaiterState waiterState() {
                return WaiterState.FAILURE;
            }

            @Override
            public boolean matches(T response, WaiterExecutionContext context) {
                incrementAttempt(context);
                if (getAttempts(context) > maxAttempts) {
                    return true;
                }
                return false;
            }

            @Override
            public boolean matches(Throwable throwable, WaiterExecutionContext context) {
                incrementAttempt(context);
                if (getAttempts(context) > maxAttempts) {
                    return true;
                }
                return false;
            }
            
            private void incrementAttempt(WaiterExecutionContext context) {
                context.setAttribute(ATTEMPTS, getAttempts(context) + 1);
            }

            private int getAttempts(WaiterExecutionContext context) {
                Integer attempts = context.getAttribute(ATTEMPTS);
                return attempts == null ? 1 : attempts;
            }
        }
    }
```

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/waiters/DefaultWaiter.java
millems: I prefer for preconditions to be encoded in the type/constructor. "If you give me a waiter that was initialized, I'll work just fine". 

If preconditions are only validated when the Waiter is used, the user of the Waiter needs to handle the case that it's not properly configured themselves. E.g. If I have a `WaiterUser` that is injected with a `Waiter` when it's created, I either need to call `runAsync` when a `WaiterUser` is created to make sure it doesn't throw an exception, or annotate every method that uses `runAsync` to document that it will throw an exception if the configured `Waiter` wasn't configured correctly. I also probably should (but probably won't remember to) write unit tests for that case, which doubles the number of tests I should write for `WaiterUser`.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/waiters/DefaultWaiter.java
millems: That doesn't prevent the case of mutating the list when it's in the builder.

```java
List<WaiterAcceptor<T>> acceptors = new ArrayList<>();
waiterBuilder.acceptors(acceptors);
acceptors.put(someAcceptor); // Mutates the acceptors within waiterBuilder.
```

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/waiters/WaiterHandlerHelper.java
millems: I'm not convinced. That's what the strategy pattern is for. I usually pick composition over inheritance as well, but there are exceptions to the rule. That said, if you've tested it out and it's messy, I believe you. I haven't done the deep-dive necessary to tell whether it's any better than this option.

I notice a lot of duplicate code between the async and sync code paths, though. Can we move some of it to the WaiterHandlerHelper?

core/sdk-core/src/main/java/software/amazon/awssdk/core/waiters/PollingStrategy.java
millems: Is there any way to let the customer configure the `maxWaitTime` and `maxAttempts` on the generated service waiters, to override the behavior?

I don't mind waiting on customers to ask, but I have recently seen a *lot* of people using the `maxClientExecuteDuration` duration lately to limit the amount of time that the SDK will be holding their threads. I feel like SAs are recommending it pretty strongly to their customers, and I am worried that customers might think of the feature as incomplete without it.

core/sdk-core/src/main/java/software/amazon/awssdk/core/waiters/WaiterAcceptor.java
millems: That's fair, but it feels like a weird omission, even if people wouldn't use it very much. A `retryOnResponseAcceptor` would still be useful to allow someone to short-circuit a more broad acceptor later in the list:

E.g.
```java
waiterBuilder.addAcceptor(retryOnResponseAcceptor(r -> r.errorCode() == FieldValues.RETRYABLE_ERROR)
             .addAcceptor(failOnResponseAcceptor(r -> r.errorCode() != null))
```

I agree that it's not necessary if all we're using this for is the generated waiters, but we're writing it more generic for a reason.

I'd almost rather have the default behavior of the core Waiters to *not* retry by default and just have all of the generated waiters use `addAcceptor(retryOnResponseAcceptor(r -> true))` since that seems more of an intuitive default.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/waiters/WaiterHandlerHelper.java
zoewangg: Are you referring to the switch statement? It's not true duplication. For instance, for exceptional case, sync will throw exception and async will complete the future exceptionally. While it's possible to share more code, the code would be less readable, in my mind unless you have other ideas?

sync:
```java
        if (waiterState.isPresent()) {
            WaiterState state = waiterState.get();
            switch (state) {
                case SUCCESS:
                    return handlerHelper.createWaiterResponse(response, exception, attemptNumber);
                case RETRY:
                    return maybeRetry(pollingFunction);
                case FAILURE:
                    throw SdkClientException.create("A waiter acceptor was matched and transitioned the waiter "
                                                    + "to failure state.");
                default:
            }
        }

        if (exception != null) {
            throw SdkClientException.create("An exception was thrown and did not match with any waiter acceptor", exception);
        } else {
            // default to retry if there's no acceptor matched for the response
            return maybeRetry(pollingFunction);
        }
    }
```

async:

```java
            if (waiterState.isPresent()) {
                log.info(() -> "waiter state " + waiterState.get());
                WaiterState state = waiterState.get();
                switch (state) {
                    case SUCCESS:
                        future.complete(handlerHelper.createWaiterResponse(response, exception, attemptNumber));
                        break;
                    case RETRY:
                        maybeRetry(asyncPollingFunction, future);
                        break;
                    case FAILURE:
                        future.completeExceptionally(SdkClientException.create("A waiter acceptor was matched and transitioned "
                                                                               + "the waiter to failure state"));
                        break;
                    default:
                        future.completeExceptionally(new UnsupportedOperationException());
                }
            } else {
                if (exception != null) {
                    future.completeExceptionally(SdkClientException.create("An exception was thrown and did not match any "
                                                                           + "waiter acceptors"));
                } else {
                    // default to retry if there's no acceptor matched for the response
                    maybeRetry(asyncPollingFunction, future);
                }
            }
```

core/sdk-core/src/main/java/software/amazon/awssdk/core/waiters/WaiterAcceptor.java
zoewangg: Yeah, makes sense. I'll add it.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/waiters/WaiterHandlerHelper.java
millems: I think the sync ("return the result") and async ("return a completed future") behavior can be factored out of the logic itself. It's just a question of how hard it makes the code to read.

zoewangg:
> Should this be in util, not sdk-core, so that we can use it more easily in tests?

One disadvantage of using it in util is that it doesn't have access to the `SdkClientException` and some other core classes.

millems:
> > Should this be in util, not sdk-core, so that we can use it more easily in tests?
> 
> One disadvantage of using it in util is that it doesn't have access to the `SdkClientException` and some other core classes.

What other core classes? I feel like losing `SdkClientException` is very minor... It might even be nice to put it in a different package altogether.

core/sdk-core/src/main/java/software/amazon/awssdk/core/waiters/WaiterAcceptor.java
```diff
@@ -0,0 +1,214 @@
...
+            public boolean matches(T response) {
+                return responsePredicate.test(response);
+            }
+
+            @Override
+            public Optional<String> message() {
+                return Optional.of(message);
+            }
+        };
+    }
```
millems, line 171: Validate the message is not-null so that it doesn't fail slow? Or use Optional.ofNullable.

core/sdk-core/src/main/java/software/amazon/awssdk/core/waiters/WaiterBuilder.java
```diff
@@ -0,0 +1,72 @@
...
+     */
+    B addAcceptor(WaiterAcceptor<T> waiterAcceptors);
+
+    /**
+     * Defines a {@link PollingStrategy} to use when polling a resource
+     *
+     * @param pollingStrategy the polling strategy to use
+     * @return a reference to this object so that method calls can be chained together.
+     */
+    B pollingStrategy(PollingStrategy pollingStrategy);
```
millems, line 58: Do we have a naming convention established that differentiates between a Strategy (e.g. PollingStrategy) or Policy (e.g. RetryPolicy)?

core/sdk-core/src/main/java/software/amazon/awssdk/core/waiters/WaiterAcceptor.java
zoewangg: Ah, good catch!

core/sdk-core/src/main/java/software/amazon/awssdk/core/waiters/WaiterBuilder.java
zoewangg: Nope, don't think so.`PollingStrategy`, `BackoffStrategy`, and `RetryPolicy` are all existing classes.

zoewangg:
> What other core classes? I feel like losing SdkClientException is very minor... It might even be nice to put it in a different package altogether.

Discussed offline. To close the loop, waiter classes also use `PollingStrategy` and `BackoffStrategy` from the core module, so it's not easy to move them to utils.

zoewangg:
Bugs and code smells reported by sonarcloud are *not* related to this change at all. Code coverage is also incorrect!

PR #1977, opened 2020-08-07
Added an "unsafe" way to retrieve a byte array from `SdkBytes` and `ResponseBytes` without copying the data.

### LINKED ISSUE #1959: Allow to get array from BytesWrapper without copying
In performance sensitive applications in is important to reduce number of data-copies that happen during processing.
Unfortunately currently when we get data from S3 and then try to use response byte array we have to:
* Either do an array copy
* Get a readonly ByteBuffer - which doesn't provide access to array and forces us to do an array copy later

## Describe the Feature
Provide a method in `BytesWrapper` that allows users to get either underlying array directly or a non-read-only `ByteBuffer`

## Is your Feature Request related to a problem?
Current S3 client implementation does multiple data copies - this is just one more of them, and this reduces performace.

## Describe alternatives you've considered
Unfortunately there are not many alternatives: in some cases one needs access to arrays, for example if one tries top use JNI compression implementations.

## Additional Context

- [x] I may be able to implement this feature request

## Your Environment

* AWS Java SDK version used:
* JDK version used:

millems:
Reduced the number of payload data copies on some code paths.

Fixes #1959.

### DIFF
--- a/core/sdk-core/src/test/java/software/amazon/awssdk/core/ResponseBytesTest.java
+++ b/core/sdk-core/src/test/java/software/amazon/awssdk/core/ResponseBytesTest.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.core;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.junit.Test;
+
+public class ResponseBytesTest {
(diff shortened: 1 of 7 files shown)


millems:
Integration tests have passed. Running benchmarks...

millems:
Unit tests have passed. Still waiting on benchmarks...

millems:
No appreciable difference in benchmarks.

core/sdk-core/src/test/java/software/amazon/awssdk/core/ResponseBytesTest.java
```diff
@@ -0,0 +1,34 @@
+package software.amazon.awssdk.core;
```
zoewangg, line 1: missing copyright header?

.changes/next-release/feature-AWSSDKforJavav2-47fe81e.json
```diff
@@ -0,0 +1,5 @@
+{
+    "type": "feature",
+    "category": "AWS SDK for Java v2",
+    "description": "Slight performance improvement, from reduced the number of data copies on some code paths."
```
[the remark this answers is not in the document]
millems: Done, as well as removed the actual changes that use the new methods.

core/sdk-core/src/test/java/software/amazon/awssdk/core/ResponseBytesTest.java
millems: Fixed, thanks!

PR #1979, opened 2020-08-07
Remove DynamoDB and Metrics tests from the benchmark test suite, by default. This should make running benchmarks much more timely.

millems:


test/sdk-benchmarks/src/main/java/software/amazon/awssdk/benchmark/BenchmarkRunner.java
```diff
@@ -96,8 +96,11 @@ public static void main(String... args) throws RunnerException, JsonProcessingEx
         benchmarksToRun.addAll(ASYNC_BENCHMARKS);
         benchmarksToRun.addAll(PROTOCOL_BENCHMARKS);
         benchmarksToRun.addAll(COLD_START_BENCHMARKS);
-        benchmarksToRun.addAll(MAPPER_BENCHMARKS);
-        benchmarksToRun.addAll(METRIC_BENCHMARKS);
+
+        System.out.println("Skipping tests, to reduce benchmark times: \n" +
```
zoewangg, line 100: Do we need the log here? The benchmarksToRun will be printed out on line 112. If so, can we use the logger created on line 83

millems: The log is good to make it clear what *isn't* being run. It also makes it so that we don't have static private variables laying around. +1 to moving to the logger.

PR #1987, opened 2020-08-13
Enable fips region to work with Access point ARN.

Quanzzzz:
Enable fips region to work with Access point ARN;
Changed the check of matches between client configured region and arn region;
Add new method related to ```fipsEnabled``` in ```S3AccessPointBuilder``` class;

## Testing

Add tests for multiple cases.

## License


services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/S3EndpointUtils.java
```diff
@@ -131,7 +125,7 @@ private static ConfiguredS3SdkHttpRequest applyEndpointConfigurationForAccessPoi
         }
 
         if (serviceConfiguration == null || !serviceConfiguration.useArnRegionEnabled()) {
-            if (!region.id().equals(arnRegion)) {
```
zoewangg, line 134: can we define two variables here so that they can be reused below?

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/S3EndpointUtils.java
```diff
@@ -167,10 +161,11 @@ private static ConfiguredS3SdkHttpRequest applyEndpointConfigurationForAccessPoi
                                 .accountId(
                                     s3EndpointResource.accountId().orElseThrow(() -> new IllegalArgumentException(
                                         "An S3 access point ARN must have an account ID")))
-                                .region(arnRegion)
+                                .region(removeFipsIfNeeded(arnRegion))
                                 .protocol(request.protocol())
                                 .domain(clientPartitionMetadata.dnsSuffix())
                                 .dualstackEnabled(dualstackEnabled)
+                                .fipsEnabled(isFipsRegionProvided(region.toString(), arnRegion, serviceConfiguration.useArnRegionEnabled()))
```
zoewangg, line 168: let's create a variable for this

services/s3control/src/main/java/software/amazon/awssdk/services/s3control/internal/interceptors/EndpointAddressInterceptor.java
```diff
@@ -62,17 +62,15 @@ public SdkHttpRequest modifyHttpRequest(Context.ModifyHttpRequest context,
...
         String host = request.getUri().getHost();
         if (isDualstackEnabled(configuration)) {
             if (!host.contains(ENDPOINT_PREFIX)) {
                 throw SdkClientException.create(String.format("The Dual-Stack option cannot be used with custom endpoints (%s)",
                                                               request.getUri()));
             }
             host = host.replace(ENDPOINT_PREFIX, String.format("%s.%s", ENDPOINT_PREFIX, "dualstack"));
-        } else if (isFipsEnabled(configuration)) {
+        }
+        if (isFipsEnabled(configuration)) {
```
zoewangg, line 73: Don't think we should make changes in s3 control

PR #1989, opened 2020-08-14
Various codegen performance improvements.

millems:
Only potentially impacting change: intermediate model is no longer written to disk, because it is not used.


millems:
FYI, I did a codegen before and after these changes to make sure that the generated code was byte-for-byte identical.

codegen/src/main/java/software/amazon/awssdk/codegen/CodeGenerator.java
```diff
@@ -66,12 +57,14 @@ public static Builder builder() {
      * code.
      */
     public void execute() {
+        // Make sure ClassName is statically initialized before we do anything in parallel.
+        // Parallel static initialization of ClassName and TypeName can result in a deadlock:
+        // https://github.com/square/javapoet/issues/799
+        ClassName.get(Object.class);
+
```
dagnir, line 64: Should this be moved to a `static` block within this class?

codegen/src/main/java/software/amazon/awssdk/codegen/emitters/CodeWriter.java
```diff
@@ -100,22 +99,20 @@ public CodeWriter(String dir, String file, String fileNameSuffix) {
      */
     @Override
     public void flush() {
-        PrintWriter out = null;
         try {
-            File outputFile = Utils.createFile(dir, this.file);
+            File outputFile = new File(dir, this.file);
```
dagnir, line 103: nit: use `Paths.get()` so we can remove call to `toPath()` later

test/protocol-tests/src/main/resources/codegen-resources/query/service-2.json
```diff
@@ -97,12 +97,6 @@
     "BlobType":{"type":"blob"},
     "Boolean":{"type":"boolean"},
     "Double":{"type":"double"},
-    "EmptyModeledException":{
```
dagnir, line 100: Why did this get removed?

millems: It was actually a duplicated entry: there's another EmptyModeledException below. Jackson-databind ignored it, but Jackson-jr required additional flags to support duplicate entries. Since it was invalid anyway, I just fixed it.

codegen/src/main/java/software/amazon/awssdk/codegen/CodeGenerator.java
millems: Sure

millems: Fixed.

codegen/src/main/java/software/amazon/awssdk/codegen/emitters/CodeWriter.java
millems: Fixed.

PR #1995, opened 2020-08-21
Make STSCredentialsProvider prefetch and stale times configurable

Helmsdown:
## Description
Allow clients to specify the time (relative to expiration) for when to refresh STS session tokens.

## Motivation and Context
The default STSCredentialsProvider's strategy for refreshing STS sessions is to wait until the session is almost stale and to refresh it either 5 minutes (prefetch) or 1 minute (stale) before it expires. This works great for most use cases. However, it doesn't work well for pre-signed S3 urls. For example, let's say I want to pre-sign an s3 URL that will be good for 90 minutes AND I'm signing the URL with an assumed STS role's credentials. The URL's effective expiration is the min(url expiration, sts session expiration). In other words, if you have STS session credentials that last for 60 minutes, and you want to hand out a URL that will last for 30 minutes, any URLs signed with said session credentials from minutes 31 to 60 will not actually be valid for the full 30 minutes.

The solution is to allow the client to specify the stale and prefetch times of the to use when determining when a session is ready for prefetch and when it's stale. I expose these values as Java Function objects because I intend to reference runtime changeable configuration properties to compute the stale and prefetch times (e.g. prefetch = actual - ${url_length} - ${prefetch_relative_to_url_expiration})

## Testing
I modified the StsCredentialsProviderTestBase. For each current test case I replicated it and specified stale and prefetch time functions that provide the same values they did before.

## Types of changes

## Checklist

- [Not quite] Local run of `mvn install` succeeds
I've seen several failures for classes that are not related to anything I've touched. Re-running the tests in intellij succeeds
For example: aws-sdk-java-v2/test/codegen-generated-classes-test/target/surefire-reports

## License

### DIFF
--- a/services/sts/src/main/java/software/amazon/awssdk/services/sts/auth/StsCredentialsProvider.java
+++ b/services/sts/src/main/java/software/amazon/awssdk/services/sts/auth/StsCredentialsProvider.java
@@ -17,6 +17,7 @@ package software.amazon.awssdk.services.sts.auth;
 
 import java.time.Duration;
 import java.time.Instant;
+import java.util.Optional;
 import java.util.function.Function;
 import software.amazon.awssdk.annotations.NotThreadSafe;
 import software.amazon.awssdk.annotations.SdkInternalApi;
@@ -31,6 +32,7 @@ import software.amazon.awssdk.utils.cache.CachedSupplier;
 import software.amazon.awssdk.utils.cache.NonBlocking;
 import software.amazon.awssdk.utils.cache.RefreshResult;
 
+
 /**
  * An implementation of {@link AwsCredentialsProvider} that is extended within this package to provide support for periodically-
  * updating session credentials. When credentials get close to expiration, this class will attempt to update them asynchronously
@@ -40,6 +42,10 @@ import software.amazon.awssdk.utils.cache.RefreshResult;
 @ThreadSafe
 @SdkInternalApi
 abstract class StsCredentialsProvider implements AwsCredentialsProvider, SdkAutoCloseable {
+
+    private static final Duration DEFAULT_STALE_TIME = Duration.ofMinutes(1);
+    private static final Duration DEFAULT_PREFETCH_TIME = Duration.ofMinutes(5);
+
     /**
      * The STS client that should be used for periodically updating the session credentials in the background.
      */
@@ -50,8 +56,14 @@ abstract class StsCredentialsProvider implements AwsCredentialsProvider, SdkAuto
      */
     private final CachedSupplier<SessionCredentialsHolder> sessionCache;
 
+    private final Duration staleTime;
+    private final Duration prefetchTime;
+
     protected StsCredentialsProvider(BaseBuilder<?, ?> builder, String asyncThreadName) {
         this.stsClient = Validate.notNull(builder.stsClient, "STS client must not be null.");
+
+        this.staleTime = Optional.ofNullable(builder.staleTime).orElse(DEFAULT_STALE_TIME);
+        this.prefetchTime = Optional.ofNullable(builder.prefetchTime).orElse(DEFAULT_PREFETCH_TIME);
 
         CachedSupplier.Builder<SessionCredentialsHolder> cacheBuilder = CachedSupplier.builder(this::updateSessionCredentials);
         if (builder.asyncCredentialUpdateEnabled) {
@@ -67,9 +79,10 @@ abstract class StsCredentialsProvider implements AwsCredentialsProvider, SdkAuto
     private RefreshResult<SessionCredentialsHolder> updateSessionCredentials() {
         SessionCredentialsHolder credentials = new SessionCredentialsHolder(getUpdatedCredentials(stsClient));
         Instant actualTokenExpiration = credentials.getSessionCredentialsExpiration().toInstant();
+
         return RefreshResult.builder(credentials)
-                            .staleTime(actualTokenExpiration.minus(Duration.ofMinutes(1)))
-                            .prefetchTime(actualTokenExpiration.minus(Duration.ofMinutes(5)))
+                            .staleTime(actualTokenExpiration.minus(staleTime))
+                            .prefetchTime(actualTokenExpiration.minus(prefetchTime))
                             .build();
     }
 
@@ -81,6 +94,21 @@ abstract class StsCredentialsProvider implements AwsCredentialsProvider, SdkAuto
     @Override
     public void close() {
         sessionCache.close();
+    }
+
+    /**
+     * The amount of time, relative to STS token expiration, that the cached credentials are considered stale and should no longer be used.
+     * All threads will block until the value is updated.
+     */
+    public Duration staleTime() {
+        return staleTime;
+    }
+
+    /**
+     * The amount of time, relative to STS token expiration, that the cached credentials are considered close to stale and should be updated.
+     */
+    public Duration prefetchTime() {
+        return prefetchTime;
     }
 
     /**
@@ -97,6 +125,8 @@ abstract class StsCredentialsProvider implements AwsCredentialsProvider, SdkAuto
 
         private Boolean asyncCredentialUpdateEnabled = false;
         private StsClient stsClient;
+        private Duration staleTime;
+        private Duration prefetchTime;
 
         protected BaseBuilder(Function<B, T> providerConstructor) {
             this.providerConstructor = providerConstructor;
@@ -126,6 +156,31 @@ abstract class StsCredentialsProvider implements AwsCredentialsProvider, SdkAuto
             this.asyncCredentialUpdateEnabled = asyncCredentialUpdateEnabled;
             return (B) this;
         }
+
+        /**
+         * Configure the amount of time, relative to STS token expiration, that the cached credentials are considered stale and should no longer be used.
+         * All threads will block until the value is updated.
+         *
+         * <p>By default, this is 1 minute.</p>
+         */
+        @SuppressWarnings("unchecked")
+        public B staleTime(Duration staleTime) {
+            this.staleTime = staleTime;
+            return (B) this;
+        }
+
+        /**
+         * Configure the amount of time, relative to STS token expiration, that the cached credentials are considered close to stale and should be updated.
+         * See {@link #asyncCredentialUpdateEnabled}.
+         *
+         * <p>By default, this is 5 minutes.</p>
+         */
+        @SuppressWarnings("unchecked")
+        public B prefetchTime(Duration prefetchTime) {
+            this.prefetchTime = prefetchTime;
+            return (B) this;
+        }
+
 
         /**
          * Build the credentials provider using the configuration applied to this builder.
--- a/.changes/next-release/feature-AmazonSTS-289d9e7.json
+++ b/.changes/next-release/feature-AmazonSTS-289d9e7.json
@@ -1,0 +1,5 @@
+{
+    "category": "Amazon STS", 
+    "type": "feature", 
+    "description": "Make the STSCredentialsProvider stale and prefetch times configurable so clients can control when session credentials are refreshed"
+}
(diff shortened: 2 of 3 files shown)


bmaizels:
Hi @Helmsdown , thank you very much for putting this PR together. Taking a closer look at your use-case, we have concerns around using custom Functions to calculate the refresh thresholds. Primarily this is breaking the concept of immutability in the behavior of these objects which could easily lead to bad practices and unexpected race conditions, especially as these functions will have no context around why they are being invoked. We have no objections to exposing these thresholds as configurable values, but we would like to see them statically configured (with a duration of some kind) and thus immutable. What are your thoughts?

Helmsdown:
Using a java Function was a way for me to tie this to runtime changeable configuration properties. If this is undesirable to you for immutability reasons then I'll take static values (over nothing) defined once and know that I'll have to reboot my servers should I wish to have new values take effect. 

Would you like me to amend my PR or can you make the changes and get them merged?

bmaizels:
> Using a java Function was a way for me to tie this to runtime changeable configuration properties.

Understood, but we would prefer you to construct and switch to providers with different configuration on demand.

> Would you like me to amend my PR or can you make the changes and get them merged?

If you're able to amend your PR, then we can move forward with it in principle. I would still need to review the actual implementation, I wanted to having this more directional discussion first so that we're all on the same page around the goals of this change.

If you don't have time or inclination to make the change, we'd be happy for you to move this to an issue (as a feature request) and we'll work on it in priority order with other features.

Helmsdown:
I will amend my PR. How do you feel about specifying the static values for prefetch and stale time as java Duration objects?

bmaizels:
@Helmsdown 
> I will amend my PR. How do you feel about specifying the static values for prefetch and stale time as java Duration objects?

Yep I think that's what I would do. Thanks!

Helmsdown:
@bmaizels I made the changes as per our discussion and rebased to the tip of master.

Helmsdown:
Additionally, the travis CI build errors don't seem to be related to my changes. Can you confirm?

services/sts/src/main/java/software/amazon/awssdk/services/sts/auth/StsCredentialsProvider.java
```diff
@@ -127,6 +142,30 @@ public B asyncCredentialUpdateEnabled(Boolean asyncCredentialUpdateEnabled) {
...
+        public B staleTime(Duration staleTime) {
+            this.staleTime = staleTime;
+            return (B) this;
+        }
+
+        /**
+         * Configure the amount of time, relative to STS token expiration, that a new STS token should be fetched asynchronously.
+         * The value only applies if asyncCredentialUpdateEnabled is enabled.
+         *
+         * <p>By default, this is 5 minuts.</p>
```
bmaizels, line 160: Small typo here in 'minutes'.

bmaizels (commented):
Your testing strategy adds the new [override] codepath to existing tests, but I think we should also add tests that explicitly verify the behavior of the override. For instance we should have a test that proves that when I change the duration with an override that changed duration is respected and the behavior of the credentials provider changes from the default. I believe that if I removed the code from the credentials provider that actually stored and used the override, all these tests would still pass.

bmaizels:
> 
> 
> Additionally, the travis CI build errors don't seem to be related to my changes. Can you confirm?

@Helmsdown Thanks for continuing to work on this! I can confirm that this error has nothing to do with your changes. We're looking into it separately.

services/sts/src/main/java/software/amazon/awssdk/services/sts/auth/StsCredentialsProvider.java
Helmsdown: Type fixed and rebased to pick up changes to CHANGELOG.md

bmaizels:
@Helmsdown Thanks! Did you see my note about testing? I'll restate it in case you missed it in the review:

>Your testing strategy adds the new [override] codepath to existing tests, but I think we should also add tests that explicitly verify the behavior of the override. For instance we should have a test that proves that when I change the duration with an override that changed duration is respected and the behavior of the credentials provider changes from the default. I believe that if I removed the code from the credentials provider that actually stored and used the override, all these tests would still pass.

Helmsdown:
> Your testing strategy adds the new [override] codepath to existing tests, but I think we should also add tests that explicitly verify the behavior of the override. For instance we should have a test that proves that when I change the duration with an override that changed duration is respected and the behavior of the credentials provider changes from the default. I believe that if I removed the code from the credentials provider that actually stored and used the override, all these tests would still pass.

@bmaizels so there are a few things to do discuss here. Would it be a fair characterization to say that the testing of the StsCredentialsProvider implementations on master is somewhat _light_? 

1. No tests actually set the asyncCredentialUpdateEnabled flag when constructing the providers.
2. The test distantExpiringCredentialsUpdatedInBackground on master doesn't really test the async behavior. Partially because of #1 but partially because it doesn't actually get ahold of the credential provider and try and invoke it AFTER initializing it with callClientWithCredentialsProvider. In fact, the expiration time it gives, 90 seconds, means the first invocation of the provider will result in an invocation (because the provider was "empty") and the second call that happens immediately after will result in an immediate prefetch call because 90 seconds is within the default 5 minute prefetch window.

So if we were to improve upon this it would mean that the default distantExpiringCredentialsUpdatedInBackground would need to do several things:
1. Set asyncCredentialUpdateEnabled to true in at least one concrete test class
2. Get access to the provider and invoke it inside the time boxed while loop to "prove" that nothing is happening while the the credentials are in neither the prefetch or stale time windows. 
3. Modify the code such that we don't close down the provider in the try with resource block in the base test (at least in this use case)
4. Set the credential expiration to be greater than 5 minutes

So are you ok with a ~5 minute unit test testing the default behavior? A test using the overridden stale/prefetch times could be faster given we can make the prefetch time.

I'm ok do the following if you concur its right direction. If my read of the current test code is correct, and I am willing to be educated otherwise, it is disheartening to have to re-imagine the SDK test code in this case.

Helmsdown:
@bmaizels additionally, if you want to jump on a Google Meet to discuss this in a shorter feedback loop I am more than willing to set one up.

Helmsdown:
I'm already rethinking part of what I said. The test does not need to be 5 minutes.

bmaizels:
Your comments are fair, I definitely don't want to put paying down pre-existing technical debt on you (unless you think you can easily write those missing tests and want to take a stab at it anyway). I think at a minimum I would propose the following :
1. Add getters to the new properties.
2. Write tests to assert a) default values are correct, b) overridden values are correct

Anything beyond that we'll treat as a bonus. Happy to meet and discuss in person if you want to do that too!

Helmsdown:
Could you jump in here for a moment to chat:
https://meet.google.com/jux-fbug-kmb

Helmsdown:
@bmaizels updated per our discussion in the video call. Let me know if I missed anything.

services/sts/src/main/java/software/amazon/awssdk/services/sts/auth/StsCredentialsProvider.java
```diff
@@ -127,6 +150,30 @@ public B asyncCredentialUpdateEnabled(Boolean asyncCredentialUpdateEnabled) {
             return (B) this;
         }
 
+        /**
+         * Configure the amount of time, relative to STS token expiration, that a new STS token should be fetched synchronously.
```
bmaizels, line 154: Can we change this to: Configure the amount of time, relative to STS token expiration, that the cached credentials are considered stale and should not longer be used. All threads will block until the value is updated.

services/sts/src/main/java/software/amazon/awssdk/services/sts/auth/StsCredentialsProvider.java
```diff
@@ -127,6 +150,30 @@ public B asyncCredentialUpdateEnabled(Boolean asyncCredentialUpdateEnabled) {
...
+         */
+        @SuppressWarnings("unchecked")
+        public B staleTime(Duration staleTime) {
+            this.staleTime = staleTime;
+            return (B) this;
+        }
+
+        /**
+         * Configure the amount of time, relative to STS token expiration, that a new STS token should be fetched asynchronously.
+         * The value only applies if asyncCredentialUpdateEnabled is enabled.
```
bmaizels, line 166: Can we change this to: Configure the amount of time, relative to STS token expiration, that the cached credentials are considered close to stale and should be updated. See {@link #asyncCredentialUpdateEnabled}.

bmaizels (commented):
Looking good! Just a couple of doc things to take care of. Sorry I didn't flag these the first time.

services/sts/src/main/java/software/amazon/awssdk/services/sts/auth/StsCredentialsProvider.java
Helmsdown: Done

services/sts/src/main/java/software/amazon/awssdk/services/sts/auth/StsCredentialsProvider.java
Helmsdown: Done

services/sts/src/main/java/software/amazon/awssdk/services/sts/auth/StsCredentialsProvider.java
```diff
@@ -83,6 +96,14 @@ public void close() {
         sessionCache.close();
     }
 
+    public Duration staleTime() {
```
bmaizels, line 99: Can we add javadoc: The amount of time, relative to STS token expiration, that the cached credentials are considered stale and should no longer be used. All threads will block until the value is updated.

services/sts/src/main/java/software/amazon/awssdk/services/sts/auth/StsCredentialsProvider.java
```diff
@@ -83,6 +96,14 @@ public void close() {
...
+        return staleTime;
+    }
+
+    public Duration prefetchTime() {
```
bmaizels, line 103: Can we add javadoc: The amount of time, relative to STS token expiration, that the cached credentials are considered close to stale and should be updated.

services/sts/src/main/java/software/amazon/awssdk/services/sts/auth/StsCredentialsProvider.java
Helmsdown: Done

services/sts/src/main/java/software/amazon/awssdk/services/sts/auth/StsCredentialsProvider.java
Helmsdown: Done

bmaizels (approved):

services/sts/src/main/java/software/amazon/awssdk/services/sts/auth/StsCredentialsProvider.java
```diff
@@ -40,6 +42,10 @@
 @ThreadSafe
 @SdkInternalApi
 abstract class StsCredentialsProvider implements AwsCredentialsProvider, SdkAutoCloseable {
+
+    private static Duration DEFAULT_STALE_TIME = Duration.ofMinutes(1);
```
bmaizels, line 46: Can we change this to : ```private static final Duration DEFAULT_STALE_TIME = Duration.ofMinutes(1);```

services/sts/src/main/java/software/amazon/awssdk/services/sts/auth/StsCredentialsProvider.java
```diff
@@ -40,6 +42,10 @@
...
+    private static Duration DEFAULT_PREFETCH_TIME = Duration.ofMinutes(5);
```
bmaizels, line 47: Can we change this to: ```private static final Duration DEFAULT_PREFETCH_TIME = Duration.ofMinutes(5);```

bmaizels (commented):
Sorry, our static code analysis flagged a couple of trivial things.

services/sts/src/main/java/software/amazon/awssdk/services/sts/auth/StsCredentialsProvider.java
Helmsdown: No worries. In my head it was final. Pushed

services/sts/src/main/java/software/amazon/awssdk/services/sts/auth/StsCredentialsProvider.java
Helmsdown: Done

bmaizels (approved):

bmaizels:
Merged. Thanks for your contribution!

PR #2005, opened 2020-08-26
create connection health checks configuration

zoewangg:
## Description
create `ConnectionHealthChecksConfiguration` and expose it in the crt client builder

## Motivation and Context
https://github.com/aws/aws-sdk-java-v2/blob/master/docs/design/http-clients/aws-crt-client/DecisionLog.md#82420

## Testing
added unit tests

## Types of changes

## Checklist

## License

### DIFF
--- a/http-clients/aws-crt-client/src/main/java/software/amazon/awssdk/http/crt/AwsCrtAsyncHttpClient.java
+++ b/http-clients/aws-crt-client/src/main/java/software/amazon/awssdk/http/crt/AwsCrtAsyncHttpClient.java
@@ -105,11 +105,23 @@ public final class AwsCrtAsyncHttpClient implements SdkAsyncHttpClient {
 
             this.initialWindowSize = builder.initialWindowSize;
             this.maxConnectionsPerEndpoint = maxConns;
-            this.monitoringOptions = builder.monitoringOptions;
+            this.monitoringOptions = revolveHttpMonitoringOptions(builder.connectionHealthChecksConfiguration);
             this.maxConnectionIdleInMilliseconds = config.get(SdkHttpConfigurationOption.CONNECTION_MAX_IDLE_TIMEOUT).toMillis();
 
             this.proxyOptions = buildProxyOptions(builder.proxyConfiguration);
         }
+    }
+
+    private HttpMonitoringOptions revolveHttpMonitoringOptions(ConnectionHealthChecksConfiguration config) {
+        if (config == null) {
+            return null;
+        }
+
+        HttpMonitoringOptions httpMonitoringOptions = new HttpMonitoringOptions();
+        httpMonitoringOptions.setMinThroughputBytesPerSecond(config.minThroughputInBytesPerSecond());
+        int seconds = (int) config.allowableThroughputFailureInterval().getSeconds();
+        httpMonitoringOptions.setAllowableThroughputFailureIntervalSeconds(seconds);
+        return httpMonitoringOptions;
     }
 
     private HttpProxyOptions buildProxyOptions(ProxyConfiguration proxyConfiguration) {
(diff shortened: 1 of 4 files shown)


http-clients/aws-crt-client/src/main/java/software/amazon/awssdk/http/crt/AwsCrtAsyncHttpClient.java
```diff
@@ -105,13 +106,25 @@ private AwsCrtAsyncHttpClient(DefaultBuilder builder, AttributeMap config) {
...
     }
 
+    private HttpMonitoringOptions revolveHttpMonitoringOptions(ConnectionHealthChecksConfiguration config) {
+        if (config == null) {
+            return null;
+        }
+
+        HttpMonitoringOptions httpMonitoringOptions = new HttpMonitoringOptions();
+        httpMonitoringOptions.setMinThroughputBytesPerSecond(config.minThroughputInBytesPerSecond());
+        int seconds = (int) (TimeUnit.MILLISECONDS.toSeconds(config.allowableThroughputFailureInterval().toMillis()));
```
dagnir, line 123: `Duration` has `getSeconds()`

http-clients/aws-crt-client/src/main/java/software/amazon/awssdk/http/crt/ConnectionHealthChecksConfiguration.java
```diff
@@ -0,0 +1,111 @@
...
+
+    /**
+     * @return the minimum amount of throughput, in bytes per second, for a connection to be considered healthy.
+     */
+    public long minThroughputInBytesPerSecond() {
+        return minThroughputInBytesPerSecond;
+    }
+
+    /**
+     * @return How long, in seconds, a connection is allowed to be unhealthy before getting shut down.
```
dagnir, line 46: nit: `Duration` doesn't have a single unit so I don't think it would make sense to say "in seconds" here. We might want to add a note on the setter that it only supports seconds precision though

zoewangg: Yup, copy paste error

http-clients/aws-crt-client/src/main/java/software/amazon/awssdk/http/crt/AwsCrtAsyncHttpClient.java
zoewangg: `getSeconds` only get the seconds from the duration, i.e., not the total number of whole seconds in the duration

`toSeconds` is added since 9,, so we can't use it unfortunately. :(
https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/time/Duration.html#toSeconds()

dagnir: http://hg.openjdk.java.net/jdk9/jdk9/jdk/file/7f644a5d554a/src/java.base/share/classes/java/time/Duration.java they seem to have the same impl :P

dagnir: don't feel strongly about this though

zoewangg: Interesting! Didn't know they have the same impl. Why do they have two APIs then?! 🤔

zoewangg: Found this relevant link: https://bugs.openjdk.java.net/browse/JDK-8142936 
TL;DR They added `toSeconds` to be consistent with other `toXXX` methods

Issue #1801, opened 2020-04-30
DynamoDB Enhanced Client support for Immutable objects


bmaizels:
This issue represents the desire for the DynamoDB Enhanced Client to be able to support Java idiomatic immutable data classes for mapping purposes. At time of writing only mutable 'bean'-like classes with standard getters and setters are supported.

Full support for immutables would likely include :

* An implementation of TableSchema that supported the direct compile-time declaration of attributes for an immutable object, like the StaticTableSchema.
* Would support the common Java idiom of a separate Builder class for the immutable object.
* An implementation of TableSchema that supported introspection of an annotated class to build a static table schema, like the BeanTableSchema.
* Would support inherited immutable (bonus if it can support mutable too) classes with abstract super like StaticTableSchema does.
* Would support composed immutable (bonus if it can support mutable too) classes with flattened attributes like StaticTableSchema does.

bmaizels:
So, I have a work-in-progress POC going on. It has ended up looking like this:

```
    private final TableSchema<ImmutableRecord> immutableTableSchema =
        TableSchema.builder(ImmutableRecord.class, ImmutableRecord.Builder.class)
                   .newItemBuilder(ImmutableRecord::builder, ImmutableRecord.Builder::build)
                   .addAttribute(String.class, a -> a.name("id")
                                                     .getter(ImmutableRecord::id)
                                                     .setter(ImmutableRecord.Builder::id)
                                                     .tags(primaryPartitionKey()))
                   .addAttribute(String.class, a -> a.name("attribute1")
                                                     .getter(ImmutableRecord::attribute1)
                                                     .setter(ImmutableRecord.Builder::attribute1))
                   .addAttribute(int.class, a -> a.name("attribute2")
                                                  .getter(ImmutableRecord::attribute2)
                                                  .setter(ImmutableRecord.Builder::attribute2))
                   .extend(TableSchema.builder(SuperRecord.class, SuperRecord.Builder.class)
                                      .addAttribute(String.class, a -> a.name("attribute3")
                                                                        .getter(SuperRecord::attribute3)
                                                                        .setter(SuperRecord.Builder::attribute3))
                                      .build())
                   .build();
```

I have been able to successfully implement 'extend' to work with an abstract immutable tableschema as long as it has a builder that is a superclass of the child builder. This is not completely intuitive to someone building these data classes, but this is how I accomplished it on the other side :

```
 // Lots of code cut for brevity
[...]

PR #2012, opened 2020-08-30
Immutables support in DynamoDb Enhanced Client

bmaizels:
See README.md changes for description of new features. Fulfills https://github.com/aws/aws-sdk-java-v2/issues/1801

## Testing
New tests were written. All tests pass.

## License


services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/TableMetadata.java
```diff
@@ -72,9 +73,18 @@
      * attribute when using the versioned record extension.
      *
      * @return A collection of all key attribute names for the table.
+     *
+     * @deprecated Use {@link #keyAttributes()} instead.
      */
+    @Deprecated
     Collection<String> allKeys();
 
+    Collection<Index> indices();
```
bmaizels, line 82: I see I missed some javadocs for these new interface methods. Will add those shortly.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/TableSchema.java
```diff
@@ -43,6 +47,21 @@
         return StaticTableSchema.builder(itemClass);
     }
 
+    /**
+     * Returns a builder for the {@link StaticImmutableTableSchema} implementation of this interface which allows all
+     * attributes, tags and table structure to be directly delcared in the builder.
```
bmaizels, line 52: I see a typo in 'delcared'. Will fix.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/mapper/BeanTableSchema.java
```diff
@@ -41,72 +40,61 @@
...
- *     {@literal public List<String> address;}
- * }
- * }
  * </pre>
+ *
  * @param <T> The type of object that this {@link TableSchema} maps to.
  */
 @SdkPublicApi
-public final class BeanTableSchema<T> implements TableSchema<T> {
+public final class BeanTableSchema<T> extends WrappedTableSchema<T, StaticTableSchema<T>> {
```
bmaizels, line 93: I am thinking I should add 'implements TableSchema<T>' here because WrappedTableSchema is an internal class and I want the TableSchema interface to be part of the public contract of this class.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/mapper/ImmutableTableSchema.java
```diff
@@ -0,0 +1,327 @@
...
+ *
+ *         public Customer build() { ... };
+ *     }
+ * }
+ *
+ * </pre>
+ * @param <T> The type of object that this {@link TableSchema} maps to.
+ */
+@SdkPublicApi
+public final class ImmutableTableSchema<T> extends WrappedTableSchema<T, StaticImmutableTableSchema<T, ?>> {
```
bmaizels, line 92: I am thinking I should add 'implements TableSchema<T>' here because WrappedTableSchema is an internal class and I want the TableSchema interface to be part of the public contract of this class.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/TableMetadata.java
```diff
@@ -72,9 +73,36 @@
...
+    /**
+     * Returns metadata about all the known indices for this table.
+     * @return A collection of {@link IndexMetadata} containing information about the indices.
+     */
+    Collection<IndexMetadata> indices();
+
+    /**
+     * Returns all custom metadata for this table. These entries are used by extensions to the library, therefore the
+     * value type of each metadata object stored in the map is not known and is provided as {@link Object}.
+     *
```
zoewangg, line 91: nit, `<p>` 😛

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/TableSchema.java
```diff
@@ -115,4 +172,17 @@
...
+     */
+    List<String> attributeNames();
+
+    /**
+     * A boolean value that represents whether this {@link TableSchema} is abstract.
+     *
+     * @return true if it is abstract, and therefore cannot be used directly to create records but can be referred to
+     * by other schemata, and false if it is concrete and may be used to map records directly.
+     */
+    boolean isAbstract();
```
zoewangg, line 187: Can we add more javadocs to explain the definition of `abstract` here since abstract is an overloaded term?

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/immutable/ImmutableIntrospector.java
```diff
@@ -0,0 +1,226 @@
...
+        return Optional.of(buildMethod);
+    }
+
+    private Optional<Method> extractSetterMethod(String propertyName,
+                                                 Map<String, Method> indexedBuilderMethods,
+                                                 Method getterMethod,
+                                                 Class<?> builderClass) {
+        Method setterMethod = indexedBuilderMethods.get(propertyName);
+
+        if (setterMethod == null
```
zoewangg, line 213: Can we break up the if statement a bit, like creating variables with meaningful names? seems a bit hard to read with so many predicates.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/immutable/ImmutableIntrospector.java
```diff
@@ -0,0 +1,226 @@
...
+
+        if (setterName.length() > 3
+            && Character.isUpperCase(setterName.charAt(3))
+            && setterName.startsWith(SET_PREFIX)) {
+
+            return Character.toLowerCase(setterName.charAt(3)) + setterName.substring(4);
+        }
+
+        return setterName;
+    }
```
millems, line 166: Should we make "is bean style" an annotation setting, instead of trying to guess by looking at the method names? I'm worried about an attribute starting with "Set" being magically messed up because we're trying to be too clever. 

E.g. A customer probably wants a "setOfStuff" attribute to be stored as "setOfStuff", not "ofStuff".

bmaizels: Some code generators will produce immutable implementations that have 'set'. I think it would be inconvenient to require an additional annotation for all of those. I feel it would be preferable to annotate the exceptions with @DynamoDbAttribute("setOfStuff") (which is already supported)

services-custom/dynamodb-enhanced/README.md
```diff
@@ -155,6 +155,72 @@ index. Here's an example of how to do this:
...
+        private int subId;        
+        private String name;
+        private Instant createdDate;
+
+        private Builder() {}
+
+        public Builder accountId(String accountId) { this.accountId = accountId; return this; }
+        public Builder subId(int subId) { this.subId = subId; return this; }
+        public Builder name(String name) { this.name = name; return this; }
+        public Builder createdDate(Instant createdDate) { this.createdDate = createdDate; return this; }
```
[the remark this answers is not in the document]
bmaizels: It does check but it doesn't have to. The bean introspector is really strict about setters can only return void. What do you think is best for the builder? Right now I think it only allows it to return a builder, but we could also allow void if we wanted to.

services-custom/dynamodb-enhanced/README.md
```diff
@@ -155,6 +155,72 @@ index. Here's an example of how to do this:
...
+    
+    private Customer(Builder b) {
+        this.accountId = b.accountId;
+        this.subId = b.subId;
+        this.name = b.name;
+        this.createdDate = b.createdDate;
+    }   
+
+    // This method will be automatically discovered and used by the TableSchema
+    public static Builder builder() { return new Builder(); }
```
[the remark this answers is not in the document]
bmaizels: It searches for a static method called 'builder' on the class.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/TableMetadata.java
bmaizels: Will fix.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/TableSchema.java
bmaizels: Okay. +1

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/immutable/ImmutableIntrospector.java
bmaizels: Okay. +1

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/immutable/ImmutableIntrospector.java
```diff
@@ -0,0 +1,226 @@
...
+    private static final String GET_PREFIX = "get";
+    private static final String IS_PREFIX = "is";
+    private static final String SET_PREFIX = "set";
+
+    // Methods from Object are commonly overridden and confuse the mapper, automatically exclude any method with a name
+    // that matches a method defined on Object.
+    private static final Set<String> NAMES_TO_EXCLUDE =
+        Collections.unmodifiableSet(Arrays.stream(Object.class.getMethods())
+                                          .map(Method::getName)
+                                          .collect(Collectors.toSet()));
```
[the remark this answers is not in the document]
bmaizels: I was trying not to synchronize as the static class loader path is guaranteed to be thread synchronized. Do you think it would be preferable to do a synchronized double-null-check type lazy initialization here?

services-custom/dynamodb-enhanced/README.md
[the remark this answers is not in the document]
bmaizels: We could also not care about the return type at all (another option)

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/immutable/ImmutableIntrospector.java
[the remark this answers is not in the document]
millems: Since it's not a big deal if we double-initialize, doing a synchronized double-null-check lazy initialization wouldn't even need to be on a volatile variable. It would probably be the best of both worlds.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/immutable/ImmutableIntrospector.java
millems: I was thinking a boolean on the class-level annotation, instead of one on every attribute. That seems less troublesome and prevents magic confusion (I feel like this is the opposite argument we usually have - you're usually the anti-magic one!)

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/immutable/ImmutableIntrospector.java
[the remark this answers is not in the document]
bmaizels: Okay, will do that and make it non-volatile.

services-custom/dynamodb-enhanced/README.md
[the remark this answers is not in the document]
bmaizels: Looking at the code again, we do accept void as well, so both cases are covered. It will ignore methods that return an unexpected type though.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/immutable/ImmutableIntrospector.java
bmaizels: As we discussed offline, we'll bias towards making it easy for customers using code generators such as Lombok and not require an additional opt-in flag for getter/setter prefixes.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/immutable/ImmutableIntrospector.java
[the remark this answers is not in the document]
bmaizels: As discussed offline and due to spotbug's protests we'll make this volatile after all.

PR #2014, opened 2020-09-01
Add create method for Apache and Netty HttpClient

Quanzzzz:
Add create method for Apache and Netty HttpClient.

## License

### DIFF
--- a/http-clients/apache-client/src/main/java/software/amazon/awssdk/http/apache/ApacheHttpClient.java
+++ b/http-clients/apache-client/src/main/java/software/amazon/awssdk/http/apache/ApacheHttpClient.java
@@ -133,6 +133,15 @@ public final class ApacheHttpClient implements SdkHttpClient {
         return new DefaultBuilder();
     }
 
+    /**
+     * Create a {@link ApacheHttpClient} with the default properties
+     *
+     * @return an {@link ApacheHttpClient}
+     */
+    public static SdkHttpClient create() {
+        return new DefaultBuilder().build();
+    }
+
     private ConnectionManagerAwareHttpClient createClient(ApacheHttpClient.DefaultBuilder configuration,
                                                           AttributeMap standardOptions) {
         ApacheConnectionManagerFactory cmFactory = new ApacheConnectionManagerFactory();
--- a/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/NettyNioAsyncHttpClient.java
+++ b/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/NettyNioAsyncHttpClient.java
@@ -126,6 +126,15 @@ public final class NettyNioAsyncHttpClient implements SdkAsyncHttpClient {
         return new DefaultBuilder();
     }
 
+    /**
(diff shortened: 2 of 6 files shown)


http-clients/apache-client/src/main/java/software/amazon/awssdk/http/apache/ApacheHttpClient.java
```diff
@@ -133,6 +134,15 @@ public static Builder builder() {
         return new DefaultBuilder();
     }
 
+    /**
+     * Create a {@link HttpConnection} client with the default properties
+     *
+     * @return an {@link ApacheHttpClient}
+     */
+    public static SdkHttpClient create() {
+        return new DefaultBuilder().build();
```
joviegas, line 143: Query : Just wondering if we need to create it with Default value as below
```
    public static SdkHttpClient create() {
        return new DefaultBuilder().buildWithDefaults(AttributeMap.empty());
    }
```
This will end up in initializing the values with [GLOBAL_HTTP_DEFAULTS](https://github.com/aws/aws-sdk-java-v2/blob/cdfb7d3f6a7dafb37be84f6e95e49feb0d0af273/http-client-spi/src/main/java/software/amazon/awssdk/http/SdkHttpConfigurationOption.java#L130)

else we end up with Default build as [SdkHttpClient Builder](https://github.com/aws/aws-sdk-java-v2/blob/cdfb7d3f6a7dafb37be84f6e95e49feb0d0af273/http-client-spi/src/main/java/software/amazon/awssdk/http/SdkHttpClient.java#L67-L69)

When we create should it   created with Empty values or with GLOBAL_HTTP_DEFAULTS ?

http-clients/apache-client/src/main/java/software/amazon/awssdk/http/apache/ApacheHttpClient.java
```diff
@@ -133,6 +134,15 @@ public static Builder builder() {
...
+    }
```
millems, line 144: Can we just say ApacheHttpClient instead of HttpConnection? They shouldn't care about implementation details. Same with Netty below.

http-clients/apache-client/src/main/java/software/amazon/awssdk/http/apache/ApacheHttpClient.java
```diff
@@ -43,6 +43,7 @@
 import javax.net.ssl.TrustManager;
 import javax.net.ssl.X509TrustManager;
 import org.apache.http.Header;
+import org.apache.http.HttpConnection;
```
millems, line 46: Unused import.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/NettyNioAsyncHttpClient.java
```diff
@@ -24,6 +24,7 @@
 
 import io.netty.channel.ChannelOption;
 import io.netty.channel.EventLoopGroup;
+import io.netty.handler.codec.http2.Http2Connection;
```
millems, line 27: Unused import.

PR #2018, opened 2020-09-03
Build method for concrete classes of StsCredentialsProvider was not s…

joviegas:
## Motivation and Context
 All  concrete classes of [StsCredentialsProvider link](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/sts/auth/package-summary.html) did not show Public Build method in Java docs.
 
Why is this change required? What problem does it solve? -->  Public methods should be showed in Java Docs and build method which was public was not shown.
 
### Root Cause :- 
The BaseBuilder is protected inner in Abstract parent class StsCredentialsProvider  which is of type @SdkInternalApi. Since BaseBuilder was protected all the concrete methods of StsCredentialsProvider didnot show in Java Docs
 
### Solution :- 
Overriden Build method as public in all the concrete methods where they just call super.buuild.
 
#### Alternate Approaches
1. Make BaseBuilder as Public from protected :- This approach is not feasible since BaseBuilder is inner class StsCredentialsProvider @SdkInternalApi. Now since BaseBuilder is in a class which is of @SdkInternalApi the java Docs Links are not functional so java doc no generated for BaseBuilder.
 
2. Make BaseBuilder as separate independent Public Static class :- This would disrupt the structure in which the package is written where all the Builder are inner classes of the abstract classes. Also there are other two methods like stsClient and asyncCredentialUpdateEnabled which looks are protected builder methods for factory methods and not for external usage.
 
## Testing
```
mvn install -P quick -T1C && mvn install javadoc:aggregate -B -Ppublic-javadoc -Dcheckstyle.skip -Dspotbugs.skip -DskipTests -Ddoclint=none -pl '!:protocol-tests,!:protocol-tests-core,!:codegen-generated-classes-test,!:sdk-benchmarks,!:module-path-tests,!:test-utils,!:http-client-tests,!:tests-coverage-reporting'
```
```
build
public StsAssumeRoleCredentialsProvider build()
Build the credentials provider using the configuration applied to this builder.
```
 
 
 
 Bug fix (non-breaking change which fixes an issue)
 
## Checklist
 
## License


services/sts/src/main/java/software/amazon/awssdk/services/sts/auth/StsAssumeRoleCredentialsProvider.java
```diff
@@ -117,5 +117,13 @@ public Builder refreshRequest(Supplier<AssumeRoleRequest> assumeRoleRequestSuppl
         public Builder refreshRequest(Consumer<AssumeRoleRequest.Builder> assumeRoleRequest) {
             return refreshRequest(AssumeRoleRequest.builder().applyMutation(assumeRoleRequest).build());
         }
+
+        /**
+         * {@inheritDoc}
+         */
+        public StsAssumeRoleCredentialsProvider build() {
```
millems, line 124: @Override? Also, inheritDoc is the default behavior, so we shouldn't need to say it explicitly.

millems (approved):

PR #2031, opened 2020-09-11
Update codegen module to generate service waiter implementation classes

zoewangg:
## Description
Update codegen module to generate service waiter implementation classes
Note that it currently only has one `WaiterAcceptor` and we will add more once waiterAcceptor generation part is done.
More waiter tests will be added as well after we have the complete waiter code

## Types of changes

## Checklist

## License


codegen/src/main/java/software/amazon/awssdk/codegen/poet/waiters/AsyncWaiterClassSpec.java
```diff
@@ -0,0 +1,122 @@
...
+    @Override
+    protected Optional<String> additionalWaiterConfig() {
+        return Optional.of(".scheduledExecutorService(executorService)");
+    }
+
+    @Override
+    protected void additionalConstructorInitialization(MethodSpec.Builder method) {
+        method.beginControlFlow("if (builder.executorService == null)")
+              .addStatement("this.executorService = $T.newScheduledThreadPool(5, new $T().threadNamePrefix"
+                            + "(\"waiters-ScheduledExecutor\").build())",
```
dagnir, line 88: minor: use `$S` here so you don't need to escape the quotes

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/waiters/query-async-waiter-interface.java
```diff
@@ -35,11 +37,23 @@
...
+         * Sets a custom {@link ScheduledExecutorService} that will be used to schedule async polling attempts
+         * <p>
+         * This executorService must be closed by the caller when it is ready to be disposed. The SDK will not close the
+         * executorService when the waiter is closed
+         *
+         * @param executorService
+         *        the executorService to set
+         * @return a reference to this object so that method calls can be chained together.
+         */
+        Builder executorService(ScheduledExecutorService executorService);
```
dagnir, line 55: Should we have provider overload, where the returned executors are managed by the SDK?

codegen/src/main/java/software/amazon/awssdk/codegen/poet/waiters/AsyncWaiterClassSpec.java
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    @Override
+    protected Optional<String> additionalWaiterConfig() {
+        return Optional.of(".scheduledExecutorService(executorService)");
+    }
+
+    @Override
+    protected void additionalConstructorInitialization(MethodSpec.Builder method) {
+        method.beginControlFlow("if (builder.executorService == null)")
```
dagnir, line 86: Hmm why does the async waiter need a threadpool? Would it be enough to just rely on the async client?

zoewangg: We still need a `ScheduledExecutorService` to schedule the attempts.

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/waiters/query-async-waiter-interface.java
zoewangg: I like the idea, but we don't have it for the futureCompletionExecutorService override either. I feel this is additive and we can add it in the future if people are asking for it. I'd expect most customers use the waiters retrieved the service clients, where they don't need to configure it at all.

codegen/src/main/java/software/amazon/awssdk/codegen/poet/waiters/AsyncWaiterClassSpec.java
dagnir: Ah gotcha forgot about the backoffs

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/waiters/query-async-waiter-interface.java
dagnir: Sounds good

dagnir (approved):

PR #2035, opened 2020-09-14
DDB Enhanced:Added support to read Nested objects in attributesToProj…

joviegas:
## Motivation and Context
- [DDB Enhanced: Added support for projection expression #1756](https://github.com/aws/aws-sdk-java-v2/pull/1756) missed out to support projecting Nested Attributes
- with this fix we can project Nested attributes

## Fix

- Added a new Class NestedAttributeName , this can be used to represent Nested Variables
```

/**
* Public API methods for creating NestedAttributeName
*/

//Simple Attribute creation "foo"
NestedAttributeName s1 = AttributeName.builder().elements("foo").build();
s1 = NestedAttributeName.builder().addElement("foo").build()
s1 = NestedAttributeName.create("foo")

//Nested Attribute creation "foo.bar"
s2 = NestedAttributeName.builder().elements("foo", "bar").build();
s2 = NestedAttributeName.create("foo", "bar");


//Getter Output for s1
s1.elements() = ["foo"]
//Getter Output for s2
s2.elements() = ["foo", "bar"]

```

EnhancedRequest
```
/**
*SAMPLE Builder usage And GETTERS for EnhancedRequest
*/   

/***Case1*: One Attribute and attributesToProject(String) used**/
 final QueryEnhancedRequest q1 = 
 QueryEnhancedRequest.builder().attributesToProject("foo").build();
 
 q1.attributesToProject() = ["foo"].
 q1.nestedAttributeNameToProject() = ["foo"]
 
 /***Case2* :One Nested Attribute **/
 final QueryEnhancedRequest q1 = 
 QueryEnhancedRequest.builder()
    .addNestedAttributesToProject(NestedAttributeName.create("foo", "bar")).build();
 
 q1.attributesToProject() = ["foo.bar"].
 q1.nestedAttributeNameToProject() = [["foo","bar"]]
 
  /***Case3* :One Nested Attribute and One Non Nested Attribute **/
 final QueryEnhancedRequest q1 = 
 QueryEnhancedRequest.builder()
    .attributesToProject("bazz")
    .addNestedAttributesToProject(NestedAttributeName.create("foo", "bar")).build();
 
 q1.attributesToProject() = ["bazz","foo.bar"].
 q1.nestedAttributeNameToProject() = [["bazz"], ["foo","bar"]]
 
 
 
```

## Testing

Added JUnit for following casses
1.Projection of a nested Attribute .
2. Projection of a nested Attribute and not nested attribute (using older builder method).
3. Projection of Multiple Attributes
4. Projection of Null attributes passing Null  NestedAttributeName --> Returns all elements as per the doc.

5. Tested getter for attributeForProject.

## Types of changes

## Checklist

## License

### DIFF
--- a/services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/NestedAttributeName.java
+++ b/services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/NestedAttributeName.java
@@ -1,0 +1,235 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.enhanced.dynamodb;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import software.amazon.awssdk.annotations.SdkPublicApi;
+import software.amazon.awssdk.utils.Validate;
+
+/**
+ * High-level representation of a DynamoDB 'NestedAttributeName' that can be used in various situations where the API requires
+ * or accepts an Nested Attribute Name.
+ * Simple Attribute Name can be represented by passing just the name of the attribute.
+ * Nested Attributes are represented by List of String where each index of list corresponds to Nesting level Names.
+ * <p> While using attributeToProject in {@link software.amazon.awssdk.enhanced.dynamodb.model.QueryEnhancedRequest}
+ * and {@link software.amazon.awssdk.enhanced.dynamodb.model.ScanEnhancedRequest} we need  way to represent Nested Attributes.
+ * The normal DOT(.) separator is not recognized as a Nesting level separator by DynamoDB request,
+ * thus we need to use NestedAttributeName
+ * which can be used to represent Nested attributes.
+ * <p> Example : NestedAttributeName.create("foo") corresponds to a NestedAttributeName with elements list
+ * with single element foo which represents Simple attribute name "foo" without nesting.
+ * <p>NestedAttributeName.create("foo", "bar") corresponds to a NestedAttributeName with elements list "foo", "bar"
+ * respresenting nested attribute name "foo.bar".
+ */
+@SdkPublicApi
+public final class NestedAttributeName {
+
+    private final List<String> elements;
+
+    private NestedAttributeName(List<String> nestedAttributeNames) {
+        Validate.validState(nestedAttributeNames != null, "nestedAttributeNames must not be null.");
+        Validate.notEmpty(nestedAttributeNames, "nestedAttributeNames must not be empty");
+        Validate.noNullElements(nestedAttributeNames, "nestedAttributeNames must not contain null values");
+        this.elements =  Collections.unmodifiableList(nestedAttributeNames);
+    }
+
+    /**
+     * Creates a NestedAttributeName with a single element, which is effectively just a simple attribute name without nesting.
+     * <p>
+     * <b>Example:</b>create("foo") will create NestedAttributeName corresponding to Attribute foo.
+     *
+     * @param element Attribute Name. Single String represents just a simple attribute name without nesting.
+     * @return NestedAttributeName with attribute name as specified element.
+     */
+    public static NestedAttributeName create(String element) {
+        return new Builder().addElement(element).build();
+    }
+
+    /**
+     * Creates a NestedAttributeName from a list of elements that compose the full path of the nested attribute.
+     * <p>
+     * <b>Example:</b>create("foo", "bar") will create NestedAttributeName which represents foo.bar nested attribute.
+     *
+     * @param elements Nested Attribute Names. Each of strings in varargs represent the nested attribute name
+     *                 at subsequent levels.
+     * @return NestedAttributeName with Nested attribute name set as specified in elements var args.
+     */
+    public static NestedAttributeName create(String... elements) {
+        return new Builder().elements(elements).build();
+    }
+
+    /**
+     * Creates a NestedAttributeName from a list of elements that compose the full path of the nested attribute.
+     * <p>
+     * <b>Example:</b>create(Arrays.asList("foo", "bar")) will create NestedAttributeName
+     * which represents foo.bar nested attribute.
+     *
+     * @param elements List of Nested Attribute Names. Each of strings in List represent the nested attribute name
+     *                 at subsequent levels.
+     * @return NestedAttributeName with Nested attribute name set as specified in elements Collections.
+     */
+    public static NestedAttributeName create(List<String> elements) {
+        return new Builder().elements(elements).build();
+    }
+
+    /**
+     * Create a builder that can be used to create a {@link NestedAttributeName}.
+     */
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    /**
+     * Gets elements of NestedAttributeName in the form of List. Each element in the list corresponds
+     * to the subsequent Nested Attribute name.
+     *
+     * @return List of nested attributes, each entry in the list represent one level of nesting.
+     * Example, A Two level Attribute name foo.bar will be represented as  ["foo", "bar"]
+     */
+    public List<String> elements() {
+        return elements;
+    }
+
+    /**
+     * Returns a builder initialized with all existing values on the request object.
+     */
+    public Builder toBuilder() {
+        return builder().elements(elements);
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        NestedAttributeName that = (NestedAttributeName) o;
+
+        return elements != null
+                ? elements.equals(that.elements) : that.elements == null;
+    }
+
+    @Override
+    public int hashCode() {
+        return elements != null ? elements.hashCode() : 0;
+    }
+
+    /**
+     * A builder for {@link NestedAttributeName}.
+     */
+    public static class Builder {
+        private List<String> elements = null;
+
+        private Builder() {
+
+        }
+
+        /**
+         * Adds a single element of NestedAttributeName.
+         * Subsequent calls to this method can add attribute Names at subsequent nesting levels.
+         * <p>
+         * <b>Example:</b>builder().addElement("foo").addElement("bar")  will add elements in NestedAttributeName
+         * which represent a Nested Attribute Name foo.bar
+         *
+         * @param element Attribute Name.
+         * @return Returns a reference to this object so that method calls can be chained together.
+         */
+        public Builder addElement(String element) {
+            if (elements == null) {
+                elements = new ArrayList<>();
+            }
+            elements.add(element);
+            return this;
+        }
+
+        /**
+         * Adds a single element of NestedAttributeName.
+         * Subsequent calls to this method will append the new elements to the end of the existing chain of elements
+         * creating new levels of nesting.
+         * <p>
+         * <b>Example:</b>builder().addElements("foo","bar")  will add elements in NestedAttributeName
+         * which represent a Nested Attribute Name foo.bar
+         *
+         * @param elements Nested Attribute Names. Each of strings in varargs represent the nested attribute name
+         *                 at subsequent levels.
+         * @return Returns a reference to this object so that method calls can be chained together.
+         */
+        public Builder addElements(String... elements) {
+            if (this.elements == null) {
+                this.elements = new ArrayList<>();
+            }
+            this.elements.addAll(Arrays.asList(elements));
+            return this;
+        }
+
+        /**
+         * Adds a List of elements to NestedAttributeName.
+         * Subsequent calls to this method will append the new elements to the end of the existing chain of elements
+         * creating new levels of nesting.
+         * <p>
+         * <b>Example:</b>builder().addElements(Arrays.asList("foo","bar"))  will add elements in NestedAttributeName
+         * to represent a Nested Attribute Name foo.bar
+         *
+         * @param elements List of Strings where each string corresponds to subsequent nesting attribute name.
+         * @return Returns a reference to this object so that method calls can be chained together.
+         */
+        public Builder addElements(List<String> elements) {
+            if (this.elements == null) {
+                this.elements = new ArrayList<>();
+            }
+            this.elements.addAll(elements);
+            return this;
+        }
+
+        /**
+         * Set elements of NestedAttributeName with list of Strings. Will overwrite any existing elements stored by this builder.
+         * <p>
+         * <b>Example:</b>builder().elements("foo","bar") will set the elements in NestedAttributeName
+         * to represent a nested attribute name of 'foo.bar'
+         *
+         * @param elements a list of strings that correspond to the elements in a nested attribute name.
+         * @return Returns a reference to this object so that method calls can be chained together.
+         */
+        public Builder elements(String... elements) {
+            this.elements = new ArrayList<>(Arrays.asList(elements));
+            return this;
+        }
+
+        /**
+         * Sets the elements that compose a nested attribute name. Will overwrite any existing elements stored by this builder.
+         * <p>
+         * <b>Example:</b>builder().elements(Arrays.asList("foo","bar")) will add elements in NestedAttributeName
+         * which represent a Nested Attribute Name foo.bar
+         *
+         * @param elements a list of strings that correspond to the elements in a nested attribute name.
+         * @return Returns a reference to this object so that method calls can be chained together.
+         */
+        public Builder elements(List<String> elements) {
+            this.elements = new ArrayList<>(elements);
+            return this;
+        }
+
+
+        public NestedAttributeName build() {
+            return new NestedAttributeName(elements);
+        }
+    }
+}
--- a/services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/ProjectionExpressionConvertor.java
+++ b/services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/ProjectionExpressionConvertor.java
@@ -1,0 +1,107 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.enhanced.dynamodb.internal;
+
+import static software.amazon.awssdk.enhanced.dynamodb.internal.EnhancedClientUtils.cleanAttributeName;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.function.UnaryOperator;
+import java.util.stream.Collectors;
+import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.enhanced.dynamodb.NestedAttributeName;
+
+/**
+ * Wrapper method to get Projection Expression Name map and Projection Expressions from NestedAttributeNames.
+ */
+@SdkInternalApi
+public class ProjectionExpressionConvertor {
+
+    private static final String AMZN_MAPPED = "#AMZN_MAPPED_";
+    private static final UnaryOperator<String> PROJECTION_EXPRESSION_KEY_MAPPER = k -> AMZN_MAPPED + cleanAttributeName(k);
+    private final List<NestedAttributeName> nestedAttributeNames;
+
+    private ProjectionExpressionConvertor(List<NestedAttributeName> nestedAttributeNames) {
+        this.nestedAttributeNames = nestedAttributeNames;
+    }
+
+    public static ProjectionExpressionConvertor create(List<NestedAttributeName> nestedAttributeNames) {
+        return new ProjectionExpressionConvertor(nestedAttributeNames);
+    }
+    
+    private static Optional<Map<String, String>> convertToExpressionNameMap(NestedAttributeName attributeName) {
+        List<String> nestedAttributeNames = attributeName.elements();
+        if (nestedAttributeNames != null) {
+            Map<String, String> resultNameMap = new LinkedHashMap<>();
+            nestedAttributeNames.stream().forEach(nestedAttribute ->
+                            resultNameMap.put(PROJECTION_EXPRESSION_KEY_MAPPER.apply(nestedAttribute), nestedAttribute));
+            return Optional.of(resultNameMap);
+        }
+        return Optional.empty();
+    }
+
+    private static Optional<String> convertToNameExpression(NestedAttributeName nestedAttributeName) {
+
+        String name = nestedAttributeName.elements().stream().findFirst().orElse(null);
+
+        List<String> nestedAttributes = null;
+        if (nestedAttributeName.elements().size() > 1) {
+            nestedAttributes = nestedAttributeName.elements().subList(1, nestedAttributeName.elements().size());
+        }
+        if (name != null) {
+            List<String> hashSeparatedNestedStringList =
+                    new ArrayList<>(Arrays.asList(PROJECTION_EXPRESSION_KEY_MAPPER.apply(name)));
+            if (nestedAttributes != null) {
+                nestedAttributes.stream().forEach(hashSeparatedNestedStringList::add);
+            }
+            return Optional.of(String.join(".".concat(AMZN_MAPPED), hashSeparatedNestedStringList));
+        }
+        return Optional.empty();
+    }
+
+    public List<NestedAttributeName> nestedAttributeNames() {
+        return nestedAttributeNames;
+    }
+
+    public Map<String, String> convertToExpressionMap() {
+        Map<String, String> attributeNameMap = new LinkedHashMap<>();
+        if (this.nestedAttributeNames() != null) {
+            this.nestedAttributeNames().stream().forEach(attribs -> convertToExpressionNameMap(attribs)
+                    .ifPresent(attributeNameMap::putAll));
+        }
+        return attributeNameMap;
+    }
+
+    public Optional<String> convertToProjectionExpression() {
+        if (nestedAttributeNames != null) {
+            List<String> expressionList = new ArrayList<>();
+            this.nestedAttributeNames().stream().filter(Objects::nonNull)
+                    .filter(item -> item.elements() != null && !item.elements().isEmpty())
+                    .forEach(attributeName -> convertToNameExpression(attributeName)
+                            .ifPresent(expressionList::add));
+            String joinedExpression = String.join(",", expressionList.stream()
+                    .distinct().collect(Collectors.toList()));
+            return Optional.ofNullable(joinedExpression.isEmpty() ? null : joinedExpression);
+        }
+        return Optional.empty();
+    }
+
+}
--- a/services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/model/QueryEnhancedRequest.java
+++ b/services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/model/QueryEnhancedRequest.java
@@ -22,11 +22,14 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 import software.amazon.awssdk.annotations.SdkPublicApi;
 import software.amazon.awssdk.enhanced.dynamodb.DynamoDbAsyncIndex;
 import software.amazon.awssdk.enhanced.dynamodb.DynamoDbTable;
 import software.amazon.awssdk.enhanced.dynamodb.Expression;
+import software.amazon.awssdk.enhanced.dynamodb.NestedAttributeName;
 import software.amazon.awssdk.services.dynamodb.model.AttributeValue;
+import software.amazon.awssdk.utils.Validate;
 
 /**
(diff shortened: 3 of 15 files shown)


services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/AttributeName.java
```diff
@@ -0,0 +1,140 @@
...
+ * Optionally, nested attribute names can be added with tokens using the addNestedAttributeName .
+ * Every call to addNestedAttributeName represent next level of Nesting Attribute.
+ * If we want to denote three level of Nesting attribute name then addNestedAttributeName should be called thrice.
+ *
+ * <p>
+ * Example, if we want to represent a two level nested attribute outerAttribute.innerAttributeLevel1.innerAttributeLevel2
+ * then it can be represented as :-
+ * {@code
+ * AttributeName myAttributeName = AttributeName.builder()
+ * .name("outerAttribute).addNested("innerAttributeLevel1).nestedAttributeName("innerAttributeLevel2").build();
```
bmaizels, line 38: `  * .name("outerAttribute).addNested("innerAttributeLevel1).addNested("innerAttributeLevel2").build();`

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/AttributeName.java
```diff
@@ -0,0 +1,140 @@
...
+ * }
+ */
+@SdkPublicApi
+public final class AttributeName {
+
+    private final String name;
+    private final List<String> nestedAttributeNames;
```
bmaizels, line 45: Ignoring the builder entirely, I think we should store this just as a single list. If it's a single element, then it's not nested, if there are multiple elements then it's nested.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/AttributeName.java
```diff
@@ -0,0 +1,140 @@
...
+     * @return Name of the attribute.
+     */
+    public String getName() {
+        return name;
+    }
+
+    /**
+     * @return List of nested attributes , each entry in the list represent one level of nesting.
+     */
+    public List getNestedAttributeNames() {
```
bmaizels: Why are you returning a raw-type list here?

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/AttributeName.java
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    /**
+     * This is the final step in charge of building the AttributeName Object.
+     */
+    public interface BuildStep {
+        BuildStep addNested(String nestedAttributeName);
+
+        AttributeName build();
+    }
```
bmaizels, line 91: We avoid this builder pattern, we think it creates more complexity and confusion than it solves. We should try and design a simple builder that intuitively works for both use cases. This may not be the best solution but whatever solution we come up with has to work for all the scenarios. E.g.:

```
// All of these are equivalent
AttributeName.builder().name("foo").addNestedNames("bar").build();
AttributeName.builder().addNestedNames("foo", bar");
AttributeName.builder().addNestedNames("foo").addNestedNames("bar");
```

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/AttributeName.java
```diff
@@ -0,0 +1,140 @@
...
+     * A builder for {@link AttributeName}
+     * This builder follows BuilderStep pattern.
+     * The Attribute Name always comes first followed by optional Nested Attribute Name.
+     */
+    public static class Builder implements AttributeNameBuilder, BuildStep {
+        private final List<String> nestedAttributeNames = new ArrayList<>();
+        private String attributeToProject;
+
+        private Builder() {
+
```
bmaizels, line 103: Nit: remove this empty line.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/AttributeName.java
```diff
@@ -0,0 +1,140 @@
...
+        AttributeName build();
+    }
+
+    /**
+     * A builder for {@link AttributeName}
+     * This builder follows BuilderStep pattern.
+     * The Attribute Name always comes first followed by optional Nested Attribute Name.
+     */
+    public static class Builder implements AttributeNameBuilder, BuildStep {
+        private final List<String> nestedAttributeNames = new ArrayList<>();
```
bmaizels, line 99: Our pattern is to initialize this to null like other parameters, and when they add something check if it's null and create a new ArrayList then if necessary. This also allows us to provide a setter method, although it's important when we let them set the whole list we always create our own new ArrayList to avoid mutation of the list outside of our control after object created : 

```
public Builder nestedAttributeNames(List<String> nestedAttributeNames) { ... }
```

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/AttributeName.java
```diff
@@ -0,0 +1,140 @@
...
+ * .name("outerAttribute).addNested("innerAttributeLevel1).nestedAttributeName("innerAttributeLevel2").build();
+ * }
+ */
+@SdkPublicApi
+public final class AttributeName {
+
+    private final String name;
+    private final List<String> nestedAttributeNames;
+
+    private AttributeName(String name, List<String> nestedAttributeNames) {
```
bmaizels, line 47: Minor: Our standard pattern is to have a constructor that takes an instance of a builder. Eg:

```
private AttributeName(Builder b) { ... }
```

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/AttributeName.java
```diff
@@ -0,0 +1,140 @@
...
+        return new Builder();
+    }
+
+    /**
+     * @return Name of the attribute.
+     */
+    public String getName() {
+        return name;
+    }
+
```
bmaizels, line 68: We should add a toBuilder() method and override equals() and hashcode().

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/model/QueryEnhancedRequest.java
```diff
@@ -47,6 +48,7 @@
     private final Boolean consistentRead;
     private final Expression filterExpression;
     private final List<String> attributesToProject;
+    private final List<AttributeName> attributeNamesToProject;
```
bmaizels, line 51: My opinion is that we should deprecate the List<String> and just have a List<AttributeName>. That will simplify our behavioral logic around this.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/model/QueryEnhancedRequest.java
```diff
@@ -347,6 +366,92 @@ public Builder addAttributeToProject(String attributeToProject) {
...
+         * @param attributeNameToProject
+         *        A collection of the attributes names to be retrieved from the database.
+         *        Nested levels of Attributes can be added using AttributeName class.
+         *        AttributeName.builder() can be used to created AttributeNames.
+         *        AttributeName.builder().name("simpleAttribute").build
+         *        or for Nested attribute
+         *        AttributeName.builder().name("outerField").addNested("innerLevel1")..addNested("innerLevel2")build();
+         * @return Returns a reference to this object so that method calls can be chained together.
+         */
+        public Builder attributeNameToProject(Collection<AttributeName> attributeNameToProject) {
```
bmaizels, line 394: Can we talk more offline about the design of this.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/operations/utils/OperationUtils.java
```diff
@@ -0,0 +1,121 @@
...
+import java.util.function.UnaryOperator;
+import java.util.stream.Collectors;
+import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.enhanced.dynamodb.AttributeName;
+
+/**
+ * Contains Util methods used in Operations.
+ */
+@SdkInternalApi
+public final class OperationUtils {
```
bmaizels, line 36: Can we figure out a way not to have this class at all. Seems like these methods can be divided into two groups, one that acts on a List of AttributeNames and one that acts on a single AttributeName. Maybe we can create two dynamic classes that are constructed from those respective entities and provide these methods.

dagnir:
Looks like the previous PR was closed; for future PR's please use the existing pull request when making updates so we can maintain the history

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/model/QueryEnhancedRequest.java
```diff
@@ -125,9 +128,20 @@ public Expression filterExpression() {
 
     /**
      * Returns the list of projected attributes on this request object, or an null if no projection is specified.
+     * The Nested Attributes are represented using DOT separator.
+     * Example : foo.bar is represnted as "foo.bar". Use attributeNamesToProject whenever Nested attributes are used.
```
bmaizels, line 132:
```
Example : foo.bar is represented as "foo.bar" which is indistinguishable from a non-nested attribute with the name "foo.bar". Use {@link #nestedAttributesToProject} if you have a use-case that requires discrimination between these two cases.
```

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/model/QueryEnhancedRequest.java
```diff
@@ -125,9 +128,20 @@ public Expression filterExpression() {
...
      */
     public List<String> attributesToProject() {
-        return attributesToProject;
+        return nestedAttributesToProject != null ? nestedAttributesToProject.stream().filter(Objects::nonNull)
+                        .map(item -> String.join(".", item.elements())).collect(Collectors.toList()) : null;
+    }
+
+    /**
+     * Returns the list of projected Nested attributes Name on this request object, or an null if no projection is specified.
```
bmaizels, line 140:
```
Returns the list of projected attribute names, in the form of {@link NestedAttributeName} objects, for this request object, or null if no projection is specified.
```

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/model/QueryEnhancedRequest.java
```diff
@@ -193,6 +205,7 @@ public int hashCode() {
         private Boolean consistentRead;
         private Expression filterExpression;
         private List<String> attributesToProject;
```
bmaizels, line 207: Can we remove this and convert attribute name strings as they are added? If we don't think could cause unexpected and surprising behavior when they overwrite existing attributes

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/model/QueryEnhancedRequest.java
```diff
@@ -347,7 +360,101 @@ public Builder addAttributeToProject(String attributeToProject) {
             return this;
         }
 
+        /**
+         * <p>
+         * Sets a collection of the NestedAttributeNames to be retrieved from the database. These attributes can include
```
bmaizels, line 365: Change 'sets' to 'adds'

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/model/QueryEnhancedRequest.java
```diff
@@ -347,7 +360,101 @@ public Builder addAttributeToProject(String attributeToProject) {
...
+                if (nestedAttributeNames != null) {
+                    this.nestedAttributesToProject.addAll(nestedAttributeNames);
+                }
+            }
+            return this;
+        }
+
+        /**
+         * <p>
+         * Sets one or more attribute names to be retrieved from the database. These attributes can include
```
bmaizels, line 401: Copy changes I recommended for the other similar method.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/model/QueryEnhancedRequest.java
```diff
@@ -347,7 +360,101 @@ public Builder addAttributeToProject(String attributeToProject) {
...
+        /**
+         * <p>
+         * Sets a collection of the NestedAttributeNames to be retrieved from the database. These attributes can include
+         * scalars, sets, or elements of a JSON document.
+         * This method takes arguments in form of NestedAttributeName which supports representing nested attributes.
+         * The NestedAttributeNames is specially created for projecting Nested Attribute names.
+         * The DOT characters are not recognized as nesting separator by DDB thus for Emnhance request NestedAttributeNames
+         * should be created to project Nested Attribute name at various levels.
+         * This Builder adds NestedAttributes to existing NestedAttributes
+         * or create a new list if NestedAttributes is not created.
```
bmaizels, line 372: ```This method will add new attributes to project to the existing list of attributes to project stored by this builder.```

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/model/QueryEnhancedRequest.java
```diff
@@ -347,7 +360,101 @@ public Builder addAttributeToProject(String attributeToProject) {
...
+         * </p>
+         * <p>
+         * If no attribute names are specified, then all attributes will be returned. If any of the requested attributes
+         * are not found, they will not appear in the result.
```
bmaizels, line 376: I think this is confusing to have here.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/model/ScanEnhancedRequest.java
```diff
@@ -284,7 +301,102 @@ public Builder addAttributeToProject(String attributeToProject) {
             return this;
         }
 
+        /**
+         * <p>
+         * Sets a collection of the NestedAttributeNames to be retrieved from the database. These attributes can include
```
bmaizels, line 306: Please implement all changes I suggested in QueryEnhancedRequest

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/ProjectionExpressionConvertor.java
```diff
@@ -0,0 +1,152 @@
...
+     */
+    public static ProjectionExpressionConvertor create(List<NestedAttributeName> nestedAttributeNames) {
+        return new ProjectionExpressionConvertor(nestedAttributeNames);
+    }
+
+    /**
+     * @param nestedAttributeName Individual NestedAttributeName
+     * @return Instance of ProjectionExpressionConvertor.
+     */
+    public static ProjectionExpressionConvertor create(NestedAttributeName nestedAttributeName) {
```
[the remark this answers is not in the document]
joviegas: removed

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/model/QueryEnhancedRequest.java
```diff
@@ -192,7 +206,8 @@ public int hashCode() {
         private Integer limit;
         private Boolean consistentRead;
         private Expression filterExpression;
-        private List<String> attributesToProject;
+        private List<String> attributesToProjectStringList;
```
joviegas, line 209: attributesToProjectStringList attribute is required to maintain the order of the older attributeToProject in cases of resets.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/model/ScanEnhancedRequest.java
```diff
@@ -153,7 +171,8 @@ public int hashCode() {
         private Integer limit;
         private Boolean consistentRead;
         private Expression filterExpression;
-        private List<String> attributesToProject;
+        private List<String> attributesToProjectStringList;
```
joviegas, line 174: attributesToProjectStringList attribute is required to maintain the order of the older attributeToProject in cases of resets.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/NestedAttributeName.java
```diff
@@ -0,0 +1,232 @@
...
+import software.amazon.awssdk.annotations.SdkPublicApi;
+
+/**
+ * High-level representation of a DynamoDB 'NestedAttributeName' that can be used in various situations where the API requires
+ * or accepts an Nested Attribute Name.
+ * Simple Attribute Name can be represented by passing just the name of the attribute.
+ * Nested Attributes are represented by List of String where each index of list corresponds to Nesting level Names.
+ * <p> While using attributeToProject in {@link software.amazon.awssdk.enhanced.dynamodb.model.QueryEnhancedRequest}
+ * and {@link software.amazon.awssdk.enhanced.dynamodb.model.ScanEnhancedRequest} we need  way to represent Nested Attributes.
+ * The normal DOT(.) separator is not recognized as a Nesting level separator by BynamoDB request,
```
bmaizels, line 31: Typo in DynamoDB

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/NestedAttributeName.java
```diff
@@ -0,0 +1,232 @@
...
+ * respresenting nested attribute name "foo.bar".
+ */
+@SdkPublicApi
+public final class NestedAttributeName {
+
+    private final List<String> elements;
+
+    private NestedAttributeName(List<String> nestedAttributeNames) {
+        this.elements = nestedAttributeNames != null
+                ? Collections.unmodifiableList(nestedAttributeNames) : null;
```
bmaizels, line 46: Having null elements should be invalid. Can we add a validator here?

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/model/QueryEnhancedRequest.java
```diff
@@ -125,8 +128,22 @@ public Expression filterExpression() {
...
     /**
      * Returns the list of projected attributes on this request object, or an null if no projection is specified.
+     * The Nested Attributes are represented using DOT separator.
+     * Example : foo.bar is represented as "foo.bar" which is indistinguishable from a non-nested attribute
+     * with the name "foo.bar".
+     * Use {@link #nestedAttributesToProject} if you have a use-case that requires discrimination between these two cases.
      */
     public List<String> attributesToProject() {
+        return attributesToProject != null ? attributesToProject.stream().filter(Objects::nonNull)
+                .map(item -> String.join(".", item.elements())).collect(Collectors.toList()) : null;
```
bmaizels, line 138: I feel like we should add validation to the constructor of the class to avoid the possibility of having null elements in attributesToProject then there would be no need to filter them here.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/model/QueryEnhancedRequest.java
bmaizels: Is the order important? Conceptually I think of this as a 'set'.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/model/ScanEnhancedRequest.java
```diff
@@ -231,12 +251,22 @@ public Builder filterExpression(Expression filterExpression) {
...
          * @return Returns a reference to this object so that method calls can be chained together.
          */
         public Builder attributesToProject(Collection<String> attributesToProject) {
-            this.attributesToProject = attributesToProject != null ? new ArrayList<>(attributesToProject) : null;
+
+            //If there is a reset then clear the attributesToProject List.
+            if (attributesToProjectStringList != null) {
+                attributesToProjectStringList
+                        .forEach(attribute -> this.attributesToProject.remove(NestedAttributeName.create(attribute)));
+            }
```
bmaizels, line 264: I think this is really confusing. We don't want to treat these as separate lists. We should make it clear in the javadoc that addNestedAttributeToProject is adding a 'nested attribute' to the 'attributesToProject' list. Let's make it clear to the customer there is just one list, and let's treat it like one list.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/model/QueryEnhancedRequest.java
```diff
@@ -294,12 +310,21 @@ public Builder filterExpression(Expression filterExpression) {
...
+         * @param attributesToProject A collection of the attributes names to be retrieved from the database.
          * @return Returns a reference to this object so that method calls can be chained together.
          */
         public Builder attributesToProject(Collection<String> attributesToProject) {
-            this.attributesToProject = attributesToProject != null ? new ArrayList<>(attributesToProject) : null;
+            //If there is a reset then clear the attributesToProject List.
+            if (attributesToProjectStringList != null) {
+                attributesToProjectStringList
+                        .forEach(attribute -> this.attributesToProject.remove(NestedAttributeName.create(attribute)));
+            }
```
bmaizels, line 322: I think this is really confusing. We don't want to treat these as separate lists. We should make it clear in the javadoc that addNestedAttributeToProject is adding a 'nested attribute' to the 'attributesToProject' list. Let's make it clear to the customer there is just one list, and let's treat it like one list.

services-custom/dynamodb-enhanced/src/test/java/software/amazon/awssdk/enhanced/dynamodb/ProjectionExpressionConvertorTest.java
```diff
@@ -0,0 +1,74 @@
...
+    public void testAttributeNameWithNoNestedAttributes() {
+        final String keyName = "fieldKey";
+        NestedAttributeName attributeName = NestedAttributeName.builder().elements(keyName).build();
+        ProjectionExpressionConvertor expressionConvertor = ProjectionExpressionConvertor.create(Arrays.asList(attributeName));
+        final Map<String, String> stringStringMap = expressionConvertor.convertToExpressionMap();
+        final Optional<String> toNameExpression = expressionConvertor.convertToProjectionExpression();
+        Map<String, String> expectedmap = new HashMap<>();
+        expectedmap.put(MAPPED_INDICATOR + keyName, keyName);
+        Assert.assertEquals(expectedmap, stringStringMap);
+        Assert.assertEquals(MAPPED_INDICATOR + keyName, toNameExpression.get());
```
bmaizels, line 28: Our assertions libraries in order of preference are : assertJ, hamcrest, junit. You picked the least preferential, can we use assertJ here and everywhere else you have created new test files.

services-custom/dynamodb-enhanced/src/test/java/software/amazon/awssdk/enhanced/dynamodb/ProjectionExpressionConvertorTest.java
```diff
@@ -0,0 +1,74 @@
...
+        Map<String, String> expectedmap = new HashMap<>();
+        expectedmap.put(MAPPED_INDICATOR + keyName, keyName);
+        expectedmap.put(MAPPED_INDICATOR + nestedAttribute, nestedAttribute);
+        Assert.assertEquals(expectedmap, stringStringMap);
+        Assert.assertEquals(MAPPED_INDICATOR + keyName + NESTING_SEPARATOR + MAPPED_INDICATOR + nestedAttribute, toNameExpression.get());
+    }
+
+    @Test
+    public void testAttributeNameWithNullAttributeName() {
+        NestedAttributeName attributeName = NestedAttributeName.builder().addElement(null).build();
```
bmaizels, line 48: This should be invalid and throw an exception (on NestedAttributeName).

services-custom/dynamodb-enhanced/src/test/java/software/amazon/awssdk/enhanced/dynamodb/ProjectionExpressionConvertorTest.java
```diff
@@ -0,0 +1,74 @@
...
+        final Optional<String> toNameExpression = expressionConvertor.convertToProjectionExpression();
+        assertThat(stringStringMap, is(new HashMap<>()));
+        Assert.assertFalse(toNameExpression.isPresent());
+    }
+
+    @Test
+    public void testAttributeNameWithNullNestedNestedAttributes() {
+        NestedAttributeName attributeName = NestedAttributeName.builder().build();
+        NestedAttributeName nullTest =  NestedAttributeName.builder().build();
+        ProjectionExpressionConvertor expressionConvertor = ProjectionExpressionConvertor.create(Arrays.asList(attributeName, nullTest));
```
bmaizels, line 61: This should be invalid and throw an exception (on NestedAttributeName).

services-custom/dynamodb-enhanced/src/test/java/software/amazon/awssdk/enhanced/dynamodb/functionaltests/BasicQueryTest.java
```diff
@@ -121,26 +117,50 @@ public int hashCode() {
...
 
     private DynamoDbTable<Record> mappedTable = enhancedClient.table(getConcreteTableName("table-name"), TABLE_SCHEMA);
 
+    private DynamoDbTable<NestedTestRecord> mappedNestedTable = enhancedClient.table(getConcreteTableName("nested-table-name"),
+            TableSchema.fromClass(NestedTestRecord.class));
+
     private void insertRecords() {
         RECORDS.forEach(record -> mappedTable.putItem(r -> r.item(record)));
+        NESTED_TEST_RECORDS.forEach(nestedTestRecord -> mappedNestedTable.putItem(r -> r.item(nestedTestRecord)));
     }
```
bmaizels, line 147: Maybe split this up, since our tests shouldn't be doing both.

services-custom/dynamodb-enhanced/src/test/java/software/amazon/awssdk/enhanced/dynamodb/model/QueryEnhancedRequestTest.java
```diff
@@ -89,6 +90,48 @@ public void builder_maximal() {
         assertThat(builtObject.attributesToProject(), is(attributesToProject));
     }
 
+    @Test
+    public void test_nestedAttributesToProject(){
+
```
bmaizels, line 95: Can we split these into individual tests.

services-custom/dynamodb-enhanced/src/test/java/software/amazon/awssdk/enhanced/dynamodb/model/QueryEnhancedRequestTest.java
```diff
@@ -89,6 +90,48 @@ public void builder_maximal() {
...
+        assertThat(builtObjectOne.attributesToProject(), is(attributesToProjectNestedLast));
+
+
+        QueryEnhancedRequest builtObjectTwo = QueryEnhancedRequest.builder()
+                .attributesToProject(attributesToProjectArray)
+                .addAttributeToProject(additionalElement)
+                .addNestedAttributesToProject(NestedAttributeName.create("foo", "bar"))
+                .attributesToProject(additionalElement)
+                .build();
+        List<String> attributesToProjectStringAttributeOverwrittenNestedLast = Arrays.asList( "foo.bar", "three");
```
bmaizels, line 122: See my other comment in the code, I think the 'correct' state here should be "three", and the nested attribute should be overwritten with the others.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/NestedAttributeName.java
joviegas: Added for Null as well as Empty list

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/model/QueryEnhancedRequest.java
joviegas: `QueryEnhancedRequest.builder(). attributesToProject("One", "two ", "three"). addNestedAttributesToProject("foo", "bar").attributesToProject("bazz");`
  The expected output for this should be
getAttributesToProject --> [foo.bar, bazz]
Thus I have kept a Temp List on the Builder called attributesToProjectStringList where you had commented that order is not required. I will rename to attributesToProjectTempList;

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/model/QueryEnhancedRequest.java
joviegas: I will rename attributesToProjectStringList to ** attributesToProjectTempList** , the intention of this list is to keep the backward compatibility , earlier if the user called attributesToProject() twice it used to reset the old values with new one we need to keep this behavior.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/model/ScanEnhancedRequest.java
joviegas: I will rename attributesToProjectStringList to ** attributesToProjectTempList** , the intention of this list is to keep the backward compatibility , earlier if the user called attributesToProject() twice it used to reset the old values with new one we need to keep this behavior.

services-custom/dynamodb-enhanced/src/test/java/software/amazon/awssdk/enhanced/dynamodb/ProjectionExpressionConvertorTest.java
joviegas: updated

bmaizels (approved):

PR #2036, opened 2020-09-14
Add support for generating waiter acceptors.

millems:
This required the following changes:
1. Added a JMES Path interpreter based on the JMES Path parser from a previous commit. This interpreter is in the code generator and creates the Java instructions for executing the acceptor logic.
2. Added the code generator logic for specifying the acceptors.
3. Added a Waiters runtime that is currently shared between all service clients, used by the acceptors of all services.
4. Added a "MemberName" field to the SDK field metadata, so that it can be referenced by the waiter code (waiters refer to the member names from the C2J).
5. Added tests for the two most complex waiter classes: AutoScaling and ECS
6. Added support for acceptors matching supertypes of the response type.

Waiters should mostly be working now, but there are a few remaining tasks:
1. There's a WaitersRuntime component in sdk-core that should be copied to each of the service to remove them as a protected API.
2. Waiters currently only support the JMESPath syntax currently used by services. We should add support for some of the other 'easier' JMESPath syntax so that we're future-compatible with more waiters.


codegen/src/main/java/software/amazon/awssdk/codegen/poet/waiters/BaseWaiterClassSpec.java
```diff
@@ -294,4 +328,143 @@ private OperationModel operationModel(WaiterDefinition waiterDefinition) {
...
+                      .add(pathAcceptorBody(acceptor))
+                      .add(")");
+                break;
+            case "pathAll":
+                result.add("OnResponseAcceptor(")
+                      .add(pathAllAcceptorBody(acceptor))
+                      .add(")");
+                break;
+            case "pathAny":
+                result.add("OnResponseAcceptor(")
```
dagnir, line 374: Can this be move out of the switch? Looks like we always add except for `status` which doesn't use the `result` that's being built

codegen/src/main/java/software/amazon/awssdk/codegen/poet/waiters/BaseWaiterClassSpec.java
```diff
@@ -294,4 +328,143 @@ private OperationModel operationModel(WaiterDefinition waiterDefinition) {
...
+
+    private CodeBlock pathAcceptorBody(Acceptor acceptor) {
+        String expected = acceptor.getExpected().asText();
+        String expectedType = acceptor.getExpected() instanceof JrsString ? "$S" : "$L";
+        return CodeBlock.builder()
+                        .add("response -> {")
+                        .add("$1T input = new $1T(response);", WaitersRuntime.Value.class)
+                        .add("return $T.equals(", Objects.class)
+                        .add(JmesPathInterpreter.interpret(acceptor.getArgument(), "input"))
+                        .add(".value(), " + expectedType + ");", expected)
```
dagnir, line 416: minor: use `addStatement` instead of adding `;` manually

codegen/src/main/java/software/amazon/awssdk/codegen/poet/waiters/BaseWaiterClassSpec.java
```diff
@@ -294,4 +328,143 @@ private OperationModel operationModel(WaiterDefinition waiterDefinition) {
...
+            case "retry":
+                return WaiterState.RETRY.name();
+            default:
+                throw new IllegalArgumentException("Unsupported acceptor state: " + acceptor.getState());
+        }
+    }
+
+    private CodeBlock pathAcceptorBody(Acceptor acceptor) {
+        String expected = acceptor.getExpected().asText();
+        String expectedType = acceptor.getExpected() instanceof JrsString ? "$S" : "$L";
```
dagnir, line 410: minor: extract to helper method?

codegen/src/main/java/software/amazon/awssdk/codegen/poet/waiters/JmesPathInterpreter.java
```diff
@@ -0,0 +1,306 @@
...
+import software.amazon.awssdk.codegen.jmespath.parser.JmesPathParser;
+import software.amazon.awssdk.codegen.jmespath.parser.JmesPathVisitor;
+import software.amazon.awssdk.core.SdkPojo;
+import software.amazon.awssdk.core.waiters.WaitersRuntime;
+import software.amazon.awssdk.utils.Validate;
+
+/**
+ * A code interpreter for JMESPath expressions.
+ *
+ * This can convert a JMESPath expression into a statement that executes against an {@link SdkPojo}.
```
dagnir, line 56: Can we add more detail in this javadoc? I was confused at first thinking the statements are jut basic Java expressions, but looks like the output is an intermediate representation that is further interpreted by the Waiters Runtime.

codegen/src/test/java/software/amazon/awssdk/codegen/poet/waiters/JmesPathInterpreterTest.java
```diff
@@ -0,0 +1,53 @@
...
+ */
+
+package software.amazon.awssdk.codegen.poet.waiters;
+
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.junit.Test;
+
+public class JmesPathInterpreterTest {
```
dagnir, line 23: Not a block but these are quite hard to read in line. Would be nice to move them to text files that are easier to read, especially if we end up writing more test cases (like when we support more JMESPath features).

core/sdk-core/src/main/java/software/amazon/awssdk/core/waiters/WaitersRuntime.java
```diff
@@ -0,0 +1,523 @@
...
+import software.amazon.awssdk.core.SdkResponse;
+import software.amazon.awssdk.core.exception.SdkClientException;
+import software.amazon.awssdk.core.exception.SdkServiceException;
+import software.amazon.awssdk.utils.ToString;
+
+/**
+ * Contains classes used at runtime by the code generator classes for waiter acceptors generated from JMESPath expressions.
+ */
+@SdkProtectedApi
+public final class WaitersRuntime {
```
dagnir, line 38: Can we add tests for the runtime?

dagnir (commented):
Looks like JMESPath has a suite of acceptance tests. Can we integrate them into our implementation somehow?

https://github.com/jmespath/jmespath.test

codegen/src/main/java/software/amazon/awssdk/codegen/poet/waiters/BaseWaiterClassSpec.java
millems: +1

codegen/src/main/java/software/amazon/awssdk/codegen/poet/waiters/BaseWaiterClassSpec.java
millems: We're pretty inconsistent with this in general. 

I think on these 4 lines it would be a bit less readable, since the first line would be addStatement but the later ones wouldn't be (they're partial statements), so the code wouldn't line up nicely.

codegen/src/main/java/software/amazon/awssdk/codegen/poet/waiters/BaseWaiterClassSpec.java
millems: I feel like customizing the type used in poet is unusual, so putting it away in a helper method would make the reader always have to go to that method to see what it's doing. It seems easier to understand inlined with the method, even if it's less dry.

codegen/src/main/java/software/amazon/awssdk/codegen/poet/waiters/JmesPathInterpreter.java
millems: Will do! As some history, I tried to just emit java expressions, but it was way too difficult. The waiters runtime made it much easier.

codegen/src/test/java/software/amazon/awssdk/codegen/poet/waiters/JmesPathInterpreterTest.java
millems: Whoops, I meant do to that before publishing the review. Good catch.

core/sdk-core/src/main/java/software/amazon/awssdk/core/waiters/WaitersRuntime.java
millems: That's a good idea, but I'm wondering how that will jive when we move the runtime to being copied into each of the service clients. I was hoping testing it in the context it's used (i.e. testing the generated waiters and making sure they worked, which indirectly tests the runtime) would be sufficient to side-step this problem. Do you have any ideas on how to make that work?

millems (commented):
For the JMESPath acceptance tests, there's two issues:
1. We don't support all of JMESPath, so we'd have to pick out the tests that we do support and use those, or expand to support all of JMESPath.
2. Those tests are written to work against JSON, and our implementation works against generated SdkPojos. To bridge the gap, we'd have to build a test harness to map the two, as well as parse and execute the acceptance tests against the SdkPojo -> JSON mapping.

I think that would add another week or more to the task. While I don't mind us doing it, it might be a bit much to do as part of this code review.

core/sdk-core/src/main/java/software/amazon/awssdk/core/waiters/WaitersRuntime.java
dagnir: > Do you have any ideas on how to make that work?

I think if we just accessed a `WaitersRuntime` class from within `codegen-generated-classes-test` that could work.

dagnir:
> I think that would add another week or more to the task. While I don't mind us doing it, it might be a bit much to do as part of this code review.

Yep I think doing that as a separate task is fine; I imagine the middle layer would look something like our protocol test harness. We should integrate any tests that we can to give us extra confidence :).

core/sdk-core/src/main/java/software/amazon/awssdk/core/waiters/WaitersRuntime.java
millems: Cool, so generate it into each service as well? That could work...

codegen/src/main/java/software/amazon/awssdk/codegen/poet/waiters/BaseWaiterClassSpec.java
millems: Actually, the error path is different, so I can't do this.

dagnir (approved):
LGTM! Please remember to add a task in the backlog if we don't have one for integrating the JMESPath suite

PR #2039, opened 2020-09-16
Append User-Agent for Waiters operations

Quanzzzz:
Currently, there is no signs for Waiters in the ```User-Agent``` header, and to track the usage of waiters, we should append a specific user agent for waiters in the header.

## Description

The ```codegen``` module is modified, adding the method appending userAgent for waiters in all the operations related to the waiters.

## Testing

The generated dynamodb client is tested with the waiters. There are two WireMock tests verifying if the requests triggered by the waiters include the ```WAITERS``` user agent in them.

## Types of changes

## License

### DIFF
--- a/codegen/src/test/resources/software/amazon/awssdk/codegen/poet/waiters/query-sync-waiter-class.java
+++ b/codegen/src/test/resources/software/amazon/awssdk/codegen/poet/waiters/query-sync-waiter-class.java
@@ -4,10 +4,13 @@ import java.time.Duration;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
+import java.util.function.Consumer;
 import software.amazon.awssdk.annotations.Generated;
 import software.amazon.awssdk.annotations.SdkInternalApi;
 import software.amazon.awssdk.annotations.ThreadSafe;
+import software.amazon.awssdk.awscore.AwsRequestOverrideConfiguration;
 import software.amazon.awssdk.awscore.exception.AwsServiceException;
+import software.amazon.awssdk.core.ApiName;
 import software.amazon.awssdk.core.internal.waiters.WaiterAttribute;
 import software.amazon.awssdk.core.retry.backoff.FixedDelayBackoffStrategy;
 import software.amazon.awssdk.core.waiters.PollingStrategy;
@@ -18,6 +21,7 @@ import software.amazon.awssdk.core.waiters.WaiterState;
 import software.amazon.awssdk.services.query.QueryClient;
 import software.amazon.awssdk.services.query.model.APostOperationRequest;
 import software.amazon.awssdk.services.query.model.APostOperationResponse;
+import software.amazon.awssdk.services.query.model.QueryRequest;
 import software.amazon.awssdk.services.query.waiters.internal.WaitersRuntime;
 import software.amazon.awssdk.utils.AttributeMap;
 import software.amazon.awssdk.utils.SdkAutoCloseable;
@@ -59,7 +63,7 @@ final class DefaultQueryWaiter implements QueryWaiter {
 
     @Override
     public WaiterResponse<APostOperationResponse> waitUntilPostOperationSuccess(APostOperationRequest aPostOperationRequest) {
-        return postOperationSuccessWaiter.run(() -> client.aPostOperation(aPostOperationRequest));
+        return postOperationSuccessWaiter.run(() -> client.aPostOperation(applyWaitersUserAgent(aPostOperationRequest)));
     }
 
     private static List<WaiterAcceptor<? super APostOperationResponse>> postOperationSuccessWaiterAcceptors() {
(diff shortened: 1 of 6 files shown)


codegen/src/test/resources/software/amazon/awssdk/codegen/poet/waiters/query-sync-waiter-class.java
```diff
@@ -84,6 +89,15 @@ public void close() {
         return new DefaultBuilder();
     }
 
+    private <T extends QueryRequest> T applyWaitersUserAgent(T request) {
+        Consumer<AwsRequestOverrideConfiguration.Builder> userAgentApplier = b -> b.addApiName(ApiName.builder()
+                                                                                                      .version(VersionInfo.SDK_VERSION).name("WAITERS").build());
```
millems, line 94: In metric publisher we do

```
ApiName.builder().name("hll").version("cw-mp").build();
```

With the mapper we do 

```
ApiName.builder().version("ddb-enh").name("hll").build();
```

Waiters is kind of a high-level library. Should we do 

```
ApiName.builder().name("hll").version("waiter").build();
```

?

services/dynamodb/src/test/java/software/amazon/awssdk/services/dynamodb/WaitersUserAgentTest.java
```diff
@@ -0,0 +1,97 @@
...
+                                       .overrideConfiguration(c -> c.addExecutionInterceptor(interceptor))
+                                       .build();
+
+        dynamoDbAsyncClient = DynamoDbAsyncClient.builder()
+                                                 .credentialsProvider(StaticCredentialsProvider.create(AwsBasicCredentials
+                                                                                                           .create("test",
+                                                                                                                   "test")))
+                                                 .region(Region.US_WEST_2)
+                                                 .overrideConfiguration(c -> c.addExecutionInterceptor(interceptor))
+                                                 .build();
```
millems, line 69: Is this an integration test now?

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/waiters/query-sync-waiter-class.java
Quanzzzz: Yea sure, I can change that. I was imitating the `Paginator` user agent so it looked like this.

services/dynamodb/src/test/java/software/amazon/awssdk/services/dynamodb/WaitersUserAgentTest.java
Quanzzzz: No, these are just the clients with the interceptor. The integration tests are not here.

millems: It looks like these are calling the real DynamoDB, though, not the wiremock server.

Quanzzzz: Oh I got what you mean, yea I can add the wireMock server.

services/dynamodb/src/test/java/software/amazon/awssdk/services/dynamodb/WaitersUserAgentTest.java
```diff
@@ -0,0 +1,106 @@
...
+        stubFor(any(urlEqualTo("/")).willReturn(aResponse().withStatus(500)));
+
+        DynamoDbWaiter waiter = dynamoDbClient.waiter();
+        assertThatThrownBy(() -> waiter.waitUntilTableExists(DescribeTableRequest.builder().tableName("table").build())).isNotNull();
+
+        ArgumentCaptor<Context.BeforeTransmission> context = ArgumentCaptor.forClass(Context.BeforeTransmission.class);
+        Mockito.verify(interceptor, times(9)).beforeTransmission(context.capture(), Matchers.any());
+
+        assertTrue(context.getValue().httpRequest().headers().get("User-Agent").toString().contains("waiter"));
+    }
```
millems, line 90: How long do these tests take to run?

Quanzzzz: The sync one takes around 7s, while the async one only takes 98ms, because the sync one does wait for the response. For the async one, the CompletableFuture is cancelled once the user agent is checked.

zoewangg: Should we use mock clients instead? `codegen-generated-classes-test` seems like a good place for this test. https://github.com/aws/aws-sdk-java-v2/tree/waiters-development/test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/waiters

nit, you can just use wiremock's verify method rather than the interceptor
```
       verify(postRequestedFor(anyUrl()).withHeader("User-Agent", contains(...));
```

millems: You could also throw an exception from the interceptor if you want to cancel the calls altogether and just validate the request.

Quanzzzz: @zoewangg I just tried testing with the mock clients and Mockito.verify(), and it took around 8 mins to get the result, not sure what caused that but it seems the WireMock test is much more efficient.

Quanzzzz:
Test updated with adding an exception in the `beforeTransmission()` of the interceptor. Current testing time is around 1.5s for sync and 49ms for async.

millems (approved):

PR #2041, opened 2020-09-17
Parse XML errors like in S3

cenedhryn:
Changes the definition of the document error root to what is used in AWS S3, so that errors are fully parsed. 

## Motivation and Context
To this point, errors returned from S3Control have lost their XML content and only the error type has been returned, showing null: 
~~~
software.amazon.awssdk.services.s3control.model.S3ControlException: null
~~~
With this change, errors are parsed correctly as:
~~~
software.amazon.awssdk.services.s3control.model.S3ControlException: Missing role arn 
(Service: S3Control, Status Code: 400, Request ID: 4T5H9GEV7H3K9MCJ, Extended Request ID: USkN9APZoEVT0...)
~~~

## Testing
- Unit tests and functional tests

## Types of changes

## Checklist

## License

### DIFF
--- a/.changes/next-release/bugfix-AWSS3Control-1ad7752.json
+++ b/.changes/next-release/bugfix-AWSS3Control-1ad7752.json
@@ -1,0 +1,5 @@
+{
+    "category": "AWS S3 Control", 
+    "type": "bugfix", 
+    "description": "Changes document error root definition for XML based errors to same as AWS S3, so that errors are fully parsed."
+}
--- a/services/s3control/pom.xml
+++ b/services/s3control/pom.xml
@@ -78,5 +78,10 @@
             <version>${awsjavasdk.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>com.github.tomakehurst</groupId>
+            <artifactId>wiremock</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 </project>
--- a/services/s3control/src/main/resources/codegen-resources/customization.config
+++ b/services/s3control/src/main/resources/codegen-resources/customization.config
@@ -3,5 +3,6 @@
     "customResponseMetadata": {
         "EXTENDED_REQUEST_ID": "x-amz-id-2",
         "REQUEST_ID": "x-amz-request-id"
-    }
+    },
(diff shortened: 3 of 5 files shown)


services/s3control/src/main/resources/log4j.properties
```diff
@@ -0,0 +1,33 @@
+#
```
dagnir, line 1: We shouldn't add `log4j.properties` in the main resources

services/s3control/src/main/java/software/amazon/awssdk/services/s3control/internal/interceptors/EndpointAddressInterceptor.java
```diff
@@ -81,7 +86,7 @@ private String resolveHost(SdkHttpRequest request, String accountId, S3ControlCo
 
         }
         validateComponentIsHostnameCompliant(accountId, "account id");
-        return String.format("%s.%s", accountId, host);
+        return hasEndpointOverride ? host : String.format("%s.%s", accountId, host);
```
dagnir, line 89: Thinking about this some more, I'm not sure that this is the right behavior; it's possible that the user could still want this behavior even with an endpoint override.

Can we leave this change out of the client for now? For testing, we can we can just add a new client interceptor that changes the request URL to the WireMock host before it's sent.

cenedhryn: Yes, that sounds like a less intrusive change.

services/s3control/src/main/resources/log4j.properties
cenedhryn: I thought I added it in the test directory, will move it.

services/s3control/src/main/java/software/amazon/awssdk/services/s3control/internal/interceptors/EndpointAddressInterceptor.java
dagnir, line 89: New changes look good, but looks like this still needs to be reverted along with the accompanying test.

dagnir (approved):
:shipit:

Issue #1157, opened 2019-03-21
RDS IAM Authentication Token Generator


pula2000:
I'm trying to use V2 SDK for RDS to establish a JDBC connection to an RDS instance using IAM DB authentication. But the example in the tutorial uses the auth package in the V1 sdk. I don't see an equivalent package in the V2 RDS module. Would be great if someone could point me to the right place in V2 for doing the same. Thanks

varunnvs92:
V2 auth module link: https://github.com/aws/aws-sdk-java-v2/tree/master/core/auth
Can you give a link to the tutorial?

zoewangg:
Closing this due to inactivity. Feel free to reopen if you have further questions.

abatkin:
@zoewangg This issue is clearly not solved (and auto-closing after 12 days isn't very friendly) - can you reopen and assign to the right place please?

In order to connect to an RDS instance that is using IAM auth, you need to generate an auth token (the cli provides `aws rds generate-db-auth-token` for this purpose). AWS provides [documentation](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.IAMDBAuth.Connecting.Java.html) for the v1 Java SDK, but I can't find any corresponding documentation for the v2 Java SDK.

Actually I don't see any parallel classes in the v2 Java SDK, which means that either they are hidden/undocumented somewhere (hence this ticket!) or else it isn't implemented (in which case we need a different ticket to ask for that to be implemented).

At this point the only option appears to be translating the v1 docs for manually generating a token ("Manually Constructing an IAM Authentication Token" in the docs linked above) which is very cumbersome.

zoewangg:
As a general rule, we try to close an issue if there is no response from customers after 7 days to make sure we are on top of all issues; the issue list will grow uncontrollably if we leave all unresponsive issues open. You are always welcome to create new issues or comment on any issues. That said, thank you for your feedback and we might consider increasing auto-closing period.

Now that we have more information on this issue, we will take another look.

forficate:
Adding my +1 vote.

This is a crucial piece of functionality missing from the new SDK.

millems:
[...]

PR #2057, opened 2020-09-22
Add a RdsUtilities with the ability to generate an IAM auth token

abrooksv:
## Description
Add a new Rds Utilities class that can generate the auth token.
V1 Parity feature
More info: https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.IAMDBAuth.Connecting.AWSCLI.html

Utilities pattern is used instead of Presigner due to this is a pseudo request. There does not exist a modeled API call so it does not make sense to make a presigner for it.

## Motivation and Context
Upstream this code out of the AWS Toolkit for JetBrains
#1157

## Testing
Need to write the tests

## Types of changes

## Checklist

## License

### DIFF
--- a/services/rds/src/main/java/software/amazon/awssdk/services/rds/DefaultRdsUtilities.java
+++ b/services/rds/src/main/java/software/amazon/awssdk/services/rds/DefaultRdsUtilities.java
@@ -1,0 +1,156 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.services.rds;
+
+import java.time.Clock;
+import java.time.Duration;
+import java.time.Instant;
+import software.amazon.awssdk.annotations.Immutable;
+import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider;
+import software.amazon.awssdk.auth.signer.Aws4Signer;
+import software.amazon.awssdk.auth.signer.params.Aws4PresignerParams;
+import software.amazon.awssdk.awscore.client.config.AwsClientOption;
+import software.amazon.awssdk.core.client.config.SdkClientConfiguration;
+import software.amazon.awssdk.http.SdkHttpFullRequest;
+import software.amazon.awssdk.http.SdkHttpMethod;
+import software.amazon.awssdk.regions.Region;
+import software.amazon.awssdk.services.rds.model.GenerateAuthenticationTokenRequest;
+import software.amazon.awssdk.utils.StringUtils;
+
+@Immutable
+@SdkInternalApi
+final class DefaultRdsUtilities implements RdsUtilities {
+    // The time the IAM token is good for. https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.IAMDBAuth.html
+    private static final Duration EXPIRATION_DURATION = Duration.ofMinutes(15);
+
+    private final Aws4Signer signer = Aws4Signer.create();
+    private final Region region;
+    private final AwsCredentialsProvider credentialsProvider;
+    private final Clock clock;
+
+    DefaultRdsUtilities(DefaultBuilder builder) {
+        this(builder, Clock.systemUTC());
+    }
+
+    /**
+     * Test Only
+     */
+    DefaultRdsUtilities(DefaultBuilder builder, Clock clock) {
+        this.credentialsProvider = builder.credentialsProvider;
+        this.region = builder.region;
+        this.clock = clock;
+    }
+
+    /**
+     * Used by RDS low-level client's utilities() method
+     */
+    @SdkInternalApi
+    static RdsUtilities create(SdkClientConfiguration clientConfiguration) {
+        return new DefaultBuilder().clientConfiguration(clientConfiguration).build();
+    }
+
+    @Override
+    public String generateAuthenticationToken(GenerateAuthenticationTokenRequest request) {
+        SdkHttpFullRequest httpRequest = SdkHttpFullRequest.builder()
+                                            .method(SdkHttpMethod.GET)
+                                            .protocol("https")
+                                            .host(request.hostname())
+                                            .port(request.port())
+                                            .encodedPath("/")
+                                            .putRawQueryParameter("DBUser", request.username())
+                                            .putRawQueryParameter("Action", "connect")
+                                            .build();
+
+        Instant expirationTime = Instant.now(clock).plus(EXPIRATION_DURATION);
+        Aws4PresignerParams presignRequest = Aws4PresignerParams.builder()
+                                                .signingClockOverride(clock)
+                                                .expirationTime(expirationTime)
+                                                .awsCredentials(resolveCredentials(request).resolveCredentials())
+                                                .signingName("rds-db")
+                                                .signingRegion(resolveRegion(request))
+                                                .build();
+
+        SdkHttpFullRequest fullRequest = signer.presign(httpRequest, presignRequest);
+        String signedUrl = fullRequest.getUri().toString();
+
+        // Format should be: <hostname>>:<port>>/?Action=connect&DBUser=<username>>&X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Expi...
+        // Note: This must be the real RDS hostname, not proxy or tunnels
+        return StringUtils.replacePrefixIgnoreCase(signedUrl, "https://", "");
+    }
+
+    private Region resolveRegion(GenerateAuthenticationTokenRequest request) {
+        if (request.region() != null) {
+            return request.region();
+        }
+
+        if (this.region != null) {
+            return this.region;
+        }
+
+        throw new IllegalArgumentException("Region should be provided either in GenerateAuthenticationTokenRequest object " +
+                "or RdsUtilities object");
+    }
+
+    private AwsCredentialsProvider resolveCredentials(GenerateAuthenticationTokenRequest request) {
+        if (request.credentialsProvider() != null) {
+            return request.credentialsProvider();
+        }
+
+        if (this.credentialsProvider != null) {
+            return this.credentialsProvider;
+        }
+
+        throw new IllegalArgumentException("CredentialProvider should be provided either in GenerateAuthenticationTokenRequest " +
+                "object or RdsUtilities object");
+    }
+
+    @SdkInternalApi
+    static final class DefaultBuilder implements Builder {
+        private Region region;
+        private AwsCredentialsProvider credentialsProvider;
+
+        DefaultBuilder() {
+        }
+
+        Builder clientConfiguration(SdkClientConfiguration clientConfiguration) {
+            this.credentialsProvider = clientConfiguration.option(AwsClientOption.CREDENTIALS_PROVIDER);
+            this.region = clientConfiguration.option(AwsClientOption.AWS_REGION);
+
+            return this;
+        }
+
+        @Override
+        public Builder region(Region region) {
+            this.region = region;
+            return this;
+        }
+
+        @Override
+        public Builder credentialsProvider(AwsCredentialsProvider credentialsProvider) {
+            this.credentialsProvider = credentialsProvider;
+            return this;
+        }
+
+        /**
+         * Construct a {@link RdsUtilities} object.
+         */
+        @Override
+        public RdsUtilities build() {
+            return new DefaultRdsUtilities(this);
+        }
+    }
+}
--- a/services/rds/src/main/java/software/amazon/awssdk/services/rds/RdsUtilities.java
+++ b/services/rds/src/main/java/software/amazon/awssdk/services/rds/RdsUtilities.java
@@ -1,0 +1,106 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.services.rds;
+
+import java.util.function.Consumer;
+import software.amazon.awssdk.annotations.SdkPublicApi;
+import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider;
+import software.amazon.awssdk.regions.Region;
+import software.amazon.awssdk.services.rds.model.GenerateAuthenticationTokenRequest;
+
+/**
+ * Utilities for working with RDS. An instance of this class can be created by:
+ * <p>
+ * 1) Using the low-level client {@link RdsClient#utilities()} (or {@link RdsAsyncClient#utilities()}} method. This is
+ * recommended as SDK will use the same configuration from the {@link RdsClient} object to create the {@link RdsUtilities} object.
+ *
+ * <pre>
+ * RdsClient rdsClient = RdsClient.create();
+ * RdsUtilities utilities = rdsClient.utilities();
+ * </pre>
+ * </p>
+ *
+ * <p>
+ * 2) Directly using the {@link #builder()} method.
+ *
+ * <pre>
+ * RdsUtilities utilities = RdsUtilities.builder()
+ *  .credentialsProvider(DefaultCredentialsProvider.create())
+ *  .region(Region.US_WEST_2)
+ *  .build()
+ * </pre>
+ * </p>
+ *
+ * Note: This class does not make network calls.
+ */
+@SdkPublicApi
+public interface RdsUtilities {
+    /**
+     * Create a builder that can be used to configure and create a {@link RdsUtilities}.
+     */
+    static Builder builder() {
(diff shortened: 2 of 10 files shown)


abrooksv:
This was based on a conversation with Matt. Looking for feedback before writing tests and finishing testing

services/rds/src/main/java/software/amazon/awssdk/services/rds/RdsUtilities.java
```diff
@@ -0,0 +1,139 @@
...
+                                            .method(SdkHttpMethod.GET)
+                                            .protocol("https")
+                                            .host(request.hostname())
+                                            .port(request.port())
+                                            .encodedPath("/")
+                                            .putRawQueryParameter("DBUser", request.userName())
+                                            .putRawQueryParameter("Action", "connect")
+                                            .build();
+
+        Instant expirationTime = Instant.now().plus(15, ChronoUnit.MINUTES);
```
hunterwerlla, line 76: This probably needs a link to the doc where it states this must be 15 min

abrooksv: Ya, our doc pages all say 15 mins: https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.IAMDBAuth.html

abrooksv:
Ping. Still looking for feedback on direction before committing any more time to this.

services/rds/src/main/java/software/amazon/awssdk/services/rds/model/GenerateAuthenticationTokenRequest.java
```diff
@@ -0,0 +1,140 @@
...
+ */
+public final class GenerateAuthenticationTokenRequest implements ToCopyableBuilder<GenerateAuthenticationTokenRequest.Builder, GenerateAuthenticationTokenRequest> {
+    private final String hostname;
+    private final int port;
+    private final String username;
+    private final Region region;
+    private final AwsCredentialsProvider credentialsProvider;
+
+    public GenerateAuthenticationTokenRequest(BuilderImpl builder) {
+        this.hostname = Validate.notEmpty(builder.hostname, "hostname");
```
zoewangg, line 37: We probably should validate hostname to make sure it's compliant. https://github.com/aws/aws-sdk-java-v2/blob/master/utils/src/main/java/software/amazon/awssdk/utils/HostnameValidator.java#L23

services/rds/src/main/java/software/amazon/awssdk/services/rds/RdsUtilities.java
```diff
@@ -0,0 +1,139 @@
...
+    /**
+     * Builder class to construct {@link RdsUtilities} object
+     */
+    public static final class Builder {
+        private SdkClientConfiguration clientConfiguration;
+
+        private Builder() {
+        }
+
+        private Builder clientConfiguration(SdkClientConfiguration clientConfiguration) {
```
zoewangg, line 127: `SdkClientConfiguration` is a protected API, so not sure if we want to use it here
It seems we only need region and credential provider here

services/rds/src/main/java/software/amazon/awssdk/services/rds/RdsUtilities.java
```diff
@@ -0,0 +1,139 @@
...
+import software.amazon.awssdk.core.client.config.SdkClientConfiguration;
+import software.amazon.awssdk.http.SdkHttpFullRequest;
+import software.amazon.awssdk.http.SdkHttpMethod;
+import software.amazon.awssdk.regions.Region;
+import software.amazon.awssdk.services.rds.model.GenerateAuthenticationTokenRequest;
+import software.amazon.awssdk.utils.StringUtils;
+
+@Immutable
+@SdkPublicApi
+public final class RdsUtilities {
```
zoewangg, line 37: I understand the existing `S3Utitlies` doesn't have an interface, but seems a good idea to create one for new utility classes.

zoewangg (commented):
Looks good overall.

services/rds/src/main/java/software/amazon/awssdk/services/rds/RdsUtilities.java
abrooksv: To be clear, not a base interface for all utils, just the methods in this class, right?

abrooksv:
Sorry, re:invent kept me from getting back to this Zoe. Thank you for the initial review. Will work on marching this forward

services/rds/src/main/java/software/amazon/awssdk/services/rds/model/GenerateAuthenticationTokenRequest.java
abrooksv: Will change

services/rds/src/main/java/software/amazon/awssdk/services/rds/RdsUtilities.java
abrooksv: Was worried about future compatibility, but since this is private we can worry about that if/when that time comes

services/rds/src/main/java/software/amazon/awssdk/services/rds/RdsUtilities.java
zoewangg: Yup

abrooksv:
@zoewangg Should we stick region/credProvider in:

1. the RdsUtiltiies as fields
2. The request GenerateAuthenticationTokenRequest
3. Both and request takes precedence?

S3 seems to do both 3

services/rds/src/main/java/software/amazon/awssdk/services/rds/RdsUtilities.java
abrooksv: Builder now stores the cred provider and region as fields. This method is used by the package protected method `RdsUtilities#create` which is generated to be used from `RdsClient#utiltiies()`

DefaultRdsClient.java
```
    @Override
    public RdsUtilities utilities() {
        return RdsUtilities.create(clientConfiguration);
    }
```

So everything is now hidden which should be hidden.

zoewangg:
@abrooksv I'd vote # 3 to be consistent with s3 utilities.

abrooksv:
Okay, new revision pushed:

1. Interfaces are in place. Went back and forth on how to lay them out.
SDK Clients make them separate files, but things like Apache builders and request/response builders use inner classes. 

Let me know if you want them broken out into 4 files, happy to do it. Or w/e layout you feel is best.

2. CodeGen was modified to allow interfaces for utility classes, new field is `instanceType`
3. First pass of the Javadoc written.

Need to figure out how we should test this. V1 has no tests for its version of the class 😢 

Standard pre-signer methods have access to `AwsRequestOverrideConfiguration`, but this does not. Should it?  (If so, maybe actually SdkRequestOverrideConfiguration since the request already has credentials?)

Should we have a test only constructor on DefaultRdsUtilities to takes a signer?

services/rds/src/main/java/software/amazon/awssdk/services/rds/DefaultRdsUtilities.java
```diff
@@ -0,0 +1,141 @@
...
+                                            .method(SdkHttpMethod.GET)
+                                            .protocol("https")
+                                            .host(request.hostname())
+                                            .port(request.port())
+                                            .encodedPath("/")
+                                            .putRawQueryParameter("DBUser", request.userName())
+                                            .putRawQueryParameter("Action", "connect")
+                                            .build();
+
+        Instant expirationTime = Instant.now().plus(15, ChronoUnit.MINUTES);
```
zoewangg, line 64: nit, can we make it a constant and add a comment with the [link](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.IAMDBAuth.html ) there so that it's less of a magic number?

services/rds/src/main/java/software/amazon/awssdk/services/rds/DefaultRdsUtilities.java
```diff
@@ -0,0 +1,141 @@
...
+        Aws4PresignerParams presignRequest = Aws4PresignerParams.builder()
+                                                .expirationTime(expirationTime)
+                                                .awsCredentials(resolveCredentials(request).resolveCredentials())
+                                                .signingName("rds-db")
+                                                .signingRegion(resolveRegion(request))
+                                                .build();
+
+        SdkHttpFullRequest fullRequest = Aws4Signer.create().presign(httpRequest, presignRequest);
```
zoewangg, line 72: Seems like we can reuse the signer.

zoewangg (commented):
WRT testing, I don't think it should have access to `AwsRequestOverrideConfiguration` or `SdkRequestOverrideConfiguration`. A test ctor with a test singer injected sounds good to me.

services/rds/src/main/java/software/amazon/awssdk/services/rds/model/GenerateAuthenticationTokenRequest.java
```diff
@@ -0,0 +1,173 @@
...
+        HostnameValidator.validateHostnameCompliant(builder.hostname, "hostname", "GenerateAuthenticationTokenRequest");
+
+        this.hostname = Validate.notEmpty(builder.hostname, "hostname");
+        this.port = Validate.isPositive(builder.port, "port");
+        this.username = Validate.notEmpty(builder.username, "username");
+        this.region = builder.region;
+        this.credentialsProvider = builder.credentialsProvider;
+    }
+
+    public String hostname() {
```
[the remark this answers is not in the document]
abrooksv: Should be inherited from the interface definition

services/rds/src/main/java/software/amazon/awssdk/services/rds/RdsUtilities.java
```diff
@@ -0,0 +1,66 @@
...
+     */
+    static Builder builder() {
+        return new DefaultRdsUtilities.DefaultBuilder();
+    }
+
+    default String generateAuthenticationToken(Consumer<GenerateAuthenticationTokenRequest.Builder> getUrlRequest) {
+        return generateAuthenticationToken(GenerateAuthenticationTokenRequest.builder().applyMutation(getUrlRequest).build());
+    }
+
+    default String generateAuthenticationToken(GenerateAuthenticationTokenRequest request) {
```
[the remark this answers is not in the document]
abrooksv: Yep, missed this one.

services/rds/src/main/java/software/amazon/awssdk/services/rds/model/GenerateAuthenticationTokenRequest.java
```diff
@@ -0,0 +1,173 @@
...
+        return hostname;
+    }
+
+    public int port() {
+        return port;
+    }
+
+    public String userName() {
```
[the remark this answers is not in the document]
abrooksv: Changed

services/rds/src/main/java/software/amazon/awssdk/services/rds/DefaultRdsUtilities.java
abrooksv: Yep

services/rds/src/main/java/software/amazon/awssdk/services/rds/DefaultRdsUtilities.java
abrooksv: Ok, it lacked the thread safe annotation so wasnt sure.

services/rds/src/test/java/software/amazon/awssdk/services/rds/DefaultRdsUtilitiesTest.java
```diff
@@ -0,0 +1,128 @@
...
+import java.time.ZoneId;
+import java.time.ZonedDateTime;
+import software.amazon.awssdk.services.rds.DefaultRdsUtilities.DefaultBuilder;
+
+import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
+import static org.assertj.core.api.AssertionsForClassTypes.assertThatThrownBy;
+
+public class DefaultRdsUtilitiesTest {
+    private final ZoneId utcZone = ZoneId.of("UTC").normalized();
+    private final Clock fixedClock = Clock.fixed(ZonedDateTime.of(2016, 11, 7, 17, 39, 33, 0, utcZone).toInstant(), utcZone);
```
abrooksv, line 19: Test data came from AWS CLI: https://github.com/aws/aws-cli/blob/master/tests/functional/rds/test_generate_db_auth_token.py#L50

services/rds/src/main/java/software/amazon/awssdk/services/rds/RdsUtilities.java
```diff
@@ -0,0 +1,104 @@
...
+ * RdsUtilities utilities = RdsUtilities.builder()
+ *  .credentialsProvider(DefaultCredentialsProvider.create())
+ *  .region(Region.US_WEST_2)
+ *  .build()
+ * </pre>
+ * </p>
+ *
+ * <p>
+ * 2) Using the low-level client {@link RdsClient#utilities()} (or {@link RdsAsyncClient#utilities()}} method. This is
+ * recommended as SDK will use the same configuration from the {@link RdsClient} object to create the {@link RdsUtilities} object.
```
zoewangg, line 39: Nit, should we move this option up since it's recommended

zoewangg:
Left one minor comment, otherwise LGTM. Can you squash all commits?

lbruun:
Just a few clarifications.
Is it correctly understood that no network traffic is involved in generating an IAM Auth Token?

If so, then why model it as if there is?  Specifically:

* The PR has a `model.*Request` class. Such classes are used throughout the SDK for request/replies that travel across the network. Yet in this case the `GenerateAuthenticationTokenRequest` class serves a completely different purpose, namely solely as a basis for providing input to a method.  Comparing it to the SDKv1 implementation you'll notice that at least in that implementation they have _not_ lumped the Request class into the same package where network request/reply classes exist.

* In the PR the `RdsUtilities` class can be created from the RDS client class too, instead of directly. Again this tricks the user into thinking that the client is somehow involved (it isn't !). The benefits for the API user of creating the `RdsUtilities` via the client are close to zero, IMO (not having to specify CredentialsProvider and Region again). You can imagine use-cases where the SDK is indeed _only_ being used for creating an IAM Auth Token and I bet you'll then see users who thinks - partly lured by the Javadoc documentation too - that they first have to create an `RdsClient` in order to create an IAM Auth Token.

Overall, I reckon the idea is to model it after `S3Utilities` class?  Well, that implementation suffers from the same strange type of complexity in my opinion. At least that class has in its Javadoc "Note: This class does not make network calls." which is not the case for `RdsUtilities`.

Why not simply a class with static methods, named `Utils`, not `Utilities` ?  I think this would be inline with how the rest of SDK does it. Even in a `Utils` class you can still springle it with Builder pattern if the method calls would become overly elaborate.

zoewangg:
Hi @lbruun, first of all, thank you for reviewing the design and providing feedback!

> Is it correctly understood that no network traffic is involved in generating an IAM Auth Token?

Yes, it is correct.

> The PR has a model.*Request class. Such classes are used throughout the SDK for request/replies that travel across the network. Yet in this case the GenerateAuthenticationTokenRequest class serves a completely different purpose, namely solely as a basis for providing input to a method. Comparing it to the SDKv1 implementation you'll notice that at least in that implementation they have not lumped the Request class into the same package where network request/reply classes exist.

`software.amazon.awssdk.services.*.model` packages contains POJOs, but they are not limited only to AWS service requests classes. 
If a model class extends `AwsRequest` class, it means this class is a request class to send to AWS services over wire. `GenerateAuthenticationTokenRequest` is not a subclass of `AwsRequest` in this case.

> In the PR the RdsUtilities class can be created from the RDS client class too, instead of directly. Again this tricks the user into thinking that the client is somehow involved (it isn't !). The benefits for the API user of creating the RdsUtilities via the client are close to zero, IMO (not having to specify CredentialsProvider and Region again). You can imagine use-cases where the SDK is indeed only being used for creating an IAM Auth Token and I bet you'll then see users who thinks - partly lured by the Javadoc documentation too - that they first have to create an RdsClient in order to create an IAM Auth Token.

`RdsUtilities` can be created either from the builder or the RDS client class. I understand it might be a bit confusing to some customers, and we definitely should improve the Javadocs. The main reason we create a static factory method in the client class is to improve the discoverability of the utility methods. We had internal usability study before where we interviewed customers asking their opinions, and the result indicates that providing a static method to instantiate a class in the SDK client makes the class far more discoverable. When customers type `RdsClient` in their IDEs, `utilities` method will show up in the code completion popup window whereas if it's a completely standalone class, customers might never find out the existence of the utilities class. Also, we've gotten a lot of feedback on 1.11.x that it's difficult to find such features hidden in a Utils class.

We will work on the javadocs to make it clear that `RdsUtilities` does not make network calls. Please let us know if you have further questions/suggestions.

zoewangg (approved):

lbruun:
Hi @zoewangg 
Thank you for good explanation. You argue well. I think my concerns can be mitigated with some Javadoc comments:

`RdsUtilties` class:  add "This class does not perform network traffic".

`GenerateAuthenticationTokenRequest` class:  Instead of "Request object to generate an auth token for IAM database authentication." then "Input parameters for generating an auth token for IAM database authentication".  (to me at least, the word 'request' is in my brain associated with network traffic)


And speaking of hidden utility classes:   Why not take advantage of `AwsHostNameUtils.parseSigningRegion() method`  or at least reference it in comments in Javadoc ? 

use like this:

```java
AwsHostNameUtils.parseSigningRegion(hostname, "rds");
```

Of course it should be optional. User should still be able to provide the Region explicitly. But providing the Region this way is IMO very likely to be correct. In fact I struggle to see when it won't yield the correct result given that the value used for `hostname` must be the  actual RDS endpoint hostname (it can't be an alias, a CNAME). But perhaps there's some Outpost scenario I'm missing when saying it will always yield correct result?

zoewangg:
@lbruun thanks for the javadoc suggestions! I think they make sense. We have added a note in `RdsUtilities` to make it clear that it doesn't make network calls. @abrooksv can we update the javadocs for `GenerateAuthenticationTokenRequest`?

`AwsHostNameUtils` is more of an internal class, and it doesn't really work well for all endpoints such as VPC endpoint and gov endpoint. It's a a lot of maintenance to enumerate all existing patterns and keep up with the new endpoint patterns.

zoewangg (approved):

zoewangg:
I'll go ahead and merge this PR. 

@lbruun thanks again for your feedback 🙂 please let us know if you have further feedback by creating a new issue, and we will address them.

@abrooksv thank you for your contribution, Austin! 🎉

hzmj9h:
I would like to check this changes committed to 2.17.14 version. I couldn't able to locate the software.amazon.awssdk.services.rds.DefaultRdsUtilities.DefaultBuilder newer version.

Issue #1986, opened 2020-08-13
Incomplete Field Mapping for S3ControlException via S3ControlClient.createJob()


LukeHornibrookHaventec:
## Describe the bug
A failed request to `S3ControlClient.createJob()` results in an `S3ControlException` with null message and error code. However, the API response body has the expected `<Code>` and `<Message>` elements, eg:

```
<Error><Code>InvalidRequest</Code><Message>Invalid job manifest</Message><RequestId>****</RequestId><HostId>****</HostId></Error>
```

## Expected Behavior
The response should be marshalled to an `InvalidRequestException` (or equivalent based on the error code) and the `AwsErrorDetails` fully populated.

## Steps to Reproduce
Call `S3ControlClient.createJob()` with only the `accountId` populated, eg:

```
S3ControlClient.create().createJob(x -> {
	x.accountId("123456789012");
});
```

## Possible Solution
Hopefully there is a generalised solution that could resolve other related issues, eg #1983.

## Your Environment
* AWS Java SDK version used: 2.13.41

debora-ito:
Thank you for reporting @LukeHornibrookHaventec, I can reproduce it. Although I see a 
```
<Code>InvalidRequest</Code><Message>Missing role arn</Message>...
```
with the sample code you provided, I think it falls under the same issue with unmarshalling.

cenedhryn:
@LukeHornibrookHaventec 
This issue is now fixed; you will get an S3ControlException back with an InvalidRequest error code.

PR #2060, opened 2020-09-23
Adds support for parsing errors with a top level root XML structure

cenedhryn:
## Description
- Adds an interceptor to modify a generic unmarshalled S3ControlException when the message contains 'null' and the XML body contains 'Error' as the top level element. The interceptor will attempt to identify a modeled exception if it exists, otherwise return a new S3ControlException with message/error code correctly filled in. 
- Supports `InvalidRequestException` 
- Detected and tracked in https://github.com/aws/aws-sdk-java-v2/issues/1986

## Motivation and Context
Errors from the server side that contain an XML body with error description should have this structure:
~~~
<ErrorResponse>
  <Error>
    <AccountId></AccountId>
    <Code></Code>
    <Message></Message>
  </Error>
</ErrorResponse>
~~~
However at least one error type, InvalidRequest, that S3Control returns (observed through the Jobs CRUD operations) use the S3-structure with no wrapping top level tag:
~~~
  <Error>
    <AccountId></AccountId>
    <Code></Code>
    <Message></Message>
  </Error>
~~~
The error parser that S3 control uses cannot parse the XML document with the Error as the document root and therefore creates an exception with no information. 

## Testing
Added wiremock tests and integration tests, verified that existing errors unmarshall correctly. 

## Types of changes

## Checklist

## License

### DIFF
--- a/services/s3control/pom.xml
+++ b/services/s3control/pom.xml
@@ -53,6 +53,11 @@
         </dependency>
         <dependency>
             <groupId>software.amazon.awssdk</groupId>
+            <artifactId>aws-query-protocol</artifactId>
+            <version>${awsjavasdk.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
             <artifactId>protocol-core</artifactId>
             <version>${awsjavasdk.version}</version>
         </dependency>
@@ -76,6 +81,11 @@
             <groupId>software.amazon.awssdk</groupId>
             <artifactId>sts</artifactId>
             <version>${awsjavasdk.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.github.tomakehurst</groupId>
+            <artifactId>wiremock</artifactId>
             <scope>test</scope>
         </dependency>
     </dependencies>
--- a/services/s3control/src/main/java/software/amazon/awssdk/services/s3control/internal/interceptors/TopLevelXMLErrorInterceptor.java
+++ b/services/s3control/src/main/java/software/amazon/awssdk/services/s3control/internal/interceptors/TopLevelXMLErrorInterceptor.java
@@ -1,0 +1,91 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.services.s3control.internal.interceptors;
+
+import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.awscore.exception.AwsErrorDetails;
+import software.amazon.awssdk.core.interceptor.Context;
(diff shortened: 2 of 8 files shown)


services/s3control/pom.xml
```diff
@@ -51,6 +51,11 @@
             <artifactId>aws-xml-protocol</artifactId>
             <version>${awsjavasdk.version}</version>
         </dependency>
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>aws-query-protocol</artifactId>
+            <version>${awsjavasdk.version}</version>
+        </dependency>
```
millems, line 58: Is this a new dependency, or has it always been there?

services/s3control/src/main/java/software/amazon/awssdk/services/s3control/internal/interceptors/TopLevelXMLErrorInterceptor.java
```diff
@@ -0,0 +1,91 @@
...
+import software.amazon.awssdk.core.interceptor.ExecutionInterceptor;
+import software.amazon.awssdk.protocols.query.unmarshall.XmlDomParser;
+import software.amazon.awssdk.protocols.query.unmarshall.XmlElement;
+import software.amazon.awssdk.services.s3control.model.InvalidRequestException;
+import software.amazon.awssdk.services.s3control.model.S3ControlException;
+
+/**
+ * Translate S3 style exceptions, which have the Error tag at root instead of wrapped in ErrorResponse.
+ * If the exception follows this structure but isn't known, create an S3ControlException with the
+ * error code and message.
```
millems, line 31: Isn't this already implemented in the S3 client?

services/s3control/src/test/resources/log4j.properties
```diff
@@ -0,0 +1,33 @@
...
+# Adjust to see more / less logging
+#log4j.logger.com.amazonaws.ec2=DEBUG
+
+# HttpClient 3 Wire Logging
+#log4j.logger.httpclient.wire=DEBUG
+
+# HttpClient 4 Wire Logging
+#log4j.logger.org.apache.http.wire=DEBUG
+log4j.logger.org.apache.http=DEBUG
+log4j.logger.org.apache.http.wire=DEBUG
```
millems, line 32: That's a lot of logging for the default.

services/s3control/pom.xml
cenedhryn: It's a new dependency

services/s3control/src/main/java/software/amazon/awssdk/services/s3control/internal/interceptors/TopLevelXMLErrorInterceptor.java
cenedhryn: In S3 the whole client is customized to use the AwsS3ProtocolFactory instead of AwsXmlProtocolFactory. It's not possible to use it dynamically for one or two exception types.

services/s3control/src/test/resources/log4j.properties
cenedhryn: Indeed. Good catch, in a previous incarnation it was commented out but missed it in following commits.

services/s3control/pom.xml
millems: Unfortunately it's not backwards-compatible to add new dependencies on existing packages, since it causes Brazil major version conflicts. Is there any way to avoid this new dependency?

services/s3control/src/main/java/software/amazon/awssdk/services/s3control/internal/interceptors/TopLevelXMLErrorInterceptor.java
millems: Can we not use the AwsS3ProtocolFactory for all exception types?

cenedhryn:
Superceded by a different PR

PR #2062, opened 2020-09-23
Add functional tests and integration tests for Waiters

Quanzzzz:
Add integration tests for DynamoDB Waiters;
Add functional tests for RestJsonWithWaiters Waiters;

## License

### DIFF
--- a/test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/waiters/WaitersSyncFunctionalTest.java
+++ b/test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/waiters/WaitersSyncFunctionalTest.java
@@ -1,0 +1,133 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.services.waiters;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.mockito.Matchers.any;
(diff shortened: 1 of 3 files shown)


test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/waiters/WaitersSyncFunctionalTest.java
```diff
@@ -0,0 +1,93 @@
...
+import software.amazon.awssdk.core.retry.backoff.BackoffStrategy;
+import software.amazon.awssdk.core.waiters.PollingStrategy;
+import software.amazon.awssdk.core.waiters.WaiterResponse;
+import software.amazon.awssdk.http.SdkHttpResponse;
+import software.amazon.awssdk.services.restjsonwithwaiters.RestJsonWithWaitersClient;
+import software.amazon.awssdk.services.restjsonwithwaiters.model.AllTypesRequest;
+import software.amazon.awssdk.services.restjsonwithwaiters.model.AllTypesResponse;
+import software.amazon.awssdk.services.restjsonwithwaiters.waiters.RestJsonWithWaitersWaiter;
+
+public class WaitersSyncFunctionalTest {
```
dagnir, line 21: Seems like we need to test other scenarios as well such as 

 - What happens when we reach max number of attempts
 - What happens when we get an unexpected exception
 - What happens to the client and scheduled executor when the the waiter is closed?
 - etc.

cc: @zoewangg for more suggestions

Quanzzzz: The max number of attempts case is added in sync Waiters. Unexpected exception case has been covered by the tests for generic Waiters. For the third case, if the client and scheduled executor are configured while creating the Waiters then they won't be closed while the Waiters are closed, and tests related to this case are also added.

dagnir (approved):

test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/waiters/WaitersSyncFunctionalTest.java
zoewangg: @dagnir Good call! I added the test cases with unexpected exception and caught a bug in waiterRuntime 😅

Quanzzzz: 😅Guess that's my fault. I misunderstood the unmatched exception case in the `BaseWaiterTest`, thanks for making sure about it again.

PR #2066, opened 2020-09-25
add per request waiterOverrideConfiguration overloaded waiter operation

zoewangg:
## Description
- add per request waiterOverrideConfiguration overloaded waiter operation
- remove `throwOnUnmatchedExceptionWaiter` from `WaitersRuntime` because it would cause async waiter future to hang. unmatched exception is handled in the generic waiter implementations
- add try-catch in generic async waiter `whenComplete` block to fix the issue where any exception thrown there could cause the future to hang
- add more functional tests

## Types of changes

## Checklist

## License

### DIFF
--- a/codegen/src/test/resources/software/amazon/awssdk/codegen/poet/waiters/query-sync-waiter-class.java
+++ b/codegen/src/test/resources/software/amazon/awssdk/codegen/poet/waiters/query-sync-waiter-class.java
@@ -4,6 +4,7 @@ import java.time.Duration;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.function.Consumer;
 import software.amazon.awssdk.annotations.Generated;
 import software.amazon.awssdk.annotations.SdkInternalApi;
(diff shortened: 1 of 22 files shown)


codegen/src/test/resources/software/amazon/awssdk/codegen/poet/waiters/query-sync-waiter-class.java
```diff
@@ -66,6 +66,15 @@ private static String errorCode(Throwable error) {
         return postOperationSuccessWaiter.run(() -> client.aPostOperation(applyWaitersUserAgent(aPostOperationRequest)));
     }
 
+    @Override
+    public WaiterResponse<APostOperationResponse> waitUntilPostOperationSuccess(APostOperationRequest aPostOperationRequest,
+                                                                                WaiterOverrideConfiguration overrideConfig) {
+        Waiter<APostOperationResponse> waiter = Waiter.builder(APostOperationResponse.class)
+                                                      .acceptors(postOperationSuccessWaiterAcceptors())
+                                                      .overrideConfiguration(postOperationSuccessWaiterConfig(overrideConfig)).build();
+        return waiter.run(() -> client.aPostOperation(applyWaitersUserAgent(aPostOperationRequest)));
```
dagnir, line 75: Is it possible to just overload the Waiter `run()` method to take an the override instead? That way the existing waiter can be reused without instantiating a new one when a override is given. Would also simplify codegen

zoewangg: Instantiating a new waiter is actually very cheap, but I do like your idea. Well, too bad I already made the codegen change :( Anyway, I'll update it

dagnir: But hopefully less code to maintain :). Also customers using the generic `Waiter` directly would benefit as well

dagnir (approved):

PR #2075, opened 2020-09-30
Remove flaky integ test

zoewangg:
## Description
Remove flaky integ test that is testing service and not the SDK

## License

### DIFF
--- a/services/s3control/src/it/java/software.amazon.awssdk.services.s3control/S3ControlIntegrationTest.java
+++ b/services/s3control/src/it/java/software.amazon.awssdk.services.s3control/S3ControlIntegrationTest.java
@@ -103,16 +103,6 @@ public class S3ControlIntegrationTest extends AwsIntegrationTestBase {
     }
 
     @Test
-    public void getPublicAccessBlock_NoSuchConfig() {
-        try {
-            client.getPublicAccessBlock(r -> r.accountId(accountId));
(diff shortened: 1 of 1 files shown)


services/s3control/src/it/java/software.amazon.awssdk.services.s3control/S3ControlIntegrationTest.java
```diff
@@ -102,16 +102,6 @@ public void getPublicAccessBlock_NoSuchAccount() {
         }
     }
 
-    @Test
-    public void getPublicAccessBlock_NoSuchConfig() {
```
cenedhryn, line 106: I'm a bit concerned that we're diminishing our ability to detect issues with S3Control error model, since we currently have that issue of InvalidRequest having the wrong XML format. This is one of the few tests that verify we can parse the tests from the backend. On the other hand, we should follow the model. Can you think of another test that would be less flaky to replace this one? Too many tags maybe?

zoewangg: Good call. I think we should mock the service response to verify the exception unmarshalling instead of calling the service. 

Can we merge this first since this might block the release? I'll create a backlog item to add mock tests for s3 control

cenedhryn (approved):

Issue #1822, opened 2020-05-07
DynamoDB Enhanced: Add support for if_not_exists in Update operations


whablake:
DynamoDB supports preventing an overwrite of an existing value in an update operation through the use of the function `if_not_exists(path, value)` [DynamoDB Update Expression Docs](https://docs.amazonaws.cn/en_us/amazondynamodb/latest/developerguide/Expressions.UpdateExpressions.html#Expressions.UpdateExpressions.SET.PreventingAttributeOverwrites). The DynamoDB Enhanced client does not have a mechanism to achieve the same behavior. You must do a read from dynamodb, then omit the values you do not want to overwrite in a subsequent write operation.

## Describe the Feature
Allow a user to specify attributes in their table schema that should not be overwritten in an update operation.

This is the basic use case, I have a class that has attributes similar to:

```java
public class Customer {
    private String accountId;
    private String name;
    private Instant createdDate; // Once this is written, never overwrite
    private Instant updatedDate;

    static final TableSchema<Customer> CUSTOMER_TABLE_SCHEMA =
       TableSchema.builder(Customer.class)
           .newItemSupplier(Customer::new)
           .addAttribute(String.class, a -> a.name("account_id")
                                      .getter(Customer::getAccountId)
                                      .setter(Customer::setAccountId)
                                      .tags(primaryPartitionKey()))
           .addAttribute(String.class, a -> a.name("name")
                                      .getter(Customer::getName)
                                      .setter(Customer::setName))
           .addAttribute(Instant.class, a -> a.name("created_date")
                                      .getter(Customer::getCreatedDate)
                                      .setter(Customer::setCreatedDate))
           .addAttribute(Instant.class, a -> a.name("updated_date")
                                      .getter(Customer::getUpdatedDate)
                                      .setter(Customer::setUpdatedDate))
           .build();
}
```

If the name field is updated, I would like to update both "name" and "updated_date" but not "created_date" in a single operation. This is possible with the low level client, but not the enhanced client. 

The sequence of events is as follows:

Customer "xyz" specifies their name for the first time

```json
{
"accountId": "xyz",
"name" : "foo",
"createdDate" : "2020-05-01T00:00:00Z",
"updatedDate" : "2020-05-01T00:00:00Z"
}
```

[...]

PR #2076, opened 2020-10-01
DynamoDB Enhanced Client: Added support for attribute level custom up…

bmaizels:
…date behaviors

Implements suggestion in https://github.com/aws/aws-sdk-java-v2/issues/1822

## Description
Copied from README.md update:

### Changing update behavior of attributes
It is possible to customize the update behavior as applicable to individual attributes when an 'update' operation is
performed (e.g. UpdateItem or an update within TransactWriteItems).

For example, say like you wanted to store a 'created on' timestamp on your record, but only wanted its value to be
written if there is no existing value for the attribute stored in the database then you would use the 
WRITE_IF_NOT_EXISTS update behavior. Here is an example using a bean:

```java
@DynamoDbBean
public class Customer extends GenericRecord {
    private String id;
    private Instant createdOn;

    @DynamoDbPartitionKey
    public String getId() { return this.id; }
    public void setId(String id) { this.name = id; }

    @DynamoDbUpdateBehavior(UpdateBehavior.WRITE_IF_NOT_EXISTS)
    public Instant getCreatedOn() { return this.createdOn; }    
    public void setCreatedOn(Instant createdOn) { this.createdOn = createdOn; }
}
```

Same example using a static table schema:

```java
static final TableSchema<Customer> CUSTOMER_TABLE_SCHEMA =
     TableSchema.builder(Customer.class)
       .newItemSupplier(Customer::new)
       .addAttribute(String.class, a -> a.name("id")
                                         .getter(Customer::getId)
                                         .setter(Customer::setId)
                                         .tags(primaryPartitionKey()))
       .addAttribute(Instant.class, a -> a.name("createdOn")
                                          .getter(Customer::getCreatedOn)
                                          .setter(Customer::setCreatedOn)
                                          .tags(updateBehavior(UpdateBehavior.WRITE_IF_NOT_EXISTS)))
       .build();
```


services-custom/dynamodb-enhanced/src/test/java/software/amazon/awssdk/enhanced/dynamodb/functionaltests/models/RecordWithUpdateBehaviors.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    public void setId(String id) {
+        this.id = id;
+    }
+
+    @DynamoDbUpdateBehavior(WRITE_IF_NOT_EXISTS)
+    public Instant getCreatedOn() {
+        return createdOn;
+    }
```
millems, line 46: I'm confused about the use of this. Would you actually use it for "created date" if it doesn't work for putItem (only for updateItem)? Wouldn't this be `firstUpdatedDate`, which seems kind of useless?

bmaizels: The use-case, as it was put to me, was to avoid branching logic and always do a uniform update. It also allows schema evolution so they can update the record and write a value with a new attribute that didn't exist in the schema before only if it hadn't been set before in the database. I imagine it's the same rationale that lead to DDB implementing if_not_exists in their update expressions at all, this is really just a 'plumbing it through' exercise.

millems: I think it's probably more useful at the request level than the schema level, but that's a different feature request, I suppose.

millems (approved):

takarabt:
Hi @bmaizels,

Are you planning to add the possibility to ignore null values when updating a record?

I think this is a very common use case. For example, if I want to change only the description for a given record using the DynamoDBMapper (sdk v1) I could use the object {"id": "123", "description": "My new description"}, Using the same object with the DynamoDB Enhanced client, the attributes which are not included on the object will be overwritten to null.

Basically we need something similar to UPDATE_SKIP_NULL_ATTRIBUTES -> https://aws.amazon.com/blogs/developer/using-the-savebehavior-configuration-for-the-dynamodbmapper/

For more flexibility, it's probably better to add this as a configuration on the client level.

Regards,

bmaizels:
@takarabt does https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/enhanced/dynamodb/model/UpdateItemEnhancedRequest.html#ignoreNulls-- cover your use case?

takarabt:
@bmaizels this is exactly what we are looking for, thank you.

takarabt:
@bmaizels yes this is what we used, after investigation I figured out that the problem is related the usage of primitive type "int" it's set by default to 0 and not to null.

Thank you.

PR #2078, opened 2020-10-01
Fix wrongly replaced double slash in Apache Http Client

Quanzzzz:
The `sanitizeUri` method in Apache Http Client is wrongly replacing all the double slash charSet in the Uri with `"%2F"` when the path in Uri is `"//"`. This pr will fix this issue and only replace the `"//"` at the end of the Uri.

For the condition of multiple consecutive slashes path, for instance an Uri ending with `"///"`, only the first `"//"` will be replaced with `"/%2F"`. That is, `"http://localhost:80///" -> "http://localhost:80/%2F/"`.

## Types of changes

## License

### DIFF
--- a/http-clients/apache-client/src/main/java/software/amazon/awssdk/http/apache/internal/impl/ApacheHttpRequestFactory.java
+++ b/http-clients/apache-client/src/main/java/software/amazon/awssdk/http/apache/internal/impl/ApacheHttpRequestFactory.java
@@ -71,7 +71,8 @@ public class ApacheHttpRequestFactory {
      */
     private String sanitizeUri(URI uri) {
         String newPath = uri.getPath().replace("//", "/%2F");
-        return uri.toString().replace(uri.getPath(), newPath);
+        String baseUri = uri.toString();
+        return baseUri.substring(0, baseUri.length() - uri.getPath().length()) + newPath;
     }
 
     private void addRequestConfig(final HttpRequestBase base,
--- a/http-clients/apache-client/src/test/java/software/amazon/awssdk/http/apache/internal/impl/ApacheHttpRequestFactoryTest.java
+++ b/http-clients/apache-client/src/test/java/software/amazon/awssdk/http/apache/internal/impl/ApacheHttpRequestFactoryTest.java
@@ -17,6 +17,8 @@ package software.amazon.awssdk.http.apache.internal.impl;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertThat;
+import static org.junit.Assert.assertTrue;
 
 import java.net.InetAddress;
 import java.net.URI;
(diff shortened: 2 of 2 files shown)


http-clients/apache-client/src/main/java/software/amazon/awssdk/http/apache/internal/impl/ApacheHttpRequestFactory.java
```diff
@@ -71,7 +71,13 @@ public HttpRequestBase create(final HttpExecuteRequest request, final ApacheHttp
      */
     private String sanitizeUri(URI uri) {
         String newPath = uri.getPath().replace("//", "/%2F");
-        return uri.toString().replace(uri.getPath(), newPath);
+        String baseUri = uri.toString();
+
+        if (baseUri.endsWith("//")) {
```
zoewangg, line 76: Hmm, you can have a key with leading slash like `/myKey`, so the double slash might not be at the end of the string.

Quanzzzz: Resolved in the latest commit.

http-clients/apache-client/src/main/java/software/amazon/awssdk/http/apache/internal/impl/ApacheHttpRequestFactory.java
```diff
@@ -71,7 +71,8 @@ public HttpRequestBase create(final HttpExecuteRequest request, final ApacheHttp
      */
     private String sanitizeUri(URI uri) {
         String newPath = uri.getPath().replace("//", "/%2F");
-        return uri.toString().replace(uri.getPath(), newPath);
+        String baseUri = uri.toString();
```
joviegas, line 74: Query : 
`String newPath = uri.getPath().replace("//", "/%2F");`
Here we are handling only. "/" what if we havae other special characters ? 
Could you please test it for "@" or try the example given below
```

https://docs.oracle.com/javase/7/docs/api/java/net/URLEncoder.html
For example using UTF-8 as the encoding scheme the string "The string ü@foo-bar" would get converted to "The+string+%C3%BC%40foo-bar" because in UTF-8 the character ü is encoded as two bytes C3 (hex) and BC (hex), and the character @ is encoded as one byte 40 (hex).
```

URLEncoder.encode() might handle for all special characters .. What do you say ?

zoewangg: The uri has already been encoded in https://github.com/aws/aws-sdk-java-v2/blob/master/http-client-spi/src/main/java/software/amazon/awssdk/http/SdkHttpRequest.java#L108 at this point :)

I will create another PR for the fix

zoewangg:
Closing in favor of #2079

http-clients/apache-client/src/main/java/software/amazon/awssdk/http/apache/internal/impl/ApacheHttpRequestFactory.java
joviegas: got it..Thank you.

Issue #2034, opened 2020-09-11
SdkHttpFullRequest builder.URI removes query parameters


zyrain:
When calling the SdkHttpFullRequest builder using the URI default method, and query parameters already in your URI are removed.

## Expected Behavior
Existing query parameters in the URI should be kept.

## Current Behavior
	default Builder uri(URI uri) {
            return this.protocol(uri.getScheme())
                       .host(uri.getHost())
                       .port(uri.getPort())
                       .encodedPath(SdkHttpUtils.appendUri(uri.getRawPath(), encodedPath()));

Calling getRawPath() returns only the path without parameters, and encodedPath() is empty if this is your first builder call.  

## Steps to Reproduce
Create a URI with query parameters.
Call the builder.

## Possible Solution
call URI.getQuery, create a map, then call rawQueryParameters

## Context
We are calling a callback url, provided already fully formed.  Without this we need to tear apart the input url and put it back together to call it.

## Your Environment
* AWS Java SDK version used: 2.14.15
* JDK version used: 11
* Operating System and version: AWS Lambda

debora-ito:
Hi @zyrain thank you for reaching out. That is actually the designed behavior of the `uri` method, as documented in the SdkHttpFullRequest [javadoc](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/http/SdkHttpFullRequest.Builder.html#uri-java.net.URI-):

> Convenience method to set the protocol(), host(), port(), and encodedPath() from a URI object.

But I can see in your use case that you work with a fully formed URI, so this is something we can improve in the SDK. Changing this to a feature request.

Strat1987:
I took a stab to aid with this. please be gentle as this is my first contribution.

Strat1987:
@zyrain I was unable to properly merge the upstream master branch and keep only my single squashed commit as a diff so I started over which closed the previous pull request automatically. still not sure how this should typically be tackled.

PR #2082, opened 2020-10-07
feature 2034: SdkHttpFullRequest builder.URI conditionally accept que…

Strat1987:
## Description
When calling the SdkHttpFullRequest builder using the URI default method, query parameters are now supported in the provided URI if keepUriQueryParams is set to true on the builder.\nThis can be useful in case you want to provide an already fully formed URI like a callback URI.

## Motivation and Context
Query parameters that were provided to the builder in the form of the passed URI were omitted before
https://github.com/aws/aws-sdk-java-v2/issues/2034

## Testing
added test cases in SdkHttpRequestResponseTest

## Types of changes

## Checklist

## License


Strat1987:
I'm unsure as to why the CI failed, is there a way I can retrigger?

http-client-spi/src/main/java/software/amazon/awssdk/http/SdkHttpRequest.java
```diff
@@ -134,18 +138,28 @@ default URI getUri() {
...
+                                                 .host(uri.getHost())
+                                                 .port(uri.getPort())
+                                                 .encodedPath(SdkHttpUtils.appendUri(uri.getRawPath(), encodedPath()));
+            if (uri.getQuery() != null) {
+                Pattern.compile("\\s*&\\s*")
+                       .splitAsStream(uri.getQuery().trim())
+                       .map(s -> s.split("=", 2))
+                       .collect(groupingBy(a -> a[0], mapping(a -> a[1], toList())))
+                       .forEach((paramKey, paramValues) -> paramValues
+                           .forEach(paramValue -> this.appendUriQueryParameter(paramKey, paramValue)));
```
dagnir, line 160: I don't think appending would be the correct behavior here; the current set of query parameters should be cleared; otherwise something like this

```java
builder.uri(myUri1).uri(myUri2);
```

will contain the query parameters from both URIs.

dagnir, line 160: Think the query parameters need to be decoded as well

dagnir (changes requested):
Thank you for the PR!

http-client-spi/src/main/java/software/amazon/awssdk/http/SdkHttpRequest.java
Strat1987: @dagnir Thanks for the review. I've tried my best to update the upstream master but to have 2 commits now that I am unsure of how to properly squash into a single one (9d6fae0647 and 43a43be5a7)

dagnir:
Hmm it's probably easier to reset your branch, something like:

```
git fetch origin
git checkout -b reset-branch origin/master
git checkout 9d6fae0647cc2516d6cefb7ad6ea70a2b11a89dd http-client-spi
git checkout 9d6fae0647cc2516d6cefb7ad6ea70a2b11a89dd .changes
git commit
```

dagnir:
BTW, I suspect what happened was that the when attempting to squash the commits, the merge commit was included in the commit range

Strat1987:
@dagnir I understand this leaves me with a local reset-branch including the latest changes from upstream (as I have origin set to my fork). It's unclear to me how I can then make sure to force push this to the httpclientspi-uri-with-query-params branch on my origin?

Strat1987:
@dagnir I've temporary deleted my local httpclientspi-uri-with-query-params branch, checkout -b from the reset-branch to a new local httpclientspi-uri-with-query-params and then force pushed that.
Seems cumbersome to me, as well as it does not store my local changes in case something goes wrong (although I could've temporary branched of in a backup branch before deleting the local copy)

dagnir:
> I understand this leaves me with a local reset-branch including the latest changes from upstream (as I have origin set to my fork). It's unclear to me how I can then make sure to force push this to the httpclientspi-uri-with-query-params branch on my origin?

You can use a refspec to push between your branches

```
git push origin reset-branch:httpclientspi-uri-with-query-params 
```

http-client-spi/src/main/java/software/amazon/awssdk/http/SdkHttpFullRequest.java
```diff
@@ -66,9 +71,29 @@
...
+         */
+        default Builder uriWithQueryParams(URI uri) {
+            Builder builder = this.uri(uri);
+            if (uri.getQuery() != null) {
+                Pattern.compile("\\s*&|%26\\s*")
+                       .splitAsStream(uri.getQuery().trim())
+                       .map(s -> s.split("=|%3D", 2))
+                       .collect(groupingBy(a -> a[0], mapping(a -> a[1], toList())))
+                       .forEach((paramKey, paramValues) -> paramValues
+                           .forEach(paramValue -> this.appendUriQueryParameter(paramKey, paramValue)));
```
dagnir, line 94: This code is pretty complex and is duplicated here and in `SdkHttpRequest`. Let's move this out into `SdkHttpUtils` instead. Should also make the parsing easier to test in isolation

http-client-spi/src/main/java/software/amazon/awssdk/http/SdkHttpFullRequest.java
```diff
@@ -66,9 +71,29 @@
...
+         * Convenience method to set the {@link #protocol()}, {@link #host()}, {@link #port()},
+         * {@link #encodedPath()} and keep query parameters if present from a {@link URI} object.
+         *
+         * @param uri URI containing protocol, host, port, path and optional query parameters.
+         * @return This builder for method chaining.
+         */
+        default Builder uriWithQueryParams(URI uri) {
+            Builder builder = this.uri(uri);
+            if (uri.getQuery() != null) {
+                Pattern.compile("\\s*&|%26\\s*")
```
dagnir, line 89: - I don't think we should including `%26` here as the literal `&` should be the only delimiter we look for.
 - Compiling a pattern each time is inefficient. Once this logic is moved into `SdkHttpUtils`, let's extract the pattern out into a static constant

http-client-spi/src/main/java/software/amazon/awssdk/http/SdkHttpFullRequest.java
```diff
@@ -66,9 +71,29 @@
...
+                       .splitAsStream(uri.getQuery().trim())
+                       .map(s -> s.split("=|%3D", 2))
```
dagnir, line 91: Same here, should only splitting on the literal `=`.

Also seems like this assumes each query parameter is a name-value pair; we should support parameters that don't have a value  as well, like in `https://github.com/aws/aws-sdk-java-v2/pull/2082?foo`

http-client-spi/src/main/java/software/amazon/awssdk/http/SdkHttpRequest.java
```diff
@@ -337,5 +363,17 @@ default Builder putHeader(String headerName, String headerValue) {
...
+        /**
+         * Add a single un-encoded URI query parameter to be included in the created HTTP request.
+         *
+         * <p>This will <b>ADD</b> the value to any existing values already configured with this parameter name in
+         * the builder.</p>
+         *
+         * @param paramName  The name of the query parameter to add
+         * @param paramValue The un-encoded value for the query parameter.
+         */
+        Builder appendUriQueryParameter(String paramName, String paramValue);
```
dagnir, line 376: Hmm why do we need this instead of using `putRawQueryParameter`?

.changes/next-release/feature-HTTPclientspi-25d1ed9.json
```diff
@@ -0,0 +1,5 @@
+{
+    "category": "HTTP client spi",
```
dagnir, line 2: Please capitalize here to `HTTP Client SPI` to be consistent with the rest of the log entries

http-client-spi/src/main/java/software/amazon/awssdk/http/SdkHttpFullRequest.java
Strat1987: In appendUriQueryParameter impl I now decode the incoming paramValue as was suggested before, which is why I reasoned to also split on the decoded delimiter as otherwise the value would not be split properly before appending?

http-client-spi/src/main/java/software/amazon/awssdk/http/SdkHttpRequest.java
Strat1987: since I wanted to make sure not to interfere with already in place param handling

http-client-spi/src/main/java/software/amazon/awssdk/http/SdkHttpRequest.java
```diff
@@ -143,9 +148,29 @@ default URI getUri() {
          */
         default Builder uri(URI uri) {
             return this.protocol(uri.getScheme())
-                       .host(uri.getHost())
-                       .port(uri.getPort())
-                       .encodedPath(SdkHttpUtils.appendUri(uri.getRawPath(), encodedPath()));
+                                                 .host(uri.getHost())
+                                                 .port(uri.getPort())
+                                                 .encodedPath(SdkHttpUtils.appendUri(uri.getRawPath(), encodedPath()));
+        }
```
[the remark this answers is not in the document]
Strat1987: will do

http-client-spi/src/main/java/software/amazon/awssdk/http/SdkHttpFullRequest.java
dagnir: Unfortunately I don't think that's the right strategy. We should do the decoding after we have the individual components of the query parsed out. Otherwise, if the the input `URI` has and encoded `&` or `=`, then we would be parsing it incorrectly.  It's perfectly valid for the parameter name or value to have `&` or `=` it as long as it's encoded.

For example, the pair [contains=equals, foo] in a URI would be `https://foo.bar?contains%3Dequals=foo`, but the code as is will split this as `[contains, equals=foo]`

http-client-spi/src/main/java/software/amazon/awssdk/http/SdkHttpFullRequest.java
Strat1987: Agreed

http-client-spi/src/main/java/software/amazon/awssdk/http/SdkHttpFullRequest.java
dagnir: I think the general strategy should be

 - get components of query string
 - foreach component
   - decode name
   - decode value if present
 - put raw query param

.changes/next-release/feature-HTTPclientspi-25d1ed9.json
Strat1987: I made an educated guess before so was unaware of the casing significance

http-client-spi/src/main/java/software/amazon/awssdk/http/SdkHttpFullRequest.java
Strat1987: I haven't thought about it in that way, makes sense to split like that and then decode both parts before adding.

On reusing the put raw query param bullet, that would mean we cannot differ on where te params come from as I did now in:

    private Map<String, List<String>> determineQueryParameters(Builder builder) {
        if (builder.uriQueryParameters.isEmpty()) {
            return builder.queryParametersAreFromToBuilder
                   ? builder.queryParameters
                   : deepUnmodifiableMap(builder.queryParameters, () -> new LinkedHashMap<>());
        }
        return builder.uriQueryParameters;
    }

utils/src/main/java/software/amazon/awssdk/utils/http/SdkHttpUtils.java
```diff
@@ -322,4 +328,15 @@ public static String appendUri(String baseUri, String path) {
     public static boolean isSingleHeader(String h) {
         return SINGLE_HEADERS.contains(StringUtils.lowerCase(h));
     }
+
+    /**
+     * Extracts query parameters from the given URI
+     */
+    public static Map<String, List<String>> uriParams(URI uri) {
+        return Pattern.compile(QUERY_PARAM_PATTERN)
+               .splitAsStream(uri.getQuery().trim())
```
dagnir, line 337: BTW, I think we should be looking at `getRawQuery()` here, since `getQuery()` will give you the decoded form which is not what we want

http-client-spi/src/test/java/software/amazon/awssdk/http/SdkHttpRequestResponseTest.java
```diff
@@ -401,13 +402,41 @@ public BuilderProxy setMap(Map<String, List<String>> map) {
...
+        final SdkHttpFullRequest sdkHttpFullRequest = SdkHttpFullRequest.builder().method(SdkHttpMethod.POST).uri(uri).build();
+        assertThat(sdkHttpFullRequest.getUri().getQuery()).isNullOrEmpty();
+    }
+
+    @Test
+    public void testSdkHttpFullRequestBuilderUriWithQueryParams() throws URISyntaxException {
+        URI uri = new URI("https://github.com/aws/aws-sdk-java-v2/issues/2034?reqParam=1234&oParam=3456%26reqParam%3D5678");
+        final SdkHttpFullRequest sdkHttpFullRequest =
+            SdkHttpFullRequest.builder().method(SdkHttpMethod.POST).uriWithQueryParams(uri).build();
+        assertThat(sdkHttpFullRequest.getUri().getQuery()).contains("reqParam=1234", "oParam=3456", "reqParam=5678");
```
dagnir, line 424: I don't think this assertion is correct per my comment earlier re: parsing.

Splitting `reqParam=1234&oParam=3456%26reqParam%3D5678` on the `&` delimiter should give `reqParam=1234` and `oParam=3456%26reqParam%3D5678` as the two name-value pairs, then parsing each as a pair gives [reqParam, 1234] and `[oParam, 3456%26reqParam%3D5678]`

http-client-spi/src/main/java/software/amazon/awssdk/http/SdkHttpFullRequest.java
dagnir: Hmm I don't think it should be necessary to distinguish the source. `uriWithQueryParams` acts basically like a static constructor for `SdkHttpRequest`

http-client-spi/src/main/java/software/amazon/awssdk/http/DefaultSdkHttpFullRequest.java
```diff
@@ -349,6 +368,19 @@ private void copyHeadersIfNeeded() {
             }
         }
 
+        @Override
+        public SdkHttpFullRequest.Builder appendUriQueryParameter(String paramName, String paramValue) {
+            try {
+                this.uriQueryParameters.computeIfAbsent(paramName, k -> new ArrayList<>())
+                                       .add(URLDecoder.decode(paramValue, StandardCharsets.UTF_8.toString()));
+            } catch (UnsupportedEncodingException e) {
+                log.warn("Could not decode {}={} using {}", paramName, paramValue, StandardCharsets.UTF_8);
```
dagnir, line 377: We should just fail instead of logging a warning here

dagnir:
Discussed with millems@ and we think just changing the `uri()` behavior is okay; we don't need to introduce a separate method. We acknowledge the possibility of this can break some use case out there, but the opposite is more likely that customers are expecting the request to contain the query params from the input `URI`.

http-client-spi/src/main/java/software/amazon/awssdk/http/SdkHttpRequest.java
roexber: Done

http-client-spi/src/main/java/software/amazon/awssdk/http/SdkHttpRequest.java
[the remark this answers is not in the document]
roexber: Done

http-client-spi/src/main/java/software/amazon/awssdk/http/SdkHttpFullRequest.java
```diff
@@ -66,9 +71,29 @@
         @Override
         default Builder uri(URI uri) {
             return this.protocol(uri.getScheme())
-                       .host(uri.getHost())
-                       .port(uri.getPort())
-                       .encodedPath(SdkHttpUtils.appendUri(uri.getRawPath(), encodedPath()));
+                    .host(uri.getHost())
+                    .port(uri.getPort())
+                    .encodedPath(SdkHttpUtils.appendUri(uri.getRawPath(), encodedPath()));
+        }
```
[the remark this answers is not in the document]
roexber: Done

http-client-spi/src/main/java/software/amazon/awssdk/http/SdkHttpFullRequest.java
roexber: extracted

http-client-spi/src/main/java/software/amazon/awssdk/http/SdkHttpFullRequest.java
roexber: Attempted to reflect this train of thought

utils/src/main/java/software/amazon/awssdk/utils/http/SdkHttpUtils.java
roexber: You're right, updated and adapted unit test

http-client-spi/src/test/java/software/amazon/awssdk/http/SdkHttpRequestResponseTest.java
roexber: Today I learned! updated and adapted unit test

http-client-spi/src/main/java/software/amazon/awssdk/http/DefaultSdkHttpFullRequest.java
roexber: method is no longer needed

roexber:
I'm hoping the current version is aligned with the expectations

http-client-spi/src/main/java/software/amazon/awssdk/http/SdkHttpRequest.java
```diff
@@ -134,18 +134,24 @@ default URI getUri() {
...
-            return this.protocol(uri.getScheme())
+            Builder builder = this.protocol(uri.getScheme())
                        .host(uri.getHost())
                        .port(uri.getPort())
                        .encodedPath(SdkHttpUtils.appendUri(uri.getRawPath(), encodedPath()));
+            if (uri.getRawQuery() != null) {
+                SdkHttpUtils.uriParams(uri)
+                            .forEach(this::putRawQueryParameter);
+            }
+            return builder;
```
dagnir, line 154: The existing set of query parameters set on the builder should be cleared first

http-client-spi/src/main/java/software/amazon/awssdk/http/SdkHttpFullRequest.java
```diff
@@ -56,19 +56,25 @@
...
-            return this.protocol(uri.getScheme())
+            Builder builder =  this.protocol(uri.getScheme())
                        .host(uri.getHost())
                        .port(uri.getPort())
                        .encodedPath(SdkHttpUtils.appendUri(uri.getRawPath(), encodedPath()));
+            if (uri.getRawQuery() != null) {
+                SdkHttpUtils.uriParams(uri)
+                            .forEach(this::putRawQueryParameter);
+            }
+            return builder;
```
dagnir, line 77: Same here, the existing set of query parameters set on the builder should be cleared first

utils/src/test/java/software/amazon/awssdk/utils/SdkHttpUtilsTest.java
```diff
@@ -173,4 +178,16 @@ public void headersFromCollectionWorksCorrectly() {
         assertThat(SdkHttpUtils.firstMatchingHeaderFromCollection(headers, asList("foo", "nothing"))).hasValue("bar");
         assertThat(SdkHttpUtils.firstMatchingHeaderFromCollection(headers, asList("foo", "other"))).hasValue("foo");
     }
+
+    @Test
+    public void uriParams() throws URISyntaxException {
```
dagnir, line 183: We should have tests for the two requests classes that we changed as well.

utils/src/main/java/software/amazon/awssdk/utils/http/SdkHttpUtils.java
```diff
@@ -322,4 +328,15 @@ public static String appendUri(String baseUri, String path) {
...
         return SINGLE_HEADERS.contains(StringUtils.lowerCase(h));
     }
+
+    /**
+     * Extracts query parameters from the given URI
+     */
+    public static Map<String, List<String>> uriParams(URI uri) {
+        return Pattern.compile(QUERY_PARAM_PATTERN)
+               .splitAsStream(uri.getRawQuery().trim())
+               .map(s -> s.contains("=") ? s.split("=", 2) : new String[] {s, ""})
```
dagnir, line 338: Hmm I don't think using an empty string when there's no `=` is what we want; we should be setting `null` if there's no `=` for the parameter. There should be a test for the result of something like this is

```
URI myUri = URI.create("https://github.com/aws/aws-sdk-for-java-v2?foo");

SdkHttpFullRequest request = SdkHttpFullRequest.builder().uri(myUri).build();

request.getUri();
```

to make sure that the result is not `https://github.com/aws/aws-sdk-for-java-v2?foo=`, which will likely be interpreted differently by the server

utils/src/test/java/software/amazon/awssdk/utils/SdkHttpUtilsTest.java
```diff
@@ -173,4 +178,16 @@ public void headersFromCollectionWorksCorrectly() {
...
+        URI uri = new URI("https://github.com/aws/aws-sdk-java-v2/issues/2034?reqParam=1234&oParam=3456&reqParam=5678&noval"
+                          + "&decoded%26Part=equals%3Dval");
+        Map<String, List<String>> uriParams = SdkHttpUtils.uriParams(uri);
+        assertThat(uriParams).contains(entry("reqParam", Arrays.asList("1234", "5678")),
+                                       entry("oParam", Collections.singletonList("3456")),
+                                       entry("noval", Arrays.asList("")),
```
dagnir, line 189: As mentioned above, the value for `noval` should be null, not a list containing `""`.

utils/src/main/java/software/amazon/awssdk/utils/http/SdkHttpUtils.java
```diff
@@ -322,4 +328,15 @@ public static String appendUri(String baseUri, String path) {
     public static boolean isSingleHeader(String h) {
         return SINGLE_HEADERS.contains(StringUtils.lowerCase(h));
     }
+
+    /**
+     * Extracts query parameters from the given URI
+     */
+    public static Map<String, List<String>> uriParams(URI uri) {
+        return Pattern.compile(QUERY_PARAM_PATTERN)
```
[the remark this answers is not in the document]
dagnir: BTW, I think a more accurate name would be "QUERY_PARAM_DELIMETER_PATTERN"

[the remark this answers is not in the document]
roexber: the split is safe to use in my opinion since it uses a StringTokenizer if only a single character is provided since java 7 according to http://hg.openjdk.java.net/jdk7/jdk7/jdk/rev/1ff977b938e5

roexber:
Hopeful we're getting to a common understanding of the solution 🙏

dagnir (commented):
LGTM! I'm going kick off an integ test just to make sure none of those tests were affected by this change.

dagnir (approved):

dagnir:
:shipit:! Thank you for all your work on this!

roexber:
> :shipit:! Thank you for all your work on this!

I'm happy I was able to contribute and grateful for the review. I've learned a thing or two!

PR #2091, opened 2020-10-14
Include event type in visitor name

dagnir:
## Description
This commit fixes event visitor method generation by including the *event type*
in the visitor name. Note that event type here is actually the member name in
the encompassing eventstream structure, *not* the Java class of the event.

For example, in the example below, "Foo" is the event type.

```json
"MyEventStream": {
   "type": "structure",
   "eventstream": true,
   "members": {
       "Foo": {
           "shape": "Bar"
       }
   }
}
```

To ensure that this change is backwards compatible with existing services that
have event streams (Transcribe Streaming and Kinesis), we also include a
customization to supress this behavior keep the method name as "visit".

Note that by sheer luck, the existing consumer method setters, e.g.

```java
Builder onSubscribeToShardEvent(Consumer<SubscribeToShardEvent>);
```

are unaffected by this change. See an example [here](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/kinesis/model/SubscribeToShardResponseHandler.Visitor.Builder.html#onSubscribeToShardEvent-java.util.function.Consumer-).
The previous generation scheme for these methods is "on{Generation Java class
name for event shape}"; now it's "on{event type}".  However, both Transcribe and
Kinesis use the shape name as the member name (or vise versa?) in their current
iteration of their event streams.

For example, in Kinesis, SubscribeToShard's output stream is defined as

```json
"SubscribeToShardEventStream":{
  "type":"structure",
  "required":["SubscribeToShardEvent"],
  "members":{
    "SubscribeToShardEvent":{"shape":"SubscribeToShardEvent"},

    ...
    },
  "eventstream":true
}
```

## Types of changes

## Checklist

## License

### DIFF
--- a/codegen/src/test/resources/software/amazon/awssdk/codegen/poet/eventstream/test-visitor-builder.java
+++ b/codegen/src/test/resources/software/amazon/awssdk/codegen/poet/eventstream/test-visitor-builder.java
@@ -13,6 +13,10 @@ final class DefaultEventStreamOperationVisitorBuilder implements EventStreamOper
 
     private Consumer<EventTwo> onEventTwo;
 
+    private Consumer<EventOne> onSecondEventOne;
+
+    private Consumer<EventTwo> onSecondEventTwo;
+
     @Override
     public EventStreamOperationResponseHandler.Visitor.Builder onDefault(Consumer<EventStream> c) {
         this.onDefault = c;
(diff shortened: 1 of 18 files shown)


dagnir:
Note: anything that doesn't use the customization at the moment will fail to compile, e.g. the `protocol-tests` module since the model classes don't know how to call the new visitor methods yet (will be done in following PR). Right now they're hardcoded to to call `visitor.visit(this)` basically.

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/eventstream/test-visitor-builder.java
```diff
@@ -67,5 +91,16 @@ public void visit(EventOne event) {
         public void visit(EventTwo event) {
             onEventTwo.accept(event);
         }
```
zoewangg, line 93: Should we have `visitEventTwo` and `visitEventOne` and deprecate `visit`?

dagnir: I considered this, but I think I don't think that would actually work when you do the dispatch; for example,

```
eventStream.visit(visitor);
```

We don't know if `visit` is implemented, or `visitEventOne` is implemented, or both!

zoewangg: Gotcha

zoewangg (approved):

PR #2097, opened 2020-10-16
Added support for the "contentMd5Required" C2J trait, which will inject the "Content-MD5" header when an operation requires it.

millems:
This functionality was previously hard-coded to certain operations within S3. It's now driven by a trait, so that Content-MD5 does not get calculated when it isn't needed.

Other changes in this commit:
1. Cleaned up some of the code generation poet code around the content-md5-related logic.
2. Added support for empty async XML responses without content-lengths. These would previously throw an exception that they're invalid XML.
3. Fixed an issue where requestBody and asyncRequestBody were not visible in ExecutionInterceptor.afterMarshalling.


test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/HttpChecksumRequiredTest.java
```diff
@@ -0,0 +1,119 @@
...
+import software.amazon.awssdk.services.protocolrestjson.ProtocolRestJsonAsyncClient;
+import software.amazon.awssdk.services.protocolrestjson.ProtocolRestJsonClient;
+import software.amazon.awssdk.services.protocolrestxml.ProtocolRestXmlAsyncClient;
+import software.amazon.awssdk.services.protocolrestxml.ProtocolRestXmlClient;
+
+/**
+ * Verify that the "httpChecksumRequired" C2J trait results in a valid MD5 checksum of the payload being included in the HTTP
+ * request.
+ */
+public class HttpChecksumRequiredTest {
```
dagnir, line 46: Can we include a test for the case where an async content is provided?

dagnir:
Please do an integ test as well

codegen/src/main/java/software/amazon/awssdk/codegen/poet/client/specs/JsonProtocolSpec.java
```diff
@@ -235,36 +225,23 @@ public CodeBlock asyncExecutionHandler(IntermediateModel intermediateModel, Oper
...
+               .add(".withErrorResponseHandler(errorResponseHandler)\n")
+               .add(".withMetricCollector(apiCallMetricCollector)\n")
+               .add(hostPrefixExpression(opModel))
+               .add(discoveredEndpoint(opModel))
+               .add(asyncRequestBody)
+               .add(HttpChecksumRequiredTrait.putHttpChecksumAttribute(opModel))
+               .add(".withInput($L)$L);",
+                    opModel.getInput().getVariableName(), asyncResponseTransformerVariable(isStreaming, isRestJson, opModel));
+
+
```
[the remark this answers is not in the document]
millems: It killed me to add one more line to it.

millems:
We have integration tests that exercise the S3 operations with this trait on them, so it should be covered.

dagnir:
Ah I meant to do a run of the integ tests :)

millems:
Ah, I ran them locally (at least, the S3 and S3 control ones).

test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/HttpChecksumRequiredTest.java
millems: Fixed!

dagnir (approved):

PR #2118, opened 2020-10-26
Add support for SSO credentials provider

Quanzzzz:
Add support for SSO Credentials Provider.

## Description

In this pr, a credential provider that exchanges a resolved SSO login token for temporary AWS credentials is added. Besides, several new classes and tests are added to enable this credentials provider.

Main Updates since the first review:
1. Merged the `SsoCredentialsProvider` into part of the `ProfileCredentialsProvider` so the profile addressing part can be better taken care of;
2. Separate the generation of cached token path and resolution of access token to make it more flexible;
3. Updated all the tests to make them match the current code and achieve better coverage;

Besides the main changes for SSO, there are also some small changes to the code:
1. Corrected the javadocs of annotation `SdkInternalApi` to make it more accurate;
2. Moved the `userHomeDirectory`method out of `ProfileFileLocation` and put it in the `utils` module, thus it can be used by the other modules;

## Motivation and Context

Currently, SSO provides customers an interface where customers can copy-paste snippests to set the environment variables to temporary credentials that they generate.
While the interface is good for console access and fine for occasional API use, it is painful for repeated API use as customers have to go back to the SSO page every hour to get new credentials. Allowing customers to use their existing login username and password to fetch temporart AWS credentials without having to go to an external site is a much more seamless experience.

## Testing

Since this credentials provider include interaction with SSO service and it relies on another SSO login flow, so there isn’t integration tests added here. However, to make sure the functionality is sound, multiple unit tests and functional tests are added.

## Types of changes

## Checklist

## License

### DIFF
--- a/core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/ProfileCredentialsUtils.java
+++ b/core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/ProfileCredentialsUtils.java
@@ -35,6 +35,7 @@ import software.amazon.awssdk.auth.credentials.ContainerCredentialsProvider;
 import software.amazon.awssdk.auth.credentials.EnvironmentVariableCredentialsProvider;
 import software.amazon.awssdk.auth.credentials.InstanceProfileCredentialsProvider;
 import software.amazon.awssdk.auth.credentials.ProcessCredentialsProvider;
+import software.amazon.awssdk.auth.credentials.ProfileCredentialsProviderFactory;
 import software.amazon.awssdk.auth.credentials.StaticCredentialsProvider;
 import software.amazon.awssdk.auth.credentials.SystemPropertyCredentialsProvider;
 import software.amazon.awssdk.core.internal.util.ClassLoaderHelper;
@@ -50,6 +51,8 @@ import software.amazon.awssdk.utils.Validate;
 public final class ProfileCredentialsUtils {
     private static final String STS_PROFILE_CREDENTIALS_PROVIDER_FACTORY =
         "software.amazon.awssdk.services.sts.internal.StsProfileCredentialsProviderFactory";
+    private static final String SSO_PROFILE_CREDENTIALS_PROVIDER_FACTORY =
+        "software.amazon.awssdk.services.sso.auth.SsoProfileCredentialsProviderFactory";
 
     private final Profile profile;
 
@@ -95,18 +98,21 @@ public final class ProfileCredentialsUtils {
      * @param children The child profiles that source credentials from this profile.
      */
     private Optional<AwsCredentialsProvider> credentialsProvider(Set<String> children) {
+        if (properties.containsKey(ProfileProperty.ROLE_ARN) && properties.containsKey(ProfileProperty.WEB_IDENTITY_TOKEN_FILE)) {
+            return Optional.ofNullable(roleAndWebIdentityTokenProfileCredentialsProvider());
+        }
+
+        if (properties.containsKey(ProfileProperty.SSO_ROLE_NAME) || properties.containsKey(ProfileProperty.SSO_ACCOUNT_ID)
+            || properties.containsKey(ProfileProperty.SSO_REGION) || properties.containsKey(ProfileProperty.SSO_START_URL)) {
+            return Optional.ofNullable(ssoProfileCredentialsProvider());
+        }
+
         if (properties.containsKey(ProfileProperty.ROLE_ARN)) {
             boolean hasSourceProfile = properties.containsKey(ProfileProperty.SOURCE_PROFILE);
             boolean hasCredentialSource = properties.containsKey(ProfileProperty.CREDENTIAL_SOURCE);
-            boolean hasWebIdentityTokenFile = properties.containsKey(ProfileProperty.WEB_IDENTITY_TOKEN_FILE);
-            boolean hasRoleArn = properties.containsKey(ProfileProperty.ROLE_ARN);
             Validate.validState(!(hasSourceProfile && hasCredentialSource),
                                 "Invalid profile file: profile has both %s and %s.",
                                 ProfileProperty.SOURCE_PROFILE, ProfileProperty.CREDENTIAL_SOURCE);
-
-            if (hasWebIdentityTokenFile && hasRoleArn) {
-                return Optional.ofNullable(roleAndWebIdentityTokenProfileCredentialsProvider());
-            }
 
             if (hasSourceProfile) {
                 return Optional.ofNullable(roleAndSourceProfileBasedProfileCredentialsProvider(children));
@@ -162,6 +168,17 @@ public final class ProfileCredentialsUtils {
         return ProcessCredentialsProvider.builder()
                                          .command(properties.get(ProfileProperty.CREDENTIAL_PROCESS))
                                          .build();
+    }
+
+    /**
+     * Create the SSO credentials provider based on the related profile properties.
+     */
+    private AwsCredentialsProvider ssoProfileCredentialsProvider() {
+        requireProperties(ProfileProperty.SSO_ACCOUNT_ID,
+                          ProfileProperty.SSO_REGION,
+                          ProfileProperty.SSO_ROLE_NAME,
+                          ProfileProperty.SSO_START_URL);
+        return ssoCredentialsProviderFactory().create(profile);
     }
 
     private AwsCredentialsProvider roleAndWebIdentityTokenProfileCredentialsProvider() {
@@ -259,6 +276,22 @@ public final class ProfileCredentialsUtils {
         } catch (ClassNotFoundException e) {
             throw new IllegalStateException("To use assumed roles in the '" + name + "' profile, the 'sts' service module must "
                                             + "be on the class path.", e);
+        } catch (NoSuchMethodException | InvocationTargetException | InstantiationException | IllegalAccessException e) {
+            throw new IllegalStateException("Failed to create the '" + name + "' profile credentials provider.", e);
+        }
+    }
+
+    /**
+     * Load the factory that can be used to create the SSO credentials provider, assuming it is on the classpath.
+     */
+    private ProfileCredentialsProviderFactory ssoCredentialsProviderFactory() {
+        try {
+            Class<?> ssoProfileCredentialsProviderFactory = ClassLoaderHelper.loadClass(SSO_PROFILE_CREDENTIALS_PROVIDER_FACTORY,
+                                                                                 getClass());
+            return (ProfileCredentialsProviderFactory) ssoProfileCredentialsProviderFactory.getConstructor().newInstance();
+        } catch (ClassNotFoundException e) {
+            throw new IllegalStateException("To use Sso related properties in the '" + name + "' profile, the 'sso' service "
+                                            + "module must be on the class path.", e);
         } catch (NoSuchMethodException | InvocationTargetException | InstantiationException | IllegalAccessException e) {
             throw new IllegalStateException("Failed to create the '" + name + "' profile credentials provider.", e);
         }
--- a/utils/src/main/java/software/amazon/awssdk/utils/UserHomeDirectoryUtils.java
+++ b/utils/src/main/java/software/amazon/awssdk/utils/UserHomeDirectoryUtils.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.utils;
+
+import software.amazon.awssdk.annotations.SdkProtectedApi;
+
+/**
+ * Load the home directory that should be used for the stored file. This will check the same environment variables as the CLI
+ * to identify the location of home, before falling back to java-specific resolution.
+ */
+@SdkProtectedApi
+public final class UserHomeDirectoryUtils {
+
+    private UserHomeDirectoryUtils() {
+
+    }
+
+    public static String userHomeDirectory() {
+        // To match the logic of the CLI we have to consult environment variables directly.
+        // CHECKSTYLE:OFF
+        String home = System.getenv("HOME");
+
+        if (home != null) {
+            return home;
+        }
+
+        boolean isWindows = JavaSystemSetting.OS_NAME.getStringValue()
+                                                     .map(s -> StringUtils.lowerCase(s).startsWith("windows"))
+                                                     .orElse(false);
+
+        if (isWindows) {
+            String userProfile = System.getenv("USERPROFILE");
+
+            if (userProfile != null) {
+                return userProfile;
+            }
+
+            String homeDrive = System.getenv("HOMEDRIVE");
+            String homePath = System.getenv("HOMEPATH");
+
+            if (homeDrive != null && homePath != null) {
+                return homeDrive + homePath;
+            }
+        }
+
+        return JavaSystemSetting.USER_HOME.getStringValueOrThrow();
+        // CHECKSTYLE:ON
+    }
+}
--- a/utils/src/test/java/software/amazon/awssdk/testutils/EnvironmentVariableHelper.java
+++ b/utils/src/test/java/software/amazon/awssdk/testutils/EnvironmentVariableHelper.java
@@ -1,0 +1,146 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.testutils;
+
+import java.lang.reflect.Field;
+import java.security.AccessController;
+import java.security.PrivilegedActionException;
+import java.security.PrivilegedExceptionAction;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+import java.util.function.Consumer;
+import org.junit.rules.ExternalResource;
+import software.amazon.awssdk.utils.SystemSetting;
+
+/**
+ * A utility that can temporarily forcibly set environment variables and
+ * then allows resetting them to the original values.
+ */
+public class EnvironmentVariableHelper extends ExternalResource {
+
+    private final Map<String, String> originalEnvironmentVariables;
+    private final Map<String, String> modifiableMap;
+    private volatile boolean mutated = false;
+
+    public EnvironmentVariableHelper() {
+        // CHECKSTYLE:OFF - This is a specific utility around system environment variables
+        originalEnvironmentVariables = new HashMap<>(System.getenv());
+        modifiableMap = Optional.ofNullable(processEnv()).orElse(envMap());
+        // CHECKSTYLE:ON
+    }
+
+    public void remove(SystemSetting setting) {
+        remove(setting.environmentVariable());
+    }
+
+    public void remove(String key) {
+        mutated = true;
+        modifiableMap.remove(key);
+    }
+
+    public void set(SystemSetting setting, String value) {
+        set(setting.environmentVariable(), value);
+    }
+
+    public void set(String key, String value) {
+        mutated = true;
+        modifiableMap.put(key, value);
+    }
+
+    public void reset() {
+        if (mutated) {
+            synchronized (this) {
+                if (mutated) {
+                    modifiableMap.clear();
+                    modifiableMap.putAll(originalEnvironmentVariables);
+                    mutated = false;
+                }
+            }
+        }
+    }
+
+    @Override
+    protected void after() {
+        reset();
+    }
+
+    private PrivilegedExceptionAction<Void> setAccessible(Field f) {
+        return () -> {
+            f.setAccessible(true);
+            return null;
+        };
+    }
+
+    /**
+     * Static run method that allows for "single-use" environment variable modification.
+     *
+     * Example use:
+     * <pre>
+     * {@code
+     * EnvironmentVariableHelper.run(helper -> {
+     *    helper.set("variable", "value");
+     *    //run some test that uses "variable"
+     * });
+     * }
+     * </pre>
+     *
+     * Will call {@link #reset} at the end of the block (even if the block exits exceptionally).
+     *
+     * @param helperConsumer a code block to run that gets an {@link EnvironmentVariableHelper} as an argument
+     */
+    public static void run(Consumer<EnvironmentVariableHelper> helperConsumer) {
+        EnvironmentVariableHelper helper = new EnvironmentVariableHelper();
+        try {
+            helperConsumer.accept(helper);
+        } finally {
+            helper.reset();
+        }
+    }
+
+    private Map<String, String> envMap() {
+        // CHECKSTYLE:OFF - This is a specific utility around system environment variables
+        return getField(System.getenv().getClass(), System.getenv(), "m");
+        // CHECKSTYLE:ON
+    }
+
+    /**
+     * Windows is using a different process environment.
+     *
+     * See http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/windows/classes/java/lang/ProcessEnvironment.java#l235
+     */
+    private Map<String, String> processEnv() {
+        Class<?> processEnvironment;
+        try {
+            processEnvironment = Class.forName("java.lang.ProcessEnvironment");
+            return getField(processEnvironment, null, "theCaseInsensitiveEnvironment");
+        } catch (ClassNotFoundException e) {
+            return null;
+        }
+    }
+
+    @SuppressWarnings("unchecked")
+    private Map<String, String> getField(Class<?> processEnvironment, Object obj, String fieldName) {
+        try {
+            Field declaredField = processEnvironment.getDeclaredField(fieldName);
+            AccessController.doPrivileged(setAccessible(declaredField));
+
+            return (Map<String, String>) declaredField.get(obj);
+        } catch (IllegalAccessException | NoSuchFieldException | PrivilegedActionException e) {
+            return null;
+        }
+    }
+}
--- a/.changes/next-release/feature-AWSSingleSignon-9e785be.json
+++ b/.changes/next-release/feature-AWSSingleSignon-9e785be.json
@@ -1,0 +1,6 @@
+{
+    "category": "AWS Single Sign-on", 
+    "contributor": "", 
+    "type": "feature", 
+    "description": "Added support for retrieving SSO credentials: https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-sso.html."
+}
--- a/core/annotations/src/main/java/software/amazon/awssdk/annotations/SdkInternalApi.java
+++ b/core/annotations/src/main/java/software/amazon/awssdk/annotations/SdkInternalApi.java
@@ -19,7 +19,7 @@ import java.lang.annotation.ElementType;
 import java.lang.annotation.Target;
 
 /**
- * Marker interface for 'internal' APIs that should not be used outside the core module. Breaking
+ * Marker interface for 'internal' APIs that should not be used outside the same module. Breaking
  * changes can and will be introduced to elements marked as {@link SdkInternalApi}. Users of the SDK
  * and the generated clients themselves should not depend on any packages, types, fields,
  * constructors, or methods with this annotation.
--- a/core/auth/src/main/java/software/amazon/awssdk/auth/credentials/ProfileCredentialsProviderFactory.java
+++ b/core/auth/src/main/java/software/amazon/awssdk/auth/credentials/ProfileCredentialsProviderFactory.java
@@ -1,0 +1,29 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.auth.credentials;
+
+import software.amazon.awssdk.annotations.SdkProtectedApi;
+import software.amazon.awssdk.profiles.Profile;
+
+/**
+ * A factory for {@link AwsCredentialsProvider}s, which can be used to create different credentials providers with
+ * different Profile properties.
+ */
+@FunctionalInterface
+@SdkProtectedApi
+public interface ProfileCredentialsProviderFactory {
+    AwsCredentialsProvider create(Profile profile);
+}
--- a/core/profiles/src/main/java/software/amazon/awssdk/profiles/ProfileFileLocation.java
+++ b/core/profiles/src/main/java/software/amazon/awssdk/profiles/ProfileFileLocation.java
@@ -15,16 +15,15 @@
 
 package software.amazon.awssdk.profiles;
 
+import static software.amazon.awssdk.utils.UserHomeDirectoryUtils.userHomeDirectory;
+
 import java.nio.file.FileSystems;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.Optional;
 import java.util.regex.Pattern;
-import software.amazon.awssdk.annotations.SdkInternalApi;
 import software.amazon.awssdk.annotations.SdkPublicApi;
-import software.amazon.awssdk.utils.JavaSystemSetting;
-import software.amazon.awssdk.utils.StringUtils;
 
 /**
  * A collection of static methods for loading the location for configuration and credentials files.
@@ -46,7 +45,7 @@ public final class ProfileFileLocation {
     public static Path configurationFilePath() {
         return resolveProfileFilePath(
             ProfileFileSystemSetting.AWS_CONFIG_FILE.getStringValue()
-                                                    .orElse(Paths.get(ProfileFileLocation.userHomeDirectory(),
+                                                    .orElse(Paths.get(userHomeDirectory(),
                                                                       ".aws", "config").toString()));
     }
 
@@ -76,43 +75,6 @@ public final class ProfileFileLocation {
      */
     public static Optional<Path> credentialsFileLocation() {
         return resolveIfExists(credentialsFilePath());
-    }
-
-    /**
-     * Load the home directory that should be used for the profile file. This will check the same environment variables as the CLI
-     * to identify the location of home, before falling back to java-specific resolution.
-     */
-    @SdkInternalApi
-    static String userHomeDirectory() {
-        boolean isWindows = JavaSystemSetting.OS_NAME.getStringValue()
-                                                     .map(s -> StringUtils.lowerCase(s).startsWith("windows"))
-                                                     .orElse(false);
-
-        // To match the logic of the CLI we have to consult environment variables directly.
-        // CHECKSTYLE:OFF
-        String home = System.getenv("HOME");
-
-        if (home != null) {
-            return home;
-        }
-
-        if (isWindows) {
-            String userProfile = System.getenv("USERPROFILE");
-
-            if (userProfile != null) {
-                return userProfile;
-            }
-
-            String homeDrive = System.getenv("HOMEDRIVE");
-            String homePath = System.getenv("HOMEPATH");
-
-            if (homeDrive != null && homePath != null) {
-                return homeDrive + homePath;
-            }
-        }
-
-        return JavaSystemSetting.USER_HOME.getStringValueOrThrow();
-        // CHECKSTYLE:ON
     }
 
     private static Path resolveProfileFilePath(String path) {
--- a/core/profiles/src/main/java/software/amazon/awssdk/profiles/ProfileProperty.java
+++ b/core/profiles/src/main/java/software/amazon/awssdk/profiles/ProfileProperty.java
@@ -99,6 +99,28 @@ public final class ProfileProperty {
      */
     public static final String RETRY_MODE = "retry_mode";
 
+    /**
+     * Aws region where the SSO directory for the given 'sso_start_url' is hosted. This is independent of the general 'region'.
+     */
+    public static final String SSO_REGION = "sso_region";
+
+    /**
+     * The corresponding IAM role in the AWS account that temporary AWS credentials will be resolved for.
+     */
+    public static final String SSO_ROLE_NAME = "sso_role_name";
+
+    /**
+     * AWS account ID that temporary AWS credentials will be resolved for.
+     */
+    public static final String SSO_ACCOUNT_ID = "sso_account_id";
+
+    /**
+     * Start url provided by the SSO service via the console. It's the main URL used for login to the SSO directory.
+     * This is also referred to as the "User Portal URL" and can also be used to login to the SSO web interface for AWS
+     * console access.
+     */
+    public static final String SSO_START_URL = "sso_start_url";
+
     private ProfileProperty() {
     }
 }
--- a/services/sso/pom.xml
+++ b/services/sso/pom.xml
@@ -56,5 +56,23 @@
             <artifactId>aws-json-protocol</artifactId>
             <version>${awsjavasdk.version}</version>
         </dependency>
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>profiles</artifactId>
+            <version>${awsjavasdk.version}</version>
+            <scope>compile</scope>
+        </dependency>
+
+        <!-- Test Dependencies -->
+        <dependency>
+            <groupId>com.google.jimfs</groupId>
+            <artifactId>jimfs</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 </project>
--- a/services/sso/src/main/java/software/amazon/awssdk/services/sso/auth/ExpiredTokenException.java
+++ b/services/sso/src/main/java/software/amazon/awssdk/services/sso/auth/ExpiredTokenException.java
@@ -1,0 +1,90 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.services.sso.auth;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import software.amazon.awssdk.annotations.SdkPublicApi;
+import software.amazon.awssdk.core.SdkField;
+import software.amazon.awssdk.core.SdkPojo;
+import software.amazon.awssdk.core.exception.SdkClientException;
+
+/**
+ * <p>
+ * The session token that was passed is expired or is not valid.
+ * </p>
+ */
+@SdkPublicApi
+public final class ExpiredTokenException extends SdkClientException {
+
+    private static final List<SdkField<?>> SDK_FIELDS = Collections.unmodifiableList(Arrays.asList());
+
+    private ExpiredTokenException(Builder b) {
+        super(b);
+    }
+
+    @Override
+    public Builder toBuilder() {
+        return new BuilderImpl(this);
+    }
+
+    public static Builder builder() {
+        return new BuilderImpl();
+    }
+
+    public interface Builder extends SdkPojo, SdkClientException.Builder {
+        @Override
+        Builder message(String message);
+
+        @Override
+        Builder cause(Throwable cause);
+
+        @Override
+        ExpiredTokenException build();
+    }
+
+    static final class BuilderImpl extends SdkClientException.BuilderImpl implements Builder {
+        private BuilderImpl() {
+        }
+
+        private BuilderImpl(ExpiredTokenException model) {
+            super(model);
+        }
+
+        @Override
+        public BuilderImpl message(String message) {
+            this.message = message;
+            return this;
+        }
+
+        @Override
+        public BuilderImpl cause(Throwable cause) {
+            this.cause = cause;
+            return this;
+        }
+
+        @Override
+        public ExpiredTokenException build() {
+            return new ExpiredTokenException(this);
+        }
+
+        @Override
+        public List<SdkField<?>> sdkFields() {
+            return SDK_FIELDS;
+        }
+    }
+}
--- a/services/sso/src/main/java/software/amazon/awssdk/services/sso/auth/SsoCredentialsProvider.java
+++ b/services/sso/src/main/java/software/amazon/awssdk/services/sso/auth/SsoCredentialsProvider.java
@@ -1,0 +1,256 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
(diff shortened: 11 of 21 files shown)


services/sso/src/main/java/software.amazon.awssdk.services.sso/auth/SsoCredentialsProvider.java
```diff
@@ -0,0 +1,303 @@
...
+    /**
+     * A builder for creating a custom {@link SsoCredentialsProvider}.
+     */
+    public interface Builder {
+
+        /**
+         * Define the profile file that should be used by this credentials provider. By default, the
+         * {@link ProfileFile#defaultProfileFile()} is used.
+         */
+        Builder profileFile(ProfileFile profileFile);
```
abrooksv, line 170: Why tie this to the ProfileFile file instead of having the ProfileCredentialProvider create this class like the other providers?

services/sso/src/main/java/software.amazon.awssdk.services.sso/internal/SsoCredentialsUtils.java
```diff
@@ -0,0 +1,165 @@
...
+        validateAndResolveSsoProfile(profile);
+
+        try {
+            String cachedTokenPath = generateCachedTokenPath(ssoStartUrl);
+            File cachedTokenFile = new File(cachedTokenPath);
+
+            tokenProperties = new ObjectMapper().readValue(cachedTokenFile,
+                                                           SsoCredentialsTokenProperties.class);
+            if (tokenExpired(tokenProperties.getExpiresAt())) {
+                throw ExpiredTokenException.builder().message("The SSO session associated with this profile has expired or is"
```
abrooksv, line 79: Hard coding this instead of allowing users to build a callback to handle expiration seems overly limiting and prevents tools built upon the SDK to be able to use SSO

services/sso/src/test/java/software.amazon.awssdk.services.sso/auth/SsoCredentialsProviderTest.java
```diff
@@ -0,0 +1,275 @@
...
+    private static String testTokenDirectory;
+    private SsoCredentialsProvider provider;
+
+    @Rule
+    public final ExpectedException exception = ExpectedException.none();
+
+    @BeforeClass
+    public static void createTestTokenFile() throws Exception {
+        testTokenDirectory = generateCachedTokenPath(START_URL);
+        createTestTokenFile(testTokenDirectory, "2090-01-01T00:00:00Z");
```
abrooksv, line 43: CLI generates timestamps in the format `2020-10-27T03:07:06UTC`, does that lead to errors?

services/sso/src/test/java/software.amazon.awssdk.services.sso/auth/SsoCredentialsProviderTest.java
```diff
@@ -0,0 +1,275 @@
...
+            throw new Exception(String.format("Creating the test token file failed because of %s", e));
+        }
+
+    }
+
+    private static String generateCachedTokenPath(String startUrl) {
+        Validate.notNull(startUrl, "The start url shouldn't be null.");
+        byte[] startUrlBytes = startUrl.getBytes(StandardCharsets.UTF_8);
+        String encodedUrl = new String(startUrlBytes, StandardCharsets.UTF_8);
+        return "/Users/"+System.getProperty("user.name")+"/.aws/sso/cache/" +  sha1Hex(encodedUrl) + ".json";
```
abrooksv, line 255: Seems dangerous to write test data to a users home dir

services/sso/src/main/java/software.amazon.awssdk.services.sso/internal/SsoCredentialsUtils.java
abrooksv: Errors can be the default impl of the callback inside of the Builder

services/sso/src/main/java/software.amazon.awssdk.services.sso/internal/SsoCredentialsUtils.java
```diff
@@ -0,0 +1,165 @@
...
+                                                                               .build();
+        GetRoleCredentialsResponse response = ssoClient.getRoleCredentials(request);
+        RoleCredentials roleCredentials = response.roleCredentials();
+        return AwsSsoCredentials.create(roleCredentials.accessKeyId(), roleCredentials.secretAccessKey(),
+                                        roleCredentials.sessionToken(), roleCredentials.expiration());
+
+    }
+
+    private boolean tokenExpired(String expiresTime) {
+        return Instant.now().isAfter(Instant.parse(expiresTime));
```
abrooksv, line 138: Do we not want to treat expiring soon as expired?

```
When checking to see if a cached token is expired, the SDK MAY treat the token as expired if it will expire within 15 minutes of the expiresAt time
```

services/sso/src/main/java/software.amazon.awssdk.services.sso/internal/SsoCredentialsUtils.java
```diff
@@ -0,0 +1,165 @@
...
+     */
+    public AwsCredentials getUpdatedCredentials() {
+        validateAndResolveSsoProfile(profile);
+
+        try {
+            String cachedTokenPath = generateCachedTokenPath(ssoStartUrl);
+            File cachedTokenFile = new File(cachedTokenPath);
+
+            tokenProperties = new ObjectMapper().readValue(cachedTokenFile,
+                                                           SsoCredentialsTokenProperties.class);
```
millems, line 77: NEVER use ObjectMapper directly! We don't want our libraries to have a hard-dependency on Jackson-databind. Use `JacksonUtils.sensitiveJsonNodeOf` here: (1) so that we only indirectly depend on Jackson (marginally better), and (2) so that we don't log the user's access token if parsing fails.

services/sso/src/main/java/software.amazon.awssdk.services.sso/internal/SsoCredentialsUtils.java
```diff
@@ -0,0 +1,165 @@
...
+import software.amazon.awssdk.services.sso.model.GetRoleCredentialsResponse;
+import software.amazon.awssdk.services.sso.model.RoleCredentials;
+import software.amazon.awssdk.utils.Validate;
+import software.amazon.awssdk.utils.internal.SystemSettingUtils;
+
+@SdkProtectedApi
+public final class SsoCredentialsUtils {
+
+    private static final String TOKEN_DIRECTORY = "/Users/" + SystemSettingUtils.resolveSetting(USER_NAME).get() +
+                                                  "/.aws/sso/cache/";
```
millems, line 49: Is this only valid on Mac? See `ProfileFileLocation` for the proper way to find the customer's home directory.

services/sso/src/main/java/software.amazon.awssdk.services.sso/internal/SsoCredentialsUtils.java
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    private void validateAndResolveSsoProfile(Profile profile) {
+        Map<String, String> configurations = profile.properties();
+
+        if (configurations.containsKey(SSO_ACCOUNT_ID)) {
+            ssoAccountId = configurations.get(SSO_ACCOUNT_ID);
+        } else {
+            throw SdkClientException.create(String.format("This profile is invalid, %s is missing.", SSO_ACCOUNT_ID));
+        }
```
millems, line 99: This is a bit DRYer, and makes it only one hash key lookup in `configurations` instead of two:

```Java
ssoAccountId = Validate.notNull(configurations.get(SSO_ACCOUNT_ID), 
                                "The %s profile is invalid, %s is missing.", profile.name(), SSO_ACCOUNT_ID);
```

services/sso/src/main/java/software.amazon.awssdk.services.sso/internal/SsoCredentialsUtils.java
```diff
@@ -0,0 +1,165 @@
...
+    private static final String TOKEN_DIRECTORY = "/Users/" + SystemSettingUtils.resolveSetting(USER_NAME).get() +
+                                                  "/.aws/sso/cache/";
+
+    private Profile profile;
+    private SsoClient ssoClient;
+
+    private String ssoRegion;
+    private String ssoAccountId;
+    private String ssoRoleName;
+    private String ssoStartUrl;
```
millems, line 57: These not being final makes me nervous about thread-safety. The first two can easily be made final, but the next 4 only happen when we do getUpdatedCredentials. Is there any way to just do the work of getUpdatedCredentials at initialization and create a new `SsoCredentialsUtils` (there's probably a better name) whenever we need a refresh?

services/sso/src/main/java/software.amazon.awssdk.services.sso/internal/SsoCredentialsUtils.java
```diff
@@ -0,0 +1,165 @@
...
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import java.io.File;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.time.Instant;
+import java.util.Map;
+import javax.xml.bind.DatatypeConverter;
```
millems, line 32: Is this dependency always available?

services/sso/src/main/java/software.amazon.awssdk.services.sso/internal/SsoCredentialsUtils.java
```diff
@@ -0,0 +1,165 @@
...
+    private String sha1Hex(String input) {
+        MessageDigest md;
+        try {
+            md = MessageDigest.getInstance("SHA-1");
+            md.update(input.getBytes(StandardCharsets.UTF_8));
+        } catch (NoSuchAlgorithmException e) {
+            throw SdkClientException.builder().message("Unable to use \"SHA-1\" algorithm.").cause(e).build();
+        }
+
+        return DatatypeConverter.printHexBinary(md.digest());
```
millems, line 163: `BinaryUtils`?

services/sso/src/test/java/software.amazon.awssdk.services.sso/auth/SsoCredentialsProviderTest.java
```diff
@@ -0,0 +1,275 @@
...
+    private static String sha1Hex(String input) {
+        MessageDigest md;
+        try{
+            md = MessageDigest.getInstance("SHA-1");
+            md.update(input.getBytes(StandardCharsets.UTF_8));
+        } catch (NoSuchAlgorithmException e) {
+            throw SdkClientException.builder().message("Unable to use \"SHA-1\" algorithm.").cause(e).build();
+        }
+
+        return DatatypeConverter.printHexBinary(md.digest());
```
millems, line 267: BinaryUtils?

services/sso/src/test/java/software.amazon.awssdk.services.sso/auth/SsoCredentialsProviderTest.java
```diff
@@ -0,0 +1,275 @@
...
+        }
+
+    }
+
+    private static String generateCachedTokenPath(String startUrl) {
+        Validate.notNull(startUrl, "The start url shouldn't be null.");
+        byte[] startUrlBytes = startUrl.getBytes(StandardCharsets.UTF_8);
+        String encodedUrl = new String(startUrlBytes, StandardCharsets.UTF_8);
+        return "/Users/"+System.getProperty("user.name")+"/.aws/sso/cache/" +  sha1Hex(encodedUrl) + ".json";
+    }
```
millems, line 256: It's bad form to reproduce the logic you're testing within your tests (e.g. sha1Hex, generateCachedTokenPath), because you're likely going to write the same bug in both places and defeat the purpose of your tests. Can you just hard-code the values you are expecting here?

services/sso/src/test/java/software.amazon.awssdk.services.sso/auth/SsoCredentialsProviderTest.java
millems: +1 Look at how we test profile files for a better example.

services/sso/src/main/java/software.amazon.awssdk.services.sso/auth/SsoCredentialsHolder.java
```diff
@@ -0,0 +1,45 @@
...
+import software.amazon.awssdk.annotations.ThreadSafe;
+
+/**
+ * Holder class used to atomically store a SSO session credentials with its expiration time.
+ */
+@SdkInternalApi
+@ThreadSafe
+public class SsoCredentialsHolder {
+    private final AwsSsoCredentials ssoCredentials;
+    private final Date ssoCredentialsExpiration;
```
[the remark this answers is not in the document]
kiiadi: ya'll should prolly have a linting rule for the `java.util` Date/Time abominations.

services/sso/src/test/java/software.amazon.awssdk.services.sso/auth/SsoCredentialsProviderTest.java
Quanzzzz: The "expiresAt" property of the example cached token in the SEP is using "2019-11-14T04:05:45Z", so here this property is parsed by `Instant.parse()`. I checked the javadoc and it looks this method only takes strings in the form of  `DateTimeFormatter.ISO_INSTANT` so I suppose it might lead to errors. Should I add the addressing of the suffix?

services/sso/src/main/java/software.amazon.awssdk.services.sso/internal/SsoCredentialsUtils.java
Quanzzzz: This part isn't implemented in the BuilderImpl, could you please point it out if there is any other way to implement this?

services/sso/src/test/java/software.amazon.awssdk.services.sso/auth/SsoCredentialsProviderTest.java
Quanzzzz: Thanks 😁

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/ProfileCredentialsUtils.java
```diff
@@ -116,6 +119,11 @@ public ProfileCredentialsUtils(Profile profile, Function<String, Optional<Profil
             }
         }
 
+        if (properties.containsKey(ProfileProperty.SSO_ROLE_NAME) || properties.containsKey(ProfileProperty.SSO_ACCOUNT_ID)
+            || properties.containsKey(ProfileProperty.SSO_REGION) || properties.containsKey(ProfileProperty.SSO_START_URL)) {
+            return Optional.ofNullable(ssoProfileCredentialsProvider());
+        }
```
millems, line 125: Does this match the rules in the CLI? (After role_arn, before session_token, skip SSO if any of the 4 required fields aren't set, instead of raising an error?)

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/ProfileCredentialsUtils.java
```diff
@@ -262,4 +294,20 @@ private ChildProfileCredentialsProviderFactory stsCredentialsProviderFactory() {
...
         }
     }
+
+    /**
+     * Load the factory that can be used to create the SSO credentials provider, assuming it is on the classpath.
+     */
+    private SsoCredentialsProviderFactory ssoCredentialsProviderFactory() {
+        try {
+            Class<?> ssoCredentialsProviderFactory = Class.forName(SSO_PROFILE_CREDENTIALS_PROVIDER_FACTORY, true,
+                                                                   Thread.currentThread().getContextClassLoader());
```
millems, line 304: You should check with John to see if this line is correct, since he's changing this line for STS.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/SsoCredentialsProfileProperties.java
```diff
@@ -0,0 +1,87 @@
...
+
+package software.amazon.awssdk.auth.credentials.internal;
+
+import software.amazon.awssdk.annotations.SdkInternalApi;
+
+/**
+ * A container for SSO credential properties, these properties can be loaded from the profile in default.
+ */
+@SdkInternalApi
+public class SsoCredentialsProfileProperties {
```
millems, line 24: Isn't this used by SSO, so it should be protected?

services/sso/src/main/java/software.amazon.awssdk.services.sso/auth/SsoCredentialsProvider.java
```diff
@@ -0,0 +1,252 @@
...
+ * updating session credentials. The credential provider loads configuration from a {@link ProfileFile}, and use this information
+ * to generate a cached token directory. With this directory the information a {@link SsoClient} can do a
+ * {@link SsoClient#getRoleCredentials(Consumer)} call to retrieve the credentials needed.
+ *
+ * Currently, the cached token is assumed unexpired, and if it's expired then an {@link ExpiredTokenException} will be thrown. If
+ * the users want to change the behavior of this, please implement your own cached token resolving logic and override the
+ * {@link Builder#refreshRequest).
+ *
+ * When credentials get close to expiration, this class will attempt to update them asynchronously. If the credentials
+ * end up expiring, this class will block all calls to {@link #resolveCredentials()} until the credentials can be updated.
```
millems, line 51: Is this still accurate? E.g. with regard loading from a profile file.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/SessionCredentialsHolder.java
```diff
@@ -0,0 +1,44 @@
...
+import java.time.Instant;
+import software.amazon.awssdk.annotations.SdkProtectedApi;
+import software.amazon.awssdk.annotations.ThreadSafe;
+
+/**
+ * Holder class used to atomically store a session with its expiration time.
+ */
+@SdkProtectedApi
+@ThreadSafe
+public final class SessionCredentialsHolder {
```
millems, line 27: I know it hurts to duplicate this, but I'd vote for copying it to the two places it's needed instead of sharing it, and making it an API we have to maintain backwards-compatibility on.

Maybe an alternative would be to make AwsSessionCredentials have an optional expiration time? Not sure if that's better or worse...

services/sso/src/main/java/software.amazon.awssdk.services.sso/auth/SsoCredentialsProvider.java
```diff
@@ -0,0 +1,252 @@
...
+        return staleTime;
+    }
+
+    /**
+     * The amount of time, relative to session token expiration, that the cached credentials are considered close to stale
+     * and should be updated.
+     */
+    public Duration prefetchTime() {
+        return prefetchTime;
+    }
```
millems, line 125: Do these need to be public?

services/sso/src/main/java/software.amazon.awssdk.services.sso/internal/ExpiredTokenException.java
```diff
@@ -0,0 +1,87 @@
...
+import software.amazon.awssdk.core.SdkPojo;
+import software.amazon.awssdk.core.exception.SdkClientException;
+
+/**
+ * <p>
+ * The session token that was passed is expired or is not valid.
+ * </p>
+ */
+@SdkInternalApi
+public class ExpiredTokenException extends SdkClientException {
```
millems, line 32: If customers will see this, it'll have to be a public API. Should we just use a generic SdkClientException, or do you think customers will want to catch this exception specifically and do something about it?

services/sso/src/main/java/software.amazon.awssdk.services.sso/internal/SsoCachedAccessTokenProvider.java
```diff
@@ -0,0 +1,70 @@
...
+            throw ExpiredTokenException.builder().message("The SSO session associated with this profile has expired or is"
+                                                          + " otherwise invalid. To refresh this SSO session run aws sso"
+                                                          + " login with the corresponding profile.").build();
+        }
+
+        return jsonNode.get("accessToken").asText();
+    }
+
+    private boolean validateToken(String expirationTime) {
+        return Instant.now().isAfter(Instant.parse(expirationTime).minus(15, MINUTES));
```
millems, line 67: Is this 15 minute non-grace-period in the SEP?

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/ProfileCredentialsUtils.java
```diff
@@ -49,6 +50,8 @@
 public final class ProfileCredentialsUtils {
     private static final String STS_PROFILE_CREDENTIALS_PROVIDER_FACTORY =
         "software.amazon.awssdk.services.sts.internal.StsProfileCredentialsProviderFactory";
+    private static final String SSO_PROFILE_CREDENTIALS_PROVIDER_FACTORY =
+        "software.amazon.awssdk.services.sso.internal.SsoCredentialsProviderFactory";
```
millems, line 54: I don't actually see this class in this PR. Am I blind?

services/sso/src/main/java/software.amazon.awssdk.services.sso/internal/SsoProfileCredentialsProviderFactory.java
```diff
@@ -0,0 +1,135 @@
...
+import software.amazon.awssdk.services.sso.model.GetRoleCredentialsRequest;
+import software.amazon.awssdk.utils.IoUtils;
+import software.amazon.awssdk.utils.SdkAutoCloseable;
+
+/**
+ * An implementation of {@link SsoCredentialsProviderFactory} that allows users to get SSO role credentials using the startUrl
+ * sprcified in either a {@link Profile} or environment variables.
+ */
+@SdkInternalApi
+public class SsoProfileCredentialsProviderFactory implements SsoCredentialsProviderFactory {
```
millems, line 43: Is this what's supposed to be in that string I commented on earlier? If so, should this be a protected API, since its constructor is invoked by the auth module?

services/sso/src/main/java/software.amazon.awssdk.services.sso/internal/SsoProfileCredentialsProviderFactory.java
```diff
@@ -0,0 +1,135 @@
...
+    private SsoCachedAccessTokenProvider tokenProvider;
+
+    /**
+     * Default method to create the {@link SsoProfileCredentialsProvider} with a {@link SsoCachedAccessTokenProvider}
+     * object created with the start url from {@link Profile} or environment variables and the default token file directory.
+     */
+    public AwsCredentialsProvider create(SsoCredentialsProfileProperties ssoCredentialsProfileProperties) {
+        this.tokenProvider = new SsoCachedAccessTokenProvider(
+            generateCachedTokenPath(ssoCredentialsProfileProperties.ssoStartUrl(), TOKEN_DIRECTORY));
+        return new SsoProfileCredentialsProvider(ssoCredentialsProfileProperties, tokenProvider);
```
millems, line 56: Nit: can this just use the other create() method?

services/sso/src/main/java/software.amazon.awssdk.services.sso/internal/SsoProfileCredentialsProviderFactory.java
```diff
@@ -0,0 +1,135 @@
...
+/**
+ * An implementation of {@link SsoCredentialsProviderFactory} that allows users to get SSO role credentials using the startUrl
+ * sprcified in either a {@link Profile} or environment variables.
+ */
+@SdkInternalApi
+public class SsoProfileCredentialsProviderFactory implements SsoCredentialsProviderFactory {
+
+    private static final String TOKEN_DIRECTORY = Paths.get(userHomeDirectory(), ".aws", "sso", "cache").toString();
+
+    private SsoCachedAccessTokenProvider tokenProvider;
```
millems, line 47: Is this field needed?

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/SsoCredentialsProfileProperties.java
millems, line 24: Isn't this also a protected API?

millems, line 24: Should this be named with "Profile" since it's not necessarily related to a profile file? I'd honestly rather just pass a `Profile` to a `ProfileCredentialsProviderFactory` instead of a `SsoCredentialsProfileProperties` to a `SsoCredentialsProviderFactory`, and not make this specific to SSO. If we had done that for the web identity token file, we wouldn't have to create yet another protected API here.

services/sso/src/main/java/software.amazon.awssdk.services.sso/internal/SsoProfileCredentialsProviderFactory.java
```diff
@@ -0,0 +1,135 @@
...
+
+        @Override
+        public void close() {
+            IoUtils.closeQuietly(credentialsProvider, null);
+            IoUtils.closeQuietly(ssoClient, null);
+        }
+    }
+
+
+    private static final class GetRoleCredentialsRequestSupplier implements Supplier {
```
millems, line 118: Unqualified generics. Supplier of what?

services/sso/src/main/java/software.amazon.awssdk.services.sso/internal/SsoProfileCredentialsProviderFactory.java
```diff
@@ -0,0 +1,135 @@
...
+        GetRoleCredentialsRequestSupplier(GetRoleCredentialsRequest request,
+                                          String cachedToken) {
+            this.request = request;
+            this.cachedToken = cachedToken;
+        }
+
+        @Override
+        public Object get() {
+            return request.toBuilder().accessToken(cachedToken).build();
+        }
```
millems, line 131: This looks constant. Do we need to even use a supplier here? What if the token expires?

utils/src/test/java/software/amazon/awssdk/testutils/EnvironmentVariableHelper.java
```diff
@@ -0,0 +1,146 @@
...
+import java.util.Optional;
+import java.util.function.Consumer;
+import org.junit.rules.ExternalResource;
+import software.amazon.awssdk.utils.SystemSetting;
+
+/**
+ * A utility that can temporarily forcibly set environment variables and
+ * then allows resetting them to the original values.
+ */
+public class EnvironmentVariableHelper extends ExternalResource {
```
millems, line 33: Don't we already have this elsewhere?

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/ProfileCredentialsUtils.java
kiiadi: out of interest why are you doing custom class loading?

millems: What's the alternative? SSO isn't necessarily on the classpath.

utils/src/main/java/software/amazon/awssdk/utils/UserHomeDirectoryUtils.java
```diff
@@ -0,0 +1,62 @@
...
+ */
+@SdkProtectedApi
+public final class UserHomeDirectoryUtils {
+
+    private UserHomeDirectoryUtils() {
+
+    }
+
+    public static String userHomeDirectory() {
+        boolean isWindows = JavaSystemSetting.OS_NAME.getStringValue()
```
kiiadi, line 32: nit: you could probably move this down since it's not used if `HOME` is set.

services/sso/src/main/java/software.amazon.awssdk.services.sso/auth/SsoCredentialsProvider.java
Quanzzzz: Generally, we still need to load the `accountId` and `roleName` from the profile file, but that's for generating the request, not the cached token directory. I can update that.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/SessionCredentialsHolder.java
Quanzzzz: Can I know the reason why we don't want to share this between different credentials providers?

services/sso/src/main/java/software.amazon.awssdk.services.sso/auth/SsoCredentialsProvider.java
Quanzzzz: Not really for now, I made them public for consistency with the existed STSCredentialsProvider, I can change it.

services/sso/src/main/java/software.amazon.awssdk.services.sso/internal/ExpiredTokenException.java
Quanzzzz: It is mentioned in the SEP that this exception will be thrown. I guess it's safer for us to use the same exception as the SEP so the users wouldn't get confused? I can change it to public though.

services/sso/src/main/java/software.amazon.awssdk.services.sso/internal/SsoCachedAccessTokenProvider.java
Quanzzzz: Yes, this is on the SEP: `When checking to see if a cached token is expired, the SDK MAY treat the token as expired if it will expire within 15 minutes of the expiresAt time.`

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/ProfileCredentialsUtils.java
Quanzzzz: My bad, the name of this class is changed, it should be `SsoProfileCredentialsProviderFactory` now.

services/sso/src/main/java/software.amazon.awssdk.services.sso/internal/SsoProfileCredentialsProviderFactory.java
Quanzzzz: Agree, should I move it out of the internal package?

services/sso/src/main/java/software.amazon.awssdk.services.sso/internal/SsoProfileCredentialsProviderFactory.java
Quanzzzz: Not really, it was useful before and not any more after I did multiple changes. I can remove it.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/SsoCredentialsProfileProperties.java
Quanzzzz: It's not necessary to include "Profile" in the name but I can't come up with a better name. Do you have any good idea about how to name this? I named it with "Profile" because it takes the configurations from the profile file and those properties are all related to the SSO credentials provider.

utils/src/main/java/software/amazon/awssdk/utils/UserHomeDirectoryUtils.java
Quanzzzz: Good call, thanks for reminding me about this.

utils/src/test/java/software/amazon/awssdk/testutils/EnvironmentVariableHelper.java
Quanzzzz: Yea, we already have it in the `test-utils` module, I tried moving it from the `test-utils` into the `utils` module, but there are some other places using this class having the dependency on `test-utils` but not on `utils`, and that caused some issue. So I just copied one here. Should we move it to `utils` and correct all of those usages?

services/sso/src/main/java/software.amazon.awssdk.services.sso/internal/SsoProfileCredentialsProviderFactory.java
Quanzzzz: This is a debatable part. I used to put the check of whether the token expires here in this supplier, but that led to the lacking of flexibility because customers wouldn't be able to customize their own ways of resolving the access token from the cached token file. However, in my current implmenetation, I took the whole resolving access token part out of the factory(including validating the token file), and found out that this supplier became constant and it doesn't really change with the token. Could you point me out to a better way to handle both of these two here?

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/ProfileCredentialsUtils.java
Quanzzzz: This rule isn't specified in the SEP, but there is a line saying: 
`If the SSO credential provider is added to a default chain it MUST be added immediately after the WebIdentityProvider and before all other providers.` 
Even though this isn't the default providers chain, it should still follow the similar rule so I put the SSO provider here.

About skipping SSO when none of the four fields are set, there is the corresponding rule in SEP: 
`If one or more of the above configuration values is present the profile MUST be resolved by the SSO credential provider. All of the above configuration values MUST be set in the profile for the profile to be treated as a valid profile by the SSO credential provider. In the case that one or more of the above configuration values is missing from the profile the SSO credential provider MUST raise an exception that the profile is invalid.`
From my understanding, if there is one or more of the four properties is present then we need to check if all of the four are all present, and if not then we throw out an exception. However, if none of them are here then at least it means the customer isn't wanting the SSO provider so we shouldn't raise the error.

millems: Should this be above the `role_arn` logic, then?

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/SessionCredentialsHolder.java
millems: Our biggest source of technical debt in the SDK are APIs that we can't change: protected and public APIs. We try not to add new protected or public APIs unless we have a strong reason to do so, because it means we can pretty much never change our mind in the future. We can always refactor internal APIs if we want to in the future.

This seems like a protected API that we might regret adding. For example, what happens if we want to add the expiration time to AwsSessionCredentials in the future? Then, SessionCredentialsHolder is no longer needed, but we can't delete it because it's protected.

services/sso/src/main/java/software.amazon.awssdk.services.sso/auth/SsoCredentialsProvider.java
millems: Ah, if StsCredentialsProvider have them public I guess that's fine.

services/sso/src/main/java/software.amazon.awssdk.services.sso/internal/ExpiredTokenException.java
millems: I suppose someone might want to "catch" ExpiredTokenException and update their credentials. I guess it's fine to add a new public class here, then.

services/sso/src/main/java/software.amazon.awssdk.services.sso/internal/SsoCachedAccessTokenProvider.java
millems: Alright, then. I guess we can relax it if customers complain. 15 minutes seems like a lot. Does the CLI do this?

services/sso/src/main/java/software.amazon.awssdk.services.sso/internal/SsoProfileCredentialsProviderFactory.java
millems: Yeah, if it's protected it shouldn't be in the internal package.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/SsoCredentialsProfileProperties.java
millems: What do you think about my alternate suggestion - making it more generic and just passing the profile through?

services/sso/src/main/java/software.amazon.awssdk.services.sso/internal/SsoProfileCredentialsProviderFactory.java
```diff
@@ -0,0 +1,135 @@
...
+        return new SsoProfileCredentialsProvider(ssoCredentialsProfileProperties, tokenProvider);
+    }
+
+    /**
+     * Alternative method to create the {@link SsoProfileCredentialsProvider} with a customized
+     * {@link SsoCachedAccessTokenProvider}. This method is only used for testing.
+     */
+    @SdkTestInternalApi
+    public AwsCredentialsProvider create(SsoCredentialsProfileProperties ssoCredentialsProfileProperties,
+                                         SsoCachedAccessTokenProvider tokenProvider) {
```
millems, line 65: Can we rename SsoCachedAccessTokenProvider to SsoAccessTokenProvider? The fact that it works on a cache seems like an implementation detail.

services/sso/src/main/java/software.amazon.awssdk.services.sso/internal/SsoProfileCredentialsProviderFactory.java
millems: If you need to be able to reload the token in the credentials provider after it expires, you'll have to query the SsoAccessTokenProvider in this supplier to get the access token.

utils/src/test/java/software/amazon/awssdk/testutils/EnvironmentVariableHelper.java
millems: Why does this need to be in the utils module?

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/SessionCredentialsHolder.java
Quanzzzz: Got you, will do the change.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/SsoCredentialsProfileProperties.java
Quanzzzz: I understand your point of making the factory generic so we don't need to add more and more factories as protected apis here. But I think our purpose of using `ProfileCredentialsUtils` to create different credentials providers from the profile properties is just to make the reading properties part generic. If we pass a `Profile` directly to the `ProfileCredentialsProviderFactory`, then we would need to add a series of reading properties and validating properties stuffs in the factory itself. Isn't that conflicting with our original design? If you do think creating less protected apis is worthy then do you think we should make it independent from the current `ProfileCredentialsUtils`?

utils/src/test/java/software/amazon/awssdk/testutils/EnvironmentVariableHelper.java
Quanzzzz: Because we have the `UserHomeDirectoryUtilsTest` using it in the `utils` module. And as we discussed, it's better for us to extract `userHomeDirectory()` method out of `ProfileFileLocation`, thus I moved it into `UserHomeDirectoryUtils`, as well as the `UserHomeDirectoryUtilsTest`.

services/sso/src/main/java/software.amazon.awssdk.services.sso/internal/SsoProfileCredentialsProviderFactory.java
Quanzzzz: Since we are going to put the logic of resolving access token back into the inner class of the factory, we won't need this token provider any more.

Quanzzzz: Will do the change.

services/sso/src/main/java/software.amazon.awssdk.services.sso/internal/SsoProfileCredentialsProviderFactory.java
Quanzzzz: This change is done.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/ProfileCredentialsUtils.java
Quanzzzz: I checked with CLI team, they confirmed that they have the same order as us (SSO is after webIdentity).
```
def providers(self, profile_name, disable_env_vars=False):
        return [
            self._create_web_identity_provider(
                profile_name, disable_env_vars,
            ),
            self._create_sso_provider(profile_name),
            self._create_shared_credential_provider(profile_name),
            self._create_process_provider(profile_name),
            self._create_config_provider(profile_name),
        ]
```

services/sso/src/main/java/software.amazon.awssdk.services.sso/internal/SsoCachedAccessTokenProvider.java
Quanzzzz: Yes, confirmed by CLI team.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/SsoCredentialsProfileProperties.java
millems: To make sure we're on the same page, I was picturing just changing the current `SsoCredentialsProviderFactory` to `ProfileCredentialsProviderFactory`, and then we'd have an `SsoProfileCredentialsProviderFactory` that implements that interface. The responsibilities of the calling class and the factory itself would remain the same.

The `SsoProfileCredentialsProviderFactory` would indeed read the properties from the profile, but that's just calling `profile.property(foo)` instead of `ssoCredentialsProfileProperties.foo()`. Would there have to be more than that?

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/ProfileCredentialsUtils.java
millems: Which step do they check the `role_arn`? Right now, it looks like we're prioritizing `role_arn` higher than `sso`. I think the CLI is prioritizing it lower.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/SsoCredentialsProfileProperties.java
Quanzzzz: I see, sorry I misunderstood your suggestion, will do the change.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/ProfileCredentialsUtils.java
Quanzzzz: I think `role_arn` is only used while creating the web identity credentials provider so it should be checked at the very first step. But I can check again to confirm this more accurately.

millems: I guess I should be asking about STS assume-role credentials, not role_arn. I think the CLI treats STS assume-role credentials lower than SSO credentials, but this code treats it higher.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/ProfileCredentialsUtils.java
```diff
@@ -46,10 +47,12 @@
 /**
  * Utility class to load {@link #credentialsProvider()} configured in a profile.
  */
-@SdkInternalApi
+@SdkProtectedApi
```
millems, line 50: Why is this protected?

services/sso/src/main/java/software.amazon.awssdk.services.sso/auth/ExpiredTokenException.java
```diff
@@ -0,0 +1,87 @@
...
+        return new BuilderImpl();
+    }
+
+    public interface Builder extends SdkPojo, SdkClientException.Builder {
+        @Override
+        Builder message(String message);
+
+        @Override
+        Builder cause(Throwable cause);
+    }
```
millems, line 55: Can we override `build()` to return `ExpiredTokenException`? Customers might want to get hold of an actual instance of ExpiredTokenException, and without the different `build` return type they'd need to cast,

services/sso/src/main/java/software.amazon.awssdk.services.sso/auth/SsoProfileCredentialsProviderFactory.java
```diff
@@ -0,0 +1,130 @@
...
+                                      .region(Region.of(ssoRegion))
+                                      .build();
+
+            GetRoleCredentialsRequest request = GetRoleCredentialsRequest.builder()
+                                                                         .accountId(ssoAccountId)
+                                                                         .roleName(ssoRoleName)
+                                                                         .build();
+
+            GetRoleCredentialsRequestSupplier supplier = new GetRoleCredentialsRequestSupplier(request,
+                                                                                               provider);
```
millems, line 92: You could simplify this to

```
Supplier<GetRoleCredentialsRequest> supplier = () -> request.toBuilder().accessToken(tokenProvider.resolveAccessToken()).build();
```

(or even use the copy() method instead of toBuilder() and build() if you're feeling extra frisky).

services/sso/src/main/java/software.amazon.awssdk.services.sso/auth/SsoCredentialsProvider.java
```diff
@@ -0,0 +1,250 @@
...
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.services.sso.auth;
```
millems, line 16: It looks like the directory name is `software.amazon.awssdk.services.sso` instead of having a chain of parent directories: `software/amazon/awssdk/services/sso`. Is that just github being weird, or is that accurate?

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/ProfileCredentialsUtils.java
Quanzzzz: That was a mistake, I fixed it once... but I guess the fix wasn't included... will update it again.

services/sso/src/main/java/software.amazon.awssdk.services.sso/auth/ExpiredTokenException.java
Quanzzzz: Sure, will do the change.

services/sso/src/main/java/software.amazon.awssdk.services.sso/auth/SsoProfileCredentialsProviderFactory.java
Quanzzzz: Sounds good, will do the change.

services/sso/src/main/java/software.amazon.awssdk.services.sso/auth/SsoCredentialsProvider.java
Quanzzzz: It might be IntelliJ acting weird, I will recreate the directory to make it accurate.

services/sso/src/main/java/software.amazon.awssdk.services.sso/auth/ExpiredTokenException.java
```diff
@@ -0,0 +1,87 @@
...
+import software.amazon.awssdk.core.SdkPojo;
+import software.amazon.awssdk.core.exception.SdkClientException;
+
+/**
+ * <p>
+ * The session token that was passed is expired or is not valid.
+ * </p>
+ */
+@SdkPublicApi
+public class ExpiredTokenException extends SdkClientException {
```
zoewangg, line 32: `final`?

services/sso/src/main/java/software.amazon.awssdk.services.sso/auth/SsoCredentialsProvider.java
```diff
@@ -0,0 +1,250 @@
...
+import software.amazon.awssdk.utils.SdkAutoCloseable;
+import software.amazon.awssdk.utils.cache.CachedSupplier;
+import software.amazon.awssdk.utils.cache.NonBlocking;
+import software.amazon.awssdk.utils.cache.RefreshResult;
+
+/**
+ * An implementation of {@link AwsCredentialsProvider} that is extended within this package to provide support for periodically
+ * updating session credentials. This credential provider maintains a {@link Supplier<GetRoleCredentialsRequest>} for a
+ * {@link SsoClient#getRoleCredentials(Consumer)} call to retrieve the credentials needed.
+ *
```
zoewangg, line 42: nit: missing `<p>`

services/sso/src/main/java/software.amazon.awssdk.services.sso/auth/SsoCredentialsProvider.java
```diff
@@ -0,0 +1,250 @@
...
+ * When credentials get close to expiration, this class will attempt to update them asynchronously. If the credentials
+ * end up expiring, this class will block all calls to {@link #resolveCredentials()} until the credentials can be updated.
+ */
+@SdkPublicApi
+public final class SsoCredentialsProvider implements AwsCredentialsProvider, SdkAutoCloseable {
+
+    private static final Duration DEFAULT_STALE_TIME = Duration.ofMinutes(1);
+    private static final Duration DEFAULT_PREFETCH_TIME = Duration.ofMinutes(5);
+
+    private static final String ASYNC_THREAD_NAME = "sso-credentials-provider";
```
zoewangg, line 57: Would it make sense to prefix it with `sdk-`? It'd be easier to filter all SDK threads in a thread dump.

Quanzzzz: Sure, good call

services/sso/src/main/java/software.amazon.awssdk.services.sso/auth/SsoCredentialsProvider.java
```diff
@@ -0,0 +1,250 @@
...
+
+        /**
+         * Create a {@link SsoCredentialsProvider} using the configuration applied to this builder.
+         * @return
+         */
+        SsoCredentialsProvider build();
+
+    }
+
+    public static final class BuilderImpl implements Builder {
```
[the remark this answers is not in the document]
Quanzzzz: I guess it can be `protected` here, as the factory is actually calling the methods of the BuilderImpl.

millems (approved):

Issue #1991, opened 2020-08-17
Enum deserialization is suspending the process forever when the enum value is not existent in the enum class.


eheinen:
When there is a data on DynamoDB with Enum value that doesn't exist on the class model that is gonna receive the deserialization, the execution process hangs forever.

## Expected Behavior
An exception when deserializing an enum value that doesn't exist.

## Steps to Reproduce
- A document with a String `partitionKey`, an Enum with values: `A` and `B` and an LSI Instant `updatedAt`.
- Make an index query with `DynamoDbAsyncIndex` using `QueryConditional.sortBetween` like:

```
       val queryConditional = QueryConditional.sortBetween(
            Key.builder()
                .partitionValue(partitionKey)
                .sortValue(updatedAtFrom.toString())
                .build(),
            Key.builder()
                .partitionValue(partitionKey)
                .sortValue(updatedAtTo.toString())
                .build()
        )

       myTableIndex.query(
                QueryEnhancedRequest
                    .builder()
                    .queryConditional(queryConditional)
                    .build())
```

# Issue
The problem is in the deserialization of the enum class that **doesn't have** a value that is signed to a data on DynamoDB.
For this reason, the process gets running forever.

## Your Environment
  - Kotlin:  1.3.72
  - Spring Boot: 2.3.1.RELEASE
  - Webflux
  - dynamodb-enhanced: 2.13.35
 - Netty: 4.1.50.Final

In my DynamoDB environment, there was a unique record with an old state called INITIALIZED that was replaced for CREATED some days ago.
When the query runs, the data cannot be deserialized in this enum and finally hangs the process forever without any exception or log.

eheinen:
Hey guys, I have found the issue. It is a problem to deserialize an enum.
It is basically really weird, my document models are like:

```
@DynamoDbBean
data class A {
   @get:DynamoDbSortKey
    var id: String = "",

    @get:DynamoDbPartitionKey
    var secondId: String = "",

    var states: List<B> = emptyList(),

    @get:DynamoDbSecondarySortKey(indexNames = ["updatedAt_index"])
    var updatedAt: Instant? = null,
}

@DynamoDbBean
data class DeliveryStateDocument(

    var state: State? = null
)

enum class State {
     CREATED,
     DESTROYED
}
```

In my DynamoDB environment, there was a unique record with an old state INITIALIZED that was replaced for CREATED.
[...]

PR #2121, opened 2020-10-27
Fix for handling transormation error in async DynamoDb Enhanced Clien…

bmaizels:
…t operations that return an SdkPublisher

Fixes https://github.com/aws/aws-sdk-java-v2/issues/1991


.changes/next-release/bugfix-AWSDynamoDBEnhancedClient-f9d7822.json
```diff
@@ -0,0 +1,6 @@
+{
+    "type": "bugfix",
+    "category": "AWS DynamoDB Enhanced Client",
+    "contributor": "",
+    "description": "Publisher streams returned by async resources in the DynamoDB Enhanced Client now correctly handle mapping errors when they are encountered in the stream by calling onError on the subscriber and then implicitly cancelling the subscription. Previously the stream would just permanently hang and never complete."
```
dagnir, line 5: Just occurred to me, we should add a note about the fix in `SdkPublisher#map()` behavior under `AWS SDK for Java V2`

utils/src/main/java/software/amazon/awssdk/utils/internal/MappingSubscriber.java
```diff
@@ -0,0 +1,93 @@
...
+    public static <T, U> MappingSubscriber<T, U> create(Subscriber<? super U> subscriber,
+                                                        Function<T, U> mapFunction) {
+        return new MappingSubscriber<>(subscriber, mapFunction);
+    }
+
+    @Override
+    public void onSubscribe(Subscription subscription) {
+        this.subscriptions.add(subscription);
+        delegateSubscriber.onSubscribe(subscription);
+    }
```
dagnir, line 53: I'm not sure about this; any reason we're supporting multiple subscriptions? Per my reading of the spec, a `Subscriber` can only have one `Subscription`. Per [2.12](https://github.com/reactive-streams/reactive-streams-jvm/tree/v1.0.3#2.12):

> Subscriber.onSubscribe MUST be called at most once for a given Subscriber (based on object equality).

dagnir (approved):
:shipit:

PR #2126, opened 2020-10-29
Added a customization to allow underscores in shape names.

millems:
1. Throw an exception if an invalid shape name is used instead of trying to transform it. This should be backwards-compatible, because our transformation is currently broken.
2. Throw an exception if a customer-visible identifier contains underscores. Added customizations to rename any existing shapes that violate this rule.
3. Add validation to the code generator to verify that certain names will actually compile. This provides a more useful error message when services contain invalid identifiers.
4. Deprecate SsoOidc's "error_description" getters for ones with idiomatic names.


services/mediapackagevod/src/main/resources/codegen-resources/customization.config
```diff
@@ -0,0 +1,6 @@
+{
+  "renameShapes": {
+    // Do not keep adding to this list. Require the service team to name enums like they're naming their shapes.
+    "__PeriodTriggersElement": "PeriodTriggersElement"
```
zoewangg, line 4: Is this breaking change?

millems: No, the existing __s were being stripped. My change no longer strips, so I did the rename manually.

zoewangg (approved):

PR #2136, opened 2020-11-05
Increased maximum threads allowed by stability tests.

millems:
This maximum was being exceeded by an expected set of threads. Also named the threads created by `FileAsyncRequestBody` and `FileAsyncResponseTransformer` for more easy identification.

### DIFF
--- a/test/stability-tests/src/it/java/software/amazon/awssdk/stability/tests/utils/StabilityTestRunner.java
+++ b/test/stability-tests/src/it/java/software/amazon/awssdk/stability/tests/utils/StabilityTestRunner.java
@@ -74,7 +74,7 @@ public class StabilityTestRunner {
     private static final int TESTS_TIMEOUT_IN_MINUTES = 60;
     // The peak thread count might be different depending on the machine the tests are currently running on.
     // because of the internal thread pool used in AsynchronousFileChannel
-    private static final int ALLOWED_PEAK_THREAD_COUNT = 60;
+    private static final int ALLOWED_PEAK_THREAD_COUNT = 90;
 
     private ThreadMXBean threadMXBean;
     private IntFunction<CompletableFuture<?>> futureFactory;


core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/FileAsyncRequestBody.java
```diff
@@ -293,6 +298,12 @@ private void signalOnError(Throwable t) {
     }
 
     private static AsynchronousFileChannel openInputChannel(Path path) throws IOException {
-        return AsynchronousFileChannel.open(path, StandardOpenOption.READ);
+        return AsynchronousFileChannel.open(path, Collections.singleton(StandardOpenOption.READ), EXECUTOR);
+    }
+
+    private static ExecutorService executor() {
```
zoewangg, line 304: Could this be a breaking change? Because the default thread pool used is https://docs.oracle.com/javase/8/docs/api/java/nio/channels/AsynchronousChannelGroup.html and customers can configure it via system properties.

Should we close the executor when the SDK client is closed? My concern is that customers might find it surprising that after the client is closed, there are still some sdk threads lingering.

millems: 1. I guess someone might consider this to be breaking. It certainly wouldn't affect the majority of users, but I suppose it could affect someone. I'm willing to drop naming these threads from this change, since it was just for my ease of debugging and no customers have complained about it.
2. The threads will go away after 60 seconds, so they won't linger too long. That's also the current behavior.
3. A async request body is a separate abstraction from the client, so it can't be linked to the lifecycle of a client.

zoewangg (approved):

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/FileAsyncRequestBody.java
zoewangg: Yeah, +1 it's super difficult to debug right now. It'd also be nice if we could provide a configuration that allows customers to provide their own executors.

Issue #2123, opened 2020-10-28
STS not on class path when using parallel stream


BartXZX:
Running _EcrClient.listImagesPaginator()_ in a _Collection.parallelStream()_ using Spring Boot sometimes results in _WebIdentityCredentialsUtils.factory()_ not finding STS on its class path.

## Describe the issue
_Sometimes_ when I execute the above I get the error message: "To use web identity tokens, the 'sts' service module must be on the class path.". I know that parallel streams delegates work to the ForkJoin common pool, and it seems that sometimes _some_ threads from this pool might have a class loader that does not have STS on its path. (Maybe because of Spring? I noticed that the loader called "app" does not have STS)

Is there a reason you guys are using _Thread.currentThread().getContextClassLoader()_, instead of _Object.getClass().getClassLoader()_ in the WebIdentityCredentialsUtils class?

I admit I know little about class loaders and how you use them, but I cannot use parallelStream() reliably at the moment. Or do you advise against using parallelStream at all?

Not filing as bug, since it might be intended behaviour.

## Steps to Reproduce
I am listing docker images from multiple repositories in parallel.
```java
@Override
public Stream<String> listImages(List<String> repositories) {
    return repositories.parallelStream()
        .flatMap(repoName -> ecr.listImagesPaginator(builder -> builder.registryId(this.registryId).repositoryName(repoName).filter(f -> f.tagStatus(TagStatus.TAGGED)))
            .imageIds()
            .stream()
            .map(ImageIdentifier::imageTag));
}
```

## Current Behavior
I get the message "To use web identity tokens, the 'sts' service module must be on the class path." instead of it finding STS and using the token credentials.

## Your Environment

* AWS Java SDK version used: 2.14.18
* JDK version used: 13
* Operating System and version: Using the openjdk:13 image, which uses Oracle Linux 7 I believe.

zoewangg:
Thank you for reporting the issue! I think we should use class loader instead of the thread context loader. We already have the `ClassLoaderHelper#classLoader` helper class. Marking this as a bug

joviegas:
Hi @BartXZX 
Could you please help me to confirm ones, if you have added "sts" module in dependency like

```
        <dependency>
            <groupId>software.amazon.awssdk</groupId>
            <artifactId>sts</artifactId>
            <version>2.15.20</version>
        </dependency>
```

or it would be of great help if you could provide all dependencies related to <groupId>software.amazon.awssdk</groupId>

[...]

PR #2139, opened 2020-11-07
PR#2123 STS not on class path when using parallel stream

joviegas:
## Description

STS not on class path when using parallel stream

## Motivation and Context

Please check for the context in [PR#2123](https://github.com/aws/aws-sdk-java-v2/issues/2123)
Fixing this issue by changing the Classloader to use ClassLoaderHelper.

Reference material :- https://stackoverflow.com/questions/1771679/difference-between-threads-context-class-loader-and-normal-classloader

## Testing

- Added JUNit
- sudo mvn clean install -pl :sts -P quick --am
- sudo mvn clean install -pl :core -P quick --am

## Types of changes

## Checklist

## License


build-tools/src/main/resources/software/amazon/awssdk/checkstyle.xml
```diff
@@ -387,6 +387,11 @@
 
         <!-- Checks that no 'final' are added on local variables -->
         <module name="software.amazon.awssdk.buildtools.checkstyle.UnnecessaryFinalOnLocalVariableCheck" />
+
+        <!-- Checks that no forbidden methods are used in implementation -->
+        <module name="com.github.sevntu.checkstyle.checks.coding.ForbidCertainMethodCheck">
```
zoewangg, line 392: Can we use `Regexp` instead?  That way, we don't need to add another dependency

Example: https://github.com/aws/aws-sdk-java-v2/blob/8f361fde4800451d99dce4a1a22c5557fee61aee/build-tools/src/main/resources/software/amazon/awssdk/checkstyle.xml#L361-L365

services/sts/src/test/java/software/amazon/awssdk/services/sts/internal/StsWebIdentityCredentialsProviderFactoryTest.java
```diff
@@ -0,0 +1,35 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatCode;
+
+public class StsWebIdentityCredentialsProviderFactoryTest {
+
+
+    @Test
+    public void stsWebIdentityCredentialsProviderFactory_with_webIdentityCredentialsUtils(){
+        final WebIdentityTokenCredentialsProviderFactory factory = WebIdentityCredentialsUtils.factory();
```
zoewangg, line 31: nit: we don't use `final` on local variables. We have a checkstyle for it, but it doesn't apply to tests

zoewangg:
Can we add a changelog entry?

build-tools/src/main/resources/software/amazon/awssdk/checkstyle.xml
joviegas: Thanks for the reference. Changes done.

services/sts/src/test/java/software/amazon/awssdk/services/sts/internal/StsWebIdentityCredentialsProviderFactoryTest.java
joviegas: done.

joviegas:
> Can we add a changelog entry?

Done.

zoewangg (approved):

PR #2152, opened 2020-11-19
Adds call to interceptor to disable double URL encoding for S3 contro…

cenedhryn:
…l requests in addition to S3 requests

## Motivation and Context
When there are query parameters containing special characters, the SDK normally double encodes them before signing. S3 does not, so S3 requests have an interceptor that sets a signer attribute to disable double url encoding. S3 control APIs is governed by the same rule, but does not have the interceptor attached. 

Discovered through testing of non-released features.

## Testing
- Functional test to check that any method in S3 control will cause the signer attribute
- Existing integration tests work

## Types of changes

## Checklist

## License


services/s3control/src/main/resources/software/amazon/awssdk/services/s3control/execution.interceptors
```diff
@@ -1,2 +1,3 @@
+software.amazon.awssdk.services.s3.internal.handlers.DisableDoubleUrlEncodingInterceptor
```
millems, line 1: This makes it a protected API, not an internal API. If it's simple, can we just copy it?

cenedhryn: I discussed with Zoe and this was the solution we agreed on yesterday. You have a valid argument, but also S3 doesn't necessarily distinguish between S3 data plane and control plane API calls in terms on how they receive and process requests. Also we already have a dependency on S3. 

However I think I can be persuaded to copy it since it IS very small and dependencies are potentially expensive in the future...

millems: I think if we're going to share it and make it protected we should move it to a place that we can get to it without needing to go through the S3 module. Maybe in the `auth` module, since that's where it's used? We'd want to pick a more specific name as well: `DisableDoubleUrlEncodingDuringSigningInterceptor`?

I'm still on "team just copy it", but you and Zoe outnumber me :).

millems (approved):

PR #2179, opened 2020-12-03
Add amz-sdk-request header, removed amz-sdk-retry header.

millems:
Most changes are from a rename of MockHttpClient to MockSyncHttpClient so that a common parent interface can be added for generic programming purposes.

The new amz-sdk-request header does not include the TTL, because it's not at all easy to calculate with pluggable HTTP clients and the specification lists it as optional. The TTL can be added when service teams ask for it.


core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/utils/RetryableStageHelper.java
```diff
@@ -140,14 +139,8 @@ public void logBackingOff(Duration backoffDelay) {
...
     public SdkHttpFullRequest requestToSend() {
-        Integer availableRetryCapacity = TokenBucketRetryCondition.getCapacityForExecution(context.executionAttributes())
-                                                                  .map(TokenBucketRetryCondition.Capacity::capacityRemaining)
-                                                                  .orElse(null);
-        String headerValue = (attemptNumber - 1) + "/" +
-                             context.executionAttributes().getAttribute(LAST_BACKOFF_DELAY_DURATION).toMillis() + "/" +
-                             (availableRetryCapacity != null ? availableRetryCapacity : "");
         return request.toBuilder()
-                      .putHeader(SDK_RETRY_INFO_HEADER, headerValue)
+                      .putHeader(SDK_RETRY_INFO_HEADER, "attempt=" + attemptNumber + "; max=" + retryPolicy.numRetries())
```
zoewangg, line 143: I think the retry capacity is very valuable for troubleshooting; otherwise there's no way to tell if the retry gets throttled. I've seen an issue before where all requests failed of `IOException`, and I initially thought all retry attempts failed, but this header made me realized that the request did not get retried because there was no retry capacity.

zoewangg: I guess now that we have metrics, we can see it there. It's just not as obvious

millems: Maybe we could add a debug log whenever we do not retry because of throttling?

core/sdk-core/src/main/java/software/amazon/awssdk/core/retry/conditions/TokenBucketRetryCondition.java
```diff
@@ -122,9 +125,17 @@ public boolean shouldRetry(RetryPolicyContext context) {
             context.executionAttributes().putAttribute(LAST_ACQUIRED_CAPACITY, c);
             context.executionAttributes().putAttribute(RETRY_COUNT_OF_LAST_CAPACITY_ACQUISITION,
                                                        context.retriesAttempted());
+            log.trace(() -> "Successfully acquired token bucket capacity to retry this request. "
```
zoewangg, line 128: `debug`?

test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/retry/RetryHeaderTestSuite.java
```diff
@@ -0,0 +1,126 @@
...
+import org.junit.Test;
+import software.amazon.awssdk.http.HttpExecuteResponse;
+import software.amazon.awssdk.http.SdkHttpRequest;
+import software.amazon.awssdk.http.SdkHttpResponse;
+import software.amazon.awssdk.testutils.service.http.MockHttpClient;
+
+/**
+ * A set of tests that verify the behavior of retry-related headers (amz-sdk-invocation-id and amz-sdk-request).
+ */
+public abstract class RetryHeaderTestSuite<T extends MockHttpClient> {
```
zoewangg, line 32: Nice!

zoewangg (approved):

core/sdk-core/src/main/java/software/amazon/awssdk/core/retry/conditions/TokenBucketRetryCondition.java
millems: This seemed like a bit much for debug, since we're already logging when there wasn't enough capacity. I was thinking if we really needed to debug this, we can ask them to enable trace for this one class.

zoewangg (approved):

PR #2182, opened 2020-12-04
Handle TLS 1.3 bad cert errors in proxy handler

zoewangg:
## Description
Fixed the issue where BAD_CERTIFICATE issue manifested as acquire connection timeout error when using TLS1.3 and proxy.

### Requests w/ proxy

- Before this change

TLS version | BAD_CERTIFICATE error
------------ | -------------
TLS 1.2 | java.io.IOException: Unable to send CONNECT request to proxy
TLS 1.3 | Acquire operation took longer than the configured maximum time

- After this change

TLS version | BAD_CERTIFICATE error
------------ | -------------
TLS 1.2 | java.io.IOException: Unable to send CONNECT request to proxy
TLS 1.3 | java.io.IOException: Unable to send CONNECT request to proxy

###  Requests w/o proxy

- Before this change

TLS version | BAD_CERTIFICATE error
------------ | -------------
TLS 1.2 | The channel was closed...
TLS 1.3 | Server failed to send complete response

- After this change

TLS version | BAD_CERTIFICATE error
------------ | -------------
TLS 1.2 | The channel was closed...
TLS 1.3 | Server failed to send complete response. The channel was closed...
## Testing
Added unit tests.

## Types of changes

## Checklist

## License


http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/utils/NettyUtils.java
```diff
@@ -40,6 +40,14 @@
      */
     public static final SucceededFuture<?> SUCCEEDED_FUTURE = new SucceededFuture<>(null, null);
 
+    public static final String CLOSED_CHANNEL_MESSAGE = "The channel was closed. This may have been done by the client (e.g. "
+                                                        + "because the request was aborted), " +
+                                                        "by the service (e.g. because there was a handshake error[use -Djavax"
+                                                        + ".net.debug=ssl to enable SSL logs], the request "
```
millems, line 46: Is there any way we can detect that it was a handshake error so that they don't have to enable SSL logs? I wouldn't want every customer having to enable SSL logs to see if it was a handshake error just a plain-ol'-closed-channel.

zoewangg: No, unfortunately, this kind of errors are bubbled up as `ClosedChannelException` with no details at all. When I was troubleshooting the failing test, it took me while to figure out it was related to handshake.

millems: I wish we could link to docs that help customers debug these errors. Maybe something we can talk to @Carey-AWS about. In the meantime, it's really hard to read this with the parens-in-parens for the "use -D...". Can we exclude that? The other potential causes we list don't say how to diagnose them.

zoewangg: +1. Yeah, a guide on how to troubleshoot closed channel exception would be super helpful!

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ProxyTunnelInitHandler.java
```diff
@@ -98,6 +96,31 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) {
...
+            handleConnectRequestFailure(ctx, null);
+        }
+    }
+
+    @Override
+    public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) {
+        if (!initPromise.isDone()) {
+            handleConnectRequestFailure(ctx, cause);
+        } else {
+            ctx.fireExceptionCaught(cause);
```
dagnir, line 111: Any reason we propagate this event, when `initPromise` is already done, but not `channelInactive`?

zoewangg: Actually, there's really no need to propagate the event since this is the last handler in the pipeline. I think we should just close the channel context in this case. WDUT?

dagnir: I think in this case we're not likely to add any new handlers to this pipeline, but just to be safe let's just propagate the event if we don't handle it, which I think is the preferred default behavior.

zoewangg: I think we should handle the exception though, otherwise it'll be logged as something like "unhandled exception reached the end of pipeline", which is not a good customer experience imo.

dagnir: Yeah that's what I meant; I think we agree. Let's just propagate the event if we don't call `handleConnectRequestFailure` in `channelInactive`

.changes/next-release/bugfix-NettyNIOHTTPClient-dbb46f2.json
```diff
@@ -0,0 +1,6 @@
+{
+    "category": "Netty NIO HTTP Client", 
+    "contributor": "", 
+    "type": "bugfix", 
+    "description": "Fixed the issue where BAD_CERTIFICATE issue manifested as acquire connection timeout error when using TLS1.3 and proxy."
```
dagnir, line 5: nit: is this issue unique to `BAD_CERTIFICATE`?

dagnir (approved):

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/utils/NettyUtils.java
Carey-AWS: Thanks for calling this out, @millems. I've added it to the backlog.

.changes/next-release/bugfix-NettyNIOHTTPClient-dbb46f2.json
zoewangg: Probably not, I'll update it

PR #2202, opened 2020-12-15
Treat zero-byte responses from HTTP clients as not having content when it comes to creating an SDK HTTP full request.

millems:
This reverts commit b8cffe6df26ae9e0717140bc852f40c98d19683e.


millems:
Integration tests passed.

Quanzzzz (approved):

PR #2205, opened 2020-12-16
Adding Crosslink redirect for API reference document

joviegas:
## Description

- Adding Crosslink redirect for API reference document
## Motivation and Context

- As of now we donot have Links to Java Doc link for API reference
- This PR will will create a static page which will extract the ClientName and service name from the API refenence and redirect to the s3 location where we have Java V2 docs,

### generate_cross_link_data.py 
-  this loops through all the services service-2.json  and gets the service name and client name
- The python code is same as Java code where we generate the [DefaultNamingStrategy-Link](https://github.com/aws/aws-sdk-java-v2/blob/35c5f89ae29b3e5bbf3ab711c1b7b7024b76bd97/codegen/src/main/java/software/amazon/awssdk/codegen/naming/DefaultNamingStrategy.java#L116-L130)

### crosslink_redirect.html
- This will redirect the SPI refernced link via Java Scripts the maps are populated during publish java docs from above python scripts.

### release-javadoc.yml
- This will create the static redirect html  in the S3 Java while releasing Java docs.

## Testing

- Tested end to end by running publish-javadoc builds
- Tested error cases when incorrect redirect in this case it will land to index.html

## Types of changes

## Checklist

## License

### DIFF
--- a/buildspecs/release-javadoc.yml
+++ b/buildspecs/release-javadoc.yml
@@ -22,3 +22,5 @@ phases:
     - aws s3 sync $DOC_PATH/$RELEASE_VERSION/ $DOC_PATH/latest/ --acl=public-read --delete
     - jar cf aws-java-sdk-v2-docs.jar -C target/site/apidocs .
     - aws s3 cp aws-java-sdk-v2-docs.jar $DOC_PATH/ --acl="public-read"
+    - python ./scripts/doc_crosslinks/generate_cross_link_data.py --apiDefinitionsBasePath ./services/ --apiDefinitionsRelativeFilePath src/main/resources/codegen-resources/service-2.json  --templateFilePath ./scripts/doc_crosslinks/crosslink_redirect.html --outputFilePath ./scripts/crosslink_redirect.html
+    - aws s3 cp ./scripts/crosslink_redirect.html $DOC_PATH/latest/ --acl="public-read"
--- a/scripts/doc_crosslinks/generate_cross_link_data.py
+++ b/scripts/doc_crosslinks/generate_cross_link_data.py
@@ -1,0 +1,120 @@
+import os
+import argparse
+import io
+import codecs
+import json
+import re
+# import requests
+from os import listdir
+from os.path import isdir, exists, join
+from re import split
+
+sdks = {}
+clientClass = {}
+
+def generateDocsMap(apiDefinitionsPath, apiDefinitionsRelativeFilePath):
+
+    filesInDir = [f for f in listdir(apiDefinitionsPath) if isdir(join(apiDefinitionsPath, f))]
+    for file in filesInDir :
+        serviceJsonFileName = join(apiDefinitionsPath, join(file, apiDefinitionsRelativeFilePath))
+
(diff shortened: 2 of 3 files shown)


scripts/doc_crosslinks/generate_cross_link_data.py
```diff
@@ -0,0 +1,120 @@
...
+    parser.add_argument("--apiDefinitionsRelativePath", action="store")
+    parser.add_argument("--templatePath", action="store")
+    parser.add_argument("--outputPath", action="store")
+    
+    args = vars( parser.parse_args() )
+    argMap = {}
+    argMap[ "apiDefinitionsBasePath" ] = args[ "apiDefinitionsBasePath" ] or "./../services/"
+    argMap[ "apiDefinitionsRelativePath" ] = args[ "apiDefinitionsRelativePath" ] or "/src/main/resources/codegen-resources/service-2.json"
+    argMap[ "templatePath" ] = args[ "templatePath" ] or "./"
+    argMap[ "outputPath" ] = args[ "outputPath" ] or "/"
```
cenedhryn, line 114: Are `templatePath` and `outputPath` always supposed to reference files, rather than a directory path?
If so
1. Could there be more natural defaults than `./` or `/`?
2. It'd be nice if the names of these two arguments differed from the previous, to distinguish their type

buildspecs/release-javadoc.yml
```diff
@@ -22,3 +22,5 @@ phases:
     - aws s3 sync $DOC_PATH/$RELEASE_VERSION/ $DOC_PATH/latest/ --acl=public-read --delete
     - jar cf aws-java-sdk-v2-docs.jar -C target/site/apidocs .
     - aws s3 cp aws-java-sdk-v2-docs.jar $DOC_PATH/ --acl="public-read"
+    - python ./scripts/doc_crosslinks/generate_cross_link_data.py --apiDefinitionsBasePath ./services/ --apiDefinitionsRelativePath src/main/resources/codegen-resources/service-2.json  --templatePath ./scripts/doc_crosslinks/crosslink_redirect.html --outputPath ./scripts/crosslink_redirect.html
```
cenedhryn, line 25: Are there any alternative folders for the source of the template html file and the output? Especially the output doesn't seem like it belongs in the scripts folder, but if there is no other logical place to put them, it may be unnecessary to create a new folder.

scripts/doc_crosslinks/generate_cross_link_data.py
joviegas: >Are templatePath and outputPath always supposed to reference files, rather than a directory path?
templatePath is Directory Path and outputPath is File Path. I will rename and set defaults accordingly

buildspecs/release-javadoc.yml
joviegas: There are no alternatives fir template html.
The "doc_crosslinks" is created to make sure all the required scripts/resource can be found in a single place.
The processed html is copied to parent ./script folder when the script is executed as a part of build process this output file is a temp file which is immediately copied to S3 .

cenedhryn (approved):

PR #2209, opened 2020-12-17
Enable links to contributors' github pages

zoewangg:
## Description
Enable links to contributors' github pages

## License

### DIFF
--- a/.changes/2.15.25.json
+++ b/.changes/2.15.25.json
@@ -35,7 +35,7 @@
         {
             "type": "bugfix",
             "category": "AWS SDK for Java v2",
-            "contributor": "@frosforever",
+            "contributor": "frosforever",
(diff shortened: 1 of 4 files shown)


.changes/2.15.25.json
```diff
@@ -35,7 +35,7 @@
         {
             "type": "bugfix",
             "category": "AWS SDK for Java v2",
-            "contributor": "@frosforever",
+            "contributor": "[@frosforever](https://github.com/frosforever)",
```
dagnir, line 38: Doing the formatting in the property value itself makes changelog generation less flexible; e.g. what if we want to just collect a list of contributors without the link? We should do the formatting of the link at the generation step.

zoewangg: Fair enough. Will update

dagnir (approved):

Issue #1684, opened 2020-03-04
Attempt timeouts causing non-retryable AbortedExceptions


isaac-jordan:
## Expected Behavior
I would expect a `ApiCallAttemptTimeoutException` to be thrown.

## Current Behavior

After setting a custom value for `ClientOverrideConfiguration.builder().apiCallAttemptTimeout` for a DynamoDB client I began to see occasional  `SdkClientException` errors with a message of `Unable to unmarshall response (null). Response Code: 200, Response Text: OK`, and a cause of `AbortedException`.

These exceptions are not retried by the SDK, causing my calls to error out completely without following my retry policy (like it would if a `ApiCallAttemptTimeoutException` was thrown). This is causing failures for my service.

Logs showing the stack trace internal to the SDK are available [here](https://github.com/aws/aws-sdk-java-v2/files/4288366/dynamodb.sdkclientexception.log.txt)

## Possible Solutions
- Catch `AbortedException`s that are caused by the thread being interrupted when an API call attempt times out and re-throw an `ApiCallAttemptTimeoutException`.
- Or, add `AbortedException` as a retryable exception in `SdkDefaultRetrySetting`

## Steps to Reproduce (for bugs)
This Kotlin snippet manages to reproduce it for me most of the time, since it keeps retrying until it decides to stop (often because of an AbortedException). If this doesn't work for you, try tweaking the `apiCallAttemptTimeout` until you see an `AbortedException`. 

Note that I can also reproduce this more reliably by setting a very low timeout (e.g. 2ms), but that's definitely a more artificial test (but should also be fixed).

```
        val retryPolicy = RetryPolicy.defaultRetryPolicy().toBuilder()
            .numRetries(30)
            .retryCondition(OrRetryCondition.create(
                RetryCondition.defaultRetryCondition(),
                RetryCondition { c: RetryPolicyContext ->
                    c.exception().printStackTrace() // Print out error for each failure to make retries obvious
                    println("Retries attempt: ${c.retriesAttempted()}")
                    false
                }
                )
            )
            .build()

        val client = DynamoDbClient.builder()
            .overrideConfiguration {
                it.retryPolicy(retryPolicy)

                // Should retry up to 30 times, but often exits early due to an AbortedException (depends on your luck and internet connection)
[...]

PR #2220, opened 2020-12-25
Fix for Issue #1684 Changes in handleTimeoutCausedException in Timeou…

joviegas:
…tExceptionHandlingStage

## Description

Some of the Retry attempts which failed due to the API TimeOuts did not successfully retried but ended up with AbortedException.
## Motivation and Context

Refer [Issue 1684](https://github.com/aws/aws-sdk-java-v2/issues/1684)

### Root Cause
- This is intermittent timing issue.
-Occurs in below scenario
1. ApiCallTimeoutTrackingStage Times out as its not getting the response. This Interrupts the thread.
2. Request is still in progress but in some places we check the Thread interrupt status it then fails because the Thread is interrupted and throws Aborted Exception which is then thrown with SDKCLient Exception
3. Now ApiCallTimeoutTrackingStage's handleInterruptedException resets the interrupt flag.
4. And when the TimeoutExceptionHandlingStage resumes and tries to translatePipelineException() it fails to handle the exception because thread Interrupt is RESET and its SdkClientException (Hadled ones are InterruptedException, AbortedException, IOException)
5. Thus fails to handle the Exception

### Fix Options

1. Do changes in ApiCallTimeoutTrackingStage's handleInterruptedException() not to clear the interrupt .
Cons :- However this fix doesnot looks feasible since as per ApiCallTimeoutTrackingStage it rightly resets the Interrupt of the current Thread as part of had handleInterruptedException

2. translatePipelineException  of TimeoutExceptionHandlingStage should check (e.getCause instance of AbortedException)
Cons :- We are getting too specifics to AbortedException and assuming that all the APIs fail with AbortedException whenever the request is aborted due to TimeOuts. So this looks we might miss some other exceptions or new exception added in future/

3. In translatePipelineException check if the isCausedByApiCallAttemptTimeout and its SdkClientException
Pros : This handles all the cases also in handleTimeoutCausedException we have if clauses which does the filtering.
Cons : not aware.

## Testing

- Added Junits
- Integ Test added and Integ Testing done

## Types of changes

## Checklist

## License

### DIFF
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/TimeoutExceptionHandlingStage.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/TimeoutExceptionHandlingStage.java
@@ -25,6 +25,7 @@ import software.amazon.awssdk.core.client.config.ClientOverrideConfiguration;
 import software.amazon.awssdk.core.client.config.SdkClientOption;
 import software.amazon.awssdk.core.exception.AbortedException;
 import software.amazon.awssdk.core.exception.ApiCallAttemptTimeoutException;
+import software.amazon.awssdk.core.exception.SdkClientException;
 import software.amazon.awssdk.core.exception.SdkInterruptedException;
 import software.amazon.awssdk.core.internal.http.HttpClientDependencies;
 import software.amazon.awssdk.core.internal.http.RequestExecutionContext;
@@ -90,7 +91,8 @@ public final class TimeoutExceptionHandlingStage<OutputT> implements RequestToRe
      */
     private Exception translatePipelineException(RequestExecutionContext context, Exception e) {
         if (e instanceof InterruptedException || e instanceof IOException ||
-            e instanceof AbortedException || Thread.currentThread().isInterrupted()) {
+                e instanceof AbortedException || Thread.currentThread().isInterrupted()
+                || (e instanceof SdkClientException && isCausedByApiCallAttemptTimeout(context))) {
             return handleTimeoutCausedException(context, e);
         }
         return e;
--- a/.changes/next-release/bugfix-AWSSDKforJavav2-b5a2807.json
+++ b/.changes/next-release/bugfix-AWSSDKforJavav2-b5a2807.json
@@ -1,0 +1,6 @@
+{
+    "category": "AWS SDK for Java v2", 
+    "contributor": "", 
+    "type": "bugfix", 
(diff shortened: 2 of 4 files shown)


core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/TimeoutExceptionHandlingStage.java
```diff
@@ -90,7 +91,8 @@ public TimeoutExceptionHandlingStage(HttpClientDependencies dependencies, Reques
      */
     private Exception translatePipelineException(RequestExecutionContext context, Exception e) {
         if (e instanceof InterruptedException || e instanceof IOException ||
-            e instanceof AbortedException || Thread.currentThread().isInterrupted()) {
+                e instanceof AbortedException || Thread.currentThread().isInterrupted()
+                || (e instanceof SdkClientException && isCausedByApiCallAttemptTimeout(context))) {
```
dagnir, line 95: We check `isCausedByApiCallAttemptTimeout` but can this be caused by the API call timer as well?

joviegas: The timeout exception is caused while  [ApiCallAttemptTimeoutExecute](https://github.com/aws/aws-sdk-java-v2/blob/35267ca707c3fb5cdf7e3e98758a8ef969269183/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/ApiCallAttemptTimeoutTrackingStage.java#L73) and this is the same check which resets the interrupts 
[ApiCallAttemptTimeoutTrackingStageInterruptReset](https://github.com/aws/aws-sdk-java-v2/blob/35267ca707c3fb5cdf7e3e98758a8ef969269183/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/ApiCallAttemptTimeoutTrackingStage.java#L135).

API call timer  causes context.apiCallTimeoutTracker() while the one used above is 
context.apiCallAttemptTimeoutTracker()

dagnir: Okay gotcha, so if the API call timer fired (and _not_ the attempt timer), then the exception would be `AbortedException` which we already handle.

dagnir (approved):

Issue #2154, opened 2020-11-20
Presigned URL resolves in a SignatureDoesNotMatch


TVDSC:
## Describe the issue

Hi,
I have functionalities in our web application to upload and download a certain configuration file (xml).
The upload works.
But we want to enforce the download by using a presigned URL that is valid for 5 minutes.

When using that url we receive a SignatureDoesNotMatch.
<img width="1632" alt="Screenshot 2020-11-20 at 11 00 09" src="https://user-images.githubusercontent.com/26546590/99787278-d5730e80-2b1f-11eb-995c-113a1a3f5b68.png">

## Steps to Reproduce
> Code to upload a configuration file:
```
public Boolean uploadConfiguration(Part file) throws IOException {
        String bucketName = System.getenv("AWS_BUCKET");
        PutObjectRequest request = PutObjectRequest.builder()
                .bucket(bucketName)
                .key("BeFirstConfiguration.xml")
                .build();
        PutObjectResponse response = s3Client.putObject(request, RequestBody.fromBytes(file.getInputStream().readAllBytes()));
        return StringUtils.isNotEmpty(response.eTag());
    }
```

> Code to receive a presigned URL:
```
String bucketName = System.getenv("AWS_BUCKET");

        GetObjectRequest getUrlRequest = GetObjectRequest.builder()
                .bucket(bucketName)
                .key("BeFirstConfiguration.xml")
                .build();
        GetObjectPresignRequest getObjectPresignRequest = GetObjectPresignRequest.builder()
                .getObjectRequest(getUrlRequest)
                .signatureDuration(Duration.ofMinutes(10))
                .build();

        PresignedGetObjectRequest request = s3Presigner.presignGetObject(getObjectPresignRequest);

        log.info("Generating pre-signed URL.");
        return request.url().toString();
```

> JavaConfig configuration
```
public S3Configuration s3Configuration() {
        return S3Configuration.builder().build();
    }

    @Bean
    public S3Presigner s3Presigner() {
        return S3Presigner.builder()
                .region(Region.EU_WEST_1)
                .serviceConfiguration(s3Configuration())
                .build();
    }

    @Bean
    public S3Client s3Client() {
        return S3Client.create();
    }
```

## Current Behavior

After using that presigned url we get an SignatureDoesNotMatch.

## Your Environment

* AWS Java SDK version used:  2.15.31
* JDK version used: openjdk:11 (docker image based on openjdk:11-jre-slim)
[...]

PR #2231, opened 2021-01-08
Fix inconsistency of checksunValidationEnabled between default S3Configuration and no S3Configuration

Quanzzzz:
Change the default value of `checksumValidationEnabled` as `false` to achieve parity between presigners with the default S3Configuration and no S3Cofiguration, to avoid `signatureNotMatched` error while using the presigned url in browser with the default S3Configuration.

## Motivation and Context

#2154 

## Testing

Unit tests added.

## Types of changes

## Checklist

## License

### DIFF
--- a/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/presigner/DefaultS3Presigner.java
+++ b/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/presigner/DefaultS3Presigner.java
@@ -90,6 +90,7 @@ import software.amazon.awssdk.services.s3.transform.GetObjectRequestMarshaller;
 import software.amazon.awssdk.services.s3.transform.PutObjectRequestMarshaller;
 import software.amazon.awssdk.services.s3.transform.UploadPartRequestMarshaller;
 import software.amazon.awssdk.utils.IoUtils;
+import software.amazon.awssdk.utils.Logger;
 import software.amazon.awssdk.utils.Validate;
 
 /**
@@ -97,10 +98,12 @@ import software.amazon.awssdk.utils.Validate;
  */
 @SdkInternalApi
 public final class DefaultS3Presigner extends DefaultSdkPresigner implements S3Presigner {
+    private static final Logger log = Logger.loggerFor(DefaultS3Presigner.class);
+
     private static final AwsS3V4Signer DEFAULT_SIGNER = AwsS3V4Signer.create();
     private static final S3Configuration DEFAULT_S3_CONFIGURATION = S3Configuration.builder()
-            .checksumValidationEnabled(false)
-            .build();
+                                                                                   .checksumValidationEnabled(false)
+                                                                                   .build();
     private static final String SERVICE_NAME = "s3";
     private static final String SIGNING_NAME = "s3";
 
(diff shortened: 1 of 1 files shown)


services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Configuration.java
```diff
@@ -50,7 +50,7 @@
     /**
      * S3 payload checksum validation is by default enabled;
      */
-    private static final boolean DEFAULT_CHECKSUM_VALIDATION_ENABLED = true;
+    private static final boolean DEFAULT_CHECKSUM_VALIDATION_ENABLED = false;
```
zoewangg, line 53: Hmm, I think we should keep checksum enabled by default.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/presigner/DefaultS3Presigner.java
```diff
@@ -98,9 +98,7 @@
 @SdkInternalApi
 public final class DefaultS3Presigner extends DefaultSdkPresigner implements S3Presigner {
     private static final AwsS3V4Signer DEFAULT_SIGNER = AwsS3V4Signer.create();
-    private static final S3Configuration DEFAULT_S3_CONFIGURATION = S3Configuration.builder()
-            .checksumValidationEnabled(false)
-            .build();
+    private static final S3Configuration DEFAULT_S3_CONFIGURATION = S3Configuration.builder().build();
```
zoewangg, line 101: Instead of changing the default configuration, should we just validate the customized service configuration and log a warning message or something if checksum is enabled on line 117 ?

services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Configuration.java
cenedhryn: +1; what are the implications of changing the default apart from the presigner?

zoewangg: The implication is that the SDK will no longer do data integrity validation for get and put object by default, which is not what we want :)

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/presigner/DefaultS3Presigner.java
```diff
@@ -117,6 +120,11 @@ private DefaultS3Presigner(Builder b) {
         super(b);
 
         this.serviceConfiguration = b.serviceConfiguration != null ? b.serviceConfiguration : DEFAULT_S3_CONFIGURATION;
+        if (serviceConfiguration.checksumValidationEnabled()) {
+            log.warn(() -> "Please note that the checksum validation is enabled by default, so if you are passing a "
+                           + "default S3Configuration, the presigned url won't be browser-executable.");
```
zoewangg, line 125: Can we change it to something like the following?
```
The provided S3Configuration has ChecksumValidationEnabled set to true. Please note the pre-signed request can't be executed using a web browser if checksum is enabled.
```
Because they could pass a non-default S3Configuration with checksumValidation enabled explicitly

On second thought, maybe we should change the level to `debug` since this doesn't concern customers who do not need it to be browser executable.

Quanzzzz: Sounds reasonable, will do the change.

zoewangg (approved):

PR #2261, opened 2021-01-26
adding integ tests for S3Control to get better coverage for inconsist…

cenedhryn:
…ent exception unmarshalling

## Description
Adds integration tests for s3control

## Motivation and Context
There is an ongoing issue which is now partially fixed:
https://github.com/aws/aws-sdk-java-v2/issues/1986

These tests show that the error code is now correctly returned, but it does not return an InvalidRequestException, it's parsed as an S3ControlException, so the fix is not complete. 

## Testing
Local integ tests


services/s3control/src/it/java/software.amazon.awssdk.services.s3control/S3ControlIntegrationTest.java
```diff
@@ -20,6 +20,7 @@
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
 
+import org.assertj.core.api.ThrowableAssert;
```
Quanzzzz, line 23: Looks like this dependency wasn't used?

services/s3control/src/it/java/software.amazon.awssdk.services.s3control/S3ControlIntegrationTest.java
```diff
@@ -29,6 +30,7 @@
 import software.amazon.awssdk.http.SdkHttpFullRequest;
 import software.amazon.awssdk.services.s3control.model.DeletePublicAccessBlockRequest;
 import software.amazon.awssdk.services.s3control.model.GetPublicAccessBlockResponse;
+import software.amazon.awssdk.services.s3control.model.InvalidRequestException;
```
Quanzzzz, line 33: Looks like this dependency also wasn't used?

services/s3control/src/it/java/software.amazon.awssdk.services.s3control/S3ControlIntegrationTest.java
```diff
@@ -102,6 +104,40 @@ public void getPublicAccessBlock_NoSuchAccount() {
         }
     }
 
+    @Test
+    public void getPublicAccessBlock_NoSuchPublicAccessBlock() {
+        try {
+            client.getPublicAccessBlock(r -> r.accountId(accountId));
+            fail("Expected exception");
+        } catch (S3ControlException e) {
+            assertEquals("NoSuchPublicAccessBlockConfiguration", e.awsErrorDetails().errorCode());
```
Quanzzzz, line 113: Can we change this to `assertThat`? And same as all the other usages below? `assertThat` is more readable and flexible, and it's used more often in our code base.

services/s3control/src/it/java/software.amazon.awssdk.services.s3control/S3ControlIntegrationTest.java
cenedhryn: Missed that, built with the quick flag...

services/s3control/src/it/java/software.amazon.awssdk.services.s3control/S3ControlIntegrationTest.java
cenedhryn: I prefer assertThat too. Changed all assertEquals in the file (but did not touch the assertTrue etc)

Quanzzzz (approved):

PR #2313, opened 2021-03-04
Add support for adapting the sdk configurations to crt configurations.

Quanzzzz:
Add support for adapting the sdk configurations to crt configurations.

## Motivation and Context

The java sdk configurations are not exactly the same as the crt configurations, so we need to adapt the sdk configurations to the crt configurations.

## Testing

A unit test testing if the crt credentials provider can be created correctly is added in this pr.

### DIFF
--- a/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/S3CrtUtils.java
+++ b/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/S3CrtUtils.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.services.s3.internal;
+
+import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.auth.credentials.AwsCredentials;
+import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider;
+import software.amazon.awssdk.auth.credentials.AwsSessionCredentials;
+import software.amazon.awssdk.crt.auth.credentials.CredentialsProvider;
+import software.amazon.awssdk.crt.auth.credentials.StaticCredentialsProvider;
+
+@SdkInternalApi
+public final class S3CrtUtils {
+
+    private S3CrtUtils() {
+    }
+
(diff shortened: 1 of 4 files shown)


services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/S3CrtUtils.java
```diff
@@ -0,0 +1,47 @@
...
+
+import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.auth.credentials.AwsCredentials;
+import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider;
+import software.amazon.awssdk.auth.credentials.AwsSessionCredentials;
+import software.amazon.awssdk.crt.auth.credentials.CredentialsProvider;
+import software.amazon.awssdk.crt.auth.credentials.StaticCredentialsProvider;
+
+@SdkInternalApi
+public class S3CrtUtils {
```
zoewangg, line 26: nit, missing final

services/s3/src/test/java/software/amazon/awssdk/services/s3/internal/S3CrtUtilsTest.java
```diff
@@ -0,0 +1,49 @@
...
+import software.amazon.awssdk.crt.auth.credentials.CredentialsProvider;
+
+public class S3CrtUtilsTest {
+
+    private final String ACCESS_KEY = "accessKey";
+    private final String SECRET_ACCESS_KEY = "secretAccessKey";
+    private final String SESSION_TOKEN = "sessionToken";
+
+    @Test
+    public void createS3CrtClientTest() throws ExecutionException, InterruptedException {
```
zoewangg, line 37: nit, this is a test to create s3 crt client

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/S3CrtUtils.java
```diff
@@ -0,0 +1,47 @@
...
+
+    private S3CrtUtils() {
+    }
+
+    /**
+     * Adapter between the sdk credentials provider and the crt credentials provider.
+     */
+    public static CredentialsProvider createCrtCredentialsProvider(AwsCredentialsProvider awsCredentialsProvider) {
```
zoewangg, line 34: nit, can we add a TODO here to make sure we revisit this once crt adds more credential provider?

zoewangg (approved):

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/S3CrtUtils.java
Quanzzzz: My bad, will add the `final`.

services/s3/src/test/java/software/amazon/awssdk/services/s3/internal/S3CrtUtilsTest.java
Quanzzzz: I copied the test here but forgot to change the name, will make the change.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/S3CrtUtils.java
Quanzzzz: Makes sense, will do.

PR #2322, opened 2021-03-09
Append user agent for retry mode

Quanzzzz:
Append user agent for retry mode so we can collect the metrics of the usage of each retry mode.

## Testing

A unit test is added


core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/ApplyUserAgentStage.java
```diff
@@ -94,6 +96,15 @@ private StringBuilder getUserAgent(SdkClientConfiguration config, List<ApiName>
                  .append("/")
                  .append(SdkHttpUtils.urlEncode(clientName));
 
+        String retryMode = config.option(SdkClientOption.RETRY_POLICY).retryMode().toString();
```
millems, line 99: Can retry mode be null? E.g. if the customer does `RetryPolicy.builder((RetryMode) null).build()`? Arguably that should probably not be allowed, but I feel like it can happen.

core/sdk-core/src/test/java/software/amazon/awssdk/core/http/AmazonHttpClientTest.java
```diff
@@ -183,6 +185,34 @@ public void testUserAgentContainsHttpClientInfo() {
...
+              .executionContext(ClientExecutionAndRequestTimerTestUtils.executionContext(null))
+              .execute(combinedSyncResponseHandler(handler, null));
+
+        ArgumentCaptor<HttpExecuteRequest> httpRequestCaptor = ArgumentCaptor.forClass(HttpExecuteRequest.class);
+        verify(sdkHttpClient).prepareRequest(httpRequestCaptor.capture());
+
+        final String userAgent = httpRequestCaptor.getValue().httpRequest().firstMatchingHeader("User-Agent")
+                                                  .orElseThrow(() -> new AssertionError("User-Agent header was not found"));
+
+        System.out.println(userAgent);
```
millems, line 212: Not needed

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/ApplyUserAgentStage.java
Quanzzzz: It will trigger another exception while the retry mode is being used, but to make it more clear that the retry mode should not be set as null, a new exception is added to the `RetryPolicy` class.

core/sdk-core/src/main/java/software/amazon/awssdk/core/retry/RetryPolicy.java
```diff
@@ -103,6 +104,9 @@ public static Builder builder() {
      * Create a {@link RetryPolicy.Builder} populated with the defaults from the provided {@link RetryMode}.
      */
     public static Builder builder(RetryMode retryMode) {
+        Validate.notNull(retryMode, "The retry mode cannot be set as null. If you don't want to set the retry mode,"
```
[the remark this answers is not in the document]
Quanzzzz: Sure, will do

zoewangg (approved):

Issue #2301, opened 2021-02-26
Thoughts on exposing endpoint prefix in client interface


abrooksv:
## Describe the Feature
Add new public String to each generated client interface containing the endpoint prefix

## Is your Feature Request related to a problem?
If we look at ECR we get the service name (or ID) is "ecr":

```
public interface EcrClient extends SdkClient {
    String SERVICE_NAME = "ecr";
```

but ECRs endpoint data is "api.ecr"
https://github.com/aws/aws-sdk-java-v2/blob/41a86004660eb94974b8a1459bf5254c59248806/core/regions/src/main/resources/software/amazon/awssdk/regions/internal/region/endpoints.json#L324
https://github.com/aws/aws-sdk-java-v2/blob/8db24fd31b68207fa0a001995c4608dfc5bcbce2/services/ecr/src/main/resources/codegen-resources/service-2.json#L5

## Proposed Solution
```
public interface EcrClient extends SdkClient {
    String SERVICE_NAME = "ecr";
	String REGION_ENDPOINT_PREFIX = "api.ecr";
```

This is similar to what is exposed in .NET: https://github.com/aws/aws-sdk-net/blob/master/sdk/src/Services/ECR/Generated/AmazonECRConfig.cs#L54

This string can then be passed to the public API: 
https://github.com/aws/aws-sdk-java-v2/blob/93f2f1389d0811f06c7d5e58c9643629b96d517a/core/regions/src/main/java/software/amazon/awssdk/regions/ServiceMetadataProvider.java#L32

We can bikeshed on the name later

## Additional Context

- [x] I may be able to implement this feature request

## Your Environment

* AWS Java SDK version used:
* JDK version used:
* Operating System and version:

debora-ito:
@abrooksv it's a reasonable ask, marking as a feature request. Looking forward to see your PR 😃

abrooksv:
PR created #2327

PR #2327, opened 2021-03-11
Expose the metadata key on client interfaces

abrooksv:
## Description
Add a new static field on the client interfaces with metadata ID (aka API endpoint prefix)

## Motivation and Context
#2301 

## Testing
```
public interface ApiGatewayClient extends SdkClient {
    String SERVICE_NAME = "apigateway";

    /**
     * Value for looking up the service's metadata from the
     * {@link software.amazon.awssdk.regions.ServiceMetadataProvider}.
     */
    String SERVICE_METADATA_ID = "apigateway";
```

```
public interface ApiGatewayAsyncClient extends SdkClient {
    String SERVICE_NAME = "apigateway";

    /**
     * Value for looking up the service's metadata from the
     * {@link software.amazon.awssdk.regions.ServiceMetadataProvider}.
     */
    String SERVICE_METADATA_ID = "apigateway";
```

```
public interface EcrAsyncClient extends SdkClient {
    String SERVICE_NAME = "ecr";

    /**
     * Value for looking up the service's metadata from the
     * {@link software.amazon.awssdk.regions.ServiceMetadataProvider}.
     */
    String SERVICE_METADATA_ID = "api.ecr";
```

## Types of changes

## Checklist

## License


codegen/src/test/resources/software/amazon/awssdk/codegen/poet/client/c2j/json/service-2.json
```diff
@@ -2,12 +2,13 @@
...
     "apiVersion": "2010-05-08",
-    "endpointPrefix": "json-service",
+    "endpointPrefix": "json-service-endpoint",
     "globalEndpoint": "json-service.amazonaws.com",
     "protocol": "rest-json",
     "serviceAbbreviation": "Json Service",
     "serviceFullName": "Some Service That Uses Json Protocol",
-    "serviceId":"Json Service",
+    "serviceId": "Json Service",
+    "signingName": "json-service",
```
abrooksv, line 11: Added so it tests when singing name != endpoint prefix

abrooksv:
I am a bit stumped why the sync client code gen test is failing with extra spaces in the JavaDoc, ideas

Screen Shot 2021-03-11 at 1 23 54 PM

abrooksv:
Digging into this more it seems there is a 

```
    static ServiceMetadata serviceMetadata() {
        return ServiceMetadata.of(SERVICE_METADATA_ID);
    }
```
on the sync interface (it is missing from the async interface).

This is basically what I needed the ID for in the first place, but on order to make what I need work (dynamic service info in the IDEs), I basically need to find a way to update `MetadataLoader` with an updated endpoints.json data that the toolkits team vends from CloudFront (https://idetoolkits.amazonwebservices.com/endpoints.json)

```
/**
 * Internal class for determing where to load region and service
 * metadata from. Currently only generated region metadata is supported.
 */
@SdkInternalApi
final class MetadataLoader {
```

Is there any thoughts on how to make the `Currently only generated region metadata is supported.` a bit less true of a statement 😉 To be clear, I am not advocating for dropping the codegen, but a way to load newer data into the `MetadataLoader` using the SDKs endpoints metadata format. Note: The toolkits are onboarded to the SDK release platform, that is how we get this data today. 

The endpoints.json parser currently lives in `codegen-lite`, so it would have to be added to new module that lives below `codegen-lite` and `regions`.

Quanzzzz:
Sorry for delay. This pr should be good to go once we get another review from the team, as this is a change to the public apis. About the extra feature request mentioned in the comment, we can implement a new utillity method to read the service metadata directly from the json file so you can get "dynamic service info". We've created a new feature request task in our backlog, will implement it once it's prioritized.

Quanzzzz (approved):

zoewangg:
@all-contributors please add @abrooksv for code

PR #2331, opened 2021-03-13
Implement getObject in S3CrtAsyncClient

zoewangg:
## Description
Implement getObject in S3CrtAsyncClient

~~This depends on https://github.com/awslabs/aws-crt-java/pull/292~~

~~Still need to do more testing and clean up unnecessary the logging statements~~

The change in CRT https://github.com/awslabs/aws-crt-java/pull/296 has been merged, but it hasn't been released yet. So the build will fail because `aws-crt-java: 1.0.0-SNAPSHOT` is not available in Maven

## Types of changes

## Checklist

## License

### DIFF
--- a/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/CrtResponseDataConsumerAdapter.java
+++ b/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/CrtResponseDataConsumerAdapter.java
@@ -1,0 +1,85 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.services.s3.internal;
+
+import com.amazonaws.s3.ResponseDataConsumer;
+import com.amazonaws.s3.model.GetObjectOutput;
+import java.nio.ByteBuffer;
+import java.util.concurrent.CompletableFuture;
+import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.core.SdkStandardLogger;
+import software.amazon.awssdk.core.async.AsyncResponseTransformer;
+import software.amazon.awssdk.crt.CrtRuntimeException;
+import software.amazon.awssdk.crt.http.HttpHeader;
+import software.amazon.awssdk.http.HttpStatusFamily;
+import software.amazon.awssdk.services.s3.model.GetObjectResponse;
+import software.amazon.awssdk.utils.Logger;
+
+/**
+ * Adapt the SDK API {@link AsyncResponseTransformer} to the CRT API {@link ResponseDataConsumer}.
+ */
+@SdkInternalApi
+public final class CrtResponseDataConsumerAdapter<ReturnT> implements ResponseDataConsumer<GetObjectOutput> {
+    private static final Logger log = Logger.loggerFor(CrtResponseDataConsumerAdapter.class);
+    private final AsyncResponseTransformer<GetObjectResponse, ReturnT> transformer;
+    private final CompletableFuture<ReturnT> future;
+    private final S3CrtDataPublisher publisher;
+
+    public CrtResponseDataConsumerAdapter(AsyncResponseTransformer<GetObjectResponse, ReturnT> transformer) {
+        this.transformer = transformer;
+        this.future = transformer.prepare();
+        this.publisher = new S3CrtDataPublisher();
+    }
+
+    public CompletableFuture<ReturnT> transformerFuture() {
+        return future;
+    }
+
+    @Override
+    public void onResponseHeaders(int statusCode, HttpHeader[] headers) {
+        if (HttpStatusFamily.of(statusCode) == HttpStatusFamily.SUCCESSFUL) {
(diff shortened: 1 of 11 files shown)


services/s3/src/test/java/software/amazon/awssdk/services/s3/crt/S3CrtGetObjectIntegrationTest.java
```diff
@@ -0,0 +1,121 @@
...
+    private static final String BUCKET = temporaryBucketName(S3CrtGetObjectIntegrationTest.class);
+    private static final String KEY = "key";
+    private static S3CrtAsyncClient crtClient;
+    private static File file;
+
+    @BeforeClass
+    public static void setup() throws IOException {
+        createBucket(BUCKET);
+        crtClient = S3CrtAsyncClient.builder()
+                                    .region(Region.US_WEST_2)
```
dagnir, line 51: Nit: use `DEFAULT_REGION`

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/CrtResponseDataConsumerAdapter.java
```diff
@@ -0,0 +1,84 @@
...
+    @Override
+    public void onResponseData(ByteBuffer byteBuffer) {
+        log.debug(() -> "Received data: " + byteBuffer.remaining());
+        publisher.deliverData(byteBuffer);
+    }
+
+    @Override
+    public void onException(CrtRuntimeException e) {
+        log.debug(() -> "An error occurred ", e);
+        transformer.exceptionOccurred(e);
```
dagnir, line 76: Do we need to make sure the subscriber sees this error?

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/CrtResponseDataConsumerAdapter.java
```diff
@@ -0,0 +1,84 @@
...
+
+    @Override
+    public void onResponse(GetObjectOutput output) {
+        GetObjectResponse response = S3CrtUtils.adaptGetObjectOutput(output);
+        transformer.onResponse(response);
+        transformer.onStream(publisher);
+    }
+
+    @Override
+    public void onResponseData(ByteBuffer byteBuffer) {
```
dagnir, line 68: There's no back pressure right? CRT will just call this as fast as possible?

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/CrtResponseDataConsumerAdapter.java
zoewangg: Good call. Just checked the javadocs of `onException`, and it actually means ` Exceptions indicate the operation is finished.`, so we should definitely notify the subscriber


https://github.com/awslabs/aws-crt-java/blob/main/s3-native-client/src/main/java/com/amazonaws/s3/OperationHandler.java#L16-L20

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/CrtResponseDataConsumerAdapter.java
zoewangg: Yeah, CRT is planning to add backpressure by making the API return the number of consumed bytes from this API and adjust the sliding window size accordingly.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/S3CrtDataPublisher.java
```diff
@@ -0,0 +1,141 @@
...
+    public void notifyErrorIfNeeded(Exception exception) {
+        if (subscriberRef.get() != null) {
+            subscriberRef.get().onError(exception);
+        }
+    }
+
+    public void deliverData(ByteBuffer byteBuffer) {
+        buffer.add(byteBuffer);
+        flushBuffer();
+    }
```
dagnir, line 71: We probably want to check `isDone` here if the subscription was cancelled early, especially for cases where there's lots of data still being downloaded.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/S3CrtDataPublisher.java
```diff
@@ -0,0 +1,141 @@
...
+
+    public void notifyStreamingFinished() {
+        buffer.add(COMPLETE);
+        flushBuffer();
+    }
+
+    public void notifyErrorIfNeeded(Exception exception) {
+        if (subscriberRef.get() != null) {
+            subscriberRef.get().onError(exception);
```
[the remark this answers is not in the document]
zoewangg: +1 Updated

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/S3CrtDataPublisher.java
zoewangg: Good catch. Updated

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/S3CrtDataPublisher.java
```diff
@@ -0,0 +1,222 @@
...
+    }
+
+    public void notifyError(Exception exception) {
+        // If the subscription is cancelled, no op
+        if (isDone) {
+            return;
+        }
+
+        isDone = true;
+        buffer.clear();
```
dagnir, line 72: Hmm we might not want to clear here since if the consumer is currently draining, then you could get an NPE on line 108 

```
            while (!buffer.isEmpty() && outstandingDemand.get() > 0) {
                log.debug(() -> "buffer is not empty & there's request demand. " + "Buffer size: " + buffer.size() + " "
                               + "Demand: " + outstandingDemand.get());
               
               // buffer might get cleared before poll()
                Event event = buffer.poll();
```

Probably better to just let `flushBuffer()` do it when it sees `ERROR`

dagnir (approved):

zoewangg:
The build failed because aws-crt-java: 1.0.0-SNAPSHOT is not available in Maven. We will update crt version once it's released.

Issue #2280, opened 2021-02-11
DynamoDbEnhancedClient: BeanTableSchema.mapToItem returns null when given an empty map or an all-null map, should return an uninitialized bean


bill-phast:
## Expected Behavior
A map with all null attributes, or with no non-null attributes, should return an uninitialized bean. Instead it returns null. Especially when beans contain other beans, this differences is significant and can be hard to work around.

## Current Behavior
This application shows the problem. It takes an all-null bean, encodes it, then decodes it; instead of getting an all-null bean back, it gets just null:
```
import software.amazon.awssdk.enhanced.dynamodb.TableSchema;
import software.amazon.awssdk.enhanced.dynamodb.mapper.annotations.DynamoDbBean;

import java.util.Map;

public class MainClass {
    @DynamoDbBean
    public static class BeanClass {
        private String name;
        public String getName() { return name; }
        public void setName(String name) { this.name = name; }
    }
    public static void main(String[] args) {
        TableSchema<BeanClass> schema = TableSchema.fromBean(BeanClass.class);
        BeanClass b = new BeanClass();
        System.out.println("All-null bean cannot be decoded, returns null instead: " +
                               schema.mapToItem(schema.itemToMap(b, true)));
        System.out.println("Should print an empty bean, will print null: " +
                               schema.mapToItem(Map.of()));
    }
}
```
Output:
```
All-null bean cannot be decoded, returns null instead: null
Should print an empty bean, will print null: null
```

## Possible Solution
The bug is in StaticImmutableTableSchema.mapToItem. It lazily constructs the bean, which means that if no setters ever need to be called, the bean is never constructed and null is returned. If it is changed to construct the bean at the start of the call, then the problem is solved.

If this seems acceptable I would be happy to write a pull request.

## Steps to Reproduce (for bugs)
Compile and run the example code above.

## Context
This is blocking us from moving to SDK 2. SDK 1 does not have this bug, and we do store objects with uninitialized beans, and we need to preserve the objects as-is, not null out these beans.

## Your Environment
* AWS Java SDK version used: 2.15.79
* JDK version used: 1.3
* Operating System and version: MacOS, and whatever AWS Lambdas runs

debora-ito:
Hi @bill-phast can you tell us more about your use case? Please be as specific as you can. 

[...]

PR #2334, opened 2021-03-15
Empty beans are now preserved through serialization and deserialization

reed53:
## Description

Previously when a bean that had all null attributes, but itself was not null, was serialized and deserialized, the bean would become null. With this change, the bean is not lost through serialization and deserialization.

## Motivation and Context

This pull request aims to fix the following issue: https://github.com/aws/aws-sdk-java-v2/issues/2280
My team and I were trying to move to SDK 2, however this issue caused a loss in preservation through Dynamo, so this aims to fix that.

## Testing

Previously existing tests were modified so that if given an empty bean, it expects an empty bean in return.

## Types of changes

## Checklist

## License


zoewangg:
Thank you for your PR and apologies for the delayed response. We will take a look shortly

zoewangg (commented):
I understand the use case, but this change might be backwards-incompatible. I'll discuss with the team.

zoewangg:
Thank you for your patience! We can't merge this change for backwards-compatibility reasons. We think customizing it at the attribute level might be a better way to support this. It would be something like `DynamoDbUpdateBehavior`, say `DynamoDbPreserveEmptyBean` (name subject to change) and you can put it on the attribute where you want to preserve the empty beans.

https://github.com/aws/aws-sdk-java-v2/blob/master/services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/mapper/annotations/DynamoDbUpdateBehavior.java

Would you able to make the change? If not, we will add this task to our backlog, but we don't have a timeline for this change at the moment.

reed53:
> Thank you for your patience! We can't merge this change for backwards-compatibility reasons. We think customizing it at the attribute level might be a better way to support this. It would be something like `DynamoDbUpdateBehavior`, say `DynamoDbPreserveEmptyBean` (name subject to change) and you can put it on the attribute where you want to preserve the empty beans.
> 
> https://github.com/aws/aws-sdk-java-v2/blob/master/services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/mapper/annotations/DynamoDbUpdateBehavior.java
> 
> Would you be able to make the change? If not, we will add this task to our backlog, but we don't have a timeline for this change at the moment.

That sounds good! I will start working on that now, and let you know when I finish.

reed53:
Hi @zoewangg, I've been working on updating the pull request but I've been running into issues as to how to pass information into the inner bean. At first, I tried to implement an annotation on a method, however since I need the attribute key to access tags, it would be impossible to get the tag from the inner bean, since I don't have access to the attribute key the outer bean uses. I then thought to instead implement an annotation on the class and set a flag on the table schema for the inner bean upon its creation in order to tell whether the bean should be preserved when empty or not. The issue with this is that the only way to pass down this information would be to either modify the parameters of public methods, or to modify how the schemas are recursively created to make it avoid the public methods. Is there a better approach to this that you can think of?

zoewangg:
Yeah, this is bit tricky. I looked into it yesterday, and I can't think of other approaches. @bmaizels, can you think of anything?

> The issue with this is that the only way to pass down this information would be to either modify the parameters of public methods, or to modify how the schemas are recursively created to make it avoid the public methods

Could you elaborate on this? More specifically, what public methods need to be modified? We can update the public APIs in the internal classes (classes with `@SdkInternalApi` annotation)

reed53:
> Yeah, this is bit tricky. I looked into it yesterday, and I can't think of other approaches. @bmaizels, can you think of anything?
> 
> > The issue with this is that the only way to pass down this information would be to either modify the parameters of public methods, or to modify how the schemas are recursively created to make it avoid the public methods
> 
> Could you elaborate on this? More specifically, what public methods need to be modified? We can update the public APIs in the internal classes (classes with `@SdkInternalApi` annotation)

The public method create for BeanTableSchema is referenced by recursiveCreate, and BeanTableSchema is unfortunately labeled as public api. The reason I wanted to use this method specifically is because that is what is used when the inner bean schemas are created, which would be the ideal spot to include the preserve empty behavior.

zoewangg:
I talked with @bmaizels offline. It seems there are some other approaches we can look into:
- create a client-level configuration for this behavior, although it means all beans will have this behavior.
- introduce a new `mapToItem` that takes a boolean flag to preserve empty bean, similar to what we have in `Map<String, AttributeValue> itemToMap(T item, boolean ignoreNulls)`

@reed53 WDUT? Would either of those work for your use-case? I'll also look into it a bit more to see which approach is better and get back to you later.

reed53:
> I talked with @bmaizels offline. It seems there are some other approaches we can look into:
> 
> * create a client-level configuration for this behavior, although it means all beans will have this behavior.
> * introduce a new `mapToItem` that takes a boolean flag to preserve empty bean, similar to what we have in `Map<String, AttributeValue> itemToMap(T item, boolean ignoreNulls)`
> 
> @reed53 WDUT? Would either of those work for your use-case? I'll also look into it a bit more to see which approach is better and get back to you later.

The behavior for every bean is ideal, so the client sounds good, however it will still likely be pretty messy to implement, since we need to pass around a flag everywhere. I recently sent an email about the issue and got a response asking to schedule a call with a member of the SDK team. I will wait until after this call to continue working on this issue, as the call may help to clear some things up.

zoewangg:
Hi @reed53 I investigated a bit more and the `#2` option that I mentioned above (introduce a new mapToItem that takes a boolean flag to preserve empty bean) would not work because there's no way to pass the flag to `DocumentAttributeConverter#transformTo` with current implementation.

I'm not a big fan of client-level configuration as well because it's not very flexible, so I went ahead and implemented the class level annotation. Here is the PR #2364. Feedback is welcome! 🙂

reed53:
> Hi @reed53 I investigated a bit more and the `#2` option that I mentioned above (introduce a new mapToItem that takes a boolean flag to preserve empty bean) would not work because there's no way to pass the flag to `DocumentAttributeConverter#transformTo` with current implementation.
> 
> I'm not a big fan of client-level configuration as well because it's not very flexible, so I went ahead and implemented the class level annotation. Here is the PR #2364. Feedback is welcome! 🙂

I have no comments beyond what is already mentioned there, but thank you very much for doing this!

zoewangg:
Closing in favor of #2364

PR #2366, opened 2021-04-01
Fix buffer leak with go awaway packets

bondj:
When processing Http2GoAwayEvents Netty reports buffer leaks

## Description
The buffer does not appear to be released, and I think it's acceptable to convert to a string without bumping the refCount

## Motivation and Context
Potential memory leak

## Testing
Currently performing long run test. The leak warnings usually show up in a couple of days, so will confirm a good fix then.
UnitTests updated

## Types of changes

## Checklist

- [X ] I have read the **CONTRIBUTING** document
- [ X] My code follows the code style of this project
- [ X] My change requires a change to the Javadoc documentation
- [ X] I have updated the Javadoc documentation accordingly
- [ X] I have read the **README** document
- [ X] I have added tests to cover my changes
- [ X] All new and existing tests passed
- [ X] A short description of the change has been added to the **CHANGELOG**
- [ X] My change is to implement 1.11 parity feature and I have updated [LaunchChangelog](https://github.com/aws/aws-sdk-java-v2/blob/master/docs/LaunchChangelog.md)

## License

- [ X] I confirm that this pull request can be released under the Apache 2 license


http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/Http2GoAwayEventListener.java
```diff
@@ -41,7 +42,7 @@ public Http2GoAwayEventListener(Channel parentChannel) {
     @Override
     public void onGoAwayReceived(int lastStreamId, long errorCode, ByteBuf debugData) {
         Http2MultiplexedChannelPool channelPool = parentChannel.attr(ChannelAttributeKey.HTTP2_MULTIPLEXED_CHANNEL_POOL).get();
-        GoAwayException exception = new GoAwayException(errorCode, debugData.retain());
+        GoAwayException exception = new GoAwayException(errorCode, debugData.toString(StandardCharsets.UTF_8));
```
zoewangg, line 45: Should we release the ByteBuf after this? Preferably in a finally block?

zoewangg (commented):
Thanks for the PR! Could you add a changelog entry? You can generate one by running the `scripts/new-change` script and following the instructions.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/Http2GoAwayEventListener.java
bondj: I do not believe so... The original bug was that by retaining the debugData, the refCount was increased, but never released.  I restructured the code, to convert to a String and pass into the GoAwayException, so we always know the lifecycle of that bytebuf, and it seemed tidier than try/final the modifying of the refCount for processing that happens all within this method..

The Javadoc for https://netty.io/4.1/api/io/netty/handler/codec/http2/DefaultHttp2Connection.html goAwayReceived says 
> debugData - The Additional Debug Data in the GOAWAY frame. Note that reference count ownership belongs to the caller (ownership is not transferred to this method).
So I think it's safe to just use the debugData content within the scope of the method, and the code shouldn't release it.

I will do a quick check, to see if this bug appears elsewhere.

bondj: I will update the changelog once we've run the fix for a week, and I'm confident that it fixes our problem :)

bondj: Looks like the fix, fixes the problem

zoewangg: Nice, thanks for verifying.

zoewangg (approved):

zoewangg:
Thank you again for your contribution!

@all-contributors please add @bondj for code

PR #2370, opened 2021-04-02
Add an Optional getter for the contentLength and a new fromContentProvider method in RequestBody

Quanzzzz:
Add an Optional getter for the `contentLength`  a new `fromContentProvider` method allowing contentLength in class `RequestBody` to be null.

## Description

Some APIs don't require the `Content-Length` header and support receiving the request body using Transfer-Encoding: chunked. For example,` Lex postContent`.

Currently, `RequestBody` returns a primitive long from its `contentLength()` method, and all static constructors on `RequestBody` require a content length parameter if it can't be determined automatically; e.g. `fromInputStream`.

With the optional getter we can omit the `content-length` from `RequestBody` so that we can fully use service operations that don't require it upfront; especially in cases where the content length can't be determined beforehand.

Also, with the newly added `fromContentProvider` method the `RequestBody` can be created with a `null` content-length when the `content-length` is not needed.

## Testing

Corresponding unit test is added.


core/sdk-core/src/main/java/software/amazon/awssdk/core/sync/RequestBody.java
```diff
@@ -47,13 +48,15 @@
...
     private final ContentStreamProvider contentStreamProvider;
     private final long contentLength;
+    private final Optional<Long> optionalContentLength;
     private final String contentType;
 
-    private RequestBody(ContentStreamProvider contentStreamProvider, long contentLength, String contentType) {
+    private RequestBody(ContentStreamProvider contentStreamProvider, Long contentLength, String contentType) {
         this.contentStreamProvider = paramNotNull(contentStreamProvider, "contentStreamProvider");
-        this.contentLength = contentLength;
+        this.contentLength = contentLength != null ? contentLength : 0;
```
Quanzzzz, line 56: I'm still not sure if setting the null contentLength as 0 is a good idea, but that's the only way I can think of without breaking change.

core/sdk-core/src/main/java/software/amazon/awssdk/core/sync/RequestBody.java
```diff
@@ -205,6 +216,19 @@ public static RequestBody fromContentProvider(ContentStreamProvider provider, lo
...
+    /**
+     * Creates a {@link RequestBody} from the given {@link ContentStreamProvider}.
+     *
+     * @param provider The content provider.
+     * @param contentLength The content length, using non primitive data type.
+     * @param mimeType The MIME type of the content.
+     *
+     * @return The created {@code RequestBody}.
+     */
+    public static RequestBody fromContentProvider(ContentStreamProvider provider, Long contentLength, String mimeType) {
```
zoewangg, line 228: Should we just omit the contentLength param?

core/sdk-core/src/main/java/software/amazon/awssdk/core/sync/RequestBody.java
```diff
@@ -70,6 +73,13 @@ public long contentLength() {
         return contentLength;
     }
```
zoewangg, line 74: Should we deprecate this method?

core/sdk-core/src/main/java/software/amazon/awssdk/core/sync/RequestBody.java
zoewangg: Since we require content length to be non negative, we can probably use a negative value to denote that content length is absent.

Quanzzzz: Sounds good, will do the change.

core/sdk-core/src/main/java/software/amazon/awssdk/core/sync/RequestBody.java
Quanzzzz: Yea, we can also do that. But it's going to break the customers that are using this method currently. I can do the change and bump up the minor version though.

core/sdk-core/src/main/java/software/amazon/awssdk/core/sync/RequestBody.java
Quanzzzz: Same as above, this will be a breaking change. But I can do the change if that matches with our standard better. I agree omitting this param can be a better customer experience.

zoewangg: Why would adding `RequestBody fromContentProvider(ContentStreamProvider provider, String mimeType)` be a breaking change?

core/sdk-core/src/main/java/software/amazon/awssdk/core/sync/RequestBody.java
zoewangg: I meant adding `@deprecate` annotation, so it's not a breaking change.

core/sdk-core/src/main/java/software/amazon/awssdk/core/sync/RequestBody.java
Quanzzzz: Oh I was thinking about removing the `contentLength` param in the original `fromContentProvider` method, because the getter of `contentLength` is going to be removed, and the only source of `contentLength` would be the length of files or byte arrays, so I thought it's doable to remove the `contentLength` parameter from the origin `fromContentProvider` method, it's just a breaking change. But anyway, adding a new method should be totally fine.

core/sdk-core/src/main/java/software/amazon/awssdk/core/sync/RequestBody.java
Quanzzzz: Got you, will do.

core/sdk-core/src/main/java/software/amazon/awssdk/core/sync/RequestBody.java
```diff
@@ -49,11 +50,10 @@
     private final long contentLength;
     private final String contentType;
 
-    private RequestBody(ContentStreamProvider contentStreamProvider, long contentLength, String contentType) {
+    private RequestBody(ContentStreamProvider contentStreamProvider, Long contentLength, String contentType) {
         this.contentStreamProvider = paramNotNull(contentStreamProvider, "contentStreamProvider");
-        this.contentLength = contentLength;
+        this.contentLength = contentLength != null ? contentLength : Integer.MIN_VALUE;
```
zoewangg, line 55: I think we should still validate it that it's not negative if not null so that it fails fast. 

Something like this:
https://github.com/aws/aws-sdk-java-v2/blob/master/utils/src/main/java/software/amazon/awssdk/utils/Validate.java#L622

core/sdk-core/src/main/java/software/amazon/awssdk/core/sync/RequestBody.java
```diff
@@ -64,12 +64,22 @@ public ContentStreamProvider contentStreamProvider() {
     }
 
     /**
-     * @return Content length of {@link RequestBody}.
+     * @deprecated *by {*@link *#optionalContentLength()}
```
zoewangg, line 67: nit, can we keep `@return` statement

core/sdk-core/src/main/java/software/amazon/awssdk/core/sync/RequestBody.java
```diff
@@ -49,11 +50,10 @@
     private final long contentLength;
```
zoewangg, line 50: Can we change it to `Long contentLength`? That way, we can just check null instead in the `contentLength` method

zoewangg (approved):

PR #2373, opened 2021-04-05
Fix ApacheHttpClient to use socket keep alive when appropriate

ajs139:
## Description
This enables the underlying [`Socket.SO_KEEPALIVE`](https://docs.oracle.com/javase/7/docs/api/java/net/SocketOptions.html#SO_KEEPALIVE) when the `CONNECTION_KEEPALIVE` configuration option is set to `true`.

## Motivation and Context
There's a [known issue](https://docs.aws.amazon.com/vpc/latest/userguide/nat-gateway-troubleshooting.html#nat-gateway-troubleshooting-timeout) when using a NAT gateway that it times out after 350 seconds if the connection is not kept alive. If using ApacheHttpClient to create connections to an AWS Lambda from a machine within the VPC it appears that this connection goes through the NAT. This means that any Lambda invocations taking more than 350 seconds will never get a response.

## Testing
Deployed the code to environment that was experiencing the timeout issue, and verified that setting `setSoKeepAlive(true)` on the `SocketConfig` enabled keep-alive and allowed connections running longer than 350 seconds to respond.

## Types of changes

## Checklist

## License


zoewangg:
Thank you for the PR! We will take a look shortly

http-clients/apache-client/src/main/java/software/amazon/awssdk/http/apache/ApacheHttpClient.java
```diff
@@ -385,6 +387,11 @@ public String clientName() {
          */
         Builder connectionMaxIdleTime(Duration maxIdleConnectionTimeout);
 
+        /**
+         * Configure the connection to use java.net.SocketOptions.SO_KEEPALIVE.
+         */
+        Builder connectionKeepalive(Boolean keepConnectionAlive);
```
zoewangg, line 393: Same as above. Can we change it to `tcpKeepAlive`?

http-clients/apache-client/src/main/java/software/amazon/awssdk/http/apache/ApacheHttpClient.java
```diff
@@ -675,10 +693,10 @@ public void checkServerTrusted(X509Certificate[] x509Certificates, String s) thr
             };
         }
 
-        private SocketConfig buildSocketConfig(AttributeMap standardOptions) {
+        private SocketConfig buildSocketConfig(AttributeMap standardOptions,
+                                                  Boolean soKeepAlive) {
```
zoewangg, line 697: Can we get soKeepAlive from `standardOptions.get(SdkHttpConfigurationOption.TCP_KEEPALIVE)` directly instead of passing it?

http-client-spi/src/main/java/software/amazon/awssdk/http/SdkHttpConfigurationOption.java
```diff
@@ -59,6 +59,12 @@
     public static final SdkHttpConfigurationOption<Duration> CONNECTION_MAX_IDLE_TIMEOUT =
             new SdkHttpConfigurationOption<>("ConnectionMaxIdleTimeout", Duration.class);
 
+    /**
+     * Whether or not to use keepalive on the connection.
+     */
+    public static final SdkHttpConfigurationOption<Boolean> CONNECTION_KEEPALIVE =
```
zoewangg, line 65: Can we rename it to `TCP_KEEPALIVE` so that users don't confuse it with HTTP keep alive?

http-client-spi/src/main/java/software/amazon/awssdk/http/SdkHttpConfigurationOption.java
```diff
@@ -116,6 +122,7 @@
     private static final Duration DEFAULT_CONNECTION_TIMEOUT = Duration.ofSeconds(2);
     private static final Duration DEFAULT_CONNECTION_ACQUIRE_TIMEOUT = Duration.ofSeconds(10);
     private static final Duration DEFAULT_CONNECTION_MAX_IDLE_TIMEOUT = Duration.ofSeconds(60);
+    private static final Boolean DEFAULT_CONNECTION_KEEPALIVE = Boolean.TRUE;
```
zoewangg, line 125: We should probably make it default to false so that it has the same behavior as now.

zoewangg (commented):
Could you add a changelog entry? You can generate one by running the `scripts/new-change` script and following the instructions.

http-client-spi/src/main/java/software/amazon/awssdk/http/SdkHttpConfigurationOption.java
ajs139: Sorry, thought it was, not sure what happened there.

ajs139:
@zoewangg - thanks for the code review, it's much cleaner now.

zoewangg (commented):
Thanks for updating the PR! Just have one minor comment, otherwise lgtm.

http-clients/apache-client/src/main/java/software/amazon/awssdk/http/apache/ApacheHttpClient.java
```diff
@@ -559,6 +564,16 @@ public void setCredentialsProvider(CredentialsProvider credentialsProvider) {
             credentialsProvider(credentialsProvider);
         }
 
+        @Override
+        public Builder tcpKeepAlive(Boolean keepConnectionAlive) {
+            standardOptions.put(SdkHttpConfigurationOption.TCP_KEEPALIVE, keepConnectionAlive);
+            return this;
+        }
+
+        public void setTcpKeepalive(Boolean keepConnectionAlive) {
```
zoewangg, line 573: It seems the setter also needs to be updated to `setTcpKeepAlive`

ajs139: Missed that one, thank you.

zoewangg (approved):
LGTM. Thank you for your contribution!

zoewangg:
@all-contributors please add @ajs139 for code

PR #2375, opened 2021-04-05
Resubmit Add Execution Attributes Overrides for Client and Request Configurations

ndejaco2:
## Description
This change adds getters/setters for execution attributes on the client and request override configuration objects in order to allow these attributes to be added prior to a request being made. The additional execution attributes are added to the collection at the time of a request.

## Motivation and Context
The motivation for this change is to allow a request to pass a custom object override instead of relying on thread local storage for passing data to a client interceptor.

## Testing
Unit tests added for getters and setters.

## Types of changes

## Checklist

- [x ] I have read the **CONTRIBUTING** document
- [x ] My code follows the code style of this project
- [x ] All new and existing tests passed
- [x ] A short description of the change has been added to the **CHANGELOG**

## License


core/sdk-core/src/main/java/software/amazon/awssdk/core/interceptor/ExecutionAttributes.java
```diff
@@ -48,11 +66,61 @@
         return this;
     }
 
+    /**
+     * Merge attributes of a higher precedence into the current lower precedence collection.
+     */
+    public void merge(ExecutionAttributes higherPrecedenceExecutionAttributes) {
+        higherPrecedenceExecutionAttributes.getAttributes().forEach(this.attributes::put);
+    }
+
```
dagnir, line 75: This is opposite of what `AttributeMap` does; they should be identical so that `this` has higher precedence. Then in `AwsClientHandlerUtils` we can use the attributes on the request as the 'base', and merge the lower precedence attributes into it.

core/sdk-core/src/main/java/software/amazon/awssdk/core/RequestOverrideConfiguration.java
```diff
@@ -371,6 +388,22 @@ default B putRawQueryParameter(String name, String value) {
 
         List<MetricPublisher> metricPublishers();
 
+        /**
+         * Sets the additional execution attributes collection for this request.
+         * @param executionAttributes Execution attributes map for this request
+         * @return This object for method chaining.
+         */
+        B executionAttributes(Map<ExecutionAttribute<?>, ?> executionAttributes);
```
dagnir, line 396: Any reason we don't just accept `ExecutionAttributes`? Same for the client config

core/sdk-core/src/main/java/software/amazon/awssdk/core/interceptor/UnmodifiableExecutionAttributes.java
```diff
@@ -0,0 +1,35 @@
...
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.core.interceptor;
+
+import software.amazon.awssdk.annotations.SdkPublicApi;
+
+@SdkPublicApi
+public  class UnmodifiableExecutionAttributes extends ExecutionAttributes {
+    public UnmodifiableExecutionAttributes(ExecutionAttributes executionAttributes) {
```
dagnir, line 22: Minor: please make this a private class, and add a static constructor on `ExecutionAttributes` instead, so customers can do something like

```
ExecutionAttributes.unmodifiableExecutionAttributes(myAttributes);
```

Being on `ExecutionAttributes` will make it more discoverable

core/aws-core/src/main/java/software/amazon/awssdk/awscore/client/handler/AwsClientHandlerUtils.java
```diff
@@ -60,6 +62,24 @@ private AwsClientHandlerUtils() {
 
     }
 
+    static <InputT extends SdkRequest, OutputT extends SdkResponse> ExecutionAttributes addInitialExecutionAttributes(
+            ClientExecutionParams<InputT, OutputT> executionParams, SdkClientConfiguration clientConfig) {
```
dagnir, line 66: Doesn't look like this is being used in the sync clients

test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/ExecutionAttributesTest.java
```diff
@@ -69,4 +67,131 @@ public void asyncClient_disableHostPrefixInjection_isPresent() {
             async.close();
         }
     }
-}
+
+    @Test
+    public void asyncClient_clientOverrideExecutionAttribute() {
```
dagnir, line 72: Please add tests for the sync clients as well

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/handler/BaseClientHandler.java
```diff
@@ -195,8 +195,9 @@ private static InterceptorContext runModifyHttpRequestAndHttpContentInterceptors
                     (OutputT) response.toBuilder().sdkHttpResponse(httpFullResponse).build());
     }
 
-    static ExecutionAttributes addInitialExecutionAttributes(ExecutionAttributes executionAttributes) {
-        return executionAttributes.putAttribute(InternalCoreExecutionAttribute.EXECUTION_ATTEMPT, 1);
```
dagnir, line 199: Can we avoid making this change from `static` to `protected`? My preference would be to make least amount of changes structurally to support this

dagnir (changes requested):

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/handler/BaseClientHandler.java
ndejaco2: this change cant be done with it being static. This is required due to insistence on making the merge method work like the attribute map. Otherwise I would revert to what I had previously which needed to be changed.

test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/ExecutionAttributesTest.java
ndejaco2: Initial change was to do this only for async clients as the pointer you sent me is only called by async clients. Would you like me to implement for sync clients as well?

dagnir: Sorry about that. Yes, since the options are available through the sync paths, the sync clients should support it as well.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/handler/BaseClientHandler.java
dagnir: I don't follow why it can't be without changing these from static. Can you explain?

ndejaco2:
The changes require per client and per request overrides. The client
configuration that needs to be accessed is a non static member variable.

dagnir:
LGTM! Can you squash your commits and rebase?

dagnir (approved):

PR #2388, opened 2021-04-07
Add support for event stream requests over RPC

rschmitt:
This commit adds support for RPC services that use client->server or
bidirectional event streaming. In the RPC formats (i.e. AWS JSON), the
marshalled request and response objects are sent as the first event in
the request and response stream respectively. This is different from the
REST formats, such as REST JSON, which bind the request and response
fields to non-payload positions (headers, URI, method, status code,
query string) when event streaming is used.

The implementation of this feature is mainly performed by a new global
interceptor, `EventStreamInitialRequestInterceptor`. The interceptor
detects when an `initial-request` event needs to be sent and transforms
the async response body by constructing the `initial-request` message
and prepending it to the event stream. Due to the high precedence of
global interceptors, any subsequent service or override interceptors
that run on the body will see the transformed version. This also applies
to the `EventStreamAws4Signer`, which will produce a signature over the
`initial-request` event.

I confirm that this pull request can be released under the Apache 2 license.

### DIFF
--- a/core/aws-core/src/main/java/software/amazon/awssdk/awscore/eventstream/EventStreamInitialRequestInterceptor.java
+++ b/core/aws-core/src/main/java/software/amazon/awssdk/awscore/eventstream/EventStreamInitialRequestInterceptor.java
@@ -1,0 +1,108 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.awscore.eventstream;
+
+import static software.amazon.awssdk.core.interceptor.SdkInternalExecutionAttribute.HAS_INITIAL_REQUEST_EVENT;
+import static software.amazon.awssdk.core.internal.util.Mimetype.MIMETYPE_EVENT_STREAM;
+import static software.amazon.awssdk.http.Header.CONTENT_TYPE;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.ByteBuffer;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+import org.reactivestreams.Publisher;
+import software.amazon.awssdk.annotations.SdkProtectedApi;
+import software.amazon.awssdk.core.async.AsyncRequestBody;
+import software.amazon.awssdk.core.interceptor.Context.ModifyHttpRequest;
+import software.amazon.awssdk.core.interceptor.ExecutionAttributes;
+import software.amazon.awssdk.core.interceptor.ExecutionInterceptor;
+import software.amazon.awssdk.core.internal.async.AsyncStreamPrepender;
+import software.amazon.awssdk.core.sync.RequestBody;
+import software.amazon.awssdk.http.SdkHttpRequest;
+import software.amazon.eventstream.HeaderValue;
+import software.amazon.eventstream.Message;
+
+/**
+ * An interceptor for event stream requests sent over RPC. This interceptor will prepend the initial request (i.e. the
+ * serialized request POJO) to the stream of events supplied by the caller.
+ */
+@SdkProtectedApi
+public class EventStreamInitialRequestInterceptor implements ExecutionInterceptor {
+    @Override
+    public SdkHttpRequest modifyHttpRequest(
+        ModifyHttpRequest context, ExecutionAttributes executionAttributes
+    ) {
+        if (!Boolean.TRUE.equals(executionAttributes.getAttribute(HAS_INITIAL_REQUEST_EVENT))) {
+            return context.httpRequest();
+        }
+
+        return context.httpRequest().toBuilder()
+            .removeHeader(CONTENT_TYPE)
+            .putHeader(CONTENT_TYPE, MIMETYPE_EVENT_STREAM)
+            .build();
+    }
+
+    @Override
+    public Optional<AsyncRequestBody> modifyAsyncHttpContent(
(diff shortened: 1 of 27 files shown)


dagnir:
@rschmitt Awesome! Appreciate the contribution.  We will need schedule some time on our side to review this so there will be a delay, but we'll have a look soon.

core/aws-core/src/main/resources/software/amazon/awssdk/global/handlers/execution.interceptors
```diff
@@ -0,0 +1 @@
+software.amazon.awssdk.awscore.eventstream.EventStreamInitialRequestInterceptor
```
zoewangg, line 1: I think we need to add `reflect-config.json ` file in this module for this interceptor.  That should fix the `aws-sdk-java-v2-native-image-test` build

Example: https://github.com/aws/aws-sdk-java-v2/blob/master/core/sdk-core/src/main/resources/META-INF/native-image/software.amazon.awssdk/sdk-core/reflect-config.json

rschmitt: OK, I'll try that.

dagnir:
I think for this change, it makes sense to add some codegen reftests, just to make sure we're actually generating things like `.withInitialRequestEvent(true)` for the `json` services. I think you can just a JSON-RPC specific test here https://github.com/aws/aws-sdk-java-v2/blob/a402b403997e7531f000b8e5d795224c9e62c7df/codegen/src/test/java/software/amazon/awssdk/codegen/poet/client/PoetClientFunctionalTests.java#L27

core/aws-core/src/main/java/software/amazon/awssdk/awscore/eventstream/EventStreamInitialRequestInterceptor.java
```diff
@@ -0,0 +1,107 @@
...
+
+        Map<String, HeaderValue> initialRequestEventHeaders = new HashMap<>();
+        initialRequestEventHeaders.put(":message-type", HeaderValue.fromString("event"));
+        initialRequestEventHeaders.put(":event-type", HeaderValue.fromString("initial-request"));
+        initialRequestEventHeaders.put(":content-type", HeaderValue.fromString(contentType));
+
+        ByteBuffer initialRequest = new Message(initialRequestEventHeaders, payload).toByteBuffer();
+
+        Publisher<ByteBuffer> asyncRequestBody = context.asyncRequestBody()
+            .orElseThrow(() -> new IllegalStateException("This request should have an asyncRequestBody"));
```
dagnir, line 84: minor: can we indicate in the error message that it's because it's an event streaming request? Would help in debugging

dagnir:
I think ideally, we'd leverage the marshallers to marshall the request directly to a suitable `initial-request` instead of "remarshalling" in the interceptor, but I'm fine with this approach as well since the marshaller route would require a bit more work in the codegen; we could save that for a second iteration in the future.

dagnir:
Integ tests seem to be failing. Can you take a look?

Locally, I'm seeing `software.amazon.awssdk.services.s3.S3PresignerIntegrationTest` fail, but there may be others. You can run them using

`mvn clean install -P integration-tests`

rschmitt:
> Integ tests seem to be failing. Can you take a look?
> 
> Locally, I'm seeing `software.amazon.awssdk.services.s3.S3PresignerIntegrationTest` fail, but there may be others. You can run them using
> 
> `mvn clean install -P integration-tests`

I wasn't immediately able to reproduce this particular failure.

```
[INFO] Running software.amazon.awssdk.services.s3.S3PresignerIntegrationTest
[INFO] Tests run: 11, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 4.305 s - in software.amazon.awssdk.services.s3.S3PresignerIntegrationTest
```

rschmitt:
> I think for this change, it makes sense to add some codegen reftests, just to make sure we're actually generating things like `.withInitialRequestEvent(true)` for the `json` services. I think you can just a JSON-RPC specific test here
> 
> https://github.com/aws/aws-sdk-java-v2/blob/a402b403997e7531f000b8e5d795224c9e62c7df/codegen/src/test/java/software/amazon/awssdk/codegen/poet/client/PoetClientFunctionalTests.java#L27

It looks like the JSON tests are actually `rest-json` tests; I need RPC tests, and I'm not sure any exist currently.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/AsyncStreamPrepender.java
```diff
@@ -0,0 +1,85 @@
...
+package software.amazon.awssdk.core.internal.async;
+
+import org.reactivestreams.Processor;
+import org.reactivestreams.Publisher;
+import org.reactivestreams.Subscriber;
+import org.reactivestreams.Subscription;
+import software.amazon.awssdk.annotations.SdkInternalApi;
+
+@SdkInternalApi
+public class AsyncStreamPrepender<T> implements Processor<T, T> {
```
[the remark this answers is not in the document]
rschmitt: This was a great suggestion; now I need to go back and add these for lots more reactive code I've written elsewhere 😬

core/sdk-core/src/test/java/software/amazon/awssdk/core/internal/async/AyncStreamPrependerTckTest.java
```diff
@@ -0,0 +1,43 @@
...
+
+package software.amazon.awssdk.core.internal.async;
+
+import org.reactivestreams.Publisher;
+import org.reactivestreams.tck.PublisherVerification;
+import org.reactivestreams.tck.TestEnvironment;
+
+import io.reactivex.Flowable;
+
+public class AyncStreamPrependerTckTest extends PublisherVerification<Long> {
```
dagnir, line 24: Please add a Subscriber TCK test as well, since the prepender is a `Processor` (publisher and subscriber)

dagnir (commented):
LGTM overall!

core/sdk-core/src/test/java/software/amazon/awssdk/core/internal/async/AyncStreamPrependerTckTest.java
rschmitt: This class should really only be used as a `Publisher`, so I'll try to push the `Subscriber` part down into a private inner class.

dagnir (approved):
:shipit:

PR #2423, opened 2021-04-19
Rewrote generated copiers to support builders for arbitrarily nested containers.

millems:
This fixes an issue where DynamoDB types like BatchGetResponse.Builder could not be serialized using
a bean-based serializer, because it contained a Map<K, List<Map<K2, V>>.

### DIFF
--- a/codegen/src/main/java/software/amazon/awssdk/codegen/poet/model/MemberCopierSpec.java
+++ b/codegen/src/main/java/software/amazon/awssdk/codegen/poet/model/MemberCopierSpec.java
@@ -15,18 +15,19 @@
 
 package software.amazon.awssdk.codegen.poet.model;
 
+import static software.amazon.awssdk.codegen.poet.model.TypeProvider.ShapeTransformation.NONE;
+import static software.amazon.awssdk.codegen.poet.model.TypeProvider.ShapeTransformation.USE_BUILDER;
+
 import com.squareup.javapoet.ClassName;
 import com.squareup.javapoet.CodeBlock;
 import com.squareup.javapoet.MethodSpec;
-import com.squareup.javapoet.ParameterizedTypeName;
 import com.squareup.javapoet.TypeName;
 import com.squareup.javapoet.TypeSpec;
-import com.squareup.javapoet.WildcardTypeName;
-import java.util.Collection;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.Map;
-import java.util.Optional;
 import java.util.stream.Collectors;
 import javax.lang.model.element.Modifier;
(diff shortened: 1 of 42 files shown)


pom.xml
```diff
@@ -512,6 +512,11 @@
                         <onlyModified>true</onlyModified>
                         <excludes>
                             <exclude>*.internal.*</exclude>
+                            <exclude>*.SubscribeToShardEvent$BuilderImpl</exclude>
+                            <exclude>*.ConfigurationEvent$BuilderImpl</exclude>
+                            <exclude>*.IntentResultEvent$BuilderImpl</exclude>
+                            <exclude>*.TextResponseEvent$BuilderImpl</exclude>
+                            <exclude>*.IntentResultEvent$BuilderImpl</exclude>
```
millems, line 519: Bean-style getter response type had to be changed.

codegen/src/main/java/software/amazon/awssdk/codegen/poet/model/MemberCopierSpec.java
```diff
@@ -482,4 +350,17 @@ private String memberParamName() {
         }
         return Utils.unCapitalize(memberModel.getC2jShape()) + "Param";
     }
+
+    private static final class VariableSource {
```
cenedhryn, line 354: This class name is too generic for me, I'd prefer if it was renamed to be more specific to its task. 
IDK if I love disambiguator but it's the only think I can think of now :-)

codegen/src/main/java/software/amazon/awssdk/codegen/poet/model/TypeProvider.java
```diff
@@ -142,42 +112,115 @@ public FieldSpec asField(MemberModel memberModel, Modifier... modifiers) {
...
 
-    private static boolean isContainerType(MemberModel m) {
-        return m.isList() || m.isMap();
+    public enum ShapeTransformation {
+        USE_BUILDER,
+        USE_BUILDER_IMPL,
+        NONE
+    }
+
+    public static final class TypeNameOptions {
```
cenedhryn, line 188: This class makes it much easier to reason around how to build types

codegen/src/main/java/software/amazon/awssdk/codegen/poet/model/MemberCopierSpec.java
millems: I can't think of a better name. It provides unique variable names... UniqueVariableSource? VariableUniquifier?

cenedhryn: UniqueVariableSource is better than VariableSource

cenedhryn (approved):

Issue #1966, opened 2020-07-31
DynamoDbEnhancedClient: Delete table operation


alexfuksenko-toast:
The DynamoDbEnhancedClient returns a DynamoDbTable<T> that allows you to do various dynamodb operations, including createTable(), but there is no way to delete the table. 

This means for repeated cases, especially for integration tests, I need to maintain a DynamoDbClient and a DynamoDbEnhancedClient reference. 

Even the enhanced client unit tests need to use a the normal client for only delete table operations like:
https://github.com/aws/aws-sdk-java-v2/blob/master/services-custom/dynamodb-enhanced/src/test/java/software/amazon/awssdk/enhanced/dynamodb/functionaltests/AsyncBasicQueryTest.java#L139

## Describe the Feature

I think it would be nice if the DynamoDbEnhancedClient was able to delete tables itself, either through the table reference or through the enhanced client

## Is your Feature Request related to a problem?

Awkward to create two different DynamoDbClients (enhanced and regular for usage)

## Additional Context

- [x] I may be able to implement this feature request

## Your Environment

* AWS Java SDK version used:
* JDK version used:
* Operating System and version:

debora-ito:
Thank you for reaching out @alexfuksenko-toast, marking as a feature request.

amrittuladhar:
Similarly, other table-level operations also seem to be missing from the v2 API, e.g. "describe table".

hiramf:
+1

bigunyak:
Are there any plans on implementing support for deleteTable operation yet?

debora-ito:
No timeline for it yet, will comment here when we have an update.

PR #2461, opened 2021-05-06
DDBEnhanced - Support delete-table operation

joviegas:
## Description

- DDBEnhanced - Support delete-table operation
- As of now deleteTable API was not available in DynamoDBTable interface. Added this new API and its default implementation

## Motivation and Context

#1966 requested this operation

## Testing

- Added JUnit test cases using DDBLocals and Mocks.

## Types of changes

## Checklist

## License

### DIFF
--- a/services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/DynamoDbAsyncTable.java
+++ b/services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/DynamoDbAsyncTable.java
@@ -667,4 +667,28 @@ public interface DynamoDbAsyncTable<T> extends MappedTableResource<T> {
     default CompletableFuture<T> updateItem(T item) {
         throw new UnsupportedOperationException();
     }
+
+    /**
+     * Deletes a table in DynamoDb with the name and schema already defined for this DynamoDbTable.
+     * <p>
+     * Use {@link DynamoDbEnhancedClient#table(String, TableSchema)} to define the mapped table resource.
+     * <p>
+     * This operation calls the low-level DynamoDB API DeleteTable operation.
+     * Note that this is an asynchronous operation and that the table may not immediately deleted. You can use
+     * {@link software.amazon.awssdk.services.dynamodb.waiters.DynamoDbAsyncWaiter#waitUntilTableNotExists}
+     * in the underlying client.
+     * <p>
+     * Example:
+     * <pre>
+     * {@code
+     *
+     * mappedTable.deleteTable().join();
+     * }
+     * </pre>
+     *
+     * @return a {@link CompletableFuture} of {@link Void}.
+     */
+    default CompletableFuture<Void> deleteTable() {
+        throw new UnsupportedOperationException();
+    }
 }
--- a/.changes/next-release/feature-AmazonDynamoDB-1607a75.json
+++ b/.changes/next-release/feature-AmazonDynamoDB-1607a75.json
@@ -1,0 +1,6 @@
+{
+    "category": "Amazon DynamoDB", 
+    "contributor": "", 
+    "type": "feature", 
+    "description": "Support delete-table operation in DDBEnhanced."
+}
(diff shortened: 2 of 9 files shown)


services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/DynamoDbAsyncTable.java
```diff
@@ -667,4 +667,28 @@
...
     }
+
+    /**
+     * Deletes a table in DynamoDb with the name and schema already defined for this DynamoDbTable.
+     * <p>
+     * Use {@link DynamoDbEnhancedClient#table(String, TableSchema)} to define the mapped table resource.
+     * <p>
+     * This operation calls the low-level DynamoDB API DeleteTable operation. Note that this is an asynchronous operation and that
+     * the table may not immediately deleted. Currently, there is no mechanism supported within this
+     * library to wait for/check the status of a deleted table. You must provide this functionality yourself.
```
millems, line 678: They can use the waiters in the underlying client to do it.

zoewangg: While we are at it, should we add `deleteTableUntilTableNotExists` instead of `deleteTable` here?

zoewangg: We have `waitUntilTableNotExists` in the `DynamoDbWaiter`.

If we decide to use the waiter in this operation, we can name it `deleteTableAndWaitUntilTableNotExists` or something similar.

millems: I'd personally rather just have deleteTable and have them call the waiter afterwards if they want to wait.

joviegas: Agree. Updated the comment , directing them to use waitUntilTableNotExists in the Waiters.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/DynamoDbAsyncTable.java
```diff
@@ -667,4 +667,27 @@
...
+     * <pre>
+     * {@code
+     *
+     * mappedTable.deleteTable().join();
+     * }
+     * </pre>
+     *
+     * @return a {@link CompletableFuture} of {@link Void}.
+     */
+    default CompletableFuture<Void> deleteTable() {
```
zoewangg, line 690: Should we return the `DeleteTableResponse`? Same as sync. I think we have feature requests on some existing APIs that are void.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/DynamoDbAsyncTable.java
zoewangg: Yeah, good point. People might be more interested to wait for a table to be created than deleted.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/DynamoDbAsyncTable.java
cenedhryn: +1 to returning a response, even if it's empty.

millems: I had this argument a lot with put-item and create-table to try to add a return type, but we settled on not returning something. I still don't agree with that decision, but it'd be inconsistent to return something from this one.

millems: (I'm not saying we shouldn't return something here, just that we'd need a good reason to be inconsistent with the other APIs, or a plan to make the other APIs consistent with this one.)

zoewangg: Oh, I completely forgot we have createTable already in this class that does not have a return type. In this case, I'd vote for consistency

joviegas: Had taken createTable as a reference for deleteTable , thus returning void.
I too will prefer keeping deleteTable consistent with createTable return type.

cenedhryn (approved):

PR #2476, opened 2021-05-12
Do not allow cross-region requests for S3 client with Fips Region.

joviegas:
## Description

Do not allow cross-region requests for Fips Enabled Client Regions

## Motivation and Context

Below are the cross region scenario and their expected behavior

 |ScenarioNo.|ClientRegion | ArnSourceRgion | useArnRegionFlag | ExpectedBehvior |
|-|-|-|-|-|
|1| Region.AWS_GLOBAL| us-east-1 |true | Success |
|2|Region.AWS_GLOBAL|us-east-1 | false |illegalArgumentException|
|3|s3-external-1 | us-east-1|true | Success| 
|4|s3-external-1 | us-east-1|false | illegalArgumentException| 
|5|fips-us-gov-east-1 | fips-us-gov-east-1|true | illegalArgumentException| 
|6|us-gov-east-1 | fips-us-gov-west-1|true | illegalArgumentException| 
|7|fips-us-gov-east-1 | us-gov-west-1|false | illegalArgumentException| 
|8|fips-us-gov-east-1 | us-gov-east-1|true | Success| 

 |ScenarioNo.|OldEndPoint | NewEndpoint | 
|-|-|-|
|1| myendpoint-123456789012.s3-accesspoint.dualstack.fips-us-gov-east-1.amazonaws.com| myendpoint-123456789012.s3-accesspoint-fips.dualstack.us-gov-east-1.amazonaws.com |
|2| myendpoint-123456789012.s3-object-lambda.fips-us-gov-east-1.amazonaws.com| myendpoint-123456789012.s3-object-lambda-fips.us-gov-east-1.amazonaws.com |

{accesspoint-name}-{account-id}.s3-accesspoint[-fips][.dualstack].{region}.{partition}

## Testing

- Added Junits for above mentioned scenarios,

## Types of changes

## Checklist

## License

### DIFF
--- a/.changes/next-release/bugfix-AWSS3-c21fe33.json
+++ b/.changes/next-release/bugfix-AWSS3-c21fe33.json
@@ -1,0 +1,6 @@
+{
+    "category": "AWS S3", 
+    "contributor": "", 
+    "type": "bugfix", 
+    "description": "Do not allow cross-region requests for S3 client with Fips Region."
+}
--- a/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/endpoints/S3AccessPointEndpointResolver.java
+++ b/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/endpoints/S3AccessPointEndpointResolver.java
@@ -19,7 +19,6 @@ import static software.amazon.awssdk.services.s3.internal.endpoints.S3EndpointUt
 import static software.amazon.awssdk.services.s3.internal.endpoints.S3EndpointUtils.isArnRegionEnabled;
 import static software.amazon.awssdk.services.s3.internal.endpoints.S3EndpointUtils.isDualstackEnabled;
 import static software.amazon.awssdk.services.s3.internal.endpoints.S3EndpointUtils.isFipsRegion;
-import static software.amazon.awssdk.services.s3.internal.endpoints.S3EndpointUtils.isFipsRegionProvided;
 import static software.amazon.awssdk.services.s3.internal.endpoints.S3EndpointUtils.isPathStyleAccessEnabled;
 import static software.amazon.awssdk.services.s3.internal.endpoints.S3EndpointUtils.removeFipsIfNeeded;
 
@@ -122,6 +121,9 @@ public final class S3AccessPointEndpointResolver implements S3EndpointResolver {
             "An S3 access point ARN must have a region"));
 
         S3Configuration serviceConfiguration = context.serviceConfiguration();
+
+        validateFipsRegionConfiguration(region, arnRegion);
+
         if (isAccelerateEnabled(serviceConfiguration)) {
             throw new IllegalArgumentException("An access point ARN cannot be passed as a bucket parameter to an S3 "
                                                + "operation if the S3 client has been configured with accelerate mode"
@@ -134,6 +136,7 @@ public final class S3AccessPointEndpointResolver implements S3EndpointResolver {
                                                + "addressing enabled.");
         }
 
+
         if (!isArnRegionEnabled(serviceConfiguration) && clientRegionDiffersFromArnRegion(region, arnRegion)) {
             throw new IllegalArgumentException(
                 String.format("The region field of the ARN being passed as a bucket parameter to an S3 operation "
@@ -142,6 +145,7 @@ public final class S3AccessPointEndpointResolver implements S3EndpointResolver {
                               + "configuration when building the S3 client. Provided region: '%s'; client region:"
                               + " '%s'.", arnRegion, region));
         }
+
 
         String clientPartition = PartitionMetadata.of(region).id();
 
@@ -155,8 +159,20 @@ public final class S3AccessPointEndpointResolver implements S3EndpointResolver {
         return arnRegion;
     }
 
+    private void validateFipsRegionConfiguration(Region region, String arnRegion) {
+        validateIsTrue(!isFipsRegion(arnRegion),
+                "Invalid ARN, FIPS region is not allowed in ARN."
+                        + " Provided arn region: '" + arnRegion + "'.");
+
+        validateIsTrue(!(isFipsRegion(region.id()) && clientRegionDiffersFromArnRegion(region, arnRegion)),
+                String.format("The region field of the ARN being passed as a bucket parameter to an S3 operation "
+                        + "does not match the region the client was configured with. " +
+                        "Cross region access not allowed for fips region in client or arn."
+                        + " Provided region: '%s'; client region:'%s'.", arnRegion, region));
+    }
+
     private boolean clientRegionDiffersFromArnRegion(Region clientRegion, String arnRegion) {
-        return !removeFipsIfNeeded(clientRegion.id()).equals(removeFipsIfNeeded(arnRegion));
+        return !removeFipsIfNeeded(clientRegion.id()).equals(arnRegion);
     }
 
     private boolean illegalPartitionConfiguration(S3Resource s3Resource, String clientPartition) {
@@ -183,14 +199,13 @@ public final class S3AccessPointEndpointResolver implements S3EndpointResolver {
         }
 
         boolean dualstackEnabled = isDualstackEnabled(context.serviceConfiguration());
-        boolean fipsRegionProvided = isFipsRegionProvided(context.region().toString(), arnRegion,
-                                                          isArnRegionEnabled(context.serviceConfiguration()));
+        boolean fipsRegionProvided = isFipsRegion(context.region().toString());
         return S3AccessPointBuilder.create()
                                    .endpointOverride(context.endpointOverride())
                                    .accessPointName(accessPointName)
                                    .accountId(accountId)
                                    .fipsEnabled(fipsRegionProvided)
-                                   .region(removeFipsIfNeeded(arnRegion))
+                                   .region(arnRegion)
                                    .protocol(context.request().protocol())
                                    .domain(clientPartitionMetadata.dnsSuffix())
                                    .dualstackEnabled(dualstackEnabled)
@@ -260,5 +275,11 @@ public final class S3AccessPointEndpointResolver implements S3EndpointResolver {
         }
 
         return Optional.empty();
+    }
+
+    private void validateIsTrue(boolean condition, String error, Object... params) {
+        if (!condition) {
+            throw new IllegalArgumentException(String.format(error, params));
+        }
     }
 }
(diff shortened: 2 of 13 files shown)


.changes/next-release/bugfix-AWSS3-c21fe33.json
```diff
@@ -0,0 +1,6 @@
+{
+    "category": "AWS S3", 
+    "contributor": "", 
+    "type": "bugfix", 
+    "description": "Donot allow cross-region requests for S3 client with Fips Region."
```
millems, line 5: Do not*

services/s3/src/test/java/software/amazon/awssdk/services/s3/S3EndpointResolutionTest.java
```diff
@@ -508,4 +509,62 @@ private S3Configuration withDualstackAndAccelerateEnabled() {
...
+    @Test
+    public void accessPointArn_us_gov_west_1_clientRegion_fips_us_gov_east_1_useArnRegionTrue() throws Exception {
+        mockHttpClient.stubNextResponse(mockListObjectsResponse());
+        S3Client s3Client = clientBuilder().region(Region.of("fips-us-gov-east-1"))
+                .serviceConfiguration(S3Configuration.builder().useArnRegionEnabled(true).build()).build();
+        String accessPointArn = "arn:aws-us-gov:s3:us-gov-west-1:123456789012:accesspoint:myendpoint";
+
+        assertThatThrownBy(() -> s3Client.getObject(GetObjectRequest.builder().bucket(accessPointArn).key("someKey").build()))
+                .isInstanceOf(IllegalArgumentException.class)
+                .hasMessageContaining("region");
```
millems, line 554: useArnRegionEnabled is true for here. What's the error message that's thrown? Is it still the "enable use-arn-region" one?

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/endpoints/S3AccessPointEndpointResolver.java
```diff
@@ -134,13 +134,15 @@ private String validateConfiguration(S3EndpointResolverContext context, S3Resour
...
 
-        if (!isArnRegionEnabled(serviceConfiguration) && clientRegionDiffersFromArnRegion(region, arnRegion)) {
-            throw new IllegalArgumentException(
-                String.format("The region field of the ARN being passed as a bucket parameter to an S3 operation "
-                              + "does not match the region the client was configured with. To enable this "
-                              + "behavior and prevent this exception set 'useArnRegionEnabled' to true in the "
-                              + "configuration when building the S3 client. Provided region: '%s'; client region:"
-                              + " '%s'.", arnRegion, region));
+        if (!isArnRegionEnabled(serviceConfiguration) || isFipsRegion(region.toString())) {
+            if (clientRegionDiffersFromArnRegion(region, arnRegion)) {
```
millems, line 138: clientRegionDiffersFromArnRegion currently strips "fips" from arn regions as well, but arn regions should never contain fips, correct?

We probably don't want people setting the client region to non-fips, then modifying the ARN region to fips and having it successfully go to a non-fips endpoint.

joviegas: Agree.. Using similar check as [isFipsRegionProvided](https://github.com/aws/aws-sdk-java-v2/blob/584ccb59e770177aeaa4c3b6bda4e24015b8ece9/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/endpoints/S3EndpointUtils.java#L56-L61) , but I cannot use this method directlty since this method relies on useArnRegion to check if the arn is fips.
And it will fail for the below case
arn.--> fips region client --> non fips  and useArnRegion enabled.

services/s3/src/test/java/software/amazon/awssdk/services/s3/S3EndpointResolutionTest.java
joviegas: Adding separate error message for fips case

services/s3/src/test/java/software/amazon/awssdk/services/s3/S3EndpointResolutionTest.java
```diff
@@ -508,4 +512,74 @@ private S3Configuration withDualstackAndAccelerateEnabled() {
...
+    public void accessPointArnRegion_fips_clientRegion_non_fips_useArnRegionTrue() throws Exception {
+        mockHttpClient.stubNextResponse(mockListObjectsResponse());
+        S3Client s3Client = clientBuilder().region(Region.of("us-west-1"))
+                .serviceConfiguration(S3Configuration.builder().useArnRegionEnabled(true).build()).build();
+        String accessPointArn = "arn:aws-us-gov:s3:fips-us-gov-east-1:123456789012:accesspoint:myendpoint";
+
+        assertThatThrownBy(() -> s3Client.getObject(GetObjectRequest.builder().bucket(accessPointArn).key("someKey").build()))
+                .isInstanceOf(IllegalArgumentException.class)
+                .hasMessageContaining(CROSS_REGION_FIPS_ERROR);
+    }
```
millems, line 582: Can we add a test similar to this one, but with a client region of us-gov-east-1 (and an ARN region of fips-us-gov-east-1)? The customer would probably expect this call to go to the fips endpoint, so it might break compliance for them if it goes to a non-fips endpoint.

I'd honestly vote for throwing an exception in this case, but I'd be okay with it going to the FIPS endpoint as well. Either way, we should align that behavior with the other SDKs.

joviegas: The scenario mentioned by you is same as Scenario5 as mentioned in above table in Motivation and Context section.
Current V2 code does not gets the partition correctly for Fips region/Gov regions and assigns them as default partition "aws'. And when we do a partition check which is right below Region match check it fails.

millems: I'm suggesting the inverse of scenario 5, swapping the client and arn region columns. It might be more interesting if it's in the AWS partition, so something like fips-us-west-2 for the ARN region and us-west-2 for the client region.

joviegas: Got it . It successfully resolved endpoints in this case. Added  test cases for this.

services/s3/src/test/java/software/amazon/awssdk/services/s3/S3EndpointResolutionTest.java
```diff
@@ -508,4 +512,100 @@ private S3Configuration withDualstackAndAccelerateEnabled() {
...
+    public void accessPointArn_non_region_fips_useArnRegionTrue() throws Exception {
+        mockHttpClient.stubNextResponse(mockListObjectsResponse());
+        S3Client s3Client = clientBuilder().region(Region.of("us-west-2-fips"))
+                .serviceConfiguration(S3Configuration.builder().useArnRegionEnabled(true).build()).build();
+        String accessPointArn = "arn:aws:s3:us-west-2:12345678910:accesspoint:foobar";
+        s3Client.getObject(GetObjectRequest.builder().bucket(accessPointArn).key("someKey").build());
+        assertThat(mockHttpClient.getLastRequest().getUri().getHost())
+                .isEqualTo("foobar-12345678910.s3-accesspoint.us-west-2.amazonaws.com");
+
+    }
```
millems, line 595: 1. Shouldn't this resolve to the FIPS endpoint? The customer has configured the client for FIPS, I thought the point of these changes were to make sure that if the customer thinks they're using FIPS that we shouldn't ever use a non-FIPS endpoint because it might result in the customer accidentally being uncompliant with FIPS.
2. This wasn't really the test I was asking for. I was interested in seeing where the ARN region is us-west-2-fips, not the client region. This is no different than scenario 5.

joviegas: Thanks Matt , I was too focused on Not allowing the request if mismatch and missed to observe the resolved endpoints for this case.


1. Yeap I looked at the original table and found that it should actually resolve to fips
The issue is because of [this code](https://github.com/aws/aws-sdk-java-v2/blob/584ccb59e770177aeaa4c3b6bda4e24015b8ece9/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/endpoints/S3EndpointUtils.java#L56-L61) where it says isFipsRegionProvided is false if useArnRegion  is true and  Arn not in fips and even though client is in fips
Will correct this.

2.  The actual spec table does not mention about the case where to resolve the endpoints if Client is Non fips and Arn is in fips. I will try to get the requirement.
3. 



```
    // "arn:aws:s3:fips-us-west-1:123456789012:accesspoint:myendpoint"
    //"us-west-1"
    // "myendpoint-123456789012.s3-accesspoint.us-west-1.amazonaws.com"
    // "myendpoint-123456789012.s3-accesspoint.fips.us-west-1.amazonaws.com"   
```

joviegas: Thanks Matt , updated the PR with the fips endpoint.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/resource/S3AccessPointBuilder.java
```diff
@@ -121,11 +121,11 @@ public URI toUri() {
...
         String uri;
         if (endpointOverride == null) {
-            String fipsSegment = Boolean.TRUE.equals(fipsEnabled) ? "fips-" : "";
+            String fipsSegment = Boolean.TRUE.equals(fipsEnabled) ? "-fips" : "";
             String dualStackSegment = Boolean.TRUE.equals(dualstackEnabled) ? ".dualstack" : "";
 
-            uri = String.format("%s://%s-%s.s3-accesspoint%s.%s%s.%s", protocol, urlEncode(accessPointName),
-                                accountId, dualStackSegment, fipsSegment, region, domain);
+            uri = String.format("%s://%s-%s.s3-accesspoint%s%s.%s.%s", protocol, urlEncode(accessPointName),
+                                accountId, fipsSegment, dualStackSegment, region, domain);
```
millems, line 128: Haha, whoops! It was totally broken...

millems (approved):

services/s3/src/test/java/software/amazon/awssdk/services/s3/S3EndpointResolutionTest.java
joviegas: For ARN region of fips-us-gov-east , now we throw exception ```Invalid ARN, FIPS region not allowed in ARN```

PR #2483, opened 2021-05-14
refactor: remove jackson-databind dependency

### LINKED ISSUE #2466: Remove Jackson-databind requirement, make databinding pluggable
## Describe the Feature

Jackson databind is required by SDK-core package. Jackson databind adds ~10MB to the runtime memory requirement and should be replaceable by jackson-jr or other smaller databinding packages such as gson. I believe that only the JacksonUtils class would need to be changed https://github.com/aws/aws-sdk-java-v2/blob/584ccb59e770177aeaa4c3b6bda4e24015b8ece9/core/sdk-core/src/main/java/software/amazon/awssdk/core/util/json/JacksonUtils.java#L32, as long as the API was changed so that `JsonNode` type isn't needed. The parent class `TreeNode` should be fine as a replacement.

## Is your Feature Request related to a problem?
The AWS SDK adds a large amount of additional memory requirement which is a large requirement for IoT use cases.

## Proposed Solution
Just like how the SDK HTTP clients are pluggable, jackson-databind should not be a hard requirement. Even better, Jackson-core shouldn't be required and instead there should be APIs for generic serialization and deserialization.

## Additional Context

- [x] I may be able to implement this feature request

## Your Environment

* AWS Java SDK version used:
* JDK version used:

MikeDombo:
## Description
This PR removes the dependency on jackson-databind from the sdk-core package. It is replaced with jackson-jr which is faster to initialize and uses significantly less memory. With this change, our application (AWS Greengrass Nucleus) is able to save 6MB of RAM at runtime and 2MB of disk space. This may sound trivial, but it is nearly a 10% improvement, which is really meaningful for our IoT use cases.

While this does _technically_ change the `JacksonUtils` interface, I believe that it is considered internal, and thus changing it should be alright.

## Motivation and Context
Closes #2466

## Testing
All tests are passing locally.

## Types of changes

## Checklist

## License

### DIFF
--- a/.changes/next-release/feature-AWSSDKforJavav2-4e08f09.json
+++ b/.changes/next-release/feature-AWSSDKforJavav2-4e08f09.json
@@ -1,0 +1,6 @@
+{
+    "category": "AWS SDK for Java v2", 
+    "contributor": "MikeDombo", 
+    "type": "feature", 
+    "description": "Removes usage of Jackson-Databind and replaces it with Jackson-jr in order to reduce memory requirement"
+}
--- a/bom-internal/pom.xml
+++ b/bom-internal/pom.xml
@@ -50,6 +50,11 @@
                 <version>${jacksonjr.version}</version>
             </dependency>
             <dependency>
+                <groupId>com.fasterxml.jackson.jr</groupId>
+                <artifactId>jackson-jr-annotation-support</artifactId>
+                <version>${jacksonjr.version}</version>
+            </dependency>
+            <dependency>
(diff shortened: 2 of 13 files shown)


debora-ito:
Hi @MikeDombo thank you for taking the time to submit this. I'll brig it up to the team for discussion.

MikeDombo:
Thank you @debora-ito. Note that this PR as provided requires jackson-jr 2.13 which isn't yet released, but it will be soonish. The reason why 2.13 is required right now is to support case insensitive property matching.

Issue #1881, opened 2020-06-08
Ability to customize stsClient in `StsWebIdentityCredentialsProvider` 


marcinczeczko:
At the moment if using client for any AWS service with `WebIdentityTokenFileCredentialsProvider` as credentials provider it forces to use sync HTTP client. Additionally, STS client instead always assumes region based on the `DefaultAwsRegionProviderChain` without the ability to override it via the code. In majority cases it's fine as it can be controlled at some extent via system properties or env variables. However, the use case I have
-  Using e.g. DynamoDB Async client with `WebIdentityTokenFileCredentialsProvider` as credentials provider
- It forces me to add any sync HTTP client (URL connection or Apache HTTP client) on the classpath, as the `StsWebIdentityCredentialsProviderFactory` uses sync client only.
- Additionally, if my code uses Dynamodb client has configured that client to use specific region via the code, I need to repeat the same via sysprops or env vars for sts client.
- For local testing I can't define STS endpoint override as there is no way to customize/configure sts client used by credentials provider.

## Describe the Feature
Ability to specify STS client built in my code or STS client builder in the `WebIdentityTokenFileCredentialsProvider.Builder`

## Is your Feature Request related to a problem?
Explained at the top

## Proposed Solution
- The best would be to extend the `WebIdentityTokenFileCredentialsProvider.Builder` to accept stsClient built somewhere else or stsClient builder instead, so I can build either sync or async client.

## Describe alternatives you've considered
N/A

## Additional Context
I was implementing Quarkus extensions to support AWS SDK v2 client and the web identity token file credentials provider causes troubles to make it easily configurable. And since `StsWebIdentityCredentialsProviderFactory` is final with private methods it's hard to override it's logic to control how stsClient is being created.

- [x] I may be able to implement this feature request

## Your Environment
* AWS Java SDK version used: 2.13.27
* JDK version used: 1.8, 11, Graalvm 19.3.1
* Operating System and version: macOs Catalina

debora-ito:
It's a reasonable ask, marking as a feature request.

nhoughto:
It also prevents using a non-AWS STS endpoint, we are trying to override it for local testing (calling localstack) and the current approach makes that hard.

hawkpatrick:
[...]

PR #2488, opened 2021-05-19
Adding New StsWebIdentityTokenFileCredentialsProvider in sts module that accepts STSClient

joviegas:
Adding New WebIdentityTokenFileCredentialsProvider  in sts that accepts STSClient

## Description

- Currently the WebIdentityTokenFileCredentialsProvider is in core module and it creates AwsCredentialsProvider with Default Sts Client.
- We cannot modify the [WebIdentityTokenCredentialsProviderFactory](https://github.com/aws/aws-sdk-java-v2/blob/584ccb59e770177aeaa4c3b6bda4e24015b8ece9/core/auth/src/main/java/software/amazon/awssdk/auth/credentials/WebIdentityTokenCredentialsProviderFactory.java) because  the interface is marked as @FunctionalInterface  and also StsClient is not available in core package so would need to pass SdkClient which would be confusing.
- Thus we have added a new WebIdentityTokenFileCredentialsProvider in Sts module and named it as StsWebIdentityTokenFileCredentialsProvider

### Usage

1. Credential provider that will read web identity token file path, aws role arn, aws session name from system properties or environment variables And DefaultStsClient
 ```
 StsWebIdentityTokenFileCredentialsProvider.create();
 ```

2. Credential provider that will read web identity token file path, aws role arn, aws session name from system properties or environment variables And StsClient will be used as it is passed in the constructor.

 ```
 StsWebIdentityTokenFileCredentialsProvider.create(stsClient);
 ```

3 . Credential provider that will read web identity token file path, aws role arn, aws session name from Builders And StsClient will be used as it is passed in the Builder.
 ```
                StsWebIdentityTokenFileCredentialsProvider.builder()
                        .stsClient(stsClient)
                        .roleArn("someRole")
                        .webIdentityTokenFile(Paths.get(webIdentityTokenPath))
                        .roleSessionName("tempRoleSession")
                        .build();
 ```

## Motivation and Context

- Please refer Pr #1881 

## Testing

- Added JUnit Test cases.

## Types of changes

## Checklist

## License

### DIFF
--- a/services/sts/src/main/java/software/amazon/awssdk/services/sts/internal/StsWebIdentityCredentialsProviderFactory.java
+++ b/services/sts/src/main/java/software/amazon/awssdk/services/sts/internal/StsWebIdentityCredentialsProviderFactory.java
@@ -50,30 +50,39 @@ import software.amazon.awssdk.utils.SdkAutoCloseable;
 public final class StsWebIdentityCredentialsProviderFactory implements WebIdentityTokenCredentialsProviderFactory {
 
     public AwsCredentialsProvider create(WebIdentityTokenCredentialProperties credentialProperties) {
-        return new StsWebIdentityCredentialsProvider(credentialProperties);
+        return new StsWebIdentityCredentialsProvider(credentialProperties, null);
+    }
+
+    public AwsCredentialsProvider create(WebIdentityTokenCredentialProperties credentialProperties,
+                                         StsClient stsClient) {
+        return new StsWebIdentityCredentialsProvider(credentialProperties, stsClient);
     }
 
     /**
      * A wrapper for a {@link StsAssumeRoleWithWebIdentityCredentialsProvider} that is returned by this factory when
      * {@link #create(WebIdentityTokenCredentialProperties)} is invoked. This wrapper is important because it ensures the parent
      * credentials provider is closed when the assume-role credentials provider is no longer needed.
+     * It creates default StsClient if StsClient is null or if StsClient is not passed while creating the instance.
      */
     private static final class StsWebIdentityCredentialsProvider implements AwsCredentialsProvider, SdkAutoCloseable {
         private final StsClient stsClient;
         private final StsAssumeRoleWithWebIdentityCredentialsProvider credentialsProvider;
 
-        private StsWebIdentityCredentialsProvider(WebIdentityTokenCredentialProperties credentialProperties) {
+        private StsWebIdentityCredentialsProvider(WebIdentityTokenCredentialProperties credentialProperties,
+                                                  StsClient stsClient) {
             String roleSessionName = credentialProperties.roleSessionName();
             String sessionName = roleSessionName != null ? roleSessionName : "aws-sdk-java-" + System.currentTimeMillis();
 
             OrRetryCondition retryCondition = OrRetryCondition.create(new StsRetryCondition(),
                                                                       RetryCondition.defaultRetryCondition());
 
-            this.stsClient = StsClient.builder()
+            this.stsClient = stsClient == null ?
+                    StsClient.builder()
                                       .applyMutation(this::configureEndpoint)
                                       .credentialsProvider(AnonymousCredentialsProvider.create())
                                       .overrideConfiguration(o -> o.retryPolicy(r -> r.retryCondition(retryCondition)))
-                                      .build();
+                                      .build()
+                    : stsClient;
 
             AssumeRoleWithWebIdentityRequest request = AssumeRoleWithWebIdentityRequest.builder()
                                                                                        .roleArn(credentialProperties.roleArn())
@@ -86,7 +95,7 @@ public final class StsWebIdentityCredentialsProviderFactory implements WebIdenti
 
             this.credentialsProvider =
                 StsAssumeRoleWithWebIdentityCredentialsProvider.builder()
-                                                               .stsClient(stsClient)
+                                                               .stsClient(this.stsClient)
                                                                .refreshRequest(supplier)
                                                                .build();
         }
--- a/services/sts/src/main/java/software/amazon/awssdk/services/sts/internal/StsWebIdentityTokenFileCredentialsProvider.java
+++ b/services/sts/src/main/java/software/amazon/awssdk/services/sts/internal/StsWebIdentityTokenFileCredentialsProvider.java
@@ -1,0 +1,192 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.services.sts.internal;
+
+import static software.amazon.awssdk.utils.StringUtils.trim;
+
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Optional;
+import software.amazon.awssdk.annotations.SdkPublicApi;
+import software.amazon.awssdk.auth.credentials.AwsCredentials;
+import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider;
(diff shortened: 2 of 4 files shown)


millems:
Thinking out-loud here: I'm trying to make sense out of the different options we provide to customers in terms of web identity token credentials providers.

Today, we have:
1. The `DefaultCredentialsProvider`, which is just a proxy for `WebIdentityTokenFileCredentialsProvider` with the right configuration
2. The `WebIdentityTokenFileCredentialsProvider`, which loads credentials based on three environment variables or system properties (and allows direct configuration): AWS_WEB_IDENTITY_TOKEN_FILE, AWS_ROLE_ARN, AWS_ROLE_SESSION_NAME
3. The `StsAssumeRoleWithWebIdentityCredentialsProvider`, which loads credentials based on any values in a `AssumeRoleWithWebIdentityRequest` (including those 3 above in the WebIdentityTokenFileCredentialsProvider, except that the web identity token is a string instead of a file from which it should be read).

This is proposing adding a fourth option:
4. The `StsWebIdentityTokenFileCredentialsProvider` that is essentially a combination of (2) and (3) - it's configured in the same way as (2), but it allows configuring the STS client like (3) does.

This feels like a lot of options. We talked about deprecating (2) since it's strictly less powerful than (4), and that seems reasonable, but I'm wondering if we can do it in a way that doesn't add even more options for the customer. Is there a way we can reuse (3)?

I mean, (4) is pretty much just short for:
```java
StsAssumeRoleWithWebIdentityCredentialsProvider.builder()
    .refreshRequest(r -> r.roleArn(SdkSystemSetting.AWS_ROLE_ARN.getStringValueOrThrow())
                          .roleSessionName(SdkSystemSetting.AWS_ROLE_SESSION_NAME.getStringValueOrThrow())
                          .webIdentityToken(readFile(SdkSystemSetting.AWS_WEB_IDENTITY_TOKEN_FILE.getStringValueOrThrow())))
    .build();
```

That provides a benefit, but do we really need a new credentials provider to do it?

We could do something like ...

```java
StsAssumeRoleWithWebIdentityCredentialsProvider.builder()
    .refreshRequest(WebIdentityTokenFileRefreshRequestSupplier.create())
    .build()
```

I don't know if that's better or worse than

```java
StsWebIdentityTokenFileCredentialsProvider.create()
```

What do you think? My gut says that this CR is probably correct, and we should just add (4) and deprecate (2).

services/sts/src/main/java/software/amazon/awssdk/services/sts/internal/StsWebIdentityCredentialsProviderFactory.java
```diff
@@ -50,30 +50,39 @@
 public final class StsWebIdentityCredentialsProviderFactory implements WebIdentityTokenCredentialsProviderFactory {
 
     public AwsCredentialsProvider create(WebIdentityTokenCredentialProperties credentialProperties) {
-        return new StsWebIdentityCredentialsProvider(credentialProperties);
+        return new StsWebIdentityCredentialsProvider(credentialProperties, null);
+    }
+
+    public AwsCredentialsProvider create(WebIdentityTokenCredentialProperties credentialProperties,
+                                         StsClient stsClient) {
+        return new StsWebIdentityCredentialsProvider(credentialProperties, stsClient);
```
millems, line 58: Do we need to add this? Create methods are usually to make things more convenient for our customers, but this is a protected API that they won't use, so I'd prefer just to rely on the builder.

services/sts/src/main/java/software/amazon/awssdk/services/sts/internal/StsWebIdentityCredentialsProviderFactory.java
```diff
@@ -50,30 +50,39 @@
...
     }
 
     /**
      * A wrapper for a {@link StsAssumeRoleWithWebIdentityCredentialsProvider} that is returned by this factory when
      * {@link #create(WebIdentityTokenCredentialProperties)} is invoked. This wrapper is important because it ensures the parent
      * credentials provider is closed when the assume-role credentials provider is no longer needed.
+     * It creates default StsClient if StsClient is null or if StsClient is not passed while creating the instance.
      */
     private static final class StsWebIdentityCredentialsProvider implements AwsCredentialsProvider, SdkAutoCloseable {
```
millems, line 67: This feels like a copy of StsAssumeRoleWithWebIdentityTokenCredentialsProvider with extra construction logic. I don't see why this has to be a class.

Two questions:
1. Should we just copy the configuration logic into the StsWebIdentityCredentialsProviderFactory, and just use the StsAssumeRoleWithWebIdentityTokenCredentialsProvider?
2. If we allow customers to override the StsClient, won't that remove the StsRetryCondition that they need? Should we move that StsRetryCondition to be retried universally so that customers won't override it by accident?

services/sts/src/main/java/software/amazon/awssdk/services/sts/internal/StsWebIdentityTokenFileCredentialsProvider.java
```diff
@@ -0,0 +1,192 @@
...
+import software.amazon.awssdk.core.SdkSystemSetting;
+import software.amazon.awssdk.services.sts.StsClient;
+import software.amazon.awssdk.utils.ToString;
+
+
+/**
+ * A credential provider that will read web identity token file path, aws role arn, aws session name
+ * from system properties or environment variables  and StsClient for using
+ * web identity token credentials with STS. If StsClient is not passed or null then it will use Default Sts client.
+ */
```
millems, line 36: Can we compare and contrast this with the other web identity credential providers and specify here (and in them) when to use each one?

services/sts/src/main/java/software/amazon/awssdk/services/sts/internal/StsWebIdentityTokenFileCredentialsProvider.java
```diff
@@ -0,0 +1,192 @@
...
+                                                    .build();
+
+            credentialsProviderLocal = new StsWebIdentityCredentialsProviderFactory().create(credentialProperties,
+                    builder.stsClient);
+        } catch (RuntimeException e) {
+            // If we couldn't load the credentials provider for some reason, save an exception describing why. This exception
+            // will only be raised on calls to getCredentials. We don't want to raise an exception here because it may be
+            // expected (eg. in the default credential chain).
+            loadExceptionLocal = e;
+        }
```
millems, line 74: Do we need to use the StsWebIdentityCredentialsProviderFactory here? Can we just move the logic from StsWebIdentityCredentialsProviderFactory into this class and call this class from that one? It's hard for me to wrap my head around these factory-generates-a-delegate-for-me-to-use approach.

services/sts/src/main/java/software/amazon/awssdk/services/sts/internal/StsWebIdentityTokenFileCredentialsProvider.java
```diff
@@ -0,0 +1,192 @@
...
+import software.amazon.awssdk.utils.ToString;
+
+
+/**
+ * A credential provider that will read web identity token file path, aws role arn, aws session name
+ * from system properties or environment variables  and StsClient for using
+ * web identity token credentials with STS. If StsClient is not passed or null then it will use Default Sts client.
+ */
+@SdkPublicApi
+public class StsWebIdentityTokenFileCredentialsProvider implements AwsCredentialsProvider {
```
millems, line 38: Should this extend StsCredentialsProvider like StsAssumeRoleWithWebIdentityCredentialsProvider does?

services/sts/src/main/java/software/amazon/awssdk/services/sts/internal/StsWebIdentityTokenFileCredentialsProvider.java
joviegas: Added a comment about this in StsWebIdentityTokenFileCredentialsProvider and WebIdentityTokenFileCredentialsProvider

services/sts/src/main/java/software/amazon/awssdk/services/sts/internal/StsWebIdentityTokenFileCredentialsProvider.java
joviegas: Agree, moved it now

services/sts/src/main/java/software/amazon/awssdk/services/sts/internal/StsWebIdentityTokenFileCredentialsProvider.java
joviegas: doen , new changes in StsWebIdentityTokenFileCredentialsProvider

joviegas:
Please refer https://github.com/aws/aws-sdk-java-v2/pull/3220

Issue #2283, opened 2021-02-15
DynamoDbEnhancedClient transaction write missing support for return values


openwebsolns:
Support for [ReturnValuesOnConditionCheckFailure](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_Put.html#DDB-Type-Put-ReturnValuesOnConditionCheckFailure) is missing from the `PutItemEnhancedRequest` object, but exists in the vanilla `PutItemRequest`.

## Describe the bug

Unlike V1's `DynamoDBMapper` or V2's plain `DynamoDbClient`, there is no way to specify `returnValues` on a put item request in `DynamoDbEnhancedClient`. Specifically, the `PutItemEnhancedRequest` object [is missing the field](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/enhanced/dynamodb/model/PutItemEnhancedRequest.Builder.html) which otherwise exists in [PutItemRequest](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/dynamodb/model/PutItemRequest.Builder.html#returnValues-software.amazon.awssdk.services.dynamodb.model.ReturnValue-).

## Expected Behavior

I should be able to specify `ALL_OLD` (or `NONE`) for return values in a transaction write using `DynamoDbEnhancedClient`:

```java
TransactionWriteItemsEnhancedRequest request = TransactionWriteItemsEnhancedRequest.builder()
    .addPutItem(
        itemTable, // MappedTableResource
        PutItemEnhancedRequest.builder()
            .item(someItem)
            .conditionExpression(someExpression)
            .returnValues(ReturnValues.ALL_OLD)    // <<<< MISSING FUNCTIONALITY
            .build()
    ).build();
```

## Current Behavior

There is no way to specify `returnValues` [as can be done in plain PutItemRequest](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/dynamodb/model/PutItemRequest.Builder.html#returnValues-software.amazon.awssdk.services.dynamodb.model.ReturnValue-).

## Steps to Reproduce

See above.

## Possible Solution

Add support as part of the `PutItemEnhancedRequest`; not sure if depending on `services.dynamodb` package is considered correct when working in `enhanced.dynamodb`.

## Context

Unable to migrate from `DynamoDBMapper` in V1 to `DynamoDbEnhancedClient` because of this missing functionality.

## Your Environment

* AWS Java SDK version used:
[...]

PR #2503, opened 2021-05-28
DynamoDB Enhanced: Adds support for configuring the ReturnValue for PutItem and UpdateItem

marcdejonge:
Adds support for configuring the ReturnValue for PutItem and UpdateItem

## Description
When executing a PutItem or UpdateItem request, there is an option to change the return value. This can be useful in many situation, for example to see what the values were before the change was applied.

## Motivation and Context
#2283 

In our project, we'd like to execute an UpdateItem while only getting the old values back of fields that have actually changed. With the current enhanced client this is actually not possible. That's why this change is needed.

## Testing
- Added a Unit test for the Request themselves
- Added a Unit test for the relevant Operation object
- Manually tried out the change in a personal project

## Types of changes

## Checklist

## License

### DIFF
--- a/.changes/feature-AWSDynamoDBEnhanced-2283.json
+++ b/.changes/feature-AWSDynamoDBEnhanced-2283.json
@@ -1,0 +1,6 @@
+{
+  "category": "DynamoDB Enhanced Client",
+  "contributor": "marcdejonge",
+  "type": "feature",
+  "description": "Adds the option to change the ReturnValue for PutItemEnhancedRequest and UpdateItemEnhancedRequest."
+}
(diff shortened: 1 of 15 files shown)


marcdejonge:
This seems to work for the UpdateItem request just fine, but for the PutItem I can't return the old value right now. This requires an API change, where the putItem functions in the DynamoDbTable should return something (right now they are void). What is the process to change this?

marcdejonge:
Anyone able to help what should change here? I guess changing the API like this won't work (that's the reason the build is failing right now). For the UpdateItem it's fairly straightforward, but for the PutItem maybe we should add another function?

debora-ito:
Hi @marcdejonge I'm sorry for the long silence here in the PR.

Yes, the changes you proposed are breaking changes, that's probably why the build was failing.

Since then we implemented new methods `putItemWithResponse`, `updateItemWithResponse` and `deleteItemWithResponse` that should fulfill this request. Closing this PR.

PR #2512, opened 2021-06-04
Add user agent for transfer manager

zoewangg:
## Description
Add user agent for transfer manager

## License

### DIFF
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/util/SdkUserAgent.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/util/SdkUserAgent.java
@@ -13,15 +13,15 @@
  * permissions and limitations under the License.
  */
 
-package software.amazon.awssdk.core.internal.util;
+package software.amazon.awssdk.core.util;
 
 import java.util.Optional;
 import java.util.jar.JarInputStream;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.annotations.SdkProtectedApi;
+import software.amazon.awssdk.annotations.SdkTestInternalApi;
 import software.amazon.awssdk.annotations.ThreadSafe;
-import software.amazon.awssdk.core.util.VersionInfo;
 import software.amazon.awssdk.utils.IoUtils;
 import software.amazon.awssdk.utils.JavaSystemSetting;
 
@@ -29,8 +29,8 @@ import software.amazon.awssdk.utils.JavaSystemSetting;
  * Utility class for accessing AWS SDK versioning information.
  */
 @ThreadSafe
-@SdkInternalApi
-public final class UserAgentUtils {
+@SdkProtectedApi
+public final class SdkUserAgent {
 
     private static final String UA_STRING = "aws-sdk-{platform}/{version} {os.name}/{os.version} {java.vm.name}/{java.vm"
                                             + ".version} Java/{java.version}{language.and.region}{additional.languages} "
@@ -40,12 +40,27 @@ public final class UserAgentUtils {
     private static final String UA_BLACKLIST_REGEX = "[() ,/:;<=>?@\\[\\]{}\\\\]";
 
     /** Shared logger for any issues while loading version information. */
-    private static final Logger log = LoggerFactory.getLogger(UserAgentUtils.class);
+    private static final Logger log = LoggerFactory.getLogger(SdkUserAgent.class);
     private static final String UNKNOWN = "unknown";
-    /** User Agent info. */
-    private static volatile String userAgent;
+    private static volatile SdkUserAgent instance;
 
-    private UserAgentUtils() {
+    /** User Agent info. */
+    private String userAgent;
+
+    private SdkUserAgent() {
+        initializeUserAgent();
+    }
+
+    public static SdkUserAgent create() {
+        if (instance == null) {
+            synchronized (SdkUserAgent.class) {
+                if (instance == null) {
+                    instance = new SdkUserAgent();
+                }
+            }
+        }
+
+        return instance;
     }
 
     /**
@@ -53,14 +68,7 @@ public final class UserAgentUtils {
      *     the AWS services.  The User Agent encapsulates SDK, Java, OS and
      *     region information.
      */
-    public static String getUserAgent() {
-        if (userAgent == null) {
-            synchronized (UserAgentUtils.class) {
-                if (userAgent == null) {
-                    initializeUserAgent();
-                }
-            }
-        }
+    public String userAgent() {
         return userAgent;
     }
 
@@ -69,11 +77,12 @@ public final class UserAgentUtils {
      * {@code InternalConfig} and filling in the detected version/platform
      * info.
      */
-    private static void initializeUserAgent() {
-        userAgent = userAgent();
+    private void initializeUserAgent() {
+        userAgent = getUserAgent();
     }
 
-    static String userAgent() {
+    @SdkTestInternalApi
+    String getUserAgent() {
 
         String ua = UA_STRING;
 
--- a/services-custom/s3-transfermanager/src/main/java/software/amazon/awssdk/custom/s3/transfer/S3TransferManager.java
+++ b/services-custom/s3-transfermanager/src/main/java/software/amazon/awssdk/custom/s3/transfer/S3TransferManager.java
@@ -138,6 +138,7 @@ public interface S3TransferManager extends SdkAutoCloseable {
         default Builder s3ClientConfiguration(Consumer<S3ClientConfiguration.Builder> builderConsumer) {
             S3ClientConfiguration.Builder builder = S3ClientConfiguration.builder();
             builderConsumer.accept(builder);
+            s3ClientConfiguration(builder.build());
             return this;
         }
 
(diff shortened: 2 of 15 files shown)


services-custom/s3-transfermanager/src/main/java/software/amazon/awssdk/custom/s3/transfer/internal/S3CrtUtils.java
```diff
@@ -238,19 +237,28 @@ private static String encodedQueryString(Map<String, List<String>> rawQueryParam
...
+                String encodedQueryString = encodedQueryString(requestOverrideConfiguration.rawQueryParameters());
+                queryParametersConsumer.accept(encodedQueryString);
+            }
         }
+    }
 
-        if (!requestOverrideConfiguration.rawQueryParameters().isEmpty()) {
-            String encodedQueryString = encodedQueryString(requestOverrideConfiguration.rawQueryParameters());
-            queryParametersConsumer.accept(encodedQueryString);
+    private static void addCustomHeaders(Optional<AwsRequestOverrideConfiguration> requestOverrideConfiguration,
```
zoewangg, line 253: Using `Optional` as input param might be a bit controversial; I can't think of better ways for this method and it's a private method.

services-custom/s3-transfermanager/src/main/java/software/amazon/awssdk/custom/s3/transfer/internal/S3CrtUtils.java
```diff
@@ -48,6 +50,8 @@
 
 @SdkInternalApi
 public final class S3CrtUtils {
+    private static final String HEADER_USER_AGENT = "User-Agent";
+    private static final String USER_AGENT_STRING = "ft/s3-transfer";
```
millems, line 54: What does ft mean?

services-custom/s3-transfermanager/src/main/java/software/amazon/awssdk/custom/s3/transfer/internal/S3CrtUtils.java
```diff
@@ -238,19 +237,28 @@ private static String encodedQueryString(Map<String, List<String>> rawQueryParam
...
         }
+    }
 
-        if (!requestOverrideConfiguration.rawQueryParameters().isEmpty()) {
-            String encodedQueryString = encodedQueryString(requestOverrideConfiguration.rawQueryParameters());
-            queryParametersConsumer.accept(encodedQueryString);
+    private static void addCustomHeaders(Optional<AwsRequestOverrideConfiguration> requestOverrideConfiguration,
+                                         Consumer<HttpHeader[]> headersConsumer) {
+        List<HttpHeader> crtHeaders = new ArrayList<>();
+        crtHeaders.add(new HttpHeader(HEADER_USER_AGENT, USER_AGENT_STRING));
```
millems, line 256: Won't this override the other user-agent headers?

services-custom/s3-transfermanager/src/main/java/software/amazon/awssdk/custom/s3/transfer/internal/S3CrtUtils.java
zoewangg: `feature` per the SEP :)

services-custom/s3-transfermanager/src/main/java/software/amazon/awssdk/custom/s3/transfer/internal/S3CrtUtils.java
zoewangg: This is no existing `User-Agent` header set on the CRT side, I think. I'll double check with CRT team. S3 CRT client doesn't have APIs to append user agent string unfortunately.

services-custom/s3-transfermanager/src/main/java/software/amazon/awssdk/custom/s3/transfer/S3TransferManager.java
```diff
@@ -135,11 +135,7 @@ static S3TransferManager create() {
 
         Builder s3ClientConfiguration(S3ClientConfiguration configuration);
 
-        default Builder s3ClientConfiguration(Consumer<S3ClientConfiguration.Builder> builderConsumer) {
-            S3ClientConfiguration.Builder builder = S3ClientConfiguration.builder();
-            builderConsumer.accept(builder);
-            return this;
-        }
+        Builder s3ClientConfiguration(Consumer<S3ClientConfiguration.Builder> builderConsumer);
```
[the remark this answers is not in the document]
zoewangg: Yeah, the consumer builder method should be in the impl because otherwise `Default#s3ClientConfiguration` will be null and NPE will be thrown. (I should've created a separate PR for this change)

services-custom/s3-transfermanager/src/main/java/software/amazon/awssdk/custom/s3/transfer/internal/S3CrtUtils.java
millems: Ah, was looking at an old revision. Cool.

services-custom/s3-transfermanager/src/main/java/software/amazon/awssdk/custom/s3/transfer/internal/S3CrtUtils.java
zoewangg: Yeah, confirmed that the CRT S3 client will add a User-Agent header if one doesn't exist or append to the existing User-Agent header

millems: If the CRT doesn't set the user agent, we should set it. We don't want to lose all of the other information we usually include in the user agent...

zoewangg: Yeah, good point, will add other stuff as well

services-custom/s3-transfermanager/src/main/java/software/amazon/awssdk/custom/s3/transfer/S3TransferManager.java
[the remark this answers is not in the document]
millems: Couldn't we just call s3ClientConfiguration(builder.build())? That's what we usually do with other consumer-builder methods.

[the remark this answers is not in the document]
zoewangg: 😅 D'oh!

services-custom/s3-transfermanager/src/main/java/software/amazon/awssdk/custom/s3/transfer/internal/S3CrtUtils.java
zoewangg: Yeah, I thought about doing null check in this method because it feels a bit unnecessary with the existing optional object. But I agree using optional here is a code smell. I'll update it.

services-custom/s3-transfermanager/src/main/java/software/amazon/awssdk/custom/s3/transfer/internal/S3CrtUtils.java
```diff
@@ -48,6 +49,8 @@
 
 @SdkInternalApi
 public final class S3CrtUtils {
+    private static final String HEADER_USER_AGENT = "User-Agent";
+    private static final String USER_AGENT_STRING = UserAgentUtils.getUserAgent() + " ft/s3-transfer";
```
millems, line 53: UserAgentUtils is currently an internal API.

zoewangg: As discussed offline, changed it to protected API

core/sdk-core/src/main/java/software/amazon/awssdk/core/util/SdkUserAgent.java
```diff
@@ -40,27 +40,35 @@
...
+    }
+
+    public static SdkUserAgent create() {
+        if (instance == null) {
+            synchronized (SdkUserAgent.class) {
+                if (instance == null) {
+                    instance = new SdkUserAgent();
+                }
+            }
+        }
```
millems, line 61: Any reason to move to an instance instead of having the user just call `SdkUserAgent.getUserAgent()`?

zoewangg: Yeah, I feel utility class is in general not flexible and I personally would only use it for something that is truly stateless such as calculation or string manipulation. I'd be fine if it's an internal API, and now that it's protected API, I changed it to an instance class.

millems: Can you give an example of something in the future that we might want to do with this that would benefit from it being an instance?

zoewangg: For example, we might want to support retrieving the partial user agent string, say sdk version, via something like`sdkUserAgent.metadata().get("SdkVeresion")`. The main issue I have with this being a util class is that it holds a state and using static methods is not OOP. To me, a util class should be stateless, returning a result based on the input and nothing else.

millems: I usually would agree, but in this case the state that it stores is immutable, so I personally wouldn't call the class stateful. This feels closer to a class with a public static final field, except the value is loaded lazily so we have to go through a method to get to the data.

`SdkUserAgent.metadata().get("SdkVersion")` is still possible without metadata being an instance method.

The injectability could be useful, but I don't foresee us ever using it. It feels like we're just adding burden to the caller of calling create() before they call getUserAgent().

Either way, good chat! I don't think this is a blocker for me.

millems (approved):

Issue #2510, opened 2021-06-03
[Help]Getting error in WellArchitected aws service


niraj-ranjan-comprinno:
I am getting error while calling listWorkloads.
Error is ``Unsupported Media Type (Service: WellArchitected, Status Code: 415, Request ID: 8acd63db-8a7f-4adf-9188-3b060da968d6, Extended Request ID: null)``
## Steps to Reproduce

AwsCredentials credentials = AWSCustomCredentialHandler.getCredential(credential);

WellArchitectedClient wellArchitectedClient = WellArchitectedClient.builder()
.credentialsProvider(new AWSStaticCredentialsProvider(credentials)).region(Region.AP_SOUTH_1)
.build();

ListWorkloadsRequest.Builder listWorkloadsRequest = ListWorkloadsRequest.builder();
ListWorkloadsResponse listWorkloadsResponse = wellArchitectedClient.listWorkloads(listWorkloadsRequest.build());

## Your Environment

* AWS Java SDK version used:
* JDK version used:
* Operating System and version:

niraj-ranjan-comprinno:
@joviegas Please help

debora-ito:
Hi @niraj-ranjan-comprinno thank you for reporting this, we'll work on a fix.

joviegas:
Hi @niraj-ranjan-comprinno 
Appreciate your patience 👍 
The issue is fixed and available in  Release version 2.16.82.

niraj-ranjan-comprinno:
> Hi @niraj-ranjan-comprinno
> Appreciate your patience 
> The issue is fixed and available in Release version 2.16.82.

Hi @joviegas , Many many thanks

PR #2519, opened 2021-06-09
Support customization of contentTypeOverride for Header ContentType

joviegas:
## Description
- Support customization of contentTypeOverride for Header ContentType

## Motivation and Context

- By default the ContentType Header for Json type is set to  ```application/x-amz-json-1.1``` for json protocol.
- While wellarchitected accepted  ```application/json```  ContentType Header.
- These changes will add a customization for services which accept ```application/json```ContentType header.

Why is this change required? What problem does it solve? 
- Customization required for  ```application/json```ContentType header. Else the API calls to Services which accept  ```application/json```ContentType header will not be successful.

 If it fixes an open issue, please link to the issue here  :  PR #2510 

### JsonProtocolSpec.java :  
contentType attribute is populated only if CustomizationConfig is set. If not then we look for metadata.getContentType() else we donot add any ContentType, in this case it uses from DefaultJsonContentTypeResolver.

### AwsJsonProtocolMetadata
Added contentType which is used while we resolve the ContentType.

### BaseAwsJsonProtocolFactory
The Factory method that creates Builders for AwsJsonProtocolFactory. ContentType is added in the attributes which will be set by codegen. Here we select from ContentType if the ContentType is set else we select the default one as we used to do before.

### services/wellarchitected/src/main/resources/codegen-resources/customization.config 
The Config file to set wellarchitected services ContentType as application/json

### codegen/src/test/resources/software/amazon/awssdk/codegen/poet/client/test-customservicemetadata-async.java
Expected generated files when ContentType Customization is set.

Rest of the changes are for testing.

## Types of changes

## Checklist

## License

### DIFF
--- a/codegen/src/main/java/software/amazon/awssdk/codegen/poet/client/specs/JsonProtocolSpec.java
+++ b/codegen/src/main/java/software/amazon/awssdk/codegen/poet/client/specs/JsonProtocolSpec.java
@@ -33,6 +33,7 @@ import software.amazon.awssdk.awscore.eventstream.EventStreamAsyncResponseTransf
 import software.amazon.awssdk.awscore.eventstream.EventStreamTaggedUnionPojoSupplier;
 import software.amazon.awssdk.awscore.eventstream.RestEventStreamAsyncResponseTransformer;
 import software.amazon.awssdk.awscore.exception.AwsServiceException;
+import software.amazon.awssdk.codegen.model.config.customization.MetadataConfig;
 import software.amazon.awssdk.codegen.model.intermediate.IntermediateModel;
 import software.amazon.awssdk.codegen.model.intermediate.Metadata;
 import software.amazon.awssdk.codegen.model.intermediate.OperationModel;
@@ -94,8 +95,13 @@ public class JsonProtocolSpec implements ProtocolSpec {
                       .addCode(".protocolVersion($S)\n", metadata.getJsonVersion())
                       .addCode("$L", customErrorCodeFieldName());
 
-        if (metadata.getContentType() != null) {
-            methodSpec.addCode(".withContentTypeOverride($S)", metadata.getContentType());
+
+        String contentType = Optional.ofNullable(model.getCustomizationConfig().getCustomServiceMetadata())
+                .map(MetadataConfig::getContentType)
+                .orElse(metadata.getContentType());
+
+        if (contentType != null) {
+            methodSpec.addCode(".contentType($S)", contentType);
         }
 
         registerModeledExceptions(model, poetExtensions).forEach(methodSpec::addCode);
--- a/core/protocols/aws-json-protocol/src/main/java/software/amazon/awssdk/protocols/json/BaseAwsJsonProtocolFactory.java
+++ b/core/protocols/aws-json-protocol/src/main/java/software/amazon/awssdk/protocols/json/BaseAwsJsonProtocolFactory.java
@@ -147,7 +147,8 @@ public abstract class BaseAwsJsonProtocolFactory {
 
     @SdkTestInternalApi
     protected final String getContentType() {
-        return getContentTypeResolver().resolveContentType(protocolMetadata);
+        return protocolMetadata.contentType() != null ? protocolMetadata.contentType()
+                : getContentTypeResolver().resolveContentType(protocolMetadata);
     }
 
     /**
@@ -239,6 +240,18 @@ public abstract class BaseAwsJsonProtocolFactory {
          */
         public final SubclassT protocolVersion(String protocolVersion) {
             protocolMetadata.protocolVersion(protocolVersion);
+            return getSubclass();
+        }
+
+        /**
+         * ContentType  of the client (By default it is used from {@link #AWS_JSON} ).
+         * Used to determine content type.
+         *
+         * @param contentType JSON protocol contentType.
+         * @return This builder for method chaining.
+         */
+        public final SubclassT contentType(String contentType) {
+            protocolMetadata.contentType(contentType);
             return getSubclass();
         }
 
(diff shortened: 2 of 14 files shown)


core/protocols/aws-json-protocol/src/main/java/software/amazon/awssdk/protocols/json/BaseAwsJsonProtocolFactory.java
```diff
@@ -242,6 +242,18 @@ public final SubclassT protocolVersion(String protocolVersion) {
             return getSubclass();
         }
 
+        /**
+         * ContentType  of the client (By default it is application/x-amz-json-<JSON version>).
```
millems, line 246: This default will change soon in the future (like, the next few months - there's a SEP on the way), and we'll probably forget to update this comment. Should we link to where the default is defined instead?

core/protocols/aws-json-protocol/src/main/java/software/amazon/awssdk/protocols/json/DefaultJsonContentTypeResolver.java
```diff
@@ -33,6 +33,7 @@ public DefaultJsonContentTypeResolver(String prefix) {
     @Override
     public String resolveContentType(AwsJsonProtocolMetadata protocolMetadata) {
         //Changing this to 'application/json' may break clients expecting 'application/x-amz-json-1.1'
-        return prefix + protocolMetadata.protocolVersion();
+        return protocolMetadata.contentTypeOverride() != null ? protocolMetadata.contentTypeOverride()
+                : prefix + protocolMetadata.protocolVersion();
```
millems, line 37: I don't like "overrides" being bound so late. It means we have to pipe a lot information through from the code generator, which adds a lot more cross-module API surface area. Can we just have this line be `return protocolMetadata.contentType();` and have the code generator define the content-type that's used?

We can have the default value be "prefix + protocolMetadata.protocolVersion()", but do that upstream in the protocol factory. I'd honestly prefer it to be defaulted in the code generator, but that would break people who just upgrade the JSON protocol module without updating their service clients (which we try not to do outside of minor version bumps).

millems (commented):
Can we add a comment to the customization to make it clear that it only works for JSON protocols?

codegen/src/main/java/software/amazon/awssdk/codegen/poet/client/specs/JsonProtocolSpec.java
```diff
@@ -94,16 +95,24 @@ public MethodSpec initProtocolFactory(IntermediateModel model) {
...
     }
 
+    private void addContentTypeOverride(IntermediateModel model, Metadata metadata, MethodSpec.Builder methodSpec) {
+        MetadataConfig customServiceMetadata = model.getCustomizationConfig().getCustomServiceMetadata();
+        // Preference is given to CustomizationConfig
+        if (customServiceMetadata != null && customServiceMetadata.getContentType() != null) {
+            methodSpec.addCode(".contentTypeOverride($S)", customServiceMetadata.getContentType());
+        } else if (metadata.getContentType() != null) {
+            methodSpec.addCode(".contentTypeOverride($S)", metadata.getContentType());
+        }
```
millems, line 113: Not a blocker, but this is easier for me to read:

```java
String contentType = Optional.ofNullable(model.getCustomizationConfig().getCustomServiceMetadata())
                             .map(MetadataConfig::getContentType)
                             .orElse(metadata.getContentType()));
methodSpec.addCode(".contentTypeOverride($S)", contentType);
```

core/protocols/aws-json-protocol/src/main/java/software/amazon/awssdk/protocols/json/DefaultJsonContentTypeResolver.java
joviegas: Agree.  Tried adding this changes in Codegen it self but there we have dependency of Protocol Version too and it make code complicated as as you said might break upgrade the JSON protocol module .
Moved this logic to BaseAwsJsonProtocolFactory.

core/protocols/aws-json-protocol/src/main/java/software/amazon/awssdk/protocols/json/BaseAwsJsonProtocolFactory.java
joviegas: Added link to  code where we set the default value ```AWS_JSON = new DefaultJsonContentTypeResolver("application/x-amz-json-");```

core/protocols/aws-json-protocol/src/main/java/software/amazon/awssdk/protocols/json/BaseAwsJsonProtocolFactory.java
```diff
@@ -242,6 +243,18 @@ public final SubclassT protocolVersion(String protocolVersion) {
...
+         * ContentType  of the client (By default it is used from {@link #AWS_JSON} ).
+         * Used to determine content type.
+         *
+         * @param contentTypeOverride JSON protocol contentTypeOverride.
+         * @return This builder for method chaining.
+         */
+        public final SubclassT contentTypeOverride(String contentTypeOverride) {
+            protocolMetadata.contentTypeOverride(contentTypeOverride);
+            return getSubclass();
+        }
```
millems, line 256: I think in the future we will move contentType to be always controlled from the codegen, because we'll need to treat each protocol differently. Should we just name this "contentType" since in the future it won't really be an "override"?

joviegas: Changed all occurrences of contentTypeOverride to contentType.

core/protocols/aws-json-protocol/src/main/java/software/amazon/awssdk/protocols/json/DefaultJsonContentTypeResolver.java
joviegas: Moved the logic where the  check for customized ContentType is present  is moved to protocol factory.

millems (approved):

PR #2522, opened 2021-06-09
Moved Jackson from being an external SDK dependency to an internal dependency.

millems:
Jackson-databind was removed from being a dependency (except in the code generator). Jackson-core and
Jackson-dataformat-cbor were moved to shaded dependencies. This was done because Jackson versions
are not 100% compatible between minor versions and keeping them up to date were problematic for SDK
customers.

Customers will see a change in artifact sizes. Customers which do not use Jackson-databind outside of
the SDK today will see a ~1 MB artifact size decrease. Customers which do use Jackson-databind outside
of the SDK will see a ~0.5 MB artifact size increase.

This change required breaking inter-module (protected) APIs, so it should not be released outside of an
SDK minor version bump. Customers will not be able to use older client versions with this version of
the core libraries.

Change summary:
1. Added 'jackson-core' and 'jackson-dataformat-cbor' which contain the SDK's shaded versions with the
   Jackson packages of the same names.
2. Removed 'aws-ion-protocol' module. This was not used by any public AWS services and can be re-added when
   service support is needed.
3. Added 'json-utils', a library for reading and parsing JSON.


codegen/pom.xml
```diff
@@ -143,6 +138,14 @@
...
             <artifactId>jackson-jr-stree</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-annotations</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+        </dependency>
```
millems, line 148: The code generator was already using this. Because it was removed from the aws-json-protocol package, we have to re-add it here. It's fine if the code generator uses Jackson, because it's just a build-time dependency.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/HttpCredentialsProvider.java
```diff
@@ -98,11 +103,6 @@ protected boolean isLocalCredentialLoadingDisabled() {
                                 .build();
         } catch (SdkClientException e) {
             throw e;
-        } catch (JsonMappingException e) {
-            throw SdkClientException.builder()
-                                    .message("Unable to parse response returned from service endpoint.")
-                                    .cause(e)
-                                    .build();
```
millems, line 105: The same failure scenario (invalid response JSON) is now thrown as an UncheckedIOException (caught below).

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/ProcessCredentialsProvider.java
```diff
@@ -174,21 +174,10 @@ private Instant credentialExpirationTime(JsonNode credentialsJson) {
...
-            return null;
-        }
-
-        if (!subNode.isTextual()) {
-            throw new IllegalStateException(nodeName + " from credential process should be textual, but was " +
-                                            subNode.getNodeType());
-        }
-
-        return subNode.asText();
+        return jsonObject.get(nodeName).map(JsonNode::text).orElse(null);
```
millems, line 180: We validate the result elsewhere, so the invalid response type will be handled elsewhere.

core/json-utils/pom.xml
```diff
@@ -18,32 +18,32 @@
...
         <groupId>software.amazon.awssdk</groupId>
         <version>2.16.82-SNAPSHOT</version>
     </parent>
     <modelVersion>4.0.0</modelVersion>
 
-    <artifactId>aws-ion-protocol</artifactId>
-    <name>AWS Java SDK :: Core :: Protocols :: AWS Ion Protocol</name>
-    <description>The AWS SDK for Java - module holds the classes for AWS Ion protocol
-    </description>
+    <artifactId>json-utils</artifactId>
```
millems, line 27: Not actually a file rename, just git being over-zealous.

core/json-utils/src/main/java/software/amazon/awssdk/protocols/jsoncore/JsonNodeParser.java
```diff
@@ -0,0 +1,231 @@
...
+import software.amazon.awssdk.thirdparty.jackson.core.JsonToken;
+import software.amazon.awssdk.thirdparty.jackson.core.json.JsonReadFeature;
+
+/**
+ * Parses an JSON document into a simple DOM-like structure, {@link JsonNode}.
+ *
+ * <p>This is created using {@link #create()} or {@link #builder()}.
+ */
+@SdkProtectedApi
+public final class JsonNodeParser {
```
millems, line 46: Logic mostly copied from JsonDomParser.

third-party/jackson-core/pom.xml
```diff
@@ -0,0 +1,124 @@
...
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>third-party</artifactId>
+        <groupId>software.amazon.awssdk</groupId>
+        <version>2.16.82-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>jackson-core</artifactId>
```
millems, line 27: This matches Jackson's artifact ID. Should we pick something else?

zoewangg: `sdk-jackson-core`?

test/module-path-tests/src/main/java/module-info.java
```diff
@@ -30,5 +30,4 @@
 
     requires org.slf4j;
     requires slf4j.simple;
-    requires com.fasterxml.jackson.core;
```
zoewangg, line 33: Hmm, does it mean this change would break customers who use the java SDK via module path and do not use Jackson for anything else?

millems: If they're defining "requires com.fasterxml.jackson.core" (or literally making any other in-code reference to Jackson) without including it themselves (or another dependency that does so), and we remove Jackson, then yes they will break. 

In general people should not rely on transitive dependencies without directly depending on it themselves. We have the maven dependency plugin to make sure we don't do that. I'm sure some people are, though, but there's nothing we can do about that. "We'll only break people that are already broken" is the name of the game here, I think.

zoewangg: I think currently they do have to add `requires com.fasterxml.jackson.core;` otherwise they'll get `java.lang.NoClassDefFoundError: com/fasterxml/jackson/core/JsonFactory`; That's why we have it here in the first place. (If we had module-info in the SDK specifying `requires transitive com.fasterxml.jackson.core;`, then customers wouldn't need to.)

I don't think this is a blocker, but we should probably make it clear in the changelog entry.

core/json-utils/src/main/java/software/amazon/awssdk/protocols/jsoncore/internal/BooleanJsonNode.java
```diff
@@ -0,0 +1,78 @@
...
+    @Override
+    public Object asEmbeddedObject() {
+        throw new UnsupportedOperationException("A JSON boolean cannot be converted to an embedded object.");
+    }
+
+    @Override
+    public String text() {
+        return Boolean.toString(value);
+    }
+
```
joviegas, line 73: Query : I saw that SDKJsonNode implementations had equals and hashCode overrides , should we have the same for JsonNode Concrete classes ?

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/ProcessCredentialsProvider.java
millems: This change introduces a bit of a relaxing of the error messages. Before, they could not specify a value of 1 or null. Now they can, but it'll fail when it hits the service.

core/json-utils/src/main/java/software/amazon/awssdk/protocols/jsoncore/JsonNode.java
```diff
@@ -0,0 +1,185 @@
...
+     * When {@link #isString()}, {@link #isBoolean()}, or {@link #isNumber()} is true, this will return the value of this node
+     * as a textual string. If this is any other type, this will return null.
+     */
+    String text();
+
+    /**
+     * When {@link #isObject()} is true, this will return the result of {@code Optional.ofNullable(asObject().get(child))}. If
+     * this is any other type, this will return {@link Optional#empty()}.
+     */
+    default Optional<JsonNode> get(String child) {
```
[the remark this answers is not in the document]
millems: Sure

core/json-utils/src/main/java/software/amazon/awssdk/protocols/jsoncore/internal/EmbeddedObjectJsonNode.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    @Override
+    public Object asEmbeddedObject() {
+        return embeddedObject;
+    }
+
+    @Override
+    public String text() {
+        return null;
```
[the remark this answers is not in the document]
millems: Throwing an exception defeats `text()`'s purpose - it's a way for the caller to not care about whether it's a string, number or boolean and treat them all the same (same as Jackson's `asText()`). If embedded objects returned `embeddedObject.toString()` from `text()`, why not do the same for array and object types? I don't think that's the purpose of `text()`.

core/json-utils/src/main/java/software/amazon/awssdk/protocols/jsoncore/JsonNodeParser.java
```diff
@@ -0,0 +1,231 @@
...
+            case VALUE_EMBEDDED_OBJECT:
+                return new EmbeddedObjectJsonNode(parser.getEmbeddedObject());
+            default:
+                throw new IllegalArgumentException("Unexpected JSON token - " + token);
+        }
+    }
+
+    private JsonNode parseObject(JsonParser parser) throws IOException {
+        JsonToken currentToken = parser.nextToken();
+        Map<String, JsonNode> object = new LinkedHashMap<>();
```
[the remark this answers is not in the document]
millems: Yeah, it's worth benchmarking these changes.

core/json-utils/src/main/java/software/amazon/awssdk/protocols/jsoncore/internal/EmbeddedObjectJsonNode.java
```diff
@@ -0,0 +1,80 @@
...
+
+/**
+ * An embedded object {@link JsonNode}.
+ */
+@SdkInternalApi
+public final class EmbeddedObjectJsonNode implements JsonNode {
+    private final Object embeddedObject;
+
+    public EmbeddedObjectJsonNode(Object embeddedObject) {
+        Validate.paramNotNull(embeddedObject, "embeddedObject");
```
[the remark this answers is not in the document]
millems: Because these constructors are only called internal to the module, I was fine trusting that Jackson would never return invalid types. I agree that this is inconsistent, though. I'd be fine going either way, because this check is pretty much free from a performance perspective. What do you think would be better?

core/json-utils/src/main/java/software/amazon/awssdk/protocols/jsoncore/internal/BooleanJsonNode.java
millems: +1

core/json-utils/src/main/java/software/amazon/awssdk/protocols/jsoncore/internal/EmbeddedObjectJsonNode.java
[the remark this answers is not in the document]
joviegas: Then I think it's fine if we skip the null check since it's internal and Jackson doesn't return invalid types

core/json-utils/src/main/java/software/amazon/awssdk/protocols/jsoncore/JsonNodeParser.java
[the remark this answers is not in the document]
joviegas: We can benchmark this as another task. This is not a blocker to this PR .

joviegas (approved):

PR #2536, opened 2021-06-16
Fix a race condition in FileAsyncRequestBody

zoewangg:
## Description
Fix a race condition in FileAsyncRequestBody where a request demand comes in at the same time reading is progress, which causes the request to hang.

## Types of changes

## Checklist

## License

### DIFF
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/FileAsyncRequestBody.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/FileAsyncRequestBody.java
@@ -31,6 +31,7 @@ import software.amazon.awssdk.annotations.SdkInternalApi;
 import software.amazon.awssdk.core.async.AsyncRequestBody;
 import software.amazon.awssdk.core.internal.util.Mimetype;
 import software.amazon.awssdk.core.internal.util.NoopSubscription;
+import software.amazon.awssdk.utils.Logger;
 import software.amazon.awssdk.utils.builder.SdkBuilder;
 
 /**
@@ -41,6 +42,7 @@ import software.amazon.awssdk.utils.builder.SdkBuilder;
  */
 @SdkInternalApi
 public final class FileAsyncRequestBody implements AsyncRequestBody {
+    private static final Logger log = Logger.loggerFor(FileAsyncRequestBody.class);
 
     /**
      * Default size (in bytes) of ByteBuffer chunks read from the file and delivered to the subscriber.
@@ -169,10 +171,11 @@ public final class FileAsyncRequestBody implements AsyncRequestBody {
         private final Subscriber<? super ByteBuffer> subscriber;
         private final int chunkSize;
 
-        private long position = 0;
-        private AtomicLong outstandingDemand = new AtomicLong(0);
-        private boolean writeInProgress = false;
+        private final AtomicLong position = new AtomicLong(0);
+        private long outstandingDemand = 0;
+        private boolean readInProgress = false;
         private volatile boolean done = false;
+        private final Object lock = new Object();
 
         private FileSubscription(AsynchronousFileChannel inputChannel, Subscriber<? super ByteBuffer> subscriber, int chunkSize) {
             this.inputChannel = inputChannel;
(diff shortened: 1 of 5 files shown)


core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/FileAsyncRequestBody.java
```diff
@@ -271,34 +274,22 @@ private void closeFile() {
...
-        private void signalOnNext(ByteBuffer bb) {
-            synchronized (this) {
-                if (!done) {
-                    subscriber.onNext(bb);
-                }
-            }
-        }
-
         private void signalOnComplete() {
-            synchronized (this) {
```
zoewangg, line 283: `synchronized` is removed because it doesn't seem to be necessary here imo.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/FileAsyncRequestBody.java
```diff
@@ -271,34 +274,22 @@ private void closeFile() {
...
-                    done = true;
-                }
+            if (!done) {
+                done = true;
+                subscriber.onComplete();
             }
         }
 
         private void signalOnError(Throwable t) {
-            synchronized (this) {
```
zoewangg, line 292: `synchronized` is removed because it doesn't seem to be necessary here imo.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/FileAsyncRequestBody.java
```diff
@@ -271,34 +274,22 @@ private void closeFile() {
             }
         }
 
-        private void signalOnNext(ByteBuffer bb) {
-            synchronized (this) {
```
zoewangg, line 275: `synchronized` is removed because it doesn't seem to be necessary here imo.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/FileAsyncRequestBody.java
```diff
@@ -169,10 +169,11 @@ public FileAsyncRequestBody build() {
         private final Subscriber<? super ByteBuffer> subscriber;
         private final int chunkSize;
 
-        private long position = 0;
-        private AtomicLong outstandingDemand = new AtomicLong(0);
-        private boolean writeInProgress = false;
```
dagnir, line 174: Why did we have this? lol

zoewangg: copy paste error from FileAsyncResponseTransformer I guess, lol

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/FileAsyncRequestBody.java
```diff
@@ -169,10 +169,11 @@ public FileAsyncRequestBody build() {
...
+        private final AtomicLong position = new AtomicLong(0);
```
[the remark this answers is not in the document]
zoewangg: I tried with volatile long, but IntelliJ was complaining `non-atomic operation on volatile variable` for `position += attachment.remaining());`, so I figured we should use AtomicLong to be safe

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/FileAsyncRequestBody.java
zoewangg: Can `signalOnComplete` and `signalOnError` be invoked at the same time though? It's either complete or error right?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/FileAsyncRequestBody.java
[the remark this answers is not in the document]
dagnir: Ah I missed that; this looks good then 👍

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/FileAsyncRequestBody.java
dagnir: hmm, one (albeit rare possibility) if the file is complete, it's possible that the reading thread calls `onComplete()`, and at the same time, `request(-1)` is called so we try to signal `onError()`.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/FileAsyncRequestBody.java
```diff
@@ -271,34 +274,22 @@ private void closeFile() {
             }
```
millems, line 274: `closeFile` calls `signalOnError` if it fails, but we may have already called `onComplete` or `onError` which is a spec violation. Should we just log the error as a warning instead of signaling it?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/FileAsyncRequestBody.java
millems: I believe this was needed, in case a failure happened in `request(long)` and the `inputChannel.read` simultaneously. It guards against double-sending `onError`.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/FileAsyncRequestBody.java
millems: +1 I believe this was needed to ensure mutual execution with signalOnError

zoewangg: Yeah, good point 👍🏽

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/FileAsyncRequestBody.java
zoewangg: Ah, good catch! Yeah, we should just log the error

dagnir (approved):

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/FileAsyncRequestBody.java
millems: Technically I think this is still needed. onError and onNext must be called serially, so the synchronized made sure that it would always happen.

Without the synchronized, we may get onError and onNext in parallel, e.g. if someone calls subscription.request(-1) while onNext is in progress.

zoewangg: Yeah, that's fair.

millems (approved):

PR #2558, opened 2021-06-25
Wrap CrtS3RuntimeException to SdkServiceException

joviegas:
## Description

- Wrap CrtS3RuntimeException to SdkServiceException

## Motivation and Context

- Currently CrtClient throws CrtS3RuntimeException whenever there is S3 Service side exceptions.
- We need to Wrap these exception indicating these are Service side exceptions

## Testing

- Added Integration test
- Added Junits

## Types of changes

## Checklist

## License

### DIFF
--- a/services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/CrtErrorHandler.java
+++ b/services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/CrtErrorHandler.java
@@ -1,0 +1,117 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.transfer.s3.internal;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.awscore.exception.AwsErrorDetails;
+import software.amazon.awssdk.core.exception.SdkClientException;
+import software.amazon.awssdk.core.exception.SdkServiceException;
+import software.amazon.awssdk.crt.s3.CrtS3RuntimeException;
+import software.amazon.awssdk.services.s3.model.BucketAlreadyExistsException;
+import software.amazon.awssdk.services.s3.model.BucketAlreadyOwnedByYouException;
+import software.amazon.awssdk.services.s3.model.InvalidObjectStateException;
+import software.amazon.awssdk.services.s3.model.NoSuchBucketException;
+import software.amazon.awssdk.services.s3.model.NoSuchKeyException;
+import software.amazon.awssdk.services.s3.model.NoSuchUploadException;
+import software.amazon.awssdk.services.s3.model.ObjectAlreadyInActiveTierErrorException;
+import software.amazon.awssdk.services.s3.model.S3Exception;
+import software.amazon.awssdk.utils.StringUtils;
+
+@SdkInternalApi
+public class CrtErrorHandler {
+
+    private final Map<String, S3Exception.Builder> s3ExceptionBuilderMap;
+
+    public CrtErrorHandler() {
+        s3ExceptionBuilderMap = getS3ExceptionBuilderMap();
+    }
+
+    /**
+     * This class transform a crtTRunTimeException to the S3 service Exceptions.
+     * CrtS3RuntimeException are the exceptions generated due to failures in CRTClient due to S3 Service errors.
+     *
+     * @param crtRuntimeException Exception that is thrown by CrtClient.
+     * @return
+     */
+    public Exception transformException(Exception crtRuntimeException) {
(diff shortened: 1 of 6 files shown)


services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/CrtErrorHandler.java
```diff
@@ -0,0 +1,61 @@
...
+     * @return
+     */
+    public Exception transformException(Exception crtRuntimeException) {
+        Optional<CrtS3RuntimeException> crtS3RuntimeExceptionOptional = getCrtS3RuntimeException(crtRuntimeException);
+
+        return crtS3RuntimeExceptionOptional.isPresent()
+                ? crtS3RuntimeExceptionOptional
+                .filter(item -> StringUtils.isNotBlank(item.getAwsErrorCode())
+                        || StringUtils.isNotBlank(item.getAwsErrorMessage())).map(e -> SdkServiceException.builder()
+                        .statusCode(crtS3RuntimeExceptionOptional.get().getStatusCode())
```
zoewangg, line 43: Can we create specific S3Exception based on the error code?
For example:
`NoSuchBucket` -> `NoSuchBucketException`. 

That way, customers can handle specific cases without checking the error code by themselves. 

If the error code can't be mapped, I think we should just return a generic `S3ServiceException`

If there's no error code, we should wrap it with `SdkClientException`. 

WDUT?

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/CrtErrorHandler.java
```diff
@@ -0,0 +1,61 @@
...
+     * This class transform a CRTRunTimeException to Meaningful SdkServiceException.
+     * CrtS3RuntimeException are the exceptions generated due to failures in CRTClient due to S3 Service errors.
+     *
+     * @param crtRuntimeException Exception that is thrown by CRTClient.
+     * @return
+     */
+    public Exception transformException(Exception crtRuntimeException) {
+        Optional<CrtS3RuntimeException> crtS3RuntimeExceptionOptional = getCrtS3RuntimeException(crtRuntimeException);
+
+        return crtS3RuntimeExceptionOptional.isPresent()
```
zoewangg, line 39: Minor: instead of checking if it's present and doing `crtS3RuntimeExceptionOptional.get()`, we can do the following:
```
return crtS3RuntimeExceptionOptional.filter(xxxx).orElse(crtRuntimeException)
```

Can we break up the code block within `filter(xx)`? It's a bit hard to read

services-custom/s3-transfer-manager/src/test/java/software/amazon/awssdk/transfer/s3/internal/CrtExceptionTransformationIntegrationTest.java
```diff
@@ -0,0 +1,140 @@
...
+                        .build();
+    }
+
+    @Test
+    public void getObjectNoSuchKey() throws IOException {
+        String randomBaseDirectory = Files.createTempDirectory(getClass().getSimpleName()).toString();
+        assertThatThrownBy(() -> s3Crt.getObject(GetObjectRequest.builder().bucket(BUCKET).key("randomKey").build(),
+                Paths.get(randomBaseDirectory).resolve("testFile")).get())
+                .hasCauseInstanceOf(SdkServiceException.class)
+                .hasMessageContaining("Failed with Aws Error code \"NoSuchKey\" and Aws Error Message \"The specified key does not exist.\"");
```
zoewangg, line 80: Minor: we probably don't need to verify the whole message because services can change them, which will make the tests flaky.

Same as other tests

services-custom/s3-transfer-manager/src/test/java/software/amazon/awssdk/transfer/s3/internal/CrtErrorHandlerTest.java
```diff
@@ -0,0 +1,74 @@
...
+import software.amazon.awssdk.crt.s3.CrtS3RuntimeException;
+
+@RunWith(MockitoJUnitRunner.class)
+public class CrtErrorHandlerTest {
+
+    @Mock
+    private CrtS3RuntimeException mockCrtS3RuntimeException;
+
+    @Test
+    public void testCrtS3ExceptionAreTransformed(){
```
zoewangg, line 36: nit, we don't normally use `test` prefix in the test names, seems a bit redundant, imo.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/CrtErrorHandler.java
joviegas: Implemented to get Service side exception.
We might need to update the CrtS3RuntimeException to send requestId so that service exception  is built with this.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/CrtErrorHandler.java
joviegas: Done

services-custom/s3-transfer-manager/src/test/java/software/amazon/awssdk/transfer/s3/internal/CrtExceptionTransformationIntegrationTest.java
joviegas: removed this

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/CrtErrorHandler.java
zoewangg: Can we write down each exception type rather than using reflection? I don't think there are many specific exceptions for get and put. https://github.com/aws/aws-sdk-java-v2/blob/master/services/s3/src/main/resources/codegen-resources/service-2.json#L509 In the future, we will generate the code. I think we should be very careful about using reflection.

PR #2560, opened 2021-06-26
Update s3 crt client tests to verify resources are closed properly

zoewangg:
## Description
Update s3 crt client tests to verify resources are closed properly

`CrtResource.waitForNoResources();` will time out if there are remaining resources after 1 min

## Types of changes

## Checklist

## License

### DIFF
--- a/services-custom/s3-transfer-manager/src/it/java/software/amazon/awssdk/transfer/s3/S3IntegrationTestBase.java
+++ b/services-custom/s3-transfer-manager/src/it/java/software/amazon/awssdk/transfer/s3/S3IntegrationTestBase.java
@@ -15,7 +15,9 @@
 
 package software.amazon.awssdk.transfer.s3;
 
+import org.junit.AfterClass;
 import org.junit.BeforeClass;
+import software.amazon.awssdk.crt.CrtResource;
 import software.amazon.awssdk.regions.Region;
 import software.amazon.awssdk.services.s3.S3AsyncClient;
 import software.amazon.awssdk.services.s3.S3AsyncClientBuilder;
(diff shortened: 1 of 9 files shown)


services-custom/s3-transfer-manager/src/it/java/software/amazon/awssdk/transfer/s3/S3IntegrationTestBase.java
```diff
@@ -55,10 +57,16 @@
...
     @BeforeClass
     public static void setUp() throws Exception {
+        System.setProperty("aws.crt.debugnative", "true");
         s3 = s3ClientBuilder().build();
         s3Async = s3AsyncClientBuilder().build();
     }
 
+    @AfterClass
+    public static void cleanUp() {
+        CrtResource.waitForNoResources();
```
Bennett-Lynch, line 67: Nit/off-topic: Long-term, I think we should consider moving away from static resources like this, since it prohibits us from being able to run these tests in parallel one day (which is supported in JUnit 5).

Bennett-Lynch (approved):

services-custom/s3-transfer-manager/src/it/java/software/amazon/awssdk/transfer/s3/S3IntegrationTestBase.java
zoewangg: Yeah, we might be able to run some tests (eg: tests for idempotent operations) in parallel but for some service tests that require certain resources to be created first such as s3 bucket, it might be better to set up static resources to be used by all tests within the same class. If we create one bucket per test, we might run out of buckets, and it'll take more time for each test to finish because of the resource creation and deletion.

PR #2583, opened 2021-07-09
Add proposal for swappable metadata providers

rli:
## Description
A proposal for allowing service, partition and region metadata to be loaded and used dynamically at runtime instead of only the static generated metadata.


debora-ito:
Talked to @rli, this should be supported by the new rule-based endpoint providers.

PR #2588, opened 2021-07-13
Add support for marshalling lists of strings in HTTP headers...

Bennett-Lynch:
...for JSON/XML protocols.

## Motivation and Context

We currently lack support for marshalling lists of strings (and enums)
in HTTP headers. Other IDLs and modeling languages, e.g., Smithy, do
support such bindings and services may come to expect such support:

https://awslabs.github.io/smithy/1.0/spec/core/http-traits.html#httpheader-trait

## Description

* Add support for marshalling a list of strings to relevant JSON and XML
marshalling classes
* Consistent with the existing XML HeaderUnmarshaller map
implementation, unmarshalling only supports string value types (this can
be extended in the future if needed)
* Marshalling implementations are changed to use appendHeader rather
than putHeader
* Add convenience method SdkField#getRequiredTrait(..) to support the
common use case of ensuring a trait is present; update other existing
use cases to utilize this method
* Update relevant test classes to correctly recognize headers as a
`Map<String, List<String>>`
* Only wrap marshalling assertion exceptions as runtime when needed,
minimizing stack trace noise

## Testing

New tests added to:
* rest-json-input.json
* rest-json-output.json
* rest-xml-input.json
* rest-xml-output.json

## Types of changes

## Checklist

## License


core/protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/internal/marshall/HeaderMarshaller.java
```diff
@@ -66,6 +68,24 @@ protected boolean shouldEmit(Map map) {
...
+            SdkField memberFieldInfo = sdkField.getRequiredTrait(ListTrait.class).memberFieldInfo();
+            for (Object listValue : list) {
+                XmlMarshaller marshaller = context.marshallerRegistry().getMarshaller(MarshallLocation.HEADER, listValue);
+                marshaller.marshall(listValue, context, paramName, memberFieldInfo);
+            }
+        }
+
+        @Override
+        protected boolean shouldEmit(List list) {
+            return list != null && !list.isEmpty();
```
zoewangg, line 86: nit: should we just use CollectionUtils.isNullOrEmpty?

core/protocols/aws-json-protocol/src/main/java/software/amazon/awssdk/protocols/json/internal/marshall/HeaderMarshaller.java
```diff
@@ -45,6 +48,17 @@
     public static final JsonMarshaller<Instant> INSTANT
         = new SimpleHeaderMarshaller<>(JsonProtocolMarshaller.INSTANT_VALUE_TO_STRING);
 
+    public static final JsonMarshaller<List<?>> LIST = (list, context, paramName, sdkField) -> {
+        if (list.isEmpty()) {
```
zoewangg, line 52: Do we need to check null here?

zoewangg (commented):
Do we need to support it in query protocol?

Bennett-Lynch:
> Do we need to support it in query protocol?

Discussed offline and we believe this is not applicable to query protocol.

zoewangg (approved):

joviegas (approved):

PR #2598, opened 2021-07-15
Release preparation for removing Jackson as an external dependency

millems:
Code changes were already reviewed as part of https://github.com/aws/aws-sdk-java-v2/pull/2522


joviegas (approved):

johnduffell:
Hello, this sounds like a good idea on the face of it, but it makes life very hard for vulnerability checking (which jackson databind is renowned for).  We use snyk.io to look for vulnerabilities which won't spot them, and even if it did, there would be no way to override/evict the dependencies in favour of a patch version.  Unless aws-sdk-v2 can commit to releasing patched versions within 14 days of a vulnerabaility, this may leave users in breach of GDPR obligations to keep systems secure.
We raised it with snyk directly and they mentioned that Elasticsearch have actually gone the other way and stopped using shaded dependencies for the reasons I mentioned: https://www.elastic.co/blog/to-shade-or-not-to-shade

Do you have anything to add on this?  Cheers,

millems:
We're aware that shading has those (and other) drawbacks, which is why we held off doing it so long. 

We did this precisely *because* we previously could not upgrade to new Jackson versions without breaking backwards-compatibility. Jackson is not backwards-compatible between minor versions, so our customers did not want us to upgrade, even though we needed to upgrade to get new security updates. Shading allows us to use the latest, secure Jackson version without worrying about breaking customer applications.

We also removed Jackson-databind as a dependency - we don't use it anymore (even shaded) because of the issues you're referencing. Jackson-core and Jackson-dataformat-cbor do not have nearly as many open CVEs. We also run the types of validations you reference in order to detect vulnerabilities in Jackson, so that we will be alerted and can assess whether the vulnerability impacts our specific usage of Jackson-core and Jackson-dataformat-cbor. This will allow us to quickly patch issues if we're impacted.

johnduffell:
Thanks for your detailed response and it's great that you have a plan to release patches quickly.
Just a small point of note that the people behind jackson do seem to patch several minor versions in parallel, however I am fully aware of the classpath conflicts issue and the difficulty of shading libraries at the app level, so I can understand you wanting to solve the issue for customers.

PR #2611, opened 2021-07-20
Remove buffer copy since it's no longer needed

zoewangg:
## Motivation and Context

- Bump up crt version
- Remove buffer copy because it's no longer needed. https://github.com/awslabs/aws-crt-java/pull/352

## License


dagnir (approved):

zoewangg:
Integ tests and stability tests passed. The code smells reported are not related to this change.

Issue #1656, opened 2020-02-18
CopyObject in SDK V2- how to specify the copy version


damogallagher:
## Expected Behavior
In version 1 of the sdk - you could specify the following code to perform a copy on a particular version of a file
```
 CopyObjectRequest copyObjectRequest =new CopyObjectRequest(fromBucketName, fromObjectKey, toBucketName, toObjectKey).withSourceVersionId(versionId);

s3Client.copyObject(copyObjectRequest);
```

## Current Behavior
I cant seem to figure out with version 2 of the SDK how to specify the version when performing a copy

## Possible Solution
Is it possible to specify a version on the copyObject command? If so could the documentation be updated with the correct syntax to use

## Your Environment

* AWS Java SDK version used: 2.10.65
* JDK version used: Amazon 11 Corretto
* Operating System and version: Mac OS Catalina 10.15.3

debora-ito:
Hi @damogallagher, to copy a specified version of an object you can pass the version id in the CopyObjectRequest [copySource](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/s3/model/CopyObjectRequest.html#copySource--):
```
/bucket/my-image.jpg?versionId=<myVersionId>
```

We've requested the documentation team to add this info to the docs some time ago, I'll ask them again.

damogallagher:
Does the versionId need to be encoded?
I have the following code
```
  try {
            String encodedUrl = null;
            try {
                if (StringUtils.isEmpty(versionId)) {
                    encodedUrl = URLEncoder.encode(fromBucketName + "/" + fromObjectKey, StandardCharsets.UTF_8.toString());
                } else {
                    String url = fromBucketName + "/" + fromObjectKey + "?versionId=" + versionId;
                     encodedUrl = URLEncoder.encode(url, StandardCharsets.UTF_8.toString());
                }
                logger.debug("encodedUrl:{}", encodedUrl);
            } catch (UnsupportedEncodingException e) {
                logger.error("An UnsupportedEncodingException has occured. Exception is: {}", e.getMessage());
                return fileCopied;
            }

            CopyObjectRequest copyObjectRequest = CopyObjectRequest.builder()
                    .copySource(encodedUrl)
                    .destinationBucket(toBucketName)
                    .destinationKey(toObjectKey)
                    .build();

            CopyObjectResponse copyObjectResponse = s3Client.copyObject(copyObjectRequest);

            if (copyObjectResponse == null) {
[...]

Issue #1452, opened 2019-10-03
Copy objects between S3 buckets - Specific version (Like SDK 1)


mani100101:
## Current Behavior
In the original SDK, there is a support to copy only the specific version of a object from source to destination.

`new CopyObjectRequest(sourceBucketName, sourceKey, sourceVersionId, destinationBucketName, destinationKey)`

In SDK2, support for copying an object of specific version is removed. 

## Expected Behavior
`CopyObjectRequest.Builder` should have methods like `sourceVersionId(sourceVersionId)` to copy a specific version of the object.

## Context
Need to copy or move a specific version of object between buckets instead of duplicating all versions of the object.

zoewangg:
Hi @mani100101, you can pass the version in `copySource`, something like the following: `bucket/key/?versionId=1` 

https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/s3/model/CopyObjectRequest.Builder.html#copySource-java.lang.String-

mani100101:
@zoewangg, thanks for helping out. It would be great if the same is captured in doc as well.

Anyways, will check and close the issue.

debora-ito:
@mani100101 Thank you for your feedback, I've forwarded it to our documentation team.

Please feel free to reopen this if you have any further questions.

Bennett-Lynch:
Hi @mani100101. Please take a look at the new copy source support introduced by https://github.com/aws/aws-sdk-java-v2/pull/2612 and let me know if this helps satisfy your use case.

PR #2612, opened 2021-07-21
[S3] Add support for more user-friendly CopyObject source parameters

Bennett-Lynch:
## Motivation and Context

The current S3Client interface has a cumbersome API for invoking
CopyObjectRequests. We require users to define the bucket name, key
name, and version ID in a raw string format. We require that the string
conform to the S3 API, which forces users to know the intricate details
for how to join these values together. Additionally, portions (but not
all) of the value must be URL encoded, further increasing the burden.

https://docs.aws.amazon.com/AmazonS3/latest/API/API_CopyObject.html#API_CopyObject_RequestParameters

In the Java SDK v1, users are given explicit parameters for the
different copy source attributes. But in v2, parity for this support is
clearly lacking. E.g.,

v1:
```
s3.copyObject(new CopyObjectRequest()
        .withSourceBucketName(SOURCE_BUCKET)
        .withSourceKey(key)
        .withSourceVersionId(versionId)
        .withDestinationBucketName(DESTINATION_BUCKET)
        .withDestinationKey(key));
```

v2:
```
s3.copyObject(CopyObjectRequest.builder()
        .copySource(SOURCE_BUCKET + "/" + key + "?versionId=" + versionId)
        .destinationBucket(DESTINATION_BUCKET)
        .destinationKey(key)
        .build());
```

The v1 SDK will also URL encode on the user's behalf, allowing users to
use the same input values as they would for a PutObjectRequest. The v2
code snippet above may appear to work for most users until they run into
unexpected source keys that require URL encoding, at which point they
will typically be given `NoSuchKey` errors.

This API deficiency has been called out by users in at least the
following issues:

* https://github.com/aws/aws-sdk-java-v2/issues/1313
* https://github.com/aws/aws-sdk-java-v2/issues/1452
* https://github.com/aws/aws-sdk-java-v2/issues/1656
* https://github.com/awsdocs/aws-doc-sdk-examples/issues/740

## Description
* For both CopyObjectRequest and UploadPartCopyRequest, add explicit
parameters for: SourceBucket, SourceKey, SourceVersionId
* If specified, these values will be used to construct a CopySource on
the user's behalf, including URL encoding the relevant portions.
* These values are introduced in a backwards compatible fashion. Users
who are already using CopySource today will see no change in behavior,
but these new fields may not be used in conjunction with CopySource.
* A follow-up PR will be submitted to propose deprecating the current
CopySource parameter. It is excluded from this PR since our current code
gen configuration lacks appropriate support.
* Add support for "DestinationBucket" & "DestinationKey" to
UploadPartCopyRequest (this support already existed for
CopyObjectRequest)
* Utility function added to detect if an ARN is for a particular S3
resource type. This is to conform with the S3 API requirements of
inserting "/object" in the path of these requests.

## Testing
* New unit tests added
* New integration tests added

## Types of changes

## Checklist

## License


services/s3/src/it/java/software/amazon/awssdk/services/s3/CopySourceIntegrationTest.java
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    @After
+    public void tearDown() {
+        deleteBucketAndAllContents(SOURCE_BUCKET_NAME);
+        deleteBucketAndAllContents(DESTINATION_BUCKET_NAME);
+    }
+
+    @Parameters
+    public static Collection parameters() throws Exception {
```
millems, line 69: Nit: `Collection<?>` or `Collection<String>`

services/s3/src/it/java/software/amazon/awssdk/services/s3/CopySourceIntegrationTest.java
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    private final String key;
+
+    public CopySourceIntegrationTest(String key) {
+        this.key = key;
+    }
+
+    @Test
+    public void copyObject_WithoutExplicitVersion_AcceptsSameKeyAsPut() throws Exception {
```
millems, line 86: Can any of these integration test cases be converted to build-time tests, now that we know how the service behaves?

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/handlers/CopySourceInterceptor.java
```diff
@@ -0,0 +1,127 @@
...
+                      .build();
+    }
+
+    private static String constructCopySource(String sourceBucket, String sourceKey, String sourceVersionId) {
+        StringBuilder copySource = new StringBuilder();
+        copySource.append("/");
+        copySource.append(urlEncodeIgnoreSlashes(sourceBucket));
+        if (isArnFor(S3ResourceType.ACCESS_POINT, sourceBucket) || isArnFor(S3ResourceType.OUTPOST, sourceBucket)) {
+            copySource.append("/object");
+        }
```
millems, line 103: Yikes, didn't know we had to go this specific.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/handlers/CopySourceInterceptor.java
```diff
@@ -0,0 +1,127 @@
...
+                                                             paramName));
+        }
+    }
+
+    private static <T> T requireSet(T value, String paramName) {
+        if (value == null) {
+            throw new IllegalArgumentException(String.format("Parameter '%s' must not be null",
+                                                             paramName));
+        }
+        return value;
```
millems, line 125: Nit: You could have used `Validate.isTrue` here. I don't think it's much easier to read, so no real need to change this unless you want to.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/resource/S3ArnUtils.java
```diff
@@ -72,4 +72,14 @@ public static IntermediateOutpostResource parseOutpostArn(Arn arn) {
                                           .outpostSubresource(ArnResource.fromString(subresource))
                                           .build();
     }
+
+    public static boolean isArnFor(S3ResourceType s3ResourceType, String arnString) {
```
millems, line 76: Nit: Should this be getArnType so that we don't have to parse the ARN twice?

services/s3/src/it/java/software/amazon/awssdk/services/s3/CopySourceIntegrationTest.java
Bennett-Lynch: I did try to capture most of these test cases in `CopySourceInterceptorTest` as well. They have mostly the same input for their `@Parameters`.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/handlers/CopySourceInterceptor.java
Bennett-Lynch: Unfortunate, I agree. This logic was borrowed from .NET's [CopyObjectRequestMarshaller](https://github.com/aws/aws-sdk-net/blob/969f2c5f26e593fbb1c622eb2a9b16806b125e54/sdk/src/Services/S3/Custom/Model/Internal/MarshallTransformations/CopyObjectRequestMarshaller.cs#L159-L161).

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/handlers/CopySourceInterceptor.java
Bennett-Lynch: Will change. I like not having to specify `String.format(..)`.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/resource/S3ArnUtils.java
Bennett-Lynch: Good idea. Especially since we're relying on exception handling here, which is not the most performant to begin with.

services/s3/src/it/java/software/amazon/awssdk/services/s3/CopySourceIntegrationTest.java
millems: That's good! Is there a reason to have the integration tests, then? Fewer tests mean faster releases, and less changes for internet hiccups to delay our release.

Bennett-Lynch: What I really like about the integration tests is that they validate that we can use identical input for both `PutObjectRequest` and `CopyObjectRequest`, as most users would expect (and where users have been burned before). But I understand your point. If you still feel they are redundant I can remove them though.

millems: How long do they take to run?

Bennett-Lynch: On my machine they take about 30 seconds to run. This is primarily because they're creating buckets per parameter. I can speed this up by reusing buckets between parameters.

Bennett-Lynch: Reusing buckets speeds this up to ~3-4 seconds. I was biasing for simplicity rather than speed before. Will update.

millems: Great! Much better. Fewer service calls mean fewer points of failure.

millems (approved):

PR #2622, opened 2021-07-23
Annotate and document members that are modeled as deprecated...

Bennett-Lynch:
...and add support for using customization to override specific members
as deprecated.

## Motivation and Context

As part of a recent change to repurpose the S3 CopyObjectRequest
copySource parameter into more user-friendly parameters, we lacked the
ability in our ShapeModifiersProcessor to also deprecate the old
parameter name. Furthermore, after additional investigation, we also
appear to not have been correctly honoring the different service-2.json
models that tag certain members as deprecated. E.g., the following
parameter is marked as deprecated:

https://github.com/aws/aws-sdk-java-v2/blob/54fd5a69c74af0124e97830bef53b3690dfa2015/services/cloudwatchlogs/src/main/resources/codegen-resources/service-2.json#L1393-L1398

But the corresponding generated code and Javadoc (at the time of this
writing) do not mark any of the member's methods as deprecated, or
expose the deprecatedMessage:

https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/cloudwatchlogs/model/FilterLogEventsRequest.Builder.html#interleaved-java.lang.Boolean-

## Description
* For service-2.json members that are marked as deprecated, correctly
annotate and document these fields as deprecated in their various
accessors and builder setters.
* Add support to ShapeModifiersProcessor for customization.config files
to be able to manually modify the `deprecated` and `deprecatedMessage`
attributes.

(A follow-up PR will propose marking the before-mentioned copySource
parameter as deprecated.)

## Testing
* Add a new operation with members that are modeled as deprecated and
members that are modified as deprecated, and add accompanying expected
generated code (see expected generated code for an example of how the
fields are annotated as deprecated).

## Types of changes

## Checklist

## License


codegen/src/main/java/software/amazon/awssdk/codegen/poet/model/DeprecationUtils.java
```diff
@@ -0,0 +1,37 @@
...
+    public static MethodSpec checkDeprecated(MemberModel member, MethodSpec method) {
+        if (!member.isDeprecated() || method.annotations.contains(DEPRECATED)) {
+            return method;
+        }
+        MethodSpec.Builder builder = method.toBuilder().addAnnotation(DEPRECATED);
+        if (!method.javadoc.isEmpty()) {
+            builder.addJavadoc(LF + "@deprecated");
+            if (StringUtils.isNotBlank(member.getDeprecatedMessage())) {
+                builder.addJavadoc("$L", " " + member.getDeprecatedMessage());
+            }
```
Bennett-Lynch, line 29: Should we add a default message if there is no `deprecatedMessage`?

codegen/src/main/java/software/amazon/awssdk/codegen/poet/model/AwsServiceModel.java
```diff
@@ -509,7 +510,7 @@ private void addCasesForMember(MethodSpec.Builder methodBuilder, MemberModel mem
 
         result.add(memberGetter(member));
 
-        return result.stream();
+        return checkDeprecated(member, result).stream();
```
Bennett-Lynch, line 513: I struggled a little bit with the decision on where to implement the `checkDeprecated` logic. Ideally it might make more sense to be defined at the `MemberModel` level, but I could not find a common entry or exit point that they could all safely share, and there exists other logic that will do something like `MemberModel#getDocuemtnation(..)` and then append more to the end, when it seems we want to guarantee `@deprecated` is consistently at the end. Ideas/opinions welcome.

codegen/src/main/java/software/amazon/awssdk/codegen/poet/model/DeprecationUtils.java
dagnir: I think I'm fine just leaving it blank

codegen/src/main/java/software/amazon/awssdk/codegen/poet/model/AwsServiceModel.java
dagnir: I actually prefer it the way you have it; we should keep the model objects as dumb as possible

codegen/src/main/java/software/amazon/awssdk/codegen/poet/model/DeprecationUtils.java
```diff
@@ -0,0 +1,37 @@
...
+     */
+    public static MethodSpec checkDeprecated(MemberModel member, MethodSpec method) {
+        if (!member.isDeprecated() || method.annotations.contains(DEPRECATED)) {
+            return method;
+        }
+        MethodSpec.Builder builder = method.toBuilder().addAnnotation(DEPRECATED);
+        if (!method.javadoc.isEmpty()) {
+            builder.addJavadoc(LF + "@deprecated");
+            if (StringUtils.isNotBlank(member.getDeprecatedMessage())) {
+                builder.addJavadoc("$L", " " + member.getDeprecatedMessage());
```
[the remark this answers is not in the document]
Bennett-Lynch: This works. Will update.

codegen/src/main/java/software/amazon/awssdk/codegen/poet/model/DeprecationUtils.java
```diff
@@ -0,0 +1,37 @@
+package software.amazon.awssdk.codegen.poet.model;
```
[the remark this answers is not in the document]
Bennett-Lynch: Thanks... need to configure my IDE for this.

dagnir (approved):

PR #2502, opened 2021-05-27
Update getCanonicalizedHeaderString to follow SigV4 spec

skrueger:
- Update to handle leading and trailing whitespace in header value
- Update to handle multiple header values

See Step 4 of https://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html
                                                                                                                         
Without this change a request to S3 with multiple header values or
a header value with leading and trailing spaces
(e.g.,
```
Header1: Cat\n
Header1: Dog\n
Header2:     "a   b   c"  \n
```
)
that is signed by the `AwsS3V4Signer` will receive a 403 Forbidden
response with an Error Code of SignatureDoesNotMatch from S3.

## Description
Currently my HTTP requests that contain multiple values for a specific header are receiving an 403 Forbidden response from S3 with an Error Code of `SignatureDoesNotMatch`. The S3's response would also return code would return the expected `<CanonicalRequest>`, which lead me to solving the issue.

## Motivation and Context
My requests to S3 are failing because the library does not correctly follow the SigV4 spec.

## Testing
* Added Unit Tests that cover the example given in the [SigV4 instructions on the AWS documentation website](https://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html), and also a test that covers multiple header values.
* Tested locally by sending requests to S3

## Screenshots

N/A

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)

## Checklist

- [x] I have read the **CONTRIBUTING** document
- [x] Local run of `mvn install` succeeds
- [x] My code follows the code style of this project
- [ ] My change requires a change to the Javadoc documentation
- [ ] I have updated the Javadoc documentation accordingly
- [x] I have read the **README** document
- [x] I have added tests to cover my changes
- [x] All new and existing tests passed
- [x] A short description of the change has been added to the **CHANGELOG**
- [ ] My change is to implement 1.11 parity feature and I have updated [LaunchChangelog](https://github.com/aws/aws-sdk-java-v2/blob/master/docs/LaunchChangelog.md)

## License

- [x] I confirm that this pull request can be released under the Apache 2 license

PR #2629, opened 2021-07-28
Correctly handle multi-value headers in Aws4Signer

Bennett-Lynch:
## Motivation and Context

The current AbstractAws4Signer does not correctly handle multi-value
headers. When creating the canonical header string to sign, it was
placing each value on a separate line, when the documentation requires
that they be comma-separated (even if the actual headers are serialized
on the wire as multiple header fields).

> Build the canonical headers list by sorting the (lowercase) headers by character code and then iterating through the header names. Construct each header according to the following rules:
> 
>* Append the lowercase header name followed by a colon.
> 
>* Append a comma-separated list of values for that header. Do not sort the values in headers that have multiple values.
> 
>* Append a new line ('\n').

https://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html

## Description

* Update AbstractAws4Signer to correctly join multiple header values
with a comma.
* Replace "appendCompactedString" with "trimAll", where "trimAll" is the
function name as specified in the documentation. This includes adding
String#trim() functionality.

## Result

Canonical header string example before:

```
foo:bar
foo:baz
```

after:

```
foo:bar,baz
```

## Testing

* Add new unit tests for multi-value headers
* Add new unit tests for the "Trimall" function, borrowing the example
from the docs

## Types of changes

## Checklist

## License

### DIFF
--- a/core/auth/src/main/java/software/amazon/awssdk/auth/signer/internal/AbstractAws4Signer.java
+++ b/core/auth/src/main/java/software/amazon/awssdk/auth/signer/internal/AbstractAws4Signer.java
@@ -25,6 +25,7 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.TreeMap;
+import java.util.stream.Collectors;
 import software.amazon.awssdk.annotations.SdkInternalApi;
 import software.amazon.awssdk.auth.credentials.AwsCredentials;
 import software.amazon.awssdk.auth.credentials.AwsSessionCredentials;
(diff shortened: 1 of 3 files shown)


core/auth/src/test/java/software/amazon/awssdk/auth/signer/Aws4SignerTest.java
```diff
@@ -156,6 +156,48 @@ public void xAmznTraceId_NotSigned() throws Exception {
                           "Signature=581d0042389009a28d461124138f1fe8eeb8daed87611d2a2b47fd3d68d81d73");
     }
 
+    /**
+     * Multi-value headers should be comma separated.
+     */
+    @Test
+    public void testMultiValueHeadersAreCommaSeparated() throws Exception {
```
zoewangg, line 163: Nit, `test` seems a bit redundant. We don't have a consistent naming pattern for unit tests in v2, but most of them follow similar format of `methodUnderTest_when_expectedBehavior` I think

zoewangg (approved):

core/auth/src/test/java/software/amazon/awssdk/auth/signer/Aws4SignerTest.java
Bennett-Lynch: Ack. Remnants of a JUnit 3 naming convention. I was trying to be mostly consistent with other unit tests in this class which did use `test...`, but let me update to use the more standard format.

Bennett-Lynch:
Also related to: https://github.com/aws/aws-sdk-java-v2/pull/2502

core/auth/src/main/java/software/amazon/awssdk/auth/signer/internal/AbstractAws4Signer.java
```diff
@@ -323,46 +324,51 @@ private String getCanonicalizedHeaderString(Map<String, List<String>> canonicali
...
+                sb.append(ch);
                 previousIsWhiteSpace = false;
             }
         }
+
+        return sb.toString().trim();
+    }
+
+    private List<String> trimAll(List<String> source) {
+        return source.stream().map(this::trimAll).collect(Collectors.toList());
```
skrueger: See the other comment about doing what https://github.com/aws/aws-sdk-java-v2/pull/2502/commits/632fea82eae1e3cafad6ab8bf7028d12eb45b31e to reduce allocations

But on the topic of streams: create a JMH micro bench harness yourself but I believe stream are going to perform worse than a `for` loop here. Also, I think `Collectors.toList()` allocates the default `ArrayList` size, which may be the wrong size (either too much or too little) which requires extra instructions.

core/auth/src/main/java/software/amazon/awssdk/auth/signer/internal/AbstractAws4Signer.java
```diff
@@ -323,46 +324,51 @@ private String getCanonicalizedHeaderString(Map<String, List<String>> canonicali
...
-                destination.append(ch);
+                sb.append(ch);
                 previousIsWhiteSpace = false;
             }
         }
+
+        return sb.toString().trim();
+    }
+
+    private List<String> trimAll(List<String> source) {
```
skrueger: Nitpick: static?

core/auth/src/main/java/software/amazon/awssdk/auth/signer/internal/AbstractAws4Signer.java
```diff
@@ -323,46 +324,51 @@ private String getCanonicalizedHeaderString(Map<String, List<String>> canonicali
...
-                appendCompactedString(buffer, headerName);
-                buffer.append(":");
-                if (headerValue != null) {
-                    appendCompactedString(buffer, headerValue);
-                }
-                buffer.append("\n");
-            }
+            buffer.append(headerName);
+            buffer.append(":");
+            buffer.append(String.join(",", trimAll(headerValues)));
```
skrueger, line 329: Again This is doing some extra memory allocations for the `String.join` that can be avoided. See https://github.com/aws/aws-sdk-java-v2/pull/2502/commits/632fea82eae1e3cafad6ab8bf7028d12eb45b31e

skrueger (commented):
Good job fixing this!

I'd prefer an implementation that reduces the number of memory allocations to perform the algorithm.

Issue #2625, opened 2021-07-26
Release 2.17 wrongfully removed EC2MetadataUtils#getInstanceInfo


slandelle:
## Describe the bug

[This commit](https://github.com/aws/aws-sdk-java-v2/commit/51cf791235c6bf408d40df35411eb11f12839eef#diff-aa35187dc0d550f218a761a04b65ab1c79c6804221535a1c66da49feb8475a1aL227), described as some internal clean up in order to remove Jackson dependencies.

Sadly, it also performed some other unrelated **functional** changes. In particular, it removed `EC2MetadataUtils#getInstanceInfo` (and some other methods too).

## Expected Behavior

* No functional change should occur in a PR that claims to just be a refactoring in order to change the JSON parser.
* Removing features shouldn't be performed without providing a migration path.

## Current Behavior

We contributed a change in `EC2MetadataUtils` 2 years ago in order to expose the AWS MarkerPlace product codes that were in the response payload but not mapped in the POJO.

This data is critical for us as AWS MarketPlace vendors.
This feature was removed without any further notice, without considering impacts (in particular for us who appear in the git history of this class) and without providing a migration path.

## Steps to Reproduce

Please read the changes in [EC2MetatadataUtils](https://github.com/aws/aws-sdk-java-v2/commit/51cf791235c6bf408d40df35411eb11f12839eef#diff-aa35187dc0d550f218a761a04b65ab1c79c6804221535a1c66da49feb8475a1aL227).

## Possible Solution

Someone might argue that this method is an internal and users shouldn't it it. However:

* using this REST API was the means the AWS MarketPlace people told us to use to fetch the AWS MarketPlace product codes
* the AWS Java SDK doesn't provide any other means to achieve this
* our contributions were reviewed and merged, you can't just simply remove them without considering the impact

Ideally, this SDK should provide a public stable API for fetching Marketplace ProductCodes. Or make `EC2MetatadataUtils` a stable API.

In the meantime, please restore this method.

## Context

We've been using this class for years as it's the only way for AWS MarketPlace AMI vendors to fetch the AWS MarketPlace product codes. We even [contributed](https://github.com/aws/aws-sdk-java-v2/commit/f6e9cca46e7744b31ce6fbf263fc7e9287dc01b3#diff-aa35187dc0d550f218a761a04b65ab1c79c6804221535a1c66da49feb8475a1a) to expose the field in the `InstanceInfo` object.

[...]

Issue #61, opened 2017-07-11
EC2 Metadata Client


davidmankin:
Since all access to EC2MetadataUtils is static, the class is very difficult to use in a properly tested application.  It can't be mocked; you can't give canned answers when running in a non-EC2 environment.

If there were an EC2Metadata interface or abstract class of which there were a default implementation, then it would be easier to use.  

A contrived example:
```java
public class Application {
    EC2Metadata metadata;
    public static void main(String args[]) {
        new Application(EC2MetadataUtils.getDefaultInstance());
    }
    public Application(EC2Metadata metadata) {
        this.metadata = metadata;
        LOG.info("We're running on " + metadata.getInstanceId());
    }
}

public class ApplicationTest {
    @Mock EC2Metadata metadata;
    @Test public testApp() {
        when(metadata.getDefaultInstance()).then(return "i-fakeinstance");
        new Application(metadata);
        verifyLoggedLine("We're running on i-fakeinstance");
    }
}
```

As it stands today, I need to run a whole HTTP process that mimics the metadata server and somehow get the system property to point to it; and even that is a global setting for the whole JVM: parallel tests can interfere with each other.  Or I can write my own EC2Metadata interface and use it throughout my app, but it's not something each developer should have to do.

millems:
Using this issue to track building a public-facing EC2 Metadata Client.

dagnir:
Sorry for the delay in our response. We aim to respond to all comments, feedback and pull requests our customers make and we apologize that we failed to do so in this instance.

We have been working with the EC2 team to figure out the best possible customer experience for exposing this information, and in the meantime we welcome all suggestions and will use them to help us in that decision making process.

We will keep updating this issue as we move towards delivery.

Thank you for your patience!

injeniero:
@dagnir Any update on this?

spender0:
any progress? It turned out that we need to change metadata_service_timeout as the connection in our case takes more than 1 second. But it is hardcoded in the SDK

sanjeeb26:
@dagnir do you have timeline when this will be available in the SDK?

dFarras:
@millems @dagnir is there any new about this issue?

Thanks in advance

millems:
@dFarras Sorry, not right now. What specific feature do you need (and why do you need it), so we can make sure that's included when this gets done?

dFarras:
My use case is that I need to retrieve an instance id but I can not be sure were the instance is deployed in advance. 

[...]

PR #2630, opened 2021-07-28
Temporarily restore EC2MetadataUtils.getInstanceInfo for #2625

millems:
…until #61 can be completed, at which time it will have to go away again.

### DIFF
--- a/core/regions/src/main/java/software/amazon/awssdk/regions/internal/util/EC2MetadataUtils.java
+++ b/core/regions/src/main/java/software/amazon/awssdk/regions/internal/util/EC2MetadataUtils.java
@@ -283,6 +283,63 @@ public final class EC2MetadataUtils {
         return getData(EC2_USERDATA_ROOT);
     }
 
+    /**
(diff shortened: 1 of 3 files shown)


core/regions/src/main/java/software/amazon/awssdk/regions/internal/util/EC2MetadataUtils.java
```diff
@@ -587,4 +638,120 @@ public URI endpoint() {
             return requestHeaders;
         }
     }
+
+
```
joviegas, line 642: Optional : The comment for this class is missing as compared to the original one

joviegas (approved):

core/regions/src/main/java/software/amazon/awssdk/regions/internal/util/EC2MetadataUtils.java
millems: Added a comment

PR #2644, opened 2021-08-05
Improve documentation of has* methods based on customer feedback.

millems:


cenedhryn (approved):

codegen/src/main/java/software/amazon/awssdk/codegen/internal/DocumentationUtils.java
```diff
@@ -36,10 +36,12 @@
...
     private static final String DEFAULT_GETTER_PARAM = "The value of the %s property for this object.";
 
-    private static final String DEFAULT_EXISTENCE_CHECK = "Returns true if the %s property was specified by the sender "
-                                                          + "(it may be empty), or false if the sender did not specify "
-                                                          + "the value (it will be empty). For responses returned by the SDK, "
-                                                          + "the sender is the AWS service.";
+    private static final String DEFAULT_EXISTENCE_CHECK =
+        "For responses, this returns true if the service returned a value for the %s property. This DOES NOT check that the "
+        + "value is non-empty (for which, you should check the {@code isEmpty()} method on the property). This is useful because "
+        + "the SDK will never return an empty collection or map, but you may need to differentiate between the service returning "
```
Bennett-Lynch, line 42: I think you meant `null` here?

> `the SDK will never return an empty collection or map`

millems: +1, good catch

millems: Fixed

Bennett-Lynch (approved):

PR #2646, opened 2021-08-05
 Fixed an issue where checksum validation only considered the first 4 bytes of the 16 byte checksum...

millems:
...creating the potential for corrupted downloads to go undetected.

### DIFF
--- a/services/s3/src/main/java/software/amazon/awssdk/services/s3/checksums/ChecksumValidatingInputStream.java
+++ b/services/s3/src/main/java/software/amazon/awssdk/services/s3/checksums/ChecksumValidatingInputStream.java
@@ -17,11 +17,12 @@ package software.amazon.awssdk.services.s3.checksums;
 
 import java.io.IOException;
 import java.io.InputStream;
-import java.nio.ByteBuffer;
+import java.util.Arrays;
 import software.amazon.awssdk.annotations.SdkInternalApi;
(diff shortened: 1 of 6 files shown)


services/s3/src/main/java/software/amazon/awssdk/services/s3/checksums/ChecksumValidatingInputStream.java
```diff
@@ -162,26 +162,15 @@ public void close() throws IOException {
...
+            computedChecksum = checkSum.getChecksumBytes();
         }
 
-        if (streamChecksumInt != computedChecksum) {
+        if (!Arrays.equals(computedChecksum, streamChecksum)) {
             throw SdkClientException.builder().message(
-                String.format("Data read has a different checksum than expected. Was %d, but expected %d",
-                              computedChecksum, streamChecksumInt)).build();
+                String.format("Data read has a different checksum than expected. Was %s, but expected %s",
+                              Arrays.toString(computedChecksum), Arrays.toString(streamChecksum))).build();
```
dagnir, line 173: minor: should we represent as a hex string instead? Would make the error message shorter and bit easier to digest easier to digest

millems: +1

dagnir (approved):

PR #2652, opened 2021-08-10
Move IntelliJ code style & copyright configuration to project-level settings

Bennett-Lynch:
## Motivation

We currently expose IntelliJ code style & copyright configuration as
standalone config files that users must manually import into their IDE.
The import process is not too tedious, since the files are stored in
Git (and thus readily available at the project level), but it requires
users to know to find and import these files, and it's an unnecessary
manual step for new users to be able to contribute.

IntelliJ makes it easy to share configuration at the project-level. With
its .idea directory structure, we can manually choose which files to
share in Git. Then any user cloning the project will automatically use
our recommended settings.

Another benefit of this approach is that any changes to the existing
code style will be automatically shared and synced with users when
they pull the latest version.

## Modifications

* Copy intellij-codestyle.xml to project-level settings (and remove the
old file)
* Copy intellij-copyright-profile.xml to project-level settings (and
remove the old file)
* Update gitignore to not ignore the directories for these specific
project-level settings
* Update the GettingStarted guide accordingly

## Notes

To generate the project-level code style, I used the existing
intellij-codestyle.xml and performed the IntelliJ "copy to project"
action. This resulted in some slight differences, like reordering of
elements and elements like "GENERATE_FINAL_LOCALS" no longer being
present, but from what I can tell this is due to them specifying the
same as the default behavior.

## Resources

https://stackoverflow.com/questions/9087439/how-to-share-code-style-settings-between-developers-in-intellij

https://intellij-support.jetbrains.com/hc/en-us/articles/206544839

https://stackoverflow.com/questions/5533050/gitignore-exclude-folder-but-include-specific-subfolder

## License


millems (approved):

dagnir (approved):

Bennett-Lynch:
Removed copyright header from the config files themselves since it will be clobbered/overwritten whenever we change a setting preference, creating unnecessary churn (either in the diff or in having to manually preserve it).

PR #2657, opened 2021-08-11
Fixed an issue where request-level overrides (e.g. credentials) were not applied to endpoint discovery calls.

millems:


codegen/src/test/resources/software/amazon/awssdk/codegen/poet/endpointdiscovery/test-async-cache-loader.java
```diff
@@ -27,18 +28,18 @@ public static EndpointDiscoveryTestAsyncEndpointDiscoveryCacheLoader create(Endp
 
     @Override
     public CompletableFuture<EndpointDiscoveryEndpoint> discoverEndpoint(EndpointDiscoveryRequest endpointDiscoveryRequest) {
+        AwsRequestOverrideConfiguration requestConfig = AwsRequestOverrideConfiguration.from(endpointDiscoveryRequest
+                                                                                                 .overrideConfiguration().orElse(null));
```
dagnir, line 32: One issue might be that request-level overrides like headers/query params could have unintended effects on the `DescribeEndpoints` request. What do you think about erring on the conservative side and just copying the credentials provider if it's present?

millems: I'm worried about only copying some things, because different people might have different expectations about what is used and not used. Why *shouldn't* the headers and query parameters be used for the discovery call? What about the API names (for user agent)? Signer? Metrics publishers? Some of these things we almost certainly want to include on the discovery call, so I'd favor just including them all so that there's no confusion.

Also, if we try to pick-and-choose now, when we add new stuff in the future, we'll definitely forget to ask "does this make sense to copy over for endpoint discovery?".

dagnir: Yep good points; I can't think of a use case where the other things on the override config would cause issues with the request, but I thought it was a valid concern. I'm in favor of just copying the override config as well.

dagnir (approved):

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/endpointdiscovery/test-async-cache-loader.java
millems: I don't mean to disregard your concern. I totally agree it's a valid concern. I'm just less fond of the other alternatives.

PR #2660, opened 2021-08-12
Fixed connection pooling in the UrlConnectionHttpClient.

millems:


test/http-client-tests/src/main/java/software/amazon/awssdk/http/SdkHttpClientTestSuite.java
```diff
@@ -113,6 +120,30 @@ public void validatesHttpsCertificateIssuer() throws Exception {
...
+            HttpExecuteResponse response =
+                client.prepareRequest(HttpExecuteRequest.builder()
+                                                        .request(req)
+                                                        .contentStreamProvider(req.contentStreamProvider().orElse(null))
+                                                        .build())
+                      .call();
+            response.responseBody().ifPresent(IoUtils::drainInputStream);
+        }
+
+        assertThat(CONNECTION_COUNTER.openedConnections()).isLessThan(initialOpenedConnections + 5);
```
Bennett-Lynch, line 144: Can we be more precise than `lessThan` here? It seems like this should be deterministic. Do we expect exactly `initialOpenedConnections`, `initialOpenedConnections + 1`?

http-clients/url-connection-client/src/main/java/software/amazon/awssdk/http/urlconnection/UrlConnectionHttpClient.java
```diff
@@ -71,18 +72,18 @@
...
             this.connectionFactory = connectionFactory;
         } else {
-            this.sslContext = getSslContext(options);
-            this.connectionFactory = this::createDefaultConnection;
-        }
+            // Note: This socket factory MUST be reused between requests because the connection pool in the JVM is keyed by both
+            // URL and SSLSocketFactory. If the socket factory is not reused, connections will not be reused between requests.
+            SSLSocketFactory socketFactory = getSslContext(options).getSocketFactory();
 
+            this.connectionFactory = url -> createDefaultConnection(url, socketFactory);
```
Bennett-Lynch, line 85: Awesome find and nice documentation.

Bennett-Lynch (approved):

test/http-client-tests/src/main/java/software/amazon/awssdk/http/SdkHttpClientTestSuite.java
Bennett-Lynch: I guess the number may depend (and thus vary) upon the client implementation, and how many times we're willing to reuse a connection. Is it fair to say that no client ought to use more than 1 *new* connection here though? Can we do: `isLessThan(initialOpenedConnections + 1)`?

millems: In my tests it was fairly consistently `+1` when pooling was working, or `+5` when it was not working. 

Because the test is so indirect (involves cooperation of various HTTP client implementations as well as Wiremock/Jetty, I wasn't confident that hard-coding it to `+1` was going to create a 100% deterministic test, and non-deterministic tests are worse than no test (IMO). I think if it's < 5 we can confidently say that connection pooling is working, but you're right that we don't know to what extent it's working. With the test how it currently is, we could make a change that makes pooling "worse" (whatever that means) without this test catching it, unless it completely breaks pooling altogether.

Let me try running it a few thousand times to see if it ever fails at `+1`.

millems: I switched it to +1, it seemed to work just fine. We can reassess if it becomes a problem.

PR #2451, opened 2021-05-01
Revert "Add support for event stream requests over RPC"

dagnir:
This reverts commit 8763931a0fe92081f7ddfc7f6d18213da55c92fd.

## Description
Reverting due to an issue with the content-type on input events being incorrect.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)

## Checklist

- [ ] I have read the **CONTRIBUTING** document
- [ ] Local run of `mvn install` succeeds
- [ ] My code follows the code style of this project
- [ ] My change requires a change to the Javadoc documentation
- [ ] I have updated the Javadoc documentation accordingly
- [ ] I have read the **README** document
- [ ] I have added tests to cover my changes
- [ ] All new and existing tests passed
- [ ] A short description of the change has been added to the **CHANGELOG**
- [ ] My change is to implement 1.11 parity feature and I have updated [LaunchChangelog](https://github.com/aws/aws-sdk-java-v2/blob/master/docs/LaunchChangelog.md)

## License

- [x] I confirm that this pull request can be released under the Apache 2 license

PR #2662, opened 2021-08-12
Add support for event stream requests over RPC

millems:
This reverts the revert in https://github.com/aws/aws-sdk-java-v2/pull/2451/files, which was done because the original change broke Lex's event streaming APIs: it changed the content-type in the event stream to application/x-amz-json-1.1, which Lex couldn't handle because it is a rest-json service.

This will no longer break Lex because of the second commit included in this customizing the content-type for Lex to application/json. This also affects non-event-streaming operations.

I verified that after this change, Lex's deleteSession (non-streaming) and startConversation (event streaming) APIs both work.

In the long-term, we should change the default content-type for all rest-json services to application/json and remove this customization, but that's a longer-term project.


dagnir:
Can you run the integ tests to make sure Transcribe is working?

millems:
Transcribe Streaming and Kinesis integration tests pass.

dagnir (approved):

PR #2666, opened 2021-08-18
Adding batchManager() method into code generated sync and async clients

16lim21:
## Motivation and Context
Users should have the option to be able to create a `batchManager()` directly from a low level sqs client as outlined in the design document. Changes also had to be made to the generated clients to ensure that they contained the necessary executor/scheduledExecutor that would be used by the `batchManager`.

## Description
Added some changes to how the sync and async clients (as well as their corresponding interfaces) are generated. Specifically added in the `batchManager()` method to create a `batchManager` for that service. `Executors` and `ScheduledExecutors` were also added to the underlying clients if they supported the `batchManager()` method.

## Testing
Added some codegen tests to check the generated classes and interfaces matched expectations. Also modified existing SqsBatchManager tests to use the client's batchManager instead of creating one separately.

## Types of changes

## Checklist

## License

### DIFF
--- a/codegen/src/main/java/software/amazon/awssdk/codegen/poet/client/AsyncClientClass.java
+++ b/codegen/src/main/java/software/amazon/awssdk/codegen/poet/client/AsyncClientClass.java
@@ -50,6 +50,7 @@ import software.amazon.awssdk.awscore.client.handler.AwsAsyncClientHandler;
 import software.amazon.awssdk.awscore.client.handler.AwsClientHandlerUtils;
 import software.amazon.awssdk.awscore.eventstream.EventStreamTaggedUnionJsonMarshaller;
 import software.amazon.awssdk.codegen.emitters.GeneratorTaskParams;
+import software.amazon.awssdk.codegen.model.config.customization.BatchManagerMethod;
 import software.amazon.awssdk.codegen.model.config.customization.UtilitiesMethod;
 import software.amazon.awssdk.codegen.model.intermediate.IntermediateModel;
 import software.amazon.awssdk.codegen.model.intermediate.MemberModel;
@@ -79,6 +80,7 @@ import software.amazon.awssdk.utils.CompletableFutureUtils;
 import software.amazon.awssdk.utils.FunctionalUtils;
 
 public final class AsyncClientClass extends AsyncClientInterface {
+    private boolean hasScheduledExecutor = false;
     private final IntermediateModel model;
     private final PoetExtensions poetExtensions;
     private final ClassName className;
@@ -144,7 +146,17 @@ public final class AsyncClientClass extends AsyncClientInterface {
             classBuilder.addField(FieldSpec.builder(ClassName.get(ScheduledExecutorService.class), "executorService")
                                                    .addModifiers(PRIVATE, FINAL)
                                                    .build());
+            hasScheduledExecutor = true;
             classBuilder.addMethod(waiterImplMethod());
+        }
+
+        if (model.getCustomizationConfig().getBatchManagerMethod() != null) {
+            classBuilder.addMethod(batchMangerMethod());
+            if (!hasScheduledExecutor) {
+                classBuilder.addField(FieldSpec.builder(ClassName.get(ScheduledExecutorService.class), "executorService")
+                                               .addModifiers(PRIVATE, FINAL)
+                                               .build());
+            }
         }
 
         return classBuilder.build();
@@ -196,6 +208,11 @@ public final class AsyncClientClass extends AsyncClientInterface {
         }
 
         if (model.hasWaiters()) {
+            builder.addStatement("this.executorService = clientConfiguration.option($T.SCHEDULED_EXECUTOR_SERVICE)",
+                                 SdkClientOption.class);
+        }
+
+        if (model.getCustomizationConfig().getBatchManagerMethod() != null && !hasScheduledExecutor) {
             builder.addStatement("this.executorService = clientConfiguration.option($T.SCHEDULED_EXECUTOR_SERVICE)",
                                  SdkClientOption.class);
         }
@@ -484,5 +501,20 @@ public final class AsyncClientClass extends AsyncClientInterface {
     private boolean hasStreamingV4AuthOperations() {
         return model.getOperations().values().stream()
                 .anyMatch(this::shouldUseAsyncWithBodySigner);
+    }
+
+    private MethodSpec batchMangerMethod() {
+        String scheduledExecutor = "executorService";
+
+        BatchManagerMethod config = model.getCustomizationConfig().getBatchManagerMethod();
+        ClassName returnType = PoetUtils.classNameFromFqcn(config.getAsyncReturnType());
+
+        return MethodSpec.methodBuilder(BatchManagerMethod.METHOD_NAME)
+                         .returns(returnType)
+                         .addModifiers(Modifier.PUBLIC)
+                         .addAnnotation(Override.class)
+                         .addStatement("return $T.builder().client(this).scheduledExecutor($N).build()",
+                                       returnType, scheduledExecutor)
+                         .build();
     }
 }
--- a/codegen/src/main/java/software/amazon/awssdk/codegen/poet/client/AsyncClientClass.java
+++ b/codegen/src/main/java/software/amazon/awssdk/codegen/poet/client/AsyncClientClass.java
@@ -23,6 +23,7 @@ import static javax.lang.model.element.Modifier.STATIC;
 import static software.amazon.awssdk.codegen.poet.client.ClientClassUtils.addS3ArnableFieldCode;
 import static software.amazon.awssdk.codegen.poet.client.ClientClassUtils.applyPaginatorUserAgentMethod;
 import static software.amazon.awssdk.codegen.poet.client.ClientClassUtils.applySignerOverrideMethod;
+import static software.amazon.awssdk.codegen.poet.client.ClientClassUtils.batchMangerMethod;
 import static software.amazon.awssdk.codegen.poet.client.SyncClientClass.getProtocolSpecs;
 
 import com.squareup.javapoet.ClassName;
@@ -50,7 +51,6 @@ import software.amazon.awssdk.awscore.client.handler.AwsAsyncClientHandler;
 import software.amazon.awssdk.awscore.client.handler.AwsClientHandlerUtils;
 import software.amazon.awssdk.awscore.eventstream.EventStreamTaggedUnionJsonMarshaller;
 import software.amazon.awssdk.codegen.emitters.GeneratorTaskParams;
-import software.amazon.awssdk.codegen.model.config.customization.BatchManagerMethod;
 import software.amazon.awssdk.codegen.model.config.customization.UtilitiesMethod;
 import software.amazon.awssdk.codegen.model.intermediate.IntermediateModel;
 import software.amazon.awssdk.codegen.model.intermediate.MemberModel;
@@ -80,7 +80,6 @@ import software.amazon.awssdk.utils.CompletableFutureUtils;
 import software.amazon.awssdk.utils.FunctionalUtils;
 
 public final class AsyncClientClass extends AsyncClientInterface {
-    private boolean hasScheduledExecutor = false;
     private final IntermediateModel model;
     private final PoetExtensions poetExtensions;
     private final ClassName className;
@@ -146,18 +145,12 @@ public final class AsyncClientClass extends AsyncClientInterface {
             classBuilder.addField(FieldSpec.builder(ClassName.get(ScheduledExecutorService.class), "executorService")
                                                    .addModifiers(PRIVATE, FINAL)
                                                    .build());
-            hasScheduledExecutor = true;
             classBuilder.addMethod(waiterImplMethod());
+            addBatchManagerMethodAndScheduledExecutorIfNeeded(classBuilder, true);
+        } else {
+            addBatchManagerMethodAndScheduledExecutorIfNeeded(classBuilder, false);
         }
 
-        if (model.getCustomizationConfig().getBatchManagerMethod() != null) {
-            classBuilder.addMethod(batchMangerMethod());
-            if (!hasScheduledExecutor) {
-                classBuilder.addField(FieldSpec.builder(ClassName.get(ScheduledExecutorService.class), "executorService")
-                                               .addModifiers(PRIVATE, FINAL)
-                                               .build());
-            }
-        }
 
         return classBuilder.build();
     }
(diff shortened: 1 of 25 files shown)


codegen/src/main/java/software/amazon/awssdk/codegen/poet/client/AsyncClientClass.java
```diff
@@ -144,9 +146,19 @@ public TypeSpec poetSpec() {
...
         }
 
+        if (model.getCustomizationConfig().getBatchManagerMethod() != null) {
+            classBuilder.addMethod(batchMangerMethod());
+            if (!hasScheduledExecutor) {
+                classBuilder.addField(FieldSpec.builder(ClassName.get(ScheduledExecutorService.class), "executorService")
+                                               .addModifiers(PRIVATE, FINAL)
+                                               .build());
+            }
+        }
```
zoewangg, line 160: Can we extract the logic to a separate method, say `addScheduledExecutorIfNeeded` and set/check `hasScheduledExecutor` there? It's easier to manage that way

codegen/src/main/java/software/amazon/awssdk/codegen/poet/client/AsyncClientClass.java
```diff
@@ -200,6 +212,11 @@ private MethodSpec constructor(Builder classBuilder) {
                                  SdkClientOption.class);
         }
 
+        if (model.getCustomizationConfig().getBatchManagerMethod() != null && !hasScheduledExecutor) {
```
zoewangg, line 215: Same comment here

codegen/src/main/java/software/amazon/awssdk/codegen/poet/client/SyncClientClass.java
```diff
@@ -390,4 +404,19 @@ private MethodSpec waiterMethod() {
...
+        ClassName returnType = PoetUtils.classNameFromFqcn(config.getReturnType());
+
+        return MethodSpec.methodBuilder(BatchManagerMethod.METHOD_NAME)
+                         .returns(returnType)
+                         .addModifiers(Modifier.PUBLIC)
+                         .addAnnotation(Override.class)
+                         .addStatement("return $T.builder().client(this).executor($N).scheduledExecutor($N).build()",
+                                       returnType, scheduledExecutor, scheduledExecutor)
+                         .build();
+    }
```
zoewangg, line 421: This looks very similar to the one in async client class. Can we reuse it?

services/sqs/src/test/java/software/amazon/awssdk/services/sqs/SqsAsyncBatchManagerTest.java
```diff
@@ -104,15 +98,15 @@ public void tearDown() {
...
+        SqsAsyncBatchManager batchManager = new DefaultSqsAsyncBatchManager(client,
                                                                             mockSendMessageBatchManager,
                                                                             mockDeleteMessageBatchManager,
                                                                             mockChangeVisibilityBatchManager);
         batchManager.close();
         verify(mockSendMessageBatchManager).close();
         verify(mockDeleteMessageBatchManager).close();
         verify(mockChangeVisibilityBatchManager).close();
-        assertThat(scheduledExecutor.isShutdown()).isFalse();
+        assertThatCode(() -> client.serviceName()).doesNotThrowAnyException(); // To make sure client is not closed
```
zoewangg, line 109: I think if you close the client, you can still get the serviceName. Can we create a mock client and verify that client.close() is not invoked?

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/client/test-batchmanager-async-class.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    private HttpResponseHandler<AwsServiceException> createErrorResponseHandler(BaseAwsJsonProtocolFactory protocolFactory,
+                                                                                JsonOperationMetadata operationMetadata) {
+        return protocolFactory.createErrorResponseHandler(operationMetadata);
+    }
+
+    @Override
+    public SqsAsyncBatchManager batchManager() {
+        return SqsAsyncBatchManager.builder().client(this).scheduledExecutor(executorService).build();
```
zoewangg, line 89: Can we use some other fake class here for testing? It is a bit confusing to use sqs batch manager here, to me

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/client/test-batchmanager-sync-class.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    @Override
+    public void close() {
+        clientHandler.close();
+    }
+
+    @Override
+    public SqsBatchManager batchManager() {
+        return SqsBatchManager.builder().client(this).executor(executorService).scheduledExecutor(executorService).build();
```
zoewangg, line 86: I'm not sure if using the scheduled executor service to send blocking request is a good idea. I was initially thinking of using the existing async response completion pool, but looks like that only gets set in async client,  so we should create a default executor service in each sync batch manager and close it when the batch manager gets closed.

zoewangg (commented):
Don't we need to modify the customization config in sqs?

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/client/test-batchmanager-sync-class.java
16lim21: Yeah I noticed that too, that's why I switched to using the scheduled executor service. Will modify to create a default executor service.

codegen/src/main/java/software/amazon/awssdk/codegen/poet/client/AsyncClientClass.java
16lim21: Extracted to a `addBatchManagerMethodAndScheduledExecutorIfNeeded` method. hasScheduledExecutor is passed as method parameter.

codegen/src/main/java/software/amazon/awssdk/codegen/poet/client/AsyncClientClass.java
16lim21: Addressed with change above.

codegen/src/main/java/software/amazon/awssdk/codegen/poet/client/SyncClientClass.java
16lim21: Reused the code by moving method into `ClientClassUtils`.

services/sqs/src/test/java/software/amazon/awssdk/services/sqs/SqsAsyncBatchManagerTest.java
16lim21: Created mock client and verified c`lient.close() `is not invoked.

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/client/test-batchmanager-async-class.java
16lim21: Used another fake class for testing.

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/client/test-batchmanager-sync-class.java
16lim21: Currently creating a single thread executor inside sync batch manager and closing it when the client is closed. Not sure if single thread should be the expected behavior, could also make a executor with a maxPoolSize equal to the client's maxConnections?

codegen/src/main/java/software/amazon/awssdk/codegen/poet/client/AsyncClientClass.java
```diff
@@ -207,12 +200,7 @@ private MethodSpec constructor(Builder classBuilder) {
...
 
-        if (model.hasWaiters()) {
-            builder.addStatement("this.executorService = clientConfiguration.option($T.SCHEDULED_EXECUTOR_SERVICE)",
-                                 SdkClientOption.class);
-        }
-
-        if (model.getCustomizationConfig().getBatchManagerMethod() != null && !hasScheduledExecutor) {
+        if (model.hasWaiters() || model.getCustomizationConfig().getBatchManagerMethod() != null) {
             builder.addStatement("this.executorService = clientConfiguration.option($T.SCHEDULED_EXECUTOR_SERVICE)",
                                  SdkClientOption.class);
```
zoewangg, line 205: Can we call `addScheduledExecutorIfNeeded` here?

codegen/src/main/java/software/amazon/awssdk/codegen/poet/client/AsyncClientClass.java
```diff
@@ -503,18 +491,14 @@ private boolean hasStreamingV4AuthOperations() {
...
-        ClassName returnType = PoetUtils.classNameFromFqcn(config.getAsyncReturnType());
-
-        return MethodSpec.methodBuilder(BatchManagerMethod.METHOD_NAME)
-                         .returns(returnType)
-                         .addModifiers(Modifier.PUBLIC)
-                         .addAnnotation(Override.class)
-                         .addStatement("return $T.builder().client(this).scheduledExecutor($N).build()",
-                                       returnType, scheduledExecutor)
-                         .build();
+    private void addBatchManagerMethodAndScheduledExecutorIfNeeded(Builder classBuilder, boolean hasScheduledExecutor) {
```
zoewangg, line 494: Would it make sense to separate them so that waiter can reuse the `addScheduledExecutorIfNeeded` and `hasScheduledExecutor` will be managed by `addScheduledExecutorIfNeeded` only

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/client/test-batchmanager-sync-class.java
```diff
@@ -0,0 +1,97 @@
...
+
+    private final ExecutorService executor;
+
+    protected DefaultBatchManagerClient(SdkClientConfiguration clientConfiguration) {
+        this.clientHandler = new AwsSyncClientHandler(clientConfiguration);
+        this.clientConfiguration = clientConfiguration;
+        this.protocolFactory = init(AwsJsonProtocolFactory.builder()).build();
+        this.executorService = clientConfiguration.option(SdkClientOption.SCHEDULED_EXECUTOR_SERVICE);
+        ThreadFactory threadFactory = new ThreadFactoryBuilder().threadNamePrefix("DefaultBatchManagerClient").build();
+        this.executor = Executors.newSingleThreadExecutor(threadFactory);
```
zoewangg, line 54: We should only create it in the SqsBatchManager class and not the client class. Also, single thread executor might not be the best choice here since all requests will be blocking call.  Let's create the same one as defined here https://github.com/aws/aws-sdk-java-v2/blob/39d6f2a7d3c98bffff57c574c5bf1c63bc72e698/services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/S3NativeClientConfiguration.java#L118

16lim21: Ok that makes sense, I'll remove it from the codegen and add it in the handcoded SqsBatchManager for now.

codegen/src/main/java/software/amazon/awssdk/codegen/poet/client/AsyncClientClass.java
16lim21: I'm not entirely sure why this would need to be refactored into another method? Adding the statement is only done here.

codegen/src/main/java/software/amazon/awssdk/codegen/poet/client/AsyncClientClass.java
16lim21: Separated method into `addScheduledExecutorIfNeeded` and `addBatchManagerMethod`

16lim21:
Sonarcloud fails due to 4 lines of duplicated code across the SyncClientInterface and AsyncClientInterface. The duplicated code just gets the batchManagerMethod which and adds it.

codegen/src/main/java/software/amazon/awssdk/codegen/poet/client/AsyncClientClass.java
zoewangg: Oh I got confused 😅 . I thought this was the same as adding scheduled executor service

services/sqs/src/main/java/software/amazon/awssdk/services/sqs/internal/batchmanager/DefaultSqsBatchManager.java
```diff
@@ -54,18 +60,17 @@
...
                                                                                      .maxBatchItems(config.maxBatchItems())
                                                                                      .maxBatchOpenInMs(config.maxBatchOpenInMs())
                                                                                      .build();
-
-
         ScheduledExecutorService scheduledExecutor = builder.scheduledExecutor;
-        this.executor = builder.executor;
-
+        ThreadFactory threadFactory = new ThreadFactoryBuilder().threadNamePrefix("DefaultSqsBatchManager").build();
+        this.executor = createDefaultExecutor(threadFactory);
```
zoewangg, line 73: We probably should only create a default executor if `builder.executor` is null and only close the executor if it's the default one created by the class

16lim21: Only creating and shutting down default executor if builder.executor is null. Adding tests to check if executors are properly shut down.

services/sqs/src/main/java/software/amazon/awssdk/services/sqs/batchmanager/SqsBatchManager.java
```diff
@@ -134,7 +134,7 @@ static Builder builder() {
          * @param executor the executor to be used.
          * @returna reference to this object so that method calls can be chained together.
          */
-        Builder executor(Executor executor);
+        Builder executor(ExecutorService executor);
```
zoewangg: Hmm, why did we revert it back to `Executor`?  If lack of shutdown is the reason, we can cast the created executor to executor service and call shutdown. https://github.com/aws/aws-sdk-java-v2/blob/master/utils/src/main/java/software/amazon/awssdk/utils/AttributeMap.java#L91-L93

16lim21: I originally changed it to `Executor` when I was trying to access the client's `FUTURE_COMPLETION_EXECUTOR`. Since they sync client doesn't have that executor, it was ok to revert back to an `ExecutorService`.

zoewangg (approved):

PR #2693, opened 2021-09-02
Improve performance of generated enum methods

Bennett-Lynch:
## Motivation

Generated enums use a standard `fromValue` method to resolve a string
value (usually received on the wire) to the corresponding enum. The
existing implementation is not optimal for a few reasons:

1. It calls `Enum#values()`, which must allocate a new array every time
2. It iterates the entire array with O(N) complexity

The generated `knownValues` follows a similar pattern, plus it does not
take advantage of `EnumSet`, which is typically faster than HashSet.

These are minor optimization concerns, however many code paths
repeatedly call `fromValue()` (as opposed to caching it).

## Modifications

* For each generated enum, store a value map which maps the string value
to the corresponding enum constant value. Use this map for `fromValue`.
* Use an EnumSet for `knownValues`.
* Create new `CollectionUtils.index()` & `EnumUtils.index()` methods to
facilitate the common use case of constructing a unique index lookup map

## Result

* `fromValue` changes from O(N) to O(1)
* Minor increase in memory overhead from storing a static map, but
reduced allocation from creating new arrays every method call

## Types of changes

## Checklist

## License


utils/src/main/java/software/amazon/awssdk/utils/CollectionUtils.java
```diff
@@ -139,4 +140,26 @@ public static boolean isNullOrEmpty(Map<?, ?> map) {
...
+     * @param <V> the value type
+     * @return a (modifiable) map that indexes K to its unique value V
+     * @throws IllegalArgumentException if any of the values map to the same index
+     */
+    public static <K, V> Map<K, V> index(Iterable<V> values, Function<? super V, K> indexFunction) {
+        Map<K, V> map = new HashMap<>();
+        for (V value : values) {
+            K index = indexFunction.apply(value);
+            V prev = map.put(index, value);
+            Validate.isNull(prev, "No duplicate indices allowed but both %s and %s have the same index: %s",
```
joviegas, line 160: Query : Since we are indexing to unique Converted type should that be reflected in the function name ?

In some case in future I might even want to index duplicate value in where I would expect  Map<K, List<V>> might have difficulty to use the same function name in this class

joviegas (approved):

utils/src/main/java/software/amazon/awssdk/utils/CollectionUtils.java
Bennett-Lynch: Good point/question.

My thinking is that indexing like this would be the common/standard use case, where users expect unique indices, and if we wanted to extend this to something like multi-map support, then that new function name would be specialized to differentiate. E.g., `index` vs `indexMulti`, or similar.

However, since the current function can throw upon encounter duplicates, then perhaps there is an argument for giving this one more explicit naming. What do you think?

Bennett-Lynch: Renamed `index` to `uniqueIndex`.

PR #2724, opened 2021-09-17
Fix more sonarcloud warnings.

millems:
https://sonarcloud.io/project/security_hotspots?id=aws_aws-sdk-java-v2

### DIFF
--- a/utils/src/main/java/software/amazon/awssdk/utils/http/SdkHttpUtils.java
+++ b/utils/src/main/java/software/amazon/awssdk/utils/http/SdkHttpUtils.java
@@ -24,6 +24,7 @@ import java.io.UnsupportedEncodingException;
 import java.net.URI;
 import java.net.URLDecoder;
 import java.net.URLEncoder;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.LinkedHashMap;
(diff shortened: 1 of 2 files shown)


utils/src/main/java/software/amazon/awssdk/utils/http/SdkHttpUtils.java
```diff
@@ -335,10 +332,26 @@ public static boolean isSingleHeader(String h) {
      * Extracts query parameters from the given URI
      */
     public static Map<String, List<String>> uriParams(URI uri) {
-        return QUERY_PARAM_DELIMITER_PATTERN
-                      .splitAsStream(uri.getRawQuery().trim())
-                      .map(s -> s.contains("=") ? s.split("=", 2) : new String[] {s, null})
-                      .collect(groupingBy(a -> urlDecode(a[0]), mapping(a -> urlDecode(a[1]), toList())));
+        return splitQueryString(uri.getRawQuery())
+            .stream()
+            .map(s -> s.contains("=") ? s.split("=", 2) : new String[] {s, null})
```
Bennett-Lynch, line 337: Is `contains` necessary? Can we just uniformly call `split`?

Bennett-Lynch (approved):

utils/src/main/java/software/amazon/awssdk/utils/http/SdkHttpUtils.java
millems: Good question, I just copied+pasted that part. 

split("=", 2) will only return 1 element if there is no =, so we still need to account for that case, but it's a little more efficient to do it after the split than before. Will update the PR.

Bennett-Lynch (approved):

PR #2734, opened 2021-09-23
Move SDK-provided global execution.interceptors to a hard-coded list.

millems:
This reduces the need for customers to use a special transformer for execution.interceptors files when they are creating an uber-jar of all SDK modules. Customers should still consider using such an appender, in case other libraries rely on execution.interceptors files.


zoewangg (approved):

dagnir (approved):

dagnir:
Please run integ tests

PR #2736, opened 2021-09-24
Bug Fix to support Document Type for Explicit Payload members

joviegas:
## Motivation and Context

1. DocumentType was not handled for explicit payload member.
2. In doMarshall() the Documents were getting Marshalled in below snippet whenever the Payload type was explicit payload
3. Explicit payload member are shapes where we Explicitly define the Payload member using "payload key"
```
         if (val != null && field.containsTrait(PayloadTrait.class)) {
                    jsonGenerator.writeStartObject();
                    doMarshall((SdkPojo) val);
                    jsonGenerator.writeEndObject();
                }
```

## Description

1. Fixed by adding specific check for DocumentType and calling its corresponding marshaling Type 

## Testing

1. Added J-units for Explicit and Implicit payload
2. Note : test cases added in codegen-test because in protocol test there is an additional layer where the suite.json files gets marshalled , to make sure we test end-to-end added test in codegen-test .

Implicit and explicit Recursive payload test case added
```
{
  "MyDocument": [
    1,
    null,
    "end"
  ],
  "MapOfStringToString": {
    "key1": "value1",
    "key2": "value2"
  },
  "RecursiveStructType": {
    "NoRecurse": "noRecursive1",
    "MyDocument": {
      "docsL1": "docsStringL1"
    },
    "RecursiveStruct": {
      "NoRecurse": "noRecursive2",
      "MyDocument": {
        "docsL2": "docsStringL2"
      }
    }
  }
}
```

## Types of changes

## Checklist

## License

### DIFF
--- a/core/protocols/aws-json-protocol/src/main/java/software/amazon/awssdk/protocols/json/internal/marshall/JsonProtocolMarshaller.java
+++ b/core/protocols/aws-json-protocol/src/main/java/software/amazon/awssdk/protocols/json/internal/marshall/JsonProtocolMarshaller.java
@@ -16,6 +16,7 @@
 package software.amazon.awssdk.protocols.json.internal.marshall;
 
 import static software.amazon.awssdk.core.internal.util.Mimetype.MIMETYPE_EVENT_STREAM;
+import static software.amazon.awssdk.core.protocol.MarshallingType.DOCUMENT;
 import static software.amazon.awssdk.http.Header.CHUNKED;
 import static software.amazon.awssdk.http.Header.CONTENT_LENGTH;
 import static software.amazon.awssdk.http.Header.CONTENT_TYPE;
@@ -177,6 +178,8 @@ public class JsonProtocolMarshaller implements ProtocolMarshaller<SdkHttpFullReq
             Object val = field.getValueOrDefault(pojo);
             if (isBinary(field, val)) {
                 request.contentStreamProvider(((SdkBytes) val)::asInputStream);
+            } else if (isDocumentType(field) && val != null) {
+                marshalDocumentType(field, val);
             } else {
                 if (val != null && field.containsTrait(PayloadTrait.class)) {
                     jsonGenerator.writeStartObject();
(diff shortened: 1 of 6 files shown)


core/protocols/aws-json-protocol/src/main/java/software/amazon/awssdk/protocols/json/internal/marshall/JsonProtocolMarshaller.java
```diff
@@ -177,6 +178,12 @@ void doMarshall(SdkPojo pojo) {
             Object val = field.getValueOrDefault(pojo);
             if (isBinary(field, val)) {
                 request.contentStreamProvider(((SdkBytes) val)::asInputStream);
+            } else if (isDocumentType(field) && val != null) {
+                if (hasExplicitPayloadMember && field.containsTrait(PayloadTrait.class)) {
+                    jsonGenerator.writeStartObject();
```
cenedhryn, line 183: Why is `jsonGenerator.writeStartObject()` called for explicit payload member?

cenedhryn (commented):
Nice testing!

core/protocols/aws-json-protocol/src/main/java/software/amazon/awssdk/protocols/json/internal/marshall/JsonProtocolMarshaller.java
joviegas: In case of implicit payload writeStartObject will be done at
[StartMarshilling](https://github.com/aws/aws-sdk-java-v2/blob/a1f75b3807e548f3fab4300cf450524204a3e374/core/protocols/aws-json-protocol/src/main/java/software/amazon/awssdk/protocols/json/internal/marshall/JsonProtocolMarshaller.java#L171).

This check gets skipped because we do writeStartObject when the actual payload member is iterated [here]
(https://github.com/aws/aws-sdk-java-v2/blob/a1f75b3807e548f3fab4300cf450524204a3e374/core/protocols/aws-json-protocol/src/main/java/software/amazon/awssdk/protocols/json/internal/marshall/JsonProtocolMarshaller.java#L182)

writeStartObject on jsonGenerator is required to mark the context as Object which make the wtiteContext ready to accept Object else it always expects a Key Pair.

In DocumentType the members must always will be in payload location (as per specs ) and since writeObject gets skipped we need to perform explicit write object before writing the Object to jsonGenerator.

cenedhryn (approved):

core/protocols/aws-json-protocol/src/main/java/software/amazon/awssdk/protocols/json/internal/marshall/JsonProtocolMarshaller.java
joviegas: Thanks for offline review.
After the review I had done following modifications

1. WriteEndObject for DocumentType
2. Handle Null response
3. Handle Explicit Responses so that they are directly Unmrashalled as respective document rather than Maps.

Diff is as below

https://github.com/aws/aws-sdk-java-v2/compare/e95ab8ecd488290e31d42a2c9cd491887c9c40ef..2b94c81253a8d09854e5b7261a29c28d6af8e98a

PR #2739, opened 2021-09-24
WIP pipeline modification race fix

dagnir:
## Types of changes

## Checklist

## License


http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/HandlerRemovingChannelPool.java
```diff
@@ -55,37 +57,38 @@ public HandlerRemovingChannelPool(SdkChannelPool delegate) {
...
-                           LastHttpContentHandler.class,
-                           FlushOnReadHandler.class,
-                           ResponseHandler.class,
-                           ReadTimeoutHandler.class,
-                           WriteTimeoutHandler.class);
-        }
+            // Only remove per request handler if the channel is registered
+            // or open since DefaultChannelPipeline would remove handlers if
+            // channel is closed and unregistered
+            // See DefaultChannelPipeline.java#L1403
```
millems, line 81: I think this line number is out of date.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/NettyRequestExecutor.java
```diff
@@ -164,10 +165,12 @@ private void verifyMetricsWereCollected(CompletableFuture<Void> metricsFuture) {
     private void makeRequestListener(Future<Channel> channelFuture) {
         if (channelFuture.isSuccess()) {
             channel = channelFuture.getNow();
-            configureChannel();
-            if (tryConfigurePipeline()) {
-                makeRequest();
-            }
+            NettyUtils.doInEventLoop(channel.eventLoop(), () -> {
```
millems, line 168: I assume this is just an abundance of caution? We should always be in the event loop for the channel right now, right?

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/utils/NettyUtils.java
```diff
@@ -130,13 +130,19 @@ private NettyUtils() {
...
-            runnable.run();
-        } else {
-            eventExecutor.submit(runnable);
+            try {
+                runnable.run();
+                return eventExecutor.newSucceededFuture(null);
+            } catch (Throwable t) {
+                return eventExecutor.newFailedFuture(t);
+            }
         }
```
millems, line 144: This change is the one that scares me the most, but only because it might uncover some bugs we have elsewhere.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/utils/NettyUtils.java
```diff
@@ -130,13 +130,19 @@ private NettyUtils() {
...
+        return eventExecutor.submit(runnable);
```
millems, line 145: Do we need to handle the case that doInEventLoop can fail everywhere that we're currently using doInEventLoop?

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/NettyRequestExecutor.java
dagnir: Actually no, this is part of the fix as well. I also thought that we would be in the eventloop at this point, but currently this method run as a listener to the Promise that contains the leased channel:
https://github.com/aws/aws-sdk-java-v2/blob/bb47fade74ae919beaebb49e48279d57b76c6efe/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/NettyRequestExecutor.java#L91-L94

However, the eventloop that the promise calls listeners on isn't necessarily the same as the channel's eventloop.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/utils/NettyUtils.java
dagnir: See comment here: https://github.com/aws/aws-sdk-java-v2/pull/2739#discussion_r719582253. Would be a less invasive change

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/HandlerRemovingChannelPool.java
```diff
@@ -55,37 +57,38 @@ public HandlerRemovingChannelPool(SdkChannelPool delegate) {
...
-        removePerRequestHandlers(channel);
-        return delegate.release(channel);
+        return release(channel, new DefaultPromise<>(channel.eventLoop()));
     }
 
     @Override
     public Future<Void> release(Channel channel, Promise<Void> promise) {
-        removePerRequestHandlers(channel);
-        return delegate.release(channel, promise);
+        removePerRequestHandlers(channel).addListener(future -> delegate.release(channel, promise));
```
zoewangg, line 65: Is there any risk? IIRC, we've seen issues where listener callback never got called

dagnir: I think you're referring to previous deadlock issues? I think in those cases it's because we were doing blocking calls like `await()` and `sync()`, which is why in this case used a listener instead of blocking on completion.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/HandlerRemovingChannelPool.java
dagnir: Checked with Zoe who originally wrote this comment, and we actually don't know what part of the Netty code this is referring to. Looking back to this line in 4.1.33.Final (which was Netty's version when the comment was written) links to a line that doesn't have anything to do with removing handlers. Just going to remove this line reference for now

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/utils/NettyUtils.java
dagnir: I've changed the usages of `doInEventLoop` that return a future from within the runnable that's passed in fail the future in case future returned by `doInEventLoop` fails.

millems (approved):

zoewangg (approved):
Can we make sure our test suits (stability tests, integ tests, etc) pass if we haven't already?

dagnir:
@zoewangg 👍 will do

Issue #37, opened 2017-07-03
Transfer Manager


millems:
Review the inherited state of the V1 transfer manager and determine which changes are necessary for V2.

(Feel free to comment on this issue with desired changes).

abrooksv:
Upload and download return a Transfer which has getProgress(). This returns a simple way to get a percent complete but requires you to do busy loops to update things like UIs.

On the other hand, ProgressListener only returns the bytes transferred causing you to have to do the percentage manually and not very easy since the total size is not exposed.

Better feature parity on the ProgressListener would be nice since that plays better with async

millems:
+1 on parity with progress listener. There should probably just be one system between transfer manager and the rest of the SDK for monitoring progress.

millems:
Use DirectoryStream for loading files from a directory to avoid having to load all file names into memory.

kiiadi:
Unreasonable to expect the thread-pool will be unbounded in order to avoid deadlocks see : https://github.com/aws/aws-sdk-java/issues/939

spfink:
TransferManager features requests from v1:

https://github.com/aws/aws-sdk-java/issues/117
https://github.com/aws/aws-sdk-java/issues/284
https://github.com/aws/aws-sdk-java/issues/474
https://github.com/aws/aws-sdk-java/issues/645
https://github.com/aws/aws-sdk-java/issues/893
https://github.com/aws/aws-sdk-java/issues/964
https://github.com/aws/aws-sdk-java/issues/988
https://github.com/aws/aws-sdk-java/issues/1215
https://github.com/aws/aws-sdk-java/issues/1207
https://github.com/aws/aws-sdk-java/issues/1103

millems:
https://github.com/aws/aws-sdk-java/issues/1321

millems:
Allow using a finite number of threads for background processing. Currently, 1.11.x's `TransferManager` is reported to require an unbounded thread pool to prevent deadlocks.

edlunde-dnastar:
+1 for [aws/aws-sdk-java#1103](https://github.com/aws/aws-sdk-java/issues/1103), a request for the ability to limit bandwidth for S3 uploads/downloads. See also the recently closed issue from the aws-cli repo:
https://github.com/aws/aws-cli/issues/1090

This same feature would be similarly useful in the Java SDK to help avoid fees from ISPs for excessive bandwidth usage, or to prevent a single application from overwhelming a network's capacity.

[...]

PR #2743, opened 2021-09-27
Support upload directory in s3 transfer manager

zoewangg:
## Description

#37 
Support upload directory in s3 transfer manager
Sample code

```java
UploadDirectory uploadDirectory =
            tm.uploadDirectory(UploadDirectoryRequest.builder()
                                                     .sourceDirectory(directory)
                                                     .bucket("bucket")
                                                     .overrideConfiguration(o -> o.recursive(false).followSymbolicLinks(true))
                                                     .build());

CompletedUploadDirectory completedUploadDirectory = uploadDirectory.completionFuture().get(5, TimeUnit.SECONDS);

assertThat(completedUploadDirectory.failedUploads()).isEmpty();
assertThat(completedUploadDirectory.successfulObjects()).hasSize(2).containsOnly(completedUpload, completedUpload2);
```

Users can configure upload settings on the request `UploadDirectoryRequest#overrideConfiguration` or on the transfer manager itself. Client-level configuration takes precedence over request-level config

```java
        tm = S3TransferManager.builder()
                              .s3ClientConfiguration(b -> b.uploadDirectoryConfiguration(u -> u.followSymbolicLinks(true))
                              .build();
```

## Testing
Added functional tests against different file systems as well as integ tests.

## Types of changes

## Checklist

## License

### DIFF
--- a/.changes/2.17.59.json
+++ b/.changes/2.17.59.json
@@ -1,0 +1,42 @@
+{
+    "version": "2.17.59",
+    "date": "2021-10-13",
+    "entries": [
+        {
+            "type": "feature",
+            "category": "AWS SDK for Java v2",
+            "contributor": "",
+            "description": "Updated service endpoint metadata."
+        },
+        {
+            "type": "feature",
+            "category": "AWS Storage Gateway",
+            "contributor": "",
+            "description": "Adding support for Audit Logs on NFS shares and Force Closing Files on SMB shares."
+        },
+        {
+            "type": "feature",
+            "category": "Amazon WorkMail",
+            "contributor": "",
+            "description": "This release adds APIs for adding, removing and retrieving details of mail domains"
+        },
+        {
+            "type": "feature",
+            "category": "Amazon Kinesis Analytics",
+            "contributor": "",
+            "description": "Support for Apache Flink 1.13 in Kinesis Data Analytics. Changed the required status of some Update properties to better fit the corresponding Create properties."
+        },
+        {
+            "type": "feature",
+            "category": "AWS Config",
+            "contributor": "",
+            "description": "Adding Config support for AWS::OpenSearch::Domain"
+        },
+        {
+            "type": "feature",
+            "category": "Amazon Elastic Compute Cloud",
+            "contributor": "",
+            "description": "This release adds support for additional VPC Flow Logs delivery options to S3, such as Apache Parquet formatted files, Hourly partitions and Hive-compatible S3 prefixes"
+        }
+    ]
+}
--- a/.changes/next-release/bugfix-AWSSDKforJavav2-a6e2ed8.json
+++ b/.changes/next-release/bugfix-AWSSDKforJavav2-a6e2ed8.json
@@ -1,0 +1,6 @@
+{
+    "category": "AWS SDK for Java v2", 
+    "contributor": "", 
+    "type": "bugfix", 
+    "description": "Fixes a bug in XML error unmarshalling where error responses with empty body won't populate the requestId field. Affects Amazon S3 API calls such as Head object"
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,28 @@
+# __2.17.59__ __2021-10-13__
+## __AWS Config__
+  - ### Features
+    - Adding Config support for AWS::OpenSearch::Domain
+
+## __AWS SDK for Java v2__
+  - ### Features
+    - Updated service endpoint metadata.
+
+## __AWS Storage Gateway__
+  - ### Features
+    - Adding support for Audit Logs on NFS shares and Force Closing Files on SMB shares.
+
+## __Amazon Elastic Compute Cloud__
+  - ### Features
+    - This release adds support for additional VPC Flow Logs delivery options to S3, such as Apache Parquet formatted files, Hourly partitions and Hive-compatible S3 prefixes
+
+## __Amazon Kinesis Analytics__
+  - ### Features
+    - Support for Apache Flink 1.13 in Kinesis Data Analytics. Changed the required status of some Update properties to better fit the corresponding Create properties.
+
+## __Amazon WorkMail__
+  - ### Features
+    - This release adds APIs for adding, removing and retrieving details of mail domains
+
 # __2.17.58__ __2021-10-12__
 ## __AWS MediaTailor__
   - ### Features
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ To automatically manage module versions (currently all modules have the same ver
     <dependency>
       <groupId>software.amazon.awssdk</groupId>
       <artifactId>bom</artifactId>
-      <version>2.17.58</version>
+      <version>2.17.59</version>
       <type>pom</type>
       <scope>import</scope>
     </dependency>
@@ -86,12 +86,12 @@ Alternatively you can add dependencies for the specific services you use only:
 <dependency>
   <groupId>software.amazon.awssdk</groupId>
   <artifactId>ec2</artifactId>
-  <version>2.17.58</version>
+  <version>2.17.59</version>
 </dependency>
 <dependency>
   <groupId>software.amazon.awssdk</groupId>
   <artifactId>s3</artifactId>
-  <version>2.17.58</version>
+  <version>2.17.59</version>
 </dependency>
 ```
 
@@ -103,7 +103,7 @@ You can import the whole SDK into your project (includes *ALL* services). Please
 <dependency>
   <groupId>software.amazon.awssdk</groupId>
   <artifactId>aws-sdk-java</artifactId>
-  <version>2.17.58</version>
+  <version>2.17.59</version>
 </dependency>
 ```
 
--- a/archetypes/archetype-app-quickstart/pom.xml
+++ b/archetypes/archetype-app-quickstart/pom.xml
@@ -20,7 +20,7 @@
     <parent>
         <artifactId>archetypes</artifactId>
         <groupId>software.amazon.awssdk</groupId>
-        <version>2.17.59-SNAPSHOT</version>
+        <version>2.17.60-SNAPSHOT</version>
     </parent>
     <modelVersion>4.0.0</modelVersion>
 
--- a/archetypes/archetype-lambda/pom.xml
+++ b/archetypes/archetype-lambda/pom.xml
@@ -20,7 +20,7 @@
     <parent>
         <artifactId>archetypes</artifactId>
         <groupId>software.amazon.awssdk</groupId>
-        <version>2.17.59-SNAPSHOT</version>
+        <version>2.17.60-SNAPSHOT</version>
     </parent>
     <modelVersion>4.0.0</modelVersion>
     <artifactId>archetype-lambda</artifactId>
--- a/archetypes/archetype-tools/pom.xml
+++ b/archetypes/archetype-tools/pom.xml
@@ -20,7 +20,7 @@
     <parent>
         <artifactId>archetypes</artifactId>
         <groupId>software.amazon.awssdk</groupId>
-        <version>2.17.59-SNAPSHOT</version>
+        <version>2.17.60-SNAPSHOT</version>
     </parent>
     <modelVersion>4.0.0</modelVersion>
 
--- a/archetypes/pom.xml
+++ b/archetypes/pom.xml
@@ -20,7 +20,7 @@
     <parent>
         <artifactId>aws-sdk-java-pom</artifactId>
         <groupId>software.amazon.awssdk</groupId>
-        <version>2.17.59-SNAPSHOT</version>
+        <version>2.17.60-SNAPSHOT</version>
     </parent>
     <modelVersion>4.0.0</modelVersion>
     <artifactId>archetypes</artifactId>
--- a/aws-sdk-java/pom.xml
+++ b/aws-sdk-java/pom.xml
@@ -17,7 +17,7 @@
     <parent>
         <groupId>software.amazon.awssdk</groupId>
         <artifactId>aws-sdk-java-pom</artifactId>
-        <version>2.17.59-SNAPSHOT</version>
+        <version>2.17.60-SNAPSHOT</version>
         <relativePath>../pom.xml</relativePath>
     </parent>
     <artifactId>aws-sdk-java</artifactId>
--- a/bom-internal/pom.xml
+++ b/bom-internal/pom.xml
@@ -20,7 +20,7 @@
     <parent>
         <artifactId>aws-sdk-java-pom</artifactId>
         <groupId>software.amazon.awssdk</groupId>
-        <version>2.17.59-SNAPSHOT</version>
+        <version>2.17.60-SNAPSHOT</version>
     </parent>
     <modelVersion>4.0.0</modelVersion>
 
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -17,7 +17,7 @@
     <parent>
         <groupId>software.amazon.awssdk</groupId>
         <artifactId>aws-sdk-java-pom</artifactId>
-        <version>2.17.59-SNAPSHOT</version>
+        <version>2.17.60-SNAPSHOT</version>
         <relativePath>../pom.xml</relativePath>
     </parent>
     <artifactId>bom</artifactId>
--- a/bundle/pom.xml
+++ b/bundle/pom.xml
@@ -21,7 +21,7 @@
     <parent>
         <groupId>software.amazon.awssdk</groupId>
         <artifactId>aws-sdk-java-pom</artifactId>
-        <version>2.17.59-SNAPSHOT</version>
+        <version>2.17.60-SNAPSHOT</version>
     </parent>
     <artifactId>bundle</artifactId>
     <packaging>jar</packaging>
--- a/codegen-lite-maven-plugin/pom.xml
+++ b/codegen-lite-maven-plugin/pom.xml
@@ -22,7 +22,7 @@
     <parent>
         <groupId>software.amazon.awssdk</groupId>
         <artifactId>aws-sdk-java-pom</artifactId>
-        <version>2.17.59-SNAPSHOT</version>
+        <version>2.17.60-SNAPSHOT</version>
         <relativePath>../pom.xml</relativePath>
     </parent>
     <artifactId>codegen-lite-maven-plugin</artifactId>
--- a/codegen-lite/pom.xml
+++ b/codegen-lite/pom.xml
@@ -21,7 +21,7 @@
     <parent>
         <groupId>software.amazon.awssdk</groupId>
         <artifactId>aws-sdk-java-pom</artifactId>
-        <version>2.17.59-SNAPSHOT</version>
+        <version>2.17.60-SNAPSHOT</version>
     </parent>
     <artifactId>codegen-lite</artifactId>
     <name>AWS Java SDK :: Code Generator Lite</name>
--- a/codegen-maven-plugin/pom.xml
+++ b/codegen-maven-plugin/pom.xml
@@ -22,7 +22,7 @@
     <parent>
         <groupId>software.amazon.awssdk</groupId>
         <artifactId>aws-sdk-java-pom</artifactId>
-        <version>2.17.59-SNAPSHOT</version>
+        <version>2.17.60-SNAPSHOT</version>
         <relativePath>../pom.xml</relativePath>
     </parent>
     <artifactId>codegen-maven-plugin</artifactId>
--- a/codegen/pom.xml
+++ b/codegen/pom.xml
@@ -21,7 +21,7 @@
     <parent>
         <groupId>software.amazon.awssdk</groupId>
         <artifactId>aws-sdk-java-pom</artifactId>
-        <version>2.17.59-SNAPSHOT</version>
+        <version>2.17.60-SNAPSHOT</version>
     </parent>
     <artifactId>codegen</artifactId>
     <name>AWS Java SDK :: Code Generator</name>
--- a/core/annotations/pom.xml
+++ b/core/annotations/pom.xml
@@ -20,7 +20,7 @@
     <parent>
         <artifactId>core</artifactId>
         <groupId>software.amazon.awssdk</groupId>
-        <version>2.17.59-SNAPSHOT</version>
+        <version>2.17.60-SNAPSHOT</version>
     </parent>
     <modelVersion>4.0.0</modelVersion>
 
--- a/core/arns/pom.xml
+++ b/core/arns/pom.xml
@@ -20,7 +20,7 @@
     <parent>
         <artifactId>core</artifactId>
         <groupId>software.amazon.awssdk</groupId>
-        <version>2.17.59-SNAPSHOT</version>
+        <version>2.17.60-SNAPSHOT</version>
     </parent>
     <modelVersion>4.0.0</modelVersion>
 
--- a/core/auth-crt/pom.xml
+++ b/core/auth-crt/pom.xml
@@ -22,7 +22,7 @@
     <parent>
         <groupId>software.amazon.awssdk</groupId>
         <artifactId>core</artifactId>
-        <version>2.17.59-SNAPSHOT</version>
+        <version>2.17.60-SNAPSHOT</version>
     </parent>
 
     <artifactId>auth-crt</artifactId>
--- a/core/auth/pom.xml
+++ b/core/auth/pom.xml
@@ -22,7 +22,7 @@
     <parent>
         <groupId>software.amazon.awssdk</groupId>
         <artifactId>core</artifactId>
-        <version>2.17.59-SNAPSHOT</version>
+        <version>2.17.60-SNAPSHOT</version>
     </parent>
 
     <artifactId>auth</artifactId>
--- a/core/aws-core/pom.xml
+++ b/core/aws-core/pom.xml
@@ -22,7 +22,7 @@
     <parent>
         <groupId>software.amazon.awssdk</groupId>
         <artifactId>core</artifactId>
-        <version>2.17.59-SNAPSHOT</version>
+        <version>2.17.60-SNAPSHOT</version>
     </parent>
 
     <artifactId>aws-core</artifactId>
--- a/core/json-utils/pom.xml
+++ b/core/json-utils/pom.xml
@@ -20,7 +20,7 @@
     <parent>
         <artifactId>core</artifactId>
         <groupId>software.amazon.awssdk</groupId>
-        <version>2.17.59-SNAPSHOT</version>
+        <version>2.17.60-SNAPSHOT</version>
     </parent>
     <modelVersion>4.0.0</modelVersion>
 
--- a/core/metrics-spi/pom.xml
+++ b/core/metrics-spi/pom.xml
@@ -5,7 +5,7 @@
     <parent>
         <artifactId>core</artifactId>
         <groupId>software.amazon.awssdk</groupId>
-        <version>2.17.59-SNAPSHOT</version>
+        <version>2.17.60-SNAPSHOT</version>
     </parent>
     <modelVersion>4.0.0</modelVersion>
 
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -21,7 +21,7 @@
     <parent>
         <artifactId>aws-sdk-java-pom</artifactId>
         <groupId>software.amazon.awssdk</groupId>
-        <version>2.17.59-SNAPSHOT</version>
+        <version>2.17.60-SNAPSHOT</version>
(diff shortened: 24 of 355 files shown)


services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/UploadDirectoryManager.java
```diff
@@ -0,0 +1,166 @@
...
+    private static UploadRequest constructUploadRequest(UploadDirectoryRequest uploadDirectoryRequest, int directoryNameCount,
+                                                        Path path) {
+        String prefix = processPrefix(uploadDirectoryRequest.prefix());
+        String relativePathName = getRelativePathName(directoryNameCount, path);
+        String key = prefix + relativePathName;
+
+        PutObjectRequest putObjectRequest = PutObjectRequest.builder()
+                                                            .bucket(uploadDirectoryRequest.bucket())
+                                                            .key(key)
+                                                            .build();
```
zoewangg, line 160: Currently, we are just passing bucket and key and I'm thinking about introducing a transfer extension (similar to the core execution interceptor) for this to allow users to provide modify other parameters in `PutObjectRequest`.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/CompletedUploadDirectory.java
```diff
@@ -0,0 +1,40 @@
...
+     * A list of failed single file uploads associated with one upload directory transfer
+     * @return A list of failed uploads
+     */
+    List<FailedUpload> failedUploads();
+
+    /**
+     * A list of successful single file uploads associated with one upload directory transfer
+     * @return a list of successful  uploads
+     */
+    List<CompletedUpload> successfulObjects();
```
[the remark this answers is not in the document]
millems: Is there some risk in using the same object (CompletedUpload) for successful upload calls and successful uploadDirectory calls? E.g. uploadDirectory will almost certainly always be from a file, but upload might be from a publisher in the future. Not sure if that matters much, since if we want Path on CompletedUpload we could always just return null if it's from a publisher, but worth considering...

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/UploadDirectoryConfiguration.java
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    public interface Builder extends CopyableBuilder<Builder, UploadDirectoryConfiguration> {
+
+        /**
+         * Specify whether to recursively upload all files under the specified directory
+         *
+         * @param recursive whether enable recursive upload
+         * @return This builder for method chaining.
+         */
```
millems, line 121: "Default to true".

Are there any edge-cases we should document? E.g. "Links are excluded/not followed... see followSymbolicLinks", "The current user must have read/execute access to all directories and read access to all files", "hidden files are included/excluded", "wildcards like * / ** are treated literally and not expanded" (I guess this one is better on the directory parameter, but you get the point), etc.

I really think we should raise the bar on our javadoc in our high level libraries to try to anticipate most of what our customer might encounter. Us documenting something once can save hundreds of people from needing to figure it out on their own.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/DefaultCompletedUploadDirectory.java
```diff
@@ -0,0 +1,116 @@
...
+        return failedUploads;
+    }
+
+    @Override
+    public List<CompletedUpload> successfulObjects() {
+        return completedUploads;
+    }
+
+    /**
+     * Creates a default builder for {@link CompletedUpload}.
```
millems, line 48: CompletedUploadDirectory?

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/DefaultCompletedUploadDirectory.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    interface Builder {
+
+        Builder successfulUploads(List<CompletedUpload> successfulUploads);
+
+        Builder failedUploads(List<FailedUpload> failedUploads);
+
+        CompletedUploadDirectory build();
+    }
```
millems, line 93: Shouldn't this builder be in the parent interface? Or are we not allowing customers access to a builder for their own testing?

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/DefaultS3TransferManager.java
```diff
@@ -30,54 +32,93 @@
...
         }
 
         s3CrtAsyncClient = clientBuilder.build();
+        transferConfiguration = transferConfigBuilder.build();
+        uploadDirectoryManager = new UploadDirectoryManager(transferConfiguration, this::upload);
     }
 
     @SdkTestInternalApi
-    DefaultS3TransferManager(S3CrtAsyncClient s3CrtAsyncClient) {
+    DefaultS3TransferManager(S3CrtAsyncClient s3CrtAsyncClient, UploadDirectoryManager uploadDirectoryManager) {
```
millems, line 66: Why does this need to be exposed for testing purposes? Can't we get to this via the builder?

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/DefaultS3TransferManager.java
```diff
@@ -30,54 +32,93 @@
...
+        this.transferConfiguration = TransferConfiguration.builder().build();
+        this.uploadDirectoryManager = uploadDirectoryManager;
     }
 
     @Override
     public Upload upload(UploadRequest uploadRequest) {
-        PutObjectRequest putObjectRequest = uploadRequest.putObjectRequest();
-        AsyncRequestBody requestBody = requestBodyFor(uploadRequest);
+        try {
+            assertNotObjectLambdaArn(uploadRequest.putObjectRequest().bucket(), "upload");
```
millems, line 75: NPE risk on uploadRequest. Same below.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/DefaultS3TransferManager.java
```diff
@@ -30,54 +32,93 @@
...
-                                       AsyncResponseTransformer.toFile(downloadRequest.destination()));
-        CompletableFuture<CompletedDownload> future =
-            getObjectFuture.thenApply(r -> DefaultCompletedDownload.builder().response(r).build());
+    public UploadDirectory uploadDirectory(UploadDirectoryRequest uploadDirectoryRequest) {
+        try {
+            Path directory = uploadDirectoryRequest.sourceDirectory();
+
+            if (!Files.exists(directory) || Files.isRegularFile(directory)) {
+                throw new IllegalArgumentException("The source directory provided either does not exist or is not a directory");
+            }
```
millems, line 98: Can we throw different exception messages for these two cases?

Nit: Validate.isTrue for code golf.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/UploadDirectoryManager.java
```diff
@@ -0,0 +1,166 @@
...
+import software.amazon.awssdk.transfer.s3.UploadDirectory;
+import software.amazon.awssdk.transfer.s3.UploadDirectoryRequest;
+import software.amazon.awssdk.transfer.s3.UploadRequest;
+import software.amazon.awssdk.utils.CompletableFutureUtils;
+import software.amazon.awssdk.utils.Logger;
+import software.amazon.awssdk.utils.StringUtils;
+
+@SdkInternalApi
+public class UploadDirectoryManager {
+    private static final Logger log = Logger.loggerFor(S3TransferManager.class);
```
millems, line 47: +1000, we should be using more shared logging instead of exposing internal classes, but was this intentional?

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/UploadDirectoryManager.java
```diff
@@ -0,0 +1,166 @@
...
+        try (Stream<Path> entries = listFiles(directory, uploadDirectoryRequest)) {
+            uploadFutures =
+                entries.map(path -> {
+                    CompletableFuture<CompletedUpload> future =
+                        uploadSingleFile(uploadDirectoryRequest, uploads, failedUploads, path);
+                    // Forward cancellation of the return future to all individual futures.
+                    CompletableFutureUtils.forwardExceptionTo(returnFuture, future);
+                    return future;
+                }).collect(Collectors.toList());
+        }
```
millems, line 76: If I do transferManager.uploadDirectory("/") wouldn't this blow up our memory? Should we limit the number of concurrent uploads (at least, to the number of connections that we have)?

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/UploadDirectoryManager.java
```diff
@@ -0,0 +1,166 @@
...
+        CompletableFuture<CompletedUpload> future = uploadFunction.apply(uploadRequest)
+                                                                  .completionFuture();
+        future.whenComplete((r, t) -> {
+            if (t != null) {
+                failedUploads.add(DefaultFailedUpload.builder()
+                                                     .exception(t)
+                                                     .path(path)
+                                                     .build());
+            } else {
+                uploads.add(r);
```
millems, line 102: Do we have some serialization mechanism that i'm missing, or is failedUploads.add and uploads.add a concurrency risk?

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/UploadDirectoryManager.java
```diff
@@ -0,0 +1,166 @@
...
+        this.uploadFunction = uploadFunction;
+    }
+
+    public UploadDirectory uploadDirectory(UploadDirectoryRequest uploadDirectoryRequest) {
+        CompletableFuture<CompletedUploadDirectory> returnFuture = new CompletableFuture<>();
+        Path directory = uploadDirectoryRequest.sourceDirectory();
+
+        List<CompletedUpload> uploads = new ArrayList<>();
+        List<FailedUpload> failedUploads = new ArrayList<>();
+        List<CompletableFuture<CompletedUpload>> uploadFutures;
```
millems, line 65: If we get a feature request in the future where someone says "I'm uploading millions of files and I'm running out of memory", is that something we can accommodate? E.g. by adding a "don't record file-level results" flag? I don't see why not, but you're closer to this than me.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/UploadDirectoryManager.java
```diff
@@ -0,0 +1,166 @@
...
+
+        try (Stream<Path> entries = listFiles(directory, uploadDirectoryRequest)) {
+            uploadFutures =
+                entries.map(path -> {
+                    CompletableFuture<CompletedUpload> future =
+                        uploadSingleFile(uploadDirectoryRequest, uploads, failedUploads, path);
+                    // Forward cancellation of the return future to all individual futures.
+                    CompletableFutureUtils.forwardExceptionTo(returnFuture, future);
+                    return future;
+                }).collect(Collectors.toList());
```
millems, line 75: Is there any way to do a non-blocking directory listing? For large directories, this is potentially high-impact on a customer's event loop. If not, maybe we should offload this to another executor.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/UploadDirectoryManager.java
```diff
@@ -0,0 +1,166 @@
...
+        }
+        return prefix.endsWith("/") ? prefix : prefix + "/";
+    }
+
+    private static String getRelativePathName(int directoryNameCount, Path path) {
+        String relativePathName = path.subpath(directoryNameCount,
+                                               path.getNameCount()).toString();
+
+        // Replace "\" (Windows FS) with "/"
+        return relativePathName.replace('\\', '/');
```
millems, line 148: Should we rely on FileSystems.getDefault().getSeparator()? I feel like a file system with \ as a valid file/directory name character will have trouble with this line.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/UploadDirectoryManager.java
```diff
@@ -0,0 +1,166 @@
...
+        } catch (IOException e) {
+            throw SdkClientException.create("Failed to list files under the provided directory", e);
+        }
+    }
+
+    private static String processPrefix(String prefix) {
+        if (StringUtils.isEmpty(prefix)) {
+            return "";
+        }
+        return prefix.endsWith("/") ? prefix : prefix + "/";
```
millems, line 140: S3 delimiters are configurable: https://docs.aws.amazon.com/AmazonS3/latest/userguide/using-prefixes.html

Should we support that here, instead of hard-coding the delimiter as /?

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/UploadDirectoryManager.java
millems: They could do this with an ExecutionInterceptor on the underlying S3 client, once we support configuring the underlying S3 client, right? 

I suppose a `putObjectRequestSupplier(Supplier<PutObjectRequest>)` on the UploadDirectoryRequest would be more friendly, especially if they want request-level parameter configuration. +1

zoewangg: > They could do this with an ExecutionInterceptor on the underlying S3 client, once we support configuring the underlying S3 client, right?

Yes, that's right.

> I suppose a putObjectRequestSupplier(Supplier<PutObjectRequest>) on the UploadDirectoryRequest would be more friendly, especially if they want request-level parameter configuration. +1

+1. It would be much easier than what I proposed. I suppose we can use `Consumer<PutObjectRequest.Builder>` to make it clear that this is a modification.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/CompletedUploadDirectory.java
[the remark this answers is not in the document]
zoewangg: Good call out. I think separating them is more future proof

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/UploadDirectoryConfiguration.java
zoewangg: +1 to raising the bar on our Javadoc. I got overwhelmed when writing boilerplate code(builders are so verbose!), so the Javadoc is a bit sloppy 😅. Will fix it.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/UploadDirectoryConfiguration.java
```diff
@@ -0,0 +1,183 @@
...
+
+/**
+ * Configuration options for {@link S3TransferManager#uploadDirectory}. All values are optional, and not specifying them will
+ * use the SDK default values.
+ *
+ * <p>Use {@link #builder()} to create a set of options.</p>
+ */
+@SdkPublicApi
+@SdkPreviewApi
+public final class UploadDirectoryConfiguration implements ToCopyableBuilder<UploadDirectoryConfiguration.Builder,
```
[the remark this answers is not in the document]
zoewangg: We don't have any transfer-manager-level retries for now. We can definitely add it in the future, something like automatically retrying failed files.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/UploadDirectoryConfiguration.java
```diff
@@ -0,0 +1,183 @@
...
+    public UploadDirectoryConfiguration(DefaultBuilder builder) {
+        this.followSymbolicLinks = builder.followSymbolicLinks;
+        this.maxDepth = builder.maxDepth;
+        this.recursive = builder.recursive;
+    }
+
+    /**
+     * @return whether to follow symbolic links
+     */
+    public Optional<Boolean> followSymbolicLinks() {
```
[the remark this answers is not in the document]
zoewangg: >  because it's possible that the caller knows whether or not they're present:

The caller doesn't know though and would have to do null check explicitly otherwise.

https://github.com/aws/aws-sdk-java-v2/blob/c47d69e5f96885cb7f5efa868950a94a5ed73b5a/services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/TransferConfiguration.java#L42-L45

> We're inconsistent already, so we should at least make a choice.

Hmm, I think ClientOverrideConfiguartion and RequestOverrideConfiguration are consistent though; all getters are returning optional values if the values are nullable. Did I miss anything?

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/DefaultCompletedUploadDirectory.java
zoewangg: I wonder if we need the parent interface `CompletedUploadDirectory` at all; we don't use interface for SDK service POJOs. WDYT?

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/UploadDirectoryManager.java
Bennett-Lynch: I agree that offering a way to hook into the resulting `PutObjectRequest` is crucial.

There are a few other use cases I can think of:
1. I want to be able to supply my own `Path`-filters. For example, I may want to ensure I exclude sensitive file extensions. E.g., a `Predicate<Path>`.
1. I want to be able to customize a `key` name. For example, I may want to force everything to lowercase. E.g., a configurable `Function<Path,String>`, or a `UnaryOperator<String>`.
1. I want to be able to transform file contents before they are uploaded. For example, I may want to compress everything before uploading.

Not all of these are needed now, but it would be good to make sure we have some type of idea of the path forward here.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/UploadDirectoryConfiguration.java
```diff
@@ -0,0 +1,183 @@
...
+ * use the SDK default values.
+ *
+ * <p>Use {@link #builder()} to create a set of options.</p>
+ */
+@SdkPublicApi
+@SdkPreviewApi
+public final class UploadDirectoryConfiguration implements ToCopyableBuilder<UploadDirectoryConfiguration.Builder,
+    UploadDirectoryConfiguration> {
+
+    private final Boolean followSymbolicLinks;
```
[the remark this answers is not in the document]
zoewangg: > I worry the word "follow" sounds specific to directories and not files.

I think `follow` is what used in the Javadocs  https://docs.oracle.com/javase/8/docs/api/java/nio/file/LinkOption.html#NOFOLLOW_LINKS, https://docs.oracle.com/javase/8/docs/api/java/nio/file/FileVisitOption.html#FOLLOW_LINKS

Do you have any suggestions?

> On the same note, do you think there are any users that would want to follow directory-sym-links but NOT upload file-sym-links?

I think to allow maximum flexibility, we can support uploading stream of paths `Stream<Path>` directly in the future. That way, customers can filter the files or configure the traversal by themselves before passing them to the SDK.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/UploadDirectoryConfiguration.java
```diff
@@ -0,0 +1,183 @@
...
+         * @return This builder for method chaining.
+         */
+        Builder followSymbolicLinks(Boolean followSymbolicLinks);
+
+        /**
+         * Specify the maximum number of directory levels to traverse
+         * @param maxDepth the maximum number of directory levels
+         * @return This builder for method chaining.
+         */
+        Builder maxDepth(Integer maxDepth);
```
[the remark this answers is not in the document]
zoewangg: `maxDepth` means we include only the provided directory. I'll clarify this in the Javadoc

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/UploadDirectoryManager.java
zoewangg: Yeah, we have a unit test to verify simulated Windows file system. I can start the window integ test codeBuild job

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/UploadDirectoryManager.java
zoewangg: Nice summary! I'll create a backlog item for this specific feature because it seems out of scope of this PR, and is additive.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/UploadDirectory.java
```diff
@@ -0,0 +1,29 @@
...
+
+import java.util.concurrent.CompletableFuture;
+import software.amazon.awssdk.annotations.SdkPublicApi;
+
+/**
+ * An upload transfer of an directory to S3.
+ */
+@SdkPublicApi
+public interface UploadDirectory extends Transfer {
+
```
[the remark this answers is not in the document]
zoewangg: > Recommend UploadDirectoryTransfer

+1

> Could UploadDirectoryTransfer have a reference to the original UploadDirectoryRequest?

Is there any benefit? We could, but we don't currently keep the request in the response classes in other places.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/TransferConfigurationOption.java
```diff
@@ -0,0 +1,67 @@
...
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.transfer.s3.internal;
+
+import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.utils.AttributeMap;
+
+@SdkInternalApi
+public final class TransferConfigurationOption<T> extends AttributeMap.Key<T> {
```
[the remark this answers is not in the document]
millems: Imo, Javadoc should be on all classes. If it's internal classes the audience is for the developers instead of the customers, but it's still important for helping guide future developers. 

Questions we could answer in the javadoc here: Are all options we add supposed to be in this? Where is this used? Where should it not be used?

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/DefaultCompletedUploadDirectory.java
zoewangg: To close the loop, I removed the interface for the POJO classes.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/UploadDirectoryManager.java
```diff
@@ -0,0 +1,166 @@
...
+                return Files.list(directory).filter(Files::isRegularFile);
+            }
+
+            boolean followSymbolicLinks = transferConfiguration.resolveUploadDirectoryFollowSymbolicLinks(request);
+
+            int maxDepth = transferConfiguration.resolveUploadDirectoryMaxDepth(request);
+
+            if (followSymbolicLinks) {
+                return Files.walk(directory, maxDepth, FileVisitOption.FOLLOW_LINKS)
+                            .peek(p -> log.trace(() -> "Processing path: " + p))
```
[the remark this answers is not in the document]
zoewangg: Yeah, added a new logging statement to log request as well as path and removed this log statement

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/UploadDirectoryRequest.java
```diff
@@ -0,0 +1,223 @@
...
+    public String prefix() {
+        return prefix;
+    }
+
+    /**
+     * @return the optional override configuration
+     */
+    public Optional<UploadDirectoryConfiguration> overrideConfiguration() {
+        return Optional.ofNullable(overrideConfiguration);
+    }
```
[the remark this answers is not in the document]
zoewangg: We are returning Optional configuration in existing classes. https://github.com/aws/aws-sdk-java-v2/blob/master/core/sdk-core/src/main/java/software/amazon/awssdk/core/SdkRequest.java#L35

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/UploadDirectoryHelper.java
```diff
@@ -0,0 +1,208 @@
...
+
+                // Forward cancellation of the return future to all individual futures.
+                CompletableFutureUtils.forwardExceptionTo(returnFuture, future);
+            });
+        }
+
+        // Wait for all sub transfers to complete
+        phaser.arriveAndAwaitAdvance();
+        phaser.arriveAndDeregister();
+        returnFuture.complete(CompletedUploadDirectory.builder().failedUploads(failedUploads).build());
```
millems, line 110: Instead of using a phaser and blocking this thread, should this thread just be used for delegation, and then we can return `CompletableFuture.allOf(uploadSingleFileFutures)`?

That'll make this thread's work much shorter-lived, and make this executor less likely to be a bottleneck.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/DefaultS3TransferManager.java
```diff
@@ -29,79 +34,172 @@
...
+            Validate.isTrue(Files.exists(directory), "The source directory (%s) provided does not exist", directory);
+            boolean followSymbolicLinks = transferConfiguration.resolveUploadDirectoryFollowSymbolicLinks(uploadDirectoryRequest);
+            if (followSymbolicLinks) {
+                Validate.isFalse(Files.isRegularFile(directory), "The source directory (%s) provided is not a "
+                                                                 + "directory", directory);
+            } else {
+                Validate.isFalse(Files.isRegularFile(directory, LinkOption.NOFOLLOW_LINKS), "The source directory (%s) provided"
+                                                                                            + " is not a "
+                                                                                            + "directory", directory);
+            }
```
millems, line 126: Because it's easy to do, should we do these file system checks in uploadDIrectoryManager.uploadDirectory, so that they can be done on the executor instead of the calling thread? It's probably negligible performance-wise for most applications, but something like an NFS could block the calling thread for a more noticeable amount of time.

PR #2749, opened 2021-10-01
Update optional guideline

zoewangg:
## Motivation and Context
Update optional guideline

### DIFF
--- a/docs/design/UseOfOptional.md
+++ b/docs/design/UseOfOptional.md
@@ -6,12 +6,11 @@ This page describes a general guideline for the use of
 `java.util.Optional` in the AWS SDK for Java 2.x.
 
 - `Optional` SHOULD be used when it isn't obvious to a caller whether a
-  result will be null.
-- `Optional` MUST not be used when the result will never be
-  `Optional.empty()`.
-- `Optional` MUST not be used for member variables.
-- `Optional` MUST not be used for method parameters.
-- `Optional` MUST not be used in generated service model classes with "setters" and "getters", like Builders or POJOs.
+  result will be null, e.g., 
+- `Optional` MUST not be used when the result will never be null.
+- `Optional` SHOULD not be used for member variables, e.g., `private final Optional<String> field;`
+- `Optional` MUST not be used for method parameters, e.g., `private void test(Optional<String> value)`
+- `Optional` MUST not be used as a return type for "getters" in generated service model classes such as service Builders or POJOs.
 
 References:
 
--- a/docs/design/UseOfOptional.md
+++ b/docs/design/UseOfOptional.md
@@ -2,19 +2,16 @@
 
 ## Use of Optional
 
-This page describes general guidelines of how we use
-`java.util.Optional`.
+This page describes a general guideline for the use of
+`java.util.Optional` in the AWS SDK for Java 2.x.
 
-- `Optional` should be used when it isn't obvious to a caller whether a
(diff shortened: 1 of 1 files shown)


docs/design/UseOfOptional.md
```diff
@@ -2,19 +2,15 @@
...
-  `Optional.empty()`.
-- `Optional` must not be used for instance (in `@SdkPublicApi`s).
-- `Optional` must not be used for parameters (in `@SdkPublicApi`s).
-- `Optional` must not be used when the caller may know whether a value
-  is present. As a consequence:
-  - `Optional` must not be used in classes with "setters" and "getters",
-    like Builders or POJOs.
+- `Optional` SHOULD be used when it isn't obvious to a caller whether a
+  result will be null, e.g., 
+- `Optional` MUST not be used when the result will never be null.
```
cenedhryn, line 10: When the statement contains `not`, I think it should also be caps, so `MUST not` => `MUST NOT`

docs/design/UseOfOptional.md
```diff
@@ -2,19 +2,15 @@
...
-- `Optional` must not be used when the result will never be
-  `Optional.empty()`.
-- `Optional` must not be used for instance (in `@SdkPublicApi`s).
-- `Optional` must not be used for parameters (in `@SdkPublicApi`s).
-- `Optional` must not be used when the caller may know whether a value
-  is present. As a consequence:
-  - `Optional` must not be used in classes with "setters" and "getters",
-    like Builders or POJOs.
+- `Optional` SHOULD be used when it isn't obvious to a caller whether a
+  result will be null, e.g.,
```
cenedhryn, line 9: Missing the e.g.

docs/design/UseOfOptional.md
```diff
@@ -2,19 +2,15 @@
...
-- `Optional` must not be used when the caller may know whether a value
-  is present. As a consequence:
-  - `Optional` must not be used in classes with "setters" and "getters",
-    like Builders or POJOs.
+- `Optional` SHOULD be used when it is not obvious to a caller whether a
+  result will be null, e.g, `public <T> Optional<T> getValueForField(String fieldName, Class<T> clazz)`in [SdkResponse.java](https://github.com/aws/aws-sdk-java-v2/blob/aa161c564c580ced4a0381d3ed7d4d13120916fc/core/sdk-core/src/main/java/software/amazon/awssdk/core/SdkResponse.java#L59-L61)
+- `Optional` MUST NOT be used when the result will never be null.
+- `Optional` SHOULD NOT be used for member variables, e.g., `private final Optional<String> field;`
+- `Optional` MUST NOT be used for method parameters, e.g., `private void test(Optional<String> value)`
+- `Optional` MUST NOT be used as a return type for "getters" in generated service model classes such as service Builders or POJOs.
```
Bennett-Lynch, line 13: I wonder if we should separate these rules into categories:
1. Return types
2. Method parameters
3. Member variables (or just _other_)

Currently, your proposed \1 and \2 seem to refer to return types, but it doesn't explicitly state that. And then we jump back and forth a bit before returning to return types.

cenedhryn (approved):

docs/design/UseOfOptional.md
zoewangg: Good point, will clarify

Bennett-Lynch (approved):

PR #2799, opened 2021-10-26
Added UploadDirectoryRequest.Builder#uploadRequestTransformer that al…

millems:
…lows modifying the UploadRequests generated during directory uploads.


services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/UploadDirectoryRequest.java
```diff
@@ -41,13 +42,15 @@
...
     private final String delimiter;
+    private final UploadRequestTransformer uploadRequestTransformer;
 
     public UploadDirectoryRequest(DefaultBuilder builder) {
         this.sourceDirectory = Validate.paramNotNull(builder.sourceDirectory, "sourceDirectory");
         this.bucket = Validate.paramNotNull(builder.bucket, "bucket");
         this.prefix = builder.prefix;
         this.overrideConfiguration = builder.configuration;
         this.delimiter = builder.delimiter;
+        this.uploadRequestTransformer = builder.uploadRequestTransformer;
```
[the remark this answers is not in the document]
millems: Moved it to the override config.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/UploadRequestTransformer.java
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * A transformer for modifying {@link UploadRequest}s. This is commonly used as a functional interface in an upload request via
+ * {@link UploadDirectoryRequest.Builder#uploadRequestTransformer(UploadRequestTransformer)}.
+ */
+@FunctionalInterface
+@ThreadSafe
+@SdkPublicApi
+@SdkPreviewApi
+public interface UploadRequestTransformer {
```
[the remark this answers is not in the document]
millems: Changed to `Consumer<UploadFileRequest.Builder>`

services-custom/s3-transfer-manager/src/it/java/software/amazon/awssdk/transfer/s3/S3TransferManagerUploadDirectoryIntegrationTest.java
```diff
@@ -124,6 +128,23 @@ public void uploadDirectory_withDelimiter_filesSentCorrectly() {
...
     }
 
+    @Test
+    public void uploadDirectory_withRequestTransformer_usesRequestTransformer() throws Exception {
+        Path newSourceForEachUpload = Paths.get(directory.toString(), "bar.txt");
+
+        CompletedUploadDirectory result =
+            tm.uploadDirectory(r -> r.sourceDirectory(directory)
+                                     .bucket(TEST_BUCKET)
+                                     .uploadRequestTransformer(u -> u.toBuilder().source(newSourceForEachUpload).build()))
```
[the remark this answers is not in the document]
millems: Changed to `Consumer<UploadFileRequest.Builder>`. Offline we agreed to add the getters to builders that need them, when they need them.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/UploadDirectoryOverrideConfiguration.java
```diff
@@ -156,6 +173,53 @@ public static Builder builder() {
...
+         * being uploaded, and returns a (potentially modified) {@code UploadFileRequest}.
+         *
+         * <p>
+         * <b>Usage Example:</b>
+         * <pre>
+         * {@code
+         * // Add a LoggingTransferListener to every transfer within the upload directory request
+         * TransferRequestOverrideConfiguration fileUploadConfiguration =
+         *     TransferRequestOverrideConfiguration.builder()
+         *         .addListener(LoggingTransferListener.create())
```
Bennett-Lynch, line 192: Out of scope of this PR but worth noting: This works, but it's going to result in using the same listener for every invocation, when the current `LoggingTransferListener` implementation assumes that it's not going to be reused. We've so far refrained from adding listeners at the client-level for this reason. A client could implement a reusable listener by internally mapping something like `TransferRequest` to internal listener state, but this is a lot of complexity. I wonder if we should change listeners to be declared with a `Supplier` instead.

Bennett-Lynch (approved):

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/UploadDirectoryOverrideConfiguration.java
millems: I see, yeah, seems like it's easy to make that mistake, then. I certainly thought this would be fine. What do you recommend I do for the purposes of this PR?

Bennett-Lynch: If we moved the `LoggingTransferListener.create()` call to be inside your lambda, it would create a new one for every request and work as expected, but that may make the code less readable. If we can update the example to do that, without severely affecting legibility, then I would recommend that. Otherwise I think it's okay to use your example as-is and we can do a fast-follow-up on whether to accept a `Supplier` or not.

millems: It would unfortuantely be very difficult to make it readable with the lambdas. I'll leave it as-is and we can move it to a supplier asynchronously.

Bennett-Lynch (approved):

PR #2817, opened 2021-11-03
Refactor S3TransferManager to support non-file-based transfers

Bennett-Lynch:
# Revision 2 (https://github.com/aws/aws-sdk-java-v2/pull/2817/commits/ffc3debeefe93359b6548804b0694124d36955b8, https://github.com/aws/aws-sdk-java-v2/pull/2817/commits/4aac24aa544483e0d114f70524a9e7fba173d327)

1. Update change log to indicate breaking changes
2. Update README to reflect new API and also to include examples for TransferListener
4. Misc. small cleanups from feedback
5. Additional test coverage

# Revision 1 (https://github.com/aws/aws-sdk-java-v2/pull/2817/commits/995ca7d0112b0e1f73d058c8abf0e4b54228744a)

## Motivation and Context

We would like for S3TransferManager to support non-file-based uploads and downloads, such as reactive streams or in-memory byte buffers, but the current implementation is directly coupled to file/Path-based transfers. While the underlying S3AsyncClient already has support for AsyncRequestBodies and AsyncResponseTransformers, these constructs are not exposed through the current S3TransferManager interface.

Additionally, the existing S3TransferManager interfaces have made some assumptions that make it difficult to introduce this support, such as UploadRequest/DownloadRequest having a Path attribute, and not having interfaces that allow us to distinguish between transfer types, e.g., single-object and multi-object (directory) transfers. Likewise, the current interface hierarchy also makes it difficult to distinguish between file-based and non-file-based. Therefore, this refactoring requires **backwards-incompatible changes**, which are permissible while S3TransferManager is still in **PREVIEW**.

## Description

The diff here is very large, but this change set primarily introduces the following interface hierarchy:

---
* TransferRequest
    * TransferObjectRequest
        * UploadFileRequest
        * DownloadFileRequest
        * UploadRequest
        * DownloadRequest\<T\>
    * TransferDirectoryRequest
        * UploadDirectoryRequest
* Transfer
    * ObjectTransfer
        * FileUpload
        * FileDownload
        * Upload
        * Download\<T\>
    * DirectoryTransfer
        * DirectoryUpload
* CompletedTransfer
    * CompletedObjectTransfer
        * CompletedFileUpload
        * CompletedFileDownload
        * CompletedUpload
        * CompletedDownload\<T\>
    * CompletedDirectoryTransfer
        * CompletedDirectoryUpload
* FailedObjectTransfer
    * FailedFileUpload
---

These interfaces allow us to more selectively choose which attributes will be shared between different data types. Additionally, we take this opportunity to make the naming convention and class-vs-interface distinction consistent across all the above data types.

It is important that we distinguish between file-based-requests and non-file-based-requests so that the Collection<> of failed file transfers from a directory upload can be re-driven, if needed.

Note that the above Download-oriented data types are all generic. This is because of how AsyncResponseTransformer is designed to be generic, being parameterized with the type of data that a given AsyncResponseTransformer produces. Therefore, a user must declare the type at the time of instantiating a DownloadRequest, and the type will
be consistent throughout the transfer lifecycle, including the Download and CompletedDownload interfaces.

For a quick overview of the new API, refer to the upload/download integration tests:

### Arbitrary object transfers

https://github.com/aws/aws-sdk-java-v2/blob/995ca7d0112b0e1f73d058c8abf0e4b54228744a/services-custom/s3-transfer-manager/src/it/java/software/amazon/awssdk/transfer/s3/S3TransferManagerUploadIntegrationTest.java#L92-L99

https://github.com/aws/aws-sdk-java-v2/blob/995ca7d0112b0e1f73d058c8abf0e4b54228744a/services-custom/s3-transfer-manager/src/it/java/software/amazon/awssdk/transfer/s3/S3TransferManagerDownloadIntegrationTest.java#L79-L86

### File-based transfers

https://github.com/aws/aws-sdk-java-v2/blob/995ca7d0112b0e1f73d058c8abf0e4b54228744a/services-custom/s3-transfer-manager/src/it/java/software/amazon/awssdk/transfer/s3/S3TransferManagerUploadIntegrationTest.java#L69-L76

https://github.com/aws/aws-sdk-java-v2/blob/995ca7d0112b0e1f73d058c8abf0e4b54228744a/services-custom/s3-transfer-manager/src/it/java/software/amazon/awssdk/transfer/s3/S3TransferManagerDownloadIntegrationTest.java#L66-L72

## License


pkgonan:
@Bennett-Lynch 
Thank you so much. This is exactly the feature I've been waiting for. thank you.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/CompletedDirectoryTransfer.java
```diff
@@ -0,0 +1,56 @@
...
+     * {@code
+     * // Retrying failed uploads if the exception is retryable
+     * List<CompletableFuture<CompletedUpload>> futures =
+     *     completedDirectoryUpload.failedTransfers()
+     *                             .stream()
+     *                             .filter(failedSingleFileUpload -> isRetryable(failedSingleFileUpload.exception()))
+     *                             .map(failedSingleFileUpload ->
+     *                                  tm.upload(failedSingleFileUpload.request()).completionFuture())
+     *                             .collect(Collectors.toList());
+     * CompletableFuture.allOf(futures.toArray(new CompletableFuture[0])).join();
```
millems, line 49: Why not collect them directly to an array?

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/CompletedDirectoryUpload.java
```diff
@@ -0,0 +1,127 @@
...
+
+    private static final class DefaultBuilder implements Builder {
+        private Collection<FailedFileUpload> failedTransfers = Collections.emptyList();
+
+        private DefaultBuilder() {
+        }
+
+        @Override
+        public Builder failedTransfers(Collection<FailedFileUpload> failedTransfers) {
+            this.failedTransfers = failedTransfers;
```
millems, line 110: Copy or addAll

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/CompletedDirectoryUpload.java
```diff
@@ -0,0 +1,127 @@
...
+            return this;
+        }
+
+        public Collection<FailedFileUpload> getFailedTransfers() {
+            return failedTransfers;
+        }
```
millems, line 116: unmodifiableCollection

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/CompletedDownload.java
```diff
@@ -15,33 +15,52 @@
...
-     * @return the response
+     * Create a builder that can be used to create a {@link CompletedDownload}.
+     *
+     * @see UntypedBuilder
      */
-    public GetObjectResponse response() {
-        return response;
+    public static UntypedBuilder builder() {
+        return new DefaultUntypedBuilder();
+    }
```
millems, line 54: Our standard for typed builders is to accept the value class or value in the builder() method call: https://github.com/aws/aws-sdk-java-v2/blob/master/services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/model/PutItemEnhancedResponse.java#L69-L71

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/CompletedFileUpload.java
```diff
@@ -0,0 +1,115 @@
...
+        }
+
+        CompletedFileUpload that = (CompletedFileUpload) o;
+
+        return Objects.equals(response, that.response);
+    }
+
+    @Override
+    public int hashCode() {
+        return response != null ? response.hashCode() : 0;
```
millems, line 61: Shouldn't need null checks for non-null fields. Worth checking all objects for this, sorry. Hackathon project: codegen these stupid objects.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/CompletedUpload.java
```diff
@@ -56,12 +53,12 @@ public boolean equals(Object o) {
...
         CompletedUpload that = (CompletedUpload) o;
 
-        return response.equals(that.response);
+        return Objects.equals(response, that.response);
     }
 
     @Override
     public int hashCode() {
-        return response.hashCode();
+        return response != null ? response.hashCode() : 0;
```
millems, line 61: Same here.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/DefaultS3TransferManager.java
```diff
@@ -107,43 +113,111 @@ public Upload upload(UploadRequest uploadRequest) {
...
+            // Forward upload cancellation to CRT future
+            CompletableFutureUtils.forwardExceptionTo(uploadFuture, putObjFuture);
+
+            CompletableFutureUtils.forwardTransformedResultTo(putObjFuture, uploadFuture,
+                                                              r -> CompletedFileUpload.builder()
+                                                                                      .response(r)
+                                                                                      .build());
+        } catch (Throwable throwable) {
+            uploadFuture.completeExceptionally(throwable);
+        }
```
millems, line 159: Any chance that this lets the putObject keep running in the background? If so, can we cancel that as well?

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/CompletedDirectoryTransfer.java
```diff
@@ -0,0 +1,56 @@
...
+     * Failed single object transfers can be retried by calling {@link S3TransferManager#uploadFile(UploadFileRequest)} or
+     * {@link S3TransferManager#downloadFile(DownloadFileRequest)}.
+     *
+     * <pre>
+     * {@code
+     * // Retrying failed uploads if the exception is retryable
+     * List<CompletableFuture<CompletedUpload>> futures =
+     *     completedDirectoryUpload.failedTransfers()
+     *                             .stream()
+     *                             .filter(failedSingleFileUpload -> isRetryable(failedSingleFileUpload.exception()))
```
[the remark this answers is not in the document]
zoewangg: Yeah, I meant to delete this, but looks like I missed this place

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/DefaultDirectoryUpload.java
```diff
@@ -0,0 +1,66 @@
...
+
+import java.util.Objects;
+import java.util.concurrent.CompletableFuture;
+import software.amazon.awssdk.annotations.SdkPreviewApi;
+import software.amazon.awssdk.annotations.SdkPublicApi;
+import software.amazon.awssdk.transfer.s3.CompletedDirectoryUpload;
+import software.amazon.awssdk.transfer.s3.DirectoryUpload;
+import software.amazon.awssdk.utils.ToString;
+
+@SdkPublicApi
```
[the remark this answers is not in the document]
Bennett-Lynch: Good catch. Artifact from renaming & extracting interfaces. Will double-check others.

zoewangg (commented):
Can we update the sample code in the README? cc @Carey-AWS, can we update dev guide?

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/DefaultS3TransferManager.java
Bennett-Lynch: Hmm, I was thinking that line 151 (`forwardExceptionTo(..)`) would do this for us, but I'm not sure if `completeExceptionally(..)` will also implicitly `cancel(..)`?

If not, then perhaps we can extend `forwardExceptionTo(..)` to check for `CancellationException`, and attempt to `cancel(..)` first?

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/CompletedFileUpload.java
```diff
@@ -0,0 +1,115 @@
...
+            return false;
+        }
+
+        CompletedFileUpload that = (CompletedFileUpload) o;
+
+        return Objects.equals(response, that.response);
+    }
+
+    @Override
+    public int hashCode() {
```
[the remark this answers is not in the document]
Bennett-Lynch: I will double-check. I will also add some additional test coverage for TM itself.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/DefaultFileDownload.java
```diff
@@ -0,0 +1,80 @@
...
+import java.util.Objects;
+import java.util.concurrent.CompletableFuture;
+import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.transfer.s3.CompletedFileDownload;
+import software.amazon.awssdk.transfer.s3.FileDownload;
+import software.amazon.awssdk.transfer.s3.progress.TransferProgress;
+import software.amazon.awssdk.utils.ToString;
+
+@SdkInternalApi
+public final class DefaultFileDownload implements FileDownload {
```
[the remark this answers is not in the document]
Bennett-Lynch: I believe we want this to be an interface since we anticipate adding more complex logic here, like pause/resume/cancel, at which point it is no longer a simple data class. On the other hand, the request and completed objects should always be simple data classes.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/DefaultS3TransferManager.java
millems: If forwardExceptionTo throws an exception, it could, but I agree that's unlikely... unless someone comes along and adds more lines in there.

zoewangg: This is the cancellation logic in S3 CRT client https://github.com/awslabs/aws-crt-java/blob/main/src/main/java/s3NativeClient/com/amazonaws/s3/S3NativeClient.java#L158 Looks like it checks if the future is cancelled or not.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/DefaultFileDownload.java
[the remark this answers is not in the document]
zoewangg: Makes sense.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/CompletedFileUpload.java
zoewangg: A tool that will stop working for future Java versions :) (the bear has been poked)

Bennett-Lynch: A tool that's a build-time dependency only, and for which we'll typically have years to react to any such discontinuation (which I'm convinced won't happen without an official replacement)!

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/CompletedFileUpload.java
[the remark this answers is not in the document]
Bennett-Lynch: Made sure all of the new classes have equals/hashCode coverage.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/DefaultS3TransferManager.java
Bennett-Lynch: So I think it's probably a potential miss that `forwardExceptionTo(..)` is not forwarding/propagating cancellation. However, I'm not sure if it's safe to add that logic since the method is used in other places, including in generated clients. Alternatively, we create a new `forwardExceptionAndCancellationTo(..)` method, but it seems unfortunate to have to duplicate several of the existing methods, rather than just fix them. To prevent this PR from growing even further, I'd like to propose deferring this change to a separate, follow-up PR (which I will track).

Bennett-Lynch:
> Can we update the sample code in the README?

I updated the README. I wasn't sure if we should include examples for the arbitrary upload/download, or just file transfers (to keep it simple). Please take a look at let me know what you prefer.

zoewangg (approved):
Thanks for updating the README, LGTM. I think we should have sample code for common use-cases like what you added. The DDB enhanced client README is a good example, imo and we've got positive feedback on it. https://github.com/aws/aws-sdk-java-v2/tree/master/services-custom/dynamodb-enhanced

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/CompletedDirectoryTransfer.java
```diff
@@ -0,0 +1,56 @@
...
+     *                             .map(failedSingleFileUpload ->
+     *                                  tm.upload(failedSingleFileUpload.request()).completionFuture())
+     *                             .collect(Collectors.toList());
+     * CompletableFuture.allOf(futures.toArray(new CompletableFuture[0])).join();
+     * }
+     * </pre>
+     *
+     * @return a list of failed transfers
+     */
+    Collection<? extends FailedObjectTransfer> failedTransfers();
```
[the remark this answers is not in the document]
millems: Ah, gotcha, didn't realize that we had public subclasses. I take back my comment!

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/CompletedDownload.java
millems: Do we need to create the `CompletedDownload` before we have the `ResponseBytes<GetObjectResponse>`? If not, we can just pass that into the `builder` method.

I guess we could also do `CompletedDownload.<ResponseBytes<GetObjectResponse>>builder()` to initialize the type parameters at creation time, if we absolutely have to. It's not particularly intuitive for customers to do, but if they don't need to do it much (since it's primarily a response type) it's simpler than the mixed builder approach (imo).

Bennett-Lynch: > Do we need to create the CompletedDownload before we have the ResponseBytes<GetObjectResponse>? If not, we can just pass that into the builder method.

We do need to specify the type. The current flow looks like:
1. `DownloadRequest<T>` contains an `AsyncResponseTransformer<GetObjectResponse,T>`
1. `Download<T>` contains a `CompleteableFuture<CompletedDownload<T>>`
1. `CompletedDownload<T>` contains a `T`

> I guess we could also do CompletedDownload.<ResponseBytes<GetObjectResponse>>builder() to initialize the type parameters at creation time, if we absolutely have to. It's not particularly intuitive for customers to do, but if they don't need to do it much (since it's primarily a response type) it's simpler than the mixed builder approach (imo).

We can definitely use that method, but that type of syntax is what I was explicitly trying to avoid, since I think most users will find it very cumbersome and unfamiliar. While the mixed builder approach is more complexity on our end, most users don't need to be concerned with it if they're declaring their builders in one sweep.

Typical user code with a single, generic builder:

```
Download<ResponseBytes<GetObjectResponse>> download =
    tm.download(DownloadRequest.<ResponseBytes<GetObjectResponse>>builder()
                               .getObjectRequest(b -> b.bucket(BUCKET).key(KEY))
                               .responseTransformer(AsyncResponseTransformer.toBytes())
                               .build());
```

Typical user code with the mixed builder:

```
Download<ResponseBytes<GetObjectResponse>> download =
    tm.download(DownloadRequest.builder()
                               .getObjectRequest(b -> b.bucket(BUCKET).key(KEY))
                               .responseTransformer(AsyncResponseTransformer.toBytes())
                               .build());
```

In this case, it only simplifies things from a user perspective.

Bennett-Lynch: I also tried to document this behavior here:

https://github.com/aws/aws-sdk-java-v2/blob/9f8e42b08a34cc32d7ba7ac92b340be8153de50c/services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/DownloadRequest.java#L132-L142

millems (approved):

PR #2808, opened 2021-10-29
Added support for endpoint variants in endpoint metadata: FIPS and DUALSTACK.

millems:
This adds the ability to resolve fips and dualstack endpoints with a "fips enabled" or "dualstack enabled" flag.

E.g.
```java
URI fipsEndpoint =
    ServiceMetadata.of(S3Client.SERVICE_METADATA_ID)
                   .endpointFor(ServiceEndpointKey.builder()
                                                  .region(Region.US_WEST_2)
                                                  .tags(EndpointTag.FIPS)
                                                  .build());
```

The following public API changes were made:
1. Added `ServiceMetadata.endpointFor(ServiceEndpointKey)`
2. Added `ServiceMetadata.signingRegion(ServiceEndpointKey)`
3. Added `PartitionMetadata.hostname(PartitionEndpointKey)`
4. Added `PartitionMetadata.dnsSuffix(PartitionEndpointKey)`
5. Added a generated `EndpointTag` class, which currently contains FIPS and DUALSTACK.
6. Removed `ServiceMetadata.computeEndpoint`, because it shouldn't have ever been public, and no one should be using it.

The following module changes were made:
1. Added `regions-testing` for verifying region resolution behavior that doesn't currently exist in the endpoints.json

PR #2818, opened 2021-11-03
Added a new `dualstackEnabled` property to every client builder.

millems:
This option can be used to make calls be invoked against AWS endpoints which return IPv6 records. This can also be enabled via the `AWS_USE_DUALSTACK_ENDPOINT` environment variable, `aws.useDualstackEndpoint` system property, or the `use_dualstack_endpoint` profile file property:

1. This PR builds on top of the 'endpoint variants' #2808, by specifying the DUALSTACK endpoint variant whenever the customer requests dualstack to be enabled.
2. May of the file changes in this PR are addressing the fact that we already have service configuration for enabling dualstack for S3 and S3 control. These existing options have been deprecated in favor of the new option that works for all services.
3. There are also a lot of changes because this is the first time we're reading the profile file during endpoint resolution, outside of S3, so that information had to be piped through in all places we're resolving endpoints.

### DIFF
--- a/codegen/src/main/java/software/amazon/awssdk/codegen/poet/builder/BaseClientBuilderClass.java
+++ b/codegen/src/main/java/software/amazon/awssdk/codegen/poet/builder/BaseClientBuilderClass.java
@@ -34,6 +34,7 @@ import javax.lang.model.element.Modifier;
 import software.amazon.awssdk.annotations.SdkInternalApi;
 import software.amazon.awssdk.auth.signer.Aws4Signer;
 import software.amazon.awssdk.awscore.client.builder.AwsDefaultClientBuilder;
+import software.amazon.awssdk.awscore.client.config.AwsClientOption;
 import software.amazon.awssdk.codegen.internal.Utils;
 import software.amazon.awssdk.codegen.model.intermediate.IntermediateModel;
 import software.amazon.awssdk.codegen.model.intermediate.OperationModel;
@@ -54,6 +55,7 @@ import software.amazon.awssdk.protocols.query.interceptor.QueryParametersToBodyI
 import software.amazon.awssdk.utils.AttributeMap;
 import software.amazon.awssdk.utils.CollectionUtils;
 import software.amazon.awssdk.utils.StringUtils;
+import software.amazon.awssdk.utils.Validate;
 
 public class BaseClientBuilderClass implements ClassSpec {
     private final IntermediateModel model;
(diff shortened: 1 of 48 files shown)


codegen/src/main/java/software/amazon/awssdk/codegen/poet/builder/BaseClientBuilderClass.java
```diff
@@ -241,9 +243,20 @@ private MethodSpec finalizeServiceConfigurationMethod() {
         if (StringUtils.isNotBlank(clientConfigClassName)) {
             ClassName clientConfigClass = ClassName.bestGuess(clientConfigClassName);
             builder.addCode("$1T.Builder c = (($1T) config.option($2T.SERVICE_CONFIGURATION)).toBuilder();" +
-                            "c.profileFile(c.profileFile() != null ? c.profileFile() : config.option($2T.PROFILE_FILE))" +
-                            " .profileName(c.profileName() != null ? c.profileName() : config.option($2T.PROFILE_NAME));",
+                            "c.profileFile(c.profileFile() != null ? c.profileFile() : config.option($2T.PROFILE_FILE));" +
+                            "c.profileName(c.profileName() != null ? c.profileName() : config.option($2T.PROFILE_NAME));",
                             clientConfigClass, SdkClientOption.class);
+
+            if (model.getCustomizationConfig().getServiceConfigHasDualstackConfig()) {
```
joviegas, line 250: Can this be replaced with a static public function in some util class and the implementation of if else can go there

```java
AwsConfigUtils.validateAndSetDualstackEnabledOption(ServiceConfiguration.Builder builder ,boolean dualEndpoint)
```

which can be called 

```java
AwsConfigUtils.validateAndSetDualstackEnabledOption(c,config.option($T.DUALSTACK_ENDPOINT_ENABLED))
```

millems: What would be the pros/cons?

joviegas: Pros : 
1. Readability in codegen code.
2. In future if we want to add some validation  or change the message of Validations , we need not change the codegen or its test cases.

millems: A few cons I see:
1. It adds a new protected API, which means we have a bigger backwards-compatibility surface area to worry about. As-is, the code is entirely within the service. We could get around that by finding a way to copy the implementation into every service module that needs it, but I think that would hurt codegen readability more than what we have here.
2. The service configurations do not share an interface, so we'd have to add a parent interface to the builder that allows reading and writing the dualstack endpoint... That one would be a new public API, which means even more backwards-compatibility surface area.

As for your pros:
1. I'm not entirely convinced it's much more readable to put into a separate function, since it's just 3 lines of code and it needs such a long function name to describe what it does.
2. It definitely makes it more testable in isolation, but we'd still have to test the fact that it is wired up in codegen, so we'd still have to test it indirectly like the test cases I added. If it was more than 3 lines, then I think there's a stronger argument for testability, since testing that it's wired up would be less work than actually testing every edge case within the function.

joviegas (approved):

codegen/src/main/java/software/amazon/awssdk/codegen/poet/builder/BaseClientBuilderClass.java
joviegas: Thanks for listing the details  👍 , good leanings for me

PR #2825, opened 2021-11-04
[Default Configuration Part 6]: apply default s3 us-east-1 regional setting

zoewangg:
- Add `ServiceMetadataAdvancedOption` to support configuring s3 use us-east-1 regional setting
- Apply default s3 us-east-1 regional setting

### DIFF
--- a/core/aws-core/src/main/java/software/amazon/awssdk/awscore/endpoint/DefaultServiceEndpointBuilder.java
+++ b/core/aws-core/src/main/java/software/amazon/awssdk/awscore/endpoint/DefaultServiceEndpointBuilder.java
@@ -17,13 +17,16 @@ package software.amazon.awssdk.awscore.endpoint;
 
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import software.amazon.awssdk.annotations.NotThreadSafe;
 import software.amazon.awssdk.annotations.SdkProtectedApi;
 import software.amazon.awssdk.core.exception.SdkClientException;
 import software.amazon.awssdk.profiles.ProfileFile;
 import software.amazon.awssdk.regions.Region;
 import software.amazon.awssdk.regions.ServiceMetadata;
+import software.amazon.awssdk.regions.ServiceMetadataAdvancedOption;
 import software.amazon.awssdk.utils.Validate;
 
 /**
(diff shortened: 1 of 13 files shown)


core/aws-core/src/main/java/software/amazon/awssdk/awscore/endpoint/DefaultServiceEndpointBuilder.java
```diff
@@ -40,6 +43,7 @@
     private Region region;
     private ProfileFile profileFile;
     private String profileName;
+    private final Map<ServiceMetadataAdvancedOption<?>, Object> advancedOptions = new HashMap<>();
```
millems, line 46: Does AttributeMap.Builder work for this?

core/regions/src/main/java/software/amazon/awssdk/regions/servicemetadata/EnhancedS3ServiceMetadata.java
```diff
@@ -93,7 +95,8 @@ private boolean useUsEast1RegionalEndpoint(Supplier<ProfileFile> profileFile, Su
             return REGIONAL_SETTING.equalsIgnoreCase(profile);
         }
 
-        return false;
+        return config.advancedOption(ServiceMetadataAdvancedOption.S3_US_EAST_1_REGIONAL_ENDPOINT)
+                     .filter(REGIONAL_SETTING::equalsIgnoreCase).isPresent();
```
millems, line 99: Usually directly-configured options are higher priority than environment variables and system properties. I know in this case it's actually a default instead of a directly configured option (the customer can't configure this directly, right?), but should we name it accordingly to clarify it's the default, not the actual setting?

zoewangg: Yeah, good point. Customers can't configure this if they are using the SDK client. I'll rename it to reflect that it's the default.

core/aws-core/src/main/java/software/amazon/awssdk/awscore/endpoint/DefaultServiceEndpointBuilder.java
zoewangg: It does, but the public API `ServiceMetadataConfiguration` would need to take AttributeMap or AttributeMap.Builder instead, which is not consistent with `ClientOverrideConfiguration#advancedOptions` 

https://github.com/aws/aws-sdk-java-v2/blob/040b7fde9da28c58b6d57cfad76b40de719369a9/core/regions/src/main/java/software/amazon/awssdk/regions/ServiceMetadataConfiguration.java#L116

https://github.com/aws/aws-sdk-java-v2/blob/040b7fde9da28c58b6d57cfad76b40de719369a9/core/sdk-core/src/main/java/software/amazon/awssdk/core/client/config/ClientOverrideConfiguration.java#L350

joviegas (approved):

PR #2826, opened 2021-11-04
Added support for enabling FIPS endpoints.

millems:
This option can be used to make calls be invoked against FIPS-compliant AWS endpoints. This can also be enabled via the AWS_USE_FIPS_ENDPOINT environment variable, aws.useFipsEndpoint system property, or the use_fips_endpoint profile file property.

1. This builds on top of #2808 and #2818
2. This replaces (deprecates) the existing "fipsModeEnabled" option in the S3ControlConfiguration.

### DIFF
--- a/codegen/src/main/java/software/amazon/awssdk/codegen/poet/builder/BaseClientBuilderClass.java
+++ b/codegen/src/main/java/software/amazon/awssdk/codegen/poet/builder/BaseClientBuilderClass.java
@@ -102,7 +102,7 @@ public class BaseClientBuilderClass implements ClassSpec {
         builder.addMethod(defaultSignerMethod());
         builder.addMethod(signingNameMethod());
 
-        if (model.getCustomizationConfig().getServiceSpecificClientConfigClass() != null) {
+        if (model.getCustomizationConfig().getServiceConfig().getClassName() != null) {
             builder.addMethod(setServiceConfigurationMethod())
                    .addMethod(beanStyleSetServiceConfigurationMethod());
         }
@@ -167,7 +167,7 @@ public class BaseClientBuilderClass implements ClassSpec {
                                                         + ".CRC32_FROM_COMPRESSED_DATA_ENABLED, $L)",
                                                         SdkClientOption.class, crc32FromCompressedDataEnabled);
 
-        String clientConfigClassName = model.getCustomizationConfig().getServiceSpecificClientConfigClass();
+        String clientConfigClassName = model.getCustomizationConfig().getServiceConfig().getClassName();
         if (StringUtils.isNotBlank(clientConfigClassName)) {
             builder.addCode(".option($T.SERVICE_CONFIGURATION, $T.builder().build())",
                             SdkClientOption.class, ClassName.bestGuess(clientConfigClassName));
@@ -239,7 +239,7 @@ public class BaseClientBuilderClass implements ClassSpec {
                    .endControlFlow();
         }
 
-        String clientConfigClassName = model.getCustomizationConfig().getServiceSpecificClientConfigClass();
+        String clientConfigClassName = model.getCustomizationConfig().getServiceConfig().getClassName();
         if (StringUtils.isNotBlank(clientConfigClassName)) {
             ClassName clientConfigClass = ClassName.bestGuess(clientConfigClassName);
             builder.addCode("$1T.Builder c = (($1T) config.option($2T.SERVICE_CONFIGURATION)).toBuilder();" +
@@ -247,7 +247,7 @@ public class BaseClientBuilderClass implements ClassSpec {
                             "c.profileName(c.profileName() != null ? c.profileName() : config.option($2T.PROFILE_NAME));",
                             clientConfigClass, SdkClientOption.class);
 
-            if (model.getCustomizationConfig().getServiceConfigHasDualstackConfig()) {
+            if (model.getCustomizationConfig().getServiceConfig().hasDualstackProperty()) {
                 builder.addCode("if (c.dualstackEnabled() != null) {")
                        .addCode("    $T.validState(config.option($T.DUALSTACK_ENDPOINT_ENABLED) == null, \"Dualstack has been "
                                 + "configured on both $L and the client/global level. Please limit dualstack configuration to "
@@ -257,11 +257,30 @@ public class BaseClientBuilderClass implements ClassSpec {
                        .addCode("    c.dualstackEnabled(config.option($T.DUALSTACK_ENDPOINT_ENABLED));", AwsClientOption.class)
                        .addCode("}");
             }
+
+            if (model.getCustomizationConfig().getServiceConfig().hasFipsProperty()) {
+                builder.addCode("if (c.fipsModeEnabled() != null) {")
+                       .addCode("    $T.validState(config.option($T.FIPS_ENDPOINT_ENABLED) == null, \"Fips has been "
+                                + "configured on both $L and the client/global level. Please limit fips configuration to "
+                                + "one location.\");",
+                                Validate.class, AwsClientOption.class, clientConfigClassName)
+                       .addCode("} else {")
+                       .addCode("    c.fipsModeEnabled(config.option($T.FIPS_ENDPOINT_ENABLED));", AwsClientOption.class)
+                       .addCode("}");
+            }
         }
 
         // Update configuration
 
         builder.addCode("return config.toBuilder()\n");
+
+        if (model.getCustomizationConfig().getServiceConfig().hasDualstackProperty()) {
+            builder.addCode(".option($T.DUALSTACK_ENDPOINT_ENABLED, c.dualstackEnabled())", AwsClientOption.class);
+        }
+
+        if (model.getCustomizationConfig().getServiceConfig().hasFipsProperty()) {
+            builder.addCode(".option($T.FIPS_ENDPOINT_ENABLED, c.fipsModeEnabled())", AwsClientOption.class);
+        }
 
         if (model.getEndpointOperation().isPresent()) {
             builder.addCode(".option($T.ENDPOINT_DISCOVERY_ENABLED, endpointDiscoveryEnabled)\n",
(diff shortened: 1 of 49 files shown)


codegen/src/main/java/software/amazon/awssdk/codegen/poet/builder/BaseClientBuilderClass.java
```diff
@@ -257,12 +257,31 @@ private MethodSpec finalizeServiceConfigurationMethod() {
                        .addCode("    c.dualstackEnabled(config.option($T.DUALSTACK_ENDPOINT_ENABLED));", AwsClientOption.class)
                        .addCode("}");
             }
+
+            if (model.getCustomizationConfig().getServiceConfigHasFipsConfig()) {
```
zoewangg, line 261: Can we add a codegen test for this? I guess we can probably test both `serviceConfigHasFipsConfig` and `serviceConfigHasDualstackConfig` together

core/sdk-core/src/main/java/software/amazon/awssdk/core/client/builder/SdkDefaultClientBuilder.java
```diff
@@ -353,10 +353,14 @@ private ScheduledExecutorService resolveScheduledExecutorService() {
 
     @Override
     public final B endpointOverride(URI endpointOverride) {
-        Validate.paramNotNull(endpointOverride, "endpointOverride");
-        Validate.paramNotNull(endpointOverride.getScheme(), "The URI scheme of endpointOverride");
-        clientConfiguration.option(SdkClientOption.ENDPOINT, endpointOverride);
-        clientConfiguration.option(SdkClientOption.ENDPOINT_OVERRIDDEN, true);
```
zoewangg, line 359: Why are we removing the null check for endpointoverride?

services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Utilities.java
```diff
@@ -86,20 +91,39 @@
...
+                                               b.dualstackEnabled(builder.dualstackEnabled);
+                                           }
+
+                                           if (b.dualstackEnabled() == null) {
+                                               b.dualstackEnabled(DualstackEnabledProvider.builder()
+                                                                                          .profileFile(profileFile)
+                                                                                          .profileName(profileName)
+                                                                                          .build()
+                                                                                          .isDualstackEnabled()
+                                                                                          .orElse(false));
```
zoewangg, line 114: Might be readable if it's something like below imo

```java
if (b.dualstackEnabled() == null) {
    boolean dualstackEnabled = Optional.ofNullable(builder.dualstackEnabled)
                                       .orElseGet(() -> DualStckEnabledProvider.builder()...)
   
    b.dualstackEnabled(dualstackEnabled);
}
```

services/s3/src/test/java/software/amazon/awssdk/services/s3/S3EndpointResolutionTest.java
```diff
@@ -278,6 +344,15 @@ public void accelerateEnabled_UsesVirtualAddressingWithAccelerateEndpoint() thro
                               String.format("https://%s.s3-accelerate.amazonaws.com", BUCKET));
     }
 
+    /**
+     * S3 accelerate and FIPS should not be allowed in use together.
+     */
+    @Test(expected = IllegalStateException.class)
+    public void accelerateAndFips_Fails() throws Exception {
+        mockHttpClient.stubNextResponse(mockListObjectsResponse());
+        clientBuilder().fipsEnabled(true).serviceConfiguration(withAccelerateEnabled()).build().listObjects(r -> r.bucket(BUCKET));
```
zoewangg, line 353: Can we verify the exception message to make sure the correct exception is thrown?

services/s3control/src/main/resources/codegen-resources/customization.config
```diff
@@ -1,6 +1,7 @@
 {
     "serviceSpecificClientConfigClass": "S3ControlConfiguration",
     "serviceConfigHasDualstackConfig": true,
+    "serviceConfigHasFipsConfig": true,
```
zoewangg, line 4: Have we considered changing `serviceSpecificClientConfigClass` to an object and making  `serviceConfigHasFipsConfig` and `serviceConfigHasDualstackConfig` its fields because they seem to only apply to services that have service config

test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/EndpointVariantResolutionTest.java
```diff
@@ -0,0 +1,71 @@
+package software.amazon.awssdk.services;
```
zoewangg, line 1: nit, missing copyright header

core/aws-core/src/main/java/software/amazon/awssdk/awscore/client/builder/AwsDefaultClientBuilder.java
```diff
@@ -237,6 +240,29 @@ private Boolean dualstackEndpointFromDefaultProvider(SdkClientConfiguration conf
...
+     */
+    private Boolean resolveUseFipsFromDefaultProvider(SdkClientConfiguration config) {
+        ProfileFile profileFile = config.option(SdkClientOption.PROFILE_FILE);
+        String profileName = config.option(SdkClientOption.PROFILE_NAME);
+        return FipsEnabledProvider.builder()
+                                  .profileFile(() -> profileFile)
+                                  .profileName(profileName)
+                                  .build()
+                                  .isFipsEnabled()
+                                  .orElse(null);
```
zoewangg, line 263: Is there any reason we defer to `DefaultServiceEndpointBuilder` to resolve the default value?

millems: Old modules that don't currently specify those fields. It's for backwards compatibility.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/endpoints/S3AccessPointEndpointResolver.java
```diff
@@ -199,7 +201,7 @@ private URI getUriForAccessPointResource(S3EndpointResolverContext context,
                                    .endpointOverride(context.endpointOverride())
                                    .accessPointName(s3EndpointResource.accessPointName())
                                    .accountId(s3EndpointResource.accountId().get())
-                                   .fipsEnabled(isFipsRegion(context.region().toString()))
+                                   .fipsEnabled(context.fipsEnabled() || isFipsRegion(context.region().toString()))
```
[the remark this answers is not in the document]
millems: Done

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/endpoints/S3EndpointResolverContext.java
```diff
@@ -150,6 +177,25 @@ public Builder disableHostPrefixInjection(boolean disableHostPrefixInjection) {
             return this;
         }
 
+        public Builder fipsEnabled(Boolean fipsEnabled) {
+            if (fipsEnabled == null) {
+                this.fipsEnabled = false;
+            } else {
+                this.fipsEnabled = fipsEnabled;
+            }
```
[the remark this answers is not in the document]
millems: We can't ignore null, because `context.fipsEnabled(true); context.fipsEnabled(null);` would be fipsEnabled = true, not false which is what I believe is correct.

If this pattern is weird, I can delay resolving null to false until the context is built instead of doing it in the builder.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/endpoints/S3EndpointResolverContext.java
```diff
@@ -72,6 +84,15 @@ public boolean isDisableHostPrefixInjection() {
         return disableHostPrefixInjection;
     }
 
+    public Supplier<ProfileFile> profileFile() {
+        // TODO: Remove?
```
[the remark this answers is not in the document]
millems: Nice! Glad my TODO got caught, even if it wasn't by me 😆

codegen/src/main/java/software/amazon/awssdk/codegen/poet/builder/BaseClientBuilderClass.java
millems: Done

core/sdk-core/src/main/java/software/amazon/awssdk/core/client/builder/SdkDefaultClientBuilder.java
millems: Because we're now handling null as "reset to default". Essentially, endpointOverride won't ever be null in this statement block.

services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Utilities.java
millems: I just tried that out, and it ends up being too much nesting to keep track of in my head at once. I tried breaking it out into other methods and ended up just rewriting it, since I assume this comment was because that it was hard to understand.

services/s3/src/test/java/software/amazon/awssdk/services/s3/S3EndpointResolutionTest.java
millems: Done

test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/EndpointVariantResolutionTest.java
millems: Fixed

services/s3control/src/main/resources/codegen-resources/customization.config
millems: That likely wouldn't break anyone. I'd be surprised if anyone is using serviceSpecificClientConfigClass outside of us, and code search seems to support that. I'll make that change.

core/aws-core/src/main/java/software/amazon/awssdk/awscore/client/builder/AwsDefaultClientBuilder.java
millems: If you're asking why we use null here instead of false: it's so that service-specific config like S3ControlConfiguration resolution logic can differentiate between the customer setting false, and the default value for fips/dualstack being false.

zoewangg (approved):

Issue #2609, opened 2021-07-20
DynamoDbEnhancedClient: Describe table operation 


reply2srij:
The DynamoDbEnhancedClient returns a DynamoDbTable that allows you to perform various dynamodb operations. Good to have an operation similar to Describe table as well. 

## Your Environment
* AWS Java SDK version used: 2.16.99
* JDK version used: java-openjdk-11.0.8.10
* Operating System and version: Windows 10

debora-ito:
@reply2srij thank you for reaching out, marking as a feature request.

zoewangg:
@reply2srij thank you for your patience. We have implemented the feature in #2844. Please try with the latest Java SDK.

PR #2844, opened 2021-11-12
Support DescribeTable operation in DynamoDbTable and DynamodbAsyncTable

zoewangg:
## Motivation and Context
#2609

## Description
Support DescribeTable operation in DynamoDbTable and DynamodbAsyncTable

## Testing
Added unit tests and functional tests

## Types of changes

## Checklist

## License


services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/model/DescribeTableEnhancedResponse.java
```diff
@@ -0,0 +1,87 @@
...
+    private final DescribeTableResponse response;
+
+    private DescribeTableEnhancedResponse(Builder builder) {
+        this.response = Validate.paramNotNull(builder.response, "response");
+    }
+
+    /**
+     * The {@link DescribeTableResponse} from the low level DynamoDB client
+     */
+    public DescribeTableResponse describeTableResponse() {
```
dagnir, line 41: Can we remove these? The other `Enhanced*Response` objects don't have it. I think ideally we don't want this on the enhanced response because we should expose the relevant stuff on the enhanced object if we need to.

zoewangg: I think all fields in the `DescribeTableResponse` is relevant to a describeTable call. You mean copying all fields in `DescribeTableResponse` here?  My concern is that we will need to make code change in the future if new fields are added

dagnir: Since the enhanced client is supposed to replace the majority of the use cases for the low level client, I feel like asking the customer to work with the low level response goes against the spirit of that.

```
enhancedResponse.describeTableResponse().tableName();
```

Also looks awkward to me

zoewangg: For other operations such as getItem, yes, but I'm not sure about control plane APIa. Do you have any suggestions? I'd be hesitant to manually copying all the fields. It's a lot of ops burden because they can be out-of-date soon

dagnir: Would you be okay with just keeping a reference to the low level response internally, and just exposing direct getters on the object? Best middle ground I can think of

```
public String tableName() {
    return lowLevelResponse.tableName();
}
```

zoewangg: `DescribeTableResponse` only has `TableDescription` object actually. I guess we can do 
```java
public String table() {
    return lowLevelResponse.table();
}
```

dagnir (approved):

PR #2846, opened 2021-11-16
Adds operation name labels for operations to identify calling op when required

cenedhryn:
## Motivation and Context
Adding these labels make it possible for the client to distinguish between different operations which is useful for the extensions framework. 

## Description
- Added the label 'operationType' to all operations
- Added operationType to the DynamoDbExtensionContext.BeforeWrite interface (the read framework isn't as tied to a specific operation and at this point in time doesn't require it nor is it easy to retrofit)

## Types of changes

## Checklist

## License


services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/operations/OperationType.java
```diff
@@ -0,0 +1,50 @@
...
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.enhanced.dynamodb.internal.operations;
+
+import software.amazon.awssdk.annotations.SdkInternalApi;
+
+@SdkInternalApi
+public enum OperationType {
+    ANY("Any"), //unknown value
```
zoewangg, line 22: Should it be `UNKNOWN_TO_SDK_VERSION(null);` to be consistent with the service enum?

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/extensions/DefaultDynamoDbExtensionContext.java
```diff
@@ -96,6 +107,7 @@ public int hashCode() {
         result = 31 * result + (operationContext != null ? operationContext.hashCode() : 0);
         result = 31 * result + (tableMetadata != null ? tableMetadata.hashCode() : 0);
         result = 31 * result + (tableSchema != null ? tableSchema.hashCode() : 0);
+        result = 31 * result + (operationType != null ? operationType.hashCode() : 0);
```
zoewangg, line 110: Do we have equals hashcode test? If not, can we add one using equalsVerifier?

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/operations/OperationType.java
cenedhryn: Sounds better, lm check that out.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/extensions/DefaultDynamoDbExtensionContext.java
cenedhryn: good suggestion!

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/operations/OperationType.java
millems: Maybe just UNKNOWN? Because this is internal, why would we even need an UNKNOWN value? Can't we just always specify it for our operations?

cenedhryn: We don't always know the operation in all code paths flowing through an operation. Especially on the read path, we make transforms and use extensions without coming directly from an operation. An example is the `readAndTransform..Item` operations in `EnhancedClientUtils`. 
We could use null instead, which I don't prefer but I don't feel strongly about it.

zoewangg (approved):

PR #2862, opened 2021-11-19
Adds all fields in RequestOverrideConfiguration in toBuilder

cenedhryn:
## Motivation and Context
When a `requestOverrideConfiguration` is modified from an existing instance all values of existing fields should be copied. For instance, `metricPublishers` was missing, leading to the paginated operations code path missing metric publishers added as a request override and no metrics are published. 

## Description
Updated the `BuilderImpl` constructor that takes a `RequestOverrideConfiguration` as an argument. 

## Testing
- Added unit tests to cover minimal and maximal cases of converting to builder and back
- Added functional test to verify usage of publishers

## Types of changes

## Checklist

## License

### DIFF
--- a/core/sdk-core/src/test/java/software/amazon/awssdk/core/RequestOverrideConfigurationTest.java
+++ b/core/sdk-core/src/test/java/software/amazon/awssdk/core/RequestOverrideConfigurationTest.java
@@ -15,7 +15,6 @@
 
 package software.amazon.awssdk.core;
 
-
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.Assert.fail;
 import static org.mockito.Mockito.mock;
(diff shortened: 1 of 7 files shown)


dagnir (approved):

core/sdk-core/src/test/java/software/amazon/awssdk/core/RequestOverrideConfigurationTest.java
```diff
@@ -27,16 +26,53 @@
...
+                .putRawQueryParameter(QUERY_PARAM, "foo")
+                .addApiName(a -> a.name("test1").version("1"))
+                .apiCallTimeout(Duration.ofSeconds(1))
+                .apiCallAttemptTimeout(Duration.ofSeconds(1))
+                .signer(new NoOpSigner())
+                .executionAttributes(ExecutionAttributes.builder().put(testAttribute, expectedValue).build())
+                .addMetricPublisher(mock(MetricPublisher.class))
+                .build();
+
+        assertThat(configuration.toBuilder().build()).isEqualTo(configuration);
```
Bennett-Lynch, line 73: Could we use something like `usingRecursiveComparison()` here?

https://assertj.github.io/doc/#assertj-core-recursive-comparison

Or does the combination of `EqualsVerifier` and `isEqualTo` provide similar guarantees?

cenedhryn: In this situation they yield similar results, but I do like the more verbose output of `usingRecursiveComparison()`

Bennett-Lynch (approved):

core/sdk-core/src/test/java/software/amazon/awssdk/core/RequestOverrideConfigurationTest.java
```diff
@@ -27,16 +26,53 @@
...
+                .putRawQueryParameter(QUERY_PARAM, "foo")
+                .addApiName(a -> a.name("test1").version("1"))
+                .apiCallTimeout(Duration.ofSeconds(1))
+                .apiCallAttemptTimeout(Duration.ofSeconds(1))
+                .signer(new NoOpSigner())
+                .executionAttributes(ExecutionAttributes.builder().put(testAttribute, expectedValue).build())
+                .addMetricPublisher(mock(MetricPublisher.class))
+                .build();
+
+        assertThat(configuration.toBuilder().build()).usingRecursiveComparison().isEqualTo(configuration);
```
Bennett-Lynch, line 73: It would be nice if we could find a way to enforce this validation for all of our classes that implement `toBuilder()`.

PR #2863, opened 2021-11-20
Enable BlockHound for integration tests and stability tests

Bennett-Lynch:
This PR enables [BlockHound](https://github.com/reactor/BlockHound) for the SDK integration test and stability tests. BlockHound can be used to detect blocking behavior in threads that must be treated as non-blocking, like Netty's EventLoop threads.

If BlockHound detects an illegal blocking operation being performed, it will throw a BlockingOperationError. Doing so in an integration test would typically fail the integration test and allow us to investigate the behavior.

This PR enables BlockHound by leveraging BlockHound's provided [BlockHoundTestExecutionListener](https://github.com/reactor/BlockHound/blob/2071ca8dbb06eba9ffa65596f327e3182d3b9730/junit-platform/src/main/java/reactor/blockhound/junit/platform/BlockHoundTestExecutionListener.java), which uses `@AutoService` to automatically register with JUnit 5's ServiceLoader support:

https://junit.org/junit5/docs/current/user-guide/#launcher-api-listeners-custom

## License

### DIFF
--- a/services/s3/pom.xml
+++ b/services/s3/pom.xml
@@ -42,6 +42,15 @@
                     </archive>
                 </configuration>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-dependency-plugin</artifactId>
+                <configuration>
+                    <ignoredUsedUndeclaredDependencies>
+                        <ignoredUnusedDeclaredDependency>org.junit.jupiter:*</ignoredUnusedDeclaredDependency>
+                    </ignoredUsedUndeclaredDependencies>
+                </configuration>
+            </plugin>
         </plugins>
     </build>
 
@@ -115,6 +124,11 @@
             <groupId>software.amazon.awssdk</groupId>
             <artifactId>auth-crt</artifactId>
             <version>${awsjavasdk.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.netty</groupId>
+            <artifactId>netty-transport</artifactId>
             <scope>test</scope>
         </dependency>
     </dependencies>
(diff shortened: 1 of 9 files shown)


services/s3/pom.xml
```diff
@@ -42,6 +42,15 @@
                     </archive>
                 </configuration>
             </plugin>
+            <plugin>
```
joviegas, line 45: Should we add this in Kinesis service ?I see that in kinesis we have integration test related to a async clients .

services/s3/src/it/java/software/amazon/awssdk/services/BlockHoundInstalledTest.java
```diff
@@ -0,0 +1,132 @@
...
+                         latch.countDown();
+                         ctx.fireChannelActive();
+                     }
+                 });
+
+        int port = getUnusedPort();
+        serverChannel = bootstrap.bind(port).sync().channel();
+        clientSocket = new Socket("localhost", port);
+
+        latch.await(5, TimeUnit.SECONDS);
```
joviegas, line 117: Just curious to know in general if addition of this will cause impact on tests execution time , like will total time taken to execute the build with test remain same or any impacts?

joviegas (approved):

services/s3/src/it/java/software/amazon/awssdk/services/BlockHoundInstalledTest.java
Bennett-Lynch: Do you mean the latency introduced by BlockHound in general? It's a valid concern. This PR offers some benchmarks: https://github.com/reactor/BlockHound/pull/46. Overall the overhead appears to be trivial and should not be a significant concern for tests.

services/s3/pom.xml
Bennett-Lynch: I'm not sure if it matters terribly much since all services need to support async clients. As the Javadoc mentions, it's admittedly somewhat arbitrary, as long as we have it somewhere. I'm fine with either location.

joviegas: > As the Javadoc mentions, it's admittedly somewhat arbitrary, as long as we have it somewhere
Looks fine, then I think no need to move to kinesis.

PR #2877, opened 2021-11-30
Handle exceptions thrown by the consumer passed to SdkPublisher.subscribe or filter.

millems:
Previously, the future returned by `subscribe` or `filter` would never be completed in this scenario. Now, the result future is completed exceptionally with the exception raised by the consumer.


utils/src/main/java/software/amazon/awssdk/utils/async/FilteringSubscriber.java
```diff
@@ -40,11 +40,17 @@ public void onSubscribe(Subscription subscription) {
...
+        try {
+            if (predicate.test(t)) {
+                subscriber.onNext(t);
+            } else {
+                // Consumed a demand but didn't deliver. Request other to make up for it
+                subscription.request(1);
+            }
+        } catch (RuntimeException e) {
+            // Handle the predicate throwing an exception
+            subscription.cancel();
```
joviegas, line 52: Query just for my curiosity : Earlier these exceptions were thrown higher in stack and now we are catching it , do we need to worry about any backward compatibility issue were Users have expecting this exceptions to be caught directly.

joviegas (approved):

utils/src/main/java/software/amazon/awssdk/utils/async/FilteringSubscriber.java
millems: It's against "the rules" to throw out of an onNext (except for null parameters), so we're out of compliance already here. It's possible someone is relying on the old behavior: https://xkcd.com/1172/

joviegas: Thanks Matt.

PR #2879, opened 2021-12-02
Fix to correctly set TokenBucket capacity when requested amount greater than capacity

joviegas:
## Motivation and Context

When the requested number of tokens is greater than the current available tokens, the code sets the capacity to `0` and then sleeps the required amount of time to accumulate the remaining amount; however that amount slept is not reflected in the last recorded refill time of the bucket. Refilling the bucket is time based, and each call to `acquire()` refills the bucket, using the difference between the current time and the last recorded refill time. In this case, the next call to `refill()` will the bucket with more tokens than intended.

## Description
Rather than setting the amount fo `0`, we should just let it go into the negative.

```
_TokenBucketAcquire(amount)
  # Client side throttling is not enabled until we see a throttling error.
  if not enabled
    return

  _TokenBucketRefill()
  # Next see if we have enough capacity for the requested amount.
  if amount <= current_capacity
    current_capacity = current_capacity - amount
  else
    sleep((amount - current_capacity) / fill_rate)
    current_capacity = current_capacity - amount
  return

```

## Testing

- Added Junit

## Types of changes

## Checklist

## License

### DIFF
--- a/core/sdk-core/src/test/java/software/amazon/awssdk/core/internal/retry/RateLimitingTokenBucketTest.java
+++ b/core/sdk-core/src/test/java/software/amazon/awssdk/core/internal/retry/RateLimitingTokenBucketTest.java
@@ -15,6 +15,7 @@
 
 package software.amazon.awssdk.core.internal.retry;
 
+import static org.assertj.core.api.Assertions.as;
 import static org.assertj.core.api.Assertions.assertThat;
 
 import java.time.Duration;
@@ -66,6 +67,8 @@ public class RateLimitingTokenBucketTest {
         long elapsed = System.nanoTime() - a;
 
         assertThat(acquired).isTrue();
+        assertThat(tb.getCurrentCapacity()).isNegative();
+        assertThat(tb.getCurrentCapacity()).isEqualTo(-1.0);
         assertThat(Duration.ofNanos(elapsed).getSeconds()).isEqualTo(1);
     }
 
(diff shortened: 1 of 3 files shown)


core/sdk-core/src/test/java/software/amazon/awssdk/core/internal/retry/RateLimitingTokenBucketTest.java
```diff
@@ -96,6 +96,17 @@ public void tryAcquireCapacity_capacitySufficient_returns0() {
         tb.setCurrentCapacity(5.0);
 
         assertThat(tb.tryAcquireCapacity(5.0)).isZero();
+        assertThat(tb.getCurrentCapacity()).isZero();
+    }
+
+    @Test
+    public void tryAcquireCapacity_amountGreaterThanCapacity_returnsNonZero() {
```
dagnir, line 103: Can you add a test where the `currentCapacity` is negative when `acquire()` is called as well?

joviegas: done

core/sdk-core/src/test/java/software/amazon/awssdk/core/internal/retry/RateLimitingTokenBucketTest.java
```diff
@@ -96,6 +99,62 @@ public void tryAcquireCapacity_capacitySufficient_returns0() {
...
+        Mockito.doAnswer(invocationOnMock -> null).when(tb).refill();
+
+        tb.setFillRate(1.0);
+        tb.setCurrentCapacity(1.0);
+        tb.enable();
+
+        // 1 token to wait for at a rate of 1 per second should sleep for approx 1s
+        long a = System.nanoTime();
+        boolean acquired = tb.acquire(3.0);
+        long elapsed = System.nanoTime() - a;
```
dagnir, line 130: Can you just mock the `RateLimitingTokenBucket.sleep()` method so we don't actually need to wait 3 seconds?

dagnir (approved):

dagnir (approved):

PR #2882, opened 2021-12-02
[netty-nio-client] Ensure initial channel used for protocol detection is released before re-acquiring

Bennett-Lynch:
HttpOrHttp2ChannelPool establishes an initial connection to determine if H1 or H2 should be used for the ChannelPool. It then releases the Channel to the Pool to allow it to be re-acquired and reused. However, in some (perhaps rare) circumstances, it's possible for a subsequent acquire to race against the release, causing the second acquire to establish a new channel if the release had not yet completed. This may add unnecessary delay and incur unnecessary channel overhead. Delay because the release is likely to complete in sub-millisecond time while a new connection handshake may take many milliseconds to complete. Overhead because a second channel may linger in the pool even if the user only ever uses one channel at a time.

## License


http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/utils/NettyUtils.java
```diff
@@ -221,4 +222,48 @@ private static void configureSslEngine(SSLEngine sslEngine) {
...
+     * <p>
+     * Note that if you do not need the value returned by a successful completion (or if it returns {@link Void}) you may use
+     * {@link #runOrPropagate(Promise, Runnable)} instead.
+     *
+     * @param destination the Promise to notify upon failure or cancellation
+     * @param onSuccess   the Consumer to invoke upon success
+     */
+    public static <T> GenericFutureListener<Future<T>> consumeOrPropagate(Promise<?> destination, Consumer<T> onSuccess) {
+        return f -> {
+            if (f.isSuccess()) {
```
joviegas, line 239: > it's possible for a subsequent acquire to race against the release

Is it possible to write Junits to force this and test this scenarion ?

Bennett-Lynch: Race conditions like this are difficult to unit-test, but let me add a unit test to ensure we await the release success.

joviegas (approved):

PR #2886, opened 2021-12-03
Update lambda service integration test to wait for function to be Active after createFunction

joviegas:
## Motivation and Context

- Lambda released a backwards incompatible feature, function states https://aws.amazon.com/blogs/compute/coming-soon-expansion-of-aws-lambda-states-to-all-functions/. 
- The current integration test did not wait for Function to go Active after Function is created.

## Description

- Added a waiter to wait until function is active.

## Testing

- tested in Integ account.

## Types of changes

- [ x]Integration Test update

## Checklist

## License


services/lambda/src/it/java/software/amazon/awssdk/services/lambda/ServiceIntegrationTest.java
```diff
@@ -81,6 +84,9 @@ public void uploadFunction() throws IOException {
                                                                     .timeout(10)
                                                                     .role(lambdaServiceRoleArn)).join();
 
+        lambda.waiter()
```
zoewangg, line 87: I wonder if we should make it a static variable and use `@BeforeClass` instead because creating a function for each individual test can prolong the test time

joviegas: done

services/lambda/src/it/java/software/amazon/awssdk/services/lambda/ServiceIntegrationTest.java
```diff
@@ -81,11 +81,14 @@ public void uploadFunction() throws IOException {
...
+              .waitUntilFunctionActive(r -> r.functionName(FUNCTION_NAME))
+              .whenComplete((r, t) -> {}).join();
```
zoewangg, line 86: nit: It seems `.whenComplete((r, t) -> {})` is not needed

zoewangg (approved):

PR #2891, opened 2021-12-06
[UrlConnectionHttpClient] propagate IOExceptions during request body sending (to benefit from the aws sdk retry policy)

lestephane:
Sending the request body may fail due to an IOException when the server
has closed the underlying socket *without* a preliminary hint such as
a `Connection: close` response header.

This closure comes in the form of a server-initiated TLS alert
(close_notify). The cause of these intermittent non-orderly closures is
unclear. It could be based on the volume of data sent over the socket,
or the count of requests submitted. I suspect AWS uses an ELB fronting
an API Gateway on the endpoint side, and those decisions are made based
on some policy that the client has no control over, therefore there 
closures will happen, and the requests need to be retried when that
happens.

However, When this does, the invokeSafely() wraps the IOException into an
UncheckedIOException which makes the retriable error non-retriable,
because UncheckedIOException extends RuntimeException, which to the AWS
SDK looks like a nondescript error condition that could / should not be
retried. 

This Bugfix propagates the UncheckedIOException's inner IOException cause
so that the UrlConnectionHttpClient may benefit from the same retry policy
as the ApacheHttpClient for request body sending.

## Motivation and Context

The UrlConnectionClient is not working as reliably as the ApacheClient
if it cannot benefit from the same retry capability.

The Pull Request https://github.com/aws/aws-sdk-java-v2/pull/2848
improved the situation by allowing the use of the UrlConnectionClient
for requests with very large bodies. But this won't help if the client
cannot handle occasionally closed connections.

## Description

Catch UncheckedIOException, and rethrow the IOException cause.

## Testing

It is not possible to have the AWS endpoint close a connection in a
deterministic way.

I made the change in a private project, and was able to see such
IOExceptions be considered retriable, and be retried through the SDK
retry policy.

## Types of changes

## Checklist

  -> ran `install -pl :bom-internal,:s3 -P quick --am -T 2C`
  -> can't find the code style information

## License

### DIFF
--- a/http-clients/url-connection-client/src/main/java/software/amazon/awssdk/http/urlconnection/UrlConnectionHttpClient.java
+++ b/http-clients/url-connection-client/src/main/java/software/amazon/awssdk/http/urlconnection/UrlConnectionHttpClient.java
@@ -23,6 +23,7 @@ import static software.amazon.awssdk.utils.NumericUtils.saturatedCast;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.UncheckedIOException;
 import java.net.HttpURLConnection;
 import java.net.URI;
 import java.security.KeyManagementException;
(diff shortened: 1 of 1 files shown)


lestephane:
Closing until I have a test case for the abrupt-closure-from-the-server-side failure mode.

Bennett-Lynch:
Hi @lestephane. Instead of trying to propagate an `IOException`, what do you think about just adding `UncheckedIOException` to the list of the SDK's default retryable exceptions?

Bennett-Lynch:
Actually, the default retry policy should retry on `UncheckedIOException`s because `RetryOnExceptionsCondition` checks the cause of a given exception by default, and `UncheckedIOException` always wraps an `IOException`, which is part of our default retryable exception list.

https://github.com/aws/aws-sdk-java-v2/blob/dbe8b17b7f8e16dc5638019c2275684c11a703df/core/sdk-core/src/main/java/software/amazon/awssdk/core/retry/conditions/SdkRetryCondition.java#L30

https://github.com/aws/aws-sdk-java-v2/blob/dbe8b17b7f8e16dc5638019c2275684c11a703df/core/sdk-core/src/main/java/software/amazon/awssdk/core/retry/conditions/RetryOnExceptionsCondition.java#L59-L60

https://github.com/aws/aws-sdk-java-v2/blob/dbe8b17b7f8e16dc5638019c2275684c11a703df/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/retry/SdkDefaultRetrySetting.java#L80

PR #2892, opened 2021-12-07
Log channel ID when logging error

dagnir:
## Motivation and Context
Better logging.

## Testing
Added unit tests, ran integration tests.

## Types of changes

## Checklist

## License


http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/NettyRequestExecutor.java
```diff
@@ -297,8 +301,8 @@ private URI endpoint() {
         return context.executeRequest().request().getUri();
     }
 
-    private void handleFailure(Supplier<String> msg, Throwable cause) {
-        log.debug(msg.get(), cause);
+    private void handleFailure(Channel channel, Supplier<String> msgSupplier, Throwable cause) {
+        ChannelLogHelper.debug(log, channel, msgSupplier, cause);
```
Bennett-Lynch, line 305: Are there any mechanisms we can consider to help guarantee that `ChannelLogHelper` is used? E.g., a Checkstyle rule in Netty-specific classes that forbids regular logger calls.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/utils/ChannelLogHelper.java
```diff
@@ -0,0 +1,63 @@
...
+import io.netty.channel.Channel;
+import java.util.function.Supplier;
+import org.slf4j.Logger;
+import software.amazon.awssdk.annotations.SdkInternalApi;
+
+/**
+ * Utility class for logging the channel ID along with the message when within the context of a channel.
+ */
+@SdkInternalApi
+public final class ChannelLogHelper {
```
Bennett-Lynch, line 27: This class LGTM, but we could also consider making it object-based rather than utility-based, so that calls don't need to repeatedly include a `logger` argument, and to reduce the likelihood of using a direct logger call.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/utils/ChannelLogHelper.java
```diff
@@ -0,0 +1,63 @@
...
+    public static void trace(Logger logger, Channel channel, Supplier<String> msgSupplier) {
+        if (!logger.isTraceEnabled()) {
+            return;
+        }
+
+        Supplier<String> finalMessage = prependChannelIdIfAvailable(msgSupplier, channel);
+        logger.trace(finalMessage.get());
+    }
+
+    private static Supplier<String> prependChannelIdIfAvailable(Supplier<String> msgSupplier, Channel channel) {
```
Bennett-Lynch, line 58: Nit: When is it _not_ available?

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/utils/ChannelLogHelper.java
```diff
@@ -0,0 +1,63 @@
...
+        String id = channel.id().asShortText();
+
+        return () -> String.format("[Channel %s] %s", id, msgSupplier.get());
```
Bennett-Lynch, line 61: We may also wish to consider putting the channel ID in the MDC:

http://www.slf4j.org/api/org/slf4j/MDC.html

This would allow users to have more fine grained control in its formatting and whether it's enabled or not. I suppose a downside is that some user's logging configurations may not emit the MDC, when we may wish to be opinionated that the channel ID should always be included.

Bennett-Lynch (approved):

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/utils/ChannelLogHelper.java
zoewangg: Can we just use `channel.toString()`? I think it includes the ip and other stuff that can help troubleshoot.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/NettyRequestExecutor.java
dagnir: Yeah I thought about that, but a module-wide checkstyle rule  wouldn't work since there are lots of logging situations that don't happen in the context of a channel: e.g. failure to allocate a channel. I think if we create a more complete logging facade and allow the `channel` parameter to be `null` then that would work, in exchange for some slight clunkiness.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/utils/ChannelLogHelper.java
dagnir: Good callout, forgot to rename :)

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/utils/ChannelLogHelper.java
Bennett-Lynch: I think `channel.toString()` is quite verbose compared to `id()`. Maybe we could use it for debug/trace only?

dagnir: I agree `toString()` is pretty verbose; using for debug and trace makes sense

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/NettyRequestExecutor.java
Bennett-Lynch: Makes sense. I suppose if we take the object-based approach, we could forbid directly instantiating a normal logger (via Checkstyle), and then add more explicit methods that do not require a channel. Users could still bypass channel ID logging, but it would be a much more explicit decision.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/utils/ChannelLogHelper.java
zoewangg: Can we add IP then? We have a backlog item to add it; might as well add it as part of this PR.

dagnir: Sure

Bennett-Lynch: Logging IP-by-default might make sense at the exception-level, but it may be somewhat verbose for normal log statements. My vote would be to do something like:

```
String id;
if (log.debugEnabled()) {
    id = channel.toString();
} else {
    id = channel.id().asShortText();
}
```

zoewangg: +1

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/NettyRequestExecutor.java
```diff
@@ -130,10 +128,7 @@ public NettyRequestExecutor(RequestContext context) {
                         }
                     });
                 } catch (Throwable exc) {
-                    ChannelLogHelper.warn(log,
-                                          ch,
-                                          () -> "Unable to add a task to cancel the request to channel's EventLoop",
-                                          exc);
+                    log.warn(() -> "Unable to add a task to cancel the request to channel's EventLoop", exc);
```
Bennett-Lynch, line 131: We seem to have lost the `ch` parameter here.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/utils/NettyClientLogger.java
```diff
@@ -0,0 +1,183 @@
...
+import org.slf4j.LoggerFactory;
+import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.annotations.SdkTestInternalApi;
+
+/**
+ * Logger facade similar to {@link software.amazon.awssdk.utils.Logger}, that also includes the Channel ID in the message when
+ * provided.
+ */
+@SdkInternalApi
+public final class NettyClientLogger {
```
Bennett-Lynch, line 30: We have a few inspections that warn about incorrect logger usage. Let's add the FQCN for this logger if possible:

https://github.com/aws/aws-sdk-java-v2/blob/b953c8bd36d2e804a2701d144f25ac65c9197005/.idea/inspectionProfiles/AWS_Java_SDK_2_0.xml#L197-L200

(There are a few more occurrences in that file.)

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/utils/NettyClientLogger.java
```diff
@@ -0,0 +1,183 @@
...
+     * @param channel The channel for this message is being logged
+     * @param msgSupplier Supplier for the log message
+     * @param t The throwable to log
+     */
+    public void warn(Channel channel, Supplier<String> msgSupplier, Throwable t) {
+        if (!delegateLogger.isWarnEnabled()) {
+            return;
+        }
+
+        Supplier<String> finalMessage = prependChannelInfo(msgSupplier, channel);
```
Bennett-Lynch, line 137: Even though this is a warning statement, I think we should use the configured log level to determine the verbosity of the channel string. So that even in warnings, if debug level is enabled, we will use the more verbose form.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/utils/NettyClientLogger.java
```diff
@@ -0,0 +1,183 @@
...
+
+    private static Supplier<String> prependChannelInfo(boolean useShortId, Supplier<String> msgSupplier, Channel channel) {
+        String id;
+        if (useShortId) {
+            id = channel.id().asShortText();
+        } else {
+            id = channel.toString();
+        }
+
+        return () -> String.format("[Channel ID: %s] %s", id, msgSupplier.get());
```
Bennett-Lynch, line 181: Nit: Maybe just: `Channel:`.

Worried "ID" won't be fully accurate for the verbose form. WDYT?

Bennett-Lynch (commented):
We seem to have quite a few other Netty-based classes that could benefit from this logger. E.g., [here](https://github.com/aws/aws-sdk-java-v2/blob/b953c8bd36d2e804a2701d144f25ac65c9197005/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/FutureCancelHandler.java#L56-L56) is one I chose at random. How can we go about making sure this logger is more consistently used?

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/NettyRequestExecutor.java
Bennett-Lynch: Wondering how we can make this less error-prone in the future. Since it's an internal API, should we require an explicit, null `Channel` parameter?

dagnir:
> We seem to have quite a few other Netty-based classes that could benefit from this logger. E.g., here is one I chose at random. How can we go about making sure this logger is more consistently used?

Now that we have switched to full-fledged facade, it would be easier to do that by enforcing the use of this logger only within the Netty module.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/NettyRequestExecutor.java
dagnir: yeah I went back and forth on this; I'm going to err on caution and just remove the override that passes in `null`, and force the caller to do that.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/utils/NettyClientLogger.java
```diff
@@ -85,33 +62,10 @@ public void debug(Channel channel, Supplier<String> msgSupplier, Throwable t) {
             return;
         }
 
-        Supplier<String> finalMessage = prependChannelInfo(true, msgSupplier, channel);
+        Supplier<String> finalMessage = prependChannelInfo(msgSupplier, channel);
         delegateLogger.debug(finalMessage.get(), t);
```
Bennett-Lynch: Nit: Since we always immediately call `get()` here, it might be simpler to just return a `String`.

Bennett-Lynch (approved):

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/utils/NettyClientLogger.java
```diff
@@ -0,0 +1,124 @@
...
+import java.util.function.Supplier;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.annotations.SdkTestInternalApi;
+
+/**
+ * Logger facade similar to {@link software.amazon.awssdk.utils.Logger}, that also includes the Channel ID in the message when
+ * provided.
+ */
```
Bennett-Lynch, line 28: Late nit (sorry, my brain works in phases when doing reviews): We might want to explain why this is important. Namely, that it allows users to correlate logs together, including debug logs.

zoewangg (approved):

Bennett-Lynch (approved):

Issue #2386, opened 2021-04-07
Incorrect s3 object replication status name


abaranovv:
## Describe the bug
There is a wrong completed replication status name in the ReplicationStatus enum. According to [docs ](https://docs.aws.amazon.com/AmazonS3/latest/API/API_HeadObject.html#API_HeadObject_ResponseSyntax)name should be COMPLETED

## Expected Behavior
Status name should be `COMPLETED`

## Current Behavior
status is `COMPLETE`, so `ReplicationStatus.fromValue(String value)` returns `UNKNOWN_TO_SDK_VERSION`

## Steps to Reproduce
```
s3Client.headObject(HeadObjectRequest.builder()
                .bucket(bucketName)
                .key(key)
                .versionId(versionId)
                .build())
.replicationStatus();
```

## Possible Solution
fix status name in [service-2.json](https://github.com/aws/aws-sdk-java-v2/blob/master/services/s3/src/main/resources/codegen-resources/service-2.json#L8417)

## Your Environment

* AWS Java SDK version used: 2.13.49
* JDK version used: 11
* Operating System and version:

debora-ito:
Hi @abaranovv thank you for reaching out, I can see the issue running a quick test in `aws-cli`:
```
> aws s3api head-object --bucket bucket --key key
{
    "AcceptRanges": "bytes",
    "LastModified": "2021-04-07T18:54:04+00:00",
    "ContentLength": 12346,
    "ETag": "\"etag\"",
    "VersionId": "version",
    "ContentType": "image/jpeg",
    "Metadata": {},
    "ReplicationStatus": "COMPLETED"
}
```

 I'll forward the issue to the S3 team. Will update here when I hear back.

debora-ito:
For visibility to other SDKs, I'm moving this to the `aws-sdk` repository.

PR #2894, opened 2021-12-07
Fixed incorrect S3 object replication status name

ybuzko:
Fixed incorrect S3 object replication status name

## Motivation and Context
Solves #2386.

## Description
Corrected the enum value in [services-2.json](https://github.com/aws/aws-sdk-java-v2/blob/master/services/s3/src/main/resources/codegen-resources/service-2.json#L8417).
Created a JUnit test that verifies the response of this API.

## Testing
Unit test does the following:
- creates 2 S3 buckets, source and target
- enables versioning on both (required for replication)
- creates a test IAM role
- creates and assigns policies allowing the test role to apply replication
- applies replication configuration from source bucket to target
- creates a test key in the source bucket
- waits for the replication to complete
- verifies that the replication status matches the spec (is "COMPLETED") and replicationStatus() API returns that value instead of null.

## Types of changes

## Checklist

## License

### DIFF
--- a/services/s3/src/it/java/software/amazon/awssdk/services/s3/ReplicationStandardResponsesIntegrationTest.java
+++ b/services/s3/src/it/java/software/amazon/awssdk/services/s3/ReplicationStandardResponsesIntegrationTest.java
@@ -1,0 +1,179 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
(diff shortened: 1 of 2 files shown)


debora-ito:
@ybuzko thank you for the contribution but we cannot merge it.

The json models must be fixed upstream so every SDK can benefit from the change, not only the Java SDK. 
The models are owned by the service teams, so I'll reach out to S3 again and ask for an update.

Issue #2360, opened 2021-03-31
Blocking call while resolving credentials inside an EKS cluster using DynamoDB async client (Blockhound)


rp199:
## Describe the bug
While testing a spring boot webflux app inside a EKS cluster I've detected a blocking call within AWS libraries thanks to [BlockHound](https://github.com/reactor/BlockHound).

It happens while resolving the credentials right before the dynamodb call using the async client. This happens even when I try to provide a custom aws credentials provider:

```
    @Bean
    fun awsCredentialsProviderChain(): AwsCredentialsProviderChain {
        return AwsCredentialsProviderChain.of(
            ContainerCredentialsProvider.builder().asyncCredentialUpdateEnabled(true)
                .build(),
            InstanceProfileCredentialsProvider.builder()
                .asyncCredentialUpdateEnabled(true)
                .build()
        )
    }
```

My dynamodb client:
```

    @Bean
    fun dynamoDbAsyncClient(awsCredentialsProviderChain: AwsCredentialsProviderChain): DynamoDbAsyncClient =
        DynamoDbAsyncClient
            .builder().credentialsProvider(awsCredentialsProviderChain).build()
```

Locally everything works fine, I think the issue lies inside the _InstanceProfileCredentialsProvider_, the other providers seem to be working fine.

## Expected Behavior
Dynamodb aync client should be non blocking so I'd expect to use it without any blocking cals

## Current Behavior

Blockhounds detects a blocking call.
Stacktrace:

`reactor.blockhound.BlockingOperationError: Blocking call! java.net.SocketOutputStream#socketWrite0
	at java.base/java.net.SocketOutputStream.socketWrite0(Unknown Source)
	Suppressed: reactor.core.publisher.FluxOnAssembly$OnAssemblyException: 
Error has been observed at the following site(s):
	|_ checkpoint ⇢ com.mypackage.MyFilter [DefaultWebFilterChain]
	|_ checkpoint ⇢ org.springframework.boot.actuate.metrics.web.reactive.server.MetricsWebFilter [DefaultWebFilterChain]
	|_ checkpoint ⇢ HTTP GET "/v1/trolleys/1001781417" [ExceptionHandlingWebHandler]
Stack trace:
		at java.base/java.net.SocketOutputStream.socketWrite0(Unknown Source)
		at java.base/java.net.SocketOutputStream.socketWrite(Unknown Source)
		at java.base/java.net.SocketOutputStream.write(Unknown Source)
[...]

Issue #2145, opened 2020-11-13
Blocking call inside of a software.amazon.awssdk.http.nio.netty.internal.BetterSimpleChannelPool.close


lanwen:
Since v2 is claimed to be non-blocking, we still facing a blocking call in `software.amazon.awssdk.http.nio.netty.internal.BetterSimpleChannelPool#close` which could be concidered as illegal in this case.
We use https://github.com/reactor/BlockHound in pair with Reactor in spring as well as localstack in testcontainers in our tests, where this blocking call is popping out.

## Describe the issue
We see in test logs this stacktrace:

```java
2020-11-13 11:15:00.917  WARN 63962 --- [entExecutor-2-1] i.n.util.concurrent.GlobalEventExecutor  : Unexpected exception from the global event executor: 

reactor.blockhound.BlockingOperationError: Blocking call! java.lang.Object#wait
        at java.base/java.lang.Object.wait(Object.java)
        at java.base/java.lang.Object.wait(Object.java:321)
        at io.netty.util.concurrent.DefaultPromise.awaitUninterruptibly(DefaultPromise.java:274)
        at io.netty.channel.DefaultChannelPromise.awaitUninterruptibly(DefaultChannelPromise.java:137)
        at io.netty.channel.DefaultChannelPromise.awaitUninterruptibly(DefaultChannelPromise.java:30)
        at io.netty.channel.pool.SimpleChannelPool.close(SimpleChannelPool.java:402)
        at software.amazon.awssdk.http.nio.netty.internal.BetterSimpleChannelPool.close(BetterSimpleChannelPool.java:38)
        at software.amazon.awssdk.http.nio.netty.internal.HonorCloseOnReleaseChannelPool.close(HonorCloseOnReleaseChannelPool.java:75)
        at software.amazon.awssdk.http.nio.netty.internal.IdleConnectionCountingChannelPool.close(IdleConnectionCountingChannelPool.java:106)
        at software.amazon.awssdk.http.nio.netty.internal.utils.BetterFixedChannelPool.lambda$close0$3(BetterFixedChannelPool.java:391)
        at io.netty.util.concurrent.GlobalEventExecutor$TaskRunner.run(GlobalEventExecutor.java:243)
        at io.netty.util.internal.ThreadExecutorMap$2.run(ThreadExecutorMap.java:74)
        at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
[...]

PR #2897, opened 2021-12-07
Allowlist known blocking methods from BlockHound

Bennett-Lynch:
There are two known occurrences where the SDK may currently block from within a Netty EventLoop:

1. https://github.com/aws/aws-sdk-java-v2/issues/2145
2. https://github.com/aws/aws-sdk-java-v2/issues/2360

Allowing BlockHound to forbid these operations may fail existing integration and stability tests. While we have outstanding issues to fix these items, until they are resolved, we need to allow our existing integration tests to continue to pass. We should explicitly allow-list these methods so that they do not interfere with existing tests and so that we maintain visibility on future regression detection.

## License

### DIFF
--- a/test/service-test-utils/src/main/resources/META-INF/services/reactor.blockhound.integration.BlockHoundIntegration
+++ b/test/service-test-utils/src/main/resources/META-INF/services/reactor.blockhound.integration.BlockHoundIntegration
@@ -13,4 +13,4 @@
 # permissions and limitations under the License.
 #
 
-software.amazon.awssdk.http.nio.netty.internal.SdkBlockHoundIntegration
+software.amazon.awssdk.testutils.service.BlockHoundAllowlist
--- a/.changes/next-release/bugfix-NettyNIOHTTPClient-07d2f87.json
+++ b/.changes/next-release/bugfix-NettyNIOHTTPClient-07d2f87.json
@@ -1,0 +1,6 @@
+{
+    "category": "Netty NIO HTTP Client", 
+    "contributor": "", 
+    "type": "bugfix", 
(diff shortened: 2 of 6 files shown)


test/service-test-utils/src/main/resources/META-INF/services/reactor.blockhound.integration.BlockHoundIntegration
```diff
@@ -0,0 +1,16 @@
...
+#
+#  http://aws.amazon.com/apache2.0
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+#
+
+software.amazon.awssdk.testutils.service.BlockHoundAllowlist
```
zoewangg, line 16: nit, should we move this to test directory?

Bennett-Lynch: I believe it needs to be in this directory to be exported as part of the jar (which we want, so other modules pick up this configuration, namely the individual service IT modules).

https://stackoverflow.com/questions/17531625/how-to-include-a-config-file-in-the-meta-inf-services-folder-of-a-jar-using-ma

zoewangg: Oh this is service-test-utils. NVM then!

zoewangg (approved):

joviegas (approved):

PR #2900, opened 2021-12-08
UpdateExpression API

cenedhryn:
## Motivation and Context
In order to support DynamoDB UpdateExpressions we need a representation of these concepts in the public API. 

https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Expressions.UpdateExpressions.html

## Description
Adding a set of classes to represent UpdateExpression actions, and an UpdateExpression class to contain them. See separate design document for motivation and architecture.

## Testing
Unit tested.

### DIFF
--- a/services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/update/AddUpdateAction.java
+++ b/services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/update/AddUpdateAction.java
@@ -15,9 +15,11 @@
 
 package software.amazon.awssdk.enhanced.dynamodb.update;
 
-import java.util.Collections;
+import static java.util.Collections.unmodifiableMap;
+
 import java.util.HashMap;
 import java.util.Map;
+import software.amazon.awssdk.annotations.SdkPublicApi;
 import software.amazon.awssdk.services.dynamodb.model.AttributeValue;
 import software.amazon.awssdk.utils.Validate;
 
@@ -45,6 +47,7 @@ import software.amazon.awssdk.utils.Validate;
  * }
  * </pre>
  */
+@SdkPublicApi
 public final class AddUpdateAction implements UpdateAction {
 
     private final String path;
@@ -52,11 +55,11 @@ public final class AddUpdateAction implements UpdateAction {
     private final Map<String, String> expressionNames;
     private final Map<String, AttributeValue> expressionValues;
 
-    public AddUpdateAction(Builder builder) {
+    private AddUpdateAction(Builder builder) {
         this.path = Validate.paramNotNull(builder.path, "path");
         this.value = Validate.paramNotNull(builder.value, "value");
-        this.expressionValues = Validate.paramNotNull(builder.expressionValues, "expressionValues");
-        this.expressionNames = builder.expressionNames != null ? builder.expressionNames : new HashMap<>();
+        this.expressionValues = unmodifiableMap(Validate.paramNotNull(builder.expressionValues, "expressionValues"));
+        this.expressionNames = unmodifiableMap(builder.expressionNames != null ? builder.expressionNames : new HashMap<>());
     }
 
     /**
@@ -77,11 +80,11 @@ public final class AddUpdateAction implements UpdateAction {
     }
 
     public Map<String, String> expressionNames() {
-        return Collections.unmodifiableMap(expressionNames);
+        return expressionNames;
     }
 
     public Map<String, AttributeValue> expressionValues() {
-        return Collections.unmodifiableMap(expressionValues);
+        return expressionValues;
     }
 
     @Override
@@ -145,8 +148,11 @@ public final class AddUpdateAction implements UpdateAction {
         }
 
         /**
-         * The 'expression values' token map that maps from value references (expression attribute values) to DynamoDB
-         * AttributeValues. The value reference should always start with ':' (colon).
+         * Sets the 'expression values' token map that maps from value references (expression attribute values) to 
+         * DynamoDB AttributeValues, overriding any existing values. 
+         * The value reference should always start with ':' (colon).
+         *
+         * @see #putExpressionValue(String, AttributeValue)
          */
         public Builder expressionValues(Map<String, AttributeValue> expressionValues) {
             this.expressionValues = expressionValues == null ? null : new HashMap<>(expressionValues);
@@ -156,7 +162,7 @@ public final class AddUpdateAction implements UpdateAction {
         /**
          * Adds a single element to the 'expression values' token map.
          *
-         * @see #expressionValues
+         * @see #expressionValues(Map)
          */
         public Builder putExpressionValue(String key, AttributeValue value) {
             if (this.expressionValues == null) {
@@ -168,12 +174,11 @@ public final class AddUpdateAction implements UpdateAction {
         }
 
         /**
-         * Optional 'expression names' token map, to be used if the attribute references in the path expression are
-         * token ('expression attribute names') prepended with the '#' (pound) sign. It should map from token name
-         * to real attribute name.
-         *
-         * @param expressionNames
-         * @return
+         * Sets the optional 'expression names' token map, overriding any existing values. Use if the attribute 
+         * references in the path expression are token ('expression attribute names') prepended with the 
+         * '#' (pound) sign. It should map from token name to real attribute name.
+         * 
+         * @see #putExpressionName(String, String) 
          */
         public Builder expressionNames(Map<String, String> expressionNames) {
             this.expressionNames = expressionNames == null ? null : new HashMap<>(expressionNames);
@@ -183,7 +188,7 @@ public final class AddUpdateAction implements UpdateAction {
         /**
          * Adds a single element to the optional 'expression names' token map.
          *
-         * @see #expressionNames
+         * @see #expressionNames(Map) 
          */
         public Builder putExpressionName(String key, String value) {
             if (this.expressionNames == null) {
--- a/services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/update/AddUpdateAction.java
+++ b/services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/update/AddUpdateAction.java
@@ -1,0 +1,203 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.enhanced.dynamodb.update;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import software.amazon.awssdk.services.dynamodb.model.AttributeValue;
+import software.amazon.awssdk.utils.Validate;
+
+/**
+ * A representation of a single {@link UpdateExpression} ADD action.
+ * <p>
+ * At a minimum, this action must contain a path string referencing the attribute that should be acted upon and a value string
+ * referencing the value to be added to the attribute. The value should be substituted with tokens using the
+ * ':value_token' syntax and values associated with the token must be explicitly added to the expressionValues map.
+ * Consult the DynamoDB UpdateExpression documentation for details on this action.
+ * <p>
+ * Optionally, attribute names can be substituted with tokens using the '#name_token' syntax. If tokens are used in the
+ * expression then the names associated with those tokens must be explicitly added to the expressionNames map
+ * that is also stored on this object.
+ * <p>
+ * Example:-
+ * <pre>
+ * {@code
+ * AddUpdateAction addAction = AddUpdateAction.builder()
+ *                                            .path("#a")
+ *                                            .value(":b")
+ *                                            .putExpressionName("#a", "attributeA")
+ *                                            .putExpressionValue(":b", myAttributeValue)
+ *                                            .build();
+ * }
+ * </pre>
+ */
+public final class AddUpdateAction implements UpdateAction {
+
+    private final String path;
+    private final String value;
+    private final Map<String, String> expressionNames;
+    private final Map<String, AttributeValue> expressionValues;
+
+    public AddUpdateAction(Builder builder) {
+        this.path = Validate.paramNotNull(builder.path, "path");
+        this.value = Validate.paramNotNull(builder.value, "value");
+        this.expressionValues = Validate.paramNotNull(builder.expressionValues, "expressionValues");
+        this.expressionNames = builder.expressionNames != null ? builder.expressionNames : new HashMap<>();
+    }
+
+    /**
+     * Constructs a new builder for {@link AddUpdateAction}.
+     *
+     * @return a new builder.
+     */
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    public String path() {
+        return path;
+    }
+
+    public String value() {
+        return value;
+    }
+
+    public Map<String, String> expressionNames() {
+        return Collections.unmodifiableMap(expressionNames);
+    }
+
+    public Map<String, AttributeValue> expressionValues() {
+        return Collections.unmodifiableMap(expressionValues);
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+
+        AddUpdateAction that = (AddUpdateAction) o;
+
+        if (path != null ? ! path.equals(that.path) : that.path != null) {
+            return false;
+        }
+        if (value != null ? ! value.equals(that.value) : that.value != null) {
+            return false;
+        }
+        if (expressionValues != null ? ! expressionValues.equals(that.expressionValues) :
+            that.expressionValues != null) {
+            return false;
+        }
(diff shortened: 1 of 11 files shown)


services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/update/UpdateAction.java
```diff
@@ -0,0 +1,32 @@
...
+ * An update action represents a part of an {@link UpdateExpression}
+ */
+public interface UpdateAction {
+
+    // String expression();
+    //
+    // Map<String, String> expressionNames();
+    //
+    // Map<String, AttributeValue> expressionValues();
+
```
millems, line 31: Leftovers

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/update/UpdateExpression.java
```diff
@@ -0,0 +1,234 @@
...
+ * DynamoDb Enhanced will convert the UpdateExpression to a format readable by DynamoDb,
+ * <p>
+ * Examples of adding actions for attributes directly:
+ * <pre>
+ * {@code
+ * RemoveUpdateAction removeAction = ...
+ * List<SetUpdateAction> setActions = ...
+ * UpdateExpression.builder()
+ *                 .addAction(removeAction)
+ *                 .actions(setActions)
```
millems, line 40: Wouldn't this override the removeAction?

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/update/UpdateExpression.java
```diff
@@ -0,0 +1,234 @@
...
+            addActions.add(action);
+            return this;
+        }
+
+        /**
+         * Adds a collection of {@link UpdateAction} of any subtype to the builder. Calling this operation
+         * repeatedly will add the new collection to existing items and not overwrite them.
+         */
+        public Builder actions(Collection<UpdateAction> actions) {
+            addActions(actions);
```
millems, line 194: should this be called addActions? Elsewhere in the SDK, the noun form would override everything.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/update/UpdateExpression.java
```diff
@@ -0,0 +1,234 @@
...
+            if (action instanceof RemoveUpdateAction) {
+                addAction((RemoveUpdateAction) action);
+            } else if (action instanceof SetUpdateAction) {
+                addAction((SetUpdateAction) action);
+            } else if (action instanceof DeleteUpdateAction) {
+                addAction((DeleteUpdateAction) action);
+            } else if (action instanceof AddUpdateAction) {
+                addAction((AddUpdateAction) action);
+            } else {
+                throw new IllegalArgumentException("Do not recognize this UpdateAction");
```
millems, line 230: Nit: specify which action item was given?

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/update/UpdateExpression.java
```diff
@@ -0,0 +1,234 @@
...
+            }
+            addActions.add(action);
+            return this;
+        }
+
+        /**
+         * Adds a collection of {@link UpdateAction} of any subtype to the builder. Calling this operation
+         * repeatedly will add the new collection to existing items and not overwrite them.
+         */
+        public Builder actions(Collection<UpdateAction> actions) {
```
millems, line 193: ? extends UpdateAction?

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/update/UpdateAction.java
cenedhryn: Not good.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/update/UpdateExpression.java
cenedhryn: No, because it's additive. I'm not aware if there is an existing pattern for this behavior (override vs add) but the implementation fell more naturally to adding.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/update/UpdateExpression.java
cenedhryn: Does it? I almost thought it was the other way around. Do you have any examples? I'm sure there's lots but it's hard to come up with a search term. 

Shouldn't we also look at the overall consistency between the action builders vs UpdateExpression?
UpdateExpression has 
- `addAction`
- `actions`

The action builders have
- `putExpressionValue` (and corresponding for expressionNames)
- `expressionValues`

The latter versions are taken from the existing `Expression` class and consistent with that naming. 
If we change `actions` to `addActions` then it's less consistent with `expressionValues`. 

What's really inconsistent is `addAction` vs `putExpressionValue`. Should we change it to `addExpressionValue`?

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/update/UpdateExpression.java
cenedhryn: Yes, I think order does matter, though the code that handles validations/creating final expressions hasn't been written yet, so that would make sense.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/update/AddUpdateAction.java
```diff
@@ -0,0 +1,203 @@
...
+ * </pre>
+ */
+public final class AddUpdateAction implements UpdateAction {
+
+    private final String path;
+    private final String value;
+    private final Map<String, String> expressionNames;
+    private final Map<String, AttributeValue> expressionValues;
+
+    public AddUpdateAction(Builder builder) {
```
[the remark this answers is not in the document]
cenedhryn: Definitely, well spotted!

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/update/UpdateExpression.java
```diff
@@ -0,0 +1,234 @@
...
+        }
+
+        /**
+         * Adds a collection of {@link UpdateAction} of any subtype to the builder. Calling this operation
+         * repeatedly will add the new collection to existing items and not overwrite them.
+         */
+        public Builder actions(UpdateAction... actions) {
+            actions(Arrays.asList(actions));
+            return this;
+        }
```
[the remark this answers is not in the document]
cenedhryn: What do you mean, existing configuration options? While it's not super important, there are quite a few builders within DDB Enhanced that utilize varargs, including the clients and some enhanced request classes.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/update/UpdateExpression.java
cenedhryn: yes

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/update/UpdateExpression.java
cenedhryn: @millems Why would the signature `? extends UpdateAction` be better in this case? We're not modifying the collection/list and in fact we do accept a mix of UpdateAction, so it's almost clearer to use UpdateAction.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/update/AddUpdateAction.java
```diff
@@ -0,0 +1,203 @@
...
+         * A string expression representing the value used in the action. The value must be represented as an
+         * expression attribute value token.
+         */
+        public Builder value(String value) {
+            this.value = value;
+            return this;
+        }
+
+        /**
+         * The 'expression values' token map that maps from value references (expression attribute values) to DynamoDB
```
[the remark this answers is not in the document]
cenedhryn: updated the documentation

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/update/AddUpdateAction.java
```diff
@@ -0,0 +1,203 @@
...
+    public String path() {
+        return path;
+    }
+
+    public String value() {
+        return value;
+    }
+
+    public Map<String, String> expressionNames() {
+        return Collections.unmodifiableMap(expressionNames);
```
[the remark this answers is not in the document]
cenedhryn: Hm, didn't think about that. Good point, updating.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/update/UpdateExpression.java
millems: `add` is for appending to lists/collections
`put` is for upserting to a map
no prefix overrides.

Prior art for add: https://github.com/aws/aws-sdk-java-v2/blob/master/http-client-spi/src/main/java/software/amazon/awssdk/http/SdkHttpRequest.java#L235-L253
Prior art for put: https://github.com/aws/aws-sdk-java-v2/blob/master/core/sdk-core/src/main/java/software/amazon/awssdk/core/client/config/ClientOverrideConfiguration.java#L528-L539

More information: https://github.com/aws/aws-sdk-java-v2/blob/master/docs/design/ClientConfiguration.md

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/update/UpdateExpression.java
millems: See comment below

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/update/UpdateExpression.java
millems: It would be so that a customer can pass in a `Collection<AddUpdateAction>`. They could still pass in a `Collection<UpdateAction>`. If we're read-only, it's proper to use covariant types.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/update/AddUpdateAction.java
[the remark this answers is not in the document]
millems: FWIW, Collections.unmodifiableMap / unmodifiableList / etc. is very very cheap.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/update/UpdateExpression.java
cenedhryn: Thank you. I had forgotten about this design specification, which one should refer to..

Will align the plural method of adding a list of `action` with guidelines.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/update/UpdateExpression.java
cenedhryn: Ah, that's the use case I didn't try. Just feeding the subtypes into the method didn't trigger a difference.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/update/AddUpdateAction.java
```diff
@@ -0,0 +1,208 @@
...
+        result = 31 * result + (value != null ? value.hashCode() : 0);
+        result = 31 * result + (expressionValues != null ? expressionValues.hashCode() : 0);
+        result = 31 * result + (expressionNames != null ? expressionNames.hashCode() : 0);
+        return result;
+    }
+
+    /**
+     * A builder for {@link AddUpdateAction}
+     */
+    public static final class Builder {
```
millems, line 126: Nit: Can we implement CopyableBuilder and ToCopyableBuilder on these classes at some point?

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/update/AddUpdateAction.java
```diff
@@ -0,0 +1,208 @@
...
+
+    private final String path;
+    private final String value;
+    private final Map<String, String> expressionNames;
+    private final Map<String, AttributeValue> expressionValues;
+
+    private AddUpdateAction(Builder builder) {
+        this.path = Validate.paramNotNull(builder.path, "path");
+        this.value = Validate.paramNotNull(builder.value, "value");
+        this.expressionValues = unmodifiableMap(Validate.paramNotNull(builder.expressionValues, "expressionValues"));
```
millems, line 61: Is it true that this must always be non-null?

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/update/AddUpdateAction.java
cenedhryn: Will add to task list.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/update/AddUpdateAction.java
cenedhryn: AFAIK; not sure if you're just referring to expressionValues or all of them. For expressionValues I believe so, but we should test this premise before release. I figured it's better to be restrictive at first.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/update/DeleteUpdateAction.java
```diff
@@ -0,0 +1,208 @@
...
+    private final String path;
+    private final String value;
+    private final Map<String, String> expressionNames;
+    private final Map<String, AttributeValue> expressionValues;
+
+    private DeleteUpdateAction(Builder builder) {
+        this.path = Validate.paramNotNull(builder.path, "path");
+        this.value = Validate.paramNotNull(builder.value, "value");
+        this.expressionValues = unmodifiableMap(Validate.paramNotNull(builder.expressionValues, "expressionValues"));
+        this.expressionNames = unmodifiableMap(builder.expressionNames != null ? builder.expressionNames : new HashMap<>());
```
zoewangg, line 62: Don't we need to wrap it with a new HashMap?

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/update/UpdateExpression.java
zoewangg: Looks like we need to update the code snippet

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/update/UpdateExpression.java
```diff
@@ -0,0 +1,234 @@
...
+     * Constructs a new builder for {@link UpdateExpression}.
+     *
+     * @return a new builder.
+     */
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    public List<RemoveUpdateAction> removeActions() {
+        return Collections.unmodifiableList(removeActions);
```
zoewangg, line 73: I think we should create a new List and wrap it with unmodifiableList in the ctor. Can we do a find and replace to fix all occurrences of list and map?

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/update/UpdateExpression.java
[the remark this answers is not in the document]
zoewangg: I meant the existing configuration options that take a collection. For example, we don't have `Builder#executionInterceptors(ExecutionInterceptor... executionInterceptors)` https://github.com/aws/aws-sdk-java-v2/blob/master/core/sdk-core/src/main/java/software/amazon/awssdk/core/client/config/ClientOverrideConfiguration.java#L536

https://github.com/aws/aws-sdk-java-v2/blob/master/docs/design/ClientConfiguration.md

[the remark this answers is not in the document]
cenedhryn: Discussed offline and decided to keep the varargs signature, since it's prevalent in the enhanced library

zoewangg (approved):

PR #2903, opened 2021-12-09
Add "ConcurrencyAcquireDuration" metric for netty-nio-client

Bennett-Lynch:
The time taken to acquire a new channel from a channel pool can be both
non-trivial and highly variable, depending upon whether a new connection
needs to be established, and depending upon the overhead of new
connection establishment (including TLS handshakes). Due to the high
variability, having an explicit metric can be helpful to give a better
picture of latency sources in impacted requests.

## License

### DIFF
--- a/http-client-spi/src/main/java/software/amazon/awssdk/http/HttpMetric.java
+++ b/http-client-spi/src/main/java/software/amazon/awssdk/http/HttpMetric.java
@@ -15,6 +15,7 @@
 
 package software.amazon.awssdk.http;
 
+import java.time.Duration;
 import software.amazon.awssdk.annotations.SdkPublicApi;
 import software.amazon.awssdk.metrics.MetricCategory;
 import software.amazon.awssdk.metrics.MetricLevel;
@@ -103,6 +104,19 @@ public final class HttpMetric {
      */
     public static final SdkMetric<Integer> HTTP_STATUS_CODE =
         metric("HttpStatusCode", Integer.class, MetricLevel.TRACE);
+
+    /**
+     * The time taken to acquire a channel from the channel pool.
+     *
+     * <p>For HTTP/1 operations, a channel is equivalent to a TCP connection. For HTTP/2 operations, a channel is equivalent to
+     * an HTTP/2 stream channel. For both protocols, the time to acquire a new channel may or may not involve establishing a new
+     * connection, depending on whether an existing channel is available in the pool or not. New connection establishment time may
+     * also include TLS handshake time, if TLS is enabled.
+     *
+     * <p>Note: This metric is currently only supported in 'netty-nio-client'.
+     */
+    public static final SdkMetric<Duration> CHANNEL_ACQUIRE_DURATION =
+        metric("ChannelAcquireDuration", Duration.class, MetricLevel.INFO);
 
     private HttpMetric() {
     }
--- a/http-client-spi/src/main/java/software/amazon/awssdk/http/HttpMetric.java
+++ b/http-client-spi/src/main/java/software/amazon/awssdk/http/HttpMetric.java
@@ -106,17 +106,21 @@ public final class HttpMetric {
         metric("HttpStatusCode", Integer.class, MetricLevel.TRACE);
 
     /**
-     * The time taken to acquire a channel from the channel pool.
+     * The time taken to acquire a channel from the connection pool.
      *
      * <p>For HTTP/1 operations, a channel is equivalent to a TCP connection. For HTTP/2 operations, a channel is equivalent to
-     * an HTTP/2 stream channel. For both protocols, the time to acquire a new channel may or may not involve establishing a new
-     * connection, depending on whether an existing channel is available in the pool or not. New connection establishment time may
-     * also include TLS handshake time, if TLS is enabled.
+     * an HTTP/2 stream channel. For both protocols, the time to acquire a new concurrency permit may include the following:
+     * <ol>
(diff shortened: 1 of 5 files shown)


http-client-spi/src/main/java/software/amazon/awssdk/http/HttpMetric.java
```diff
@@ -104,6 +105,19 @@
...
+     *
+     * <p>For HTTP/1 operations, a channel is equivalent to a TCP connection. For HTTP/2 operations, a channel is equivalent to
+     * an HTTP/2 stream channel. For both protocols, the time to acquire a new channel may or may not involve establishing a new
+     * connection, depending on whether an existing channel is available in the pool or not. New connection establishment time may
+     * also include TLS handshake time, if TLS is enabled.
+     *
+     * <p>Note: This metric is currently only supported in 'netty-nio-client'.
+     */
+    public static final SdkMetric<Duration> CHANNEL_ACQUIRE_DURATION =
+        metric("ChannelAcquireDuration", Duration.class, MetricLevel.INFO);
```
Bennett-Lynch, line 119: Unsure if "ChannelAcquireDuration" or "ConnectionAcquireDuration" would be better. "Channel" may not be familiar terminology for some users, but "connection" may be slightly misleading for HTTP/2.

http-client-spi/src/main/java/software/amazon/awssdk/http/HttpMetric.java
```diff
@@ -104,6 +105,19 @@
...
 
+    /**
+     * The time taken to acquire a channel from the channel pool.
+     *
+     * <p>For HTTP/1 operations, a channel is equivalent to a TCP connection. For HTTP/2 operations, a channel is equivalent to
+     * an HTTP/2 stream channel. For both protocols, the time to acquire a new channel may or may not involve establishing a new
+     * connection, depending on whether an existing channel is available in the pool or not. New connection establishment time may
+     * also include TLS handshake time, if TLS is enabled.
+     *
+     * <p>Note: This metric is currently only supported in 'netty-nio-client'.
```
Bennett-Lynch, line 116: Looking to see how feasible this is to add to Apache's `PoolingHttpClientConnectionManager`.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/NettyRequestMetrics.java
```diff
@@ -73,4 +83,15 @@ private static void writeHttp2RequestMetrics(MetricCollector metricCollector,
         metricCollector.reportMetric(Http2Metric.REMOTE_STREAM_WINDOW_SIZE_IN_BYTES,
                                      http2Connection.remote().flowController().windowSize(stream));
     }
+
+    /**
+     * Measure the time taken for a {@link Future} to complete. Does NOT differentiate between success/failure.
+     */
+    public static void measureTimeTaken(Future<?> future, Consumer<Duration> onDone) {
+        Instant start = Instant.now();
+        future.addListener(f -> {
```
dagnir, line 92: Do we need to check that the future is successful?

http-client-spi/src/main/java/software/amazon/awssdk/http/HttpMetric.java
dagnir: This is a good point. For H2 I could imagine wanting to know both metrics. Is it possible to emit both physical socket acquire time, as well as stream channel acquire time as different metrics?

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/NettyRequestMetrics.java
Bennett-Lynch: My thinking was that we should record this for both success/failure. In the event of a failure, especially one that took a long time (i.e., something close to the acquire timeout), I would want to have visibility into where time was spent during the request.

http-client-spi/src/main/java/software/amazon/awssdk/http/HttpMetric.java
Bennett-Lynch: `ConcurrencyAcquireDuration` sounds more consistent. Will use.

Socket acquire time should be possible. It may be a little complicated since the Netty ChannelPool is responsible for establishing a new connection. Would need to measure the time taken around the lowest-level pool, but only for when a new channel is created. I'll take a look as a follow-up to this PR.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/NettyRequestMetrics.java
dagnir: Ah right, that makes sense.

cenedhryn (commented):
LGTM, but can you make sure all builds are successful?

http-client-spi/src/main/java/software/amazon/awssdk/http/HttpMetric.java
Bennett-Lynch: This looks to be doable but non-trivial. Deferring to follow-up PR.

dagnir (approved):

zoewangg (approved):

cenedhryn (approved):

PR #2906, opened 2021-12-09
Fix @Ignored kinesis integration test to not test service behavior.

millems:
It seems like this test started to fail because Kinesis changed the way merging shards worked - with records not being lost in merging. That makes sense, but we relied on the old behavior. Instead of just updating the assertions, updated the test to test less of the service's behavior without testing much less of the SDK behavior.

### DIFF
--- a/services/kinesis/src/it/java/software/amazon/awssdk/services/kinesis/KinesisIntegrationTests.java
+++ b/services/kinesis/src/it/java/software/amazon/awssdk/services/kinesis/KinesisIntegrationTests.java
@@ -16,6 +16,7 @@
 package software.amazon.awssdk.services.kinesis;
 
 import static org.junit.Assert.assertThat;
+import static org.junit.Assert.assertTrue;
 
 import java.math.BigInteger;
 import java.time.Duration;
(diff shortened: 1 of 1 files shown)


services/kinesis/src/it/java/software/amazon/awssdk/services/kinesis/KinesisIntegrationTests.java
```diff
@@ -133,13 +124,9 @@ public void testKinesisOperations() throws Exception {
...
             Shard shard = shards.get(0);
 
-            // Just to be really sure in case of eventual consistency...
-            Thread.sleep(5000);
-
-            testPuts(streamName, shard);
-
-            // Wait a bit to make sure the records propagate.
-            Thread.sleep(5000);
+            System.out.println("Putting two records...");
```
dagnir, line 127: nit: remove println

dagnir (approved):

services/kinesis/src/it/java/software/amazon/awssdk/services/kinesis/KinesisIntegrationTests.java
millems: There's a lot of printlns. Should I remove them all?

dagnir: Ah that would be great!

millems: Done

dagnir (approved):

PR #2910, opened 2021-12-11
Pure refactoring of UpdateItemOperation tests

cenedhryn:
It's a big class and we'll need to add more tests.

### DIFF
--- a/services-custom/dynamodb-enhanced/src/test/java/software/amazon/awssdk/enhanced/dynamodb/internal/operations/UpdateItemOperationTest.java
+++ b/services-custom/dynamodb-enhanced/src/test/java/software/amazon/awssdk/enhanced/dynamodb/internal/operations/UpdateItemOperationTest.java
@@ -39,6 +39,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.UUID;
+import java.util.function.Consumer;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.mockito.Mock;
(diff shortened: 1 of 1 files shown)


services-custom/dynamodb-enhanced/src/test/java/software/amazon/awssdk/enhanced/dynamodb/internal/operations/UpdateItemOperationTest.java
```diff
@@ -164,34 +157,24 @@ public void generateRequest_withConditionExpression() {
...
-        UpdateItemRequest expectedRequest =
-            baseExpectedRequest.updateExpression("SET " + OTHER_ATTRIBUTE_1_NAME + " = " + OTHER_ATTRIBUTE_1_VALUE +
-                                                 " REMOVE " + OTHER_ATTRIBUTE_2_NAME)
-                               .build();
-
-        UpdateItemRequest request = updateItemOperation.generateRequest(FakeItemWithSort.getTableSchema(),
-                                                                        PRIMARY_CONTEXT,
-                                                                        null);
+        UpdateItemRequest.Builder expectedRequestBuilder = ddbRequestBuilder(ddbKey(item.getId(), item.getSort()));
+        expectedRequestBuilder.expressionAttributeValues(expectedValues);
```
Bennett-Lynch, line 173: Nit: I think it helps readability to chain the builder method calls together, rather than repeatedly reference `expectedRequestBuilder`. Same feedback for other locations.

services-custom/dynamodb-enhanced/src/test/java/software/amazon/awssdk/enhanced/dynamodb/internal/operations/UpdateItemOperationTest.java
```diff
@@ -164,34 +157,24 @@ public void generateRequest_withConditionExpression() {
...
-                                                                .build());
-        Map<String, AttributeValue> expectedKey = new HashMap<>();
-        expectedKey.put("id", AttributeValue.builder().s(item.getId()).build());
-        expectedKey.put("sort", AttributeValue.builder().s(item.getSort()).build());
+            UpdateItemOperation.create(requestFakeItemWithSort(item, b -> b.conditionExpression(CONDITION_EXPRESSION)));
+
+        UpdateItemRequest request = updateItemOperation.generateRequest(FakeItemWithSort.getTableSchema(), PRIMARY_CONTEXT, null);
+
+
+        String expectedUpdateExpression = "SET " + OTHER_ATTRIBUTE_1_NAME + " = " + OTHER_ATTRIBUTE_1_VALUE +
```
Bennett-Lynch, line 165: Nit: As soon as we exceed a single concatenation operator, I usually find `String.format(..)` to be more readable.

Bennett-Lynch (approved):

services-custom/dynamodb-enhanced/src/test/java/software/amazon/awssdk/enhanced/dynamodb/internal/operations/UpdateItemOperationTest.java
cenedhryn: Agreed... I stopped refactoring these types of tests after a while because I'll be adding quite a few similar ones, and I didn't want to end up having to back out of some optimization. I would like to simplify the handling of the updateExpression/expressionNames/expressionMaps/conditionals to a point where it's easy to glance at a test and quickly see the significant bits. Expect to see something in the next PR.

services-custom/dynamodb-enhanced/src/test/java/software/amazon/awssdk/enhanced/dynamodb/internal/operations/UpdateItemOperationTest.java
cenedhryn: Yeah, didn't love this setup either but again, because we're releasing support of the full range of UpdateExpression, there will be a lot of different types of expression strings so it'll be rectified when those are added.

PR #2912, opened 2021-12-11
Add "ConcurrencyAcquireDuration" metric for apache-client

Bennett-Lynch:
The time taken to acquire a new channel from a channel pool can be both non-trivial and highly variable, depending upon whether a new connection needs to be established, and depending upon the overhead of new connection establishment (including TLS handshakes).

The same metric was recently added for netty-nio-client: https://github.com/aws/aws-sdk-java-v2/pull/2903

Ideally we would avoid the use of `ThreadLocal`, but the Apache interfaces make it very difficult to expose a request-level `MetricsCollector` to the connection acquire call. While we can easily wrap a `HttpClientConnectionManager` to include metrics, the `HttpClientConnectionManager`, is declared at the Apache-client-level, meaning we would need to create (not just wrap) a new Apache client instance for every request. Apache offers the `HttpContext` class to help propagate arbitrary attributes, but the context object is not exposed to [HttpClientConnectionManager#requestConnection](https://www.javadoc.io/static/org.apache.httpcomponents/httpclient/4.5.4/org/apache/http/conn/HttpClientConnectionManager.html#requestConnection(org.apache.http.conn.routing.HttpRoute,%20java.lang.Object)) or [ConnectionRequest#get](https://www.javadoc.io/static/org.apache.httpcomponents/httpclient/4.5.4/org/apache/http/conn/ConnectionRequest.html#get(long,%20java.util.concurrent.TimeUnit)). To lease & establish new connection, Apache calls `requestConnection`/`connect`. To lease an already-established connection, Apache calls only `requestConnection`. And only `connect` is given the context object.

## License

### DIFF
--- a/http-clients/apache-client/src/main/java/software/amazon/awssdk/http/apache/internal/conn/ClientConnectionRequestFactory.java
+++ b/http-clients/apache-client/src/main/java/software/amazon/awssdk/http/apache/internal/conn/ClientConnectionRequestFactory.java
@@ -87,8 +87,8 @@ public final class ClientConnectionRequestFactory {
         }
 
         @Override
-        public HttpClientConnection get(long timeout, TimeUnit timeUnit) throws InterruptedException, ExecutionException,
-                                                                                ConnectionPoolTimeoutException {
+        public HttpClientConnection get(long timeout, TimeUnit timeUnit)
+                throws InterruptedException, ExecutionException, ConnectionPoolTimeoutException {
             return delegate.get(timeout, timeUnit);
         }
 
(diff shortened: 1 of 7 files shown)


http-clients/apache-client/src/main/java/software/amazon/awssdk/http/apache/internal/conn/ClientConnectionRequestFactory.java
```diff
@@ -69,22 +80,27 @@ static ConnectionRequest wrap(ConnectionRequest orig) {
...
+                if ("get".equals(method.getName())) {
+                    startGet = Instant.now();
+                }
+                try {
+                    return method.invoke(orig, args);
+                } finally {
+                    if (startGet != null) {
+                        recordGetTime(startGet);
+                    }
+                }
```
millems, line 93: Why do we need to use a dynamic proxy instead of just implementing the ConnectionRequest interface?

Bennett-Lynch: Good point. I don't see any reason we must use a proxy here since this is an interface and not a final class. I was just taking advantage of the existing infrastructure that was seemingly awaiting this metric to be plugged in. :) But let me modify it to just use a simple delegation scheme.

Bennett-Lynch: Updated to use delegation instead. I think the primary reason to use a proxy is that it would automatically support any new interface methods that are added. I don't know how likely that is to happen with these interfaces.

millems (approved):

PR #2932, opened 2021-12-21
Use NettyClientLogger everywhere

dagnir:
## Motivation and Context
Improved logging.

## Description
Using `NettyClientLogger` everywhere in the Netty client helps ensure we log the associated `Channel` details when available and keep that logging uniform throughout.

## Testing
New unit tests.

## Types of changes

## Checklist

## License


http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/NettyNioAsyncHttpClient.java
```diff
@@ -69,7 +68,7 @@
 
     private static final String CLIENT_NAME = "NettyNio";
 
-    private static final Logger log = LoggerFactory.getLogger(NettyNioAsyncHttpClient.class);
+    private static final NettyClientLogger log = NettyClientLogger.getLogger(NettyNioAsyncHttpClient.class);
```
joviegas, line 71: Could you please help me to confirm if there will be no changes in the way the User can enable Debug logs as mentioned in https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/logging-slf4j.html ?
apart from adding 
`log4j.logger.io.netty.handler.logging.LoggingHandler=DEBUG`
If there is a difference in how user can enable the logs now ,can we document this in this PR  ?

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/NettyNioAsyncHttpClient.java
```diff
@@ -55,6 +53,7 @@
 import software.amazon.awssdk.http.nio.netty.internal.SdkChannelPool;
 import software.amazon.awssdk.http.nio.netty.internal.SdkChannelPoolMap;
 import software.amazon.awssdk.http.nio.netty.internal.SharedSdkEventLoopGroup;
+import software.amazon.awssdk.http.nio.netty.internal.utils.NettyClientLogger;
```
joviegas, line 56: Should we also mention this changes in changelogs ?

dagnir: I don't think so, this is an implementation/internal detail

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/NettyNioAsyncHttpClient.java
dagnir: No it doesn't change those details. The logger name is based on the class name given to the `getLogger` method, which stays the same.

joviegas (approved):

dagnir:
Coverage is low because this change updates logging lines that were previously uncovered.

PR #2947, opened 2022-01-04
Migrate Log4j test dependencies from version 1.2.17 to 2.17.1

Bennett-Lynch:
Note: The SDK only depends on log4j as a test-time dependency. Test dependencies aren't transitive and are only present for test class paths. Therefore, this change will have no effect on end users that depend on the SDK (and the SDK has not been exposing users to any of the recent log4j-related CVEs).

Modifications:

1. Migrate Log4j from version `1.2.17` to `2.17.1`. This replaces Maven artifact `log4j`/`log4j` with `org.apache.logging.log4j`/`log4j-api` and `org.apache.logging.log4j`/`log4j-core`.
2. Migrate our logging bridge/binding from `org.slf4j`/`slf4j-log4j12` to `org.apache.logging.log4j`/`log4j-slf4j-impl`.
3. Replace `log4j.properties` files with `log4j2.properties` files (using the new Log4j2 property syntax).
4. Update `LogCaptor` (used for unit testing logging events) to Log4j2.
5. Remove a few explicit and unnecessary calls to Log4j1's `BasicConfigurator.configure()`.

Helpful resources:

1. https://logging.apache.org/log4j/2.x/manual/migration.html
2. https://logging.apache.org/log4j/2.x/manual/configuration.html#Properties
3. https://logging.apache.org/log4j/2.x/log4j-slf4j-impl/

## License


test/test-utils/src/main/java/software/amazon/awssdk/testutils/LogCaptor.java
```diff
@@ -52,13 +56,16 @@
 
 public interface LogCaptor extends SdkAutoCloseable {
```
cenedhryn, line 57: Do we have unit tests that cover this class?

cenedhryn (commented):
Great initiative. Not only does it assist our users in how our dependencies are perceived, it's also an improvement on its own.

test/test-utils/src/main/java/software/amazon/awssdk/testutils/LogCaptor.java
Bennett-Lynch: We have other tests depending on it that I used to test the new behavior, but let me add some dedicated unit tests as well.

cenedhryn (approved):

PR #2957, opened 2022-01-07
Validate that the response content-length header matches what the ser…

millems:
…vice actually returns.

Also update error messaging in the event of service-side connection closes, and add testing to protect these error messages against regression.

### DIFF
--- a/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ChannelDiagnostics.java
+++ b/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ChannelDiagnostics.java
@@ -1,0 +1,50 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.http.nio.netty.internal;
+
+import io.netty.channel.Channel;
+import java.time.Duration;
+import java.time.Instant;
+import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.utils.ToString;
+
+/**
+ * Diagnostic information that may be useful to help with debugging during error scenarios.
+ */
+@SdkInternalApi
+public class ChannelDiagnostics {
+    private final Channel channel;
+    private final Instant channelCreationTime;
+    private int requestCount = 0;
+
+    public ChannelDiagnostics(Channel channel) {
+        this.channel = channel;
+        this.channelCreationTime = Instant.now();
+    }
+
+    public void incrementRequestCount() {
+        ++this.requestCount;
+    }
+
+    @Override
+    public String toString() {
+        return ToString.builder("ChannelDiagnostics")
+                       .add("channel", channel)
+                       .add("channelAge", Duration.between(channelCreationTime, Instant.now()))
+                       .add("requestCount", requestCount)
+                       .build();
+    }
+}
--- a/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/NettyRequestExecutor.java
+++ b/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/NettyRequestExecutor.java
@@ -16,6 +16,7 @@
 package software.amazon.awssdk.http.nio.netty.internal;
 
 import static software.amazon.awssdk.http.HttpMetric.CONCURRENCY_ACQUIRE_DURATION;
+import static software.amazon.awssdk.http.nio.netty.internal.ChannelAttributeKey.CHANNEL_DIAGNOSTICS;
 import static software.amazon.awssdk.http.nio.netty.internal.ChannelAttributeKey.EXECUTE_FUTURE_KEY;
 import static software.amazon.awssdk.http.nio.netty.internal.ChannelAttributeKey.EXECUTION_ID_KEY;
 import static software.amazon.awssdk.http.nio.netty.internal.ChannelAttributeKey.IN_USE;
@@ -23,8 +24,9 @@ import static software.amazon.awssdk.http.nio.netty.internal.ChannelAttributeKey
 import static software.amazon.awssdk.http.nio.netty.internal.ChannelAttributeKey.LAST_HTTP_CONTENT_RECEIVED_KEY;
 import static software.amazon.awssdk.http.nio.netty.internal.ChannelAttributeKey.REQUEST_CONTEXT_KEY;
 import static software.amazon.awssdk.http.nio.netty.internal.ChannelAttributeKey.RESPONSE_COMPLETE_KEY;
+import static software.amazon.awssdk.http.nio.netty.internal.ChannelAttributeKey.RESPONSE_CONTENT_LENGTH;
+import static software.amazon.awssdk.http.nio.netty.internal.ChannelAttributeKey.RESPONSE_DATA_READ;
 import static software.amazon.awssdk.http.nio.netty.internal.NettyRequestMetrics.measureTimeTaken;
-import static software.amazon.awssdk.http.nio.netty.internal.utils.NettyUtils.CLOSED_CHANNEL_MESSAGE;
 
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.Unpooled;
@@ -38,9 +40,7 @@ import io.netty.handler.codec.http.HttpHeaders;
 import io.netty.handler.codec.http.HttpMethod;
 import io.netty.handler.codec.http.HttpRequest;
 import io.netty.handler.codec.http.HttpVersion;
-import io.netty.handler.timeout.ReadTimeoutException;
 import io.netty.handler.timeout.ReadTimeoutHandler;
-import io.netty.handler.timeout.WriteTimeoutException;
 import io.netty.handler.timeout.WriteTimeoutHandler;
 import io.netty.util.concurrent.Future;
 import io.netty.util.concurrent.GenericFutureListener;
@@ -48,12 +48,10 @@ import io.netty.util.concurrent.Promise;
 import java.io.IOException;
 import java.net.URI;
 import java.nio.ByteBuffer;
-import java.nio.channels.ClosedChannelException;
 import java.time.Duration;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.Supplier;
 import org.reactivestreams.Publisher;
@@ -176,9 +174,13 @@ public final class NettyRequestExecutor {
         if (channelFuture.isSuccess()) {
             channel = channelFuture.getNow();
             NettyUtils.doInEventLoop(channel.eventLoop(), () -> {
-                configureChannel();
-                if (tryConfigurePipeline()) {
+                try {
+                    configureChannel();
+                    configurePipeline();
                     makeRequest();
+                } catch (Throwable t) {
+                    closeAndRelease(channel);
+                    handleFailure(channel, () -> "Failed to initiate request to " + endpoint(), t);
                 }
             });
         } else {
@@ -192,10 +194,13 @@ public final class NettyRequestExecutor {
         channel.attr(REQUEST_CONTEXT_KEY).set(context);
         channel.attr(RESPONSE_COMPLETE_KEY).set(false);
         channel.attr(LAST_HTTP_CONTENT_RECEIVED_KEY).set(false);
+        channel.attr(RESPONSE_CONTENT_LENGTH).set(null);
+        channel.attr(RESPONSE_DATA_READ).set(null);
+        channel.attr(CHANNEL_DIAGNOSTICS).get().incrementRequestCount();
         channel.config().setOption(ChannelOption.AUTO_READ, false);
     }
 
-    private boolean tryConfigurePipeline() {
+    private void configurePipeline() throws IOException {
         Protocol protocol = ChannelAttributeKey.getProtocolNow(channel);
         ChannelPipeline pipeline = channel.pipeline();
 
@@ -210,10 +215,7 @@ public final class NettyRequestExecutor {
                 requestAdapter = REQUEST_ADAPTER_HTTP1_1;
                 break;
             default:
-                String errorMsg = "Unknown protocol: " + protocol;
-                closeAndRelease(channel);
-                handleFailure(channel, () -> errorMsg, new RuntimeException(errorMsg));
-                return false;
+                throw new IOException("Unknown protocol: " + protocol);
         }
 
         pipeline.addLast(LastHttpContentHandler.create());
@@ -227,13 +229,8 @@ public final class NettyRequestExecutor {
         // handler (which will monitor for it going inactive from now on).
         // Make sure it's active here, or the request will never complete: https://github.com/aws/aws-sdk-java-v2/issues/1207
         if (!channel.isActive()) {
-            String errorMessage = "Channel was closed before it could be written to.";
-            closeAndRelease(channel);
-            handleFailure(channel, () -> errorMessage, new IOException(errorMessage));
-            return false;
+            throw new IOException(NettyUtils.closedChannelMessage(channel));
         }
-
-        return true;
     }
 
     private void makeRequest() {
@@ -308,78 +305,9 @@ public final class NettyRequestExecutor {
 
     private void handleFailure(Channel channel, Supplier<String> msgSupplier, Throwable cause) {
         log.debug(channel, msgSupplier, cause);
-        cause = decorateException(cause);
+        cause = NettyUtils.decorateException(channel, cause);
         context.handler().onError(cause);
         executeFuture.completeExceptionally(cause);
-    }
-
-    private Throwable decorateException(Throwable originalCause) {
-        if (isAcquireTimeoutException(originalCause)) {
-            return new Throwable(getMessageForAcquireTimeoutException(), originalCause);
-        } else if (isTooManyPendingAcquiresException(originalCause)) {
-            return new Throwable(getMessageForTooManyAcquireOperationsError(), originalCause);
-        } else if (originalCause instanceof ReadTimeoutException) {
-            return new IOException("Read timed out", originalCause);
-        } else if (originalCause instanceof WriteTimeoutException) {
-            return new IOException("Write timed out", originalCause);
-        } else if (originalCause instanceof ClosedChannelException) {
-            return new IOException(CLOSED_CHANNEL_MESSAGE, originalCause);
-        }
-
-        return originalCause;
-    }
-
-    private boolean isAcquireTimeoutException(Throwable originalCause) {
-        String message = originalCause.getMessage();
-        return originalCause instanceof TimeoutException &&
-                message != null &&
-                message.contains("Acquire operation took longer");
-    }
-
-    private boolean isTooManyPendingAcquiresException(Throwable originalCause) {
-        String message = originalCause.getMessage();
-        return originalCause instanceof IllegalStateException &&
-               message != null &&
-               originalCause.getMessage().contains("Too many outstanding acquire operations");
-    }
-
-    private String getMessageForAcquireTimeoutException() {
-        return "Acquire operation took longer than the configured maximum time. This indicates that a request cannot get a "
-                + "connection from the pool within the specified maximum time. This can be due to high request rate.\n"
-
-                + "Consider taking any of the following actions to mitigate the issue: increase max connections, "
-                + "increase acquire timeout, or slowing the request rate.\n"
-
-                + "Increasing the max connections can increase client throughput (unless the network interface is already "
-                + "fully utilized), but can eventually start to hit operation system limitations on the number of file "
-                + "descriptors used by the process. If you already are fully utilizing your network interface or cannot "
-                + "further increase your connection count, increasing the acquire timeout gives extra time for requests to "
-                + "acquire a connection before timing out. If the connections doesn't free up, the subsequent requests "
-                + "will still timeout.\n"
-
-                + "If the above mechanisms are not able to fix the issue, try smoothing out your requests so that large "
-                + "traffic bursts cannot overload the client, being more efficient with the number of times you need to "
-                + "call AWS, or by increasing the number of hosts sending requests.";
-    }
-
-    private String getMessageForTooManyAcquireOperationsError() {
-        return "Maximum pending connection acquisitions exceeded. The request rate is too high for the client to keep up.\n"
-
-                + "Consider taking any of the following actions to mitigate the issue: increase max connections, "
-                + "increase max pending acquire count, decrease pool lease timeout, or slowing the request rate.\n"
-
-                + "Increasing the max connections can increase client throughput (unless the network interface is already "
-                + "fully utilized), but can eventually start to hit operation system limitations on the number of file "
-                + "descriptors used by the process. If you already are fully utilizing your network interface or cannot "
-                + "further increase your connection count, increasing the pending acquire count allows extra requests to be "
-                + "buffered by the client, but can cause additional request latency and higher memory usage. If your request"
-                + " latency or memory usage is already too high, decreasing the lease timeout will allow requests to fail "
-                + "more quickly, reducing the number of pending connection acquisitions, but likely won't decrease the total "
-                + "number of failed requests.\n"
-
-                + "If the above mechanisms are not able to fix the issue, try smoothing out your requests so that large "
-                + "traffic bursts cannot overload the client, being more efficient with the number of times you need to call "
-                + "AWS, or by increasing the number of hosts sending requests.";
     }
 
     /**
--- a/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ResponseHandler.java
+++ b/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ResponseHandler.java
@@ -22,9 +22,10 @@ import static software.amazon.awssdk.http.nio.netty.internal.ChannelAttributeKey
 import static software.amazon.awssdk.http.nio.netty.internal.ChannelAttributeKey.LAST_HTTP_CONTENT_RECEIVED_KEY;
 import static software.amazon.awssdk.http.nio.netty.internal.ChannelAttributeKey.REQUEST_CONTEXT_KEY;
 import static software.amazon.awssdk.http.nio.netty.internal.ChannelAttributeKey.RESPONSE_COMPLETE_KEY;
+import static software.amazon.awssdk.http.nio.netty.internal.ChannelAttributeKey.RESPONSE_CONTENT_LENGTH;
+import static software.amazon.awssdk.http.nio.netty.internal.ChannelAttributeKey.RESPONSE_DATA_READ;
 import static software.amazon.awssdk.http.nio.netty.internal.utils.ExceptionHandlingUtils.tryCatch;
 import static software.amazon.awssdk.http.nio.netty.internal.utils.ExceptionHandlingUtils.tryCatchFinally;
-import static software.amazon.awssdk.http.nio.netty.internal.utils.NettyUtils.CLOSED_CHANNEL_MESSAGE;
 
 import io.netty.buffer.ByteBuf;
 import io.netty.channel.Channel;
@@ -33,12 +34,11 @@ import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.SimpleChannelInboundHandler;
 import io.netty.handler.codec.http.FullHttpResponse;
 import io.netty.handler.codec.http.HttpContent;
+import io.netty.handler.codec.http.HttpHeaderNames;
 import io.netty.handler.codec.http.HttpHeaders;
 import io.netty.handler.codec.http.HttpObject;
 import io.netty.handler.codec.http.HttpResponse;
 import io.netty.handler.codec.http.HttpUtil;
-import io.netty.handler.timeout.ReadTimeoutException;
-import io.netty.handler.timeout.WriteTimeoutException;
 import io.netty.util.ReferenceCountUtil;
 import java.io.IOException;
 import java.nio.ByteBuffer;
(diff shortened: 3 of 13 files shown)


http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ResponseHandler.java
```diff
@@ -106,14 +109,52 @@ protected void channelRead0(ChannelHandlerContext channelContext, HttpObject msg
...
+        if (length == null) {
+            return null;
+        }
+
+        return Long.parseLong(length);
+    }
+
+    private static Optional<IOException> validateResponseContentLength(ChannelHandlerContext ctx) {
+        Long contentLengthHeader = ctx.channel().attr(RESPONSE_CONTENT_LENGTH).get();
+        Long actualContentLength = ctx.channel().attr(RESPONSE_DATA_READ).get();
```
Bennett-Lynch, line 139: Can we consolidate this logic to a standalone handler, and that handler raises an exception if it detects this issue? The `ResponseHandler` would then propagate any channel exception.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ResponseHandler.java
```diff
@@ -106,14 +109,52 @@ protected void channelRead0(ChannelHandlerContext channelContext, HttpObject msg
...
+        if (actualContentLength == null) {
+            actualContentLength = 0L;
+        }
+
+        if (actualContentLength.equals(contentLengthHeader)) {
+            return Optional.empty();
+        }
+
+        return Optional.of(new IOException("Response had content-length of " + contentLengthHeader + " bytes, but only received "
+                                           + actualContentLength + " before the connection was closed."));
```
Bennett-Lynch, line 154: What if `Transfer-Encoding: chunked` is being used, but we don't receive the final (0 byte) chunk?

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ResponseHandler.java
millems: Successes and errors are reported in multiple places in the netty client. I can see about how that works, but it's harder for me to guarantee that it will work because of how messed up our success/error reporting is.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ResponseHandler.java
Bennett-Lynch: Also what happens if an HTTP/2 channel is closed without an `endOfStream` flag? Does Netty throw an exception?

millems: That sounds like a different edge case we might need to add validation for. That's much rarer, since we rarely use chunked encoding. This particular code only goes into effect when the content-length is specified.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ResponseHandler.java
millems: Looking at it, I can already tell it might require me to rewrite our FullHttpResponse handling, since that currently handles does its own finalizeResponse call in parallel with this publisher to complete the future successfully...

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ResponseHandler.java
millems: Yeah, maybe we also need to test the end-of-stream flag as well.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ResponseHandler.java
Bennett-Lynch: Got it. Out of scope if it requires any significant refactoring. In general I would consider channel attribute usage a code smell and prefer to try to isolate behavior to standalone handlers, but understood if it's not feasible with the current setup.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ChannelDiagnostics.java
```diff
@@ -0,0 +1,45 @@
...
+
+import java.time.Instant;
+import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.utils.ToString;
+
+/**
+ * Diagnostic information that may be useful to help with debugging during error scenarios.
+ */
+@SdkInternalApi
+public class ChannelDiagnostics {
```
Bennett-Lynch, line 26: This is an interesting way to expose more helpful information. I like the idea. However, if it grows in size over time, we may want to consider only emitting it if, e.g., DEBUG logging is enabled.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ResponseHandler.java
```diff
@@ -106,14 +109,52 @@ protected void channelRead0(ChannelHandlerContext channelContext, HttpObject msg
 
                 fullContent = ((FullHttpResponse) msg).content();
                 ByteBuffer bb = copyToByteBuffer(fullContent);
-                requestContext.handler().onStream(new FullResponseContentPublisher(channelContext, bb, ef));
-                finalizeResponse(requestContext, channelContext);
+                requestContext.handler().onStream(new DataCountingPublisher(channelContext,
+                                                                            new FullResponseContentPublisher(channelContext,
+                                                                                                             bb, ef)));
+
+                Optional<IOException> contentLengthFailure = validateResponseContentLength(channelContext);
```
Bennett-Lynch, line 116: Nit: I think try/catch may be more conventional than `Optional` here.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ResponseHandler.java
```diff
@@ -106,14 +109,52 @@ protected void channelRead0(ChannelHandlerContext channelContext, HttpObject msg
...
+                    finalizeResponse(requestContext, channelContext);
+                }
             } finally {
                 Optional.ofNullable(fullContent).ifPresent(ByteBuf::release);
             }
         }
     }
 
+    private Long responseContentLength(HttpResponse response) {
+        String length = response.headers().get("content-length");
```
Bennett-Lynch, line 129: Nit: `io.netty.handler.codec.http.HttpHeaderNames.CONTENT_LENGTH`

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ResponseHandler.java
Bennett-Lynch, line 154: nit: "but only received X **bytes** before"

nit: should we emit a different/special message if we received 0 bytes?

nit: should we wrap the `IOException` inside of Netty's `PrematureChannelClosureException`?

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ResponseHandler.java
millems: That's good feedback. I'll see what I can do...

millems: Yeah, it seems like FullResponseContentPublisher and PublisherAdapter do their own error handling, so I can't decorate them with something else and expect it to do anything. The onError just gets fired out to the customer's publisher and the netty client ignores it. I'd have to break up those two publishers to put their error handling downstream from the publishers themselves. It would be good to do that, but it's a significant refactor.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ChannelDiagnostics.java
millems: Yeah, I agree we need to be careful about it.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ResponseHandler.java
millems: Changed

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ResponseHandler.java
millems: Fixed

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ResponseHandler.java
millems: It would need to be the other way around: IOException on the outside, so it's seen a retryable. PrematureChannelClosureException is a RuntimeException. Does that change your mind on it? Should I use that inside?

Bennett-Lynch: Doesn't our default retry logic check 1-level of causes?

What happens if Netty does actually throw a `PrematureChannelClosureException` itself? E.g., [here](https://github.com/netty/netty/blob/f1eb920467ebad658fc0a108f73a98017fa28612/codec-http/src/main/java/io/netty/handler/codec/http/HttpObjectDecoder.java#L482). Do we retry on that? It seems like we would want to.

millems: It does check 1 level of causes, but I was under the impression that the contract for `ExecutableHttpRequest.call` is that only IOException will be retried, not necessarily one level of exception. I don't know if I have the wrong impression, or if there's further wrapping that happens upstream before the retry logic that necessitates checking 1 level deeper.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/NettyRequestExecutor.java
```diff
@@ -176,9 +174,13 @@ private void makeRequestListener(Future<Channel> channelFuture) {
         if (channelFuture.isSuccess()) {
             channel = channelFuture.getNow();
             NettyUtils.doInEventLoop(channel.eventLoop(), () -> {
-                configureChannel();
-                if (tryConfigurePipeline()) {
-                    makeRequest();
+                try {
+                    configureChannel();
+                    if (tryConfigurePipeline()) {
```
Bennett-Lynch, line 179: Should we make `tryConfigurePipeline` throw directly? This would wrap all of its exceptions with `Failed to initiate request to...`. Or do you want to reserve that for less explicit failure cases?

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ChannelDiagnostics.java
```diff
@@ -0,0 +1,45 @@
...
+    public void setChannelCreationTime(Instant channelCreationTime) {
+        this.channelCreationTime = channelCreationTime;
+    }
+
+    public void incrementRequestCount() {
+        ++this.requestCount;
+    }
+
+    @Override
+    public String toString() {
```
Bennett-Lynch, line 39: Should we include the `Channel` itself here? Primarily for remote address information.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/utils/NettyUtils.java
```diff
@@ -44,19 +53,106 @@
...
     private static final Logger log = Logger.loggerFor(NettyUtils.class);
 
     private NettyUtils() {
     }
 
+    public static Throwable decorateException(Channel channel, Throwable originalCause) {
+        if (isAcquireTimeoutException(originalCause)) {
+            return new Throwable(getMessageForAcquireTimeoutException(), originalCause);
+        } else if (isTooManyPendingAcquiresException(originalCause)) {
+            return new Throwable(getMessageForTooManyAcquireOperationsError(), originalCause);
```
Bennett-Lynch, line 65: Unrelated to this PR specifically but I don't think we should instantiate a literal `Throwable` here. Would be more conventional to use `Exception` or `RuntimeException` for the generic case.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ResponseHandler.java
```diff
@@ -106,14 +110,52 @@ protected void channelRead0(ChannelHandlerContext channelContext, HttpObject msg
...
                 ByteBuffer bb = copyToByteBuffer(fullContent);
-                requestContext.handler().onStream(new FullResponseContentPublisher(channelContext, bb, ef));
-                finalizeResponse(requestContext, channelContext);
+                requestContext.handler().onStream(new DataCountingPublisher(channelContext,
+                                                                            new FullResponseContentPublisher(channelContext,
+                                                                                                             bb, ef)));
+
+                try {
+                    validateResponseContentLength(channelContext);
+                    finalizeResponse(requestContext, channelContext);
```
Bennett-Lynch, line 119: If `validateResponseContentLength` throws we do not call `finalizeResponse`. Likewise for streaming. Do we still release the channel in that scenario?

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ResponseHandler.java
```diff
@@ -382,28 +435,59 @@ public void cancel() {
...
             runAndLogError(handlerCtx.channel(), "Fail to execute SdkAsyncHttpResponseHandler#onError",
                            () -> requestCtx.handler().onError(err));
             executeFuture(handlerCtx).completeExceptionally(err);
             runAndLogError(handlerCtx.channel(), "Could not release channel", () -> closeAndRelease(handlerCtx));
         }
     }
+
+    private static final class DataCountingPublisher implements Publisher<ByteBuffer> {
+        private final ChannelHandlerContext ctx;
+        private final Publisher<ByteBuffer> delegate;
```
Bennett-Lynch, line 455: Nit: Maybe we could consider trying to reuse [NotifyingPublisher](https://github.com/aws/aws-sdk-java-v2/blob/738f53b7e30f65080c5194d9e879e5235b5e85d5/services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/progress/NotifyingAsyncResponseTransformer.java#L69-L118) here.

We could create a publisher-specific listener interface, and `AsyncResponseTransformerListener` would then extend that, so this code would only depend on the publisher-specific listener.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/MultiplexedChannelRecord.java
```diff
@@ -111,6 +112,7 @@ void acquireClaimedStream(Promise<Channel> promise) {
                 Http2StreamChannel channel = future.getNow();
                 channel.pipeline().addLast(UnusedChannelExceptionHandler.getInstance());
                 channel.attr(ChannelAttributeKey.HTTP2_FRAME_STREAM).set(channel.stream());
+                channel.attr(ChannelAttributeKey.CHANNEL_DIAGNOSTICS).set(new ChannelDiagnostics());
```
Bennett-Lynch, line 115: Is it intentional to not initialize channel creation time here, like we do in `ChannelPipelineInitializer`?

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/utils/NettyUtils.java
```diff
@@ -44,19 +53,106 @@
...
-                                                        + "took too long, or the client tried to write on a read-only socket), " +
-                                                        "or by an intermediary party (e.g. because the channel was idle for too"
-                                                        + " long).";
-
     private static final Logger log = Logger.loggerFor(NettyUtils.class);
 
     private NettyUtils() {
     }
 
+    public static Throwable decorateException(Channel channel, Throwable originalCause) {
```
Bennett-Lynch, line 61: Nit: We almost have enough exception-related logic in this class to consider a `NettyExceptionUtils` or similar.

http-clients/netty-nio-client/src/test/java/software/amazon/awssdk/http/nio/netty/fault/ServerConnectivityErrorMessageTest.java
```diff
@@ -0,0 +1,356 @@
...
+import software.amazon.awssdk.http.SdkHttpResponse;
+import software.amazon.awssdk.http.async.AsyncExecuteRequest;
+import software.amazon.awssdk.http.async.SdkAsyncHttpClient;
+import software.amazon.awssdk.http.async.SdkAsyncHttpResponseHandler;
+import software.amazon.awssdk.http.nio.netty.NettyNioAsyncHttpClient;
+import software.amazon.awssdk.http.nio.netty.SdkEventLoopGroup;
+import software.amazon.awssdk.utils.AttributeMap;
+import software.amazon.awssdk.utils.Logger;
+
+@RunWith(Parameterized.class)
```
Bennett-Lynch, line 83: Nit: Consider JUnit 5.

http-clients/netty-nio-client/src/test/java/software/amazon/awssdk/http/nio/netty/fault/ServerConnectivityErrorMessageTest.java
```diff
@@ -0,0 +1,356 @@
...
+    @Parameterized.Parameters(name = "{0}")
+    public static Collection<TestCase> testCases() {
+        return Arrays.asList(new TestCase(CloseTime.DURING_INIT, "The connection was closed during the request."),
+                             new TestCase(CloseTime.BEFORE_SSL_HANDSHAKE, "The connection was closed during the request."),
+                             new TestCase(CloseTime.DURING_SSL_HANDSHAKE, "The connection was closed during the request."),
+                             new TestCase(CloseTime.BEFORE_REQUEST_PAYLOAD, "The connection was closed during the request."),
+                             new TestCase(CloseTime.DURING_REQUEST_PAYLOAD, "The connection was closed during the request."),
+                             new TestCase(CloseTime.BEFORE_RESPONSE_HEADERS, "The connection was closed during the request."),
+                             new TestCase(CloseTime.BEFORE_RESPONSE_PAYLOAD, "Response had content-length"),
+                             new TestCase(CloseTime.DURING_RESPONSE_PAYLOAD, "Response had content-length"));
```
Bennett-Lynch, line 101: Awesome test cases. Can we come up with more distinctive wording for the first 6 test cases, or out of scope of this PR?

http-clients/netty-nio-client/src/test/java/software/amazon/awssdk/http/nio/netty/fault/ServerConnectivityErrorMessageTest.java
```diff
@@ -0,0 +1,356 @@
...
+        if (netty != null) {
+            netty.close();
+        }
+        netty = null;
+    }
+
+    @Test
+    public void closeTimeHasCorrectMessage() {
+        server.closeTime = testCase.closeTime;
+        assertThat(captureException()).hasMessageContaining(testCase.errorMessage);
```
Bennett-Lynch, line 132: Nit: Consider `assertThatThrownBy`.

http-clients/netty-nio-client/src/test/java/software/amazon/awssdk/http/nio/netty/fault/ServerConnectivityErrorMessageTest.java
```diff
@@ -0,0 +1,356 @@
...
+                                                                                .build())
+                                                     .requestContentPublisher(new EmptyPublisher())
+                                                     .build();
+
+        return netty.execute(req);
+    }
+
+    private static class TestCase {
+        private CloseTime closeTime;
+        private String errorMessage;
```
Bennett-Lynch, line 181: Nit: Maybe `errorSubstring`

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ChannelDiagnostics.java
```diff
@@ -0,0 +1,45 @@
...
+        return ToString.builder("ChannelDiagnostics")
+                       .add("creationTime", channelCreationTime)
```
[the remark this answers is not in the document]
millems: +1 I can remove creationTime and replace it with channelAge. I'd avoid both for what you mention - redundant information.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/NettyRequestExecutor.java
millems: Sorry, I'm not following. 

I added this try/catch because exceptions raised within this runnable go unreported and the completable future returned to the customer never returns (likely causing a deadlock in the customer's code). This change could be in a separate commit, since it's unrelated/unneeded to this change.

I just added it here because I ran into this bug during implementation.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ChannelDiagnostics.java
millems: +1

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/utils/NettyUtils.java
millems: Yeah, not sure why we use throwable

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ResponseHandler.java
millems: Yep, exceptionCaught closes and releases the channel.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ResponseHandler.java
millems: Sorry, I'm not familiar with NotifyingPublisher. What would be the benefit to adding the additional layer of abstraction provide, since Publisher is a well known abstraction? Where would we put that shared code (since AsyncResponseTransformerListener and NotifyingPublisher aren't in netty's dependency closure)?

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/MultiplexedChannelRecord.java
millems: No, and I should probably initialize it at creation time to prevent making that mistake in the future. Good catch.

http-clients/netty-nio-client/src/test/java/software/amazon/awssdk/http/nio/netty/fault/ServerConnectivityErrorMessageTest.java
millems: Do you have an example what you mean?

http-clients/netty-nio-client/src/test/java/software/amazon/awssdk/http/nio/netty/fault/ServerConnectivityErrorMessageTest.java
millems: Are we on junit 5 for sure now? I honestly need to learn it, if so.

http-clients/netty-nio-client/src/test/java/software/amazon/awssdk/http/nio/netty/fault/ServerConnectivityErrorMessageTest.java
millems: I didn't want the assertions running against the Error throwables that could be raised by the logic in captureException.

http-clients/netty-nio-client/src/test/java/software/amazon/awssdk/http/nio/netty/fault/ServerConnectivityErrorMessageTest.java
millems: Sure

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/NettyRequestExecutor.java
Bennett-Lynch: `tryConfigurePipeline` raises several exceptions itself (by also invoking `handleFailure`). I was just wondering if we should wrap all of them with this new logic for the sake of consistency (by changing `tryConfigurePipeline` to simply `throw`). But perhaps that just unnecessarily nests the more pertinent error message.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ResponseHandler.java
Bennett-Lynch: It's been a reoccurring pattern to wrap a publisher/subscriber with a delegating implementation simply to listen in on events. Something like `NotifyingPublisher` would reduce boilerplate code for this common use case and make it less error prone. Just a nit/idea, we can defer this to future refactoring.

http-clients/netty-nio-client/src/test/java/software/amazon/awssdk/http/nio/netty/fault/ServerConnectivityErrorMessageTest.java
Bennett-Lynch: Both 4 and 5 work fine, but IMO we should try to ensure newly created tests are on 5 so that we don't create more future migration work. [JUnit 5 parameterization](https://www.baeldung.com/parameterized-tests-junit-5) is also a lot cleaner IMO.

http-clients/netty-nio-client/src/test/java/software/amazon/awssdk/http/nio/netty/fault/ServerConnectivityErrorMessageTest.java
Bennett-Lynch: Something like this (numbers are not aligned):

1. The connection was closed while it was being initialized.
2. The connection was closed during the TLS handshake process.
3. The connection was closed before it could be written to.
4. The connection was closed while sending the request body.
5. The connection was closed while waiting to receive a response.
6. The connection was closed while receiving the response body.

This type of wording could greatly help debugging certain failure scenarios.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ChannelDiagnostics.java
```diff
@@ -0,0 +1,50 @@
...
+
+    public void incrementRequestCount() {
+        ++this.requestCount;
+    }
+
+    @Override
+    public String toString() {
+        return ToString.builder("ChannelDiagnostics")
+                       .add("channel", channel)
+                       .add("channelAge", Duration.between(channelCreationTime, Instant.now()).getSeconds() + " seconds")
```
Bennett-Lynch, line 46: I would prefer to just use the default `Duration#toString` format, even if it's not the prettiest. Flooring to seconds may lose a lot of granularity.

millems: Fair enough!

millems: Fixed

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/NettyRequestExecutor.java
millems: Ahh, I see. We can combine them, yeah. Good idea.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ResponseHandler.java
millems: SdkPublisher is probably the place to put such a construct, imo. Unfortunately netty wouldn't have access to that, since it's in sdk-core.

http-clients/netty-nio-client/src/test/java/software/amazon/awssdk/http/nio/netty/fault/ServerConnectivityErrorMessageTest.java
millems: Updated it to junit 5. I don't like not being able to use the test parameter in the before-each statement, but otherwise it seems similar to me.

http-clients/netty-nio-client/src/test/java/software/amazon/awssdk/http/nio/netty/fault/ServerConnectivityErrorMessageTest.java
millems: The test case already provides something similar through its toString. I feel like that should be enough without needing to maintain a string description.

Screen Shot 2022-01-19 at 11 47 36 AM

Bennett-Lynch: I meant can we change the user-facing wording used in the actual exception?

millems: Gotcha, we can talk offline. I tried super hard to do that originally (that's why these test cases are structured like that), and it just doesn't work like that. The client thinks it's waiting for a response when the service hasn't even finished the SSL handshake.

Bennett-Lynch (approved):

PR #2960, opened 2022-01-11
Update Apache to not reuse connections if they received a 5xx error.

millems:
This was already the behavior for Netty, and cannot be made the behavior for the URL connection client.

### DIFF
--- a/test/http-client-tests/src/main/java/software/amazon/awssdk/http/SdkHttpClientTestSuite.java
+++ b/test/http-client-tests/src/main/java/software/amazon/awssdk/http/SdkHttpClientTestSuite.java
@@ -145,6 +145,30 @@ public abstract class SdkHttpClientTestSuite {
     }
 
     @Test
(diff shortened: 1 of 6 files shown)


dagnir (approved):

test/http-client-tests/src/main/java/software/amazon/awssdk/http/SdkHttpClientTestSuite.java
```diff
@@ -144,6 +144,30 @@ public void connectionPoolingWorks() throws Exception {
         assertThat(CONNECTION_COUNTER.openedConnections()).isEqualTo(initialOpenedConnections + 1);
     }
 
+    @Test
+    public void connectionsAreNotReusedOn5xxErrors() throws Exception {
```
zoewangg, line 148: Nice!

zoewangg (approved):

PR #2963, opened 2022-01-12
Retry on InternalError error codes.

millems:
This fixes an issue where 200s followed by a failure were not retried for S3.


zoewangg (approved):

core/aws-core/src/main/java/software/amazon/awssdk/awscore/internal/AwsErrorCode.java
```diff
@@ -64,6 +64,7 @@
         retryableErrorCodes.add("PriorRequestNotComplete");
         retryableErrorCodes.add("RequestTimeout");
         retryableErrorCodes.add("RequestTimeoutException");
+        retryableErrorCodes.add("InternalError");
```
Bennett-Lynch, line 67: Nit: Looks like we used to specify initial capacity size here, but that is probably an over-optimization.

Bennett-Lynch (approved):

PR #2982, opened 2022-01-21
Revert the rollback of #2957

millems:
#2957 was rolled back because it did not account for HEAD responses, and broke for S3 integration tests. This fixes that issue: Do not validate content-length for 304 and HEAD responses.

This PR is separated into two different commits so it's easier to ignore the previously-reviewed code and the fixes to that code.


http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ResponseHandler.java
```diff
@@ -155,6 +163,23 @@ private static void validateResponseContentLength(ChannelHandlerContext ctx) thr
...
+            return false;
+        }
+
+        // 304 responses may contain Content-Length without a payload
+        Integer responseStatusCode = ctx.channel().attr(RESPONSE_STATUS_CODE).get();
+        if (responseStatusCode != null && responseStatusCode == HttpResponseStatus.NOT_MODIFIED.code()) {
+            return false;
+        }
+
+        return false;
```
Bennett-Lynch, line 180: Should `return true` -- can we add a test case?

Bennett-Lynch (approved):

PR #2989, opened 2022-01-25
Improve availability of instance profile credentials provider during outages.

millems:
This is done by allowing services to validate credential expiration time during IMDS outages instead of assuming the SDK has access to the latest valid expiration time.

Additional changes:
1. Removed unnecessary intellij inspections that have frequent false positives.
2. Include SDK user-agent in container credential provider calls.
3. Allow specifying the profile file and name used by the instance profile credentials provider.
4. Use the client's profile file and name for instance profile credentials when the default credentials provider is not overridden.
5. Convert the HttpCredentialsProvider protected API to a public API (with some breaking changes), so that there's fewer public classes extending internal classes.


core/auth/src/main/java/software/amazon/awssdk/auth/credentials/InstanceProfileCredentialsProvider.java
```diff
@@ -69,56 +105,91 @@ public static InstanceProfileCredentialsProvider create() {
...
+        if (expiration == null || expiration.isAfter(oneHourFromNow)) {
+            return oneHourFromNow;
+        }
+
+        // If expiration time is within 15 minutes (or in the past), wait 15 minutes and warn the customer that they'll be using
+        // expired credentials.
+        Instant fifteenMinutesFromNow = now.plus(15, MINUTES);
+        if (expiration.isBefore(fifteenMinutesFromNow)) {
+            log.warn(() -> "IMDS credential expiration has been extended due to an IMDS availability outage. A refresh"
+                           + " of these credentials will be attempted again in 15 minutes.");
```
zoewangg, line 167: Should we log warning here? I wonder if this is necessary because there's nothing users can do when they notice the log

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/InstanceProfileCredentialsProvider.java
```diff
@@ -69,56 +105,91 @@ public static InstanceProfileCredentialsProvider create() {
...
-            return HttpResourcesUtils.instance()
-                    .readResource(new TokenEndpointProvider(getImdsEndpoint()), "PUT");
-        } catch (Exception e) {
+            credentials = httpCredentialsLoader.loadCredentials(createEndpointProvider());
+            this.cachedCredentials = credentials;
+        } catch (RuntimeException | IOException e) {
+            credentials = this.cachedCredentials;
+
+            if (credentials != null) {
+                credentials.getExpiration().ifPresent(expiration -> {
```
zoewangg, line 127: Can we add some tests to verify this behavior?

.changes/next-release/bugfix-AWSSDKforJavav2-2a8f703.json
```diff
@@ -0,0 +1,6 @@
+{
```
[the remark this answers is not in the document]
millems: Hah, that's fair. If it's any consolation, all but 1 of these are super tiny.

zoewangg (approved):

millems:
Closing without issues, because this will be merged in the future.

PR #2990, opened 2022-01-25
Merge s3 crt client change back in and fix region

zoewangg:
- Merge s3 crt client change back in
- Fix region issue in 7ae114c82488247a7a2b627dc09db6b644d86deb 
RCA: `S3CrtAsyncHttpClient` relies `SdkHttpRequest#uri` to set host header and since the region was not set on the S3AsyncClient before, the SdkHttpRequest#uri used the region resolved from the default region chain.
- Changed the integ test to use us-west-1 so that this issue can be caught in our integration test in the test account before the release


services-custom/s3-transfer-manager/src/it/java/software/amazon/awssdk/transfer/s3/S3IntegrationTestBase.java
```diff
@@ -43,7 +44,7 @@
  */
 public class S3IntegrationTestBase extends AwsTestBase {
 
-    protected static final Region DEFAULT_REGION = Region.US_WEST_2;
+    protected static final Region DEFAULT_REGION = Region.US_WEST_1;
```
dagnir, line 47: nice!

dagnir (approved):

PR #2993, opened 2022-01-26
[TM DownloadDirectory Part1] Create POJO classes for download directory

zoewangg:
## Modifications
[TM DownloadDirectory Part1] Create POJO classes for download directory. Implementations will be in a separate PR.
Changes will be merged to a feature branch for now once approved.

## Testing
Added unit tests

## Types of changes

## Checklist

## License


services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/CompletedDirectoryDownload.java
```diff
@@ -0,0 +1,142 @@
...
+ */
+@SdkPublicApi
+@SdkPreviewApi
+public final class CompletedDirectoryDownload implements CompletedDirectoryTransfer {
+
+    private final Collection<FailedFileDownload> failedTransfers;
+
+    private CompletedDirectoryDownload(DefaultBuilder builder) {
+        this.failedTransfers = Collections.unmodifiableCollection(
+            Validate.paramNotNull(builder.failedTransfers, "failedTransfers"));
```
millems, line 41: Defensive copy in case of builder reuse?

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/S3TransferManager.java
```diff
@@ -276,6 +276,54 @@ default DirectoryUpload uploadDirectory(Consumer<UploadDirectoryRequest.Builder>
         return uploadDirectory(UploadDirectoryRequest.builder().applyMutation(requestBuilder).build());
     }
 
+    /**
+     * Download all objects under a specific prefix and bucket to the provided directory. By default, all objects in the entire
+     * bucket will be downloaded.
```
zoewangg, line 281: More documentation such as whether we require the directory to pre-exist or not will be added as we start to work on implementation.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/CompletedDirectoryDownload.java
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    @Override
+    public Collection<FailedFileDownload> failedTransfers() {
+        return failedTransfers;
+    }
```
millems, line 47: Discussion, non-blocking: IMO, it's more polite to return the most concrete abstract type, e.g. List instead of Collection here so that people can have constant-time random access to its members. I know that's controversial "because order doesn't matter" here, but I think it's more likely that a user would want to use failedTransfers().get(n) to get the n'th failure than a user would get confused about the order being "inconsistent" just because we returned a List instead of a Collection.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/CompletedDirectoryDownload.java
```diff
@@ -0,0 +1,142 @@
...
+
+/**
+ * Represents a completed download directory transfer to Amazon S3. It can be used to track
+ * failed single file downloads.
+ *
+ * @see S3TransferManager#downloadDirectory(DownloadDirectoryRequest) 
+ */
+@SdkPublicApi
+@SdkPreviewApi
+public final class CompletedDirectoryDownload implements CompletedDirectoryTransfer {
```
millems, line 35: `, ToCopyableBuilder`?

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/CompletedDirectoryDownload.java
```diff
@@ -0,0 +1,142 @@
...
+        return ToString.builder("CompletedDirectoryDownload")
+                       .add("failedTransfers", failedTransfers)
+                       .build();
+    }
+
+    public static Class<? extends Builder> serializableBuilderClass() {
+        return DefaultBuilder.class;
+    }
+
+    public interface Builder {
```
millems, line 86: `extends CopyableBuilder`?

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/DirectoryDownload.java
```diff
@@ -0,0 +1,30 @@
...
+import software.amazon.awssdk.annotations.SdkPublicApi;
+
+/**
+ * A download transfer of a directory of objects from S3
+ */
+@SdkPublicApi
+@SdkPreviewApi
+public interface DirectoryDownload extends DirectoryTransfer {
+    @Override
+    CompletableFuture<CompletedDirectoryDownload> completionFuture();
```
millems, line 29: Nit: Javadoc

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/DownloadDirectoryRequest.java
```diff
@@ -0,0 +1,323 @@
...
+    public DownloadDirectoryRequest(DefaultBuilder builder) {
+        this.destinationDirectory = Validate.paramNotNull(builder.destinationDirectory, "destinationDirectory");
+        this.bucket = Validate.paramNotNull(builder.bucket, "bucket");
+        this.prefix = builder.prefix;
+        this.delimiter = builder.delimiter;
+        this.overrideConfiguration = builder.configuration;
+    }
+
+    /**
+     * The destination directory to download
```
millems, line 55: Nit: The destination directory to which files should be downloaded.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/CompletedDirectoryDownload.java
millems: We should also probably treat failedTransfers == null as being empty. CompletedDirectoryDownload.builder().build() should probably be valid.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/CompletedDirectoryDownload.java
zoewangg: Haha, @Bennett-Lynch and I had the same debate before! We can definitely open the discussion. The main reason I think Collection is better because it gives us flexibility to change the implementation (although it's very unlikely we will change it).
I think customers can still do `failedFileUploads.stream().collect(Collectors.toList())` if they want to iterate the list by index.

millems: If you've already convinced Bennett that this is the correct decision, then I don't want to make you have to debate this again.

Bennett-Lynch: I was pro-`List` as well, and still somewhat am, but was mostly persuaded to use `Collection` since it allows us to use our [ConcurrentLinkedQueue](https://github.com/aws/aws-sdk-java-v2/blob/acda1370233a7f22fb19fb59faec86176f4d5a62/services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/UploadDirectoryHelper.java#L105-L105) directly. I agree that ordering doesn't strictly matter here, but it still seems like a nice-to-have.

zoewangg: @millems Everyone's opinion matters! Let's vote on it. :) 
@Bennett-Lynch We are still copying it to an ArrayList technically.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/DirectoryDownload.java
zoewangg: It has inherited Javadoc, do you think we still need to add more?
```
    /**
     * @return The future that will be completed when this transfer is complete.
     */
```

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/CompletedDirectoryDownload.java
zoewangg: The array is copied in the builder setter

millems: My concern was that this behavior seems possible, unless I'm misreading:

```java
CompletedDirectoryDownload.Builder builder = CompletedDirectoryDownload.builder().addFailedTransfer(foo);
CompletedDirectoryDownload download = builder.build();

System.out.println(download.failedTransfers().size()); // Prints 1
builder.addFailedTransfer(foo);
System.out.println(download.failedTransfers().size()); // Prints 2
```

zoewangg: Ah yeah, good catch

millems (approved):
Just two minor non-blocking comments. Sorry I didn't notice them in the first pass.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/DownloadDirectoryRequest.java
```diff
@@ -0,0 +1,323 @@
...
+        return Optional.ofNullable(delimiter);
+    }
+
+    /**
+     * @return the optional override configuration
+     * @see Builder#overrideConfiguration(DownloadDirectoryOverrideConfiguration)
+     */
+    public Optional<DownloadDirectoryOverrideConfiguration> overrideConfiguration() {
+        return Optional.ofNullable(overrideConfiguration);
+    }
```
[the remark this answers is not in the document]
zoewangg: We currently return `Optional<AwsRequestOverrideConfiguration>` for the request override configuration getter though 

https://github.com/aws/aws-sdk-java-v2/blob/acda1370233a7f22fb19fb59faec86176f4d5a62/core/aws-core/src/main/java/software/amazon/awssdk/awscore/AwsRequest.java#L36-L38

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/DownloadDirectoryOverrideConfiguration.java
```diff
@@ -0,0 +1,179 @@
...
+    private DownloadDirectoryOverrideConfiguration(DefaultBuilder builder) {
+        this.downloadFileRequestTransformer = builder.downloadFileRequestTransformer;
+    }
+
+    /**
+     * @return the optional download request transformer
+     * @see DownloadDirectoryOverrideConfiguration.Builder#downloadFileRequestTransformer(Consumer)
+     */
+    public Optional<Consumer<DownloadFileRequest.Builder>> downloadFileRequestTransformer() {
+        return Optional.ofNullable(downloadFileRequestTransformer);
```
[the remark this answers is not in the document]
zoewangg: +1 We'd need to change it in uploadDirectoryOverrideConfiguration as well. I'll submit a separate PR for this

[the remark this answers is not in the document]
zoewangg: https://github.com/aws/aws-sdk-java-v2/pull/2997

PR #2999, opened 2022-01-28
Delete stale API Gateway REST APIs when running integration tests

Bennett-Lynch:
This test sometimes fails due to accumulating stale REST APIs and eventually exceeding the maximum limit (600). E.g.,

```
[ERROR]   Run 1: ServiceIntegrationTest.createRestApi:60 ? BadRequest This operation would brea...
```

While this test does attempt to clean up after itself, after-class execution may not be run for one reason or another.

Example run:

```
01:55:13.952 [main] INFO software.amazon.awssdk.services.apigateway.ServiceIntegrationTest - Searching for stale REST APIs older than 7 days...
01:55:16.513 [main] INFO software.amazon.awssdk.services.apigateway.ServiceIntegrationTest - Successfully deleted REST API java-sdk-integration-1639773333450 (ftmsvj3rfk) which was 42 days old.
01:55:16.687 [main] INFO software.amazon.awssdk.services.apigateway.ServiceIntegrationTest - Finished searching for stale REST APIs. Successfully deleted: 1. Failed to delete: 0.
```


services/apigateway/src/it/java/software/amazon/awssdk/services/apigateway/ServiceIntegrationTest.java
```diff
@@ -71,6 +77,24 @@ public static void createRestApi() {
...
+    private static void deleteStaleRestApis() {
+        Duration maxAge = Duration.ofDays(7);
+        log.info(() -> String.format("Searching for REST APIs older than %s...", maxAge));
+        apiGateway.getRestApisPaginator().items().forEach(api -> {
+            Duration age = Duration.between(api.createdDate(), Instant.now());
+            if (api.name().startsWith(NAME_PREFIX) && age.compareTo(maxAge) > 0) {
+                log.info(() -> String.format("Deleting REST API %s (%s) which is %s days old",
+                                             api.name(), api.id(), age.toDays()));
+                try {
+                    apiGateway.deleteRestApi(r -> r.restApiId(api.id()));
```
zoewangg, line 90: Should we add some sleep statement here? This will likely only delete one API in my experience...

Bennett-Lynch: Rate-limited deletes to once every 10 seconds

PR #3010, opened 2022-02-02
[TM DownloadDirectory Part2] Implement download directory in transfer manager

zoewangg:
## Modifications
Implement download directory in transfer manager

**Implementation Notes:**

- ListObjects is performed using DFS. See `ListObjectsRecursivelyHelper`
- The SDK sends out individual downloadFileRequest as it iterates the s3 objects returned from ListObjectsV2

~~### Pending Items:~~
~~- Add tests for `ListObjectsRecursivelyHelper`~~
~~- Add more Javadocs~~

## Testing

## Types of changes

## Checklist

## License


services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/DownloadDirectoryHelper.java
```diff
@@ -0,0 +1,225 @@
...
+
+        downloadDirectoryRequest.overrideConfiguration()
+                                .map(DownloadDirectoryOverrideConfiguration::downloadFileRequestTransformer)
+                                .ifPresent(c -> c.accept(requestBuilder));
+        return requestBuilder.build();
+    }
+
+    private static void createParentDirectoriesIfNeeded(Path destinationPath) {
+        Path parentDirectory = destinationPath.getParent();
+        if (parentDirectory != null && !Files.exists(parentDirectory)) {
```
Bennett-Lynch, line 208: Similar feedback here on `Files.exists`.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/ListObjectsRecursivelyHelper.java
```diff
@@ -0,0 +1,110 @@
...
+import software.amazon.awssdk.services.s3.model.S3Object;
+import software.amazon.awssdk.utils.CollectionUtils;
+import software.amazon.awssdk.utils.Logger;
+
+/**
+ * A helper class that returns all objects within a bucket given a {@link ListObjectsV2Request} recursively.
+ */
+@SdkInternalApi
+public class ListObjectsRecursivelyHelper {
+    private static final Logger logger = Logger.loggerFor(ListObjectsRecursivelyHelper.class);
```
Bennett-Lynch, line 41: Other "helpers" use the TM namespace. Should we be consistent here?

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/ListObjectsRecursivelyHelper.java
```diff
@@ -0,0 +1,110 @@
...
+    private final Function<ListObjectsV2Request, CompletableFuture<ListObjectsV2Response>> listObjectsFunction;
+
+    public ListObjectsRecursivelyHelper(Function<ListObjectsV2Request,
+                                        CompletableFuture<ListObjectsV2Response>> listObjectsFunction) {
+
+        this.listObjectsFunction = listObjectsFunction;
+    }
+
+    public SdkPublisher<S3Object> s3Objects(ListObjectsV2Request firstRequest) {
+        Function<ListObjectsV2Response, Iterator<S3Object>> getIterator = response -> {
```
Bennett-Lynch, line 51: Nit: This lambda/anonymous class feels large enough that it should probably be an inner class.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/ListObjectsRecursivelyHelper.java
```diff
@@ -0,0 +1,110 @@
...
+import software.amazon.awssdk.services.s3.model.ListObjectsV2Response;
+import software.amazon.awssdk.services.s3.model.S3Object;
+import software.amazon.awssdk.utils.CollectionUtils;
+import software.amazon.awssdk.utils.Logger;
+
+/**
+ * A helper class that returns all objects within a bucket given a {@link ListObjectsV2Request} recursively.
+ */
+@SdkInternalApi
+public class ListObjectsRecursivelyHelper {
```
Bennett-Lynch, line 40: This class looks like it could be useful as a public API/utility some day.

services-custom/s3-transfer-manager/src/test/java/software/amazon/awssdk/transfer/s3/internal/DownloadDirectoryHelperParameterizedTest.java
```diff
@@ -0,0 +1,177 @@
...
+import software.amazon.awssdk.transfer.s3.DownloadFileRequest;
+import software.amazon.awssdk.transfer.s3.FileDownload;
+import software.amazon.awssdk.transfer.s3.internal.progress.DefaultTransferProgress;
+import software.amazon.awssdk.transfer.s3.internal.progress.DefaultTransferProgressSnapshot;
+import software.amazon.awssdk.utils.IoUtils;
+
+/**
+ * Testing {@link DownloadDirectoryHelper} with different file systems.
+ */
+@RunWith(Parameterized.class)
```
Bennett-Lynch, line 61: Nit: Consider JUnit5 for new tests.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/ListObjectsRecursivelyHelper.java
```diff
@@ -0,0 +1,110 @@
...
+            }
+
+            return Collections.emptyIterator();
+        };
+
+        return PaginatedItemsPublisher.builder().nextPageFetcher(new ListObjectsV2ResponseFetcher(firstRequest))
+                                      .iteratorFunction(getIterator).isLastPage(false).build();
+    }
+
+    private final class ListObjectsV2ResponseFetcher implements AsyncPageFetcher<ListObjectsV2Response> {
```
Bennett-Lynch, line 70: Nice! I might need to stare at this a little bit longer...

services-custom/s3-transfer-manager/src/test/java/software/amazon/awssdk/transfer/s3/internal/DownloadDirectoryHelperParameterizedTest.java
```diff
@@ -0,0 +1,177 @@
...
+    public void tearDown() {
+        if (jimfs != null) {
+            IoUtils.closeQuietly(jimfs, null);
+        } else {
+            FileUtils.cleanUpTestDirectory(directory);
+        }
+    }
+
+    @Test
+    public void downloadDirectory_shouldRecursivelyDownload() {
```
Bennett-Lynch, line 104: Can we assert (either here or in integ tests) that the keys are visited in DFS order?

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/DownloadDirectoryHelper.java
```diff
@@ -0,0 +1,225 @@
...
+                                                            .key(s3Object.key())
+                                                            .build();
+        DownloadFileRequest.Builder requestBuilder = DownloadFileRequest.builder()
+                                                                        .destination(destinationPath)
+                                                                        .getObjectRequest(
+                                                                            getObjectRequest);
+
+
+        downloadDirectoryRequest.overrideConfiguration()
+                                .map(DownloadDirectoryOverrideConfiguration::downloadFileRequestTransformer)
```
Bennett-Lynch, line 201: _\<Bit off topic but want to write down my thoughts so we can capture them somewhere\>_

Even though it's not an immediately planned feature/concern, I'm trying to think what it would look like to extend this download logic to simulate "sync" logic. E.g.,

> A s3 object will require copying if the sizes of the two s3 objects differ, the last modified time of the source is newer than the last modified time of the destination, or the s3 object does not exist under the specified bucket and prefix destination.

https://docs.aws.amazon.com/cli/latest/reference/s3/sync.html#examples

I'm not sure if the current `downloadFileRequestTransformer` is extensible enough to support something like this. Should we consider something like a `BiPredicate<GetObjectResponse, Path>` that allows users to declare if an individual download should be allowed to execute or not? If we supported something like that, then in theory, implementing sync-download logic would be as simple as implementing it to replicate the behavior quoted above.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/DownloadDirectoryHelper.java
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    private CompletableFuture<CompletedFileDownload> downloadSingleFile(DownloadDirectoryRequest downloadDirectoryRequest,
+                                                                        Collection<FailedFileDownload> failedFileDownloads,
+                                                                        S3Object s3Object) {
+        String directory = downloadDirectoryRequest.destinationDirectory().toString();
+        String relativePath = getRelativePath(getDelimiter(downloadDirectoryRequest),
+                                              s3Object.key());
+
+        Path destinationPath = fileSystem.getPath(directory, relativePath);
```
dagnir, line 161: I think this will potentially be non-intuitive behavior.  `DownloadDirectoryRequest.destinationDirectory()` is a `Path`, and those are tied to a specific filesystem (they have `Path#getFileSystem()`). So if their `destinationDirectory` was created using something like `myCustomFs.getPath(..)`, I'd expect the file to end up in `myCustomFs`, instead of the system default filesystem.

I think in all cases, anywhere we need to refer to the filesystem, it should be through `destinationDirectory.getFileSystem()`. I think that also means you don't need to take a `FileSystem` parameter and can handle any number of them.

services-custom/s3-transfer-manager/src/test/java/software/amazon/awssdk/transfer/s3/internal/DownloadDirectoryHelperParameterizedTest.java
zoewangg: It seems it's not possible to get the parameter in BeforeEach and AfterEach callbacks in Junit5 :( 
https://github.com/junit-team/junit5/issues/944 I'll work around it

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/S3TransferManager.java
```diff
@@ -281,8 +281,11 @@ default DirectoryUpload uploadDirectory(Consumer<UploadDirectoryRequest.Builder>
      * bucket will be downloaded.
      *
      * <p>
+     * The SDK will create the destination directory if it does not already exist.
```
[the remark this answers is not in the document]
Bennett-Lynch: I think it will be almost guaranteed that there will be user-demand for this feature on something like download directory. It allows users to treat it like "sync" and it would allow it to behave symmetrical to upload directory (where S3 will not reject overwrites by default).

To be honest, I wish the default behavior were to overwrite-if-exists. I know that's not easily changeable, but I wonder if we should reconsider the default for new features like this.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/DownloadDirectoryHelper.java
```diff
@@ -0,0 +1,225 @@
...
+        ListObjectsV2Request request = ListObjectsV2Request.builder()
+                                                           .bucket(bucket)
+                                                           .prefix(prefix)
+                                                           .delimiter(delimiter)
+                                                           .build();
+
+        Collection<FailedFileDownload> failedFileDownloads = new ConcurrentLinkedQueue<>();
+        List<CompletableFuture<?>> futures = new ArrayList<>();
+
+        listObjectsRecursivelyHelper.s3Objects(request).subscribe(s3Object -> {
```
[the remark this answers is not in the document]
Bennett-Lynch: I think a publisher is fine.

services-custom/s3-transfer-manager/src/test/java/software/amazon/awssdk/transfer/s3/internal/DownloadDirectoryHelperParameterizedTest.java
Bennett-Lynch: Probably because you can have different parameter sources for different tests. A little inconvenient for the simple case but makes sense for when you want to use multiple sources.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/S3TransferManager.java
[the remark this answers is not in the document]
zoewangg: Yeah, I think we have to add overwrite-if-exists to support pause and resume before GA anyway. We can revisit the default behavior for download directory once we have that.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/DownloadDirectoryHelper.java
```diff
@@ -127,16 +121,17 @@ private void doDownloadDirectory(CompletableFuture<CompletedDirectoryDownload> r
...
+        listObjectsHelper.listS3ObjectsRecursively(request).subscribe(s3Object -> {
             log.debug(() -> "s3Object key: " + s3Object.key());
             futures.add(downloadSingleFile(downloadDirectoryRequest,
                                            failedFileDownloads,
                                            s3Object));
         }).whenComplete((r, t) -> {
             if (t != null) {
-                returnFuture.completeExceptionally(SdkClientException.create("Failed to call ListObjectsV2  ", t));
+                returnFuture.completeExceptionally(SdkClientException.create("Failed to call ListObjectsV2", t));
+                cleanUpDirectory(downloadDirectoryRequest.destinationDirectory());
```
Bennett-Lynch, line 134: We should probably only call this if the SDK was responsible for creating the directory. If the directory already existed, I don't think we should delete it.

zoewangg: To close the loop, discussed offline and agreed that we should move the logic to create directories in FileAsyncResponseTransformer and FileResponseTransformer. https://github.com/aws/aws-sdk-java-v2/pull/3018/

dagnir (approved):

Bennett-Lynch (approved):

joviegas (approved):

PR #3011, opened 2022-02-03
Add commented-out placeholders for wire logging

Bennett-Lynch:
When testing, we often want to enable wire logging or request logging without enabling all debug logging. This commit adds the relevant syntax to enable such logging in our log4j2.properties files, commented-out, so that developers can easily uncomment sections when needed.


cenedhryn (approved):
I like it

PR #3020, opened 2022-02-09
[TM DownloadDirectory Part3] Various updates on downloadDirectory

zoewangg:
## Modifications
- Limit the number of concurrent downloads by creating a dedicated S3ObjectSubscriber with backpressure support
- Create nonexistent parent directories 
- Strip prefix from the path if it's not empty. Feature request from v1 https://github.com/aws/aws-sdk-java/issues/1321
- Add more Javadocs
- Don't pass delimiter by default to avoid potentially high number of listObjectsV2 calls. Feature request from v1: https://github.com/aws/aws-sdk-java/issues/1215

## Testing
Added unit tests

## Types of changes

## Checklist

## License


services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/DownloadDirectoryHelper.java
```diff
@@ -110,78 +112,167 @@ private void doDownloadDirectory(CompletableFuture<CompletedDirectoryDownload> r
...
+            this.request = request;
+            this.returnFuture = returnFuture;
+            this.downloadFileFunction = downloadFileFunction;
+            this.bufferSize = concurrentDownload;
+        }
+
+        @Override
+        public void onSubscribe(Subscription subscription) {
+            if (subscription == null) {
+                throw new NullPointerException("Subscription must not be null");
```
joviegas, line 148: Validate.paramNotNull(subscription, "subscription") ?

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/DownloadDirectoryHelper.java
```diff
@@ -110,78 +112,167 @@ private void doDownloadDirectory(CompletableFuture<CompletedDirectoryDownload> r
...
+        }
+
+        /**
+         * Normalizing the key by stripping the prefix from the s3 object key if the prefix is not empty. For example: given a
+         * request with prefix = "notes/2021", delimiter = "/" and key = "notes/2021/1.txt", the returned string should be
+         * "1.txt"
+         */
+        private static String normalizeKey(DownloadDirectoryRequest downloadDirectoryRequest, S3Object s3Object,
+                                           String delimiter) {
+            return downloadDirectoryRequest.prefix()
```
joviegas, line 242: NonBlocker , just suggestion :
was curious if we can use regex for this , something like `^(.+)\/([^\/]+)$` where /  6th character is the delimiter
```
String regexExpression = String.format("^(.+)\\%s([^\\/]+)$", delimiter);
```

joviegas (approved):

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/DownloadDirectoryHelper.java
```diff
@@ -110,78 +112,167 @@ private void doDownloadDirectory(CompletableFuture<CompletedDirectoryDownload> r
...
+                                                                                           failedFileDownloads,
+                                                                                           s3Object);
+            currentBuffer.add(singleFileFuture);
+
+            if (currentBuffer.size() == bufferSize) {
+                CompletableFuture.allOf(currentBuffer.toArray(new CompletableFuture[0]))
+                                 .whenComplete((response, throwable) -> {
+                                     subscription.request(1);
+                                     currentBuffer.clear();
+                                 });
```
Bennett-Lynch, line 173: I think this buffer implementation may result in creating bursts of demand, rather than a smooth continuous demand. E.g., once the buffer is full, we will entirely cease to list objects, until all N objects finish, at which point we will resume listing (but be briefly stuck awaiting a list response). This may be partially offset by the fact that List is a paginated operation, so we only really need 1 call to resume meaningful work, but there will still be a small period of delay that we could probably avoid...

So rather than buffering in batches, could we possibly use a `Queue` that we continuously write to and read from, and only cease requesting more objects when the queue reaches a max size?

Also, would it be possible to expose this functionality via a convenience method like [`SdkPublisher#buffer`](https://github.com/aws/aws-sdk-java-v2/blob/8c79c6b1ecc313e535065e86abaf0d43c12437c5/core/sdk-core/src/main/java/software/amazon/awssdk/core/async/SdkPublisher.java#L98-L107)?

zoewangg: Fancy! The implementation would be more complicated though. I'm not sure if the benefit justifies the efforts because there will not be any delay for the case where there are still s3Objects from the previous ListObjectsV2Response in memory, and even if we need to call listObjectsV2, the latency seems to be trivial in the process of downloading multiple objects. 

> So rather than buffering in batches, could we possibly use a Queue that we continuously write to and read from, and only cease requesting more objects when the queue reaches a max size?

It seems `size()` is not a constant-time operation in the concurrent queue implementations and we probably don't want to use blocking queue or add locks before inserting and removing. There are ways to make it work, but more complex.

 > Also, would it be possible to expose this functionality via a convenience method like [SdkPublisher#buffer](https://github.com/aws/aws-sdk-java-v2/blob/8c79c6b1ecc313e535065e86abaf0d43c12437c5/core/sdk-core/src/main/java/software/amazon/awssdk/core/async/SdkPublisher.java#L98-L107)?

I'll see if I can generalize it.

Bennett-Lynch: Good point that most of the time there will be a remaining/partial list response still in memory. Also agree that the list delay would be mostly amortized in the context of bulk object transfers, but I think it's still something we could consider optimizing for.

I think another, perhaps more concerning scenario, is when the 100th download takes a very long time to download, and all other 99 downloads have finished, there will only be 1 in-flight transfer until the 100th completes.

zoewangg: Yeah, that is an interesting scenario. Good point, I'm convinced. I'll make the change.

utils/src/main/java/software/amazon/awssdk/utils/async/AsyncBufferingSubscriber.java
```diff
@@ -0,0 +1,202 @@
...
+
+    public AsyncBufferingSubscriber(Function<T, CompletableFuture<?>> consumer,
+                                    CompletableFuture<Void> future,
+                                    int maxConcurrentRequests) {
+        this.buffer = new ConcurrentLinkedQueue<>();
+        this.future = future;
+        this.consumer = consumer;
+        this.maxConcurrentRequests = maxConcurrentRequests;
+        this.numRequestsInFlight = new AtomicInteger(0);
+        this.futures = new ArrayList<>();
```
Bennett-Lynch, line 60: Are we using this?

utils/src/main/java/software/amazon/awssdk/utils/async/AsyncBufferingSubscriber.java
```diff
@@ -0,0 +1,202 @@
...
+    private final List<CompletableFuture<?>> futures;
+
+    public AsyncBufferingSubscriber(Function<T, CompletableFuture<?>> consumer,
+                                    CompletableFuture<Void> future,
+                                    int maxConcurrentRequests) {
+        this.buffer = new ConcurrentLinkedQueue<>();
+        this.future = future;
+        this.consumer = consumer;
+        this.maxConcurrentRequests = maxConcurrentRequests;
+        this.numRequestsInFlight = new AtomicInteger(0);
```
Bennett-Lynch, line 59: We don't provide synchronization to guarantee that `buffer` and `numRequestsInFlight` are updated atomically. Can this result in any race conditions with the current logic?

utils/src/main/java/software/amazon/awssdk/utils/async/AsyncBufferingSubscriber.java
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    private enum EventType {
+        DATA,
+        COMPLETE
+    }
+    
+    private interface Event<T> {
+        EventType type();
+    }
```
Bennett-Lynch, line 176: I think this approach is fine (and pretty clean), but we could also consider avoiding the wrapper entirely by using a `Queue<Object>`, and declaring a `private static final COMPLETE_EVENT = new Object()`. If the direct reference comparison fails, we know the object is a `T`.

utils/src/main/java/software/amazon/awssdk/utils/async/AsyncBufferingSubscriber.java
```diff
@@ -0,0 +1,202 @@
...
+        if (buffer.isEmpty()) {
+            if (isStreamingDone && numRequestsInFlight.get() == 0) {
+                future.complete(null);
+            } else {
+                subscription.request(1);
+            }
+            return;
+        }
+
+        if (isDelivering.compareAndSet(false, true)) {
```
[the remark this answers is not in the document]
Bennett-Lynch: It may also be more conventional to use a `Semaphore` and `tryAcquire` here.

utils/src/main/java/software/amazon/awssdk/utils/async/AsyncBufferingSubscriber.java
```diff
@@ -0,0 +1,202 @@
...
+    private static final Logger log = Logger.loggerFor(AsyncBufferingSubscriber.class);
+    private final Event<T> completeEvent = new CompleteEvent<>();
+    private final Queue<Event<T>> buffer;
+    private final CompletableFuture<?> future;
+    private final Function<T, CompletableFuture<?>> consumer;
+    private final int maxConcurrentRequests;
+    private final AtomicInteger numRequestsInFlight;
+    private final AtomicBoolean isDelivering = new AtomicBoolean(false);
+    private volatile boolean isStreamingDone;
+    private Subscription subscription;
```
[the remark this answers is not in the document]
zoewangg: Yes, this needs to be volatile. Will fix

utils/src/main/java/software/amazon/awssdk/utils/async/AsyncBufferingSubscriber.java
```diff
@@ -0,0 +1,202 @@
...
+/**
+ * An implementation of {@link Subscriber} that delivers data to async consumer and limits the number of concurrent requests by
+ * only requesting data if the number of inflight request is less than the configured maxConcurrentRequests.
+ *
+ * @param <T> Type of data requested
+ */
+@SdkProtectedApi
+public class AsyncBufferingSubscriber<T> implements Subscriber<T> {
+    private static final Logger log = Logger.loggerFor(AsyncBufferingSubscriber.class);
+    private final Event<T> completeEvent = new CompleteEvent<>();
```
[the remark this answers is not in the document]
zoewangg: Generic can't be used in a static field. :(

utils/src/main/java/software/amazon/awssdk/utils/async/AsyncBufferingSubscriber.java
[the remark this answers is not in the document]
zoewangg: CAS is pretty common imo and works perfect here, and CAS is all over the place in the SDK. :)

utils/src/main/java/software/amazon/awssdk/utils/async/AsyncBufferingSubscriber.java
Bennett-Lynch: Thanks, I confused myself. So we have `numRequestsInFlight` and `maxConcurrentRequests`, and we only request more objects when `cur` < `max`, right? Won't this result in the queue/buffer being almost always empty?

Should we have an additional parameter, e.g., `maxBacklog`, that dictates the max the queue is allowed to grow? It seems like an easy addition since the queue logic already exists. Otherwise, I'm not if we even really need a queue at all.

zoewangg: The Javadoc I wrote is wrong...I'll update it.

zoewangg: To close the loop, we discussed offline and decided to defer this and keep this class as an internal API in s3-transfer-manager module for now.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/DownloadDirectoryHelper.java
zoewangg: I'm not sure if it's possible to define a regex because we don't know the length of the prefix beforehand.

PR #3021, opened 2022-02-10
Standardize the way SdkBytes are unmarshalled when they're part of a payload and the service returns no content.

millems:
After this change, when SdkBytes are modeled as a payload, the SDK will always return a non-null empty SdkBytes. When SdkBytes are modeled as a field, they will be null if the service did not specify them.

This issue was brought to our attention when #2982 standardized the Netty HTTP client's behavior. Previously, chunked encoded responses would behave differently than non-chunked encoded payloads. While that change made the behavior consistent, it had the unintentional effect of the SDK now returning null for blob-returning operations (like Lambda's invoke) instead of an empty SdkBytes (unless the Lambda response used chunked encoding, in which case the result would have been null even under the old behavior).

This change standardizes this behavior across all protocols. The table at the end of this PR summarizes the current and new behavior.

Other changes in this PR:
1. Upgrade wiremock to wiremock-jre8. The old package is deprecated, and the newest package was required for junit 5 extension support. (Note: Wiremock could not be upgraded for the HTTP client tests, because of a bug in the latest wiremock that caused wiremock to not detect HEAD requests properly)

| Client Type | SDK Protocol | SdkBytes Location | Response Encoding | Previous Value | New Value |
|---|---|---|---|---|---|
| SYNC | JSON | PAYLOAD | CONTENT_LENGTH | empty | empty |
| SYNC | JSON | PAYLOAD | CHUNKED | empty | empty |
| SYNC | JSON | FIELD | CONTENT_LENGTH | null | null |
| SYNC | JSON | FIELD | CHUNKED | null | null |
| SYNC | XML | PAYLOAD | CONTENT_LENGTH | empty | empty |
| SYNC | XML | PAYLOAD | CHUNKED | empty | empty |
| SYNC | XML | FIELD | CONTENT_LENGTH | null | null |
| SYNC | XML | FIELD | CHUNKED | null | null |
| SYNC | QUERY | PAYLOAD | CONTENT_LENGTH | exception | empty |
| SYNC | QUERY | PAYLOAD | CHUNKED | exception | empty  |
| SYNC | QUERY | FIELD | CONTENT_LENGTH | exception | null |
| SYNC | QUERY | FIELD | CHUNKED | exception | null |
| ASYNC | JSON | PAYLOAD | CONTENT_LENGTH | null* | empty |
| ASYNC | JSON | PAYLOAD | CHUNKED | null | empty |
| ASYNC | JSON | FIELD | CONTENT_LENGTH | null | null |
| ASYNC | JSON | FIELD | CHUNKED | null | null |
| ASYNC | XML | PAYLOAD | CONTENT_LENGTH | null | empty |
| ASYNC | XML | PAYLOAD | CHUNKED | null | empty |
| ASYNC | XML | FIELD | CONTENT_LENGTH | null | null |
| ASYNC | XML | FIELD | CHUNKED | null | null |
| ASYNC | QUERY | PAYLOAD | CONTENT_LENGTH | null | empty |
| ASYNC | QUERY | PAYLOAD | CHUNKED | null | empty |
| ASYNC | QUERY | FIELD | CONTENT_LENGTH | null | null |
| ASYNC | QUERY | FIELD | CHUNKED | null | null |

\* This was the behavior that was changed with #2982. Prior to that change, it would have been 'empty'.


utils/src/main/java/software/amazon/awssdk/utils/LookaheadInputStream.java
```diff
@@ -0,0 +1,122 @@
...
+
+package software.amazon.awssdk.utils;
+
+import java.io.FilterInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import software.amazon.awssdk.annotations.SdkProtectedApi;
+
+@SdkProtectedApi
+public class LookaheadInputStream extends FilterInputStream {
```
cenedhryn, line 24: nice utility

dagnir (approved):

PR #3037, opened 2022-02-16
Allow SDK to build on JDK 17

millems:
The following changes were required:
1. Update environment variable mocking approach to no longer rely on reflection. Instead, all environment variable reading in the SDK (via SystemSetttings) supports value overrides via a testing backdoor added to SystemSettingsUtils.
2. Update to Mockito 4.x from 1.x.
3. Do not rely on reflection for comparing Instants in protocol-test-core
4. Run tests with ` -XX:+AllowRedefinitionToAddDeleteMethods` enabled, because blockhound requires it for Java 13+: https://github.com/reactor/BlockHound/issues/33

### DIFF
--- a/build-tools/src/main/resources/software/amazon/awssdk/checkstyle.xml
+++ b/build-tools/src/main/resources/software/amazon/awssdk/checkstyle.xml
@@ -345,8 +345,18 @@
         <module name="Regexp">
           <property name="format" value="System\s*(\.|::)\s*(getenv|getProperty)"/>
           <property name="illegalPattern" value="true"/>
-          <property name="message" value="Use SystemSetting instead of System.getenv and System.getProperty"/>
+          <property name="message" value="NEVER use System.getenv or System.getProperty. Create and use a
+          SystemSetting, instead."/>
           <property name="ignoreComments" value="true"/>
+        </module>
+
+        <module name="Regexp">
+            <property name="format" value="SystemSetting\s*(\.|::)\s*getStringValueFromEnvironmentVariable"/>
+            <property name="illegalPattern" value="true"/>
+            <property name="message" value="Are you being naughty and not creating a SystemSetting? Don't you know that
+            your customers want both system properties AND environment variables? Are you REALLY sure you want to use
+            this and not support a system property?"/>
+            <property name="ignoreComments" value="true"/>
         </module>
 
         <!-- Checks that we don't implement AutoCloseable/Closeable -->
--- a/core/auth/src/main/java/software/amazon/awssdk/auth/credentials/EnvironmentVariableCredentialsProvider.java
+++ b/core/auth/src/main/java/software/amazon/awssdk/auth/credentials/EnvironmentVariableCredentialsProvider.java
@@ -39,7 +39,7 @@ public final class EnvironmentVariableCredentialsProvider extends SystemSettings
     protected Optional<String> loadSetting(SystemSetting setting) {
         // CHECKSTYLE:OFF - Customers should be able to specify a credentials provider that only looks at the environment
         // variables, but not the system properties. For that reason, we're only checking the environment variable here.
-        return Optional.ofNullable(System.getenv(setting.environmentVariable()));
+        return SystemSetting.getStringValueFromEnvironmentVariable(setting.environmentVariable());
         // CHECKSTYLE:ON
     }
 
--- a/http-client-spi/src/main/java/software/amazon/awssdk/http/AbortableInputStream.java
+++ b/http-client-spi/src/main/java/software/amazon/awssdk/http/AbortableInputStream.java
@@ -20,6 +20,7 @@ import static software.amazon.awssdk.utils.Validate.paramNotNull;
 import java.io.FilterInputStream;
 import java.io.InputStream;
 import software.amazon.awssdk.annotations.SdkProtectedApi;
+import software.amazon.awssdk.annotations.SdkTestInternalApi;
 
 /**
  * Input stream that can be aborted. Abort typically means to destroy underlying HTTP connection
@@ -63,6 +64,14 @@ public final class AbortableInputStream extends FilterInputStream implements Abo
     @Override
     public void abort() {
         abortable.abort();
+    }
+
+    /**
+     * Access the underlying delegate stream, for testing purposes.
+     */
+    @SdkTestInternalApi
+    public InputStream delegate() {
+        return in;
     }
 
 }
--- a/http-clients/url-connection-client/src/test/java/software/amazon/awssdk/http/urlconnection/UrlConnectionHttpClientWithCustomCreateWireMockTest.java
+++ b/http-clients/url-connection-client/src/test/java/software/amazon/awssdk/http/urlconnection/UrlConnectionHttpClientWithCustomCreateWireMockTest.java
@@ -15,13 +15,18 @@
 package software.amazon.awssdk.http.urlconnection;
 
 import static org.assertj.core.api.AssertionsForClassTypes.assertThatThrownBy;
-import static org.mockito.Mockito.doThrow;
-import static org.mockito.Mockito.spy;
 import static software.amazon.awssdk.utils.FunctionalUtils.invokeSafely;
 import static software.amazon.awssdk.utils.FunctionalUtils.safeFunction;
 
 import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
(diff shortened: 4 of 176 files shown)


build-tools/src/main/resources/software/amazon/awssdk/checkstyle.xml
```diff
@@ -345,10 +345,20 @@
...
         </module>
 
+        <module name="Regexp">
+            <property name="format" value="SystemSetting\s*(\.|::)\s*getStringValueFromEnvironmentVariable"/>
+            <property name="illegalPattern" value="true"/>
+            <property name="message" value="Are you being naughty and not creating a SystemSetting? Don't you know that
+            your customers want both system properties AND environment variables? Are you REALLY sure you want to use
+            this and not support a system property?"/>
+            <property name="ignoreComments" value="true"/>
+        </module>
```
millems, line 360: Note for reviewers: This was modified to really push people to use SystemSetting, because that's the only mechanism that supports environment variable mocking, now. We expect everyone to create new SystemSettings to support both properties and environment variables. If they really want to break that, then they can use the static `SystemSetting.getStringValueFromEnvironmentVariable` (long to be annoying to write), and they still have to add a checkstyle suppression to do it.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/EnvironmentVariableCredentialsProvider.java
```diff
@@ -39,7 +39,7 @@ public static EnvironmentVariableCredentialsProvider create() {
     protected Optional<String> loadSetting(SystemSetting setting) {
         // CHECKSTYLE:OFF - Customers should be able to specify a credentials provider that only looks at the environment
         // variables, but not the system properties. For that reason, we're only checking the environment variable here.
-        return Optional.ofNullable(System.getenv(setting.environmentVariable()));
+        return SystemSetting.getStringValueFromEnvironmentVariable(setting.environmentVariable());
```
millems, line 42: Note for reviewer: Non-Mockito change. This uses SystemSetting to support mocking.

http-client-spi/src/main/java/software/amazon/awssdk/http/AbortableInputStream.java
```diff
@@ -65,4 +66,12 @@ public void abort() {
         abortable.abort();
     }
 
+    /**
+     * Access the underlying delegate stream, for testing purposes.
+     */
+    @SdkTestInternalApi
+    public InputStream delegate() {
+        return in;
+    }
```
millems, line 75: Note to reviewer: Our existing tests relied on reflection to access 'in', which does not work in JDK 17. Now, we expose it via a test API. This should be okay, since this is a protected (not public) API.

http-clients/url-connection-client/src/test/java/software/amazon/awssdk/http/urlconnection/UrlConnectionHttpClientWithCustomCreateWireMockTest.java
```diff
@@ -63,15 +68,290 @@ public void connectionsAreNotReusedOn5xxErrors() throws Exception {
...
     public void testGetResponseCodeNpeIsWrappedAsIo() throws Exception {
-        connectionInterceptor = safeFunction(connection -> {
-            connection = spy(connection);
-            doThrow(new NullPointerException()).when(connection).getResponseCode();
-            return connection;
+        connectionInterceptor = safeFunction(connection -> new DelegateHttpURLConnection(connection) {
+            @Override
+            public int getResponseCode() {
+                throw new NullPointerException();
+            }
```
millems, line 75: Note to reviewer: Mockito can no longer spy on classes loaded by the bootstrap class loader in JDK 17, so we had to do the method overriding manually.

services-custom/dynamodb-enhanced/src/test/java/software/amazon/awssdk/enhanced/dynamodb/converters/attribute/InstantAsStringAttributeConvertersTest.java
```diff
@@ -79,12 +79,6 @@ public void InstantAsStringAttributeConverterNotAcceptLocalDateTimeTest() {
                                                                        .toAttributeValue()));
     }
 
-    @Test
-    public void InstantAsStringAttributeConverterNotAcceptOffsetTimeTest() {
-        assertFails(() -> transformTo(CONVERTER, EnhancedAttributeValue.fromString("1988-05-21T00:12:00+01:00")
-                                                                       .toAttributeValue()));
-    }
-
```
millems, line 87: Note to reviewer: JDK 17 now supports parsing offset times. There shouldn't be a problem supporting this now, so I removed the assertion that we do not support it.

test/http-client-tests/pom.xml
```diff
@@ -119,6 +119,12 @@
             <artifactId>netty-tcnative-boringssl-static</artifactId>
             <scope>compile</scope>
         </dependency>
+        <dependency>
+            <groupId>org.bouncycastle</groupId>
+            <artifactId>bcpkix-jdk15on</artifactId>
+            <version>1.70</version>
+            <scope>compile</scope>
+        </dependency>
```
millems, line 127: Note to reviewer: Netty uses JDK internals to generate self-signed certs for our netty services we spin up for testing purposes. This no longer works on JDK 17, because those classes aren't exposed. We add this (test, since it's a module that's only used for testing) dependency on bouncycastle, which netty will automatically use instead of the JDK classes.

test/protocol-tests-core/src/main/java/software/amazon/awssdk/protocol/asserts/unmarshalling/UnmarshalledResultAssertion.java
```diff
@@ -62,7 +71,11 @@ private void assertFieldEquals(Field field, Object actual, Object expectedResult
             assertTrue(IOUtils.contentEquals((InputStream) field.get(expectedResult),
                                              (InputStream) field.get(actual)));
         } else {
-            assertReflectionEquals(field.get(expectedResult), field.get(actual));
+            Difference difference = CustomComparatorFactory.getComparator()
```
millems, line 74: Note to reviewer: this fixes an issue where the comparator is using reflection to compare instants (which fails on JDK 17), because our comparator library is old enough that it doesn't support Instant types. (It hasn't been updated since 2017).

pom.xml
```diff
@@ -657,6 +656,17 @@
             </properties>
         </profile>
 
+        <profile>
```
joviegas, line 659: 👍

test/protocol-tests-core/src/main/java/software/amazon/awssdk/protocol/asserts/unmarshalling/UnmarshalledResultAssertion.java
```diff
@@ -77,4 +90,36 @@ private String getOutputClassName(UnmarshallingTestContext context) {
...
+            comparators.add(new InstantComparator());
+            comparators.addAll(ReflectionComparatorFactory.getComparatorChain(Collections.emptySet()));
+            return new ReflectionComparator(comparators);
+        }
+    }
+
+    private static class InstantComparator implements Comparator {
+        @Override
+        public boolean canCompare(Object left, Object right) {
+            return left instanceof Instant || right instanceof Instant;
```
joviegas, line 106: Query :why isn't  it ?
```java
return left instanceof Instant && right instanceof Instant;
```

joviegas (approved):

test/protocol-tests-core/src/main/java/software/amazon/awssdk/protocol/asserts/unmarshalling/UnmarshalledResultAssertion.java
millems: I had the same thought at first. I looked at other comparator implementations and they did ||. It actually makes sense after I thought about it. This comparator knows whether two instants are equal, but it also knows whether one instant is equal to a non-instant, so it makes sense to say it can compare an instant and a non-instant (to return false).

joviegas: got it, thanks  👍

PR #3039, opened 2022-02-17
Add Copy support to S3 Transfer Manager

Bennett-Lynch:
This adds support for Copy to `S3TransferManager`, allowing users to create a copy of an object that is already stored in S3. Under the hood, by leveraging the [AWS Common Runtime (CRT)](https://docs.aws.amazon.com/sdkref/latest/guide/common-runtime.html), `S3TransferManager` will intelligently use plain `CopyObjectRequest`s for smaller objects, or multiple parallel `UploadPartCopyRequest`s for larger objects.

This PR is marked as a draft until two small-but-blocking issues in the CRT's copy implementation are resolved and released in a new version:

1. https://github.com/awslabs/aws-c-s3/pull/178
2. https://github.com/awslabs/aws-c-s3/pull/180


services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/CopyRequest.java
```diff
@@ -0,0 +1,227 @@
...
+import software.amazon.awssdk.annotations.SdkPublicApi;
+import software.amazon.awssdk.core.async.AsyncRequestBody;
+import software.amazon.awssdk.services.s3.model.CopyObjectRequest;
+import software.amazon.awssdk.utils.ToString;
+import software.amazon.awssdk.utils.Validate;
+import software.amazon.awssdk.utils.builder.CopyableBuilder;
+import software.amazon.awssdk.utils.builder.ToCopyableBuilder;
+
+/**
+ * Copy the given {@link AsyncRequestBody} to an object in S3. For file-based copys, you may use {@link CopyFileRequest} instead.
```
zoewangg, line 34: This doc seems to be wrong

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/CopyRequest.java
```diff
@@ -0,0 +1,227 @@
...
+        return DefaultBuilder.class;
+    }
+
+    @Override
+    public Builder toBuilder() {
+        return new DefaultBuilder(this);
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
zoewangg, line 87: Can we add tests for equals and hashcode for all model classes? We've run into issues before where we forgot to update them after we added a new field

zoewangg (approved):
Can we make sure stability tests and integ tests pass before merging?

PR #3042, opened 2022-02-18
Add support for filtering objects as part of S3 Transfer Manager's DownloadDirectory API

Bennett-Lynch:
This allows users to specify a filter that will be used to evaluate which objects should be downloaded from the target S3 directory.

You can use a filter, for example, to only download objects of a given size, of a given file extension, of a given last-modified date, etc.

By default, if no filter is specified, all objects will be downloaded.

In the future, this may allow us to more easily support "[sync](https://docs.aws.amazon.com/cli/latest/reference/s3/sync.html)" functionality, but support for that feature is out of scope currently.


services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/DownloadFilter.java
```diff
@@ -0,0 +1,80 @@
...
+
+/**
+ * {@link DownloadFilter} allows you to filter out which objects should be downloaded as part of a {@link
+ * DownloadDirectoryRequest}. You could use it, for example, to only download objects of a given size, of a given file extension,
+ * of a given last-modified date, etc. Multiple {@link DownloadFilter}s can be composed together via {@link #and(BiPredicate)} and
+ * {@link #or(BiPredicate)} methods.
+ */
+@SdkPublicApi
+@SdkPreviewApi
+public interface DownloadFilter extends BiPredicate<S3Object, Path> {
```
Bennett-Lynch, line 34: 1. This extends `BiPredicate` so that we can directly take advantage of the default methods it offers, such as `and` and `or`.

2. I opted to call it `DownloadFilter`, as opposed to `DownloadDirectoryFilter`, since filtering would presumably only ever be relevant for batch-transfer-style operations.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/DownloadFilter.java
```diff
@@ -0,0 +1,80 @@
...
+
+    /**
+     * A {@link DownloadFilter} that only downloads objects that are newer (i.e., modified more recently) than local files. This
+     * relies on comparing the remote object's last modified time to the local file's last modified time. It does not compare
+     * content checksums.
+     * <p>
+     * This filter is considered preview/experimental and may be modified or removed before release.
+     */
+    @SdkPreviewApi
+    static DownloadFilter newerObjects() {
```
Bennett-Lynch, line 54: This filter implementation currently lacks tests. I wanted to add it here as a sort of proof-of-concept. If we like the idea, I can add more thorough tests.

zoewangg: I like this idea, but we don't support overwriting existing files yet. I think we should compare checksum as well

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/DownloadDirectoryRequest.java
```diff
@@ -84,6 +86,14 @@ public String bucket() {
         return Optional.ofNullable(delimiter);
     }
 
+    /**
+     * @return the optional filter
+     * @see Builder#filter(DownloadFilter) 
+     */
+    public Optional<DownloadFilter> filter() {
+        return Optional.ofNullable(filter);
```
zoewangg, line 94: Should we just return `(remote, local) -> true;` if it's null? https://github.com/aws/aws-sdk-java-v2/blob/master/services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/UploadDirectoryOverrideConfiguration.java#L82

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/DownloadDirectoryRequest.java
```diff
@@ -236,8 +251,20 @@ public String toString() {
...
+         * last-modified date, etc. See {@link DownloadFilter} for some ready-made implementations. Multiple {@link
+         * DownloadFilter}s can be composed together via the {@code and} and {@code or} methods.
+         * <p>
+         * By default, if no filter is specified, all objects will be downloaded.
+         *
+         * @param filter the filter
+         * @return This builder for method chaining.
+         * @see DownloadFilter
+         */
+        Builder filter(DownloadFilter filter);
```
zoewangg, line 267: Should we move it to the DownloadDirectoryRequestOverrideConfiguration? The `uploadFileRequestTransformer` is in the `UploadDirectoryOverrideConfiguration`
https://github.com/aws/aws-sdk-java-v2/blob/master/services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/UploadDirectoryRequest.java#L94

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/DownloadDirectoryHelper.java
```diff
@@ -114,13 +116,16 @@ private void doDownloadDirectory(CompletableFuture<CompletedDirectoryDownload> r
...
 
         CompletableFuture<Void> allOfFutures = new CompletableFuture<>();
-        AsyncBufferingSubscriber<S3Object> asyncBufferingSubscriber =
-            new AsyncBufferingSubscriber<>(s3Object -> downloadSingleFile(downloadDirectoryRequest,
-                                                                          failedFileDownloads,
-                                                                          s3Object),
+        AsyncBufferingSubscriber<Pair<S3Object, Path>> asyncBufferingSubscriber =
+            new AsyncBufferingSubscriber<>(objectAndPath -> downloadSingleFile(downloadDirectoryRequest,
+                                                                               failedFileDownloads,
+                                                                               objectAndPath),
```
cenedhryn, line 122: I like the pairing of object and path (don't like the name but can't think of a better one; it IS descriptive) which allows for a more efficient implementation of the the single file download; despite the new functionality the class only grew by 3 lines.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/DownloadFilter.java
Bennett-Lynch: Hmm, `DownloadFileRequest` does capture more information (`TransferRequestOverrideConfiguration`, `GetObjectRequest`), but I'm not sure how likely that information is to be relevant for decision-making here. I feel like `<S3Object, Path>` is more symmetrical and elegant, and I'd prefer to keep that unless we have reason to believe users would need the additional information.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/DownloadFilter.java
Bennett-Lynch: Good point. Let me remove this for now. Will defer til we at least have support for overwriting existing files.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/DownloadDirectoryRequest.java
Bennett-Lynch: Good point, will do.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/DownloadDirectoryRequest.java
Bennett-Lynch: I suppose `OverrideConfiguration` is intended for parameters that we expect to be configurable at both the client and request level. Do you think it makes sense to have `DownloadFilter` configurable at the client level?

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/DownloadFilter.java
Bennett-Lynch: Ack and mostly agree. However, there is one other small issue I just realized: `DownloadFileRequest` does not include `S3Object`, where the latter is needed for metadata like last modified time. So we would need to return a tuple/wrapper containing both `DownloadFileRequest` and `S3Object`, which feels quite cumbersome.

zoewangg: Yeah, we'd have to use `BiPredicate<S3Object, DownloadFileRequest>` or Predicate<DownloadFileMetadata>`. Now I think about it, `DownloadFileMetadata` might be better. In the initial release, it can just have S3Object and Path, and we can add more if needed in the future.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/DownloadDirectoryRequest.java
zoewangg: `UploadDirectoryOverrideConfiguration` is a request-level configuration though, but yeah, it's a bit confusing right now with this naming. We should definitional revisit it before GA.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/DownloadFilter.java
Bennett-Lynch: Oh, right, whoops! `DownloadFileRequest` has the `Path`. Realized that before but forgot.

Overall I'm still inclined towards `BiPredicate<S3Object, Path>` but I understand the argument towards `BiPredicate<S3Object, DownloadFileRequest>` or `Predicate<DownloadFileMetadata>`. Maybe this is a good one to let democracy decide.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/DownloadDirectoryRequest.java
Bennett-Lynch: Understood it's request-level. What I mean is, it seems that the rule of thumb ought to be that anything that can be configured at *both* the client and request level ought to have its request-level option be in an `OverrideConfiguration`.

If we agree on that premise, do we think it would make sense to also make `DownloadFilter` configurable at the client-level, or is it more intuitive as only a request-level option?

zoewangg: Gotcha. We don't currently have a way to determine if an option should be part of the `request` itself or the `overrideConfiguration`, but what you said makes sense to me. We just need to make sure it's consistent across the TM. I'll add it to the API surface area review checklist

IMO it's more intuitive as only a request-level option.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/DownloadFilter.java
Bennett-Lynch: Going with `Predicate<DownloadFileMetadata>`!

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/DownloadDirectoryRequest.java
Bennett-Lynch: If we agree it's more intuitive as a request-level-only option then I personally think it would make more sense as a request parameter. But definitely want to be consistent over everything else. Since you've captured it in the surface area review checklist, I'll opt to leave it as is and we can revisit it in a broader context later.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/DownloadFileContext.java
```diff
@@ -0,0 +1,36 @@
...
+
+import java.nio.file.Path;
+import software.amazon.awssdk.services.s3.model.S3Object;
+
+/**
+ * Context object for determining which objects should be downloaded as part of a {@link DownloadDirectoryRequest}.
+ *
+ * @see DownloadFilter
+ */
+public interface DownloadFileContext {
```
Bennett-Lynch, line 26: Created an interface, primarily in order to hide the constructor as an internal-API.

Alternatively, can not use an interface, but would need to expose a proper builder. Let me know if any preference.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/DownloadFileContext.java
```diff
@@ -0,0 +1,36 @@
...
+    /**
+     * @return A description of the remote S3 object
+     */
+    S3Object source();
```
Bennett-Lynch, line 30: Opted to call this `source` so it reads symmetrical with `destination`.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/DownloadFileContext.java
```diff
@@ -0,0 +1,36 @@
...
+
+    /**
+     * @return A path representing the local file destination
+     */
+    Path destination();
```
Bennett-Lynch, line 35: For now I'm proposing *not* including the `GetObjectRequest` here because I think it's unlikely to be useful. In the event we wish to add it in the future, it would be safe to add.

zoewangg (approved):

PR #3050, opened 2022-02-22
Improved Expect: 100-continue response handling in the URL connection client.

millems:
HTTPURLConnection currently throws a ProtocolException if the service returns anything other than a 100 or 417 response code to an expect-100-continue request. S3 sometimes triggers this behavior. This change catches this exception and treats it according to its response code, assuming that it can verify that no information was lost (e.g. because of a response payload being included in the response). If it still fails, it instructs the customer that the apache HTTP client could be a true solution to the problem.

### DIFF
--- a/http-clients/url-connection-client/src/main/java/software/amazon/awssdk/http/urlconnection/UrlConnectionHttpClient.java
+++ b/http-clients/url-connection-client/src/main/java/software/amazon/awssdk/http/urlconnection/UrlConnectionHttpClient.java
@@ -23,6 +23,8 @@ import static software.amazon.awssdk.utils.NumericUtils.saturatedCast;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.OutputStream;
+import java.io.UncheckedIOException;
 import java.net.HttpURLConnection;
 import java.net.URI;
 import java.security.KeyManagementException;
@@ -31,6 +33,9 @@ import java.security.cert.X509Certificate;
 import java.time.Duration;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.function.Supplier;
 import java.util.stream.Collectors;
 import javax.net.ssl.HostnameVerifier;
 import javax.net.ssl.HttpsURLConnection;
@@ -42,6 +47,7 @@ import javax.net.ssl.TrustManager;
 import javax.net.ssl.X509TrustManager;
 import software.amazon.awssdk.annotations.SdkPublicApi;
 import software.amazon.awssdk.http.AbortableInputStream;
+import software.amazon.awssdk.http.ContentStreamProvider;
 import software.amazon.awssdk.http.ExecutableHttpRequest;
 import software.amazon.awssdk.http.HttpExecuteRequest;
 import software.amazon.awssdk.http.HttpExecuteResponse;
@@ -195,9 +201,14 @@ public final class UrlConnectionHttpClient implements SdkHttpClient {
     }
 
     private static class RequestCallable implements ExecutableHttpRequest {
-
         private final HttpURLConnection connection;
         private final HttpExecuteRequest request;
+
+        /**
+         * Whether we encountered the 'bug' in the way the HttpURLConnection handles 'Expect: 100-continue' cases. See
+         * {@link #getAndHandle100Bug} for more information.
+         */
+        private boolean expect100BugEncountered = false;
 
         private RequestCallable(HttpURLConnection connection, HttpExecuteRequest request) {
             this.connection = connection;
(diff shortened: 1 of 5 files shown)


http-clients/url-connection-client/src/test/java/software/amazon/awssdk/http/urlconnection/Expect100ContinueTest.java
```diff
@@ -0,0 +1,256 @@
...
+                response.addHeader("x-amz-test-header", "foo");
+                response.flushBuffer();
+            }
+        };
+
+        try (SdkHttpClient client = UrlConnectionHttpClient.create();
+             EmbeddedServer server = new EmbeddedServer(handler)) {
+            HttpExecuteResponse response = sendRequest(client, server);
+            assertThat(response.httpResponse().statusCode()).isEqualTo(500);
+            assertThat(response.httpResponse().firstMatchingHeader("x-amz-test-header")).hasValue("foo");
```
Bennett-Lynch, line 159: Nit: Could also assert a warning was logged with `LogCaptor`.

http-clients/url-connection-client/src/main/java/software/amazon/awssdk/http/urlconnection/UrlConnectionHttpClient.java
```diff
@@ -228,6 +239,85 @@ public HttpExecuteResponse call() throws IOException {
...
+
+        private Optional<InputStream> tryGetInputStream() {
+            return getAndHandle100Bug(() -> invokeSafely(connection::getInputStream), Expect100BugHandlingBehavior.EXCEPTION);
+        }
+
+        private Optional<InputStream> tryGetErrorStream() {
+            return getAndHandle100Bug(() -> invokeSafely(connection::getErrorStream), Expect100BugHandlingBehavior.LOG);
+        }
+
+        private <T> Optional<T> getAndHandle100Bug(Supplier<T> supplier, Expect100BugHandlingBehavior bugBehavior) {
```
Bennett-Lynch, line 254: Nice helper method! A small description of the bug may be helpful for posterity's sake.

Bennett-Lynch (approved):

http-clients/url-connection-client/src/main/java/software/amazon/awssdk/http/urlconnection/UrlConnectionHttpClient.java
```diff
@@ -228,6 +244,90 @@ public HttpExecuteResponse call() throws IOException {
...
+         */
+        private <T> Optional<T> getAndHandle100Bug(Supplier<T> supplier, boolean failOn100Bug) {
+            try {
+                return Optional.ofNullable(supplier.get());
+            } catch (RuntimeException e) {
+                if (!exceptionCausedBy100HandlingBug(e)) {
+                    throw e;
+                }
+
+                expect100BugEncountered = true;
```
Bennett-Lynch, line 288: Should we put `responseHasNoContent()` before this, so that we only emit warnings when we're actually missing some potential behavior?

http-clients/url-connection-client/src/main/java/software/amazon/awssdk/http/urlconnection/UrlConnectionHttpClient.java
```diff
@@ -228,6 +244,90 @@ public HttpExecuteResponse call() throws IOException {
...
+        }
+
+        private Optional<InputStream> tryGetInputStream() {
+            return getAndHandle100Bug(() -> invokeSafely(connection::getInputStream), true);
+        }
+
+        private Optional<InputStream> tryGetErrorStream() {
+            InputStream result = invokeSafely(connection::getErrorStream);
+            if (result == null && expect100BugEncountered) {
+                log.debug(() -> "The response payload has been dropped because of a limitation of the JDK's URL Connection "
```
Bennett-Lynch, line 258: Is there anyway we can trace/correlate this to a given request? Should we include the status line?

Bennett-Lynch (approved):

http-clients/url-connection-client/src/main/java/software/amazon/awssdk/http/urlconnection/UrlConnectionHttpClient.java
millems: Since it's a sync client, they can use the thread name?

http-clients/url-connection-client/src/test/java/software/amazon/awssdk/http/urlconnection/Expect100ContinueTest.java
```diff
@@ -0,0 +1,268 @@
...
+        try (SdkHttpClient client = UrlConnectionHttpClient.create();
+             EmbeddedServer server = new EmbeddedServer(handler)) {
+            HttpExecuteResponse response = sendRequest(client, server);
+            assertThat(response.httpResponse().statusCode()).isEqualTo(417);
+            assertThat(response.httpResponse().firstMatchingHeader("x-amz-test-header")).hasValue("foo");
+        }
+    }
+
+    @Test
+    public void expect100ContinueWorksWithZeroContentLength500() throws Exception {
```
Bennett-Lynch, line 124: Nit: Should we assert that no warnings were logged?

http-clients/url-connection-client/src/main/java/software/amazon/awssdk/http/urlconnection/UrlConnectionHttpClient.java
millems: Sure. responseHasNoContent() is relatively slow to check, so I'll have to rearrange some things to ensure we don't check it twice in the success case.

http-clients/url-connection-client/src/test/java/software/amazon/awssdk/http/urlconnection/Expect100ContinueTest.java
millems: Is that to prevent someone upgrading the log to warn?

Bennett-Lynch: I meant to assert that nothing was logged at all (if there was no missing behavior).

Bennett-Lynch (approved):

PR #3062, opened 2022-03-01
Extend client interface (sync) if extension methods exist in classpath

cenedhryn:


codegen/src/main/java/software/amazon/awssdk/codegen/poet/PoetUtils.java
```diff
@@ -109,4 +112,15 @@ public static JavaFile buildJavaFile(ClassSpec spec) {
         return builder.build();
     }
 
+    public static Optional<Class<?>> findExtensionClass(ClassName clientInterfaceName, Class<?> clazz) {
```
Bennett-Lynch, line 115: Can `clientInterfaceName` be inferred from `clazz`?

codegen/src/test/java/software/amazon/awssdk/codegen/poet/utils/SomeClientExtensionMethods.java
```diff
@@ -0,0 +1,23 @@
...
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.codegen.poet.utils;
+
+/**
+ * Test class used by {@link PoetUtilsTest}
+ */
+public class SomeClientExtensionMethods {
```
Bennett-Lynch, line 21: Should we require that this be an interface?

Bennett-Lynch (approved):

codegen/src/main/java/software/amazon/awssdk/codegen/poet/PoetUtils.java
cenedhryn: No, the clazz is determining how the class loader works. I meant to rename it to make it clearer lm do that.

codegen/src/test/java/software/amazon/awssdk/codegen/poet/utils/SomeClientExtensionMethods.java
cenedhryn: Hah, good point. I had mixed and matched in my tests but IRL it doesn't work if it's a class...

Bennett-Lynch (approved):

PR #3067, opened 2022-03-01
Extension detected by customization flag

cenedhryn:


services/s3/src/it/java/software/amazon/awssdk/services/s3/extensions/BucketExistsIntegrationTest.java
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    @AfterClass
+    public static void tearDownFixture() {
+        deleteBucketAndAllContents(BUCKET);
+    }
+
+    @Test
+    public void bucketExists() {
+        assertThat(s3.doesBucketExist(BUCKET)).isTrue();
```
Bennett-Lynch, line 44: v1 users rejoice!

It would be nice to test error conditions here as well. E.g., bucket exists but not owned by you.

services/s3/src/main/resources/codegen-resources/customization.config
```diff
@@ -159,5 +159,6 @@
     "createMethodParams": [
       "clientConfiguration"
     ]
-  }
+  },
+  "extensionInterface": "software.amazon.awssdk.services.s3.extensions.S3ClientExtensionMethods"
```
Bennett-Lynch, line 163: When we eventually support async extensions as well, do you foresee having 2 fields here? Would it maybe be simpler to just have the package prefix, and we can infer the rest? Or even a simple boolean if we can infer all of it.

Bennett-Lynch (approved):

services/s3/src/it/java/software/amazon/awssdk/services/s3/extensions/BucketExistsIntegrationTest.java
cenedhryn: This is primitive for sure!

services/s3/src/main/resources/codegen-resources/customization.config
cenedhryn: I think a name is better, at least a package name; leads to less inference. When I added this, I had in mind that we just slap Async on, but that's quite ugly. We could either infer both names or add a more complex structure.

PR #3073, opened 2022-03-02
[Hackathon] Adds wrappers providing shortcuts to S3 getUrl and S3Presigner

cenedhryn:


services/s3/src/main/java/software/amazon/awssdk/services/s3/extensions/S3ClientSdkExtension.java
```diff
@@ -86,4 +91,35 @@ default void deleteBucketAndAllContents(String bucket) {
...
+        return ((S3Client) this).utilities().getUrl(getUrlRequest);
+    }
+
+    /**
+     * Convenience method to retrieve the {@link S3Presigner} in order to generate presigned methods.
+     *
+     * @return The presigner
+     */
+    @SdkExtensionMethod
+    default S3Presigner presigner() {
```
Bennett-Lynch, line 122: We could also expose individual methods here, like `presignGetObject`, `presignPutObject`, etc. But it would be a lot of duplication of interfaces. At the same time, it's more in-line with the methods being discovered as a first-class citizen on the native API. WDYT?

Bennett-Lynch (approved):

services/s3/src/main/java/software/amazon/awssdk/services/s3/extensions/S3ClientSdkExtension.java
cenedhryn: I debated back and forth on this question and decided to not do it for the hackathon, since it is a lot of boilerplate. For "real" it's up for discussion but should be easy enough to add.

PR #3074, opened 2022-03-02
[Hackathon] Add extension support to async clients

Bennett-Lynch:


codegen/src/main/java/software/amazon/awssdk/codegen/poet/client/SyncClientInterface.java
```diff
@@ -97,8 +97,8 @@ public TypeSpec poetSpec() {
                                               ServiceMetadataProvider.class)
                                   .build());
 
-        Optional<ClassName> extensionClass = findExtensionInterface(className, model.getCustomizationConfig());
-        extensionClass.ifPresent(result::addSuperinterface);
+        Optional<ClassName> extensionInterface = findExtensionInterface(className, model.getCustomizationConfig());
```
cenedhryn, line 100: The overloaded meaning of class in Java makes naming difficult sometimes

cenedhryn (approved):

codegen/src/main/java/software/amazon/awssdk/codegen/poet/client/SyncClientInterface.java
Bennett-Lynch: Very true, this is still a "class" by most definitions.

PR #3076, opened 2022-03-03
[Hackathon] Add extension method: verifyBucketOwnership

Bennett-Lynch:


services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/extensions/DefaultS3ClientSdkExtension.java
```diff
@@ -53,4 +55,16 @@ public boolean doesObjectExist(String bucket, String key) {
             return false;
         }
     }
+
+    @Override
+    public void verifyBucketOwnership(String bucket) {
+        boolean isBucketOwner = s3.listBuckets()
```
cenedhryn, line 61: What if there are many many buckets?
Is it possible to use a headBucket request, or is that too inconclusive?

Bennett-Lynch: 1. The maximum number of buckets is 1,000. `listBuckets` isn't a paginated operation.
2. Verifying access to a bucket via `headBucket` doesn't verify that we own it. The intent here is to verify that we own a bucket that we may be potentially writing to.
3. To use `headBucket` we would need to use the `expectedBucketOwner` parameter, but the difficulty here is that we don't necessarily know the caller's AWS Account ID. However, `listBuckets` works as a clever workaround here because it "returns a list of all buckets owned by the authenticated sender of the request".

cenedhryn: Sounds good! Thanks for clarifying.

cenedhryn (approved):

PR #3083, opened 2022-03-07
Make API Gateway integ test's deleteRestApi more reliable & best-effort

Bennett-Lynch:
## License


services/apigateway/src/it/java/software/amazon/awssdk/services/apigateway/ServiceIntegrationTest.java
```diff
@@ -84,7 +89,14 @@ public static void createRestApi() {
     @AfterClass
     public static void deleteRestApiKey() {
         if (restApiId != null) {
-            apiGateway.deleteRestApi(DeleteRestApiRequest.builder().restApiId(restApiId).build());
+            DELETE_RATE_LIMITER.getValue().acquire();
+            try {
+                apiGateway.deleteRestApi(DeleteRestApiRequest.builder().restApiId(restApiId).build());
+            } catch (Exception e) {
```
joviegas, line 95: Query : Why aren't we catching just the specific exception `TooManyRequestsException`  here?

Bennett-Lynch: I wanted to catch any potential transient failure (`TooManyRequestsException`, `500`, `503`, etc.) and not ever fail the tests on those occurring. But I agree it could be potentially dangerous to catch something like a `400`.

I'll assume that most transient failures will be covered by retries and change this to just catch `TooManyRequestsException`.

joviegas (approved):

services/apigateway/src/it/java/software/amazon/awssdk/services/apigateway/ServiceIntegrationTest.java
joviegas: Thank you

PR #3086, opened 2022-03-09
Adding AtomicCounter extension to DDB Enhanced

cenedhryn:
## Motivation and Context
Writing extensions can be complicated. Providing an AtomicCounter extension gives customers the ability to easily increment attributes in DDB just by adding tags to attributes and doesn't require them to learn how to write an extension, nor does it require them to write low level DDB UpdateExpressions

## Modifications
- Adds tags representing an AtomicCounter for bean based and static schemas
- Adding the `AtomicCounterExtension` and adding this to the default extension chain (if we want, the extension can be excluded for manual inclusion only)
- Adding an AtomicCounter object for storing start and delta values

## Behavior
- When using `putItem`, a counter attribute will be initialized with its start value
- When using `updateItem`, a counter attribute will be initialized with its start value if it doesn't exist, otherwise incremented by the increment delta value

## Testing
Tested using unit and functional tests


services-custom/dynamodb-enhanced/src/test/java/software/amazon/awssdk/enhanced/dynamodb/functionaltests/UpdateBehaviorTest.java
```diff
@@ -141,4 +145,29 @@ public void updateBehaviors_transactWriteItems_secondUpdate() {
         assertThat(persistedRecord.getLastAutoUpdatedOn()).isAfterOrEqualTo(INSTANT_2);
         assertThat(persistedRecord.getCreatedAutoUpdateOn()).isEqualTo(firstUpdatedRecord.getCreatedAutoUpdateOn());
     }
+
+    /**
+     * Currently, nested records are not updated through extensions.
```
cenedhryn, line 150: Added this test case for future changes to behavior of nested records with tagged attributes that are manipulated through extensions and operations.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/mapper/StaticAttributeTags.java
```diff
@@ -117,6 +118,25 @@ public static StaticAttributeTag updateBehavior(UpdateBehavior updateBehavior) {
...
     }
 
+    /**
+     * Used to explicitly designate an attribute to be an auto-generated counter updated unconditionally in DynamoDB.
+     * By supplying a negative integer delta value, the attribute works as a decreasing counter.
+     *
+     * @param delta The value to increment (positive) or decrement (negative) the counter with for each update.
+     * @param startValue The starting value of the counter.
+     */
+    public static StaticAttributeTag atomicCounter(long delta, long startValue) {
```
zoewangg, line 128: Do you expect that there will be other parameters in the future?

services-custom/dynamodb-enhanced/src/test/java/software/amazon/awssdk/enhanced/dynamodb/functionaltests/models/AtomicCounterRecord.java
```diff
@@ -0,0 +1,68 @@
...
+
+    @DynamoDbPartitionKey
+    public String getId() {
+        return id;
+    }
+    public void setId(String id) {
+        this.id = id;
+    }
+
+    @DynamoDbAtomicCounter
```
zoewangg, line 38: Can we also add tests for other types of table schemas?

zoewangg (commented):
Can we add

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/mapper/StaticAttributeTags.java
cenedhryn: No, I cannot think of any

services-custom/dynamodb-enhanced/src/test/java/software/amazon/awssdk/enhanced/dynamodb/functionaltests/models/AtomicCounterRecord.java
cenedhryn: The `AtomicCounterExtensionTest` uses a `StaticTableSchema`

zoewangg: Yeah, I meant other table schemas like the ImmutableTableSchema and StaticImmutableTableSchema

cenedhryn: A really good point. I looked in the test code and it seems like we are in the habit of testing with StaticTableSchema and BeanTableSchema in general, and skipping ImmutableTableSchema. 

Both StaticTableSchema and BeanTableSchema wrap StaticImmutableTableSchema, therefore IMO testing those also test StaticImmutableTableSchema. 

It's really great that we have a lot of tests for the enhanced client, but I do find that the sheer number of tests make it confusing to find relevant tests as well as making sure we are covering all features as well as permutations of them. In this case (AtomicCounters), what the schemas do is that they should interpret tags correctly, so I created a new test that test custom metadata tags for all 3 schemas users tend to use directly. We have three different types of custom metadata right now so I tested each of them.

zoewangg: Nice, thanks for creating parameterized tests. Yeah, I had not discovered the tests either before Ben told me about them. I wonder if there's a way to enforce the tests so that we don't forget in the future. Currently, it's purely relying on code reviews and tribal knowledge

zoewangg (approved):

PR #3105, opened 2022-03-17
Add Transfer Manager integ tests for local dirs with Unicode characters

Bennett-Lynch:
## License

- [x] I confirm that this pull request can be released under the Apache 2 license

PR #3111, opened 2022-03-19
Add Transfer Manager integ tests for local dirs with Unicode characters

zoewangg:
## Modifications
- Add Transfer Manager integ tests for local dirs with Unicode characters added through #3105 
- Fix failed unit test
- Fix bucket name
- Migrate some JUnit5 eligible tests to JUnit5

## Testing
Verified the tests passed on Windows and Linux

### DIFF
--- a/services-custom/s3-transfer-manager/src/it/java/software/amazon/awssdk/transfer/s3/S3TransferManagerUploadDirectoryUnicodeIntegrationTest.java
+++ b/services-custom/s3-transfer-manager/src/it/java/software/amazon/awssdk/transfer/s3/S3TransferManagerUploadDirectoryUnicodeIntegrationTest.java
@@ -1,0 +1,169 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.transfer.s3;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static software.amazon.awssdk.testutils.service.S3BucketUtils.temporaryBucketName;
+import static software.amazon.awssdk.utils.IoUtils.closeQuietly;
+
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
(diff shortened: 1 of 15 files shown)


services-custom/s3-transfer-manager/src/it/java/software/amazon/awssdk/transfer/s3/S3TransferManagerUploadDirectoryUnicodeIntegrationTest.java
```diff
@@ -0,0 +1,169 @@
...
+import software.amazon.awssdk.services.s3.model.S3Object;
+import software.amazon.awssdk.testutils.FileUtils;
+import software.amazon.awssdk.utils.Logger;
+
+/**
+ * Tests the behavior of traversing local directories with special Unicode characters in their path name. These characters have
+ * known to be problematic when using Java's old File API or with Windows (which uses UTF-16 for file-name encoding).
+ */
+@RunWith(Parameterized.class)
+public class S3TransferManagerUploadDirectoryUnicodeIntegrationTest extends S3IntegrationTestBase {
```
joviegas, line 52: Just wondering why haven't we use Junit 5 here ?

services-custom/s3-transfer-manager/src/it/java/software/amazon/awssdk/transfer/s3/S3TransferManagerUploadDirectoryUnicodeIntegrationTest.java
```diff
@@ -0,0 +1,169 @@
...
+        createBucket(TEST_BUCKET);
+        testDirectory = createLocalTestDirectory();
+    }
+
+    @After
+    public void testTeardown() {
+        try {
+            FileUtils.cleanUpTestDirectory(testDirectory);
+        } catch (Exception exception) {
+            log.warn(() -> "Failed to clean up test directory " + testDirectory, exception);
```
joviegas, line 109: Should't this be an error log ?

joviegas (approved):

services-custom/s3-transfer-manager/src/it/java/software/amazon/awssdk/transfer/s3/S3TransferManagerUploadDirectoryUnicodeIntegrationTest.java
zoewangg: I'm not sure(this test is written by Bennett). The reason I didn't migrate this particular test to JUnit 5 as part of this PR is that it will introduce more changes, which may make code review more difficult. I do have plans to migrate all TM integration tests to use JUnit 5 in the the next individual day. :)

services-custom/s3-transfer-manager/src/it/java/software/amazon/awssdk/transfer/s3/S3TransferManagerUploadDirectoryUnicodeIntegrationTest.java
zoewangg: IMO it doesn't really matter, the log level for all tests is `warn`.

PR #3119, opened 2022-03-23
Add union type design overview.

millems:


docs/design/core/tagged-unions/README.md
```diff
@@ -0,0 +1,150 @@
...
+3. A `void accept(X.VoidVisitor)` method for branching out on the union type. e.g. `getItemResponse.item().accept(new AttributeValue.VoidVisitor() { ... })`
+4. A `X.Type type()` method for determining the type contained within the union. e.g. `AttributeValue.Type type = getItemResponse.item().type()`
+
+For example, an `AttributeValue` union type containing `String s`, `String n` and `SdkBytes b` would have the following public methods:
+```java
+class AttributeValue {
+    static AttributeValue.Builder builder();
+    static fromS(String s);
+    static fromN(String s);
+    static fromB(String s);
```
millems, line 59: Oops: SdkBytes b

docs/design/core/tagged-unions/README.md
```diff
@@ -0,0 +1,150 @@
...
+    
+    String s();
+    String n();
+    SdkBytes b();
+
+    AttributeValue.Type type();
+    void accept(AttributeValue.VoidVisitor visitor);
+    <T> T accept(AttributeValue.Visitor<T> visitor);
```
millems, line 67: I was wondering, should we have these? The customer can do a switch based on the `type()` for something equivalent.

dagnir (approved):

docs/design/core/tagged-unions/README.md
```diff
@@ -0,0 +1,150 @@
...
+else if (attributeValue.n() != null) { System.out.println(attributeValue.n()); }
+else if (attributeValue.b() != null) { System.out.println(attributeValue.b().asUtf8String()); }
+```
+
+**After Changes**
+```java
+attributeValue.visit(new AttributeValue.VoidVisitor() {
+    void visitS(String s) { System.out.println(s); }
+    void visitN(String n) { System.out.println(n); }
+    void visitB(SdkBytes b) { System.out.println(b.asUtf8String()); }
```
zoewangg, line 148: Or using pattern matching for switch!

zoewangg (approved):

docs/design/core/tagged-unions/README.md
millems: Conclusion: They might be nice, but the void visitor is basically just a switch on the type() and the visitor<T> is just a switch expression on the type() (added in Java 17). Admittedly, not a lot of customers are on 17 right now, but it would be good not to add it if we can avoid it. We'll end up waiting for customer feedback on this one before we add it.

Issue #3108, opened 2022-03-18
Add configuration to overwrite existing file in S3TransferManager.downloadFile


TheKeveloper:
### Describe the feature

Add a configuration option in `DownloadFileRequest` to overwrite the existing local file if it already exists. Currently if the destination file already exists locally, an exception is thrown. 

This seems similar to this PR: https://github.com/aws/aws-sdk-java-v2/pull/2595 and this corresponding issue: https://github.com/aws/aws-sdk-java-v2/issues/2300

### Is your Feature Request related to a problem?

I often need to retrieve a file from S3 that may have been updated and use that to update my local copy of the file. This feature existed in the v1 sdk but does not seem to exist in the v2 sdk.

One way around this is by saving to a temp file and then moving the temp file to replace the existing file. However, this requires a lot of additional effort to make sure that the temp file gets cleaned up properly.

Another possibility is to use a `getObject` request with `AsyncResponseTransformer.toBytes()` and doing a `Files.copy` on the `InputStream` with `StandardCopyOptions.REPLACE_EXISTING`. However, this doesn't allow using TransferManager and seems like a level of abstraction that should already exist. 

### Proposed Solution

It seems like merging https://github.com/aws/aws-sdk-java-v2/pull/2595 would be a good first step, and then just configuring TransferManager to use the corresponding `toFile` implementation based on a configuration option in the `DownloadFileRequest` should get us there. 

### Acknowledge

- [X] I may be able to implement this feature request

### AWS Java SDK version used

2.17.151-PREVIEW

### JDK version used

15.0.3

### Operating System and version

MacOSX 10.15.7

zoewangg:
Hi @TheKeveloper, good timing! We are working on this feature right now. We are planning to make overwriting existing file the default behavior in the TM since it seems to be more intuitive.

zoewangg:
We have changed the default behavior to replacing existing file if it exists in the transfer manager. This change has been released in `2.17.179`.  This behavior can be configured in `FileTransformerConfiguration` https://github.com/aws/aws-sdk-java-v2/blob/master/core/sdk-core/src/main/java/software/amazon/awssdk/core/FileTransformerConfiguration.java

Issue #2300, opened 2021-02-26
S3: GetObjectRequest toFile option to override file


mina-asham:
Migrating from SDK V1 to V2 we hit an issue where do something like this (pseudo-code):
```
Path temp = Files.createTempFile(...);
s3.getObject(bucket, key, temp);
// use temp file
```
This allows us to download a temp file, validate it and if we are happy flip it to the actual path overriding the existing previous file.

This works on V1 because: because it just opens an outputstream which by default overwrites ([here](https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/internal/ServiceUtils.java#L304))
but doesn't work on V2 because we are not passing any options to `Files.copy` [here](https://github.com/aws/aws-sdk-java-v2/blob/master/core/sdk-core/src/main/java/software/amazon/awssdk/core/sync/ResponseTransformer.java#L106)

## Describe the Feature

Allow the use of S3 client to download an object to an existing file by overwriting it

## Is your Feature Request related to a problem?

See general summary

## Proposed Solution

* Suggestion 1 (preferred): adding extra optional parameters here: https://github.com/aws/aws-sdk-java-v2/blob/master/core/sdk-core/src/main/java/software/amazon/awssdk/core/sync/ResponseTransformer.java#L102 that can be passed to the `Files.copy` call, this can be as straight as `CopyOption... options` which is passed through to `Files.copy` or something else that we translate to `CopyOptions`
* Suggestion 2 (less preferred because it's ugly but works): Add another method in `ResponseTransformer`, e.g. `ResponseTransformer.toFileOverwrite`
* Suggestion 3 (not preferred because it's a breaking change): Change the default behaviour and hardcode `StandardCopyOption.REPLACE_EXISTING`

## Describe alternatives you've considered

See suggestion 2/3 in proposed solution

## Additional Context

- [x] I may be able to implement this feature request

## Your Environment

* AWS Java SDK version used: N/A
* JDK version used: N/A
* Operating System and version: N/A

debora-ito:
Thank you for reaching out @mina-asham. 
We agree that should be a feature parity with v1 in this case, marking this as a feature request.

mina-asham:
I have submitted a PR for the suggested fix here: https://github.com/aws/aws-sdk-java-v2/pull/2595, thanks!

trathschlag:
If someone is looking for a quick workaround:

```java
[...]

PR #3125, opened 2022-03-24
[TM DownloadFile Pause and Resume] Part 1: Add configuration to enable overwriting existing files

zoewangg:
## Motivation and Context
#3108 
#2300 

## Modifications
- Expose an option to overwrite an existing file in `FileAsyncResponseTransformer`.

- Note that it's not added to `FileResponseTransformer` right now because it's not trivial to do so. I will create a backlog item for it.

## Testing

## Types of changes

## Checklist

## License

### DIFF
--- a/.changes/next-release/feature-AWSSDKforJavav2-1ee4d61.json
+++ b/.changes/next-release/feature-AWSSDKforJavav2-1ee4d61.json
@@ -1,0 +1,6 @@
+{
+    "category": "AWS SDK for Java v2", 
+    "contributor": "", 
+    "type": "feature", 
+    "description": "Expose an option in the `AsyncResponseTransformer#toFile` to allow overwriting and appending existing file."
+}
--- a/.changes/next-release/feature-AWSSDKforJavav2-1ee4d61.json
+++ b/.changes/next-release/feature-AWSSDKforJavav2-1ee4d61.json
@@ -2,5 +2,5 @@
     "category": "AWS SDK for Java v2", 
     "contributor": "", 
     "type": "feature", 
-    "description": "Expose an option in the `AsyncResponseTransformer#toFile` to allow overwriting and appending existing file."
+    "description": "Expose an option in `AsyncResponseTransformer#toFile` to allow overwriting and appending existing file."
 }
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/FileTransformerConfiguration.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/FileTransformerConfiguration.java
@@ -17,39 +17,43 @@ package software.amazon.awssdk.core;
 
 import java.nio.file.FileAlreadyExistsException;
 import java.nio.file.Path;
-import java.util.Objects;
-import java.util.Optional;
 import software.amazon.awssdk.annotations.SdkPublicApi;
 import software.amazon.awssdk.core.async.AsyncResponseTransformer;
+import software.amazon.awssdk.utils.Validate;
 import software.amazon.awssdk.utils.builder.CopyableBuilder;
 import software.amazon.awssdk.utils.builder.ToCopyableBuilder;
 
 /**
- * Configuration options for {@link AsyncResponseTransformer#toFile(Path, FileTransformerConfiguration)}. All values are optional.
+ * Configuration options for {@link AsyncResponseTransformer#toFile(Path, FileTransformerConfiguration)} to configure how the SDK
+ * should write the file and if the SDK should delete the file when an exception occurs.
+ *
+ * @see #builder()
+ * @see FileWriteOption
+ * @see FailureBehavior
  */
 @SdkPublicApi
 public final class FileTransformerConfiguration implements ToCopyableBuilder<FileTransformerConfiguration.Builder,
     FileTransformerConfiguration> {
     private final FileWriteOption fileWriteOption;
-    private final Boolean deleteOnFailure;
+    private final FailureBehavior failureBehavior;
 
     private FileTransformerConfiguration(DefaultBuilder builder) {
-        this.fileWriteOption = builder.fileWriteOption;
-        this.deleteOnFailure = builder.deleteOnFailure;
+        this.fileWriteOption = Validate.paramNotNull(builder.fileWriteOption, "fileWriteOption");
+        this.failureBehavior = Validate.paramNotNull(builder.failureBehavior, "failureBehavior");
     }
 
     /**
-     * The configured {@link FileWriteOption} or {@link Optional#empty()} if not configured
+     * The configured {@link FileWriteOption}
      */
-    public Optional<FileWriteOption> fileWriteOption() {
-        return Optional.ofNullable(fileWriteOption);
+    public FileWriteOption fileWriteOption() {
+        return fileWriteOption;
     }
 
     /**
-     * Whether to delete the file in the event of a failure.
+     * The configured {@link FailureBehavior}
      */
-    public Optional<Boolean> deleteOnFailure() {
-        return Optional.ofNullable(deleteOnFailure);
+    public FailureBehavior failureBehavior() {
+        return failureBehavior;
     }
 
     /**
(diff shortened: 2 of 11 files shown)


core/sdk-core/src/main/java/software/amazon/awssdk/core/FileTransformerConfiguration.java
```diff
@@ -0,0 +1,155 @@
...
+
+    private FileTransformerConfiguration(DefaultBuilder builder) {
+        this.fileModificationOption = builder.fileWriteOption;
+        this.deleteOnFailure = builder.deleteOnFailure;
+    }
+
+    /**
+     * The configured {@link FileWriteOption} or {@link Optional#empty()} if not configured
+     */
+    public Optional<FileWriteOption> fileModificationOption() {
```
millems, line 44: fileWriteOption*

core/sdk-core/src/main/java/software/amazon/awssdk/core/FileTransformerConfiguration.java
```diff
@@ -0,0 +1,155 @@
...
+         */
+        Builder fileWriteOption(FileWriteOption fileWriteOption);
+
+        /**
+         * Configures whether the file should be deleted in the event of an error
+         *
+         * @param deleteOnFailure whether to delete the file upon failure
+         * @return This object for method chaining.
+         */
+        Builder deleteOnFailure(Boolean deleteOnFailure);
```
millems, line 130: Maybe `failureBehavior(FailureBehavior.DELETE)` and `failureBehavior(FailureBehavior.LEAVE)` to leave room for something like `failureBehavior(FailureBehavior.REVERT)` in the future?

core/sdk-core/src/main/java/software/amazon/awssdk/core/async/AsyncResponseTransformer.java
```diff
@@ -116,11 +117,27 @@
...
+     *
+     * @param path        Path to file to write to.
+     * @param config      configuration for the transformer
+     * @param <ResponseT> Pojo Response type.
+     * @return AsyncResponseTransformer instance.
+     * @see FileTransformerConfiguration
+     */
+    static <ResponseT> AsyncResponseTransformer<ResponseT, ResponseT> toFile(Path path, FileTransformerConfiguration config) {
+        return new FileAsyncResponseTransformer<>(path, config);
+    }
```
millems, line 139: You probably considered it already, but a consumer-builder would be nice.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/FileAsyncResponseTransformer.java
```diff
@@ -44,13 +49,48 @@
...
-        return AsynchronousFileChannel.open(path, StandardOpenOption.WRITE, StandardOpenOption.CREATE_NEW);
+        switch (fileModificationOption) {
+            case CREATE_OR_APPEND_EXISTING:
+                return AsynchronousFileChannel.open(path, StandardOpenOption.WRITE, StandardOpenOption.CREATE);
+            case CREATE_OR_REPLACE_EXISTING:
+                return AsynchronousFileChannel.open(path, StandardOpenOption.WRITE, StandardOpenOption.CREATE,
+                                                    StandardOpenOption.TRUNCATE_EXISTING);
+            case CREATE_NEW:
+            default:
+                return AsynchronousFileChannel.open(path, StandardOpenOption.WRITE, StandardOpenOption.CREATE_NEW);
```
millems, line 92: Imo, we might want to fail by default to make sure it's an explicit choice when we add future modification options.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/FileAsyncResponseTransformer.java
```diff
@@ -44,13 +49,48 @@
     private volatile AsynchronousFileChannel fileChannel;
     private volatile CompletableFuture<Void> cf;
     private volatile ResponseT response;
+    private final long position;
+    private final FileWriteOption fileModificationOption;
```
millems, line 53: nit: fileWriteOption

.changes/next-release/feature-AWSSDKforJavav2-1ee4d61.json
```diff
@@ -0,0 +1,6 @@
+{
+    "category": "AWS SDK for Java v2", 
+    "contributor": "", 
+    "type": "feature", 
+    "description": "Expose an option in the `AsyncResponseTransformer#toFile` to allow overwriting and appending existing file."
```
dagnir, line 5: nit: remove "the"

core/sdk-core/src/main/java/software/amazon/awssdk/core/FileTransformerConfiguration.java
```diff
@@ -0,0 +1,155 @@
...
+
+        /**
+         * Create a new file if it doesn't exist, otherwise replace the existing file.
+         * <p>
+         * In the event of an error, the SDK will NOT attempt to delete the file.
+         */
+        CREATE_OR_REPLACE_EXISTING,
+
+        /**
+         * Create a new file if it doesn't exist, otherwise append the existing file.
```
dagnir, line 107: minor , "append to". Appending the existing file would be very different behavior!

core/sdk-core/src/main/java/software/amazon/awssdk/core/FileTransformerConfiguration.java
```diff
@@ -0,0 +1,155 @@
...
+import java.util.Objects;
+import java.util.Optional;
+import software.amazon.awssdk.annotations.SdkPublicApi;
+import software.amazon.awssdk.core.async.AsyncResponseTransformer;
+import software.amazon.awssdk.utils.builder.CopyableBuilder;
+import software.amazon.awssdk.utils.builder.ToCopyableBuilder;
+
+/**
+ * Configuration options for {@link AsyncResponseTransformer#toFile(Path, FileTransformerConfiguration)}. All values are optional.
+ */
```
dagnir, line 29: I think this section could benefit from explaining or at least mentioning the `FileWriteOption`s since their behavior is important to understand

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/FileAsyncResponseTransformer.java
```diff
@@ -44,13 +49,48 @@
...
+
+    private long determineFilePositionToWrite(Path path) {
+        if (fileModificationOption == CREATE_OR_APPEND_EXISTING) {
+            try {
+                return Files.size(path);
+            } catch (NoSuchFileException e) {
+                // Ignore
+            } catch (IOException exception) {
+                throw SdkClientException.create("Cannot determine the current file size " + path, exception);
+            }
```
dagnir, line 78: This seems like an expensive thing to do all the time (e.g. if the file is always being created for the first time). Is there any way we can defer this until we need it or use a solution that doesn't rely on catching `NoSuchFileException`?

core/sdk-core/src/main/java/software/amazon/awssdk/core/FileTransformerConfiguration.java
zoewangg: Good call!

core/sdk-core/src/main/java/software/amazon/awssdk/core/async/AsyncResponseTransformer.java
zoewangg: Yeah, thought about it and was concerned that having too many overloads may make it harder to use; I guess consistency takes precedence.

core/sdk-core/src/main/java/software/amazon/awssdk/core/FileTransformerConfiguration.java
zoewangg: Good idea!

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/FileAsyncResponseTransformer.java
dagnir: Hmm couldn't we do it at the time we open the file?

zoewangg: No not really, we open the file in `onStream` method, so the position we get at this point may not be correct if the previous attempt had written a few bytes already before it failed.

zoewangg:
Changed the target branch to `zoewang/tm-download-pause-resume` to make sure it works with TM download pause/resume before we merge it to mainline.

I'm still updating the PR

core/sdk-core/src/main/java/software/amazon/awssdk/core/FileTransformerConfiguration.java
```diff
@@ -0,0 +1,209 @@
...
+
+    /**
+     * Returns the default {@link FileTransformerConfiguration} for {@link FileWriteOption#CREATE_OR_APPEND_EXISTING}
+     * <p>
+     * Create a new file if it doesn't exist, otherwise append to the existing file.
+     * In the event of an error, the SDK will NOT attempt to delete the file, leaving it as-is
+     */
+    public static FileTransformerConfiguration defaultCreateOrAppend() {
+        return builder().fileWriteOption(FileWriteOption.CREATE_OR_APPEND_EXISTING)
+                        .failureBehavior(FailureBehavior.LEAVE)
```
millems, line 98: Any specific reason for this default?

millems (approved):

core/sdk-core/src/main/java/software/amazon/awssdk/core/FileTransformerConfiguration.java
zoewangg: Yeah, retry would not work as expected if we delete the file when there is an exception.

millems: But it also might not work as expected if we started appending, failed and then on a retry reappended the same data.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/FileAsyncResponseTransformer.java
```diff
@@ -44,13 +49,46 @@
...
+        if (configuration.fileWriteOption() == CREATE_OR_APPEND_EXISTING) {
+            try {
+                return Files.size(path);
+            } catch (NoSuchFileException e) {
+                // Ignore
+            } catch (IOException exception) {
+                throw SdkClientException.create("Cannot determine the current file size " + path, exception);
+            }
+        }
+        return  0L;
```
dagnir, line 77: Can we add a backlog item to revisit this? It makes me nervous to rely on catching an exception on a possibly hot code path.

dagnir (approved):

core/sdk-core/src/main/java/software/amazon/awssdk/core/FileTransformerConfiguration.java
zoewangg: The initial position is recorded in the ctor, so we will always append on the same offset on retries.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/FileAsyncResponseTransformer.java
zoewangg: We could call `Files#exists` but it uses the same try-catch though... I'm not sure how we can optimize it, but yeah, I can add a TODO

PR #3160, opened 2022-04-13
Automatically refresh async credential sources in the background.

millems:
Today, if customers have async credential refreshes enabled, refreshes only kicked off when new credentials are requested. This causes an issue for very low-traffic credential providers, because they may never refresh credentials until those credentials are actually needed, and then it's too late.

With this change, we periodically prime all CachedSuppliers configured with a NonBlocking prefetch strategy and stale credentials.

One additional change was made to simplify the prefetch timings of InstanceProfileCredentialsProvider so they are easier to understand and reason about.

### DIFF
--- a/core/auth/src/main/java/software/amazon/awssdk/auth/credentials/InstanceProfileCredentialsProvider.java
+++ b/core/auth/src/main/java/software/amazon/awssdk/auth/credentials/InstanceProfileCredentialsProvider.java
@@ -124,6 +124,7 @@ public final class InstanceProfileCredentialsProvider implements HttpCredentials
         LoadedCredentials credentials;
         try {
             credentials = httpCredentialsLoader.loadCredentials(createEndpointProvider());
+            logExpirationTime(credentials);
             this.cachedCredentials = credentials;
         } catch (RuntimeException | IOException e) {
             credentials = this.cachedCredentials;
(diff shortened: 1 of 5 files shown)


utils/src/main/java/software/amazon/awssdk/utils/cache/NonBlocking.java
```diff
@@ -36,14 +43,37 @@ public class NonBlocking implements CachedSupplier.PrefetchStrategy {
...
+    }
+
+    private void scheduleRefresh(Runnable valueUpdater) {
+        executor.schedule(() -> {
+            try {
+                valueUpdater.run();
+            } finally {
+                scheduleRefresh(valueUpdater);
+            }
+        }, 1, MINUTES);
```
joviegas, line 76: Can we make this scheduled wait time in minutes as configurable at the time of class initiation and provide default value of 1 minutes ?

millems: I considered making it configurable, but I didn't see a compelling reason to do so, since it's not a cache busting refresh. It just makes sure the cache doesn't go too cold. Functionally, configuring this to 1 second wouldn't keep the cache much fresher, and configuring it to 30 minutes doesn't really save much performance.

cenedhryn: +1, seems reasonable

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/InstanceProfileCredentialsProvider.java
```diff
@@ -163,26 +168,23 @@ private Instant prefetchTime(Instant expiration) {
             return now.plus(60, MINUTES);
         }
 
-        // If expiration time is 60+ minutes from now, refresh in 60 minutes or 60 minutes before expiration, whichever is
-        // sooner. This is the average case, where customers are using IMDS and there is no IMDS outage.
+        // If expiration time is 60+ minutes from now, refresh in 30 minutes.
```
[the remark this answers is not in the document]
millems: Yay, that was my hope.

cenedhryn (approved):

mmaltiar:
We need this, which aws-sdk-java version it will be available?

PR #3175, opened 2022-04-28
Various performance improvements

zoewangg:
## Modifications
Various performance improvements


core/sdk-core/src/main/java/software/amazon/awssdk/core/SdkStandardLogger.java
```diff
@@ -49,18 +50,25 @@ private SdkStandardLogger() {
...
+        details.add("Status Code: " + response.statusCode());
+        String requestId = SdkHttpUtils.firstMatchingHeaderFromCollection(response.headers(),
+                                                       X_AMZN_REQUEST_ID_HEADERS)
+                    .orElse(placeholder);
+
+        details.add("Request ID: " + requestId);
+
+        response.firstMatchingHeader(X_AMZ_ID_2_HEADER).ifPresent(extendedRequestId ->
+                                                                      details.add("Extended Request ID: " + extendedRequestId));
+        return details.toString();
```
millems, line 72: While we're worrying about efficiency, should we just use StringBuilder? This feels hard to read with StringJoiner.

core/sdk-core/src/main/java/software/amazon/awssdk/core/checksums/Algorithm.java
```diff
@@ -44,12 +47,16 @@ public static Algorithm fromValue(String value) {
...
         }
         String normalizedValue = StringUtils.lowerCase(value);
-        return Arrays.stream(values())
-                .filter(algorithm -> algorithm.value.equals(normalizedValue))
-                .findFirst()
-                .orElseThrow(() -> new IllegalArgumentException(String.format("Unknown Algorithm '%s'", normalizedValue)));
+        if (!VALUE_MAP.containsKey(normalizedValue)) {
+            throw new IllegalArgumentException("The provided value is not a valid algorithm " + value);
+        }
+        return VALUE_MAP.get(normalizedValue);
```
millems, line 53: Since we're being picky about performance, should we just do a get and check to see if the result is null? (I don't think we need null to be a valid algorithm)

millems (approved):

cenedhryn (approved):

core/sdk-core/src/main/java/software/amazon/awssdk/core/SdkStandardLogger.java
zoewangg: Yup, makes sense

core/sdk-core/src/main/java/software/amazon/awssdk/core/checksums/Algorithm.java
zoewangg: Yeah, updated.

PR #3177, opened 2022-05-03
Adding serialization capability to ResumableFileDownload

cenedhryn:
## Motivation and Context
With serialization and deserialization capability in ResumableFileDownload, a paused TM download can be serialized, persisted and at a later time deserialized and resumed.

## Modifications
- Adds a serializer with static methods that can serialize and deserialize between ResumableFileDownload and JSON format. 
- Local marshalling/unmarshalling classes to avoid dependencies on json protocol
- Added a simple JsonWriter to the json-utils module
- Smaller updates to ResumableFileDownload to fix bugs / add minor features

## Limitations
- Does not serialize 
  - TM configs (not needed, listeners do not need to be persisted)
  - RequestOverrideConfiguration; if this can be excluded it makes things less complicated

## Initialization
Currently, the serializer has static methods and initializes Json resources when called. Download objects may not need to be serialized, and therefore the cost of initialization is unnecessary unless needed. The ResumableFileDownload API is asymmetric, with the serialize methods being instance methods and the deserialize methods are static, which limits the initialization options.

## Testing
Unit tests

## Types of changes

## Checklist

## License


core/protocols/aws-json-protocol/src/main/java/software/amazon/awssdk/protocols/json/internal/unmarshall/JsonProtocolUnmarshaller.java
```diff
@@ -103,7 +103,7 @@ private static JsonUnmarshallerRegistry createUnmarshallerRegistry(
                 .build();
     }
 
-    private static SdkBytes unmarshallSdkBytes(JsonUnmarshallerContext context,
+    public static SdkBytes unmarshallSdkBytes(JsonUnmarshallerContext context,
```
dagnir, line 106: Hmm why are these changed to `public`?

core/protocols/aws-json-protocol/src/main/java/software/amazon/awssdk/protocols/json/internal/unmarshall/JsonUnmarshallerRegistry.java
```diff
@@ -24,7 +24,7 @@
  * Registry of {@link JsonUnmarshaller} implementations by location and type.
  */
 @SdkInternalApi
-final class JsonUnmarshallerRegistry extends AbstractMarshallingRegistry {
+public final class JsonUnmarshallerRegistry extends AbstractMarshallingRegistry {
```
dagnir, line 27: This is `@SdkInternal` so we can't use it outside of the model

core/protocols/aws-json-protocol/src/main/java/software/amazon/awssdk/protocols/json/internal/unmarshall/SimpleTypeJsonUnmarshaller.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+package software.amazon.awssdk.protocols.json.internal.unmarshall;
+
+import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.core.SdkField;
+import software.amazon.awssdk.protocols.core.StringToValueConverter;
+import software.amazon.awssdk.protocols.jsoncore.JsonNode;
+
+@SdkInternalApi
```
dagnir, line 23: Internal so shouldn't be used outside of the module

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/ResumableFileDownload.java
```diff
@@ -125,6 +143,129 @@ public Optional<Long> totalSizeInBytes() {
         return Optional.ofNullable(totalSizeInBytes);
     }
 
+    /**
+     * Persists this download object to a file in Base64-encoded JSON format.
+     *
+     * @param path The path to the file to which you want to write the serialized download object.
+     * @throws IOException If the data cannot be persisted
+     */
+    public void writeToFile(Path path) throws IOException {
```
dagnir, line 152: Shouldn't be throwing checked exceptions since this is a public interface

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/utils/ResumableFileDownloadSerializer.java
```diff
@@ -0,0 +1,269 @@
...
+
+        return generator.getBytes();
+    }
+
+    /**
+     * At this point we do not need to persist the TransferRequestOverrideConfiguration, because it only contains listeners and
+     * they are not used in the resume operation.
+     */
+    private static void marshallDownloadFileRequest(DownloadFileRequest fileRequest, JsonMarshallerContext context) {
+        context.jsonGenerator().writeFieldName("downloadFileRequest");
```
dagnir, line 92: nit: assign `jsonGenerator` to a local so the code is a bit easier to read. Ditto in other places

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/utils/ResumableFileDownloadSerializer.java
```diff
@@ -0,0 +1,269 @@
...
+import software.amazon.awssdk.protocols.jsoncore.JsonNode;
+import software.amazon.awssdk.protocols.jsoncore.JsonNodeParser;
+import software.amazon.awssdk.services.s3.model.GetObjectRequest;
+import software.amazon.awssdk.transfer.s3.DownloadFileRequest;
+import software.amazon.awssdk.transfer.s3.ResumableFileDownload;
+import software.amazon.awssdk.transfer.s3.S3TransferManager;
+import software.amazon.awssdk.utils.Logger;
+
+@SdkInternalApi
+public final class ResumableFileDownloadSerializer {
```
dagnir, line 51: The usage of the marshaller classes requires exposing some previously internal only classes. Is it possible to use the JSON generator directly in all areas without making things too tedious?

dagnir:
Can we add tests that reference a pre-generated serialized `ResumableFileDownload`?

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/ResumableFileDownload.java
```diff
@@ -125,6 +143,129 @@ public Optional<Long> totalSizeInBytes() {
...
+     * Writes the serialized JSON data representing this object to an output stream.
+     *
+     * @param outputStream The output stream to write the serialized object to.
+     * @throws IOException If the data cannot be copied from the input to output stream
+     */
+    public void writeToOutputStream(OutputStream outputStream) throws IOException {
+        byte[] bytes = ResumableFileDownloadSerializer.toJson(this);
+        ByteArrayInputStream byteArrayInputStream = new ByteArrayInputStream(bytes);
+        IoUtils.copy(byteArrayInputStream, outputStream);
+        IoUtils.closeQuietly(byteArrayInputStream, null);
```
zoewangg, line 166: We probably should not close outputstream to be consistent with Responsetransformer#toOutputStream

https://github.com/aws/aws-sdk-java-v2/blob/master/core/sdk-core/src/main/java/software/amazon/awssdk/core/sync/ResponseTransformer.java#L150

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/ResumableFileDownload.java
```diff
@@ -125,6 +143,129 @@ public Optional<Long> totalSizeInBytes() {
...
+     *
+     * @param bytes the serialized data
+     * @return the deserialized {@link ResumableFileDownload}
+     */
+    public static ResumableFileDownload fromBytes(byte[] bytes) {
+        return ResumableFileDownloadSerializer.fromJson(bytes);
+    }
+
+    /**
+     * Deserialize contents of an input stream with JSON data into a {@link ResumableFileDownload}.
```
zoewangg, line 240: Can we add docs to mention that caller is responsible to close the inputstream?

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/ResumableFileDownload.java
```diff
@@ -125,6 +143,129 @@ public Optional<Long> totalSizeInBytes() {
...
+        return ResumableFileDownloadSerializer.fromJson(toByteArray(inputStream));
+    }
+
+    /**
+     * Deserialize a string with JSON data into a {@link ResumableFileDownload}.
+     *
+     * @param contents the serialized data
+     * @return the deserialized {@link ResumableFileDownload}
+     */
+    public static ResumableFileDownload fromString(String contents) {
```
zoewangg, line 255: I just noticed we have `utf8` in the method name else where for this kind of method. Should it be `fromUtf8String`? `toString` would be `toUtf8String`, and then the overridden `toString` would be the standard SDK toString format. WDYT?

https://github.com/aws/aws-sdk-java-v2/blob/master/utils/src/main/java/software/amazon/awssdk/utils/IoUtils.java#L58

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/ResumableFileDownload.java
```diff
@@ -140,6 +281,26 @@ public interface Builder extends CopyableBuilder<Builder, ResumableFileDownload>
...
+         *
+         * <p>
+         * This is a convenience method that creates an instance of the {@link DownloadFileRequest} builder avoiding the
+         * need to create one manually via {@link DownloadFileRequest#builder()}.
+         *
+         * @param downloadFileRequestBuilder the download file request builder
+         * @return a reference to this object so that method calls can be chained together.
+         * @see #downloadFileRequest(DownloadFileRequest)
+         */
+        default ResumableFileDownload.Builder downloadFileRequest(Consumer<DownloadFileRequest.Builder>
```
zoewangg, line 295: Why do we need this? Can't we use `ResumableFileDownload.toBuilder()`?

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/utils/ResumableFileDownloadSerializer.java
```diff
@@ -0,0 +1,269 @@
...
+        context.jsonGenerator().writeFieldName("getObjectRequest");
+        context.jsonGenerator().writeStartObject();
+        validateNoRequestOverrideConfiguration(getObjectRequest);
+        getObjectRequest.sdkFields().forEach(field -> marshallPojoField(field, getObjectRequest, context));
+        context.jsonGenerator().writeEndObject();
+    }
+
+    private static void validateNoRequestOverrideConfiguration(GetObjectRequest getObjectRequest) {
+        if (getObjectRequest.overrideConfiguration().isPresent()) {
+            log.warn(() -> "ResumableFileDownload GetObjectRequest contains an override configuration that will not be "
```
zoewangg, line 110: Should it be `debug`? We can mention this behavior in the Javadocs

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/utils/ResumableFileDownloadSerializer.java
```diff
@@ -0,0 +1,269 @@
...
+                                                                                (SdkField<Object>) instantSdkField());
+                f.set(getObjectBuilder, instantValue);
+            } else if (marshallingType.equals(MarshallingType.INTEGER)) {
+                JsonUnmarshaller<Object> integerUnmarshaller =
+                    unmarshallerRegistry.getUnmarshaller(MarshallLocation.PAYLOAD, MarshallingType.INTEGER);
+                Integer integerValue = (Integer) integerUnmarshaller.unmarshall(context, value, (SdkField<Object>) f);
+                f.set(getObjectBuilder, integerValue);
+            } else {
+                throw new IllegalStateException("should not happen");
+            }
```
zoewangg, line 189: Would it make sense to use switch?

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/utils/ResumableFileDownloadSerializer.java
zoewangg: Maybe we can create a copy of `SimpleTypeJsonMarshaller`and `AwsStructuredPlainJsonFactory`with unnecessary stuff removed in s3-transfer-manager module? It seems they are all we needed. `StructuredJsonGenerator` is a protected API, so we should be fine

zoewangg, line 51: Can we make this class flexible enough to support `ResumableFileUpload`?

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/ResumableFileDownload.java
cenedhryn: ok, thanks

core/protocols/aws-json-protocol/src/main/java/software/amazon/awssdk/protocols/json/internal/unmarshall/JsonProtocolUnmarshaller.java
cenedhryn: so that they could be used, but I gather from other comments that it's not desirable

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/ResumableFileDownload.java
zoewangg: nvm, I got confused

core/protocols/aws-json-protocol/src/main/java/software/amazon/awssdk/protocols/json/internal/unmarshall/JsonUnmarshallerRegistry.java
cenedhryn: I didn't know the rule was that strict (but I can see from the annotation javadoc that is the definition) and it seemed like it was more efficient to use existing code. Discussion in separate thread.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/utils/ResumableFileDownloadSerializer.java
cenedhryn: will do

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/ResumableFileDownload.java
cenedhryn: I had missed that, will change.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/utils/ResumableFileDownloadSerializer.java
cenedhryn: I'll see how much code needs to be copied.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/ResumableFileDownload.java
cenedhryn: I am not sure what you mean with toBuilder? I added the consumer method for more fluent construction of a ResumableFileDownload instance, because it was bugging me that within the DownloadFileRequest, it was possible to add a consumer of GetObjectRequest.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/utils/ResumableFileDownloadSerializer.java
cenedhryn: Is that more consistent with our standard? What is our standard for these types of messages, would you say?

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/utils/ResumableFileDownloadSerializer.java
cenedhryn: Lol, didn't see that one coming! Sure!

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/utils/ResumableFileDownloadSerializer.java
cenedhryn: It's not a switch compatible type, so I'd have to convert to string which in this case seemed awkward but will look into it with the copy rewrite.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/utils/ResumableFileDownloadSerializer.java
zoewangg: On second thoughts, we don't know what fields `ResumableFileUpload` will have right now... Ignore my comment, we can refactor this class when we get to it.

core/protocols/aws-json-protocol/src/main/java/software/amazon/awssdk/protocols/json/internal/unmarshall/JsonProtocolUnmarshaller.java
cenedhryn: Have restored this file, and copied relevant code locally instead.

core/protocols/aws-json-protocol/src/main/java/software/amazon/awssdk/protocols/json/internal/unmarshall/JsonUnmarshallerRegistry.java
cenedhryn: Have restored this file, and copied relevant code locally instead.

core/protocols/aws-json-protocol/src/main/java/software/amazon/awssdk/protocols/json/internal/unmarshall/SimpleTypeJsonUnmarshaller.java
cenedhryn: Deleted this file.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/ResumableFileDownload.java
cenedhryn: Updated all APIs with checked exceptions to throw SdkClientException with a relevant message.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/ResumableFileDownload.java
cenedhryn: Yes, it makes sense. 

I included a `toString(Charset)` method too, because we have a fromString(Charset) method and thought symmetry was better. Do you think it's inconsistent if it's called `toString`?

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/utils/ResumableFileDownloadSerializer.java
cenedhryn: I extracted as much generic code as possible to a separate class, so a `ResumableFileUpload` serializer should be fairly quick to write.

core/json-utils/src/main/java/software/amazon/awssdk/protocols/jsoncore/JsonWriter.java
```diff
@@ -0,0 +1,299 @@
...
+        return this;
+    }
+
+    public JsonWriter writeValue(BigDecimal value) {
+        try {
+            /**
+             * Note that this is not how the backend represents BigDecimal types. On the wire
+             * it's normally a JSON number but this causes problems with certain JSON implementations
+             * that parse JSON numbers as floating points automatically. (See API-433)
+             */
```
zoewangg, line 205: Is this still relevant?

core/json-utils/src/main/java/software/amazon/awssdk/protocols/jsoncore/JsonWriter.java
```diff
@@ -0,0 +1,299 @@
...
+        /**
+         * The {@link JsonFactory} implementation to be used when parsing the input. This allows JSON extensions like CBOR or
+         * Ion to be supported.
+         *
+         * <p>It's highly recommended us use a shared {@code JsonFactory} where possible, so they should be stored statically:
+         * http://wiki.fasterxml.com/JacksonBestPracticesPerformance
+         *
+         * <p>By default, this is {@link #DEFAULT_JSON_FACTORY}.
+         */
+        public JsonWriter.Builder jsonFactory(JsonFactory jsonFactory) {
```
zoewangg, line 277: Is this API needed?

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/serialization/TransferManagerJsonMarshaller.java
```diff
@@ -0,0 +1,117 @@
...
+    TransferManagerJsonMarshaller<Integer> INTEGER = (val, jsonGenerator) -> jsonGenerator.writeValue(val);
+    TransferManagerJsonMarshaller<Long> LONG = (val, jsonGenerator) -> jsonGenerator.writeValue(val);
+    TransferManagerJsonMarshaller<Float> FLOAT = (val, jsonGenerator) -> jsonGenerator.writeValue(val);
+    TransferManagerJsonMarshaller<Double> DOUBLE = (val, jsonGenerator) -> jsonGenerator.writeValue(val);
+    TransferManagerJsonMarshaller<BigDecimal> BIG_DECIMAL = (val, jsonGenerator) -> jsonGenerator.writeValue(val);
+    TransferManagerJsonMarshaller<Boolean> BOOLEAN = (val, jsonGenerator) -> jsonGenerator.writeValue(val);
+    TransferManagerJsonMarshaller<Instant> INSTANT =
+        (val, jsonGenerator) -> jsonGenerator.writeNumber(DateUtils.formatUnixTimestampInstant(val));
+    TransferManagerJsonMarshaller<SdkBytes> SDK_BYTES = (val, jsonGenerator) -> jsonGenerator.writeValue(val.asByteBuffer());
+
```
zoewangg, line 52: Can we just use `writeValue(val)`?

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/serialization/TransferManagerMarshallingUtils.java
```diff
@@ -0,0 +1,127 @@
...
+        }
+        return (TransferManagerJsonUnmarshaller<Object>) unmarshaller;
+    }
+
+    public static SdkField<?> getSdkField(String key) {
+        return GetObjectRequest.builder().build()
+                               .sdkFields().stream()
+                               .filter(f -> f.locationName().equalsIgnoreCase(key))
+                               .findFirst()
+                               .orElseThrow(() -> new IllegalStateException("Could not match a field in GetObjectRequest"));
```
zoewangg, line 124: This seems a bit inefficient and `GetObjectRequest` has tons of fields. Can we store the sdk fields in a map?

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/ResumableFileDownload.java
```diff
@@ -125,6 +152,153 @@ public Optional<Long> totalSizeInBytes() {
...
+    }
+
+    /**
+     * Persists this download object to a file in Base64-encoded JSON format.
+     *
+     * @param path The path to the file to which you want to write the serialized download object.
+     */
+    public void writeToFile(Path path) {
+        try {
+            Files.write(path, toBase64Bytes(ResumableFileDownloadSerializer.toJson(this)));
```
zoewangg, line 173: It seems we are serializing every time this method is invoked. Same for other ser/deser methods. I think it might be fine for now and we can optimize later if customers report issue. WDYT?

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/ResumableFileDownload.java
```diff
@@ -125,6 +152,153 @@ public Optional<Long> totalSizeInBytes() {
...
+     * Returns the serialized JSON data representing this object as an {@link SdkBytes} object.
+     *
+     * @return the serialized JSON as {@link SdkBytes}
+     */
+    public SdkBytes toBytes() {
+        return SdkBytes.fromByteArray(ResumableFileDownloadSerializer.toJson(this));
+    }
+
+    /**
+     * Returns the serialized JSON data representing this object as a {@link ByteArrayInputStream}.
```
zoewangg, line 222: nit: IMO `ByteArrayInputStream` is implementation detail and should not be included in the docs

zoewangg (commented):
Can we add the sample code in `S3TransferManager#resumeDownloadFile`?

core/json-utils/src/main/java/software/amazon/awssdk/protocols/jsoncore/JsonWriter.java
```diff
@@ -0,0 +1,299 @@
...
+     * The default {@link JsonFactory} used for {@link #create()} or if a factory is not configured via
+     * {@link JsonNodeParser.Builder#jsonFactory(JsonFactory)}.
+     */
+    public static final JsonFactory DEFAULT_JSON_FACTORY = JsonFactory.builder().build();
+
+    private static final int DEFAULT_BUFFER_SIZE = 1024;
+    private final ByteArrayOutputStream baos = new ByteArrayOutputStream(DEFAULT_BUFFER_SIZE);
+    private final JsonGenerator generator;
+
+    public JsonWriter(Builder builder) {
```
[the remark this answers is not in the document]
cenedhryn: yes

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/ResumableFileDownload.java
cenedhryn: good point

core/json-utils/src/main/java/software/amazon/awssdk/protocols/jsoncore/JsonWriter.java
cenedhryn: It's provided for the parser, JsonNodeParser, so it's here for consistency.

core/json-utils/src/main/java/software/amazon/awssdk/protocols/jsoncore/JsonWriter.java
```diff
@@ -0,0 +1,299 @@
...
+        try {
+            generator.writeBinary(BinaryUtils.copyBytesFrom(bytes));
+        } catch (IOException e) {
+            throw new JsonGenerationException(e);
+        }
+        return this;
+    }
+
+    //TODO: This date formatting is coupled to AWS's format. Should generalize it
+    public JsonWriter writeValue(Instant instant) {
```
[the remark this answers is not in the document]
cenedhryn: Hm... the comment is also copied. Maybe we can keep it as-is for now.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/serialization/TransferManagerJsonMarshaller.java
cenedhryn: I added a convenience method to the jsonwriter:
~~~
    public JsonWriter writeValue(SdkBytes value) {
        writeValue(value.asByteBuffer());
        return this;
    }

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/ResumableFileDownload.java
cenedhryn: Yes. My reasoning is that since a `ResumableFileDownload` is a transient type of object, storing the serialized version wouldn't improve performance a lot. I expect the pattern would be pause download -> serialize to the format you want once -> garbage collect.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/serialization/TransferManagerJsonMarshaller.java
cenedhryn: Actually adding this method caused type erasure conflicts, removed it.

zoewangg: Sorry, I commented on the wrong line. I meant line 49 `TransferManagerJsonMarshaller<Instant> INSTANT = ....`

core/json-utils/src/main/java/software/amazon/awssdk/protocols/jsoncore/JsonWriter.java
```diff
@@ -0,0 +1,288 @@
...
+            throw new JsonGenerationException(e);
+        }
+        return this;
+    }
+
+    public JsonWriter writeNumber(String number) {
+        try {
+            generator.writeNumber(number);
+        } catch (IOException e) {
+            throw new JsonGenerationException(e);
```
dagnir, line 219: minor: this try-catch repeats throughout, would it help readability to have a wrapper around it?

```java
unsafeWrite(() -> generator.writeNumber(number));
```

cenedhryn: I like it!

dagnir (approved):
:shipit:

zoewangg (approved):

PR #3178, opened 2022-05-05
Various performance improvements.

millems:
1. Replace uses of String.replace with StringUtils.replace and StringUtils.replaceEach to avoid the overhead of regex parsing the search term.
2. Remove uses of SdkHttpRequest and SdkHttpResponse's headers and rawQueryParameters calls in favor of more specific methods that do not require data copying. Added spotbugs rules to prevent uses of these methods in the future.
3. Update SdkHttpRequest and SdkHttpResponse to not copy data in headers and rawQueryParameters when round-tripped to/from builder unless it is actually modified.
4. Reduce string copies performed during message signing.
5. Remove uses of String.format that were performance-impacting on profiler output.
6. Remove construction of log strings within the netty client unless the logging is enabled.
7. Move most user-agent construction to client creation time instead of at request time.
8. Initialize collection/map capacities to prevent resizes that will always occur.
9. Prevent creation of RateLimitingTokenBucket unless adaptive retries are enabled.


core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/RetryableStage.java
```diff
@@ -43,7 +43,7 @@ public RetryableStage(HttpClientDependencies dependencies,
                           RequestPipeline<SdkHttpFullRequest, Response<OutputT>> requestPipeline) {
         this.dependencies = dependencies;
         this.requestPipeline = requestPipeline;
-        this.rateLimitingTokenBucket = new RateLimitingTokenBucket();
+        this.rateLimitingTokenBucket = null;
```
millems, line 46: Since you're reviewing, @dagnir: Isn't the RetryableStage created with each request, so this rate limiting token bucket is never reused between requests, making it essentially worthless?

dagnir:
Awesome stuff!

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/RetryableStage.java
millems: I didn't fix that here. I can create a backlog item to fix it...

core/auth/src/main/java/software/amazon/awssdk/auth/signer/AsyncAws4Signer.java
```diff
@@ -69,7 +69,7 @@ public CompletableFuture<SdkHttpFullRequest> signWithBody(SdkHttpFullRequest req
             Aws4SignerRequestParams requestParams = new Aws4SignerRequestParams(signingParams);
 
             SdkHttpFullRequest.Builder builder = doSign(request, requestParams, signingParams,
-                    new ContentChecksum(digestHex, sdkChecksum));
+                                                        new ContentChecksum(digestHex, sdkChecksum));
```
chibenwa, line 72: Is the indent change here intended? Or an IDE fail?

core/auth/src/main/java/software/amazon/awssdk/auth/signer/internal/AbstractAws4Signer.java
```diff
@@ -333,52 +317,21 @@ private void addPreSignInformationToRequest(SdkHttpFullRequest.Builder mutableRe
...
+    /**
+     * Tests a char to see if is it whitespace.
+     * This method considers the same characters to be white
+     * space as the Pattern class does when matching \s
+     *
+     * @param ch the character to be tested
+     * @return true if the character is white  space, false otherwise.
+     */
+    private boolean isWhiteSpace(final char ch) {
+        return ch == ' ' || ch == '\t' || ch == '\n' || ch == '\u000b' || ch == '\r' || ch == '\f';
```
chibenwa, line 334: Guava's charmatcher allows denser checks, with a binary search and binary sets, which might be handy with 6 + chars to check for.... Especially if in the future you decide to add more.

core/auth/src/main/java/software/amazon/awssdk/auth/signer/internal/AbstractAws4Signer.java
```diff
@@ -426,7 +379,7 @@ private long getSignatureDurationInSeconds(Aws4SignerRequestParams requestParams
     private byte[] newSigningKey(AwsCredentials credentials,
                                  String dateStamp, String regionName, String serviceName) {
         byte[] kSecret = ("AWS4" + credentials.secretAccessKey())
-            .getBytes(Charset.forName("UTF-8"));
+            .getBytes(StandardCharsets.UTF_8);
```
chibenwa, line 382: Does this needs to be recomputed all of the time? Can't the credentials holds kSecret?

core/aws-core/src/main/java/software/amazon/awssdk/awscore/client/handler/AwsClientHandlerUtils.java
```diff
@@ -42,11 +42,9 @@ private AwsClientHandlerUtils() {
      * @return A bytebuffer representing the given request
      */
     public static ByteBuffer encodeEventStreamRequestToByteBuffer(SdkHttpFullRequest request) {
-        Map<String, HeaderValue> headers = request.headers()
-                                                  .entrySet()
-                                                  .stream()
-                                                  .collect(Collectors.toMap(Map.Entry::getKey, e -> HeaderValue.fromString(
-                                                      firstIfPresent(e.getValue()))));
+        Map<String, HeaderValue> headers = new LinkedHashMap<>();
```
chibenwa, line 45: Anything preventing us to pre-size this hash map?

core/protocols/aws-query-protocol/src/main/java/software/amazon/awssdk/protocols/query/internal/unmarshall/AwsQueryResponseHandler.java
```diff
@@ -84,12 +83,10 @@ private T unmarshallResponse(SdkHttpFullResponse response) throws Exception {
...
     private AwsResponseMetadata generateResponseMetadata(SdkHttpResponse response, Map<String, String> metadata) {
         if (!metadata.containsKey(AWS_REQUEST_ID)) {
-            metadata.put(AWS_REQUEST_ID,
-                         SdkHttpUtils.firstMatchingHeaderFromCollection(response.headers(), X_AMZN_REQUEST_ID_HEADERS)
-                                     .orElse(null));
+            metadata.put(AWS_REQUEST_ID, response.firstMatchingHeader(X_AMZN_REQUEST_ID_HEADERS).orElse(null));
         }
 
-        response.headers().forEach((key, value) -> metadata.put(key, value.get(0)));
+        response.forEachHeader((key, value) -> metadata.put(key, value.get(0)));
```
chibenwa, line 89: is firstMathcingHeader still iterating on all headers?

If so can't we concider doing one pass instead?

core/auth/src/main/java/software/amazon/awssdk/auth/signer/AsyncAws4Signer.java
millems: Not intended, but more of an IDE success. The indent is an improvement, imo.

core/auth/src/main/java/software/amazon/awssdk/auth/signer/internal/AbstractAws4Signer.java
millems: I considered doing something like a 256 byte array lookup table for chars being whitespace (with some special handling for unicode), but this method didn't show up very high on a profiler so it felt like premature optimization. It would be nice if we could have a Guava dependency, but since we have to implement it ourselves (or deal with importing it), I wasn't inclined.

core/auth/src/main/java/software/amazon/awssdk/auth/signer/internal/AbstractAws4Signer.java
millems: That's a cool idea. We'd probably want to use a weak hash map from credential to cached values for that credential so that we don't leak that information out of the signer, but that could definitely save some computation for reused credentials.

core/aws-core/src/main/java/software/amazon/awssdk/awscore/client/handler/AwsClientHandlerUtils.java
millems: Nope, good catch!

core/auth/src/main/java/software/amazon/awssdk/auth/signer/internal/AbstractAws4Signer.java
chibenwa: I thought `credentials` was immutable when commenting. If mutability is in the game it would ruin potentially ruin gains...

PR #3184, opened 2022-05-06
Add CopyableBuilder and ToCopyableBuilder interfaces to the credential providers.

### LINKED ISSUE #3129: AwsCredentialsProvider+Builder types are not instances of Buildable interface
### Describe the issue

The SDK declares a very open-ended `Buildable` interface and many of the `*Builder` types among the APIs implement it. However, the builders for various `AwsCredentialsProvider` implementations, do not:

* `AwsCredentialsProviderChain.Builder`
* `ContainerCredentialsProvider.Builder` (this one implements a different, generic, interface that defines `#build` but I think _that_ interface could extend `Buildable` with no compilation problem)
* `DefaultCredentialsProvider.Builder` (this one being a class instead of an interface challenges any sense of order and convention here)
* `InstanceProfileCredentialsProvider.Builder`
* `ProfileCredentialsProvider.Builder`

I have not checked them all, but I can say that I have not yet found any credentials provider type whose Builder _does_ extend/implement `Buildable`.

### Steps to Reproduce

N/A

### Current behavior

Everything, of course, works, but implementing this interface would help consumers of the API in working with its common patterns.

### AWS Java SDK version used

2.17.155

### JDK version used

java version "1.8.0_231" Java(TM) SE Runtime Environment (build 1.8.0_231-b11) Java HotSpot(TM) 64-Bit Server VM (build 25.231-b11, mixed mode)

macOS Monterey Version 12.3

millems:
Fixes #3129


cenedhryn (commented):
Looks good, but I don't see any tests

millems:
We don't traditionally test the toBuilder methods on any of our builders. Do you think we should test each of these?

cenedhryn:
I'm a bit surprised about that actually! It's easy to introduce bugs when converting to builder, both when the method is added as well as over time if the instance variables are expanded. 

```
public BaseBuilder(Function<B, T> providerConstructor, StsCredentialsProvider provider) {
            this.providerConstructor = providerConstructor;
            this.asyncCredentialUpdateEnabled = provider.asyncCredentialUpdateEnabled;
            this.stsClient = provider.stsClient;
           // this.staleTime = provider.staleTime;   <-- OOPS forgot to add the stale time here -->
            this.prefetchTime = provider.prefetchTime;
 }
```

Pretty sure I've seen bugs in similar cases before. You can consider it, it's not a blocker.

millems:
This was merged as part of https://github.com/aws/aws-sdk-java-v2/pull/3267

PR #3187, opened 2022-05-09
Upgrade Netty version to 4.1.77.Final

millems:


dagnir (approved):

analytically:
@millems would be good to upgrade to latest Netty again

PR #3195, opened 2022-05-16
Move S3 CRT related class out of s3-tranfer-manager module to s3 module

zoewangg:
## Modifications
Move S3 CRT related class out of s3-tranfer-manager module to s3 module

## Testing
Integ tests passed

### DIFF
--- a/services/s3/pom.xml
+++ b/services/s3/pom.xml
@@ -75,6 +75,12 @@
             <artifactId>profiles</artifactId>
             <version>${awsjavasdk.version}</version>
         </dependency>
+        <dependency>
+            <groupId>software.amazon.awssdk.crt</groupId>
+            <artifactId>aws-crt</artifactId>
+            <version>${awscrt.version}</version>
+            <scope>provided</scope>
+        </dependency>
         <!-- Test Dependencies -->
         <dependency>
             <artifactId>commons-io</artifactId>
--- a/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/AbstractS3CrtAsyncClient.java
+++ b/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/AbstractS3CrtAsyncClient.java
@@ -1,0 +1,806 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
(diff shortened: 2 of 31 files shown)


services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/S3CrtAsyncClient.java
```diff
@@ -13,7 +13,7 @@
  * permissions and limitations under the License.
  */
 
-package software.amazon.awssdk.transfer.s3.internal;
+package software.amazon.awssdk.services.s3.internal.crt;
```
zoewangg, line 16: We will make this class public API before GA

services/s3/pom.xml
```diff
@@ -75,6 +75,12 @@
             <artifactId>profiles</artifactId>
             <version>${awsjavasdk.version}</version>
         </dependency>
+        <dependency>
+            <groupId>software.amazon.awssdk.crt</groupId>
+            <artifactId>aws-crt</artifactId>
+            <version>${awscrt.version}</version>
+            <scope>provided</scope>
```
zoewangg, line 82: Dependencies of `provided` scope are not transitive.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/AbstractS3CrtAsyncClient.java
```diff
@@ -0,0 +1,806 @@
...
+import software.amazon.awssdk.services.s3.model.WriteGetObjectResponseResponse;
+import software.amazon.awssdk.services.s3.paginators.ListMultipartUploadsPublisher;
+import software.amazon.awssdk.services.s3.paginators.ListObjectVersionsPublisher;
+import software.amazon.awssdk.services.s3.paginators.ListObjectsV2Publisher;
+import software.amazon.awssdk.services.s3.paginators.ListPartsPublisher;
+import software.amazon.awssdk.services.s3.waiters.S3AsyncWaiter;
+
+@SdkInternalApi
+public abstract class AbstractS3CrtAsyncClient implements S3CrtAsyncClient {
+    abstract S3AsyncClient s3AsyncClient();
```
dagnir, line 220: Per offline discussion, should be passed as a c-tor param

zoewangg: Yeah, this is the first thing I tried when I was implementing it, but it doesn't work here because we need to instantiate `S3AsyncClient` first in the ctor of `DefaultS3CrtAsyncClient` (before `super(client)`). Another option is to use composition over inheritance, but we'd need to write the same amount of operations in the `DefaultS3CrtAsyncClient`

zoewangg: To close the loop, we decided to use `super(someStaticMethodToInitializeS3Client())`

dagnir (approved):

PR #3197, opened 2022-05-17
Updates test case to not rely on ordering of result, which is differe…

cenedhryn:
…nt in JDK8 compared to later versions

## Motivation and Context
Seems to originate here:
https://github.com/aws/aws-sdk-java-v2/blob/master/services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/mapper/StaticTableMetadata.java#L244-L245


cenedhryn:
discard

Issue #2942, opened 2021-12-30
 S3TransferManager creation fails on Apple's M1


AdamBien:
### Describe the bug

Instantiation of S3TransferManager fails

### Expected behavior

S3TransferManager is instantiated, file is uploaded.

### Current behavior

The instantiation of fails. 

var presigner = S3Presigner.builder().build();
var credentialsProvider = StaticCredentialsProvider.create(credentials);
var configuration = S3ClientConfiguration
        .builder()
        .region(Region.EU_CENTRAL_1)
        .credentialsProvider(credentialsProvider)
    .build();

This file below fails:
var transferManager = S3TransferManager.builder().s3ClientConfiguration(configuration).build();

with:

Unable to unpack AWS CRT lib: java.io.IOException: Unable to open library in jar for AWS CRT: /osx/armv8/libaws-crt-jni.dylib
java.io.IOException: Unable to open library in jar for AWS CRT: /osx/armv8/libaws-crt-jni.dylib
        at software.amazon.awssdk.crt.CRT.extractAndLoadLibrary(CRT.java:158)
        at software.amazon.awssdk.crt.CRT.loadLibraryFromJar(CRT.java:223)
        at software.amazon.awssdk.crt.CRT.<clinit>(CRT.java:33)
        at software.amazon.awssdk.crt.CrtResource.<clinit>(CrtResource.java:100)
        at software.amazon.awssdk.transfer.s3.internal.S3NativeClientConfiguration.<init>(S3NativeClientConfiguration.java:58)
        at software.amazon.awssdk.transfer.s3.internal.S3NativeClientConfiguration$Builder.build(S3NativeClientConfiguration.java:194)
        at software.amazon.awssdk.transfer.s3.internal.DefaultS3CrtAsyncClient.<init>(DefaultS3CrtAsyncClient.java:54)
        at software.amazon.awssdk.transfer.s3.internal.DefaultS3CrtAsyncClient$DefaultS3CrtClientBuilder.build(DefaultS3CrtAsyncClient.java:224)
        at software.amazon.awssdk.transfer.s3.internal.DefaultS3TransferManager.initializeS3CrtClient(DefaultS3TransferManager.java:95)
        at software.amazon.awssdk.transfer.s3.internal.DefaultS3TransferManager.<init>(DefaultS3TransferManager.java:60)
[...]

PR #3209, opened 2022-05-23
Bump CRT version to 0.16.10, which contains M1 support #2942

zoewangg:
## Motivation and Context
Bump CRT version to 0.16.10, which contains M1 support #2942

## Testing
~~Running test suites now~~ integ tests and stability tests passed

### DIFF
--- a/services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/S3CrtAsyncHttpClient.java
+++ b/services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/S3CrtAsyncHttpClient.java
@@ -67,6 +67,7 @@ public final class S3CrtAsyncHttpClient implements SdkAsyncHttpClient {
                                  .withCredentialsProvider(s3NativeClientConfiguration.credentialsProvider())
                                  .withClientBootstrap(s3NativeClientConfiguration.clientBootstrap())
                                  .withPartSize(s3NativeClientConfiguration.partSizeBytes())
+                                 .withComputeContentMd5(true)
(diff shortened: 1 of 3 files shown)


services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/S3CrtAsyncHttpClient.java
```diff
@@ -67,6 +67,7 @@ private S3CrtAsyncHttpClient(Builder builder) {
                                  .withCredentialsProvider(s3NativeClientConfiguration.credentialsProvider())
                                  .withClientBootstrap(s3NativeClientConfiguration.clientBootstrap())
                                  .withPartSize(s3NativeClientConfiguration.partSizeBytes())
+                                 .withComputeContentMd5(true)
```
zoewangg, line 70: We had to make this change because CRT has pushed fixes to honor this config in latest version. In previous versions, MD5 is always enabled and computeContentMd5 is default to false in the CRT Java binding. We will disable it once we have flexible checksum support.

millems (approved):

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/S3CrtAsyncHttpClient.java
millems: Is it imported to brazil, or will this block our build to live?

zoewangg: It should not block our build to live because this config is available in previous CRT version as well (just not honored). I'll check with CRT team to see when they will do an import.

PR #3252, opened 2022-06-15
Added customization in codegen to generate additional builder methods

joviegas:
## Motivation and Context

- Provide API in  S3AsyncClient to create instance of  S3CrtAsyncClient .

#### Customization.config
```json
  "additionalBuilderMethods": [
      {
        "methodName": "crtBuilder",
        "clientTypes": ["async"],
        "instanceType": "software.amazon.awssdk.services.s3.internal.crt.S3CrtAsyncClient",
        "returnType": "software.amazon.awssdk.services.s3.internal.crt.S3CrtAsyncClient.S3CrtAsyncClientBuilder"
      }
  ],
```

#### S3AsyncClient.java
```java

    /**
     * Create a builder that can be used to configure and create a {@link S3CrtAsyncClient}
     */
    static S3CrtAsyncClientBuilder crtBuilder() {
        return S3CrtAsyncClient.builder();
    }
```

## Modifications

- Added custom config in codegen to generate additional builders if specified.

## Testing

- Junits
- Integ test updated with the new builder.

## Types of changes

## License

### DIFF
--- a/services/s3/src/main/resources/codegen-resources/customization.config
+++ b/services/s3/src/main/resources/codegen-resources/customization.config
@@ -172,5 +172,13 @@
       "clientConfiguration"
     ]
   },
+  "additionalBuilderMethods": [
+      {
+        "methodName": "crtBuilder",
+        "clientType": "async",
+        "instanceType": "software.amazon.awssdk.services.s3.internal.crt.S3CrtAsyncClient",
+        "returnType": "software.amazon.awssdk.services.s3.internal.crt.S3CrtAsyncClient.S3CrtAsyncClientBuilder"
+      }
+  ],
   "delegateAsyncClientClass": true
 }
--- a/codegen/src/main/java/software/amazon/awssdk/codegen/model/config/customization/AdditionalBuilderMethod.java
+++ b/codegen/src/main/java/software/amazon/awssdk/codegen/model/config/customization/AdditionalBuilderMethod.java
@@ -1,0 +1,98 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.codegen.model.config.customization;
+
+import java.util.Locale;
+import software.amazon.awssdk.core.ClientType;
+
+/**
+ * Config required to generate the additional static methods in Client interface.
+ */
+public class AdditionalBuilderMethod {
+
+    /**
+     * Name of the additional static method.
+     */
+    private String methodName;
+
+    /**
+     * Fqcn of the return type
+     */
+    private String returnType;
+
+    /**
+     * Fqcn of the class that will delegate the static method call
+     */
+    private String instanceType;
+
+    /**
+     * JavaDoc for the method
+     */
+    private String javaDoc;
+
+    /**
+     * The clientType for which the builder needs to be added.
+     */
+    private ClientType clientType;
+
+    public String getReturnType() {
+        return returnType;
+    }
+
+    public void setReturnType(String returnType) {
+        this.returnType = returnType;
+    }
+
+    public String getJavaDoc() {
+        return javaDoc;
+    }
+
+    public void setJavaDoc(String javaDoc) {
+        this.javaDoc = javaDoc;
+    }
+
+    public String getMethodName() {
+        return methodName;
+    }
+
+    public void setMethodName(String methodName) {
+        this.methodName = methodName;
+    }
+
+    public String getInstanceType() {
+        return instanceType;
+    }
+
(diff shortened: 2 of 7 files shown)


services/s3/src/main/resources/codegen-resources/customization.config
```diff
@@ -172,5 +172,14 @@
       "clientConfiguration"
     ]
   },
+  "additionalBuilderMethods": [
+      {
+        "methodName": "crtClientBuilder",
```
zoewangg, line 177: `crtBuilder`?

services/s3/src/main/resources/codegen-resources/customization.config
```diff
@@ -172,5 +172,14 @@
       "clientConfiguration"
     ]
   },
+  "additionalBuilderMethods": [
```
zoewangg, line 175: Can we specify this is for async only, something like `additionalBuilderMethodsForAsyncClient`?

services/s3/src/main/resources/codegen-resources/customization.config
```diff
@@ -172,5 +172,14 @@
...
+      {
+        "methodName": "crtClientBuilder",
+        "delegateBuilder": "builder",
```
zoewangg, line 178: Do we need `delegateBuilder`? Can we assume the name is `builder`? We can add this customization if needed in the future, but I doubt we will ever need it.

services/s3/src/main/resources/codegen-resources/customization.config
```diff
@@ -172,5 +172,14 @@
...
+        "javaDoc": "{@link S3CrtAsyncClientBuilder} to create instance of {@link S3CrtAsyncClient}",
```
zoewangg, line 179: Do we need javadoc field? Can we construct the Javadoc based on existing information? We can have something similar to the javadocs for the standard builder method.
```
 Create a builder that can be used to configure and create a {@link S3CrtAsyncClient}
```

services/s3/src/main/resources/codegen-resources/customization.config
cenedhryn: Why is it restricted to async?

zoewangg: Because S3 CRT client is an async client and we don't support it for sync code path for now

cenedhryn: This use case, sure, but the additionalBuilderMethods functionality looks general enough to be used for a sync use case too.

zoewangg: We only need it for async client though for now. I guess we can add a flag in the `additionalBuilderMethods` to indicate if it's for sync or async. I'm fine with either way since it's an internal API.

services/s3/src/main/resources/codegen-resources/customization.config
joviegas: done

services/s3/src/main/resources/codegen-resources/customization.config
joviegas: Added a new field to indicate the clientType supported

services/s3/src/main/resources/codegen-resources/customization.config
joviegas: sure.. removed delegateBuilder

services/s3/src/main/resources/codegen-resources/customization.config
joviegas: done

services/s3/src/main/resources/codegen-resources/customization.config
```diff
@@ -172,5 +172,13 @@
       "clientConfiguration"
     ]
   },
+  "additionalBuilderMethods": [
+      {
+        "methodName": "crtBuilder",
+        "clientTypes": ["async"],
```
zoewangg, line 178: minor: should this be singular `clientType` since we know for sure that sync and async don't share the same instanceType and returnType?

joviegas: IMO keeping clientTypes will give us flexibility to add some builders of CustomLibraries which needs to be both in Sync and Async in future.

zoewangg: Yeah, but sync and async clients will most likely have different `instanceType`s and `returnType`s, eg: `S3AsyncClientBuilder` and `S3ClientBuilder`, so I think we will probably never use ` "clientTypes": ["async", "sync"]` in reality and we will probably have something like below:


```
"additionalBuilderMethods": [
        {
          "methodName": "builderOne",
          "clientTypes": ["sync"],
          "instanceType": "software.amazon.awssdk.services.builder.DefaultBuilder",
          "returnType": "software.amazon.awssdk.services.builder.CustomBuilder"
        },
        {
          "methodName": "builderOne",
          "clientTypes": ["async"],
          "instanceType": "software.amazon.awssdk.services.builder.DefaultAsyncBuilder",
          "returnType": "software.amazon.awssdk.services.builder.CustomAsyncBuilder"
        }
```

joviegas: Done

services/s3/src/main/resources/codegen-resources/customization.config
```diff
@@ -172,5 +172,13 @@
       "clientConfiguration"
     ]
   },
+  "additionalBuilderMethods": [
+      {
+        "methodName": "crtBuilder",
+        "clientType": "async",
```
zoewangg, line 178: Is it case sensitive?

zoewangg (approved):

PR #3258, opened 2022-06-17
S3Object based DownloadFilter and removing DownloadFileContext as destination based filter is removed

joviegas:
## Motivation and Context

- S3Object based DownloadFilter and removing DownloadFileContext as destination based filter is removed
- S3 Key can be used to do the filteration instead of Destination path,

## Modifications

- DownloadFileContext is removed.

## Testing

- Updated test  cases


services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/DownloadFilter.java
```diff
@@ -29,17 +29,16 @@
  */
 @SdkPublicApi
 @SdkPreviewApi
-public interface DownloadFilter extends Predicate<DownloadFileContext> {
+public interface DownloadFilter extends Predicate<S3Object> {
 
     /**
      * Evaluate whether the remote {@link S3Object} should be downloaded to the destination {@link Path}.
```
zoewangg, line 35: We need to update Javadocs here as well

zoewangg (approved):
Can we add a changelog entry to mention this breaking change?

PR #3264, opened 2022-06-22
Bump CRT to 0.16.16

zoewangg:
## Motivation and Context
CRT changed the API as part of https://github.com/awslabs/aws-crt-java/pull/458 
This PR bumps up the CRT version and fixes the compilation error

## Testing
Running test suites now


zoewangg:
No, the change is to support flexible checksum and they talked with us before the change and we agreed on the change.

PR #3267, opened 2022-06-23
Add missing CopyableBuilder interfaces and fixed buggy ones.

millems:
Add CopyableBuilder and ToCopyableBuilder interfaces to credential providers. Also added spotbugs rule to validate that these and all other toBuilder methods are correct. Fixed any issues found with existing methods.

### DIFF
--- a/build-tools/src/main/java/software/amazon/awssdk/buildtools/findbugs/ToBuilderIsCorrect.java
+++ b/build-tools/src/main/java/software/amazon/awssdk/buildtools/findbugs/ToBuilderIsCorrect.java
@@ -1,0 +1,456 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.buildtools.findbugs;
+
+import static java.util.Collections.emptyList;
+
+import edu.umd.cs.findbugs.BugInstance;
+import edu.umd.cs.findbugs.BugReporter;
+import edu.umd.cs.findbugs.OpcodeStack;
+import edu.umd.cs.findbugs.ba.XField;
+import edu.umd.cs.findbugs.ba.XMethod;
+import edu.umd.cs.findbugs.bcel.OpcodeStackDetector;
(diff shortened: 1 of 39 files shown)


build-tools/src/main/java/software/amazon/awssdk/buildtools/findbugs/ToBuilderIsCorrect.java
```diff
@@ -0,0 +1,459 @@
...
+
+            boolean result = false;
+            JavaClass[] parentInterfaces = obj.getAllInterfaces();
+            for (JavaClass i : parentInterfaces) {
+                if (i.getClassName().equals("software.amazon.awssdk.utils.builder.CopyableBuilder")) {
+                    builders.put(obj.getClassName(), obj);
+                    builderParents.put(obj.getClassName(),
+                                       Stream.concat(Arrays.stream(parentInterfaces), Arrays.stream(obj.getSuperClasses()))
+                                             .map(JavaClass::getClassName)
+                                             .collect(Collectors.toSet()));
```
cenedhryn, line 168: It looks like the same code will get executed for any matching `i`. Is there a way to break it out of the loop?

core/annotations/src/main/java/software/amazon/awssdk/annotations/ToBuilderIgnoreField.java
```diff
@@ -0,0 +1,35 @@
...
+import java.lang.annotation.Target;
+
+/**
+ * Used to suppress certain fields from being considered in the spot-bugs rule for toBuilder(). This annotation must be
+ * attached to the toBuilder() method to function.
+ */
+@Target(ElementType.METHOD)
+@Retention(RetentionPolicy.CLASS)
+@SdkProtectedApi
+public @interface ToBuilderIgnoreField {
```
cenedhryn, line 30: Nice addition!

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/ProcessCredentialsProvider.java
```diff
@@ -83,9 +92,11 @@ private ProcessCredentialsProvider(Builder builder) {
 
         cmd.add(builderCommand);
 
-        this.command = Collections.unmodifiableList(cmd);
+        this.executableCommand = Collections.unmodifiableList(cmd);
```
cenedhryn, line 95: why was this renamed

millems: Mostly to differentiate it from the `commandFromBuilder`.

build-tools/src/main/java/software/amazon/awssdk/buildtools/findbugs/ToBuilderIsCorrect.java
millems: Nice catch, yes!

millems: Technically it might be slower this way with it outside of the loop, since most classes won't have a matching value and even fewer will have multiple matching values.

cenedhryn (approved):

PR #3275, opened 2022-06-29
Reduced resource consumption of async credential providers.

### LINKED ISSUE #3259: StsCredentialsProvider uses excessive number of threads in multi-tenant setup
### Describe the bug

When configuring StsCredentialsProvider (or it's subclasses, e.g. StsAssumeRoleCredentialsProvider) for async credential refreshing, an instance of software.amazon.awssdk.utils.cache.NonBlocking is created for each credentials provider.  This further creates a ScheduledThreadPoolExecutor with a single thread to handle the async refreshing of credentials.

In a multi-tenant environment where a credentials-provider-per-tenant is used (to provide scoped-down per-tenant IAM policies, e.g https://aws.amazon.com/blogs/apn/isolating-saas-tenants-with-dynamically-generated-iam-policies/) there are a proliferation of ScheduledThreadPoolExecutors (each with a single thread).  Threads are expensive resources to create and have laying around at scale.

It isn't necessary to have a thread-per-credentials-provider - a shared ScheduledThreadPoolExecutor with a small pool size would suffice.

### Expected Behavior

Scaling the use of StsCredentialsProvider in a multi-tenant environment doesn't consume excessive/unnecessary thread resources.

### Current Behavior

A background thread is created for each instance of StsCredentialsProvider for async refresh of credentials.  Threads are expensive resources to create and have laying around at scale.  Each thread consumes memory (thread stack), and there are  hard limits to the number of threads that can be created (these vary based on OS, configuration, and other use of threads inside an app).

### Reproduction Steps

n/a.  This is a non-functional/scalability design defect, evident in a cursory review of the StsCredentialsProvider code.

### Possible Solution

Allow for StsCredentialsProvider builder to take an (optional) ScheduledThreadPoolExecutor (and remove cleanup logic in NonBlocking for externally provided Executors).

This allows consumers, in advanced use cases, to manage the background threads efficiently and avoid resource-starvation scenarios at scale.

_No response_

### AWS Java SDK version used

2.7.214

### JDK version used

17

Mac OS Catalina

millems:
Fixes #3259.

1. Share thread pools across async credential providers (anything using CachedSupplier's NonBlocking prefetch strategy).
2. Log a warning if an extreme number of concurrent refreshes are happening, to help users detect when they're not closing their credential providers.

Even though this is an increase in resource sharing, it should not cause increased availability risks. Because these threads are only used for background refreshes, if one particular type of credential provider has availability problems (e.g. SSO or STS high latency), it only disables background refreshes, not prefetches or synchronous fetches.

### DIFF
--- a/utils/src/main/java/software/amazon/awssdk/utils/cache/NonBlocking.java
+++ b/utils/src/main/java/software/amazon/awssdk/utils/cache/NonBlocking.java
@@ -15,17 +15,26 @@
 
 package software.amazon.awssdk.utils.cache;
 
-import static java.util.concurrent.TimeUnit.MILLISECONDS;
+import static java.time.temporal.ChronoUnit.HOURS;
+import static java.time.temporal.ChronoUnit.MINUTES;
 
 import java.time.Duration;
-import java.util.concurrent.ScheduledExecutorService;
+import java.time.Instant;
+import java.util.Optional;
+import java.util.Random;
+import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.ScheduledThreadPoolExecutor;
+import java.util.concurrent.SynchronousQueue;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Supplier;
 import software.amazon.awssdk.annotations.SdkProtectedApi;
 import software.amazon.awssdk.annotations.SdkTestInternalApi;
 import software.amazon.awssdk.utils.Logger;
 import software.amazon.awssdk.utils.ThreadFactoryBuilder;
-import software.amazon.awssdk.utils.Validate;
 
 /**
  * A {@link CachedSupplier.PrefetchStrategy} that will run a single thread in the background to update the value. A call to
@@ -38,79 +47,203 @@ public class NonBlocking implements CachedSupplier.PrefetchStrategy {
     private static final Logger log = Logger.loggerFor(NonBlocking.class);
 
     /**
+     * The maximum number of concurrent refreshes before we start logging warnings and skipping refreshes.
+     */
+    private static final int MAX_CONCURRENT_REFRESHES = 100;
+
+    /**
+     * The {@link Random} instance used for calculating jitter of the background prefetches.
+     */
+    private static final Random JITTER_RANDOM = new Random();
+
+    /**
+     * Thread used to kick off refreshes during the prefetch window. This does not do the actual refreshing. That's left for
+     * the {@link #EXECUTOR}.
+     */
+    private static final ScheduledThreadPoolExecutor SCHEDULER =
+        new ScheduledThreadPoolExecutor(1, new ThreadFactoryBuilder().daemonThreads(true).build());
+
+    /**
+     * Threads used to do the actual work of refreshing the values (because the cached supplier might block, so we don't
+     * want the work to be done by a small thread pool). This executor is created as unbounded, but we start complaining and
+     * skipping refreshes when there are more than {@link #MAX_CONCURRENT_REFRESHES} running.
+     */
+    private static final ThreadPoolExecutor EXECUTOR =
+        new ThreadPoolExecutor(0, Integer.MAX_VALUE,
+                               60L, TimeUnit.SECONDS,
+                               new SynchronousQueue<>(),
+                               new ThreadFactoryBuilder().daemonThreads(true).build());
+
+    /**
+     * An incrementing number, used to uniquely identify an instance of NonBlocking in the {@link #asyncThreadName}.
+     */
+    private static final AtomicLong THREAD_NUMBER = new AtomicLong(0);
+
+    /**
      * Whether we are currently refreshing the supplier. This is used to make sure only one caller is blocking at a time.
      */
-    private final AtomicBoolean currentlyRefreshing = new AtomicBoolean(false);
+    private final AtomicBoolean currentlyPrefetching = new AtomicBoolean(false);
 
     /**
-     * How frequently to automatically refresh the supplier in the background.
+     * Name of the thread refreshing the cache for this strategy.
      */
-    private final Duration asyncRefreshFrequency;
+    private final String asyncThreadName;
 
     /**
-     * Single threaded executor to asynchronous refresh the value.
+     * The refresh task currently scheduled for this non-blocking instance. We ensure that no more than one task is scheduled
+     * per instance.
      */
-    private final ScheduledExecutorService executor;
+    private final AtomicReference<ScheduledFuture<?>> refreshTask = new AtomicReference<>();
+
+    /**
+     * The minimum amount of time allowed between async refreshes, primarily adjustable for testing purposes.
+     */
+    private final Duration minimumRefreshFrequency;
+
+    /**
+     * Whether this strategy has been shutdown (and should stop doing background refreshes)
+     */
+    private volatile boolean shutdown = false;
+
+    /**
+     * The cached supplier using this non-blocking instance.
+     */
+    private volatile CachedSupplier<?> cachedSupplier;
+
+    static {
+        // Ensure that cancelling a task actually removes it from the queue.
+        SCHEDULER.setRemoveOnCancelPolicy(true);
+    }
 
     /**
      * Create a non-blocking prefetch strategy that uses the provided value for the name of the background thread that will be
      * performing the update.
      */
     public NonBlocking(String asyncThreadName) {
-        this(asyncThreadName, Duration.ofMinutes(1));
+        this(asyncThreadName, Duration.ofSeconds(60));
     }
 
     @SdkTestInternalApi
-    NonBlocking(String asyncThreadName, Duration asyncRefreshFrequency) {
-        this.executor = newExecutor(asyncThreadName);
-        this.asyncRefreshFrequency = asyncRefreshFrequency;
+    NonBlocking(String asyncThreadName, Duration minimumRefreshFrequency) {
+        this.asyncThreadName = asyncThreadName + "-" + THREAD_NUMBER.getAndIncrement();
+        this.minimumRefreshFrequency = minimumRefreshFrequency;
     }
 
-    private static ScheduledExecutorService newExecutor(String asyncThreadName) {
-        Validate.paramNotBlank(asyncThreadName, "asyncThreadName");
-        return new ScheduledThreadPoolExecutor(1, new ThreadFactoryBuilder().daemonThreads(true)
-                                                                            .threadNamePrefix(asyncThreadName)
-                                                                            .build());
+    @SdkTestInternalApi
+    static ThreadPoolExecutor executor() {
+        return EXECUTOR;
     }
 
     @Override
     public void initializeCachedSupplier(CachedSupplier<?> cachedSupplier) {
-        scheduleRefresh(cachedSupplier);
-    }
-
-    private void scheduleRefresh(CachedSupplier<?> cachedSupplier) {
-        executor.schedule(() -> {
-            try {
-                cachedSupplier.get();
-            } finally {
-                scheduleRefresh(cachedSupplier);
-            }
-        }, asyncRefreshFrequency.toMillis(), MILLISECONDS);
+        this.cachedSupplier = cachedSupplier;
     }
 
     @Override
     public void prefetch(Runnable valueUpdater) {
-        // Only run one async refresh at a time.
-        if (currentlyRefreshing.compareAndSet(false, true)) {
-            try {
-                executor.submit(() -> {
-                    try {
-                        valueUpdater.run();
-                    } catch (RuntimeException e) {
-                        log.warn(() -> "Exception occurred in AWS SDK background task.", e);
-                    } finally {
-                        currentlyRefreshing.set(false);
-                    }
-                });
-            } catch (Throwable e) {
-                currentlyRefreshing.set(false);
-                throw e;
-            }
+        // Only run one async prefetch at a time.
+        if (currentlyPrefetching.compareAndSet(false, true)) {
+            tryRunBackgroundTask(valueUpdater, () -> currentlyPrefetching.set(false));
+        }
+    }
+
+    @Override
+    public <T> RefreshResult<T> fetch(Supplier<RefreshResult<T>> supplier) {
+        RefreshResult<T> result = supplier.get();
+        if (result.staleTime() == null || result.prefetchTime() == null) {
+            return result;
+        }
+
+        getRefreshTime(result).ifPresent(this::schedulePrefetch);
+        return result;
+    }
+
+    private Optional<Instant> getRefreshTime(RefreshResult<?> result) {
+        Instant minStart = Instant.now().plus(minimumRefreshFrequency);
+        Instant rangeStart = result.prefetchTime().isBefore(minStart) ? minStart : result.prefetchTime();
+
+        if (Duration.between(Instant.now(), rangeStart).toDays() > 7) {
+            log.debug(() -> "Skipping background refresh because the prefetch time is too far in the future: " + rangeStart);
+            return Optional.empty();
+        }
+
+        Instant maxEnd = rangeStart.plus(1, HOURS);
+        Instant rangeEnd = result.staleTime().isAfter(maxEnd) ? maxEnd : result.staleTime().minus(1, MINUTES);
+
+        if (rangeEnd.isBefore(rangeStart)) {
+            return Optional.of(rangeStart);
+        }
+
+        return Optional.of(randomTimeBetween(rangeStart, rangeEnd));
+    }
+
+    private Instant randomTimeBetween(Instant rangeStart, Instant rangeEnd) {
+        Duration timeBetween = Duration.between(rangeStart, rangeEnd);
+        return rangeStart.plusMillis(Math.abs(JITTER_RANDOM.nextLong() % timeBetween.toMillis()));
+    }
+
+    private void schedulePrefetch(Instant refreshTime) {
+        if (shutdown) {
+            return;
+        }
+
+        Duration waitTime = Duration.between(Instant.now(), refreshTime);
+        log.debug(() -> "Scheduling refresh attempt for " + refreshTime + " (in " + waitTime.toMillis() + " ms)");
+        updateTask(SCHEDULER.schedule(() -> {
+            Thread.currentThread().setName(asyncThreadName + "-scheduler");
+            log.debug(() -> "Executing refresh attempt scheduled for " + refreshTime);
+
+            // If the supplier has already been prefetched, this will just be a cache hit.
+            tryRunBackgroundTask(cachedSupplier::get);
+        }, waitTime.toMillis(), TimeUnit.MILLISECONDS));
+
+        if (shutdown) {
+            updateTask(null);
         }
     }
 
     @Override
     public void close() {
-        executor.shutdown();
+        shutdown = true;
+        updateTask(null);
+    }
+
+    public void updateTask(ScheduledFuture<?> newTask) {
+        ScheduledFuture<?> currentTask;
+        do {
+            currentTask = refreshTask.get();
+            if (currentTask != null && !currentTask.isDone()) {
+                currentTask.cancel(false);
+            }
+        } while (!refreshTask.compareAndSet(currentTask, newTask));
+    }
+
+    public void tryRunBackgroundTask(Runnable runnable) {
+        tryRunBackgroundTask(runnable, () -> {
+        });
+    }
+
+    public void tryRunBackgroundTask(Runnable runnable, Runnable finallyRunnable) {
+        try {
+            if (EXECUTOR.getActiveCount() > MAX_CONCURRENT_REFRESHES) {
+                log.warn(() -> "Skipping a background refresh task because there are too many other tasks running.");
+                return;
+            }
+
+            EXECUTOR.submit(() -> {
+                try {
+                    Thread.currentThread().setName(asyncThreadName);
+                    runnable.run();
+                } catch (Throwable t) {
+                    log.warn(() -> "Exception occurred in AWS SDK background task.", t);
+                } finally {
+                    finallyRunnable.run();
+                }
+            });
+        } catch (Throwable t) {
+            log.warn(() -> "Exception occurred when submitting AWS SDK background task.", t);
+        } finally {
+            finallyRunnable.run();
+        }
     }
 }
--- a/utils/src/main/java/software/amazon/awssdk/utils/cache/NonBlocking.java
+++ b/utils/src/main/java/software/amazon/awssdk/utils/cache/NonBlocking.java
@@ -15,13 +15,8 @@
 
 package software.amazon.awssdk.utils.cache;
 
-import static java.time.temporal.ChronoUnit.HOURS;
-import static java.time.temporal.ChronoUnit.MINUTES;
-
 import java.time.Duration;
 import java.time.Instant;
-import java.util.Optional;
-import java.util.Random;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.ScheduledThreadPoolExecutor;
 import java.util.concurrent.Semaphore;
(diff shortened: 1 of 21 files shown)


utils/src/main/java/software/amazon/awssdk/utils/cache/NonBlocking.java
```diff
@@ -37,40 +41,107 @@
...
+    /**
+     * The {@link Random} instance used for calculating jitter. See {@link #DEFAULT_REFRESH_FREQUENCY_JITTER}.
+     */
+    private static final Random JITTER_RANDOM = new Random();
+
+    /**
+     * Threads used to periodically kick off credential refreshes based on the {@link #asyncRefreshFrequency}.
+     */
+    private static final ScheduledThreadPoolExecutor SCHEDULER =
+        new ScheduledThreadPoolExecutor(3, new ThreadFactoryBuilder().threadNamePrefix("sdk-cached-supplier-scheduler")
```
joviegas, line 72: Query : Why 3 ?

utils/src/test/java/software/amazon/awssdk/utils/cache/NonBlockingTest.java
```diff
@@ -0,0 +1,82 @@
...
+                nb.initializeCachedSupplier(slowSupplier);
+                nbs.add(nb);
+            }
+
+            Thread.sleep(1_000);
+
+            // In a perfect world this would be capped to 100, but the mechanism we use to limit concurrent refreshes usually
+            // means more than 100 get created. 150 should be a reasonable limit to check for, because without the limiter it
+            // would be ~1000.
+            assertThat(NonBlocking.executor().getLargestPoolSize()).isLessThanOrEqualTo(150);
```
joviegas, line 77: is 150 the highest possible value  ? or can we have 151 because current mechanisms to limit concurrent refreshes

utils/src/main/java/software/amazon/awssdk/utils/cache/NonBlocking.java
```diff
@@ -37,40 +41,107 @@
...
-    private final ScheduledExecutorService executor;
+    private volatile boolean shutdown = false;
 
     /**
      * Create a non-blocking prefetch strategy that uses the provided value for the name of the background thread that will be
      * performing the update.
      */
     public NonBlocking(String asyncThreadName) {
-        this(asyncThreadName, Duration.ofMinutes(1));
+        this(asyncThreadName, defaultRefreshFrequency());
```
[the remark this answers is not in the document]
millems: Where did I make it specific to credential updates?

utils/src/main/java/software/amazon/awssdk/utils/cache/NonBlocking.java
```diff
@@ -79,10 +150,35 @@ public void initializeCachedSupplier(CachedSupplier<?> cachedSupplier) {
...
-        executor.schedule(() -> {
+        SCHEDULER.schedule(() -> {
+            Thread.currentThread().setName(asyncThreadName);
+
+            if (shutdown) {
+                return;
+            }
+
+            if (!BACKGROUND_REFRESH_LEASE.tryAcquire()) {
+                log.warn(() -> "Skipped a background refresh to limit SDK resource consumption. Are you closing your SDK "
```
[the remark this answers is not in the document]
millems: Because this is a background thread, there's nowhere to throw an exception to. 

If the concern is too much logging, this should really never occur unless something is seriously wrong with what the customer is doing. At that point, the worst-case scenario is that it'll log once per minute, per resource (probably credential provider). We could build something to limit the logging further, but in this "something is seriously wrong" scenario we will already be logging a lot of other stuff in other parts of the SDK, so this is just a drop in the bucket.

utils/src/main/java/software/amazon/awssdk/utils/cache/NonBlocking.java
```diff
@@ -37,40 +41,107 @@
...
+
+    /**
+     * The {@link Random} instance used for calculating jitter. See {@link #DEFAULT_REFRESH_FREQUENCY_JITTER}.
+     */
+    private static final Random JITTER_RANDOM = new Random();
+
+    /**
+     * Threads used to periodically kick off credential refreshes based on the {@link #asyncRefreshFrequency}.
+     */
+    private static final ScheduledThreadPoolExecutor SCHEDULER =
```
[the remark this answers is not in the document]
millems: Unless I'm wrong, thread pools themselves are lazy (they don't create threads - even core ones - until they're needed), so making this lazy ourselves doesn't provide a lot of benefit. Did I miss something?

utils/src/test/java/software/amazon/awssdk/utils/cache/NonBlockingTest.java
millems: 90% of my experiments were < 100, and 100% were < 105. 

The reality is that it could be as many as the number of non-blocking instances we have (in this situation, 1000), but it's exceedingly unlikely under a fair scheduler. It would require EVERY thread to release the rate limiting lock, and then get parked while the task scheduler woke up (after ~60 seconds), acquired the rate limiting lock that got released, and then create a new thread to execute the task.

In reality, the scheduler would never park a thread for 60 seconds. The only reason we see it happening occasionally here is because we decreased the time between tasks to 0ms to be able to test more quickly.

utils/src/main/java/software/amazon/awssdk/utils/cache/NonBlocking.java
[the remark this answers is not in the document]
millems: Ctrl+f found a few results for credential. Let me fix that.

utils/src/main/java/software/amazon/awssdk/utils/cache/NonBlocking.java
joviegas: Got it , thanks 👍🏽

utils/src/main/java/software/amazon/awssdk/utils/cache/NonBlocking.java
```diff
@@ -37,40 +41,107 @@
...
     public NonBlocking(String asyncThreadName) {
-        this(asyncThreadName, Duration.ofMinutes(1));
+        this(asyncThreadName, defaultRefreshFrequency());
+    }
+
+    private static Duration defaultRefreshFrequency() {
+        // We jitter the default refresh frequency with each instance, so that objects created at the same time will not all be
+        // refreshing at the exact same times.
+        int jitter = Math.toIntExact(DEFAULT_REFRESH_FREQUENCY_JITTER.toMillis());
+        long asyncRefreshFrequency = DEFAULT_REFRESH_FREQUENCY.toMillis() +
```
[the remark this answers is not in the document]
cloudshiftchris: With SDK 2.17.214, as expected, cred providers do their async refresh all at the same time (when created at the same time) - and in separate thread pools.

With this PR, cred providers do their async refresh over an approx 1m window (when created at the same time) - and in the newly introduced shared thread pool.  However, the time distributions aren't uniform (some bucketing due to the below).

The time distributions are hard to get one's head around due to:
  -jittered time for async wakeup tasks;
  -prefetch time for STS cred provider;

Perhaps the prefetch strategy could be simplified by utilizing the actual cred expiration and the configured prefetch time (re-interpreted as a prefetch window).  This would avoid the "poll every 60-70s and then check prefetch" (which starts to align tasks), allowing for a clean distribution of tasks over a prefetch window.

utils/src/main/java/software/amazon/awssdk/utils/cache/NonBlocking.java
```diff
@@ -37,40 +41,107 @@
...
+                                                                     .daemonThreads(true)
+                                                                     .build());
+
+    /**
+     * Threads used to do the actual work of refreshing the credentials (because the cached supplier might block, so we don't
+     * want the work to be done by a small thread pool). This executor is created as unbounded, but in reality it is limited by
+     * the {@link #MAX_CONCURRENT_REFRESHES} via {@link #BACKGROUND_REFRESH_LEASE}.
+     */
+    private static final ThreadPoolExecutor EXECUTOR =
+        new ThreadPoolExecutor(0,
```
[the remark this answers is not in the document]
millems: I'd prefer for credential providers to not compete for resources, even if it means a few extra threads get created (up to the ~100 limit I've added). Most of the time, this shouldn't increase much beyond 1-2 once the requests have spread out.

utils/src/main/java/software/amazon/awssdk/utils/cache/NonBlocking.java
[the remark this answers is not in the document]
millems: Yeah, it's a bit confusing to wrap ones head around since most of the time it's just a timer that hits the cache or a timer that kicks off a prefetch.

It would be nice to be able to take the prefetch window into consideration, but that would be an API change I was hoping to avoid. I suppose it's not likely to be problematic to make that change so that the prefetch strategies can be aware of the prefetch times.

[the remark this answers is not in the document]
millems: @cloudshiftchris I've taken a shot at it. After we have a successful fetch, we schedule a hit against the cached supplier for a time after the prefetch time. We randomly jitter that time a bit between the prefetch time and the stale time.

utils/src/main/java/software/amazon/awssdk/utils/cache/NonBlocking.java
```diff
@@ -37,80 +46,204 @@
...
+        this(asyncThreadName, Duration.ofSeconds(60));
     }
 
     @SdkTestInternalApi
-    NonBlocking(String asyncThreadName, Duration asyncRefreshFrequency) {
-        this.executor = newExecutor(asyncThreadName);
-        this.asyncRefreshFrequency = asyncRefreshFrequency;
+    NonBlocking(String asyncThreadName, Duration minimumRefreshFrequency) {
+        this.asyncThreadName = asyncThreadName + "-" + THREAD_NUMBER.getAndIncrement();
+        this.minimumRefreshFrequency = minimumRefreshFrequency;
```
cloudshiftchris, line 129: Nice, like that this previous 'polling' duration is now the minimum refresh frequency.

utils/src/main/java/software/amazon/awssdk/utils/cache/NonBlocking.java
```diff
@@ -37,80 +46,204 @@
...
+                } catch (Throwable t) {
+                    log.warn(() -> "Exception occurred in AWS SDK background task.", t);
+                } finally {
+                    finallyRunnable.run();
+                }
+            });
+        } catch (Throwable t) {
+            log.warn(() -> "Exception occurred when submitting AWS SDK background task.", t);
+        } finally {
+            finallyRunnable.run();
```
cloudshiftchris, line 246: finallyRunnable.run() will be called twice - once when registering the task, a second one when executing, forcing that to be an idempotent operation (or introducing subtle bugs).

Perhaps the outer finally can be merged into the catch - either executing it on exception when registering, or whenever executed.

utils/src/main/java/software/amazon/awssdk/utils/cache/NonBlocking.java
```diff
@@ -37,80 +46,204 @@
...
+    private static final ThreadPoolExecutor EXECUTOR =
+        new ThreadPoolExecutor(0, Integer.MAX_VALUE,
+                               60L, TimeUnit.SECONDS,
+                               new SynchronousQueue<>(),
+                               new ThreadFactoryBuilder().daemonThreads(true).build());
+
+    /**
+     * An incrementing number, used to uniquely identify an instance of NonBlocking in the {@link #asyncThreadName}.
+     */
+    private static final AtomicLong THREAD_NUMBER = new AtomicLong(0);
```
cloudshiftchris, line 80: Perhaps better named "INSTANCE_NUMBER" as it tracks which instance of NonBlocking this is, separate from which thread it may be running on (which may vary over time)

utils/src/main/java/software/amazon/awssdk/utils/cache/NonBlocking.java
cloudshiftchris: This may already be a bug for this caller:

``` 
       if (currentlyPrefetching.compareAndSet(false, true)) {
            tryRunBackgroundTask(valueUpdater, () -> currentlyPrefetching.set(false));
        }
```

utils/src/main/java/software/amazon/awssdk/utils/cache/NonBlocking.java
```diff
@@ -37,80 +46,204 @@
...
 
+    /**
+     * The maximum number of concurrent refreshes before we start logging warnings and skipping refreshes.
+     */
+    private static final int MAX_CONCURRENT_REFRESHES = 100;
+
+    /**
+     * The {@link Random} instance used for calculating jitter of the background prefetches.
+     */
+    private static final Random JITTER_RANDOM = new Random();
```
cloudshiftchris, line 57: Perhaps consider ThreadLocalRandom for better performance

utils/src/main/java/software/amazon/awssdk/utils/cache/NonBlocking.java
millems: Yep, it is. I added that bug when I factored out the method. Good catch!

millems: Yep, that's what I meant to do originally. Good eye.

utils/src/main/java/software/amazon/awssdk/utils/cache/NonBlocking.java
```diff
@@ -37,80 +46,204 @@
...
+        getRefreshTime(result).ifPresent(this::schedulePrefetch);
+        return result;
+    }
+
+    private Optional<Instant> getRefreshTime(RefreshResult<?> result) {
+        Instant minStart = Instant.now().plus(minimumRefreshFrequency);
+        Instant rangeStart = result.prefetchTime().isBefore(minStart) ? minStart : result.prefetchTime();
+
+        if (Duration.between(Instant.now(), rangeStart).toDays() > 7) {
+            log.debug(() -> "Skipping background refresh because the prefetch time is too far in the future: " + rangeStart);
```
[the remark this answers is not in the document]
millems: We do that already at a higher layer today. Currently, IAM role session limits are ~12 hours, and we currently only use this for credentials. This is a "just in-case" branch that should theoretically not get hit by an SDK consumer today. We could probably do something more elegant, but these periodic refreshes are only a backup for when the credential provider is inactive (active credential providers will kick off their own background refresh when they are queried during after the prefetch time), and even if they're inactive the worst-case scenario is a blocking call.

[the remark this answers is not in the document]
cloudshiftchris: Got it.  Thanks for clarifying.

utils/src/main/java/software/amazon/awssdk/utils/cache/NonBlocking.java
```diff
@@ -37,80 +46,204 @@
...
+    }
+
+    public void tryRunBackgroundTask(Runnable runnable) {
+        tryRunBackgroundTask(runnable, () -> {
+        });
+    }
+
+    public void tryRunBackgroundTask(Runnable runnable, Runnable finallyRunnable) {
+        try {
+            if (EXECUTOR.getActiveCount() > MAX_CONCURRENT_REFRESHES) {
```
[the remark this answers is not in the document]
cloudshiftchris: I understand better now, thanks for clarifying.  My nervousness was on skipping -> sync - had missed the triggered prefetch (for my own sanity: we have pre-emptive/scheduled prefetch, triggered prefetch, and then sync fetch).

utils/src/main/java/software/amazon/awssdk/utils/cache/NonBlocking.java
millems: We should keep an eye out on our benchmarks and fix this if it appears, but it is used infrequently enough (once per credential refresh) that I'd be surprised if any parallel calls are contending over its lock.

utils/src/main/java/software/amazon/awssdk/utils/cache/NonBlocking.java
```diff
@@ -37,80 +46,204 @@
...
-                currentlyRefreshing.set(false);
-                throw e;
-            }
+    public <T> RefreshResult<T> fetch(Supplier<RefreshResult<T>> supplier) {
+        RefreshResult<T> result = supplier.get();
+        if (result.staleTime() == null || result.prefetchTime() == null) {
+            return result;
+        }
+
+        getRefreshTime(result).ifPresent(this::schedulePrefetch);
```
cloudshiftchris, line 157: Pondering the implications of having two prefetch times - this one (jittered, used for the pre-emptive/scheduled prefetching) and the one in RefreshResult (which is the lower bound of the jittered time).

The one in RefreshResult is used by CacheSupplier on each access - for an active cred provider, that would (at prefetch time) trigger an async prefetch, thereby cancelling (and rescheduling) a subsequent prefetch task.  For those active cred providers, the scheduled prefetches are rarely going to trigger.  Leaving the scheduled pre-fetches for little-used cred-providers.

Think this is OK, no obvious implications.

cloudshiftchris: Hmmm.  This does mean that the triggered-prefetches aren't jittered.  Perhaps we narrow this down to a single, jittered prefetch value in RefreshResult so that all consumers of that are prefetching at the same time (and perhaps add a few seconds for the scheduled one, as it may get handled by the triggered one).

utils/src/main/java/software/amazon/awssdk/utils/cache/NonBlocking.java
```diff
@@ -37,80 +46,204 @@
...
+     * The {@link Random} instance used for calculating jitter of the background prefetches.
+     */
+    private static final Random JITTER_RANDOM = new Random();
+
+    /**
+     * Thread used to kick off refreshes during the prefetch window. This does not do the actual refreshing. That's left for
+     * the {@link #EXECUTOR}.
+     */
+    private static final ScheduledThreadPoolExecutor SCHEDULER =
+        new ScheduledThreadPoolExecutor(1, new ThreadFactoryBuilder().daemonThreads(true).build());
```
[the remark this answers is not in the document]
millems: I think I'll just append the instance-level thread name to the base thread name, so they're both retained and searchable.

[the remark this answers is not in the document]
cloudshiftchris: Good stuff.  Seem to recall doing that a ways back for sanity in thread dumps.

Not critical for this use case - having a timestamp (Instant.now() in UTC) in the thread name also helps find tasks that have been sitting around for long periods. e.g. a task that was started 1.8h before the thread dump is likely stuck, causing all manner of downstream grief...

utils/src/main/java/software/amazon/awssdk/utils/cache/NonBlocking.java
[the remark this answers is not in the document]
millems: Switched to a semaphore to reduce the nervousness in future readers.

utils/src/main/java/software/amazon/awssdk/utils/cache/NonBlocking.java
millems: Fixed.

utils/src/main/java/software/amazon/awssdk/utils/cache/NonBlocking.java
millems: Fixed.

utils/src/main/java/software/amazon/awssdk/utils/cache/NonBlocking.java
millems: We're in this deep already, we might as well keep going!

cloudshiftchris: lol.  Seems like we're So. Very.  Close. !!

cloudshiftchris:
`StsCredentialsProvider` erroneously states that the sessionCache is async (it is strategized based on PrefetchStrategy)
```
    /**
     * The session cache that will update the credentials asynchronously in the background when they get close to expiring.
     */
    private final CachedSupplier<SessionCredentialsHolder> sessionCache;
```
Perhaps `The session cache that will update the credentials when they get close to expiring.`

utils/src/main/java/software/amazon/awssdk/utils/cache/NonBlocking.java
```diff
@@ -61,23 +67,27 @@ public class NonBlocking implements CachedSupplier.PrefetchStrategy {
      * the {@link #EXECUTOR}.
      */
     private static final ScheduledThreadPoolExecutor SCHEDULER =
-        new ScheduledThreadPoolExecutor(1, new ThreadFactoryBuilder().daemonThreads(true).build());
+        new ScheduledThreadPoolExecutor(1, new ThreadFactoryBuilder().threadNamePrefix("sdk-cache-scheduler")
```
cloudshiftchris, line 70: Perhaps a more specific name: `aws-sdk-credentials-provider-cache-scheduler`, specifying the AWS SDK and that this is a cache for credentials provider (to differentiate from other potential caches in the SDK, now or in the future).

utils/src/main/java/software/amazon/awssdk/utils/cache/NonBlocking.java
```diff
@@ -61,23 +67,27 @@ public class NonBlocking implements CachedSupplier.PrefetchStrategy {
...
      * want the work to be done by a small thread pool). This executor is created as unbounded, but we start complaining and
      * skipping refreshes when there are more than {@link #MAX_CONCURRENT_REFRESHES} running.
      */
     private static final ThreadPoolExecutor EXECUTOR =
-        new ThreadPoolExecutor(0, Integer.MAX_VALUE,
+        new ThreadPoolExecutor(1, Integer.MAX_VALUE,
                                60L, TimeUnit.SECONDS,
                                new SynchronousQueue<>(),
-                               new ThreadFactoryBuilder().daemonThreads(true).build());
+                               new ThreadFactoryBuilder().threadNamePrefix("sdk-cache")
```
cloudshiftchris, line 83: Perhaps a name of `aws-sdk-credentials-provider-cache-prefetch`

utils/src/main/java/software/amazon/awssdk/utils/cache/NonBlocking.java
millems: Okay, latest revision moves the jitter to the cached supplier and changed the prefetch strategy to schedule the async refresh for 1 second after the prefetch time.

utils/src/main/java/software/amazon/awssdk/utils/cache/NonBlocking.java
millems: The latest revision prepends this to the thread name specified when the prefetch strategy is created, so the complete thread name in logs would be something like sdk-cache-scheduler-0-sts-assume-role-credentials-provider-0. , and idle threads would be sdk-cache-scheduler-0.

Does that make this better?

utils/src/main/java/software/amazon/awssdk/utils/cache/NonBlocking.java
millems: The latest revision prepends this to the thread name specified when the prefetch strategy is created, so the complete thread name in logs would be something like sdk-cache-0-sts-assume-role-credentials-provider-0. , and idle threads would be sdk-cache-0.

Does that make this better?

millems:
Yeah, it's a bit more constrained now. The goal for capping it to 5 minutes was to make log dives easier, because you know that the refresh will happen within 5 minutes of the calculated prefetch time (which is usually less than 5 minutes before the expiration time). If it was unbounded, people are more likely to have to search more logs to find the info.

I just picked 5 minutes, because most of our default prefetch times are within 5 minutes of the stale times and most service throttling limits are windowed at one second. Some are longer, though, like IMDS's up-to-30 minutes.

What would be an ideal max time in your mind?

cloudshiftchris:
I think it's a cleaner story to say that prefetch happens in the window from the prefetch time to the stale time (with a bit of padding).  Providing a larger prefetch window simply jitters the prefetch over that larger duration.

Where E = token expiration, effective prefetch = rand([E - prefetch, E - stale - 1m]) (with the validity checks for overlaps etc)

Practically speaking, most SDK consumers won't change the prefetch (or have a need to change the prefetch), leaving them with reasonable defaults (and hence a constrained window for log analysis, etc).  

In the case of larger numbers of multi-tenant cred providers one may wish to spread that over a larger window (say 10m).

I don't practically see a use-case for bloated/unbounded pre-fetch windows - after say 10m you're starting to get into "fetching too frequently" territory.  Sure, folks can configure a prefetch of 3500s on a 3600s token expiration if they wish - if we want to prevent that perhaps the prefetch could be validated (at configuration time) to be max(5m or 10% of token expiration duration).  Alternately, the JavaDoc could be updated with that guidance, something along the lines of "There are diminishing returns for prefetch windows > 10% of token duration".

millems:
You've convinced me, I think. Let me remove the jitter maximum.

cloudshiftchris:
Thanks; reviewed everything earlier - no further comments, this addresses the goals.  Great work on this.

joviegas (approved):

PR #3290, opened 2022-07-08
 Endpoint resolution functionality added with parameterized tests

srsaikumarreddy:
Endpoint resolution functionality added with parameterized tests

## Motivation and Context

 Endpoint resolution functionality for endpoint and endpoint mode.

## Modifications

 Endpoint resolution functionality added with parameterized tests

## Testing

 Wrote parameterized tests for all test cases mentioned in the SEP and ran them in IDE

## Types of changes

## Checklist

- [X] I have read the [CONTRIBUTING](https://github.com/aws/aws-sdk-java-v2/blob/master/CONTRIBUTING.md) document

## License

### DIFF
--- a/core/imds/src/main/java/software/amazon/awssdk/imds/internal/DefaultEc2Metadata.java
+++ b/core/imds/src/main/java/software/amazon/awssdk/imds/internal/DefaultEc2Metadata.java
@@ -52,6 +52,7 @@ public final class DefaultEc2Metadata implements Ec2Metadata {
 
     private static final RequestMarshaller REQUEST_MARSHALLER = new RequestMarshaller();
 
+    private static final EndpointProvider ENDPOINT_PROVIDER = EndpointProvider.builder().build();
     private final RetryPolicy retryPolicy;
 
     private final URI endpoint;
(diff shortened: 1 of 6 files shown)


core/imds/src/main/java/software/amazon/awssdk/imds/internal/DefaultEc2Metadata.java
```diff
@@ -52,6 +52,7 @@ public final class DefaultEc2Metadata implements Ec2Metadata {
 
     private static final RequestMarshaller REQUEST_MARSHALLER = new RequestMarshaller();
 
+    private static final EndpointProvider ENDPOINT_PROVIDER = EndpointProvider.builder().build();
```
millems, line 55: Maybe just a TODO, but we should reuse the profile name and file from the client, when we're using this IMDS client as part of the IMDS credentials provider.

core/imds/src/main/java/software/amazon/awssdk/imds/internal/EndpointProvider.java
```diff
@@ -0,0 +1,168 @@
...
+        } else {
+            throw SdkClientException.create("Unknown endpoint mode: " + endpointMode);
+        }
+    }
+
+    public EndpointMode getEndpointMode() {
+
+        Optional<String> endpointMode = SdkSystemSetting.AWS_EC2_METADATA_SERVICE_ENDPOINT_MODE.getNonDefaultStringValue();
+        if (endpointMode.isPresent()) {
+            return EndpointMode.fromValue(endpointMode.get());
```
millems, line 69: Are these values case-sensitive in the SEP? It feels weird to accept iPv4, for example.

core/imds/src/main/java/software/amazon/awssdk/imds/internal/EndpointProvider.java
```diff
@@ -0,0 +1,168 @@
...
+
+        return profileFileToUse.profile(profileNameToUse);
+    }
+
+    private ProfileFile resolveProfileFile() {
+        if (profileFile != null) {
+            return profileFile.get();
+        }
+
+        return ProfileFile.defaultProfileFile();
```
millems, line 138: Should we store this value in a `Lazy<ProfileFile>` so that we don't re-read the file each time? Or is this cached somewhere else?

core/imds/src/main/java/software/amazon/awssdk/imds/internal/EndpointProvider.java
srsaikumarreddy: I have provided an ENUM for the same, when user enters endpoint Mode in the builder. That uses IPV4 or IPV6 as standards. But the SdkSystemSettings and ConfigFile is where the existing functionality , converts to lower case always to check

core/imds/src/main/java/software/amazon/awssdk/imds/internal/EndpointProvider.java
srsaikumarreddy: The SEP did not have inputs regarding re-read nor caching of config file.

PR #3293, opened 2022-07-11
Saikred/feature/output handling

srsaikumarreddy:
Created Metadata Response class to handle the response and provide convenient method.

## Motivation and Context

This change is required for response handling and parsing outputs.

## Modifications

Created Metadata Response class to handle the response and provide convenient method. Wrote Test cases for it.

## Testing

Wrote Test cases for the changes and tested in IDE.

## Types of changes

- [X ] New feature (non-breaking change which adds functionality)

## Checklist

- [X ] I have read the [CONTRIBUTING](https://github.com/aws/aws-sdk-java-v2/blob/master/CONTRIBUTING.md) document
- [ X] My code follows the code style of this project
- [X ] My change requires a change to the Javadoc documentation
- [ X] I have updated the Javadoc documentation accordingly
- [ X] I have added tests to cover my changes
- [ X] All new and existing tests passed

## License

- [ X] I confirm that this pull request can be released under the Apache 2 license

### DIFF
--- a/core/imds/pom.xml
+++ b/core/imds/pom.xml
@@ -93,6 +93,12 @@
             <version>${awsjavasdk.version}</version>
             <scope>compile</scope>
         </dependency>
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>json-utils</artifactId>
+            <version>${awsjavasdk.version}</version>
+            <scope>compile</scope>
+        </dependency>
     </dependencies>
 
     <build>
--- a/core/imds/pom.xml
+++ b/core/imds/pom.xml
@@ -99,6 +99,12 @@
             <version>${awsjavasdk.version}</version>
             <scope>compile</scope>
         </dependency>
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>aws-json-protocol</artifactId>
+            <version>${awsjavasdk.version}</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <build>
--- a/core/imds/pom.xml
+++ b/core/imds/pom.xml
@@ -103,7 +103,7 @@
             <groupId>software.amazon.awssdk</groupId>
             <artifactId>aws-json-protocol</artifactId>
             <version>${awsjavasdk.version}</version>
-            <scope>test</scope>
+            <scope>compile</scope>
         </dependency>
         <dependency>
             <groupId>software.amazon.awssdk</groupId>
--- a/core/imds/src/main/java/software/amazon/awssdk/imds/MetadataResponse.java
+++ b/core/imds/src/main/java/software/amazon/awssdk/imds/MetadataResponse.java
@@ -15,32 +15,33 @@
 
 package software.amazon.awssdk.imds;
 
+import java.io.IOException;
 import java.util.Arrays;
-import java.util.Collections;
 import java.util.List;
-import java.util.Map;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import software.amazon.awssdk.annotations.SdkPublicApi;
+import software.amazon.awssdk.core.document.Document;
+import software.amazon.awssdk.protocols.json.internal.unmarshall.document.DocumentUnmarshaller;
 import software.amazon.awssdk.protocols.jsoncore.JsonNode;
 import software.amazon.awssdk.protocols.jsoncore.JsonNodeParser;
+import software.amazon.awssdk.utils.Validate;
 
 /**
  * The class is used for Response Handling and Parsing the metadata fetched by the get call in the {@link Ec2Metadata} interface.
- * Metadata is stored in the instance variable <b>body</b>. The class provides convenience methods to the users to parse the
- * metadata as a String, List and to parse the metadata in the JsonResponse according to the key.
+ * The class provides convenience methods to the users to parse the metadata as a String, List and Document.
  */
 @SdkPublicApi
 public class MetadataResponse {
 
     private static final Logger log = LoggerFactory.getLogger(MetadataResponse.class);
 
-    private static final JsonNodeParser JSON_NODE_PARSER = JsonNode.parser();
+    private static final JsonNodeParser JSON_NODE_PARSER = JsonNode.parserBuilder().removeErrorLocations(true).build();
     
     private final String body;
 
     public MetadataResponse(String body) {
-        this.body = body;
+        this.body = Validate.notNull(body, "Metadata is null");
     }
 
     /**
@@ -54,7 +55,7 @@ public class MetadataResponse {
      * <pre>
      * {@code
      *
-     * Ec2Metadata ec2Metadata = Ec2Metadata.builder().build();
+     * Ec2Metadata ec2Metadata = Ec2Metadata.create();
      * MetadataResponse metadataResponse = client.get("/latest/meta-data/ami-id");
      * String response = metadataResponse.asString();
      *  }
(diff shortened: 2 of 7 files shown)


core/imds/src/main/java/software/amazon/awssdk/imds/MetadataResponse.java
```diff
@@ -0,0 +1,187 @@
...
+import org.slf4j.LoggerFactory;
+import software.amazon.awssdk.annotations.SdkPublicApi;
+import software.amazon.awssdk.protocols.jsoncore.JsonNode;
+import software.amazon.awssdk.protocols.jsoncore.JsonNodeParser;
+
+/**
+ * The class is used for Response Handling and Parsing the metadata fetched by the get call in the {@link Ec2Metadata} interface.
+ * Metadata is stored in the instance variable <b>body</b>. The class provides convenience methods to the users to parse the
+ * metadata as a String, List and to parse the metadata in the JsonResponse according to the key.
+ */
```
millems, line 32: We can exclude information that is internal to the class that might change, like that data is stored in the body field. That information isn't particularly relevant to a user of the class and it might change without someone updating the docs. It could be included as a comment in the field, if it's useful.

core/imds/src/main/java/software/amazon/awssdk/imds/MetadataResponse.java
```diff
@@ -0,0 +1,187 @@
...
+     * </pre>
+     */
+    public List<String> asList() {
+
+        if (null != body && body.contains("\n"))  {
+
+            return Arrays.asList(body.split("\n"));
+        }
+
+        return Collections.emptyList();
```
millems, line 92: If it's a single line, should we return that single line in a collection instead of not returning anything?

core/imds/src/main/java/software/amazon/awssdk/imds/MetadataResponse.java
```diff
@@ -0,0 +1,187 @@
...
+     * Ec2Metadata ec2Metadata = Ec2Metadata.builder().build();
+     * MetadataResponse metadataResponse = client.get("/latest/dynamic/instance-identity/document");
+     * String[] devpayProductCodes = metadataResponse.getStringArrayValuesFromJson("devpayProductCodes");
+     * }
+     * </pre>
+     */
+
+    public String[] getStringArrayValuesFromJson(String key) {
+
+        if (null != key) {
```
millems, line 117: should we throw on a null key?

core/imds/src/main/java/software/amazon/awssdk/imds/MetadataResponse.java
```diff
@@ -0,0 +1,187 @@
...
+     *
+     * Ec2Metadata ec2Metadata = Ec2Metadata.builder().build();
+     * MetadataResponse metadataResponse = client.get("/latest/meta-data/ancestor-ami-ids");
+     * List<String>response = metadataResponse.asList();
+     * }
+     * </pre>
+     */
+    public List<String> asList() {
+
+        if (null != body && body.contains("\n"))  {
```
millems, line 87: should we just validate that body is not null when an instance is created? or is there a use-case where body needs to be null?

core/imds/src/main/java/software/amazon/awssdk/imds/MetadataResponse.java
```diff
@@ -0,0 +1,187 @@
...
+     * {@code
+     *
+     * Ec2Metadata ec2Metadata = Ec2Metadata.builder().build();
+     * MetadataResponse metadataResponse = client.get("/latest/dynamic/instance-identity/document");
+     * String[] devpayProductCodes = metadataResponse.getStringArrayValuesFromJson("devpayProductCodes");
+     * }
+     * </pre>
+     */
+
+    public String[] getStringArrayValuesFromJson(String key) {
```
millems, line 115: Instead of having these json-like methods, maybe we should just have one to parse the body into a `software.amazon.awssdk.core.document.Document`? That would require making some currently-internal stuff protected (parsing JSON into a document), but it would be more flexible.

E.g.
```java
Ec2MetadataClient ec2Metadata = Ec2MetadataClient.create();
MetadataResponse metadataResponse = client.get("/latest/dynamic/instance-identity/document");
Document instanceIdentity = metadataResponse.asDocument();
```

core/imds/src/main/java/software/amazon/awssdk/imds/MetadataResponse.java
```diff
@@ -0,0 +1,187 @@
...
+ * The class is used for Response Handling and Parsing the metadata fetched by the get call in the {@link Ec2Metadata} interface.
+ * Metadata is stored in the instance variable <b>body</b>. The class provides convenience methods to the users to parse the
+ * metadata as a String, List and to parse the metadata in the JsonResponse according to the key.
+ */
+@SdkPublicApi
+public class MetadataResponse {
+
+    private static final Logger log = LoggerFactory.getLogger(MetadataResponse.class);
+
+    private static final JsonNodeParser JSON_NODE_PARSER = JsonNode.parser();
```
millems, line 38: We probably want removeErrorLocations to be true on this parser, so that if credentials are in the metadata it won't include them in the exception message (which might be logged).

core/imds/src/test/java/software/amazon/awssdk/imds/Ec2MetadataTest.java
```diff
@@ -121,9 +125,8 @@ public void get_AmiId_onMetadataResource_401Error_throws() throws IOException {
         stubFor(get(urlPathEqualTo(AMI_ID_RESOURCE)).willReturn(aResponse().withBody("{}").withStatus(401)));
 
         Ec2Metadata ec2Metadata = Ec2Metadata.builder().endpoint(URI.create("http://localhost:8080")).build();
-        String data = ec2Metadata.get("/latest/meta-data/ami-id");
-
-        assertThat(data).isNull();
+        MetadataResponse metadataResponse = ec2Metadata.get("/latest/meta-data/ami-id");
+        assertThat(metadataResponse).isNull();
```
millems, line 129: Returning null seems weird here. Any reason we wouldn't throw an exception? I'd think we'd never want get() to return null.

core/imds/src/main/java/software/amazon/awssdk/imds/MetadataResponse.java
```diff
@@ -0,0 +1,187 @@
...
+    private String[] stringArrayValue(JsonNode jsonNode) {
+
+        if (null != jsonNode && jsonNode.isArray()) {
+            try {
+                return jsonNode.asArray()
+                               .stream()
+                               .filter(JsonNode::isString)
+                               .map(JsonNode::asString)
+                               .toArray(String[]::new);
+            } catch (Exception e) {
```
millems, line 169: Any specific reason we're swallowing this exception instead of throwing it?

core/imds/src/main/java/software/amazon/awssdk/imds/MetadataResponse.java
srsaikumarreddy: Noted.

core/imds/src/main/java/software/amazon/awssdk/imds/MetadataResponse.java
srsaikumarreddy: Just learned about the Validate class. Will use it.

core/imds/src/main/java/software/amazon/awssdk/imds/MetadataResponse.java
srsaikumarreddy: Great insight. Will implement it.

core/imds/pom.xml
```diff
@@ -93,6 +93,18 @@
...
+            <artifactId>json-utils</artifactId>
+            <version>${awsjavasdk.version}</version>
+            <scope>compile</scope>
+        </dependency>
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>aws-json-protocol</artifactId>
+            <version>${awsjavasdk.version}</version>
+            <scope>compile</scope>
+        </dependency>
```
millems, line 107: We should put something in the backlog to avoid needing this dependency at GA. I know it was needed for the DocumentMarshaller, but I think it's relatively small and we could copy it to avoid the dependency.

millems (approved):

PR #3300, opened 2022-07-14
Allow pausing a resumed download even when the download hasn't already started.

millems:


services-custom/s3-transfer-manager/src/test/java/software/amazon/awssdk/transfer/s3/internal/DefaultFileDownloadTest.java
```diff
@@ -58,17 +59,34 @@ public static void tearDown() throws IOException {
...
+        ResumableFileDownload pause = fileDownload.pause();
+        assertThat(pause.downloadFileRequest()).isEqualTo(request);
+        assertThat(pause.bytesTransferred()).isEqualTo(file.length());
+        assertThat(pause.s3ObjectLastModified()).hasValue(sdkResponse.lastModified());
+        assertThat(pause.totalSizeInBytes()).hasValue(sdkResponse.contentLength());
     }
 
     @Test
-    void pause_shouldReturnCorrectly() {
+    void immediatePauseAfterResumeReturnsOriginalRequest() {
```
cenedhryn, line 89: I cannot see any difference between this test case and the previous one after refactoring

millems: You're right! Refactoring artifact. Will delete.

cenedhryn (approved):

PR #3310, opened 2022-07-18
Do not include 0 status code in exception messages, and change the default status code to 0.

millems:
This fixes an issue where event stream exception messages included a status code of zero.


core/protocols/aws-json-protocol/src/main/java/software/amazon/awssdk/protocols/json/internal/unmarshall/AwsJsonProtocolErrorUnmarshaller.java
```diff
@@ -92,14 +92,13 @@ private Duration getClockSkew(ExecutionAttributes executionAttributes) {
...
-    private int statusCode(SdkHttpFullResponse response, Optional<ExceptionMetadata> modeledExceptionMetadata) {
+    private Optional<Integer> statusCode(SdkHttpFullResponse response, Optional<ExceptionMetadata> modeledExceptionMetadata) {
         if (response.statusCode() != 0) {
-            return response.statusCode();
+            return Optional.of(response.statusCode());
         }
 
         return modeledExceptionMetadata.filter(m -> m.httpStatusCode() != null)
-                                       .map(ExceptionMetadata::httpStatusCode)
-                                       .orElse(500);
```
joviegas, line 102: - Why were we returning 500 as default value before ? 
- In what cases service sends 0 as status code.
- Should we add a change log, looks like its a bug fix.

PR #3314, opened 2022-07-20
Fix issue in IMDS credentials provider that causes expired credential…

millems:
…s to be vended for a short period of time after the credentials provider is inactive for a long time.

Before this change, if credentials are stale (because they weren't prefetched, likely due to inactivity) only one thread would block to refresh and other calling threads would be given expired credentials. This is good during an IMDS outage if the credential expiration has been extended service-side, but it's bad when the credentials are actually expired.

After this change, if credentials are stale and we go to refresh them, we'll hold other calling threads until that refresh completes. This will cause increased latency during credential refreshes during an IMDS outage, but ensure that vending expired credentials is minimized outside of outage scenarios.


utils/src/main/java/software/amazon/awssdk/utils/cache/CachedSupplier.java
```diff
@@ -156,54 +188,117 @@ private void refreshCache() {
...
+                               + "of " + newStale);
+                return fetch.toBuilder().staleTime(newStale).build(); // Refresh again in 1 second
+            case ALLOW:
+                Instant newStaleTime = jitterTime(now, Duration.ofMinutes(1), Duration.ofMinutes(10));
+                log.warn(() -> "Cached value expiration has been extended to " + newStaleTime + " because the downstream "
+                               + "service returned a time in the past: " + fetch.staleTime());
+
+                return fetch.toBuilder()
+                            .staleTime(newStaleTime)
+                            .build();
```
millems, line 239: Static stability! IMDS returned an expired value. Use that value for the next 1-10 minutes before trying to get a new value from IMDS.

utils/src/main/java/software/amazon/awssdk/utils/cache/CachedSupplier.java
```diff
@@ -156,54 +188,117 @@ private void refreshCache() {
...
+            throw e;
+        }
+
+        Instant now = clock.instant();
+        if (!now.isBefore(currentCachedValue.staleTime())) {
+            int numFailures = consecutiveStaleRetrievalFailures.incrementAndGet();
+
+            switch (staleValueBehavior) {
+                case STRICT:
+                    throw e;
```
millems, line 260: We failed to fetch from the underlying supplier and the current value is stale. Throw that exception up to the caller. This is a slight behavior change from what exists today, where we'd throw the exception even if the current value wasn't stale (i.e. it was a prefetch). This behavior felt more friendly.

utils/src/main/java/software/amazon/awssdk/utils/cache/CachedSupplier.java
```diff
@@ -156,54 +188,117 @@ private void refreshCache() {
...
+                case ALLOW:
+                    Instant newStaleTime = jitterTime(now, Duration.ofMillis(1), maxStaleFailureJitter(numFailures));
+                    log.warn(() -> "Cached value expiration has been extended to " + newStaleTime + " because calling the "
+                                   + "downstream service failed (consecutive failures: " + numFailures + ").");
+
+                    return currentCachedValue.toBuilder()
+                                             .staleTime(newStaleTime)
+                                             .build();
```
millems, line 268: Static stability! Kind of! Technically, static stability for IMDS was only supposed to account for when IMDS returned expired credentials, not when IMDS was down and we had a stale value. I took some liberties with this logic, to align with what I think would be "correct" for a cached supplier to do in this circumstance: return the stale cached value, but schedule the next 'stale' refresh to happen using exponential backoff with jitter (here, first failure: 1-100ms, second failure: 1-200ms, third failure: 1-400ms, ... up to 1-10,000 ms).

utils/src/test/java/software/amazon/awssdk/utils/cache/CachedSupplierTest.java
```diff
@@ -158,17 +195,58 @@ public void allCallsBeforeInitializationBlock() {
...
+            waitingSupplier.permits.release(50);
+            waitForAsyncGetsToFinish();
+
+            // Make extra sure all 3 "gets" actually happened.
+            waitingSupplier.waitForGetsToHaveFinished(3);
+        }
+    }
+
+    @Test
+    public void staleValueBlocksAllCallsEvenWithStaleValuesAllowed() throws InterruptedException {
```
millems, line 227: This tests the entire point of this change: when we have static stability enabled, and we call get() on the supplier with a stale value, the get()s block.

millems (commented):
Added comments to help with a review.

zoewangg (approved):

utils/src/main/java/software/amazon/awssdk/utils/cache/CachedSupplier.java
```diff
@@ -253,12 +361,32 @@ public Builder<T> prefetchStrategy(PrefetchStrategy prefetchStrategy) {
...
+        public Builder<T> staleValueBehavior(StaleValueBehavior staleValueBehavior) {
+            this.staleValueBehavior = staleValueBehavior;
+            return this;
+        }
+
+        /**
+         * Configure the clock used for this cached supplier. Configurable for testing.
+         */
+        @SdkTestInternalApi
+        public Builder<T> clock(Clock clock) {
```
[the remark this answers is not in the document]
millems: This is passed in from the InstanceProfileCredentialProvider, so it needs to be public. This is a protected class, though, so it's not something customers will see. Otherwise I wouldn't have done it this way.

PR #3331, opened 2022-07-28
Use fault trait when determining exception status code

dagnir:
AWS-JSON services use the Fault trait to designate exceptions that are the
service's fault (i.e. 500), rather than HTTP status code bindings like REST
services. In cases where the response being unmarshalled does *NOT* have a
status code (e.g. because the response was constructed rather from some other
data rather than an actual HTTP response), the JSON unmarshaller will now also
check the metadata if it's marked as a fault.

## Types of changes

## Checklist

## License


core/protocols/aws-json-protocol/src/main/java/software/amazon/awssdk/protocols/json/internal/unmarshall/AwsJsonProtocolErrorUnmarshaller.java
```diff
@@ -97,9 +97,22 @@ private int statusCode(SdkHttpFullResponse response, Optional<ExceptionMetadata>
...
 
-        return modeledExceptionMetadata.filter(m -> m.httpStatusCode() != null)
-                                       .map(ExceptionMetadata::httpStatusCode)
-                                       .orElse(500);
+        if (modeledExceptionMetadata.isPresent()) {
+            ExceptionMetadata md = modeledExceptionMetadata.get();
+
+            if (md.httpStatusCode() != null) {
+                return md.httpStatusCode();
+            }
```
millems, line 105: Should we maybe just set the status code to 400/500 based on isFault within the code generator so that we don't need to use isFault in the runtime?

dagnir: Yeah I think that makes sense. We can push the logic all the way into `ProtocolSpec`.

millems (approved):

Issue #2259, opened 2021-01-25
S3Client fails with EOFException when executing HeadObjectRequest for gzip encoded object


productivityindustries:
## Describe the bug

I am trying to execute a HeadObjectRequest for objects stored in S3. All objects are compressed and do have Content-Encoding: gzip.

## Expected Behavior

Just head info's are returned - and no exception is thrown.

## Current Behavior

This stacktrace is logged:
```
Exception in thread "main" java.io.UncheckedIOException: java.io.EOFException
	at software.amazon.awssdk.utils.FunctionalUtils.asRuntimeException(FunctionalUtils.java:180)
	at software.amazon.awssdk.utils.FunctionalUtils.lambda$safeSupplier$4(FunctionalUtils.java:110)
	at software.amazon.awssdk.utils.FunctionalUtils.invokeSafely(FunctionalUtils.java:136)
	at software.amazon.awssdk.core.http.Crc32Validation.decompressing(Crc32Validation.java:85)
	at software.amazon.awssdk.core.http.Crc32Validation.process(Crc32Validation.java:62)
	at software.amazon.awssdk.core.http.Crc32Validation.validate(Crc32Validation.java:44)
	at software.amazon.awssdk.awscore.client.handler.AwsSyncClientHandler$Crc32ValidationResponseHandler.handle(AwsSyncClientHandler.java:94)
	at software.amazon.awssdk.core.internal.handler.BaseClientHandler.lambda$successTransformationResponseHandler$6(BaseClientHandler.java:252)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.HandleResponseStage.execute(HandleResponseStage.java:40)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.HandleResponseStage.execute(HandleResponseStage.java:30)
	at software.amazon.awssdk.core.internal.http.pipeline.RequestPipelineBuilder$ComposingRequestPipelineStage.execute(RequestPipelineBuilder.java:206)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.ApiCallAttemptTimeoutTrackingStage.execute(ApiCallAttemptTimeoutTrackingStage.java:73)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.ApiCallAttemptTimeoutTrackingStage.execute(ApiCallAttemptTimeoutTrackingStage.java:42)
[...]

PR #3346, opened 2022-08-07
Fix for S3Client with httpClient as url-connection-client fails with EOFException when executing HeadObjectRequest for gzip encoded object

joviegas:
## Motivation and Context
- Fix for S3Client with httpCLient as url-connection-client fails with EOFException when executing HeadObjectRequest for gzip encoded object
- Please refer https://github.com/aws/aws-sdk-java-v2/issues/2259

## Modifications

- Similar to other http client where the responseBody is set to Null when the response doesnot have stream in its entity or in other words  if http response doesnot have payload.

- Refererence 
- Async Htttp client ==>https://github.com/aws/aws-sdk-java-v2/pull/2201
- Sync Http Client code ==> [ApacheHttpClient_sdk_code](https://github.com/aws/aws-sdk-java-v2/blob/f3265c681e68dbbadca9d8794617a067dda9959d/http-clients/apache-client/src/main/java/software/amazon/awssdk/http/apache/ApacheHttpClient.java#L286-L287)
```java
    private HttpExecuteResponse createResponse(org.apache.http.HttpResponse apacheHttpResponse,
                                               HttpRequestBase apacheRequest) throws IOException {

        // ....existing code...
        AbortableInputStream responseBody = apacheHttpResponse.getEntity() != null ?
                                   toAbortableInputStream(apacheHttpResponse, apacheRequest) : null;

        return HttpExecuteResponse.builder().response(responseBuilder.build()).responseBody(responseBody).build();

    }

```

## Testing

- Added new Junit
-  Perform all the integ test of S3 by changing the Httpclient as UrlHttpClient and all the test passes.

## Types of changes

## License


http-clients/url-connection-client/src/main/java/software/amazon/awssdk/http/urlconnection/UrlConnectionHttpClient.java
```diff
@@ -302,7 +302,10 @@ public HttpExecuteResponse call() throws IOException {
             int responseCode = getResponseCodeSafely(connection);
             boolean isErrorResponse = HttpStatusFamily.of(responseCode).isOneOf(CLIENT_ERROR, SERVER_ERROR);
             Optional<InputStream> responseContent = isErrorResponse ? tryGetErrorStream() : tryGetInputStream();
-            AbortableInputStream responseBody = responseContent.map(AbortableInputStream::create).orElse(null);
+
+            AbortableInputStream responseBody = responseHasNoContent()
```
millems, line 306: Should we do this in tryGetErrorStream and tryGetInputStream (or their downstream methods)? It seems weird that we have responseContent being present even when there's technically no content.

joviegas: - Agree, thats because of https://github.com/JetBrains/jdk8u_jdk/blob/master/src/share/classes/sun/net/www/protocol/http/HttpURLConnection.java#L1858
- Thus it make sense in adding in tryGetInputStream
- However in case of `tryGetErrorStream` we get null Stream in case of errors for all type of method like 'GET' , 'HEAD' etc , so I think it not required for `tryGetErrorStream`

http-clients/url-connection-client/src/main/java/software/amazon/awssdk/http/urlconnection/UrlConnectionHttpClient.java
```diff
@@ -320,7 +320,9 @@ private Optional<OutputStream> tryGetOutputStream() {
         }
 
         private Optional<InputStream> tryGetInputStream() {
-            return getAndHandle100Bug(() -> invokeSafely(connection::getInputStream), true);
+            return responseHasNoContent()
+                   ? Optional.empty()
+                   : getAndHandle100Bug(() -> invokeSafely(connection::getInputStream), true);
```
millems, line 325: Is it possible that we might need to check for no-content in tryGetErrorStream?

millems (approved):

http-clients/url-connection-client/src/main/java/software/amazon/awssdk/http/urlconnection/UrlConnectionHttpClient.java
joviegas: It might be possible will do some deep dive , also will need to add a test case for this . Since I am going on leave will close this PR for now and will raise a separate PR when back from leave , will keep this task  is in fast track

joviegas: As per the code walk we donot need this check for erroStreams , details in https://github.com/aws/aws-sdk-java-v2/pull/3363

PR #3357, opened 2022-08-16
[TM upload pause/resume Part 2] Implement pause and resume for uploadFile

zoewangg:
This is part 2 of pause/resume for uploadFile.

### DIFF
--- a/services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/DefaultS3TransferManager.java
+++ b/services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/DefaultS3TransferManager.java
@@ -218,7 +218,7 @@ public final class DefaultS3TransferManager implements S3TransferManager {
         boolean noResumeToken = !hasResumeToken(resumableFileUpload);
 
         if (fileModified || noResumeToken) {
-            return uploadFromBeginning(resumableFileUpload, fileModified);
+            return uploadFromBeginning(resumableFileUpload, fileModified, noResumeToken);
         }
 
         return doResumeUpload(resumableFileUpload);
@@ -243,7 +243,8 @@ public final class DefaultS3TransferManager implements S3TransferManager {
                                            .build());
     }
 
-    private FileUpload uploadFromBeginning(ResumableFileUpload resumableFileUpload, boolean fileModified) {
+    private FileUpload uploadFromBeginning(ResumableFileUpload resumableFileUpload, boolean fileModified,
+                                           boolean noResumeToken) {
         UploadFileRequest uploadFileRequest = resumableFileUpload.uploadFileRequest();
         PutObjectRequest putObjectRequest = uploadFileRequest.putObjectRequest();
         if (fileModified) {
@@ -269,6 +270,16 @@ public final class DefaultS3TransferManager implements S3TransferManager {
                                                     }
                                                 });
                                });
+        }
+
+        if (noResumeToken) {
+            log.debug(() -> String.format("No resume token is found " +
+                                          "The SDK will upload the requested object in bucket"
+                                          + " (%s) with key (%s) from "
+                                          + "the "
+                                          + "beginning.",
+                                          putObjectRequest.bucket(),
+                                          putObjectRequest.key()));
         }
 
 
--- a/services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/DefaultS3TransferManager.java
+++ b/services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/DefaultS3TransferManager.java
@@ -18,7 +18,6 @@ package software.amazon.awssdk.transfer.s3.internal;
 import static software.amazon.awssdk.core.interceptor.SdkInternalExecutionAttribute.SDK_HTTP_EXECUTION_ATTRIBUTES;
 import static software.amazon.awssdk.services.s3.internal.crt.S3InternalSdkHttpExecutionAttribute.CRT_PAUSE_RESUME_TOKEN;
 import static software.amazon.awssdk.services.s3.internal.crt.S3InternalSdkHttpExecutionAttribute.METAREQUEST_PAUSE_OBSERVABLE;
-import static software.amazon.awssdk.transfer.s3.internal.serialization.CrtUploadResumeToken.marshallResumeToken;
 import static software.amazon.awssdk.transfer.s3.internal.utils.FileUtils.fileNotModified;
 import static software.amazon.awssdk.transfer.s3.internal.utils.ResumableRequestConverter.toDownloadFileRequestAndTransformer;
 
@@ -232,7 +231,7 @@ public final class DefaultS3TransferManager implements S3TransferManager {
         CrtUploadResumeToken token = new CrtUploadResumeToken(resumableFileUpload.totalNumOfParts().getAsLong(),
                                                               resumableFileUpload.partSizeInBytes().getAsLong(),
                                                               resumableFileUpload.multipartUploadId().orElse(null));
-        String marshalledToken = marshallResumeToken(token);
+        String marshalledToken = token.marshallResumeToken();
 
         Consumer<SdkHttpExecutionAttributes.Builder> attachResumeToken =
             b -> b.put(CRT_PAUSE_RESUME_TOKEN, marshalledToken);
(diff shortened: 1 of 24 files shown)


services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/serialization/CrtUploadResumeToken.java
```diff
@@ -0,0 +1,72 @@
...
+
+import java.nio.charset.StandardCharsets;
+import java.util.Map;
+import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.protocols.jsoncore.JsonNode;
+import software.amazon.awssdk.protocols.jsoncore.JsonNodeParser;
+import software.amazon.awssdk.protocols.jsoncore.JsonWriter;
+
+@SdkInternalApi
+public class CrtUploadResumeToken {
```
zoewangg, line 26: This is the POJO class for the JSON returned from crt class`S3MetaRequest#pause`

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/S3MetaRequestPauseObservable.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+package software.amazon.awssdk.services.s3.internal.crt;
+
+import java.util.function.Function;
+import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.crt.s3.S3MetaRequest;
+
+@SdkInternalApi
+public class S3MetaRequestPauseObservable {
```
zoewangg, line 23: The naming might be confusing. LMK if you can think of a better name, but this is the class created in Transfer Manager and passed to `S3CrtAsyncHttpClient` through `SdkHttpExecutionAttribute`.

I thought about using generic types instead of String, but it's a bit awkward in  `SdkHttpExecutionAttribute` because we'd need to do casting for generic types

```
    public static final S3InternalSdkHttpExecutionAttribute<S3MetaRequestPauseObservable> METAREQUEST_PAUSE_OBSERVABLE =
        new S3InternalSdkHttpExecutionAttribute<>(S3MetaRequestPauseObservable.class);
```

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/model/ResumableFileUpload.java
```diff
@@ -71,15 +71,9 @@ public boolean equals(Object o) {
 
         ResumableFileUpload that = (ResumableFileUpload) o;
 
-        if (totalNumOfParts != that.totalNumOfParts) {
```
joviegas, line 74: I see not all the fields/attributes are present in the Equals method , any reason ? Same holds for hash , do we have test case for equals and hashCode ?

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/S3CrtAsyncHttpClient.java
```diff
@@ -87,15 +89,21 @@ public CompletableFuture<Void> execute(AsyncExecuteRequest asyncRequest) {
             new S3CrtResponseHandlerAdapter(executeFuture, asyncRequest.responseHandler());
 
         S3MetaRequestOptions.MetaRequestType requestType = requestType(asyncRequest);
+        String resumeToken = asyncRequest.httpExecutionAttributes().getAttribute(CRT_PAUSE_RESUME_TOKEN);
```
joviegas, line 92: what happen if we try to do following
1. clientOne.uploadFile_One
2. clientOne.Pause_one
3. clientOne.uploadFile_Two
4. Wait for uploadFile_Two to complete
5. Then Resume  file One that is  clientOne.Pause_one

Do we need a test case for this ?

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/S3MetaRequestPauseObservable.java
joviegas: Could you please add a Java Doc for this class to know the intent/responsibility of this task , just for internal references.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/model/ResumableFileUpload.java
zoewangg: No reason, I will fix them. Not yet, I'll add tests as I'm finalizing the PR

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/S3CrtAsyncHttpClient.java
zoewangg: Do you mean uploading the file, pausing and then uploading the same file from start again? hmm, good question, let me try it. I think the parts uploaded in the first attempt will be overridden by the second attempt.

zoewangg: I ran tests with the following order:
1. uploading the file
2. pausing
3. uploading the same file from start 
4. resuming from step2

Step 1, 2 and 3 succeeded. Step 4 threw `NoSuchUploadException`, which makes sense because the object was uploaded successfully in step 3

I'm not sure if we should add a test case for this because it is more of a service behavior

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/serialization/CrtUploadResumeToken.java
```diff
@@ -0,0 +1,101 @@
...
+        jsonGenerator.writeStartObject();
+
+        TransferManagerJsonMarshaller.LONG.marshall(resumeToken.totalNumOfParts(), jsonGenerator, "total_num_parts");
+        TransferManagerJsonMarshaller.LONG.marshall(resumeToken.partSizeInBytes(), jsonGenerator, "partition_size");
+        TransferManagerJsonMarshaller.STRING.marshall("AWS_S3_META_REQUEST_TYPE_PUT_OBJECT", jsonGenerator, "type");
+        TransferManagerJsonMarshaller.STRING.marshall(resumeToken.multipartUploadId(), jsonGenerator, "multipart_upload_id");
+
+        jsonGenerator.writeEndObject();
+        return new String(jsonGenerator.getBytes(), StandardCharsets.UTF_8);
+    }
```
millems, line 90: Should this be a virtual method instead of a static one? (`String x = resumeToken.marshall()`)

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/DefaultS3CrtAsyncClient.java
```diff
@@ -214,17 +221,17 @@ private static void validateOverrideConfiguration(SdkRequest request) {
...
                 if (overrideConfiguration.apiCallAttemptTimeout().isPresent()) {
                     throw new UnsupportedOperationException("Request-level apiCallAttemptTimeout override is not supported");
                 }
 
-                if (overrideConfiguration.executionAttributes() != null) {
-                    throw new UnsupportedOperationException("Request-level executionAttributes override is not supported");
-                }
+                // if (overrideConfiguration.executionAttributes() != null) {
+                //     throw new UnsupportedOperationException("Request-level executionAttributes override is not supported");
+                // }
```
millems, line 234: Delete commented out code?

joviegas (approved):

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/DefaultS3TransferManager.java
```diff
@@ -185,7 +204,118 @@ public FileUpload uploadFile(UploadFileRequest uploadFileRequest) {
...
+                                           .putObjectRequest(modifiedPutObjectRequest)
+                                           .build());
+    }
+
+    private FileUpload uploadFromBeginning(ResumableFileUpload resumableFileUpload, boolean fileModified,
+                                           boolean noResumeToken) {
+        UploadFileRequest uploadFileRequest = resumableFileUpload.uploadFileRequest();
+        PutObjectRequest putObjectRequest = uploadFileRequest.putObjectRequest();
+        if (fileModified) {
+            log.debug(() -> String.format("The file (%s) has been modified since "
```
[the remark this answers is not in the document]
zoewangg: I thought about this, and imo INFO may be a bit too verbose (we have customers complaining about verbosity of our logs before) and WARN may scare customers and this is probably not something we should alarm on.

PR #3380, opened 2022-08-25
Fix Trailer based Http Checksum for Async Request body with variable chunk size

joviegas:
## Motivation and Context

Fix Trailer based Http Checksum for Async Request body created from File

## Modifications

- The Non last chunks in ChecksumCalculatingAsyncRequestBody.`ChecksumCalculatingSubscriber ` was not appending the Chunk size and Carriage return to the chunks.
- The FileAsynRequest Body was not handled , this sends the data in chunks of 16KB, thus the Content-length is updated to consider additional bytes due to size and Carriage return

- Added ChunkBuffer class that will Buffer the incoming bytes until it reaches `buffer_size`
- Added flatMapIterable to wrapped netty stream of data such that we map it to ChunkBuffer
   ` wrapped.flatMapIterable(this::buffer).subscribe(new ChecksumCalculatingSubscriber(s, sdkChecksum, trailerHeader, totalBytes)); `

- Integ test with File Async body Greater and less than 16K added both for sync and async.

## Types of changes

## License


core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/ChecksumCalculatingAsyncRequestBody.java
```diff
@@ -215,6 +216,14 @@ private ByteBuffer getFinalChecksumAppendedChunk(ByteBuffer byteBuffer) {
             return checksumAppendedBuffer;
         }
 
+        private ByteBuffer appendChunkSizeAndFinalByte(ByteBuffer byteBuffer) {
+            ByteBuffer contentChunk = createChunk(byteBuffer, false);
+            ByteBuffer checksumAppendedBuffer = ByteBuffer.allocate(contentChunk.remaining());
+            checksumAppendedBuffer.put(contentChunk);
+            checksumAppendedBuffer.flip();
```
dagnir, line 223: I don't understand the need for this. Why create a copy of `contentChunk`?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/interceptor/AsyncRequestBodyHttpChecksumTrailerInterceptor.java
```diff
@@ -106,4 +112,9 @@ private static SdkHttpRequest updateHeadersForTrailerChecksum(Context.ModifyHttp
                         .putHeader(Header.CONTENT_LENGTH,
                                    Long.toString(chunkLength + checksumContentLength)));
     }
+
+    private static boolean isFileAsyncRequestBody(Context.ModifyHttpRequest context) {
+        return context.asyncRequestBody().isPresent() && context.asyncRequestBody().get() instanceof FileAsyncRequestBody;
+    }
```
dagnir, line 118: Hmm not a fan of this. Can we not fix this in a more generic way? What if a customer provides their own implementation that behaves exactly like FileAsyncRequestBody?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/ChecksumCalculatingAsyncRequestBody.java
joviegas: We need to  need to separately create `contentChunk` because we need to add following
- prefix = chunk size in hex + CRLF
- Suffix = CRLF

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/interceptor/AsyncRequestBodyHttpChecksumTrailerInterceptor.java
joviegas: Agree to this , will discuss to you offline.
We need to figure out the DEFAULT_CHUNK_SIZE which is dependent on AsyncRequestBody implementation

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/ChecksumCalculatingAsyncRequestBody.java
joviegas: Removed

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/interceptor/AsyncRequestBodyHttpChecksumTrailerInterceptor.java
joviegas: Done

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/io/AwsUnsignedChunkedEncodingInputStream.java
```diff
@@ -68,17 +66,19 @@ private static long calculateChunkLength(long originalContentLength) {
...
         }
 
-        long maxSizeChunks = originalLength / DEFAULT_CHUNK_SIZE;
-        long remainingBytes = originalLength % DEFAULT_CHUNK_SIZE;
+        long maxSizeChunks = originalLength / defaultChunkSize;
+        long remainingBytes = originalLength % defaultChunkSize;
+
+        long allChunks = maxSizeChunks * calculateChunkLength(defaultChunkSize);
+        long remainingInChunk = remainingBytes > 0 ? calculateChunkLength(remainingBytes) : 0;
+        long lastByteSize = (long) "0".length() + (long) CRLF.length();
```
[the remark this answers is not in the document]
joviegas: Updated

utils/src/main/java/software/amazon/awssdk/utils/async/ByteBufferingSubscriber.java
```diff
@@ -0,0 +1,96 @@
...
+            if (bufferedBytes.get() == 0) {
+                currentBuffer.put(t.array(), startPosition, bytesToMove);
+
+            } else {
+                currentBuffer.put(t.array(), 0, bytesToMove);
+            }
+            startPosition = startPosition + bytesToMove;
+            if (bufferedBytes.addAndGet(bytesToMove) == bufferSize) {
+                currentBuffer.position(0);
+                subscriber.onNext((T) currentBuffer);
```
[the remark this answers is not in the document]
joviegas: Doen , reusing the .flatMapIterable API to create fixed chunks.

test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/AsyncRequestBodyFlexibleChecksumInTrailerTest.java
```diff
@@ -28,12 +28,16 @@
 import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
 import static software.amazon.awssdk.http.Header.CONTENT_LENGTH;
 import static software.amazon.awssdk.http.Header.CONTENT_TYPE;
+import static software.amazon.awssdk.testutils.FileUtils.fixedLengthInKbFileWithRandomOrFixedCharacters;
```
joviegas, line 31: I will raise another PR for converting this to Junit5 .

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/ChecksumCalculatingAsyncRequestBody.java
```diff
@@ -225,4 +238,48 @@ public void onComplete() {
             wrapped.onComplete();
         }
     }
+
+    private Iterable<ByteBuffer> bufferAndCreateChunks(ByteBuffer buffer) {
+        int startPosition = 0;
+        int currentBytesRead = buffer.remaining();
+
+        List<ByteBuffer> resultBufferedList = new ArrayList<>();
```
zoewangg, line 246: Should we buffer the `ByteBuffer`s from upstream? What happens if this gets called from different threads? For example, upstream onNext gets invoked sequentially by two different threads?

This is how event stream messages are buffered: https://github.com/awslabs/aws-eventstream-java/blob/master/src/main/java/software/amazon/eventstream/MessageDecoder.java#L99

services/s3/src/test/java/software/amazon/awssdk/services/s3/utils/ChecksumUtils.java
```diff
@@ -66,4 +75,24 @@ private static MessageDigest createMessageDigest() {
             throw new RuntimeException("Unable to create SHA-256 MessageDigest instance", e);
         }
     }
+
+    public static File fixedLengthInKbFileWithRandomOrFixedCharacters(int sizeInKb, boolean isRandom) throws IOException {
```
zoewangg, line 79: It seems we can use RandomTempFile https://github.com/aws/aws-sdk-java-v2/blob/master/test/test-utils/src/main/java/software/amazon/awssdk/testutils/RandomTempFile.java

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/ChecksumCalculatingAsyncRequestBody.java
```diff
@@ -41,12 +43,16 @@
 @SdkInternalApi
 public class ChecksumCalculatingAsyncRequestBody implements AsyncRequestBody {
 
+    public static final int DEFAULT_CHUNK_SIZE = 16 * 1024;
+
     public static final byte[] FINAL_BYTE = new byte[0];
```
zoewangg, line 48: private?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/ChecksumCalculatingAsyncRequestBody.java
```diff
@@ -41,6 +42,8 @@
 @SdkInternalApi
 public class ChecksumCalculatingAsyncRequestBody implements AsyncRequestBody {
 
+    public static final int DEFAULT_CHUNK_SIZE = 16 * 1024;
+
```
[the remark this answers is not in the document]
joviegas: for Sync DEFAULT_CHUNK_SIZE in AwsChunkedEncodingInputStream was 128 * 1024 from begining.

For Async since we started by taking Async File Request body as a reference , I took the default chunk size as specified for AsyncFileRequestBody.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/ChecksumCalculatingAsyncRequestBody.java
joviegas: Done, thanks for the reference code.

dagnir (approved):

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/ChunkBuffer.java
```diff
@@ -0,0 +1,142 @@
...
+import java.util.concurrent.atomic.AtomicLong;
+import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.utils.Validate;
+import software.amazon.awssdk.utils.builder.SdkBuilder;
+
+/**
+ * Class that will buffer incoming BufferBytes of totalBytes length to chunks of bufferSize*
+ */
+@SdkInternalApi
+public final class ChunkBuffer {
```
zoewangg, line 33: As discussed offline, we should probably make this thread safe

zoewangg (approved):
Can we trigger the test suites if we haven't already

PR #3402, opened 2022-09-02
Moves HTTP URLConnectionClient S3 integration tests to separate module

cenedhryn:
## Motivation and Context
The http client modules should not depend on specific AWS Services; it breaks the architecture of the SDK and may cause issues with direct and transitive dependencies that can lead to circular dependencies 

Example: the new IMDS client which is stopped from depending on the url-connection-client because regions depend on imds and the url connection client has a transitive dependency on regions.

## Modifications
Created a new integration test module under `test`. Suggestions for better names are welcome!

## Testing
Compilation and tests successful in new module.


test/service-sdk-tests/pom.xml
```diff
@@ -0,0 +1,130 @@
...
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <artifactId>aws-sdk-java-pom</artifactId>
+        <groupId>software.amazon.awssdk</groupId>
+        <version>2.17.266-SNAPSHOT</version>
+        <relativePath>../../pom.xml</relativePath>
+    </parent>
+    <artifactId>service-sdk-tests</artifactId>
+    <name>AWS Java SDK :: Test :: Service SDK tests</name>
+    <description>Service-related tests that are too specialized to be placed in a service or core module.</description>
```
joviegas, line 29: The URL Connection Test is more Http Client tests than service test `,`Service-related tests` makes me assume this is very specific to Services while in case of Url connections client we test Http Url Connection

test/service-sdk-tests/pom.xml
```diff
@@ -0,0 +1,130 @@
...
+  ~
+  ~  http://aws.amazon.com/apache2.0
+  ~
+  ~ or in the "license" file accompanying this file. This file is distributed
+  ~ on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+  ~ express or implied. See the License for the specific language governing
+  ~ permissions and limitations under the License.
+  -->
+
+<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd"
```
zoewangg, line 17: Should we move those tests to s3 instead? I'm not sure what service tests should to added this new module and not service modules

cenedhryn: In the `s3` module Maven complains that it cannot resolve http client (2 http clients).

zoewangg: Yeah, we'd have to specify the http client for every S3Client; using system property may be easier. 

Wrt the circular dependency issue, we probably can't default to url http client for IMDS client anyway because core depends on it and it will break customers once we add url http client in the classpath

test/service-sdk-tests/pom.xml
joviegas: Also currently we have a test module called `service-test-utils` which sounds similar to `service-sdk-tests`

PR #3403, opened 2022-09-02
Implement automatic multipart copy functionality in S3 CRT async client

zoewangg:
## Motivation and Context
Implement automatic multipart copy functionality in S3CrtAsyncClient instead of relying on CRT implementation 

## Modifications
Implement automatic multipart copy functionality in S3CrtAsyncClient

Below is the workflow:
1. Invoke HeadObject on the source object and check the size
2. If the size > configured partSizeInBytes, use multipart copy, otherwise use CopyObject directly
3. For multipart copy:
   1. invoke `createMultipartUpload`
   2. invoke `multipartCopy` for all parts
   3. invoke `completeMultipartUpload`

If any error happens during 3.ii or 3.iii, the SDK will attempt to call `abortMultipartCopy` to clean up parts

## Testing


services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/CopyHelper.java
```diff
@@ -0,0 +1,248 @@
...
+                               String uploadId) {
+        long optimalPartSize = calculateOptimalPartSizeForCopy(s3NativeClientConfiguration.partSizeBytes());
+
+        int partCount = determinePartCount(contentLength, optimalPartSize);
+
+        log.debug(() -> String.format("Starting multipart copy with partCount: %s, optimalPartSize: %s",
+                                      partCount, optimalPartSize));
+
+        // The list of completed parts must be sorted
+        AtomicReferenceArray<CompletedPart> completedParts = new AtomicReferenceArray<>(partCount);
```
zoewangg, line 133: Implementation note: the reason `AtomicReferenceArray` is used instead of `ConcurrentLinkedQueue` is that the completedParts must be ordered and `AtomicReferenceArray` should be more performant in this case

zoewangg (commented):
Please ignore the test coverage and log statements; I will fix them as I finalize the PR

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/CopyRequestConversionUtils.java
```diff
@@ -0,0 +1,94 @@
...
+import software.amazon.awssdk.services.s3.model.AbortMultipartUploadRequest;
+import software.amazon.awssdk.services.s3.model.CompleteMultipartUploadResponse;
+import software.amazon.awssdk.services.s3.model.CompletedPart;
+import software.amazon.awssdk.services.s3.model.CopyObjectRequest;
+import software.amazon.awssdk.services.s3.model.CopyObjectResponse;
+import software.amazon.awssdk.services.s3.model.CopyPartResult;
+import software.amazon.awssdk.services.s3.model.CreateMultipartUploadRequest;
+import software.amazon.awssdk.services.s3.model.HeadObjectRequest;
+
+public final class CopyRequestConversionUtils {
```
dagnir, line 27: minor: annotation

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/UploadPartCopyRequestProvider.java
```diff
@@ -0,0 +1,77 @@
...
+
+import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.services.s3.model.CopyObjectRequest;
+import software.amazon.awssdk.services.s3.model.UploadPartCopyRequest;
+
+/**
+ * Factory class to create UploadPartCopyRequest
+ */
+@SdkInternalApi
+public final class UploadPartCopyRequestProvider {
```
dagnir, line 26: Do we follow this pattern elsewhere? Any reason to not just return a collection?

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/CopyHelper.java
```diff
@@ -0,0 +1,256 @@
...
+import software.amazon.awssdk.services.s3.model.UploadPartCopyRequest;
+import software.amazon.awssdk.services.s3.model.UploadPartCopyResponse;
+import software.amazon.awssdk.utils.CompletableFutureUtils;
+import software.amazon.awssdk.utils.Logger;
+
+/**
+ * An internal helper class that automatically use multipart copy based on the size of the source object
+ */
+@SdkInternalApi
+public final class CopyHelper {
```
dagnir, line 51: `CopyObjectHelper`? We have lots of I/O things and this might get confused for that

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/UploadPartCopyRequestProvider.java
zoewangg: Nope, I think I just copied what we have in v1 (not the best idea 😅 ) Will refactor it.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/S3NativeClientConfiguration.java
```diff
@@ -110,50 +99,14 @@ public URI endpointOverride() {
         return endpointOverride;
     }
 
-    public Executor futureCompletionExecutor() {
```
zoewangg, line 113: Removed this `futureCompletionExecutor` block because this is no longer needed. (Should've created a separate PR for this change 😓 I can remove this change if this makes the review difficult)

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/UploadPartCopyRequestIterable.java
```diff
@@ -0,0 +1,82 @@
...
+import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.core.pagination.sync.SdkIterable;
+import software.amazon.awssdk.services.s3.model.CopyObjectRequest;
+import software.amazon.awssdk.services.s3.model.UploadPartCopyRequest;
+
+/**
+ * Iterable class to generate {@link UploadPartCopyRequest}s
+ */
+@SdkInternalApi
+public final class UploadPartCopyRequestIterable implements SdkIterable<UploadPartCopyRequest> {
```
dagnir, line 29: Nice

dagnir (approved):

PR #3412, opened 2022-09-07
Add endpoint-rule-set.json for all services

dagnir:
Note: some of the services are placeholders, like `s3` because the rules aren't available yet.

Since `endpoints-rule-set.json` is a required model file with Endpoints 2.0, we need to have all clients have one in order to be generated.


joviegas (approved):
Just a curious , why dont we have a core file and only diff/delta in the services if they are differing from default one ?

PR #3414, opened 2022-09-07
Client endpoint integration

dagnir:
This integrates the generated endpoint providers into the request path and uses them to compute the endpoint to use for a request.

## Testing
- Unit tests pass for everything exception S3: `s3`, `s3control`, `s3outposts`
- ~~Running integ tests~~ Integ tests passed with
  ```
   mvn clean install -P integration-tests -pl \!:s3,\!:s3control,\!:s3outposts,\!:eventbridge,\!:cloudsearch,\!:machinelearning -am
  ```

EventBridge is in the same boat as S3 where there's additional work needed to make the resolution correct. CloudSearch seems to fail due to an unrelated reason, the test seems to be requesting an invalid/deprecated machine type (seems to be related to this change: https://github.com/aws/aws-sdk-java-v2/commit/8f371e0102dbe28b2de3ecfd715cb75321681c8e)


codegen/src/main/java/software/amazon/awssdk/codegen/emitters/CodeWriter.java
```diff
@@ -72,7 +75,7 @@ public CodeWriter(String dir, String file) {
      * @param fileNameSuffix
      *             suffix to be appended at the end of file name.
      */
-    public CodeWriter(String dir, String file, String fileNameSuffix) {
+    public CodeWriter(String dir, String file, String fileNameSuffix, boolean disableFormatter) {
```
dagnir, line 78: I had to add this this because the formatting time explodes on on the really big rules (e.g. S3).

dagnir:
Working on fixing broken tests

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/builder/test-client-builder-class.java
```diff
@@ -81,6 +88,10 @@ protected final String signingName() {
         return "json-service";
     }
 
+    private JsonEndpointProvider defaultEndpointProvider() {
+        return JsonEndpointProvider.defaultProvider();
+    }
```
zoewangg, line 93: nit: do we need a separate method for this?

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/builder/test-client-builder-interface.java
```diff
@@ -30,4 +16,6 @@
     default B serviceConfiguration(Consumer<ServiceConfiguration.Builder> serviceConfiguration) {
         return serviceConfiguration(ServiceConfiguration.builder().applyMutation(serviceConfiguration).build());
     }
+
+    B endpointProvider(JsonEndpointProvider endpointProvider);
```
zoewangg, line 20: Should we add Javadoc?

test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/endpointproviders/ClientBuilderTests.java
```diff
@@ -0,0 +1,119 @@
...
+                                                .region(Region.US_WEST_2)
+                                                .credentialsProvider(
+                                                    StaticCredentialsProvider.create(
+                                                        AwsBasicCredentials.create("akid", "skid")))
+                                                .build();
+
+        try {
+            client.operationWithNoInputOrOutput(r -> {
+            });
+            Thread.sleep(100);
```
zoewangg, line 104: Do we need sleep? can we use join?

test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/endpointproviders/ClientBuilderTests.java
```diff
@@ -0,0 +1,119 @@
...
+                                           .region(Region.US_WEST_2)
+                                           .credentialsProvider(StaticCredentialsProvider.create(
+                                               AwsBasicCredentials.create("akid", "skid")))
+                                           .build();
+
+        try {
+            client.operationWithNoInputOrOutput(r -> {
+            });
+        } catch (Exception ignored) {
+        }
```
zoewangg, line 66: Do we expect it to throw exception? If so, can we use `assertThatThrownBy`

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/builder/test-client-builder-class.java
dagnir: It follows the current patterns :). Don't think it's needed, but since it's here already I'll just keep it

dagnir:
Ran unit test locally and all are passing when excluding the S3 services (`s3`, `s3control`, `s3outposts`) which is expected.

Currently running integ tests.

core/sdk-core/src/main/java/software/amazon/awssdk/core/interceptor/SdkInternalExecutionAttribute.java
```diff
@@ -64,6 +64,17 @@ public final class SdkInternalExecutionAttribute extends SdkExecutionAttribute {
...
+     * The endpoint provider used to resolve the destination endpoint for a request.
+     */
+    public static final ExecutionAttribute<Object> ENDPOINT_PROVIDER =
+        new ExecutionAttribute<>("EndpointProvider");
+
+    /**
+     * Whether the endpoint on the request is the result of Endpoint Discovery.
+     */
+    public static final ExecutionAttribute<Boolean> IS_DISCOVERED_ENDPOINT =
+        new ExecutionAttribute<>("IsDiscoveredEndpoint");
```
dagnir, line 77: FYI new internal attribute

zoewangg (approved):

PR #3415, opened 2022-09-07
Fix javadoc to make it build with JDK 17

zoewangg:
## Motivation and Context
Fix javadoc so that we can use JDK 17 to generate Javadoc site

## Modifications
- Fix broken Javadoc
- Add generate-javadoc to the build process so that we can catch regressions
- Update release Javadoc buildspec

## Testing
Tested locally with JDK17 as well as CodeBuild job


pom.xml
```diff
@@ -218,6 +218,22 @@
...
+                            <configuration>
+                                <detectJavaApiLink>false</detectJavaApiLink>
+                                <source>8</source>
+                                <notree>true</notree>
+                                <includeDependencySources>false</includeDependencySources>
+                                <doclint>none</doclint>
+                            </configuration>
+                        </execution>
+                    </executions>
                 </plugin>
```
zoewangg, line 237: This will likely increase our build time because we are generating Javadocs now as part of the build.

zoewangg: Build took at least 15 mins longer after the change. We probably should not build it then or at least disable it for service modules

joviegas: - Are we going to use these Java Docs in https://github.com/aws/aws-sdk-java-v2/blob/master/buildspecs/release-javadoc.yml ?
- Why didn't we consider adding it as parallel  step , since we have failures due to intermittent test failures.

zoewangg: Nope, they are part of the build process. I tested it and it did not increase the build time w/ service modules disabled.

I considered creating another codebuild job to run the same command as release-javadoc, but the whole job can take more than one hour and it works with the largest instance, so I went with this approach.

joviegas (approved):

PR #3425, opened 2022-09-15
New RuleEngine per eval

dagnir:
RuleEngine is not threadsafe, so create a new one per evaluation.


codegen/src/main/java/software/amazon/awssdk/codegen/poet/rules/EndpointProviderSpec.java
```diff
@@ -143,8 +133,8 @@ private MethodSpec resolveEndpointMethod() {
                                          .addAnnotation(Override.class)
                                          .addParameter(endpointRulesSpecUtils.parametersClassName(), paramsName);
 
-        b.addStatement("$T res = $N.evaluate($N, toIdentifierValueMap($N))",
-                       Value.class, ENGINE_FIELD_NAME, RULE_SET_FIELD_NAME, paramsName);
+        b.addStatement("$T res = new $T().evaluate($N, toIdentifierValueMap($N))",
```
joviegas, line 136: Do we have changes in Code generated Test classes for this change ?

dagnir: Yes, this is part of a larger set of changes. Just wanted to put this one out first since :)

joviegas (approved):

PR #3437, opened 2022-09-22
Move rules engine from sdk-core to the generated clients.

millems:
1. This reduces the cross-module API service area in exchange for increased artifact size.
2. Moved rules engine testing from sdk-core to codegen-generated-classes-test module.
3. Checkstyle and spotbugs were already failing in this branch. A few failures were fixed as part of this CR, but some may not be fixed.
4. Many tests were already broken in this branch. A few are fixed as part of this CR, but most are not.
5. Because many tests were already broken, it's difficult to say with 100% certainty that this change was successful.


dagnir (approved):

dagnir:
This will probably be easier to merge after @zoewangg puts her PR out so I'll hold off on merging it.

dagnir:
Merged in internal repo. Closing this.

PR #3441, opened 2022-09-27
Add codegen and testing improvements [S3 support Pt1]

dagnir:
## Motivation and Context
 - Add support for adding static context, client context, and context params to the resolver params object
 - Support default values of the params in the params object
 - Support headers and authSchemes on Endpoint results
 - (S3) Add support for sourcing client client context params from S3Configuration
 - Improve genarated client tests to support operationInputs, and add async methods
 - Add generated endpoint provider tests for testing endpoint providers directly


codegen/src/main/java/software/amazon/awssdk/codegen/poet/builder/BaseClientBuilderClass.java
```diff
@@ -250,33 +260,13 @@ private MethodSpec finalizeServiceConfigurationMethod() {
 
         String clientConfigClassName = model.getCustomizationConfig().getServiceConfig().getClassName();
         if (StringUtils.isNotBlank(clientConfigClassName)) {
-            ClassName clientConfigClass = ClassName.bestGuess(clientConfigClassName);
```
cenedhryn, line 253: Clear improvement

core/sdk-core/src/main/java/software/amazon/awssdk/core/rules/FnNode.java
```diff
@@ -67,6 +67,10 @@ public Fn validate() {
                 return new ParseUrl(this);
             case Substring.ID:
                 return new Substring(this);
+            case UriEncodeFn.ID:
+                return new UriEncodeFn(this);
+            case IsVirtualHostableS3Bucket.ID:
```
cenedhryn, line 72: Is there any future plans to separate out service specific rules?

core/aws-core/src/main/java/software/amazon/awssdk/awscore/rules/AwsEndpointAttribute.java
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * Known endpoint attributes added by endpoint rule sets for AWS services.
+ */
+@SdkProtectedApi
+public final class AwsEndpointAttribute {
+
+    /**
+     * The auth schemes supported by the endpoint.
+     */
+    public static final EndpointAttributeKey<List<EndpointAuthScheme>> AUTH_SCHEMES =
```
cenedhryn, line 32: Referencing the conversation about auth scheme resolution in the future, will the endpoint always still have the auth scheme parameter?

core/sdk-core/src/main/java/software/amazon/awssdk/core/rules/FnNode.java
dagnir: Not how how that would be possible since the rules engine needs to be aware of it; there's no way to define and inject external rules at the moment.

core/aws-core/src/main/java/software/amazon/awssdk/awscore/rules/AwsEndpointAttribute.java
dagnir: They may be removed in future once there's a new way to resolve auth outside of endpoints.

cenedhryn (approved):

PR #3443, opened 2022-09-28
Support S3 endpoint providers

dagnir:
This adds support for using the endpoint providers in the S3 client.

## Types of changes

## Checklist

## License


services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/endpoints/UseGlobalEndpointResolver.java
```diff
@@ -0,0 +1,96 @@
...
+            ProfileFile profileFile = profileFileSupplier.get();
+            String profileName = profileNameSupplier.get();
+            if (profileFile == null || profileName == null) {
+                return null;
+            }
+            return profileFile.profile(profileName)
+                              .flatMap(p -> p.property(ProfileProperty.S3_US_EAST_1_REGIONAL_ENDPOINT))
+                              .orElse(null);
+        } catch (Exception t) {
+            //log.warn(() -> "Unable to load config file", t);
```
cenedhryn, line 91: What should it be?

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/endpoints/UseGlobalEndpointResolver.java
```diff
@@ -0,0 +1,96 @@
...
+            new Lazy<>(() -> useUsEast1RegionalEndpoint(() ->config.option(SdkClientOption.PROFILE_FILE), () -> config.option(SdkClientOption.PROFILE_NAME),
+                                                        defaultS3UsEast1RegionalEndpointFromSmartDefaults));
+    }
+
+    public boolean resolve() {
+        if (!Region.US_EAST_1.equals(sdkClientConfiguration.option(AwsClientOption.AWS_REGION))) {
+            return false;
+        }
+
+        return !useUsEast1RegionalEndpoint.getValue();
```
cenedhryn, line 54: Can we flip the internal logic in the class to reflect the name and purpose of it, so that instead of `useUsEast1RegionalEndpoint` we have `useGlobalEndpointForUsEast1` or similar? It would make reading the code more straightforward and quicker to understand, I think.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/endpoints/UseGlobalEndpointResolver.java
```diff
@@ -0,0 +1,96 @@
...
+    private final Lazy<Boolean> useUsEast1RegionalEndpoint;
+    private final SdkClientConfiguration sdkClientConfiguration;
+
+    public UseGlobalEndpointResolver(SdkClientConfiguration config) {
+        sdkClientConfiguration = config;
+        String defaultS3UsEast1RegionalEndpointFromSmartDefaults =
+            config.option(ServiceMetadataAdvancedOption.DEFAULT_S3_US_EAST_1_REGIONAL_ENDPOINT);
+        this.useUsEast1RegionalEndpoint =
+            new Lazy<>(() -> useUsEast1RegionalEndpoint(() ->config.option(SdkClientOption.PROFILE_FILE), () -> config.option(SdkClientOption.PROFILE_NAME),
+                                                        defaultS3UsEast1RegionalEndpointFromSmartDefaults));
```
cenedhryn, line 46: Nit: Can this statement be formatted so it's easier to read? Short the amount of text or line up the parameters?

The original has ` this.useUsEast1RegionalEndpoint = new Lazy<>(() -> useUsEast1RegionalEndpoint(profileFile, profileName, config))` which was easier to parse

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/handlers/RemoveBucketFromPathInterceptor.java
```diff
@@ -0,0 +1,41 @@
...
+package software.amazon.awssdk.services.s3.internal.handlers;
+
+import software.amazon.awssdk.core.interceptor.Context;
+import software.amazon.awssdk.core.interceptor.ExecutionAttributes;
+import software.amazon.awssdk.core.interceptor.ExecutionInterceptor;
+import software.amazon.awssdk.http.SdkHttpRequest;
+import software.amazon.awssdk.utils.StringUtils;
+import software.amazon.awssdk.utils.http.SdkHttpUtils;
+
+public class RemoveBucketFromPathInterceptor implements ExecutionInterceptor {
```
cenedhryn, line 25: I don't know why this interceptor is needed, can we add a few lines of documentation explaining its role?

services/s3/src/main/resources/software/amazon/awssdk/services/s3/execution.interceptors
```diff
@@ -12,4 +11,4 @@ software.amazon.awssdk.services.s3.internal.handlers.EnableTrailingChecksumInter
 software.amazon.awssdk.services.s3.internal.handlers.ExceptionTranslationInterceptor
 software.amazon.awssdk.services.s3.internal.handlers.GetObjectInterceptor
 software.amazon.awssdk.services.s3.internal.handlers.CopySourceInterceptor
-software.amazon.awssdk.services.s3.internal.handlers.SignerOverrideInterceptor
```
cenedhryn, line 15: Is the SignerOverrideInterceptor obsoleted by the endpoint auth scheme? 
Should it be removed from the code base too?

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/endpoints/UseGlobalEndpointResolver.java
dagnir: Hmm not sure I agree on this one. The way the environment variable works, a value of `regional` means we should use the regional IAD endpoint, but any other value is global. `useUsEast1RegionalEndpoint` makes it more what that value is set to IMO.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/endpoints/UseGlobalEndpointResolver.java
dagnir: done

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/endpoints/UseGlobalEndpointResolver.java
dagnir: Fixed

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/handlers/RemoveBucketFromPathInterceptor.java
dagnir: I'm just going to remove this for now. In E2.0, the S3 rules handle setting the bucket in the URI (either on the path or in the host) so we need to actually remove the one that the marshaller sets. But instead of doing this way which is actually incorrect, we will preprocess the model at codegen time instead.

services/s3/src/main/resources/software/amazon/awssdk/services/s3/execution.interceptors
dagnir: Yes

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/endpoints/UseGlobalEndpointResolver.java
cenedhryn: It's a valid point. It's still weird that the class is called useglobalendpoint and then it only talks about the regional inside... I don't love it but if you think changing it would be worse because it disguises the env var functionality that's fine.

cenedhryn (approved):

PR #3446, opened 2022-09-28
Remove Bucket param from S3 request URIs

dagnir:
## Motivation and Context
In E2.0, the rule set handles adding the Bucket to the request URI so we remove it from the model to prevent the marshallers from adding it to the path.

## Modifications
Add new processor and tests for it.


dagnir:
Forgot to note above, but also checked the generated S3 marshallers and ensure that requestUri doesn't have `{Bucket}` in them.

cenedhryn (approved):

PR #3451, opened 2022-09-29
Fixes to endpoint resolution

dagnir:
## Motivation and Context

- Don't include query params when resolving the SDK::Endpoint BuiltIn

- Correctly remove the endpoint override path from the request path before recombining the endpoints after endpoint resolution. Otherwise, the override's path will be repeated in the final endpoint.

- Fix some test expactations based around endpoint overrides in S3. Previously we did not do virtual addressing for normal bucket unless the endpoint startswith 's3'. This is no longer the case as the rules will use virtual addressing unless path style is enabled.

- Add additional tests for AwsProviderUtils

- Uncomment commented out endpoint param resolution in client builder. This was previously commented out for to make client endpoint tests pass but they will just be ignored in the generated tests instead.

### DIFF
--- a/core/aws-core/src/main/java/software/amazon/awssdk/awscore/rules/AwsProviderUtils.java
+++ b/core/aws-core/src/main/java/software/amazon/awssdk/awscore/rules/AwsProviderUtils.java
@@ -15,6 +15,8 @@
 
 package software.amazon.awssdk.awscore.rules;
 
+import static software.amazon.awssdk.utils.FunctionalUtils.invokeSafely;
+
 import java.net.URI;
 import java.util.ArrayList;
 import java.util.List;
@@ -32,6 +34,7 @@ import software.amazon.awssdk.http.SdkHttpRequest;
 import software.amazon.awssdk.regions.Region;
 import software.amazon.awssdk.utils.Logger;
 import software.amazon.awssdk.utils.StringUtils;
+import software.amazon.awssdk.utils.http.SdkHttpUtils;
 
 @SdkInternalApi
 public final class AwsProviderUtils {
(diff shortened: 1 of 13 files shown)


core/aws-core/src/main/java/software/amazon/awssdk/awscore/rules/AwsProviderUtils.java
```diff
@@ -32,6 +34,7 @@
 import software.amazon.awssdk.regions.Region;
 import software.amazon.awssdk.utils.Logger;
 import software.amazon.awssdk.utils.StringUtils;
+import software.amazon.awssdk.utils.http.SdkHttpUtils;
 
 @SdkInternalApi
 public final class AwsProviderUtils {
```
cenedhryn, line 40: Can we rename this class to AwsEndpointProviderUtils?

core/aws-core/src/main/java/software/amazon/awssdk/awscore/rules/AwsProviderUtils.java
```diff
@@ -102,23 +113,39 @@ public static Endpoint valueAsEndpointOrThrow(Value value) {
...
+    /**
+     * This sets the request URI to the resolved URI returned by the endpoint provider. There some things to be careful about
+     * to make this work properly:
+     * <p>
+     * If the client endpoint is an endpoint override, it may contain a path. In addition, the request marshaller itself may add
+     * components to the path if it's modeled for the operation. Unfortunately, {@link SdkHttpRequest#encodedPath()} returns
+     * the combined path from both the endpoint and the request. There is no way to know, just from the HTTP request object,
+     * where the override path ends (if it's even there) and where the request path starts. Additionally, the rule itself may
+     * also append other parts to the endpoint override path.
+     * <p>
```
cenedhryn, line 125: Can we add the a/b/c example

core/aws-core/src/main/java/software/amazon/awssdk/awscore/rules/AwsProviderUtils.java
```diff
@@ -102,23 +113,39 @@ public static Endpoint valueAsEndpointOrThrow(Value value) {
...
-        }
+        // [client endpoint path]/[request path]
+        String requestPath = request.getUri().getRawPath();
+
+        // [client endpoint path]/[additional path added by resolver]
+        String resolvedUriPath = resolvedUri.getRawPath();
+
+        // our goal is to construct [client endpoint path]/[additional path added by resolver]/[request path], so we just need
+        // to strip the client endpoint path from the marshalled request path to isolate just the part added by the marshaller
+        String requestPathWithClientPathRemoved = StringUtils.replaceOnce(requestPath, clientEndpointPath, "");
```
cenedhryn, line 141: can we include marshalled in the name to make it super clear?

cenedhryn (approved):

PR #3452, opened 2022-09-30
Use RFC 822 format for marshalling date shapes

dave-fn:
Add utility method so that dates are marshalled using RFC 822 format.

## Motivation and Context

The formatted representation of dates with single-digit day of month also includes a single-digit. This code fix will change the representation to two-digit day of month.

Current:
```
Expires: Fri, 1 Sep 2023 01:01:01 GMT.
```

Expected:
```
Expires: Fri, 01 Sep 2023 01:01:01 GMT
```

## Modifications
Modified DateUtils.java to add parse and format methods specific to RFC 822. Cascaded these changes to other modules.

## Types of changes

## Checklist

## License

### DIFF
--- a/utils/src/main/java/software/amazon/awssdk/utils/DateUtils.java
+++ b/utils/src/main/java/software/amazon/awssdk/utils/DateUtils.java
@@ -25,9 +25,11 @@ import java.time.Duration;
 import java.time.Instant;
 import java.time.ZoneOffset;
 import java.time.ZonedDateTime;
+import java.time.chrono.IsoChronology;
 import java.time.format.DateTimeFormatter;
 import java.time.format.DateTimeFormatterBuilder;
 import java.time.format.DateTimeParseException;
+import java.time.format.ResolverStyle;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Locale;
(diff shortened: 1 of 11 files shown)


utils/src/main/java/software/amazon/awssdk/utils/DateUtils.java
```diff
@@ -48,6 +49,18 @@ public final class DateUtils {
...
+     * RFC 822 date/time formatter.
+     */
+    static final DateTimeFormatter RFC_822_DATE_TIME = new DateTimeFormatterBuilder()
+        .parseCaseInsensitive()
+        .parseLenient()
+        .appendPattern("EEE, dd MMM yyyy HH:mm:ss")
+        .appendLiteral(' ')
+        .appendOffset("+HHMM", "GMT")
+        .toFormatter()
+        .withLocale(Locale.US);
```
joviegas, line 62: - I guess the reason for Locale.US is [V1-refernce](https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-core/src/main/java/com/amazonaws/util/DateUtils.java#L67)
- Why have we not doing 
```
toFormatter(ResolverStyle.SMART, IsoChronology.INSTANCE)
```
same as done in [RFC_1123_DATE_TIME](https://github.com/openjdk/jdk8u-dev/blob/master/jdk/src/share/classes/java/time/format/DateTimeFormatter.java#L1264) ?

dave-fn: I guess an alternative would be to set the `ResolverStye.SMART` and `IsoChonology.INSTANCE` values on the DateTimeFormatter object after it's been built. Let me know if you'd prefer me to make that change.

joviegas: Yeap , we can set it after its been built.
Can we add some test cases where we can test this too (ResolverStye and IsoChonology)?

joviegas (approved):

PR #3455, opened 2022-09-30
    Implement authSchemes support for endpoints

dagnir:
## Motivation and Context
Endpoints 2.0 support.

## Modifications
This change introduces a new ExecutionInterceptor that takes the resolved endpoints, and sets the headers and uses the auth scheme properties to modify the signing configuration for the request.

Builds off #3451


codegen/src/test/resources/software/amazon/awssdk/codegen/poet/rules/endpoint-auth-scheme-interceptor.java
```diff
@@ -0,0 +1,54 @@
...
+import software.amazon.awssdk.core.interceptor.Context;
+import software.amazon.awssdk.core.interceptor.ExecutionAttributes;
+import software.amazon.awssdk.core.interceptor.ExecutionInterceptor;
+import software.amazon.awssdk.core.interceptor.SdkInternalExecutionAttribute;
+import software.amazon.awssdk.core.rules.model.Endpoint;
+import software.amazon.awssdk.core.signer.Signer;
+
+@Generated("software.amazon.awssdk:codegen")
+@SdkInternalApi
+public final class QueryEndpointAuthSchemeInterceptor implements ExecutionInterceptor {
```
cenedhryn, line 25: Have you considered adding this logic as a part of the `AwsExecutionContextBuilder` between initial interceptors and signers instead of using an interceptor? I can't see any service-specific values at the moment.

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/rules/endpoint-auth-scheme-interceptor.java
```diff
@@ -0,0 +1,54 @@
...
+    @Override
+    public SdkRequest modifyRequest(Context.ModifyRequest context, ExecutionAttributes executionAttributes) {
+        Endpoint resolvedEndpoint = executionAttributes.getAttribute(SdkInternalExecutionAttribute.RESOLVED_ENDPOINT);
+        AwsRequest request = (AwsRequest) context.request();
+        if (resolvedEndpoint.headers() != null) {
+            request = AwsEndpointProviderUtils.addHeaders(request, resolvedEndpoint.headers());
+        }
```
cenedhryn, line 32: If all headers aren't concerned with auth schemes this is an odd place to put this logic.

core/aws-core/src/main/java/software/amazon/awssdk/awscore/rules/AwsEndpointProviderUtils.java
```diff
@@ -154,6 +160,91 @@ public static SdkHttpRequest setUri(SdkHttpRequest request, URI clientEndpoint,
                       .build();
     }
 
+    public static AwsRequest addHeaders(AwsRequest request, Map<String, List<String>> headers) {
```
cenedhryn, line 163: This method doesn't seem to belong to an endpoint utils class, it's not endpoint specific. Can we come up with a different place to put it?

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/rules/endpoint-auth-scheme-interceptor.java
dagnir: `signerProvider` is service specific :)

core/aws-core/src/main/java/software/amazon/awssdk/awscore/rules/AwsEndpointProviderUtils.java
dagnir: Meh, IMO t's a utils class *for* the providers to use. I'd rather just keep it here for now and move it if we need to to repeat this logic in other places

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/rules/endpoint-auth-scheme-interceptor.java
dagnir: Yeah I think so too. Is this a blocker? If not, I'll just keep it here to avoid adding another interceptor

core/aws-core/src/main/java/software/amazon/awssdk/awscore/rules/AwsEndpointProviderUtils.java
```diff
@@ -154,6 +160,91 @@ public static SdkHttpRequest setUri(SdkHttpRequest request, URI clientEndpoint,
...
+            .build();
+    }
+
+    /**
+     * Per the spec, the auth schemes list is ordered by preference, so we simply iterate over the list until we find an
+     * auth scheme we recognize.
+     */
+    public static EndpointAuthScheme chooseAuthScheme(List<EndpointAuthScheme> authSchemes) {
+        for (EndpointAuthScheme authScheme : authSchemes) {
+            if (SIGV4_NAME.equals(authScheme.name()) || SIGV4A_NAME.equals(authScheme.name())) {
```
cenedhryn, line 195: I'm not sure I see the need for adding the named variables like `SIGV4_NAME`. If we get a third schema then we need to add more code. Can we take the list of supported auth schemes and compare with the resolved auth schemes and pick the right one?

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/rules/endpoint-auth-scheme-interceptor.java
cenedhryn: Does it need to be an interceptor though? I know this part of the code base will continue to go through a lot of changes not only with endpoints but also with other core refactoring that's planned so I guess we can live with it if it's short/medium term. I'm just a bit concerned that we're dropping pieces of mutating logic in places where it may be hard to find them.

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/rules/endpoint-auth-scheme-interceptor.java
cenedhryn: It took me a while, I was just looking at the generated file and not the spec :-). Ok.

core/aws-core/src/main/java/software/amazon/awssdk/awscore/rules/AwsEndpointProviderUtils.java
cenedhryn: I disagree in principle but I can agree to be pragmatic in this instance. It is probably easier to find a good structure when there is another usecase.

core/aws-core/src/main/java/software/amazon/awssdk/awscore/rules/AwsEndpointProviderUtils.java
```diff
@@ -154,6 +160,91 @@ public static SdkHttpRequest setUri(SdkHttpRequest request, URI clientEndpoint,
...
+                return authScheme;
+            }
+        }
+        throw SdkClientException.create("Endpoint did not contain any known auth schemes: " + authSchemes);
+    }
+
+    public static void setSigningParams(ExecutionAttributes executionAttributes, EndpointAuthScheme authScheme) {
```
cenedhryn, line 202: Can we move the auth scheme specific code to it's own `rules/authscheme` package and also move these methods from the provider utils class? The utils class is already quite big and IMO it will make testing / reading easier. It's also easier to refactor if it's in its own package.

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/rules/endpoint-auth-scheme-interceptor.java
dagnir: The only other place it could go is the `AwsExecutionContextBuilder`, but doing so won't get rid of the interceptor. It's easier to reason about when they're all in the interceptors, as opposed to splitting them up between within the interceptors and in core IMO

cenedhryn: Let's leave it then and see what the future holds :-)

cenedhryn (approved):

Issue #3251, opened 2022-06-14
AwsCrtAsyncHttpClient - Support customization of connectionTimeout


debora-ito:
### Discussed in https://github.com/aws/aws-sdk-java-v2/discussions/3249

<div type='discussions-op-text'>

<sup>Originally posted by **mmflys** June 14, 2022</sup>
What's the default value for connectionTimeout, readTimeout in AwsCrtAsyncHttpClient and how can i change the timeout setting?</div>

mmflys:
Hi debora, could I take this feature request to implement?

zoewangg:
Hi @mmflys PRs are always welcome! :)

vaisiri:
is there any way we can set connection timeout and read timeout for AwsCrtAsyncHttpClient?

zoewangg:
This is supported now. Please try with the latest version.

PR #3456, opened 2022-10-03
Enable configuration of the connection timeout for AwsCrtAsyncHttpClient

L-Applin:
## Motivation and Context
This pull-request intent is to resolve #3251 

## Modifications
Update the `AwsCrtAsyncHttpClient.Builder` class with a method to set the connection timeout. If not specified, will default to the value in `SdkHttpConfigurationOption.GLOBAL_HTTP_DEFAULTS` which is 2 seconds. This change will therefore overwrite the default value of 3 seconds specified in the `SocketOption`.

## Testing
Created new unit test to validate field was set correctly.

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

## Checklist
- [x] I have read the [CONTRIBUTING](https://github.com/aws/aws-sdk-java-v2/blob/master/CONTRIBUTING.md) document
- [x] Local run of `mvn install` succeeds
- [x] My code follows the code style of this project
- [x] My change requires a change to the Javadoc documentation
- [x] I have updated the Javadoc documentation accordingly
- [x] I have added tests to cover my changes
- [x] All new and existing tests passed
- [x] I have added a changelog entry. Adding a new entry must be accomplished by running the `scripts/new-change` script and following the instructions. Commit the new file created by the script in `.changes/next-release` with your changes.
- [ ] My change is to implement 1.11 parity feature and I have updated [LaunchChangelog](https://github.com/aws/aws-sdk-java-v2/blob/master/docs/LaunchChangelog.md)

## License
- [x] I confirm that this pull request can be released under the Apache 2 license

PR #3457, opened 2022-10-03
Add the ability to configure SocketOptions on the AwsCrtAsyncHttpClient

nikp:
## Motivation and Context
This is necessary for any clients with long-running connections that exceed default socket timeouts of services along the call path, and need to enable keep-alive settings which the CRT client supports, but the Java client wasn't exposing to callers

## Modifications
Adds a new configuration parameter object to encapsulate SocketOptions to the AwsCrtAsyncHttpClient DefaultBuilder

## Testing

Unit testing - the change is very isolated from the rest of the package

## Types of changes

## Checklist
- [] Local run of `mvn install` succeeds
- My `mvn install` does not succeed due to `[ERROR] Failed to execute goal org.apache.maven.plugins:maven-shade-plugin:3.1.0:shade (default) on project third-party-jackson-core: Error creating shaded jar: duplicate entry: META-INF/services/software.amazon.awssdk.thirdparty.jackson.core.JsonFactory`, however the code I added is independent and I verified that the new code unit tests pass

^ I'm actually not sure about the last one - @JonathanHenson guided us to this change but I do see the LaunchChangelog references TCP Keep-Alive. Is it related?

## License


JonathanHenson (approved):
LGTM!, someone from the Java SDK team could you have a look?

L-Applin:
Just noting that this PR would also resolve resolve #3251, and make https://github.com/aws/aws-sdk-java-v2/pull/3456 obsolete as this supersedes it.

http-clients/aws-crt-client/src/main/java/software/amazon/awssdk/http/crt/SocketOptionsConfiguration.java
```diff
@@ -0,0 +1,215 @@
...
+         * @return Builder
+         */
+        Builder type(SocketOptions.SocketType type);
+
+        /**
+         * Sets the number of milliseconds before a connection will be considered timed out
+         * @param connectTimeoutMs number of milliseconds before a connection will be considered timed out
+         * @return Builder
+         */
+        Builder connectTimeoutMs(int connectTimeoutMs);
```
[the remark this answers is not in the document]
nikp: Yup, you're right. The SocketOptions expects ints to match the native code, but the builder should use Durations, and convert when setting them on SocketOptions. Will change.

http-clients/aws-crt-client/src/main/java/software/amazon/awssdk/http/crt/SocketOptionsConfiguration.java
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    /**
+     * An SDK-internal implementation of {@link Builder}.
+     */
+    private static final class DefaultSocketOptionsConfigurationBuilder implements Builder {
+
+        private SocketOptions.SocketDomain domain = SocketOptions.SocketDomain.IPv6;
+        private SocketOptions.SocketType type = SocketOptions.SocketType.STREAM;
+        private int connectTimeoutMs = 3000;
```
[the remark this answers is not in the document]
L-Applin: I discussed this with @millems and we thought that using the 2 sec default of the Global Http Default would be a better user experience than the 3 sec hard coded value hard coded in the `SocketOption` class. So I would use the value in `SdkHttpConfigurationOption.GLOBAL_HTTP_DEFAULTS`.

http-clients/aws-crt-client/src/main/java/software/amazon/awssdk/http/crt/AwsCrtAsyncHttpClient.java
```diff
@@ -343,6 +359,34 @@ Builder connectionHealthChecksConfiguration(Consumer<ConnectionHealthChecksConfi
...
+         *
+         * <p>
+         * eg: you can override socket domains and types,  connection timeouts,
+         * and enable periodic keepalive packets which are disabled out of the box, including keepalive timeouts
+         * This may be required for certain connections for longer durations than default socket timeouts
+         *
+         * @param socketOptionsConfiguration The socket configuration to use
+         * @return The builder of the method chaining.
+         */
+        Builder socketOptionsConfiguration(SocketOptionsConfiguration socketOptionsConfiguration);
```
zoewangg, line 374: This does not seem to be consistent with the APIs we have in other HTTP clients(netty and apache). I wonder if we should add more common settings like [connectionTimeout](https://github.com/aws/aws-sdk-java-v2/blob/master/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/NettyNioAsyncHttpClient.java#L280), [tcpKeepAlive](https://github.com/aws/aws-sdk-java-v2/blob/master/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/NettyNioAsyncHttpClient.java#L383) in the top level and then introduce
something like putSocketOption(SocketOption socketOption, Object value) for more "advanced setting" just like how we did in netty. This way, we don't need to make code changes if there is a new socket option and keep the number of settings less complex.


https://github.com/aws/aws-sdk-java-v2/blob/d6b02d51a69842957f19f87d8f6ef3797ad791fa/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/NettyNioAsyncHttpClient.java#L396

http-clients/aws-crt-client/src/main/java/software/amazon/awssdk/http/crt/SocketOptionsConfiguration.java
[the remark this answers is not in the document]
nikp: CC @zoewangg `SdkHttpConfigurationOption.DEFAULT_CONNECTION_TIMEOUT` is 2 seconds, but an unconfigured SocketOptions currently defaults to 3 seconds.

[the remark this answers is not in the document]
nikp: OK so I guess the real question is whether `SocketOptionsConfiguration` should have a field for `connectionTimeout` at all or if it should come from `SdkHttpConfigurationOption.CONNECTION_TIMEOUT`.

Or is that where all the KeepAlive settings should move to as well instead of a dedicated object?

http-clients/aws-crt-client/src/main/java/software/amazon/awssdk/http/crt/SocketOptionsConfiguration.java
```diff
@@ -0,0 +1,215 @@
...
+        private int keepAliveIntervalSecs = 0;
+        private int keepAliveTimeoutSecs = 0;
```
zoewangg, line 145: Instead of defining defaults here, I think we should put the configured values into the standard options map like below. The SDK will use the global defaults automatically for any options that are not configured 

https://github.com/aws/aws-sdk-java-v2/blob/b1f89898a7cd83626c0b1a01bd671594d14cd322/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/NettyNioAsyncHttpClient.java#L542-L545

https://github.com/aws/aws-sdk-java-v2/blob/b1f89898a7cd83626c0b1a01bd671594d14cd322/http-client-spi/src/main/java/software/amazon/awssdk/http/SdkHttpConfigurationOption.java#L155-L171

http-clients/aws-crt-client/src/main/java/software/amazon/awssdk/http/crt/SocketOptionsConfiguration.java
```diff
@@ -0,0 +1,215 @@
...
+     * <p>All implementations of this interface are mutable and not thread safe.</p>
+     */
+    public interface Builder {
+
+        /**
+         * Sets the socket domain
+         * @param domain socket domain
+         * @return Builder
+         */
+        Builder domain(SocketOptions.SocketDomain domain);
```
zoewangg, line 103: Do we need `domain` and `type`? If I understand correctly, there's no use-case for configuring these two with AWS services, right? We can always add them in the future if needed.

http-clients/aws-crt-client/src/main/java/software/amazon/awssdk/http/crt/AwsCrtAsyncHttpClient.java
zoewangg: If we remove `SocketDomain` and `SocketType`, I think we can add `tcpKeepAlive` and `connectTimeout` to the top level builder

http-clients/aws-crt-client/src/main/java/software/amazon/awssdk/http/crt/SocketOptionsConfiguration.java
```diff
@@ -0,0 +1,215 @@
...
+         * @return Builder
+         */
+        Builder keepAliveIntervalSecs(int keepAliveIntervalSecs);
+
+        /**
+         * Sets the number of seconds to wait for a keepalive response before considering the connection timed out
+         * @param keepAliveTimeoutSecs number of seconds to wait for a keepalive response before considering the connection timed out
+         * @return Builder
+         */
+        Builder keepAliveTimeoutSecs(int keepAliveTimeoutSecs);
```
zoewangg, line 131: Do we expect customers to configure `keepAliveIntervalSecs` and `keepAliveTimeoutSecs`? If not, should we just introduce `tcpKeepAlive(Boolean tcpKeepAlive)` and pass default values defined in the SDK to CRT? Because exposing too many knobs may make the SDK more complex to use.

http-clients/aws-crt-client/src/main/java/software/amazon/awssdk/http/crt/SocketOptionsConfiguration.java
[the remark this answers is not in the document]
zoewangg: Yeah, they should come from the SDK predefined defaults. We can just add the configured value to the standard options map in the builder setter methods. See
https://github.com/aws/aws-sdk-java-v2/pull/3457#discussion_r987157558

http-clients/aws-crt-client/src/main/java/software/amazon/awssdk/http/crt/AwsCrtAsyncHttpClient.java
nikp: I like consistency. But is there a top-level builder that is shared between Netty and Crt already?  What about keepAliveInterval? And are SocketDomain and SocketType worth exposing for configuration for other use-case (I don't have a case today)

In either case it definitely makes sense to make keepAlive a boolean since having "0" mean "off" is not idiomatic for a Java SDK

http-clients/aws-crt-client/src/main/java/software/amazon/awssdk/http/crt/SocketOptionsConfiguration.java
nikp: Can move those there, sure.

http-clients/aws-crt-client/src/main/java/software/amazon/awssdk/http/crt/SocketOptionsConfiguration.java
nikp: 👍

http-clients/aws-crt-client/src/main/java/software/amazon/awssdk/http/crt/SocketOptionsConfiguration.java
nikp: Will default to your judgement there. I suspect just being able to enable keepAlive should be enough.

nikp:
OK I think I can address the requested comments but there are 3 questions in my mind that are not clear and worth leaving a record of for future extension.

1. @L-Applin added connection timeout setting. I added keep-alive setting. Should they be top-level fields on `AwsCrtAsyncHttpClient.Builder` (peers to `connectionMaxIdleTime`), or should they be in a sub-configuration object.
2. Connection timeout already has a default in `SdkHttpConfigurationOption` of 2 seconds, however the underlying `SocketOptions` that sets the CRT options defaults to 3. Since it looks like today the DEFAULT_CONNECTION_TIMEOUT configuration isn't being applied, that means the effective default is 3 seconds. What would be considered the safe non-breaking way to resolve this inconsistency?
3. You mentioned that `NettyNioAsyncHttpClient` already supports configuration of both `connectionTimeout` and `enableTcpKeepAlive` and that the CRT client could share a common interface or base calss. Do you want to extract this interface as part of this CR (I assume to `SdkAsyncHttpClient`/`SdkAsyncHttpClient.Builder`, or leave it for a future change?

zoewangg:
> @L-Applin added connection timeout setting. I added keep-alive setting. Should they be top-level fields on AwsCrtAsyncHttpClient.Builder (peers to connectionMaxIdleTime), or should they be in a sub-configuration object.

I think they should be top-level fields. To be clear, we are going to just add `connectionTimeout(Duration connectionTimeout)` and `tcpKeepAlive(Boolean tcpKeepAlive)` right?

> Connection timeout already has a default in SdkHttpConfigurationOption of 2 seconds, however the underlying SocketOptions that sets the CRT options defaults to 3. Since it looks like today the DEFAULT_CONNECTION_TIMEOUT configuration isn't being applied, that means the effective default is 3 seconds. What would be considered the safe non-breaking way to resolve this inconsistency?

I think this should be fine. CRT HTTP Client is still in Developer Preview, so breaking change is expected

> You mentioned that NettyNioAsyncHttpClient already supports configuration of both connectionTimeout and enableTcpKeepAlive and that the CRT client could share a common interface or base calss. Do you want to extract this interface as part of this CR (I assume to SdkAsyncHttpClient/SdkAsyncHttpClient.Builder, or leave it for a future change?

They should not share a common interface actually. It's be design because not every HTTP client supports the same client options.

nikp:
> I think they should be top-level fields. To be clear, we are going to just add `connectionTimeout(Duration connectionTimeout)` and `tcpKeepAlive(Boolean tcpKeepAlive)` right?

Maybe, I'm starting to second-guess it. Your reference to NettyNioAsyncHttpClient having a boolean field made me think that we already have configured defaults somewhere for keepAliveInterval and keepAliveTimeout but we do not - actually it looks like https://github.com/aws/aws-sdk-java-v2/blob/master/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/NettyNioAsyncHttpClient.java lets `tcpKeepAlive`  be set on the Builder but isn't actually being forwarded to the client.

Which means that we either need to add sensible defaults for those values, or we let customers configure them. Since a sensible keep-alive interval very much depends on the underlying connection handler and their timeouts, I think it makes more sense to let customers configure them. Which creates weird validation edge-cases. Would it be better to just have the 2 Duration fields and if they're not set we treat it as tcpKeepAlive=false, and if they're set, true?

> I think this should be fine. CRT HTTP Client is still in Developer Preview, so breaking change is expected

That's fine, but we should be explicit. We can't change the defaults for the underlying `SocketOptions` since that's in the CRT, and would affect other languages. But is `DEFAULT_CONNECTION_TIMEOUT` in the Java SDK used for other purposes or in other packages?

> They should not share a common interface actually. It's be design because not every HTTP client supports the same client options.

An exercise for a future reader then.

http-clients/aws-crt-client/src/main/java/software/amazon/awssdk/http/crt/SocketOptionsConfiguration.java
JonathanHenson: no, keep alive is worthless without setting those values. The kernel defaults are never what anyone needs for these use-cases. So, it needs to be here.

zoewangg: My concern is that these settings may be too "advanced", so some customers may just set some random values w/o doing any research. I'm fine with adding these two, but I think we should still introduce a boolean tcpKeepAlive and define some "safe" defaults for  tcpInterval and tcpKeepAliveIntervalSecs if they are not configured. WDYT?

zoewangg:
> Your reference to NettyNioAsyncHttpClient having a boolean field made me think that we already have configured defaults somewhere for keepAliveInterval and keepAliveTimeout but we do not

Yes, we do not have defaults for keepAliveInterval and keepAliveTimeout because the existing HTTP clients don't support them. 

>  actually it looks like https://github.com/aws/aws-sdk-java-v2/blob/master/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/NettyNioAsyncHttpClient.java lets tcpKeepAlive be set on the Builder but isn't actually being forwarded to the client.

It it forwarded to the client through `AttributeMap` https://github.com/aws/aws-sdk-java-v2/blob/e547f1fdec36972f181fc807b308444c22447bff/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/NettyConfiguration.java#L104 https://github.com/aws/aws-sdk-java-v2/blob/e547f1fdec36972f181fc807b308444c22447bff/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/BootstrapProvider.java#L57 

> But is DEFAULT_CONNECTION_TIMEOUT in the Java SDK used for other purposes or in other packages?

Yes, it's used by all HTTP clients that support connect timeout, i.e., Netty and Apache HTTP client. 

> Which means that we either need to add sensible defaults for those values, or we let customers configure them. Since a sensible keep-alive interval very much depends on the underlying connection handler and their timeouts, I think it makes more sense to let customers configure them. Which creates weird validation edge-cases. Would it be better to just have the 2 Duration fields and if they're not set we treat it as tcpKeepAlive=false, and if they're set, true?

I'd vote for adding sensible defaults and keeping tcpKeepAlive. Moved this conversion to https://github.com/aws/aws-sdk-java-v2/pull/3457#discussion_r987235896

zoewangg:
> What would be considered the safe non-breaking way to resolve this inconsistency?

We could define an CRT HTTP defaults attribute map that has connect timeout of 3 seconds and merge the map in the build method. Example in netty:

https://github.com/aws/aws-sdk-java-v2/blob/e547f1fdec36972f181fc807b308444c22447bff/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/NettyNioAsyncHttpClient.java#L77-L80

https://github.com/aws/aws-sdk-java-v2/blob/e547f1fdec36972f181fc807b308444c22447bff/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/NettyNioAsyncHttpClient.java#L726-L730

http-clients/aws-crt-client/src/main/java/software/amazon/awssdk/http/crt/SocketOptionsConfiguration.java
nikp: IMO I think the right way to do this would be to separate KeepAliveConfiguration into a dedicated object.

If it's not set, that means customers want the default, disabled. If they DO set it, then they should be setting both the interval and the timeout, and they're "opting-in" to knowing what they're doing.

JonathanHenson: It's either add these values or don't bother. The kernel defaults will not meet anyone's needs. It's on the scale of hours before the first probe gets sent.  It IS too advanced, but the answer isn't to do it for the user. By the time you're tuning keep alive, both of these values need to be a conscious decision by the user.

zoewangg: @nikp I think a separate class is fine. Can we name it `TcpKeepAliveConfiguration` though? because people may confuse it with HTTP keep alive (we've had customers asking for clarification before)

@JonathanHenson I think that's fair. We'd probably need to provide some guidance on how to tune them.

nikp: I'm gonna submit a new revision with that approach - I think it'll work nicely :)

http-clients/aws-crt-client/src/main/java/software/amazon/awssdk/http/crt/SocketOptionsConfiguration.java
nikp: Based on the discussion in https://github.com/aws/aws-sdk-java-v2/pull/3457#discussion_r987166959, i think it doesn't make sense to list defaults for these

http-clients/aws-crt-client/src/main/java/software/amazon/awssdk/http/crt/SocketOptionsConfiguration.java
nikp: I ended up keeping them for the advanced users, since it's all going to the same place in the CRT, but decoupled from both connection timeouts, and the keep-alive configuration so that SDK users discovering this can get their iteratively

zoewangg: IMO we should only expose public APIs if needed to keep it a 2-way door because it'd be a lot harder to remove/change them in the future.

http-clients/aws-crt-client/src/main/java/software/amazon/awssdk/http/crt/AwsCrtAsyncHttpClient.java
```diff
@@ -138,6 +150,34 @@ private HttpProxyOptions buildProxyOptions(ProxyConfiguration proxyConfiguration
         return clientProxyOptions;
     }
 
+    private SocketOptions buildSocketOptions(DefaultBuilder builder, AttributeMap config) {
+        SocketOptions clientSocketOptions = new SocketOptions();
+
+        Duration connectionTimeout = config.get(SdkHttpConfigurationOption.CONNECTION_TIMEOUT);
+        if (connectionTimeout != null) {
+            clientSocketOptions.connectTimeoutMs = (int) Long.min(connectionTimeout.toMillis(), Integer.MAX_VALUE);
```
L-Applin, line 158: `NumericUtils.saturatedCast(connectionTimeout.toMillis())` can be used instead
(same for `keepAliveIntervalSecs` and `keepAliveTimeoutSecs` L166 and L167)

http-clients/aws-crt-client/src/main/java/software/amazon/awssdk/http/crt/SocketOptionsConfiguration.java
nikp: Are you saying I should remove `SocketOptionsConfiguration` from this change so we can decide later how to expose them? I'm fine with that

http-clients/aws-crt-client/src/main/java/software/amazon/awssdk/http/crt/SocketOptionsConfiguration.java
nikp: I submitted TcpKeepAliveConfiguration, please take a look @zoewangg

nikp:
OK I've removed SocketOptionsConfiguration to leave that for a future change, and applied the rest of the feedback.

http-clients/aws-crt-client/src/main/java/software/amazon/awssdk/http/crt/AwsCrtAsyncHttpClient.java
```diff
@@ -417,10 +500,39 @@ public Builder connectionHealthChecksConfiguration(Consumer<ConnectionHealthChec
 
         @Override
         public Builder connectionMaxIdleTime(Duration connectionMaxIdleTime) {
+            Validate.isPositive(connectionMaxIdleTime, "connectionMaxIdleTime");
```
zoewangg, line 503: Thanks for adding the validation!

http-clients/aws-crt-client/src/main/java/software/amazon/awssdk/http/crt/AwsCrtAsyncHttpClient.java
```diff
@@ -340,9 +371,58 @@ Builder connectionHealthChecksConfiguration(Consumer<ConnectionHealthChecksConfi
...
+         * <p>
+         * By default, this is disabled.
+         * <p>
+         * When enabled, the actual KeepAlive mechanism is dependent on the Operating System and therefore additional TCP
+         * KeepAlive values (like timeout, number of packets, etc) must be configured via the Operating System (sysctl on
+         * Linux/Mac, and Registry values on Windows).
+         *
+         * To enable, keepAlive interval and timeout must be additionally configured via {@link TcpKeepAliveConfiguration}
+         */
+        Builder tcpKeepAlive(Boolean keepConnectionAlive);
```
zoewangg, line 398: The documentation doesn't seem to be correct here since it does not correspond to the Java tcp keepalive implementation. I thought we are going to just remove the boolean tcpKeepAlive since we decided to require customers to provide tcpKeepAlive timeouts (Sorry for being unclear)

nikp: I tried to have it both ways - have the boolean flag matching the Netty interface, AND have the additional configuration flags since they're important.

But you're right that doesn't provide enough value, will take it out.

zoewangg:
LGTM. Running the builds now

zoewangg:
@nikp looks like there are some checkstyle errors. We can also take it over here if that works better 

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:3.1.2:check (checkstyle) on project aws-crt-client: Failed during checkstyle execution: 
There are 4 errors reported by Checkstyle 8.42 with software/amazon/awssdk/checkstyle.xml ruleset. -> [Help 1]
```

You can try locally using the following commands:

```
mvn clean install -pl :aws-crt-client -P quick --am
mvn clean install -pl :bom-internal
mvn clean install -pl :aws-crt-client
```

zoewangg (approved):
Thank you for your contribution!

PR #3460, opened 2022-10-04
Fix handling of FIPS pseudo regions

dagnir:
## Motivation and Context
Compliant with SDK standard behavior.

## Modifications
This commit adds support for automatically transforming a FIPS pesudo region (e.g. "fips-us-west-2"), into a standard region name + setting the fips_enabled flag to true.

## Testing

New unit tests.

## Types of changes

## Checklist

## License


core/aws-core/src/main/java/software/amazon/awssdk/awscore/client/builder/AwsDefaultClientBuilder.java
```diff
@@ -374,7 +377,13 @@ private RetryPolicy resolveAwsRetryPolicy(SdkClientConfiguration config) {
...
     @Override
     public final BuilderT region(Region region) {
-        clientConfiguration.option(AwsClientOption.AWS_REGION, region);
+        if (idHasFipsPattern(region)) {
+            clientConfiguration.option(AwsClientOption.AWS_REGION, removeFips(region));
+            clientConfiguration.option(AwsClientOption.FIPS_ENDPOINT_ENABLED, true);
+        } else {
+            clientConfiguration.option(AwsClientOption.AWS_REGION, region);
+            clientConfiguration.option(AwsClientOption.FIPS_ENDPOINT_ENABLED, null);
+        }
```
millems, line 386: Performance nit: can we do the starts-with and ends-with check once? Can we also do substring to remove the prefix/suffix instead of replaceOnce (which can behave weird for things like foo-fips-1-fips)?

core/aws-core/src/main/java/software/amazon/awssdk/awscore/client/builder/AwsDefaultClientBuilder.java
```diff
@@ -433,4 +442,22 @@ public final BuilderT defaultsMode(DefaultsMode defaultsMode) {
...
+    private Region removeFips(Region region) {
+        if (region.id().startsWith(FIPS_PREFIX)) {
+            String prefixRemoved = StringUtils.replaceOnce(region.id(), FIPS_PREFIX, "");
+            return Region.of(prefixRemoved);
+        }
+
+        if (region.id().endsWith(FIPS_SUFFIX)) {
+            String suffixRemoved = StringUtils.replaceOnce(region.id(), FIPS_SUFFIX, "");
+            return Region.of(suffixRemoved);
+        }
```
millems, line 459: are we supposed to clean double-fips stuff? fips-test-1-fips -> test-1?

dagnir: The spec doesn't call this out specifically, nor do the test cases. I'll assume no for now. It *does* call out some infix cases which I'll fix.

core/aws-core/src/main/java/software/amazon/awssdk/awscore/client/builder/AwsDefaultClientBuilder.java
dagnir: Weird, good callout on the behavior. Will fix. Will fix avoiding the multiple calls to starts/endsWith.

dagnir: Fixed. I optimized for the prefix/suffix case since that's likely to be the majority anyway. In any case, this only happens during client build time.

core/aws-core/src/main/java/software/amazon/awssdk/awscore/client/builder/AwsDefaultClientBuilder.java
```diff
@@ -433,4 +440,46 @@ public final BuilderT defaultsMode(DefaultsMode defaultsMode) {
     public final void setDefaultsMode(DefaultsMode defaultsMode) {
         defaultsMode(defaultsMode);
     }
+
+    /**
+     * If the region is a FIPS pseudo region (contains "fips"), this method returns a pair of values, the left side being the
+     * region with the "fips" string removed, and the right being {@code true}. Otherwise, the region is returned
+     * unchanged, and the right will be empty.
+     */
+    private static Pair<Region, Optional<Boolean>> transformFipsPseudoRegionIfNecessary(Region region) {
```
dagnir, line 449: note: `Optional<Boolean>` because `Pair` doesn't allow values to be null

core/aws-core/src/test/java/software/amazon/awssdk/awscore/client/builder/FipsPseudoRegionTest.java
```diff
@@ -0,0 +1,141 @@
...
+    @MethodSource("testCases")
+    public void verifyRegion(TestCase tc) {
+        TestClient client = testClientBuilder().region(tc.inputRegion).build();
+
+        assertThat(client.clientConfiguration.option(AwsClientOption.AWS_REGION)).isEqualTo(tc.expectedClientRegion);
+        assertThat(client.clientConfiguration.option(AwsClientOption.FIPS_ENDPOINT_ENABLED)).isTrue();
+    }
+
+    public static List<TestCase> testCases() {
+        List<TestCase> testCases = new ArrayList<>();
```
cenedhryn, line 70: Can we add a testcase where fips occurs twice?

dagnir: Done

core/aws-core/src/main/java/software/amazon/awssdk/awscore/client/builder/AwsDefaultClientBuilder.java
```diff
@@ -433,4 +440,46 @@ public final BuilderT defaultsMode(DefaultsMode defaultsMode) {
...
+        if (id.contains(FIPS_PREFIX)) {
+            String infixRemoved = StringUtils.replaceOnce(id, FIPS_PREFIX, "");
+            return Pair.of(Region.of(infixRemoved), Optional.of(true));
+        }
+
+        // "rekognition-fips.us-west-2"
+        if (id.contains(FIPS_SUFFIX)) {
+            String infixRemoved = StringUtils.replaceOnce(id, FIPS_SUFFIX, "");
+            return Pair.of(Region.of(infixRemoved), Optional.of(true));
+        }
```
millems, line 481: Since we apparently have to handle infix, can we just do...
```java
String[] find = { "fips-", "-fips" }; // Should probably make these static to avoid needing to allocate a new array each time.
String[] replace = { "", "" };
return StringUtils.replaceEach(id, find, replace);
```

It's slightly different, since it completely strips "fips-" and "-fips" regardless of how many times it occurs, but [that's what ruby seems to do](https://github.com/aws/aws-sdk-ruby/blob/45214f454e35ecc35b0e9d531ce7db7c4a7d81a3/gems/aws-sdk-core/lib/aws-sdk-core/plugins/regional_endpoint.rb#L65).

core/aws-core/src/main/java/software/amazon/awssdk/awscore/client/builder/AwsDefaultClientBuilder.java
millems: Can we use False instead of null? Seems like currently we never have Optional(False).

dagnir: `false` seems weird and would wouldn't work just setting the value directly since now we'd be setting an explicit value for fips_enabled in cases where the region is "normal", and would supercede sourcing the FIPS value from the environment https://github.com/aws/aws-sdk-java-v2/blob/master/core/aws-core/src/main/java/software/amazon/awssdk/awscore/client/builder/AwsDefaultClientBuilder.java#L324

core/aws-core/src/main/java/software/amazon/awssdk/awscore/client/builder/AwsDefaultClientBuilder.java
dagnir: Cool that works for me.

cenedhryn (approved):

PR #3466, opened 2022-10-06
Add LRU Cache to utilities

cenedhryn:
## Motivation and Context
An LRU (Least Recently Used) cache is needed when you need to store resources in memory in order of when they've been accessed. No existing utility exists for this functionality within our dependencies. 

## Modifications
The LRU cache 
- has a standard size of 100 (configurable)
- returns the value for the key that is given. 
  - If the value exists, the cache will move it to the front (most-recently-used) part of the cache before returning it
  - If no value exists, the cache will used the supplied function to retrieve a value and add it to the front of the cache. If the cache is full, the least recently value will be evicted. 
- Supports concurrency
- Does not (yet) support individual refresh of the cached values. This is a planned feature add-on.

## Testing
Unit tested

## Types of changes

## Checklist

## License

### DIFF
--- a/utils/src/main/java/software/amazon/awssdk/utils/cache/lru/LruCache.java
+++ b/utils/src/main/java/software/amazon/awssdk/utils/cache/lru/LruCache.java
@@ -20,25 +20,35 @@ import java.util.Objects;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Function;
 import software.amazon.awssdk.annotations.SdkProtectedApi;
+import software.amazon.awssdk.annotations.ThreadSafe;
 import software.amazon.awssdk.utils.Validate;
 
 /**
- * An LRU (Least Recently Used) cache implementation.
+ * A thread-safe LRU (Least Recently Used) cache implementation that returns the value for a specified key,
+ * retrieving it by either getting the stored value from the cache or using a supplied function to calculate that value
+ * and add it to the cache.
  * <p>
- * The user can configure the maximum size of the cache, which is set to a default of 100. Null values are accepted.
+ * When the cache is full, a new value will push out the least recently used value.
+ * When the cache is queried for an already stored value (cache hit), this value is moved to the back of the queue
+ * before it's returned so that the order of most recently used to least recently used can be maintained.
+ * <p>
+ * The user can configure the maximum size of the cache, which is set to a default of 100.
+ * <p>
+ * Null values are accepted.
  */
 @SdkProtectedApi
+@ThreadSafe
 public final class LruCache<K, V>  {
 
     private static final int DEFAULT_SIZE = 100;
 
     private final Map<K, CacheEntry<K, V>> cache;
     private final Function<K, V> valueSupplier;
+    private final Object listLock = new Object();
+    private final int maxCacheSize;
 
     private CacheEntry<K, V> leastRecentlyUsed = null;
     private CacheEntry<K, V> mostRecentlyUsed = null;
-    private final Object listLock = new Object();
-    private final int maxCacheSize;
 
     private LruCache(Builder<K, V> b) {
         this.valueSupplier = b.supplier;
@@ -47,15 +57,21 @@ public final class LruCache<K, V>  {
         this.cache = new ConcurrentHashMap<>();
     }
 
-
     /**
      * Get a value based on the key. If the value exists in the cache, it's returned, and it's position in the cache is updated.
      * Otherwise, the value is calculated based on the supplied function {@link Builder#builder(Function)}.
      */
     public V get(K key) {
-        CacheEntry<K, V> cachedEntry = cache.computeIfAbsent(key, this::newEntry);
-        updateCacheOrder(cachedEntry);
-        return cachedEntry.value();
+        while (true) {
+            CacheEntry<K, V> cachedEntry = cache.computeIfAbsent(key, this::newEntry);
+            synchronized (listLock) {
+                if (cachedEntry.evicted()) {
+                    continue;
+                }
+                moveToBackOfQueue(cachedEntry);
+                return cachedEntry.value();
+            }
+        }
     }
 
     private CacheEntry<K, V> newEntry(K key) {
@@ -64,55 +80,32 @@ public final class LruCache<K, V>  {
     }
 
     /**
-     * Sets an existing entry as the most recently used (MRU). If the entry is already the MRU, do nothing.
-     * <p>
-     * If the entry is the least recently used item (LRU), set the LRU pointer to the item before the current LRU; the list is
-     * guaranteed to have at least 2 items otherwise execution would have returned already (if only one item it was the MRU too).
+     * Moves an entry to the back of the queue and sets it as the most recently used. If the entry is already the
+     * most recently used, do nothing.
      * <p>
      * Summary of cache update:
      * <ol>
-     * <li>Detach the entry from its current place in the double linked list by letting its previous neighbor point to its
-     *    next neighbor, and vice versa.</li>
-     * <li>This is the new most recently used item so there is no previous entry. Point to the most recently used entry.</li>
-     * <li>Set the most recently used pointer to point to the new entry.</li>
+     * <li>Detach the entry from its current place in the double linked list.</li>
+     * <li>Add it to the back of the queue (most recently used)</li>
      *</ol>
      */
-    private void updateCacheOrder(CacheEntry<K, V> entry) {
-        synchronized (listLock) {
-            if (entry.equals(mostRecentlyUsed)) {
-                return;
-            }
-            if (entry.equals(leastRecentlyUsed)) {
-                leastRecentlyUsed = entry.previous();
-            }
-            detachIfNeeded(entry);
-            if (listIsNotInitialized()) {
-                leastRecentlyUsed = entry;
-            } else {
-                entry.setNext(mostRecentlyUsed);
-                mostRecentlyUsed.setPrevious(entry);
-            }
-            mostRecentlyUsed = entry;
-            if (size() > maxCacheSize) {
-                evict();
-            }
+    private void moveToBackOfQueue(CacheEntry<K, V> entry) {
+        if (entry.equals(mostRecentlyUsed)) {
+            return;
         }
-    }
-
-    private boolean listIsNotInitialized() {
-        return mostRecentlyUsed == null && leastRecentlyUsed == null;
+        removeFromQueue(entry);
+        addToQueue(entry);
     }
 
     /**
      * Detaches an entry from its neighbors in the cache. Remove the entry from its current place in the double linked list
      * by letting its previous neighbor point to its next neighbor, and vice versa, if those exist.
      * <p>
-     * An entry may not have a previous neighbor if it's currently the first one in the list. It may not have a next neighbor
-     * if it's the last entry.
+     * The least-recently-used and most-recently-used pointers are reset if needed.
      * <p>
      * <b>Note:</b> Detaching an entry does not delete it from the cache hash map.
      */
-    private void detachIfNeeded(CacheEntry<K, V> entry) {
+    private void removeFromQueue(CacheEntry<K, V> entry) {
         CacheEntry<K, V> previousEntry = entry.previous();
         if (previousEntry != null) {
             previousEntry.setNext(entry.next());
@@ -121,17 +114,40 @@ public final class LruCache<K, V>  {
         if (nextEntry != null) {
             nextEntry.setPrevious(entry.previous());
         }
-        entry.setPrevious(null);
+        if (entry.equals(leastRecentlyUsed)) {
+            leastRecentlyUsed = entry.previous();
+        }
+        if (entry.equals(mostRecentlyUsed)) {
+            mostRecentlyUsed = entry.next();
+        }
     }
 
     /**
-     * Removes the least recently used entry from the cache.
-     * The pointer to the least recently used entry is updated to point to the next-but-last entry.
-     * The next pointer of the new least recently used entry is updated to null, since it's last.
+     * Adds an entry to the queue as the most recently used, adjusts all pointers and triggers an evict
+     * event if the cache is now full.
+     */
+    private void addToQueue(CacheEntry<K, V> entry) {
+        if (mostRecentlyUsed != null) {
+            mostRecentlyUsed.setPrevious(entry);
+            entry.setNext(mostRecentlyUsed);
+        }
+        entry.setPrevious(null);
+        mostRecentlyUsed = entry;
+        if (leastRecentlyUsed == null) {
+            leastRecentlyUsed = entry;
+        }
+        if (size() > maxCacheSize) {
+            evict();
+        }
+    }
+
+    /**
+     * Removes the least recently used entry from the cache, marks it as evicted and removes it from the queue.
      */
     private void evict() {
+        leastRecentlyUsed.isEvicted(true);
         cache.remove(leastRecentlyUsed.key());
-        leastRecentlyUsed = leastRecentlyUsed.previous();
+        removeFromQueue(leastRecentlyUsed);
     }
 
     public int size() {
@@ -145,13 +161,13 @@ public final class LruCache<K, V>  {
     public static final class Builder<K, V> {
 
         private final Function<K, V> supplier;
-        private int maxSize;
+        private Integer maxSize;
 
         private Builder(Function<K, V> supplier) {
             this.supplier = supplier;
         }
 
-        public Builder<K, V> maxSize(int maxSize) {
+        public Builder<K, V> maxSize(Integer maxSize) {
             this.maxSize = maxSize;
             return this;
         }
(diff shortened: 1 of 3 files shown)


utils/src/main/java/software/amazon/awssdk/utils/cache/lru/LruCache.java
```diff
@@ -0,0 +1,236 @@
...
+        this.valueSupplier = b.supplier;
+        Integer customSize = Validate.isPositiveOrNull(b.maxSize, "size");
+        this.maxCacheSize = customSize != null ? customSize : DEFAULT_SIZE;
+        this.cache = new ConcurrentHashMap<>();
+    }
+
+    public V get(K key) {
+        CacheEntry<K, V> cachedEntry = cache.get(key);
+        if (cachedEntry != null) {
+            moveToFront(cachedEntry);
```
millems, line 55: nit: I think of the 'front' of the queue as the leastRecentlyUsed, not the mostRecentlyUsed. Maybe just me.

utils/src/main/java/software/amazon/awssdk/utils/cache/lru/LruCache.java
```diff
@@ -0,0 +1,236 @@
...
+            return cachedEntry.value();
+        }
+        V value = valueSupplier.apply(key);
+        CacheEntry<K, V> newEntry = new CacheEntry<>(key, value);
+        addToFront(newEntry);
+        return newEntry.value();
+    }
```
millems, line 62: Should we use `cache.computeIfAbsent` instead of `cache.get` and `cache.put`? That would lock around the key, so that the `valueSupplier` would only be executed once, even with concurrent `LruCache.get` calls.

utils/src/main/java/software/amazon/awssdk/utils/cache/lru/LruCache.java
```diff
@@ -0,0 +1,236 @@
...
+            remove(entry);
+            entry.setPrevious(null);
+            entry.setNext(mostRecentlyUsed);
+            mostRecentlyUsed.setPrevious(entry);
+            mostRecentlyUsed = entry;
+        } finally {
+            if (lockAcquired) {
+                listLock.unlock();
+            }
+        }
```
millems, line 90: You could use `private final Object listLock = new Object();` and use `synchronized(listLock)` instead of explicit lock operations on an re-entrant lock. It might read better.

utils/src/main/java/software/amazon/awssdk/utils/cache/lru/LruCache.java
```diff
@@ -0,0 +1,236 @@
...
+     * 2) This is the new most recently used item so there is no previous entry. Point to the most recently used entry.
+     * 3) Set the most recently used pointer to point to the new entry.
+     *
+     */
+    private void moveToFront(CacheEntry<K, V> entry) {
+        if (entry.equals(mostRecentlyUsed)) {
+            return;
+        }
+
+        boolean lockAcquired = listLock.tryLock();
```
millems, line 78: Shouldn't we be using `lock()` instead of `tryLock()`?

utils/src/main/java/software/amazon/awssdk/utils/cache/lru/LruCache.java
```diff
@@ -0,0 +1,236 @@
...
+import software.amazon.awssdk.annotations.SdkProtectedApi;
+import software.amazon.awssdk.utils.Validate;
+
+/**
+ * An LRU (Least Recently Used) cache implementation.
+ * <p>
+ * The user can configure the maximum size of the cache, which is set to a default of 100.
+ */
+@SdkProtectedApi
+public final class LruCache<K, V>  {
```
millems, line 33: This looks good for sync, but we may also need to write a non-blocking implementation at some point as well...

utils/src/main/java/software/amazon/awssdk/utils/cache/lru/LruCache.java
```diff
@@ -0,0 +1,236 @@
...
+     */
+    private void remove(CacheEntry<K, V> entry) {
+        CacheEntry<K, V> previousEntry = entry.previous();
+        if (previousEntry != null) {
+            previousEntry.setNext(entry.next());
+        }
+        CacheEntry<K, V> nextEntry = entry.next();
+        if (nextEntry != null) {
+            nextEntry.setPrevious(entry.previous());
+        }
```
millems, line 105: if previous or next are null, does that mean we should update leastRecentlyUsed or mostRecentlyUsed (because they're probably pointing to this entry)?

utils/src/main/java/software/amazon/awssdk/utils/cache/lru/LruCache.java
```diff
@@ -0,0 +1,236 @@
...
+     * The pointer to the least recently used entry is updated to point to the next-but-last entry.
+     * The next pointer of the new least recently used entry is updated to null, since it's last.
+     *
+     * @return The value of the evicted entry
+     */
+    private V evict() {
+        CacheEntry<K, V> evictedEntry = leastRecentlyUsed;
+        leastRecentlyUsed = evictedEntry.previous();
+        leastRecentlyUsed.setNext(null);
+        evictedEntry.setPrevious(null);
```
millems, line 145: Is this necessary?

utils/src/main/java/software/amazon/awssdk/utils/cache/lru/LruCache.java
```diff
@@ -0,0 +1,236 @@
...
+        cache.remove(evictedEntry.key());
+        return evictedEntry.value();
+    }
+
+    public Integer size() {
```
millems, line 150: `int`?

utils/src/main/java/software/amazon/awssdk/utils/cache/lru/LruCache.java
millems: I actually think this is okay for async. We'd just use a supplier that returns CompletableFuture<_>. The only place this requires blocking is for locking in the cache and queue, which should be okay.

utils/src/main/java/software/amazon/awssdk/utils/cache/lru/LruCache.java
cenedhryn: Yes, there seems to be a bug in there.

utils/src/main/java/software/amazon/awssdk/utils/cache/lru/LruCache.java
cenedhryn: Rephrasing this to fit with most developer's sense of terminology

utils/src/main/java/software/amazon/awssdk/utils/cache/lru/LruCache.java
cenedhryn: Yes, will refactor.

utils/src/main/java/software/amazon/awssdk/utils/cache/lru/LruCache.java
cenedhryn: Using synchronized instead

utils/src/main/java/software/amazon/awssdk/utils/cache/lru/LruCache.java
cenedhryn: Not entirely. Perhaps less is more.

utils/src/test/java/software/amazon/awssdk/utils/cache/lru/LruCacheTest.java
```diff
@@ -0,0 +1,372 @@
...
+            LruCache<Key, Value> cachedSupplier = LruCache.builder(waitingSupplier)
+                                                          .maxSize(5)
+                                                          .build();
+
+            waitingSupplier.permits.release(5);
+            waitFor(performAsyncGets(cachedSupplier, 8));
+
+            // Make extra sure only 2 "gets" actually happened.
+            waitingSupplier.waitForGetsToHaveFinished(2);
+        }
```
millems, line 236: The test name and choice to use 5 and 8 seem important to me when I'm reading, but they actually seem to be arbitrary. Almost any numbers would work there. Can we make it clearer that the only number that's particularly important here is 2?

utils/src/test/java/software/amazon/awssdk/utils/cache/lru/LruCacheTest.java
```diff
@@ -0,0 +1,372 @@
...
+    }
+
+    /**
+     * Asynchronously perform multiple "gets" on the provided supplier, returning the collection of futures to be completed when
+     * the "get" finishes.
+     */
+    private List<Future<?>> performAsyncGets(LruCache<Key, Value> supplier, int count) {
+        List<Future<?>> futures = new ArrayList<>();
+        for (int i = 0; i < count; ++i) {
+            futures.add(performAsyncGet(supplier, simpleTestKeys.get(count % 2)));
```
millems, line 270: The name of the test method hides the very important point that the parallel calls are only across 2 keys. Is there a way to make that clearer in the method name or by making it a parameter?

utils/src/test/java/software/amazon/awssdk/utils/cache/lru/LruCacheTest.java
```diff
@@ -0,0 +1,372 @@
...
+            }
+            Value that = (Value) o;
+            return Objects.equals(param1, that.param1);
+        }
+
+        @Override
+        public int hashCode() {
+            return param1 != null ? param1.hashCode() : 0;
+        }
+    }
```
millems, line 357: Any particular reason to not just use Integer and String directly instead of an explicit Key and Value type?

utils/src/test/java/software/amazon/awssdk/utils/cache/lru/LruCacheTest.java
```diff
@@ -0,0 +1,372 @@
...
+        result = cache.get(simpleTestKeys.get(5));
+        assertCacheState(cache, result, MAX_SIMPLE_CACHE_SIZE, 5);
+
+        //move 4 up
+        result = cache.get(simpleTestKeys.get(4));
+        assertCacheState(cache, result, MAX_SIMPLE_CACHE_SIZE, 4);
+
+        //evict 3
+        result = cache.get(simpleTestKeys.get(2));
+        assertCacheState(cache, result, MAX_SIMPLE_CACHE_SIZE, 2);
```
millems, line 196: All of our evictions are in the original ordering (1, 2, 3), so it doesn't seem like we're actually testing that we're "moving things up". Should we evict 3 earlier than 2 so that we know the 'moving up' is working?

utils/src/test/java/software/amazon/awssdk/utils/cache/lru/LruCacheTest.java
```diff
@@ -0,0 +1,372 @@
...
+        assertThat(result).isEqualTo(simpleTestValues.get(2));
+
+        cache.get(simpleTestKeys.get(2));
+
+        int onlyOneInvocation = 1;
+        verify(simpleValueSupplier, times(onlyOneInvocation)).apply(simpleTestKeys.get(2));
+    }
+
+    @Test
+    void when_cacheFillsUp_ValuesAreEvictedFromCache() {
```
millems, line 173: It seems like this test is verifying 3 things: we evict when we're full, we promote things when they're used, and we don't call the underlying supplier more than needed. Should we split it into 3 tests? The name only seems to imply we are testing 1 of those things.

utils/src/test/java/software/amazon/awssdk/utils/cache/lru/LruCacheTest.java
```diff
@@ -0,0 +1,372 @@
...
+                int numEntry = ThreadLocalRandom.current().nextInt(MAX_SIMPLE_TEST_ENTRIES - 1);
+                executor.submit(() -> cache.get(simpleTestKeys.get(numEntry)));
+            }
+
+            executor.shutdown();
+            assertThat(executor.awaitTermination(10, TimeUnit.SECONDS)).isTrue();
+        } finally {
+            executor.shutdownNow();
+        }
+    }
```
millems, line 221: It's not clear what this is testing. What's the failure case?

utils/src/test/java/software/amazon/awssdk/utils/cache/lru/LruCacheTest.java
```diff
@@ -0,0 +1,372 @@
...
+
+    @Test
+    void when_multipleCallers_oneCallerBlock_doesnt_Work() {
```
millems, line 224: I don't understand the test name (can't tell what it's supposed to be testing).

utils/src/test/java/software/amazon/awssdk/utils/cache/lru/LruCacheTest.java
cenedhryn: Since there isn't an API that extracts the cache contents verification is indirect through checking calls to the supplier. It's hard to split the test strictly.

utils/src/test/java/software/amazon/awssdk/utils/cache/lru/LruCacheTest.java
cenedhryn: This test case was refactored and improved.

utils/src/test/java/software/amazon/awssdk/utils/cache/lru/LruCacheTest.java
cenedhryn: It didn't work, and has been removed.

utils/src/test/java/software/amazon/awssdk/utils/cache/lru/LruCacheTest.java
cenedhryn: Started out generically but it reads simpler if it's using concrete types, agreed

utils/src/main/java/software/amazon/awssdk/utils/cache/lru/LruCache.java
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+
+    /**
+     * Get a value based on the key. If the value exists in the cache, it's returned, and it's position in the cache is updated.
+     * Otherwise, the value is calculated based on the supplied function {@link Builder#builder(Function)}.
+     */
+    public V get(K key) {
+        CacheEntry<K, V> cachedEntry = cache.computeIfAbsent(key, this::newEntry);
+        updateCacheOrder(cachedEntry);
```
millems, line 57: Because the cache and list are synchronized independently, it's possible that `cachedEntry` could be evicted when we call `updateCacheOrder`.

We could fix this by adding an 'evicted' flag to each entry that we set to true before we release the list lock during entry eviction, and check is false after we acquire the list lock. Since it's highly unlikely for the 'evicted' flag to be true, if it does happen we could probably just release the lock and redo the get() call.

The alternative to all of this is to just synchronize the whole thing and switch to a non-concurrent map. That might slow down the read path a bit, even if we switch to read/write locks.

utils/src/main/java/software/amazon/awssdk/utils/cache/lru/LruCache.java
```diff
@@ -0,0 +1,223 @@
...
+            if (listIsNotInitialized()) {
+                leastRecentlyUsed = entry;
+            } else {
+                entry.setNext(mostRecentlyUsed);
+                mostRecentlyUsed.setPrevious(entry);
+            }
+            mostRecentlyUsed = entry;
+            if (size() > maxCacheSize) {
+                evict();
+            }
```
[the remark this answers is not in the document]
cenedhryn: Making the logic easier to follow is a good suggestion.

utils/src/main/java/software/amazon/awssdk/utils/cache/lru/LruCache.java
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    /**
+     * Removes the least recently used entry from the cache.
+     * The pointer to the least recently used entry is updated to point to the next-but-last entry.
+     * The next pointer of the new least recently used entry is updated to null, since it's last.
+     */
+    private void evict() {
+        cache.remove(leastRecentlyUsed.key());
+        leastRecentlyUsed = leastRecentlyUsed.previous();
```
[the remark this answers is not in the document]
cenedhryn: Hm must have gotten lost.

utils/src/main/java/software/amazon/awssdk/utils/cache/lru/LruCache.java
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    public static <K, V> LruCache.Builder<K, V> builder(Function<K, V> supplier) {
+        return new Builder<>(supplier);
+    }
+
+    public static final class Builder<K, V> {
+
+        private final Function<K, V> supplier;
+        private int maxSize;
```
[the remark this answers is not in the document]
cenedhryn: Yes

utils/src/main/java/software/amazon/awssdk/utils/cache/lru/LruCache.java
cenedhryn: Evicted flag included

millems (approved):

PR #3477, opened 2022-10-11
Only enable CRT checksum for getObject and putObject

zoewangg:
## Motivation and Context
CRT does not support flexible checksum for non-streaming operation. After this change, we only pass checksum algorithm and isValidateChecksum for `getObject` and `putObject`

## Modifications
Only enable CRT checksum for getObject and putObject

## Testing
Integ tests passed

### DIFF
--- a/core/auth/src/main/java/software/amazon/awssdk/auth/signer/internal/util/SignerMethodResolver.java
+++ b/core/auth/src/main/java/software/amazon/awssdk/auth/signer/internal/util/SignerMethodResolver.java
@@ -26,6 +26,7 @@ import software.amazon.awssdk.auth.signer.internal.AbstractAwsS3V4Signer;
 import software.amazon.awssdk.core.CredentialType;
 import software.amazon.awssdk.core.interceptor.ExecutionAttributes;
 import software.amazon.awssdk.core.internal.signer.SigningMethod;
+import software.amazon.awssdk.core.signer.NoOpSigner;
 import software.amazon.awssdk.core.signer.Signer;
 
 @SdkInternalApi
@@ -56,7 +57,7 @@ public final class SignerMethodResolver {
                 signingMethod = SigningMethod.PROTOCOL_STREAMING_SIGNING_AUTH;
             } else if (isProtocolBasedUnsigned(signer, executionAttributes)) {
                 signingMethod = SigningMethod.PROTOCOL_BASED_UNSIGNED;
-            } else if (isAnonymous(credentials)) {
+            } else if (isAnonymous(credentials) || signer instanceof NoOpSigner) {
                 signingMethod = SigningMethod.UNSIGNED_PAYLOAD;
             } else {
                 signingMethod = SigningMethod.HEADER_BASED_AUTH;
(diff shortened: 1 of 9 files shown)


core/auth/src/main/java/software/amazon/awssdk/auth/signer/internal/util/SignerMethodResolver.java
```diff
@@ -56,7 +57,7 @@ public static SigningMethod resolveSigningMethodUsed(Signer signer, ExecutionAtt
                 signingMethod = SigningMethod.PROTOCOL_STREAMING_SIGNING_AUTH;
             } else if (isProtocolBasedUnsigned(signer, executionAttributes)) {
                 signingMethod = SigningMethod.PROTOCOL_BASED_UNSIGNED;
-            } else if (isAnonymous(credentials)) {
+            } else if (isAnonymous(credentials) || signer instanceof NoOpSigner) {
```
zoewangg, line 60: Per discussion with @joviegas, NoOpSigner should be treated as anonymous credentials

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/CrtChecksumUtils.java
```diff
@@ -0,0 +1,58 @@
...
+
+    /**
+     * Only validate response checksum if this is getObject operation AND it supports checksum validation AND if either of the
+     * following applies: 1. checksum validation is enabled at request level via request validation mode OR 2. checksum validation
+     * is enabled at client level
+     */
+    public static boolean validateResponseChecksum(HttpChecksum httpChecksum,
+                                                   S3MetaRequestOptions.MetaRequestType requestType,
+                                                   boolean checksumValidationEnabled) {
+        if (requestType != S3MetaRequestOptions.MetaRequestType.GET_OBJECT || httpChecksum == null) {
```
joviegas, line 52: Isn't `httpChecksum.requestValidationMode() != null` check redundant since GetObject will always have this trait set.

core/auth/src/main/java/software/amazon/awssdk/auth/signer/internal/util/SignerMethodResolver.java
joviegas, line 60: Are we planning to make NoOpSigner as final class ?

zoewangg: Yeah, I think we should do it. I'll create a PR against mainline

zoewangg: https://github.com/aws/aws-sdk-java-v2/pull/3481 is submitted to make it a final class

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/CrtChecksumUtils.java
joviegas: Yeap agree , even it has only one value the user can set it to null, so this check is required

joviegas (approved):

PR #3485, opened 2022-10-12
Fix TM integ tests set up and tear down steps

zoewangg:
## Motivation and Context
Fix TM integ tests set up and tear down steps to ensure resources are created and closed properly.

## Modifications
- Rename set up and tear down method names in `S3IntegrationTestBase` class so that they don't get hidden from child classes.
- Remove invocations of set up and tear down super methods from `S3IntegrationTestBase` in child classes because they are not needed

## Testing
Integ tests passed locally

### DIFF
--- a/services-custom/s3-transfer-manager/src/it/java/software/amazon/awssdk/transfer/s3/S3TransferManagerCopyIntegrationTest.java
+++ b/services-custom/s3-transfer-manager/src/it/java/software/amazon/awssdk/transfer/s3/S3TransferManagerCopyIntegrationTest.java
@@ -42,15 +42,12 @@ public class S3TransferManagerCopyIntegrationTest extends S3IntegrationTestBase 
 
     @BeforeAll
     public static void setUp() throws Exception {
-        S3IntegrationTestBase.setUp();
         createBucket(BUCKET);
     }
 
     @AfterAll
     public static void teardown() throws Exception {
-        tm.close();
         deleteBucketAndAllContents(BUCKET);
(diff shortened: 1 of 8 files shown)


services-custom/s3-transfer-manager/src/it/java/software/amazon/awssdk/transfer/s3/S3TransferManagerCopyIntegrationTest.java
```diff
@@ -42,15 +42,12 @@ public class S3TransferManagerCopyIntegrationTest extends S3IntegrationTestBase
 
     @BeforeAll
     public static void setUp() throws Exception {
-        S3IntegrationTestBase.setUp();
```
zoewangg, line 45: This is not needed per https://junit.org/junit5/docs/5.8.0/api/org.junit.jupiter.api/org/junit/jupiter/api/BeforeAll.html
>@BeforeAll methods are inherited from superclasses as long as they are not hidden or overridden. Furthermore, @BeforeAll methods from superclasses will be executed before @BeforeAll methods in subclasses.
Similarly, @BeforeAll methods declared in an interface are inherited as long as they are not hidden or overridden, and @BeforeAll methods from an interface will be executed before @BeforeAll methods in the class that implements the interface.

dagnir: nice good to know

dagnir (approved):

PR #3486, opened 2022-10-12
Updated aws-crt-client to use new 'utils' publishers for its request body logic.

millems:
Added two new publishers to 'utils':
1. StoringSubscriber - A subscriber that stores the events it receives in memory for retrieval.
2. ByteBufferStoringSubscriber - A subscriber that stores byte buffers it receives in memory for retrieval.

Also updated HttpRequestBodyStream to return content length. Not sure why this wasn't done before...


utils/src/main/java/software/amazon/awssdk/utils/async/StoringSubscriber.java
```diff
@@ -0,0 +1,207 @@
...
+
+    /**
+     * The active subscription. Set when {@link #onSubscribe(Subscription)} is invoked.
+     */
+    private Subscription subscription;
+
+    /**
+     * Whether the subscriber has been completed via {@link #onComplete()} or {@link #onError(Throwable)}.
+     */
+    private boolean done = false;
```
zoewangg, line 54: volatile?

utils/src/main/java/software/amazon/awssdk/utils/async/StoringSubscriber.java
```diff
@@ -0,0 +1,207 @@
...
+
+    /**
+     * The elements stored in this subscriber. The maximum size of this queue is approximately {@link #maxElements}.
+     */
+    private final Queue<Event<T>> elements;
+
+    /**
+     * The active subscription. Set when {@link #onSubscribe(Subscription)} is invoked.
+     */
+    private Subscription subscription;
```
zoewangg, line 49: volatile?

utils/src/main/java/software/amazon/awssdk/utils/async/StoringSubscriber.java
```diff
@@ -0,0 +1,207 @@
...
+    /**
+     * The maximum number of elements that can be stored in this subscriber. The number of elements in {@link #elements} may be
+     * slightly higher once {@link #onComplete()} and {@link #onError(Throwable)} events are added.
+     */
+    private final int maxElements;
+
+    /**
+     * The elements stored in this subscriber. The maximum size of this queue is approximately {@link #maxElements}.
+     */
+    private final Queue<Event<T>> elements;
```
zoewangg, line 44: nit, should we name it `events`?, same for `maxEvents`

utils/src/main/java/software/amazon/awssdk/utils/async/StoringSubscriber.java
```diff
@@ -0,0 +1,207 @@
...
+
+        this.subscription = subscription;
+        subscription.request(maxElements);
+    }
+
+    @Override
+    public void onNext(T t) {
+        Validate.notNull(t, "onNext(null) is not allowed.");
+
+        if (!done) {
```
zoewangg, line 101: Do we need done flag and done check since per [reactive streams spec](https://github.com/reactive-streams/reactive-streams-jvm), onNext should not be invoked after onComplete or onError?

> In response to a call to Publisher.subscribe(Subscriber) the possible invocation sequences for methods on the Subscriber are given by the following protocol:
onSubscribe onNext* (onError | onComplete)?

utils/src/main/java/software/amazon/awssdk/utils/async/ByteBufferStoringSubscriber.java
```diff
@@ -0,0 +1,194 @@
...
+
+    /**
+     * A delegate subscriber that we use to store the buffered bytes in the order they are received.
+     */
+    private final StoringSubscriber<ByteBuffer> storingSubscriber;
+
+    /**
+     * The active subscription. Set when {@link #onSubscribe(Subscription)} is invoked.
+     */
+    private Subscription subscription;
```
zoewangg, line 55: volatile?

utils/src/main/java/software/amazon/awssdk/utils/async/ByteBufferStoringSubscriber.java
```diff
@@ -0,0 +1,194 @@
...
+ *
+ * <p>Stored bytes can be read via {@link #transferTo(ByteBuffer)}.
+ */
+@SdkProtectedApi
+public class ByteBufferStoringSubscriber implements Subscriber<ByteBuffer> {
+    /**
+     * The minimum amount of data (in bytes) that should be buffered in memory at a time. The subscriber will request new byte
+     * buffers from upstream until the bytes received equals or exceeds this value.
+     */
+    private final int minimumBytesBuffered;
```
zoewangg, line 39: Should it be long?

utils/src/main/java/software/amazon/awssdk/utils/async/ByteBufferStoringSubscriber.java
```diff
@@ -0,0 +1,194 @@
...
+     * <p>If the data stored by this subscriber exceeds {@code out}'s {@code limit}, then {@code out} will be filled. If the data
+     * stored by this subscriber is less than {@code out}'s {@code limit}, then all stored data will be written to {@code out}.
+     *
+     * <p>If {@link #onError(Throwable)} was called on this subscriber, as much data as is available will be transferred into
+     * {@code out} before the provided exception is thrown (as a {@link RuntimeException}).
+     *
+     * <p>If {@link #onComplete()} was called on this subscriber, as much data as is available will be transferred into
+     * {@code out}, and this will return {@link TransferResult#END_OF_STREAM}.
+     *
+     * <p>Note: This method is NOT THREAD SAFE. It should not be called concurrently, and should not be called by
```
zoewangg, line 77: Should we make it thread safe? CRT `sendRequestBody` can be invoked multiple times by different threads (not concurrently)  for streaming operations, so we probably need synchronization everywhere anyway.

utils/src/test/java/software/amazon/awssdk/utils/async/ByteBufferStoringSubscriberTest.java
```diff
@@ -0,0 +1,308 @@
...
+                    }
+
+                    if (out.hasRemaining()) {
+                        carryOver.put(out);
+                    }
+                }
+                return null;
+            });
+
+            Thread.sleep(5_000);
```
zoewangg, line 286: Will this test be flaky?

millems: The sleep duration is just how long to spend trying to break it. It's not manual synchronization to make things work. 10 ms would work, just be less reliable of a test.

If there's a bug, this test will be flaky. If there's not a bug, the test should never fail.

millems: I ran it for a few minutes and never saw it fail.

utils/src/main/java/software/amazon/awssdk/utils/async/ByteBufferStoringSubscriber.java
millems: I was thinking 4gb should be plenty, but maybe you're right.

utils/src/main/java/software/amazon/awssdk/utils/async/ByteBufferStoringSubscriber.java
millems: I think I'm wrong in the docs. A call to `transfer` is safe when it happens-before another call to `transfer`. The calls to transfer can come from multiple threads.

In light of that, do you still think we should add synchronization to make sure that no one ever calls `transfer` concurrently? CRT wouldn't do that.

utils/src/main/java/software/amazon/awssdk/utils/async/ByteBufferStoringSubscriber.java
millems: Apparently it's not needed when called by well-behaved publishers: https://github.com/reactive-streams/reactive-streams-jvm/issues/53

Because `onSubscribe` happens-before `onNext`, then volatile isn't needed.

utils/src/main/java/software/amazon/awssdk/utils/async/StoringSubscriber.java
millems: Yeah, I think I'm with you and should drop it. I think I was being too cautious.

utils/src/main/java/software/amazon/awssdk/utils/async/ByteBufferStoringSubscriber.java
millems: Apparently it's not needed when called by well-behaved publishers: https://github.com/reactive-streams/reactive-streams-jvm/issues/53

Because onSubscribe happens-before onNext, then volatile isn't needed.

utils/src/main/java/software/amazon/awssdk/utils/async/ByteBufferStoringSubscriber.java
zoewangg: I was thinking about the case where thread 1 calls transferTo and while it's still in progress, another thread 2 invokes this method. It is happens-before relation but there may be an overlap. Is it possible?

utils/src/main/java/software/amazon/awssdk/utils/async/StoringSubscriber.java
millems: Removed variable

utils/src/main/java/software/amazon/awssdk/utils/async/StoringSubscriber.java
millems: See discussion below

utils/src/main/java/software/amazon/awssdk/utils/async/ByteBufferStoringSubscriber.java
millems: Yeah, that would certainly break, but CRT shouldn’t do it. The previous code makes that same assumption. We could add a synchronized block to the CRT code if you think they might do it? Not sure it’s worth including in the utils class.

utils/src/main/java/software/amazon/awssdk/utils/async/ByteBufferStoringSubscriber.java
zoewangg: Gotcha, hope our publishers are well-behaved 🤔

zoewangg (approved):

millems:
I'd like to bake these changes in our canaries for a little while, so I'm going to treat this branch as a staging branch for the whole CRT refactor. Closing this PR without merging to master. Will merge later once we have really assessed the changes.

PR #3490, opened 2022-10-14
Updated aws-crt-client to use new 'utils' publishers for its response body logic.

millems:
Added a new publisher to 'utils': SimplePublisher - A publisher that exposes simple `write`, `complete` and `error` methods, and avoids the caller needing to be concerned about backpressure or following the reactive streams specification.


utils/src/main/java/software/amazon/awssdk/utils/async/SimplePublisher.java
```diff
@@ -0,0 +1,497 @@
...
+            log.error(() -> "Failed while processing a failure. This could result in stuck futures.", t);
+        }
+    }
+
+    /**
+     * Ensure that {@link #rejectException} is null. If it is not, throw the exception.
+     */
+    private void checkRejectException() {
+        if (rejectException.get() != null) {
+            throw rejectException.get().get();
```
zoewangg, line 361: Instead of relying on throw and try-catch to complete the future exceptionally, would it be better to return an `Optional<RuntimeException>` and make the caller handle it more explicitly? Because if future developers forget to catch exceptions when calling this method, it could cause the future to be stuck.

http-clients/aws-crt-client/src/main/java/software/amazon/awssdk/http/crt/internal/response/CrtResponseAdapter.java
```diff
@@ -39,97 +40,110 @@ public final class CrtResponseAdapter implements HttpStreamResponseHandler {
...
+        });
+    }
+
+    private void onFailedResponseComplete(HttpStream stream, HttpException error) {
+        log.error(() -> "HTTP response encountered an error.", error);
+        responsePublisher.error(error);
+        die(stream, error);
+    }
+
+    private void die(HttpStream stream, Throwable error) {
```
zoewangg, line 134: nit: maybe `handleStreamError` or `handleError`?

utils/src/main/java/software/amazon/awssdk/utils/async/SimplePublisher.java
```diff
@@ -0,0 +1,497 @@
...
+ *
+ * Each of these operations returns a {@link CompletableFuture} for indicating when the message has been successfully sent.
+ *
+ * <p>Callers are expected to invoke a series of {@link #write(Object)}s followed by a single {@link #complete()} or
+ * {@link #error(Throwable)}. See the documentation on each operation for more details.
+ *
+ * <p>This publisher will store an unbounded number of messages. It is recommended that callers limit the number of in-flight
+ * {@link #write(Object)} operations in order to bound the amount of memory used by this publisher.
+ */
+@SdkPublicApi
```
zoewangg, line 60: Should it be protected API?

utils/src/main/java/software/amazon/awssdk/utils/async/SimplePublisher.java
```diff
@@ -0,0 +1,497 @@
...
+     */
+    @Override
+    public void subscribe(Subscriber<? super T> s) {
+        if (subscriber != null) {
+            s.onSubscribe(new NoOpSubscription());
+            s.onError(new IllegalStateException("Only one subscription may be active at a time."));
+        }
+        this.subscriber = s;
+        s.onSubscribe(new SubscriptionImpl());
+        processEventQueue();
```
zoewangg, line 209: Do we need to call `processEventQueue` here?

utils/src/main/java/software/amazon/awssdk/utils/async/SimplePublisher.java
```diff
@@ -0,0 +1,497 @@
...
+        @Override
+        public void request(long n) {
+            log.trace(() -> "Received request() with " + n);
+            if (n <= 0) {
+                // Create exception here instead of in supplier to preserve a more-useful stack trace.
+                IllegalArgumentException failure = new IllegalArgumentException("A downstream publisher requested an invalid "
+                                                                                + "amount of data: " + n);
+                rejectException.compareAndSet(null, () -> failure);
+                eventQueue.add(new OnErrorQueueEntry<>(failure));
+                entryTypesToFail.add(ON_NEXT);
```
zoewangg, line 388: Should it be all types of events?

utils/src/main/java/software/amazon/awssdk/utils/async/SimplePublisher.java
```diff
@@ -0,0 +1,497 @@
...
+     *
+     * <p>Here we try our best to fail all of the entries in the queue, so that no callers have "stuck" futures.
+     */
+    private void panicAndDie(Throwable cause) {
+        try {
+            // Create exception here instead of in supplier to preserve a more-useful stack trace.
+            RuntimeException failure = new IllegalStateException("Encountered fatal error in publisher", cause);
+            rejectException.compareAndSet(null, () -> failure);
+            entryTypesToFail.addAll(asList(QueueEntry.Type.values()));
+            subscriber.onError(failure);
```
zoewangg, line 341: This seems to be inconsistent with `#error()` and the case where request is <= in the subcription implementation where onError is not invoked immediately. Is this intentional?

utils/src/main/java/software/amazon/awssdk/utils/async/SimplePublisher.java
```diff
@@ -0,0 +1,497 @@
...
+     * request message in memory. Callers are recommended to limit the number of writes in progress at a time to bound the
+     * amount of memory used by this publisher.
+     *
+     * <p>The returned future will be completed exceptionally if the downstream subscriber cancels the subscription, or
+     * if the write call was performed after a {@link #complete()} or {@link #error(Throwable)} call.
+     *
+     * @param value The message to send. Must not be null.
+     * @return A future that is completed when the message is sent to the subscriber.
+     */
+    public CompletableFuture<Void> write(T value) {
```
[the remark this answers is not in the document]
millems: Yeah, I was thinking writing to the publisher, but I'm okay with `send()`.

utils/src/main/java/software/amazon/awssdk/utils/async/SimplePublisher.java
```diff
@@ -0,0 +1,497 @@
...
+        log.trace(() -> "Received write() with " + value);
```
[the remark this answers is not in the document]
millems: We can't get a size since it's untyped within this publisher. We could log the type, but that's not as useful. Since it's at the trace level, and we already log sensitive data elsewhere at the debug level, shouldn't it be safe?

utils/src/main/java/software/amazon/awssdk/utils/async/SimplePublisher.java
```diff
@@ -0,0 +1,497 @@
...
+            if (n <= 0) {
+                // Create exception here instead of in supplier to preserve a more-useful stack trace.
+                IllegalArgumentException failure = new IllegalArgumentException("A downstream publisher requested an invalid "
+                                                                                + "amount of data: " + n);
+                rejectException.compareAndSet(null, () -> failure);
+                eventQueue.add(new OnErrorQueueEntry<>(failure));
+                entryTypesToFail.add(ON_NEXT);
+                processEventQueue();
+            } else {
+                long newDemand = outstandingDemand.addAndGet(n);
```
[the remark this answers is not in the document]
millems: > A Subscription MUST support an unbounded number of calls to request and MUST support a demand up to 2^63-1 (java.lang.Long.MAX_VALUE). 

I read this "up to" as "but not greater than".

> A demand equal or greater than 2^63-1 (java.lang.Long.MAX_VALUE) MAY be considered by the Publisher as “effectively unbounded”.

Since this said "may" I was being lazy, but you're probably late - I shouldn't be lazy.

utils/src/main/java/software/amazon/awssdk/utils/async/SimplePublisher.java
```diff
@@ -0,0 +1,497 @@
...
+
+        @Override
+        public void cancel() {
+            log.trace(() -> "Received cancel()");
+
+            // Create exception here instead of in supplier to preserve a more-useful stack trace.
+            IllegalStateException failure = new IllegalStateException("A downstream publisher has cancelled the subscription.");
+            rejectException.compareAndSet(null, () -> failure);
+            eventQueue.add(new CancelQueueEntry<>());
+            entryTypesToFail.addAll(asList(ON_NEXT, ON_COMPLETE, ON_ERROR));
```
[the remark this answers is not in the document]
millems: > I wonder if we should just delete those events by removing the data from the queue to fail faster in case there are a lot of events in the queue.

And leave the futures incomplete? 

> Is there a use case for handling failure for individual data event?

Not that I can think of, but there is a use-case for handling success of each individual event, and it feels unnatural (to me) to have item-level success and publisher-level failure.

> A more general question, do we need entryTypesToFail? Would a isDone flag suffice?

We could combine some things for now, like ON_COMPLETE and ON_ERROR (we never fail one without failing the other), but we need to fail ON_NEXT without failing ON_ERROR, for example.

I originally had two flags - one for failing ON_NEXT and one for failing ON_COMPLETE and ON_ERROR, but there was more code and volatile reads under that design, so I went with this one. Is it confusing?

utils/src/main/java/software/amazon/awssdk/utils/async/SimplePublisher.java
millems: Sure

utils/src/main/java/software/amazon/awssdk/utils/async/SimplePublisher.java
millems: Yes. We support writing to this publisher before anyone subscribes, and we want any complete() or error() calls to be sent as soon as we get a subscriber.

utils/src/main/java/software/amazon/awssdk/utils/async/SimplePublisher.java
```diff
@@ -0,0 +1,497 @@
...
+                        log.trace(() -> "Calling onNext() with " + onNextEntry.value);
+                        subscriber.onNext(onNextEntry.value);
+                        long newDemand = outstandingDemand.decrementAndGet();
+                        log.trace(() -> "Decreased demand to " + newDemand);
+                        break;
+                    case ON_COMPLETE:
+                        entryTypesToFail.addAll(asList(ON_NEXT, ON_COMPLETE, ON_ERROR));
+                        log.trace(() -> "Calling onComplete()");
+                        subscriber.onComplete();
+                        outstandingDemand.set(0);
```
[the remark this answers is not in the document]
millems: No, it's not needed, because we have "skip ON_NEXT" set. Do you think it would be better to remove it?

utils/src/main/java/software/amazon/awssdk/utils/async/SimplePublisher.java
millems: Sorry, I'm not sure what you mean. 

We call `onError` here directly instead of doing it via the queue, because `panicAndDie` is used when the queue seems to be 'broken' for some unknown reason (`doProcessQueue` threw an exception, which should never happen!). It's a last-ditch attempt to not leave any incomplete futures laying around. That's why it disables all of the entry types for processing and drains the future queue itself instead of relying on the default queue processing logic.

utils/src/main/java/software/amazon/awssdk/utils/async/SimplePublisher.java
millems: Hmm, let me play with it a bit. I'm not sure how common the scenario "developers think to call this, but don't think to catch the exception it throws" is, but it's certainly possible. I was optimizing for readability in the functions that call `checkRejectException`, similar to how `Validate.isTrue` throws an exception instead of returning an `Optional<RuntimeException>`. I wanted to minimize the ceremony in the caller.

utils/src/main/java/software/amazon/awssdk/utils/async/SimplePublisher.java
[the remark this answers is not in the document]
zoewangg: > And leave the futures incomplete?

If we go with this approach, we'd probably change it to void and not return a future and that's why I was curious if there was a use-case for handling the result of each event. Reading it again, seems like we do(increasing window size). I guess alternatively, we could expose callback for onNext call, which is not necessarily better than this.

utils/src/main/java/software/amazon/awssdk/utils/async/SimplePublisher.java
millems: No, we still need to process `onError` events, but we probably should fail onComplete here, though. Good catch!

http-clients/aws-crt-client/src/main/java/software/amazon/awssdk/http/crt/internal/response/CrtResponseAdapter.java
```diff
@@ -39,97 +40,110 @@ public final class CrtResponseAdapter implements HttpStreamResponseHandler {
...
     }
 
     @Override
     public int onResponseBody(HttpStream stream, byte[] bodyBytesIn) {
-        initRespBodyPublisherIfNeeded(stream);
+        CompletableFuture<Void> writeFuture = responsePublisher.write(ByteBuffer.wrap(bodyBytesIn));
+
+        if (writeFuture.isDone() && !writeFuture.isCompletedExceptionally()) {
+            // Optimization: If write succeeded immediately, return non-zero to avoid the extra call back into the CRT.
+            return bodyBytesIn.length;
```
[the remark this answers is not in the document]
millems: Yeah, I figured we could just use both mechanisms, so that when the consumer has enough demand to accept the request immediately we can avoid the extra JNI overhead of calling back into the CRT. Do you think that's confusing?

utils/src/main/java/software/amazon/awssdk/utils/async/SimplePublisher.java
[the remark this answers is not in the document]
zoewangg: Ah yeah, I just replaced T with ByteBuffer in my mind automatically somehow. Should we not include value at all then? My concern is that if it's a 8MB byte buffer, it's going to be a lot of data to be printed out.

> we already log sensitive data elsewhere at the debug level

IIRC, the logs that could contain potential sensitive data are wirelogs generated from HTTP clients, which is out of our control. Do we log those data elsewhere in the SDK?

http-clients/aws-crt-client/src/main/java/software/amazon/awssdk/http/crt/internal/response/CrtResponseAdapter.java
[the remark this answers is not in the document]
zoewangg: Ah I see what you mean. That's smart.

utils/src/main/java/software/amazon/awssdk/utils/async/SimplePublisher.java
zoewangg: Gotcha, I was wondering if we should do the same for error cases to fail faster instead of going through the queue and skipping other events

utils/src/main/java/software/amazon/awssdk/utils/async/SimplePublisher.java
```diff
@@ -0,0 +1,497 @@
...
+     * <p>The returned future will be completed exceptionally if the downstream subscriber cancels the subscription, or
+     * if the write call was performed after a {@link #complete()} or {@link #error(Throwable)} call.
+     *
+     * @param value The message to send. Must not be null.
+     * @return A future that is completed when the message is sent to the subscriber.
+     */
+    public CompletableFuture<Void> write(T value) {
+        log.trace(() -> "Received write() with " + value);
+
+        OnNextQueueEntry<T> entry = new OnNextQueueEntry<>(value);
```
joviegas, line 123: Query : Just curious , since we are creating New instances then previous implementation does this qualifies to a quick performance/memory test ?

utils/src/main/java/software/amazon/awssdk/utils/async/SimplePublisher.java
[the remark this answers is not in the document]
millems: Logging a ByteBuffer doesn't log the contents, just the locations of the position, limit and capacity. The only chance at logging sensitive data is if we reused this elsewhere for things that do have a sensitive toString.

utils/src/main/java/software/amazon/awssdk/utils/async/SimplePublisher.java
millems: The only reason we can do this here safely is because we're holding the "processing the queue" flag. It wouldn't be safe to call `onError` directly within `error` (we might be in the middle of calling onNext, onComplete, etc.).

http-clients/aws-crt-client/src/main/java/software/amazon/awssdk/http/crt/internal/response/CrtResponseAdapter.java
```diff
@@ -39,97 +40,110 @@ public final class CrtResponseAdapter implements HttpStreamResponseHandler {
...
-        }
+    public static HttpStreamResponseHandler toCrtResponseHandler(HttpClientConnection crtConn,
+                                                                 CompletableFuture<Void> requestFuture,
+                                                                 SdkAsyncHttpResponseHandler responseHandler) {
+        return new CrtResponseAdapter(crtConn, requestFuture, responseHandler);
     }
 
     @Override
-    public void onResponseHeaders(HttpStream stream, int responseStatusCode, int blockType, HttpHeader[] nextHeaders) {
-        initRespBodyPublisherIfNeeded(stream);
```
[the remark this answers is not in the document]
millems: I didn't try to understand why it was written the way it was.

Thinking about it now:
* Theoretically, we could lazily create the publisher.
* That may make the failure path faster, but I expect that it would slow down the successful path (now all reads of the publisher need to be volatile). 
* It adds complexity.
* It may not even make the failure path faster, because initializing a lazy variable without guarantees about the calling thread is two volatile reads, a lock acquisition, and a volatile write. I'd have to test to be sure, but I expect that's slower than any initialization logic in the publisher's constructor.

I don't think it's worth the amount of time spent testing to see if it's faster, since it's more complex and I doubt that it's any significant speed improvement. The next time we do an optimization path, we can optimize it at that time if it shows up as something significant on the profiler.

utils/src/main/java/software/amazon/awssdk/utils/async/SimplePublisher.java
millems: Creating an instance in the JVM is pretty quick (mostly equivalent to work done in the constructor), and short-lived instances are very efficiently GC'd.

Creating this instance is slower than not creating one, but it's a bit faster than copying the bytes out of the buffer will be, a fair bit faster than the volatile reads/writes we're doing, and considerably faster than the network activity that all of this relies on.

That said, this is all supposition. The only way to know for sure would be to do profiling, and the output might be surprising. I intend to bake these changes in the canaries for a little while to make sure that it's not slower in any noticeable way.

I expect that this implementation might be slightly slower than the previous implementation, but I think it will make us more productive because it puts all of the confusing publisher stuff into a separate class that we (hopefully) don't have to think about again.

utils/src/main/java/software/amazon/awssdk/utils/async/SimplePublisher.java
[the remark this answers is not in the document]
zoewangg: Ah I see, then I think it should be fine

utils/src/main/java/software/amazon/awssdk/utils/async/SimplePublisher.java
zoewangg: I see. In the S3 CRT publisher adapter implementation, we just clear the queue, add ON_ERROR event to the queue and then trigger processQueue so that the error event is executed almost immediately. It doesn't apply here since we are skipping events and not deleting them. This can be marked as resolved.

utils/src/main/java/software/amazon/awssdk/utils/async/SimplePublisher.java
millems: I couldn't make it into something that I liked. I renamed the method to follow the "validate*" convention so that the caller can hopefully realize that it throws an exception.

zoewangg (approved):
Can we run integ tests and stability tests if we haven't already?

millems:
Integration and stability tests are passing. I'll be baking this in the canaries for a little while to build additional confidence before merging to main.

millems:
Looks good after baking in the canaries for a few days. Slight heap memory usage increase, but no latency impact or increase in errors.

PR #3492, opened 2022-10-17
Using the character replace() method for better performance.

RaRe2604:
Using the character replace() method instead of CharSequence replace() for better performance.

## Description
Using the character replace() instead of CharSequence replace() for better performance which avoids creating new strings of both target and replacement strings.

## Types of changes

## Checklist

## License


RaRe2604 (commented):
Performance Fix - char replace() instead charsequence replace()

debora-ito:
Hi @RaRe2604 thank you for the PR, and apologies for the delay in reply.

Can you please tell us a little more about the motivation for the change? Perhaps share some references too.

We run local tests with the change and results showed no significant performance gain in those specific cases - region, class and service names. If you could share some data (like benchmarks) showing the performance improvements we will take another look.

PR #3497, opened 2022-10-17
Enable JDK9+ code and builds to produce Java8 compatible jar

DavidOgunsAWS:
## Motivation and Context

* Newly developed Java9 code written with module-info.java/.class should still produce jars that run compatibly on JRE8 as this SDK is still required to support it.
* Developers should need to explicitly attempt to do this. Github workflows using JDK9+ should run this too.
* The publishing maven profile will want to do this by default, however the profile must be added first as calling maven with `-P <profile>` with a profile that does not exist will cause failure.

## Modifications

* Added Maven profile that triggers an explicit java-9-compile with appropriate options and java-8-compile excluding module-info.java. Java9+ class files (excluding module-info.class) are over-written with Java8 class files so the resulting jar runs on JRE8 successfully. 
* New profile has enforcement that JDK version is at least 9
* module-path-tests adjusted to suppresses both java-9-compile and java-8-compile execution, preferring how it has already been configured.

## Testing

1. Added new test sub-module under core/<module> set up with module-info.java and main method
1. Verify build succeeds when JAVA_HOME and PATH points to JDK8 and JDK9+ (tested with 17)
1. Execute main method through JRE8 classpath successfully
1. Execute main method through JRE8 with module-path successfully
1. Downstream build of in-progress Java9 module in separate fork/branch.

## Types of changes

## Checklist

## License


core/sdk-core/pom.xml
```diff
@@ -248,6 +248,14 @@
                     </archive>
                 </configuration>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <source>9</source>
+                    <target>9</target>
```
millems, line 256: Wouldn't this cause issues running this jar on JDK 8?

pom.xml
```diff
@@ -693,6 +670,145 @@
...
 
+
+        <profile>
+            <id>java9plus</id>
+            <activation>
+                <jdk>[9,)</jdk>
+            </activation>
+            <build>
+                <plugins>
+                    <!-- Don't compile module-info.java, see java 9 profile -->
```
millems, line 681: This is the Java 9 profile, right?

Edit: Oh, maybe the 9 execution? Is this comment needed?

pom.xml
```diff
@@ -693,6 +670,145 @@
...
+                                </configuration>
+                            </execution>
+                            <execution>
+                                <id>java-9-compile</id>
+                                <goals>
+                                    <goal>compile</goal>
+                                </goals>
+                                <configuration>
+                                    <release>9</release>
+                                    <!-- no excludes: compile everything to ensure module-info contains right entries -->
```
millems, line 708: What does this mean? There are excludes below.

pom.xml
```diff
@@ -693,6 +670,145 @@
...
+                                        <exclude>software/amazon/awssdk/services/kinesis/KinesisAsyncClient.java</exclude>
+                                        <exclude>software/amazon/awssdk/services/kinesis/DefaultKinesisAsyncClient.java</exclude>
+                                        <exclude>software/amazon/awssdk/services/kinesis/DefaultKinesisBaseClientBuilder.java</exclude>
+                                        <exclude>software/amazon/awssdk/services/sms/model/InternalError.java</exclude>
+                                        <exclude>software/amazon/awssdk/services/kinesis/transform/SubscribeToShardResponseUnmarshaller.java</exclude>
+                                    </excludes>
+                                </configuration>
+                            </execution>
+                            <execution>
+                                <id>java-8-compile</id>
```
millems, line 725: What do these 2 different executions do? It looks like we compile twice?

DavidOgunsAWS: Yes, that is actually the intent with this change. I added a bit of detail to the description of changes, but Java9+ compilation happens first, then relevant class files are overwritten with java8 class files aside from module-info.class. The resulting JAR works for both JRE8 classpath invocation, and JRE9+ module-path invocation.

core/sdk-core/pom.xml
DavidOgunsAWS: This is configuration that needed to be removed. It's a good catch because it probably does violate JRE8 compatibility inside the sdk-core module. I tested invocations with a new dummy module. I'm going to delete this chunk as no specific compiler plugin configuration should be needed here.

pom.xml
DavidOgunsAWS: Removing comment. I think the configuration and naming is clear.

pom.xml
DavidOgunsAWS: Removing.

pom.xml
millems: Interesting... how much does this slow down a full `mvn clean install`?

DavidOgunsAWS: Baseline on m6a.2xlarge was `1h7m`, and after dual compilation it took `1h21m`.

debora-ito:
Closing this for now.
Will reopen if necessary.

PR #3523, opened 2022-10-27
Implement Asynchronous EC2 Metadata Client

L-Applin:
Implementation of the Asynchronous Ec2 Metadata Client.

## Motivation and Context
This is part of the required features before release of the EC2Metadata Service Client

## Modifications
Add interface `Ec2MetadataAsyncClient` and one default implementation.
Slight refactoring of the sync client to keep both client aligned.

## Testing
Unit tests added for the implementation.

## Types of changes

## Checklist

## License


core/imds/src/test/java/software/amazon/awssdk/imds/Ec2MetadataClientTest.java
```diff
@@ -0,0 +1,391 @@
...
+        SdkHttpClient client = UrlConnectionHttpClient.builder().socketTimeout(Duration.ofMillis(500)).build();
+        Ec2MetadataClient ec2MetadataRequest = Ec2MetadataClient.builder()
+                                                                .httpClient(client)
+                                                                .endpoint(URI.create("http://localhost:" + mockMetadataEndpoint.port()))
+                                                                .build();
+        assertThatThrownBy(() -> ec2MetadataRequest.get("/latest/meta-data/ami-id"))
+            .hasMessageContaining("Exceeded maximum number of retries.")
+            .isInstanceOf(SdkClientException.class);
+
+        WireMock.verify(4, putRequestedFor(urlPathEqualTo(TOKEN_RESOURCE_PATH))
```
cenedhryn, line 170: Can we extract `4` to a static final instance variable to make it clear it refers to the default allowed attempts?

core/imds/src/test/java/software/amazon/awssdk/imds/Ec2MetadataClientTest.java
```diff
@@ -0,0 +1,391 @@
...
+        assertThat(metadataResponse.asString()).isEqualTo("{}");
+
+        WireMock.verify(3, putRequestedFor(urlPathEqualTo(TOKEN_RESOURCE_PATH))
+            .withHeader(EC2_METADATA_TOKEN_TTL_HEADER, equalTo("21600")));
+        WireMock.verify(exactly(1), getRequestedFor(urlPathEqualTo(AMI_ID_RESOURCE))
+            .withHeader(TOKEN_HEADER, equalTo("some-token")));
+    }
+
+    @Test
+    public void getToken_failsTwiceWithIOExceptionThenOK_shouldSucceedOnThirdAttempt() {
```
cenedhryn, line 303: Do all permutations of "fail a couple of times and then succeed" test cases test different behaviors? Could we pare them down a little bit?

core/imds/src/test/java/software/amazon/awssdk/imds/EndpointProviderTest.java
```diff
@@ -0,0 +1,136 @@
...
+import org.junit.jupiter.params.provider.MethodSource;
+import software.amazon.awssdk.core.SdkSystemSetting;
+import software.amazon.awssdk.imds.internal.Ec2MetadataEndpointProvider;
+import software.amazon.awssdk.profiles.ProfileFileSystemSetting;
+import software.amazon.awssdk.testutils.EnvironmentVariableHelper;
+
+/**
+ * Test Class to test the endpoint resolution functionality.
+ */
+class EndpointProviderTest {
```
cenedhryn, line 39: Should we also include a test where `profileFile` and `profileName` is set on the builder and not just uses the default values?

core/imds/src/test/java/software/amazon/awssdk/imds/internal/DefaultEc2MetadataClientAsyncClientTest.java
```diff
@@ -0,0 +1,334 @@
...
+import org.mockito.junit.MockitoJUnitRunner;
+import software.amazon.awssdk.core.SdkSystemSetting;
+import software.amazon.awssdk.core.retry.backoff.FixedDelayBackoffStrategy;
+import software.amazon.awssdk.http.crt.AwsCrtAsyncHttpClient;
+import software.amazon.awssdk.imds.Ec2MetadataAsyncClient;
+import software.amazon.awssdk.imds.Ec2MetadataRetryPolicy;
+import software.amazon.awssdk.imds.MetadataResponse;
+
+@RunWith(MockitoJUnitRunner.class)
+public class DefaultEc2MetadataClientAsyncClientTest {
```
cenedhryn, line 59: The names of the test classes `DefaultEc2MetadataAsyncClientTest`  and `DefaultEc2MetadataClientAsyncClientTest` confuse me. What is the difference between them?

core/imds/src/main/java/software/amazon/awssdk/imds/internal/DefaultEc2MetadataAsyncClient.java
```diff
@@ -0,0 +1,307 @@
...
+                                  .request(baseTokenRequest)
+                                  .retriesAttempted(newAttempt)
+                                  .exception(SdkClientException.create(error.getMessage(), error))
+                                  .build();
+            return scheduledRetryAttempt(path, newContext);
+        }).thenCompose(Function.identity()); // only java 12 has .exceptionallyCompose()
+    }
+
+    private CompletableFuture<String> createAsyncRequest(SdkHttpFullRequest baseRequest) {
+        StringAsyncResponseHandler responseHandler = new StringAsyncResponseHandler();
```
cenedhryn, line 124: do we need to create a new instance of the response handler for each request attempt?

core/imds/src/main/java/software/amazon/awssdk/imds/internal/DefaultEc2MetadataAsyncClient.java
```diff
@@ -0,0 +1,307 @@
...
+                RetryPolicyContext.builder()
+                                  .request(baseTokenRequest)
+                                  .retriesAttempted(newAttempt)
+                                  .exception(SdkClientException.create(error.getMessage(), error))
+                                  .build();
+            return scheduledRetryAttempt(path, newContext);
+        }).thenCompose(Function.identity()); // only java 12 has .exceptionallyCompose()
+    }
+
+    private CompletableFuture<String> createAsyncRequest(SdkHttpFullRequest baseRequest) {
```
cenedhryn, line 123: This method has a lot of responsibilities; it both creates an async request, makes the http call and transforms the result. If it makes most sense to combine these actions I suggest renaming the method to something like `makeAsyncCall` or similar (you can probably come up with a better name).

core/imds/src/main/java/software/amazon/awssdk/imds/internal/DefaultEc2MetadataAsyncClient.java
```diff
@@ -0,0 +1,307 @@
...
+    }
+
+    private CompletableFuture<MetadataResponse> get(String path, RetryPolicyContext retryPolicyContext) {
+        SdkHttpFullRequest baseTokenRequest = requestMarshaller.createTokenRequest(tokenTtl);
+
+        CompletableFuture<MetadataResponse> result = createAsyncRequest(baseTokenRequest)
+            .thenCompose(token -> {
+                SdkHttpFullRequest baseMetadataRequest = requestMarshaller.createDataRequest(path, token, tokenTtl);
+                return createAsyncRequest(baseMetadataRequest);
+            }).thenApply(MetadataResponse::create);
```
cenedhryn, line 105: Is it possible to rewrite this part of the code to make it a little easier to follow the request/response flow? I'm confused by the name `createAsyncRequest` (see other comment) and the token call is combined with the actual data call and it took me a while to realize what was happening. 

Maybe just splitting token and data futures like 
~~~
CompletableFuture<String> acquiredToken = callAsync(baseTokenRequest)
CompletableFuture<MetadataResponse> result =acquiredToken
            .thenCompose(token -> {
                SdkHttpFullRequest baseMetadataRequest = requestMarshaller.createDataRequest(path, token, tokenTtl);
                return createAsyncRequest(baseMetadataRequest);
            }).thenApply(MetadataResponse::create);
~~~

core/imds/src/main/java/software/amazon/awssdk/imds/internal/DefaultEc2MetadataAsyncClient.java
```diff
@@ -0,0 +1,307 @@
...
+            int newAttempt = retryPolicyContext.retriesAttempted() + 1;
+            log.debug(() -> "Retrying request: Attempt " + newAttempt);
+            RetryPolicyContext newContext =
+                RetryPolicyContext.builder()
+                                  .request(baseTokenRequest)
+                                  .retriesAttempted(newAttempt)
+                                  .exception(SdkClientException.create(error.getMessage(), error))
+                                  .build();
+            return scheduledRetryAttempt(path, newContext);
+        }).thenCompose(Function.identity()); // only java 12 has .exceptionallyCompose()
```
cenedhryn, line 120: is this necessary `thenCompose(Function.identity())`? Wondering if there is another way to return the future.
you can look into thenApply, CompletableFutureUtils (especially tm branch).

core/imds/src/main/java/software/amazon/awssdk/imds/internal/DefaultEc2MetadataAsyncClient.java
```diff
@@ -0,0 +1,307 @@
...
+                                                                 .build();
+        httpClient.execute(metadataRequest);
+        return responseHandler.getResult();
+    }
+
+    private CompletableFuture<MetadataResponse> scheduledRetryAttempt(String path, RetryPolicyContext retryPolicyContext) {
+        Duration retryDelay = retryPolicy.backoffStrategy().computeDelayBeforeNextRetry(retryPolicyContext);
+        Executor retryExecutor = retryAttempt -> asyncRetryScheduler.schedule(retryAttempt, retryDelay.toMillis(),
+                                                                              TimeUnit.MILLISECONDS);
+        return CompletableFuture.supplyAsync(() -> get(path, retryPolicyContext), retryExecutor)
```
cenedhryn, line 139: (optional) you could supply the function to `scheduleRetryAttempt` as well 
`return scheduledRetryAttempt(() -> get(path, newContext), newContext)` or even 
`return scheduledRetryAttempt(() -> get(path, newContext), calculateRetryDelay(newContext))`

core/imds/src/main/java/software/amazon/awssdk/imds/internal/DefaultEc2MetadataAsyncClient.java
```diff
@@ -0,0 +1,307 @@
...
+
+    private static final class StringResponseHandler implements HttpResponseHandler<String> {
+        @Override
+        public String handle(SdkHttpFullResponse response, ExecutionAttributes executionAttributes) throws Exception {
+            return IoUtils.toUtf8String(response.content().orElseThrow(() ->
+                SdkClientException.create("Unexpected error: empty response content")));
+        }
+    }
+
+    private static final class StringAsyncResponseHandler implements TransformingAsyncResponseHandler<String> {
```
cenedhryn, line 179: Nice!

core/imds/src/main/java/software/amazon/awssdk/imds/internal/DefaultEc2MetadataAsyncClient.java
```diff
@@ -0,0 +1,310 @@
...
+            this.tokenTtl = tokenTtl;
+        }
+
+        public void setEndpointMode(EndpointMode endpointMode) {
+            this.endpointMode = endpointMode;
+        }
+
+        public void setHttpClient(SdkAsyncHttpClient httpClient) {
+            this.httpClient = httpClient;
+        }
```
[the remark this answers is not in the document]
L-Applin: good catch, we dont. I will remove them. That is why you dont copy-paste code 😅 leftover from the sync client copy-paste

core/imds/src/main/java/software/amazon/awssdk/imds/internal/DefaultEc2MetadataClient.java
```diff
@@ -46,12 +49,9 @@
 @SdkInternalApi
 @Immutable
 @ThreadSafe
-public final class DefaultEc2Metadata implements Ec2Metadata {
-
-    private static final Logger log = Logger.loggerFor(DefaultEc2Metadata.class);
+public final class DefaultEc2MetadataClient implements Ec2MetadataClient, SdkClient {
```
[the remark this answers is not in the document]
L-Applin: Shouldn't it?

core/imds/src/main/java/software/amazon/awssdk/imds/internal/DefaultEc2MetadataAsyncClient.java
```diff
@@ -0,0 +1,310 @@
...
+    private final URI endpoint;
+    private final RequestMarshaller requestMarshaller;
+    private final Duration tokenTtl;
+    private final ScheduledExecutorService asyncRetryScheduler;
+
+    private DefaultEc2MetadataAsyncClient(Ec2MetadataAsyncBuilder builder) {
+        this.retryPolicy = builder.retryPolicy != null ? builder.retryPolicy
+                                                       : Ec2MetadataRetryPolicy.builder().build();
+        this.httpClient = builder.httpClient != null ? builder.httpClient
+                                                     : AwsCrtAsyncHttpClient.create();
```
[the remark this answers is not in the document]
L-Applin: That is actually a very good point. It was not discussed, no, which async http should be used by default. Whichever one we used, it will have to be included in the classpath of the user, indeed.

We make sure users have the CRT on their classpath automatically by adding the dependency to the `aws-crt-client` in the `pom.xml` of the IMDS module. Right now, users that want to use another http-client would have to include that client in their `pom.xml` (and optionally exclude the `aws-crt-client` dependency)

core/imds/src/main/java/software/amazon/awssdk/imds/internal/RequestMarshaller.java
```diff
@@ -50,28 +49,26 @@ public RequestMarshaller(URI basePath) {
         this.tokenPath = URI.create(basePath + TOKEN_RESOURCE_PATH);
     }
 
-    public HttpExecuteRequest createTokenRequest(Duration tokenTtl) {
-        SdkHttpRequest sdkHttpRequest = defaulttHttpBuilder()
+    public SdkHttpFullRequest createTokenRequest(Duration tokenTtl) {
```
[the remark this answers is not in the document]
L-Applin: To conform to the `AsyncExecuteRequest` signature which requires a `SdkHttpRequest`. `HttpExecuteRequest` does not implement that interface.

core/imds/src/main/java/software/amazon/awssdk/imds/internal/DefaultEc2MetadataClient.java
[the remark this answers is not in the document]
cenedhryn: We've viewed it as a kind of utility that isn't a regular SDK generated client. Since this is the first of its kind it makes sense to discuss this further and the pros/cons of each approach. We can add this topic to a surface API review list and not change it in this PR.

core/imds/src/main/java/software/amazon/awssdk/imds/internal/DefaultEc2MetadataAsyncClient.java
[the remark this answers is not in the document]
cenedhryn: lol

core/imds/src/main/java/software/amazon/awssdk/imds/internal/DefaultEc2MetadataAsyncClient.java
L-Applin: Yes we do, because the handler contains the future that has the actual value of the request.

core/imds/src/main/java/software/amazon/awssdk/imds/internal/DefaultEc2MetadataAsyncClient.java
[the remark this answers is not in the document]
cenedhryn: So far we've avoided forcing users into taking a dependency on the CRT without them making an active choice, except for with Transfer Manager, because we don't own CRT and there may be issues (I'm vague because it's been a while). In S3 for instance, which uses auth-crt, we check the classpath to see if it's included. Let's discuss offline what the best course of action is.

core/imds/src/main/java/software/amazon/awssdk/imds/internal/DefaultEc2MetadataAsyncClient.java
```diff
@@ -0,0 +1,310 @@
...
+    @Override
+    public CompletableFuture<MetadataResponse> get(String path) {
+        return get(path, RetryPolicyContext.builder().retriesAttempted(0).build());
+    }
+
+    private CompletableFuture<MetadataResponse> get(String path, RetryPolicyContext retryPolicyContext) {
+        SdkHttpFullRequest baseTokenRequest = requestMarshaller.createTokenRequest(tokenTtl);
+
+        CompletableFuture<MetadataResponse> result = createAsyncRequest(baseTokenRequest)
+            .thenComposeAsync(token -> {
```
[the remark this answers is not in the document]
cenedhryn: old comment!

core/imds/src/main/java/software/amazon/awssdk/imds/internal/RequestMarshaller.java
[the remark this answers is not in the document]
cenedhryn: Makes sense

core/imds/src/main/java/software/amazon/awssdk/imds/internal/DefaultEc2MetadataAsyncClient.java
L-Applin: how about `sendAsyncRequest(...)`? We have a private method `sendRequest(...)` in the sync client.

core/imds/src/main/java/software/amazon/awssdk/imds/internal/DefaultEc2MetadataAsyncClient.java
L-Applin: I haven't seen any method in the CompletableFuture API that lets me compose it with the error, when it is completed exceptionally. That feature is precisely `.exceptionallyCompose()` but only available in Java 12... because a `CompletableFuture` is returned within the `handle()` method, we have a `CompletableFuture<CompletableFuture<MetadataResponse>>`, `Function.identity()` is there just to unwrap the first `CompletableFuture`

core/imds/src/test/java/software/amazon/awssdk/imds/internal/DefaultEc2MetadataClientAsyncClientTest.java
L-Applin: this is from a git conflict resolve error, removing duplicate.

core/imds/src/main/java/software/amazon/awssdk/imds/internal/DefaultEc2MetadataAsyncClient.java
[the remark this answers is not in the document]
L-Applin: as discussed, we will instead use the `DefaultSdkHttpClientBuilder` to instanciate the http client

core/imds/src/test/java/software/amazon/awssdk/imds/Ec2MetadataClientTest.java
L-Applin: The idea is to test these cases:
- token request succeed (with or without retry) then request succeed.
- token request succeed (with or without retry) then request fails.
- token request fails (with or without retry) then request succeed. 
- token request fails (with or without retry) then request fails.

core/imds/src/main/java/software/amazon/awssdk/imds/internal/DefaultEc2MetadataAsyncClient.java
```diff
@@ -0,0 +1,287 @@
...
+        this.endpointMode = builder.endpointMode != null ? builder.endpointMode
+                                                         : DEFAULT_ENDPOINT_PROVIDER.resolveEndpointMode();
+        this.endpoint = builder.endpoint != null ? builder.endpoint
+                                                 : URI.create(DEFAULT_ENDPOINT_PROVIDER.resolveEndpoint(this.endpointMode));
+        this.requestMarshaller = new RequestMarshaller(this.endpoint);
+        this.tokenTtl = builder.tokenTtl != null ? builder.tokenTtl
+                                                 : Duration.ofSeconds(21_600);
+        this.asyncRetryScheduler = builder.scheduledExecutorService != null
+                                   ? builder.scheduledExecutorService
+                                   : Executors.newScheduledThreadPool(DEFAULT_RETRY_THREAD_POOL_SIZE);
```
zoewangg, line 86: Can we add name prefix? Probably "imds-ScheduledExecutor" to be consistent with the one used in the core.

https://github.com/aws/aws-sdk-java-v2/blob/master/core/sdk-core/src/main/java/software/amazon/awssdk/core/client/builder/SdkDefaultClientBuilder.java#L410

core/imds/src/main/java/software/amazon/awssdk/imds/internal/DefaultEc2MetadataAsyncClient.java
```diff
@@ -0,0 +1,287 @@
...
+
+    private static final class StringResponseHandler implements HttpResponseHandler<String> {
+        @Override
+        public String handle(SdkHttpFullResponse response, ExecutionAttributes executionAttributes) throws Exception {
+            return IoUtils.toUtf8String(response.content().orElseThrow(() ->
+                SdkClientException.create("Unexpected error: empty response content")));
+        }
+    }
+
+    private static final class StringAsyncResponseHandler implements TransformingAsyncResponseHandler<String> {
```
zoewangg, line 178: Do we actually need this class? Can't we check the status code in `StringResponseHandler`?

core/imds/src/main/java/software/amazon/awssdk/imds/internal/DefaultEc2MetadataAsyncClient.java
```diff
@@ -0,0 +1,287 @@
...
+            }
+            int newAttempt = retryPolicyContext.retriesAttempted() + 1;
+            log.debug(() -> "Retrying request: Attempt " + newAttempt);
+            RetryPolicyContext newContext =
+                RetryPolicyContext.builder()
+                                  .request(baseTokenRequest)
+                                  .retriesAttempted(newAttempt)
+                                  .exception(SdkClientException.create(error.getMessage(), error))
+                                  .build();
+            return scheduledRetryAttempt(() -> get(path, newContext), newContext);
```
zoewangg, line 123: It may be cleaner to create a return future, pass the future to `get` and completes it there than returning chained result IMO.

core/imds/src/test/java/software/amazon/awssdk/imds/internal/DefaultEc2MetadataAsyncClientTest.java
```diff
@@ -0,0 +1,337 @@
...
+            MetadataResponse response = res.get();
+            assertThat(res).isCompleted();
+            assertThat(response.asString()).isEqualTo("{}");
+            verify(exactly(1), putRequestedFor(urlPathEqualTo(TOKEN_RESOURCE_PATH))
+                .withHeader(EC2_METADATA_TOKEN_TTL_HEADER, equalTo("1024")));
+            verify(exactly(1), getRequestedFor(urlPathEqualTo(AMI_ID_RESOURCE))
+                .withHeader(TOKEN_HEADER, equalTo("some-token")));
+        } catch (Exception e) {
+            fail("Unexpected exception while executing test", e);
+        }
```
zoewangg, line 97: `catch` seems unnecessary because the test should fail if there is an exception thrown

core/imds/src/test/java/software/amazon/awssdk/imds/internal/DefaultEc2MetadataAsyncClientTest.java
```diff
@@ -0,0 +1,337 @@
...
+                           "http://localhost:" + mockMetadataEndpoint.port());
+        try (Ec2MetadataAsyncClient client =
+                 Ec2MetadataAsyncClient.builder()
+                                       .endpoint(URI.create("http://localhost:" + mockMetadataEndpoint.port()))
+                                       .tokenTtl(Duration.ofSeconds(1024))
+                                       .build())
+        {
+            CompletableFuture<MetadataResponse> res = client.get(AMI_ID_RESOURCE);
+            assertThat(res).isNotCompleted();
+            assertThatThrownBy(res::get).isInstanceOf(ExecutionException.class);
```
zoewangg, line 140: Can we verify the exact exception type?

core/imds/src/test/java/software/amazon/awssdk/imds/internal/DefaultEc2MetadataAsyncClientTest.java
```diff
@@ -0,0 +1,337 @@
...
+        System.setProperty(SdkSystemSetting.AWS_EC2_METADATA_SERVICE_ENDPOINT.property(),
+                           "http://localhost:" + mockMetadataEndpoint.port());
+        try (Ec2MetadataAsyncClient client =
+                Ec2MetadataAsyncClient.builder()
+                                      .endpoint(URI.create("http://localhost:" + mockMetadataEndpoint.port()))
+                                      .tokenTtl(Duration.ofSeconds(1024))
+                                      .build())
+        {
+            CompletableFuture<MetadataResponse> res = client.get(AMI_ID_RESOURCE);
+            assertThat(res).isNotCompleted();
```
zoewangg, line 87: `assertThat(res).isNotCompleted();` and `assertThat(res).isCompleted()` seem unnecessary. Same for other test code

core/imds/src/test/java/software/amazon/awssdk/imds/internal/DefaultEc2MetadataAsyncClientTest.java
```diff
@@ -0,0 +1,337 @@
...
+                .withHeader(EC2_METADATA_TOKEN_TTL_HEADER, equalTo("1024")));
+            verify(exactly(totalRequests), getRequestedFor(urlPathMatching(AMI_ID_RESOURCE + "/" + "\\d+"))
+                .withHeader(TOKEN_HEADER, equalTo("some-token")));
+        } catch (Exception e) {
+            fail("Unexpected exception while executing test", e);
+        }
+    }
+
+    @Test
+    public void get_largeResponse_shouldSucceed() {
```
zoewangg, line 304: Is this test necessary? It doesn't seem to be related to any logic in IMDS. Maybe it's more applicable to http client?

core/imds/src/main/java/software/amazon/awssdk/imds/internal/DefaultEc2MetadataAsyncClient.java
```diff
@@ -0,0 +1,287 @@
...
+    private final RequestMarshaller requestMarshaller;
+    private final Duration tokenTtl;
+    private final ScheduledExecutorService asyncRetryScheduler;
+
+    private DefaultEc2MetadataAsyncClient(Ec2MetadataAsyncBuilder builder) {
+        this.retryPolicy = builder.retryPolicy != null ? builder.retryPolicy
+                                                       : Ec2MetadataRetryPolicy.builder().build();
+        this.endpointMode = builder.endpointMode != null ? builder.endpointMode
+                                                         : DEFAULT_ENDPOINT_PROVIDER.resolveEndpointMode();
+        this.endpoint = builder.endpoint != null ? builder.endpoint
```
zoewangg, line 79: Endpoint and endpoint mode should be mutually exclusive, right? Can we add validation? We can use this method https://github.com/aws/aws-sdk-java-v2/blob/master/utils/src/main/java/software/amazon/awssdk/utils/Validate.java#L786

core/imds/src/main/java/software/amazon/awssdk/imds/internal/DefaultEc2MetadataAsyncClient.java
L-Applin: We need it to build an `AsyncExecuteRequest` instance

core/imds/src/test/java/software/amazon/awssdk/imds/internal/DefaultEc2MetadataAsyncClientTest.java
L-Applin: this test actually allowed me to catch a problem where the response bytes were "chuncked" and I was not receiving all of the data, and I had to update my code. I think it is good to keep to test for that case.

core/imds/src/main/java/software/amazon/awssdk/imds/internal/DefaultEc2MetadataAsyncClient.java
```diff
@@ -0,0 +1,287 @@
...
+    private boolean shouldRetry(RetryPolicyContext retryPolicyContext, Throwable error) {
+        boolean maxAttemptReached = retryPolicyContext.retriesAttempted() >= retryPolicy.numRetries();
+        if (maxAttemptReached) {
+            return false;
+        }
+        return error instanceof RetryableException || error.getCause() instanceof RetryableException;
+    }
+
+    @Override
+    public Builder toBuilder() {
```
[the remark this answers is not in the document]
L-Applin: It was there from the intern code in the sync client, I agree we dont really need it, I'll remove it

core/imds/src/main/java/software/amazon/awssdk/imds/internal/DefaultEc2MetadataAsyncClient.java
L-Applin: Ok, I got it, removing AsyncHandler class

core/imds/src/main/java/software/amazon/awssdk/imds/internal/DefaultEc2MetadataAsyncClient.java
L-Applin: from the SEP:

> If endpoint is specified, the client MUST use the explicitly provided endpoint.
> 
> If endpoint_mode is supported and endpoint is unspecified
> 
> If set to IPv4, use http://169.254.169.254
> If set to IPv6, use http://[fd00:ec2::254]

By that, I understand that both `endpoint_mode` and `endpoint` could be specified, in which case we ignore `endpoint_mode` and use the `endpoint` value directly. What do you think?

L-Applin: After discussion, we will implement them being mutually exclusive

core/imds/src/main/java/software/amazon/awssdk/imds/Ec2MetadataClientBuilder.java
```diff
@@ -0,0 +1,70 @@
...
+     *
+     * @param endpoint The endpoint of IMDS.
+     * @return a reference to this builder
+     */
+    B endpoint(URI endpoint);
+
+    URI getEndpoint();
+
+    /**
+     * Define the Time to live (TTL) of the token.
```
cenedhryn, line 51: Can we give more context to the user on what this parameter does? Which token, what is the default value, when would you set it?

core/imds/src/main/java/software/amazon/awssdk/imds/Ec2MetadataClientBuilder.java
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * Base shared builder interface for Ec2MetadataClient
+ * @param <B> the Builder Type
+ * @param <T> the Ec2MetadataClient Type
+ */
+@SdkPublicApi
+public interface Ec2MetadataClientBuilder<B, T> extends SdkBuilder<Ec2MetadataClientBuilder<B, T>, T> {
+    /**
+     * Define the retry policy which includes the number of retry attempts for any failed request.
```
cenedhryn, line 31: What happens if you don't provide it? What are the default values?

core/imds/src/main/java/software/amazon/awssdk/imds/Ec2MetadataClientBuilder.java
```diff
@@ -0,0 +1,70 @@
...
+     *
+     * @param retryPolicy The retry policy which includes the number of retry attempts for any failed request.
+     * @return a reference to this builder
+     */
+    B retryPolicy(Ec2MetadataRetryPolicy retryPolicy);
+
+    Ec2MetadataRetryPolicy getRetryPolicy();
+
+    /**
+     * Define the endpoint of IMDS.
```
cenedhryn, line 41: What happens if you don't provide this? What is the default value?

core/imds/src/main/java/software/amazon/awssdk/imds/Ec2MetadataClient.java
```diff
@@ -51,57 +45,23 @@ static Ec2MetadataClient create() {
...
-         * @return Returns a reference to this builder
-         */
-        Builder endpointMode(EndpointMode endpointMode);
-
-        /**
-         * Define the SdkHttpClient instance to make the http requests.
-         *
-         * @param httpClient The SdkHttpClient instance to make the http requests.
-         * @return Returns a reference to this builder
+         * Define the http client used by the Ec2 Metadata client. If provided, the Ec2MetadataClient will <em>NOT</em> manage the
```
cenedhryn, line 58: What happens if it's not provided? What is the default?

core/imds/src/main/java/software/amazon/awssdk/imds/Ec2MetadataClientBuilder.java
```diff
@@ -0,0 +1,70 @@
...
+     *
+     * @param tokenTtl The Time to live (TTL) of the token.
+     * @return a reference to this builder
+     */
+    B tokenTtl(Duration tokenTtl);
+
+    Duration getTokenTtl();
+
+    /**
+     * Define the endpoint mode of IMDS. Supported values include IPv4 and IPv6.
```
cenedhryn, line 61: What is the default value?

core/imds/src/main/java/software/amazon/awssdk/imds/internal/DefaultEc2MetadataAsyncClient.java
```diff
@@ -95,22 +86,31 @@ public static Ec2MetadataAsyncClient.Builder builder() {
...
 
-        return result.handle((response, error) -> {
-            if (response != null || !shouldRetry(retryPolicyContext, error)) {
-                return result;
+        result.whenComplete((response, error) -> {
+            if (response != null) {
+                returnFuture.complete(response);
+                return;
+            }
+            if (!shouldRetry(retryPolicyContext, error)) {
```
cenedhryn, line 111: Is it possible to have a case of response being null but there isn't an error?

core/imds/src/main/java/software/amazon/awssdk/imds/internal/DefaultEc2MetadataAsyncClient.java
```diff
@@ -120,28 +120,33 @@ private CompletableFuture<MetadataResponse> get(String path, RetryPolicyContext
...
     }
 
     private CompletableFuture<String> sendAsyncRequest(SdkHttpFullRequest baseRequest) {
-        StringAsyncResponseHandler responseHandler = new StringAsyncResponseHandler();
         SdkHttpContentPublisher requestContentPublisher = new SimpleHttpContentPublisher(baseRequest);
+        StringResponseHandler stringResponseHandler = new StringResponseHandler();
+        TransformingAsyncResponseHandler<String> responseHandler = new AsyncResponseHandler<>(stringResponseHandler,
+                                                                                              Function.identity(),
+                                                                                              new ExecutionAttributes());
+        CompletableFuture<String> future = responseHandler.prepare();
```
cenedhryn, line 133: Nit: There's a lot of futures to handle; can this future have a more descriptive name?

core/imds/src/main/java/software/amazon/awssdk/imds/internal/DefaultEc2MetadataAsyncClient.java
```diff
@@ -52,41 +49,35 @@
...
-                          : new DefaultSdkAsyncHttpClientBuilder().buildWithDefaults(AttributeMap.empty());
+        super(builder);
+        this.httpClient = Validate.getOrDefault(builder.httpClient,
+            () -> new DefaultSdkAsyncHttpClientBuilder().buildWithDefaults(AttributeMap.empty()));
+        this.asyncRetryScheduler = Validate.getOrDefault(builder.scheduledExecutorService,
+            () -> {
+                ThreadFactory threadFactory = new ThreadFactoryBuilder().threadNamePrefix("IMDS-ScheduledExecutor").build();
+                return Executors.newScheduledThreadPool(DEFAULT_RETRY_THREAD_POOL_SIZE, threadFactory);
+            });
+        this.isHttpClientManaged = builder.httpClient == null;
```
cenedhryn, line 79: Does `managed` mean managed by us? I first thought it was managed by the user but it was the opposite. Can we change it to `isInternal...` or `isDefault...` instead?

core/imds/src/test/java/software/amazon/awssdk/imds/internal/BaseEc2MetadataClientTest.java
```diff
@@ -0,0 +1,246 @@
...
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+import software.amazon.awssdk.core.exception.SdkClientException;
+import software.amazon.awssdk.core.retry.backoff.FixedDelayBackoffStrategy;
+import software.amazon.awssdk.imds.Ec2MetadataClientBuilder;
+import software.amazon.awssdk.imds.Ec2MetadataRetryPolicy;
+import software.amazon.awssdk.imds.EndpointMode;
+import software.amazon.awssdk.imds.MetadataResponse;
+
+public abstract class BaseEc2MetadataClientTest<T, B extends Ec2MetadataClientBuilder<B, T>> {
```
cenedhryn, line 53: This is a really nice set of tests, and very efficient and readable with the shared test cases.

core/imds/src/main/java/software/amazon/awssdk/imds/internal/DefaultEc2MetadataAsyncClient.java
L-Applin: I don't think that this is possible, no

cenedhryn (approved):

core/imds/pom.xml
```diff
@@ -114,7 +114,7 @@
         </dependency>
         <dependency>
             <groupId>software.amazon.awssdk</groupId>
-            <artifactId>apache-client</artifactId>
+            <artifactId>netty-nio-client</artifactId>
             <version>${awsjavasdk.version}</version>
             <scope>test</scope>
```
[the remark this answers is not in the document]
L-Applin: Added to a draft Qip doc for surface area review

core/imds/src/main/java/software/amazon/awssdk/imds/Ec2MetadataAsyncClient.java
```diff
@@ -0,0 +1,81 @@
...
+     * The builder definition for a {@link Ec2MetadataClient}. All parameters are optional and have default values if not
+     * specified. Therefore, an instance can be simply created with {@code Ec2MetadataAsyncClient.builder().build()} or
+     * {@code Ec2MetadataAsyncClient.create()}, both having the same result.
+     */
+    interface Builder extends Ec2MetadataClientBuilder<Ec2MetadataAsyncClient.Builder, Ec2MetadataAsyncClient> {
+
+        /**
+         * Define the {@link ScheduledExecutorService} used to schedule asynchronous retry attempts. If not specified,
+         * defaults to {@link Executors#newScheduledThreadPool} with a default value of 3 thread in the pool.
+         * If provided, the Ec2MetadataClient will <em>NOT</em> manage the lifetime if the httpClient and must therefore be
```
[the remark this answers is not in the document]
L-Applin: fixed in 3543

core/imds/src/main/java/software/amazon/awssdk/imds/internal/DefaultEc2MetadataAsyncClient.java
```diff
@@ -0,0 +1,279 @@
...
+
+    private final SdkAsyncHttpClient httpClient;
+    private final ScheduledExecutorService asyncRetryScheduler;
+    private final boolean isHttpClientManaged;
+    private final boolean isRetryExecutorManaged;
+
+    private DefaultEc2MetadataAsyncClient(Ec2MetadataAsyncBuilder builder) {
+        super(builder);
+        this.httpClient = Validate.getOrDefault(builder.httpClient,
+            () -> new DefaultSdkAsyncHttpClientBuilder().buildWithDefaults(AttributeMap.empty()));
```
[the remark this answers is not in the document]
L-Applin: Fixed in 3543

core/imds/src/main/java/software/amazon/awssdk/imds/internal/DefaultEc2MetadataAsyncClient.java
```diff
@@ -0,0 +1,279 @@
...
+        return error instanceof RetryableException || error.getCause() instanceof RetryableException;
+    }
+
+    @Override
+    public void close() {
+        if (isHttpClientManaged) {
+            httpClient.close();
+        }
+        if (isRetryExecutorManaged) {
+            asyncRetryScheduler.shutdown();
```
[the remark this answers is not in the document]
L-Applin: Fixed in 3543

core/imds/src/main/java/software/amazon/awssdk/imds/internal/Ec2MetadataEndpointProvider.java
```diff
@@ -36,6 +36,8 @@
 @SdkInternalApi
 public final class Ec2MetadataEndpointProvider {
 
+    public static final Ec2MetadataEndpointProvider DEFAULT_ENDPOINT_PROVIDER = builder().build();
```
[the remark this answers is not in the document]
L-Applin: Fixed in 3543

Issue #370, opened 2017-12-22
Presigned CloudFront URL Generator


marekzebrowski:
1.11.x has `com.amazonaws.services.cloudfront.CloudFrontUrlSigner` used to generate pre-signed urls
also `com.amazonaws.services.cloudfront.CloudFrontCookieSigner` does not have a counterpart

2.x lacks this functionality

selvakumar-invicara:
Is it solved ? Recently we migrated to Java 11 and trying to use AWS SDK 2.x. But couldn't find replacements for CloudFrontUrlSigner.

thereisnospoon:
Hey,
What's the status of it? We use sdk V2 and we need this feature.

debora-ito:
No updates here.

You can track the features we're currently working on and some of what we're planning to work in our backlog of public features: https://github.com/aws/aws-sdk-java-v2/projects/1

jbauerrfid:
Found the same problem today.

A possible workaround: try to generate the signed URL yourself, instructions are on https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-creating-signed-url-custom-policy.html

I see that this is not nice, but this feature will probably a bit longer to complete, and I can't wait...

jotamusik:
@jbauerrfid did you managed to do it by yourself?? 🤔 If you did, could you please share the solution? 
Thanks in advance and hope that this feature comes as soon as posible 🙌🏻

reply2srij:
Another vote for this feature. Important feature to have.

yamakz5555:
#242 
CloudFront utility has been removed.
I hope it will come back.

nowucca:
I'm trying buiild a private content repository on top of the cloudfront signing capability.
Why is this use case so difficult to bring back?
+1 for working on this.

xps2:
+1

marcolorefice:
+1

debora-ito:
Good news, everyone!

This has been released in Java SDK version `2.18.33`.

See our Blog Post that goes through the new SDK 2.x CloudFront Signed URLs and Cookies feature - https://aws.amazon.com/blogs/developer/amazon-cloudfront-signed-urls-and-cookies-are-now-supported-in-aws-sdk-for-java-2-x/

Please check it out and let us know if you have any feedback.

PR #3529, opened 2022-10-31
CloudFront Signed Url and Cookie

davidh44:
V2 Parity - CloudFront Signed Url and Cookie Feature

## Motivation and Context

Fixes https://github.com/aws/aws-sdk-java-v2/issues/370

## Modifications

Added classes and utils for CloudFront signers

## Testing

Added unit and integration tests

## Types of changes

## Checklist

## License


services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/auth/Asn1Object.java
```diff
@@ -0,0 +1,138 @@
...
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.services.cloudfront.auth;
+
+import java.io.IOException;
+import java.math.BigInteger;
+import software.amazon.awssdk.annotations.SdkInternalApi;
+
+@SdkInternalApi
```
cenedhryn, line 22: Put `SdkInternalApi` classes in the `internal` package like /internal/auth
If there are auth classes that *need* to be public, keep them in /auth

davidh44: Created internal package and moved files

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
```diff
@@ -0,0 +1,753 @@
...
+     * @param expirationDate
+     *            The expiration date of the signed URL in UTC
+     * @return The signed URL.
+     */
+    public static String getSignedUrlWithCannedPolicy(Protocol protocol,
+                                                      String distributionDomain,
+                                                      String s3ObjectKey,
+                                                      File privateKeyFile,
+                                                      String keyPairId,
+                                                      Instant expirationDate) throws InvalidKeySpecException, IOException {
```
cenedhryn, line 94: Create a POJO request class to represent the input parameters. You can put it under `/model`

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
```diff
@@ -0,0 +1,753 @@
...
+        PrivateKey privateKey = loadPrivateKey(privateKeyFile);
+        String resourceUrl = generateResourceUrl(protocol, distributionDomain, s3ObjectKey);
+        return getSignedUrlWithCannedPolicy(resourceUrl, privateKey, keyPairId, expirationDate);
+    }
+
+    /**
+     * Generate a signed URL that allows access to a specific distribution and
+     * S3 object by applying a access restrictions from a "canned" (simplified)
+     * policy document.
```
cenedhryn, line 103: This sentence doesn't read right. I also think we should be able to distinguish between the two overloaded methods. What makes them differ?

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
```diff
@@ -0,0 +1,753 @@
...
+                                                      String ipRange) throws InvalidKeySpecException, IOException {
+        PrivateKey privateKey = loadPrivateKey(privateKeyFile);
+        String resourceUrl = generateResourceUrl(protocol, distributionDomain, s3ObjectKey);
+        String policy = buildCustomPolicyForSignedUrl(resourceUrl, activeDate, expirationDate, ipRange);
+        return getSignedUrlWithCustomPolicy(resourceUrl, privateKey, keyPairId, policy);
+    }
+
+    /**
+     * Generate a signed URL that allows access to distribution and S3 objects
+     * by applying access restrictions specified in a custom policy document.
```
cenedhryn, line 188: It can be helpful to include that the custom policy itself is included as a part of the presigned url (and for the other method, that it's not).

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
```diff
@@ -0,0 +1,753 @@
...
+        } catch (InvalidKeyException e) {
+            throw SdkClientException.create("Could not sign custom policy cookie", e);
+        }
+    }
+
+    /**
+     * Contains common cookies used by Amazon CloudFront.
+     */
+    @SdkPublicApi
+    public static final class SignedCookie {
```
cenedhryn, line 362: This is a big class, it's better to extract `SignedCookie` to it's own class file.

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
```diff
@@ -0,0 +1,753 @@
...
+     * @param activeDate
+     *            The beginning valid date of the signed URL in UTC
+     * @param expirationDate
+     *            The expiration date of the signed URL in UTC
+     * @param ipRange
+     *            The allowed IP address range of the client making the GET
+     *            request, in CIDR form (e.g. 192.168.0.1/24).
+     * @return The signed URL.
+     */
+    public static String getSignedUrlWithCustomPolicy(Protocol protocol,
```
cenedhryn, line 172: I find the combination of input parameters to the two flavors of methods unintuitive (I realized it's probably an inheritance from v1). Do we know why we're combining parts of a url with a private key file as opposed to complete url with a key? Do we need the overload of parts of a url at all?


Do we even need the methods with protocol, distribution domain and object key?

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
davidh44: created CloudFrontSignedCookie class

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
davidh44: Added description to mention this

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
davidh44: Replaced

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
davidh44: Added CloudFrontSignerRequest under /model directory

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
davidh44: Condensed to one variant of each type, which takes a CloudFrontSignerRequest as parameter. Replaced protocol/ distribution domain/object key with resourceUrl, and key file with key

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/internal/auth/Pem.java
```diff
@@ -0,0 +1,117 @@
...
+import java.io.InputStreamReader;
+import java.security.PrivateKey;
+import java.security.PublicKey;
+import java.security.spec.InvalidKeySpecException;
+import java.util.ArrayList;
+import java.util.Base64;
+import java.util.List;
+import java.util.regex.Pattern;
+
+public enum Pem {
```
cenedhryn, line 30: Can we convert these empty enums to static class/private ctor instead? Empty enums are another way of achieving the same thing but I think it's mostly confusing unless you really need it. The internal annotation is also missing. 

Same for Rsa.

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/internal/auth/PemObjectType.java
```diff
@@ -0,0 +1,39 @@
...
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.services.cloudfront.internal.auth;
+
+import java.util.Arrays;
+
+public enum PemObjectType {
```
cenedhryn, line 20: Internal Sdk annotation missing

services/cloudfront/pom.xml
```diff
@@ -56,5 +56,11 @@
             <artifactId>protocol-core</artifactId>
             <version>${awsjavasdk.version}</version>
         </dependency>
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>s3</artifactId>
+            <version>2.18.8-SNAPSHOT</version>
```
cenedhryn, line 62: You have a snapshot version hardcoded here instead of ${awsjavasdk.version}

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
```diff
@@ -0,0 +1,368 @@
...
+     * @param request
+     *            A CloudFrontSignerRequest configured with the following values:
+     *            resourceUrl, privateKey, keyPairId, expirationDate, activeDate, ipRange
+     * @return A signed URL that will permit access to distribution and S3
+     *         objects as specified in the policy document.
+     */
+    public static String getSignedUrlWithCustomPolicy(CloudFrontSignerRequest request) {
+        try {
+            String policy = buildCustomPolicyForSignedUrl(request.resourceUrl(), request.activeDate(), request.expirationDate(),
+                                                                                request.ipRange());
```
cenedhryn, line 113: Nit: formatting

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/model/CloudFrontSignerRequest.java
```diff
@@ -0,0 +1,217 @@
...
+         *     HTTPS required protocol, the resource URL must start with
+         *     <tt>"https://"</tt>
+         * </p>
+         */
+        Builder resourceUrl(String resourceUrl);
+
+        /**
+         * Configure the private key to be used to sign the policy
+         */
+        Builder privateKey(PrivateKey privateKey);
```
cenedhryn, line 127: Since we're not letting the users supply the file directly anymore, the javadoc should probably point to the load.. method in the utility class

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/model/CloudFrontSignerRequest.java
```diff
@@ -0,0 +1,217 @@
...
+import software.amazon.awssdk.utils.builder.CopyableBuilder;
+import software.amazon.awssdk.utils.builder.ToCopyableBuilder;
+
+/**
+ * Request to generate a CloudFront signed URL or signed cookie
+ */
+@Immutable
+@ThreadSafe
+@SdkPublicApi
+public final class CloudFrontSignerRequest
```
cenedhryn, line 34: Missing equals, hashcode (, toString)

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontSignedCookie.java
```diff
@@ -0,0 +1,251 @@
...
+import software.amazon.awssdk.utils.builder.CopyableBuilder;
+import software.amazon.awssdk.utils.builder.ToCopyableBuilder;
+
+/**
+ * Contains common cookies used by Amazon CloudFront.
+ */
+@Immutable
+@ThreadSafe
+@SdkPublicApi
+public final class CloudFrontSignedCookie implements ToCopyableBuilder<CloudFrontSignedCookie.Builder, CloudFrontSignedCookie> {
```
cenedhryn, line 33: Missing equals, hashcode (, toString) methods

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/model/CloudFrontSignerRequest.java
```diff
@@ -0,0 +1,217 @@
...
+         * Configure the IP range of the signed URL or signed cookie - for custom policies (optional field)
+         * <p>
+         *     The allowed IP address range of the client making the GET
+         *     request, in CIDR form (e.g. 192.168.0.1/24).
+         * </p>
+         */
+        Builder ipRange(String ipRange);
+    }
+
+    @SdkInternalApi
```
cenedhryn, line 154: Nit: Not needed, having it on the main class is enough

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontSignedCookie.java
```diff
@@ -0,0 +1,251 @@
...
+        return isCustom;
+    }
+
+    /**
+     * Returns the cookie header value for the specified cookie type.
+     *
+     * @param type The type of the cookie header value to retrieve
+     * @return The cookie header value to pass into an HTTP request
+     */
+    public String cookieHeaderValue(String type) {
```
cenedhryn, line 145: Where does this type come from and how do users know what to query for?

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontSignedCookie.java
```diff
@@ -0,0 +1,251 @@
...
+                }
+                return POLICY_KEY + "=" + policyVal;
+            default:
+                throw SdkClientException.create("Did not provide a valid cookie type");
+        }
+    }
+
+    @NotThreadSafe
+    @SdkPublicApi
+    public interface Builder extends CopyableBuilder<CloudFrontSignedCookie.Builder, CloudFrontSignedCookie> {
```
cenedhryn, line 168: Combining the two types of cookies in one class comes with some awkward usage, since you never really know whether to use expires or policy. 

Consider creating a set of interfaces to represent a signed cookie, a canned signed cookie and a custom signed cookie and then add two internal default classes to represent the two implementations. In this case the main interface could define all public getters plus a builder interface. 

The benefit is that you don't need to check the isCustoms flag, and you could even have a toSdkHttpRequest method that can generate the whole request for you

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/internal/auth/Pem.java
cenedhryn, line 30: Can we convert this (copy paste) class to a static utility class instead of using an empty enum? I haven't seen this pattern used in our codebase and it seems unnecessary unless you really need it. 

Missing internal annotation

Same for rsa

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/internal/auth/PemObjectType.java
cenedhryn, line 20: internal annotation missing

services/cloudfront/src/test/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilitiesTest.java
```diff
@@ -0,0 +1,131 @@
...
+import java.time.Instant;
+import java.time.ZoneOffset;
+import java.util.Base64;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+import software.amazon.awssdk.services.cloudfront.model.CloudFrontSignerRequest;
+
+
+class CloudFrontUtilitiesTest {
```
cenedhryn, line 36: Missing tests for cookie

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
```diff
@@ -0,0 +1,368 @@
...
+ *     Signed URLs take precedence over signed cookies. If you use both signed URLs and signed cookies to control access to the
+ *     same files and a viewer uses a signed URL to request a file, CloudFront determines whether to return the file to the
+ *     viewer based only on the signed URL.
+ * </p>
+ *
+ */
+@Immutable
+@ThreadSafe
+@SdkPublicApi
+public final class CloudFrontUtilities {
```
cenedhryn, line 62: The API is looking cleaner. After deliberating, I think we should add a response object to represent the url, just as there is one for cookie. It's more work now, but it makes it possible to add more functionality and features for the user later such as letting them get the encoded url directly from the response etc, instead of just having a string.

cenedhryn (commented):
The POJO classes should have unit tests and test equals/hashcode `equalsverifier`

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/model/CloudFrontSignerRequest.java
davidh44: removed

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontSignedCookie.java
```diff
@@ -0,0 +1,251 @@
...
+        if (!this.isCustom) {
+            throw SdkClientException.create("This is a cookie with a canned policy, use expires instead of policy");
+        }
+        return POLICY_KEY;
+    }
+
+    /**
+     * Returns the key pair ID value
+     */
+    public String keyPairIdVal() {
```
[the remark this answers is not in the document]
davidh44: updated to value

services/cloudfront/pom.xml
davidh44: replaced

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
davidh44: corrected

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/model/CloudFrontSignerRequest.java
davidh44: added @link

.changes/next-release/feature-AWSSDKforJavav2-10130fc.json
```diff
@@ -0,0 +1,6 @@
+{
+    "category": "AWS SDK for Java v2", 
+    "contributor": "", 
+    "type": "feature", 
+    "description": "CloudFront brings feature parity for singed url and signed cookies"
```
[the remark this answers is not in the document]
davidh44: updated

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontSignedCookie.java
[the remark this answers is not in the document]
davidh44: updated to value

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontSignedCookie.java
davidh44: Can make an enum for this

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/internal/auth/PemObjectType.java
davidh44: added

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/internal/auth/Pem.java
davidh44: updated

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/model/CloudFrontSignerRequest.java
davidh44: added equals, hashcode

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontSignedCookie.java
davidh44: added equals, hashcode

services/cloudfront/pom.xml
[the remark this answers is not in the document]
davidh44: replaced

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontSignedCookie.java
davidh44: Implemented interfaces and default classes

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/internal/auth/Pem.java
davidh44: converted and added annotation

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/internal/auth/PemObjectType.java
davidh44: added

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
davidh44: created SignedUrl response object

services/cloudfront/src/test/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilitiesTest.java
davidh44: adding tests

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/internal/cookies/CookiesForCustomPolicy.java
```diff
@@ -0,0 +1,65 @@
...
+    String policyKey();
+
+    /**
+     * Returns the policy value
+     */
+    String policyValue();
+
+    @NotThreadSafe
+    @SdkPublicApi
+    interface Builder extends CopyableBuilder<CookiesForCustomPolicy.Builder, CookiesForCustomPolicy> {
```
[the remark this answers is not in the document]
cenedhryn: One similar example Ec2MetadataClientBuilder: https://github.com/aws/aws-sdk-java-v2/pull/3523/files#diff-6e379add12ecb08e99873e92abf4af11433f0fab588fd8faf680248d76808736

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/cookie/DefaultCookiesForCannedPolicy.java
```diff
@@ -0,0 +1,197 @@
...
+import software.amazon.awssdk.annotations.ThreadSafe;
+import software.amazon.awssdk.core.exception.SdkClientException;
+import software.amazon.awssdk.http.SdkHttpMethod;
+import software.amazon.awssdk.http.SdkHttpRequest;
+import software.amazon.awssdk.services.cloudfront.internal.cookies.CookiesForCannedPolicy;
+
+@Immutable
+@ThreadSafe
+@SdkPublicApi
+public final class DefaultCookiesForCannedPolicy implements CookiesForCannedPolicy {
```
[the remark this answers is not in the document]
davidh44: switched

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/internal/url/SignedUrl.java
```diff
@@ -0,0 +1,48 @@
...
+     */
+    String url();
+
+    /**
+     * Generates an HTTP request that can be executed by an HTTP client to access the resource
+     */
+    SdkHttpRequest generateHttpRequest();
+
+    @NotThreadSafe
+    @SdkPublicApi
```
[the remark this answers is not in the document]
davidh44: removed

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/internal/cookies/CookiesForCustomPolicy.java
```diff
@@ -0,0 +1,65 @@
...
+
+package software.amazon.awssdk.services.cloudfront.internal.cookies;
+
+import software.amazon.awssdk.annotations.NotThreadSafe;
+import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.annotations.SdkPublicApi;
+import software.amazon.awssdk.utils.builder.CopyableBuilder;
+import software.amazon.awssdk.utils.builder.ToCopyableBuilder;
+
+@SdkInternalApi
```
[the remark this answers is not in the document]
davidh44: made public

services/cloudfront/src/test/java/software/amazon/awssdk/services/cloudfront/SignedCookieTest.java
```diff
@@ -0,0 +1,106 @@
...
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.services.cloudfront;
```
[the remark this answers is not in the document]
davidh44: updated packages

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/internal/url/SignedUrl.java
```diff
@@ -0,0 +1,48 @@
...
+    interface Builder extends CopyableBuilder<Builder, SignedUrl> {
```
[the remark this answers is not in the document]
davidh44: removed

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/internal/cookies/CookiesForCustomPolicy.java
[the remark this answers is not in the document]
davidh44: Wasn't able to implement this as CookiesForCustomPolicy has a policy() method, while CookiesForCannedPolicy has an expires() method

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
```diff
@@ -67,6 +67,8 @@
 @SdkPublicApi
 public final class CloudFrontUtilities {
```
cenedhryn, line 68: Missing the SigningUtils class.

services/cloudfront/src/test/java/software/amazon/awssdk/services/cloudfront/url/SignedUrlTest.java
```diff
@@ -13,21 +13,27 @@
...
 
 import static org.assertj.core.api.Assertions.assertThat;
 
 import nl.jqno.equalsverifier.EqualsVerifier;
 import org.junit.jupiter.api.Test;
-import software.amazon.awssdk.services.cloudfront.internal.url.SignedUrl;
-import software.amazon.awssdk.services.cloudfront.url.DefaultSignedUrl;
+import software.amazon.awssdk.services.cloudfront.internal.url.DefaultSignedUrl;
 
 class SignedUrlTest {
```
cenedhryn, line 24: Can we add a test for generateHttpRequest()?

services/cloudfront/src/test/java/software/amazon/awssdk/services/cloudfront/cookie/SignedCookieTest.java
```diff
@@ -13,17 +13,14 @@
...
 import org.junit.jupiter.api.Test;
-import software.amazon.awssdk.services.cloudfront.cookie.DefaultCookiesForCannedPolicy;
-import software.amazon.awssdk.services.cloudfront.cookie.DefaultCookiesForCustomPolicy;
-import software.amazon.awssdk.services.cloudfront.internal.cookies.CookiesForCannedPolicy;
-import software.amazon.awssdk.services.cloudfront.internal.cookies.CookiesForCustomPolicy;
-import software.amazon.awssdk.services.cloudfront.internal.cookies.SignedCookie;
+import software.amazon.awssdk.services.cloudfront.internal.cookie.DefaultCookiesForCannedPolicy;
+import software.amazon.awssdk.services.cloudfront.internal.cookie.DefaultCookiesForCustomPolicy;
 
 class SignedCookieTest {
```
cenedhryn, line 25: Can we add a test for generateHttpRequest()?

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/internal/url/DefaultSignedUrl.java
```diff
@@ -0,0 +1,155 @@
...
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    public Builder toBuilder() {
+        return new Builder(this);
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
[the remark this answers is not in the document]
davidh44: moved below

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/url/SignedUrl.java
```diff
@@ -35,14 +46,4 @@ public interface SignedUrl extends ToCopyableBuilder<SignedUrl.Builder, SignedUr
      */
     SdkHttpRequest generateHttpRequest();
```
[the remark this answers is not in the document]
davidh44: renamed method and added documentation

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
davidh44: created class and moved previously private methods from CloudFrontUtilities

services/cloudfront/src/test/java/software/amazon/awssdk/services/cloudfront/url/SignedUrlTest.java
davidh44: test added

services/cloudfront/src/test/java/software/amazon/awssdk/services/cloudfront/cookie/SignedCookieTest.java
davidh44: test added

cenedhryn (approved):

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
```diff
@@ -0,0 +1,264 @@
...
+ */
+@Immutable
+@ThreadSafe
+@SdkPublicApi
+public final class CloudFrontUtilities {
+
+    private static final String CLOUDFRONT_NET = "cloudfront.net";
+
+    private CloudFrontUtilities() {
+    }
```
millems, line 69: Are intentionally not following the utilities pattern from S3, where things aren't static and they create CloudFrontUtilities directly, or indirectly from a CloudFrontClient?

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
```diff
@@ -0,0 +1,264 @@
...
+     * @return A signed URL that will permit access to a specific distribution
+     *         and S3 object.
+     */
+    public static SignedUrl getSignedUrlWithCannedPolicy(CloudFrontSignerRequest request) {
+        try {
+            String resourceUrl = request.resourceUrl();
+            String cannedPolicy = SigningUtils.buildCannedPolicy(resourceUrl, request.expirationDate());
+            byte[] signatureBytes = SigningUtils.signWithSha1Rsa(cannedPolicy.getBytes(UTF_8), request.privateKey());
+            String urlSafeSignature = SigningUtils.makeBytesUrlSafe(signatureBytes);
+            String protocol = resourceUrl.substring(0, resourceUrl.indexOf("://"));
```
millems, line 90: nit: reuse `resourceUrl.indexOf("://")` to avoid needing to do it twice

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
```diff
@@ -0,0 +1,264 @@
...
+            String domain = resourceUrl.substring(resourceUrl.indexOf("://") + 3, resourceUrl.indexOf(CLOUDFRONT_NET) + 14);
+            String encodedPath = resourceUrl.substring(resourceUrl.indexOf(CLOUDFRONT_NET) + 15)
```
millems, line 92: nit: reuse `resourceUrl.indexOf(CLOUDFRONT_NET)`

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
millems, line 90: Can we just do URI.create(resourceUrl) to extract the protocol, domain and path? That feels safer than trying to parse the URL ourselves.

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
```diff
@@ -0,0 +1,264 @@
...
+                   + (request.resourceUrl().indexOf('?') >= 0 ? "&" : "?")
+                   + "Expires=" + request.expirationDate().getEpochSecond()
```
millems, line 94: Nit: Our API for presigned URLs is signatureDuration instead of expiration, just a slight difference. Should we be consistent?

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
```diff
@@ -0,0 +1,264 @@
...
+     * "https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-creating-signed-url-custom-policy.html"
+     * >Creating a signed URL using a custom policy</a>.
+     *
+     * @param request
+     *            A CloudFrontSignerRequest configured with the following values:
+     *            resourceUrl, privateKey, keyPairId, expirationDate, activeDate, ipRange
+     * @return A signed URL that will permit access to distribution and S3
+     *         objects as specified in the policy document.
+     */
+    public static SignedUrl getSignedUrlWithCustomPolicy(CloudFrontSignerRequest request) {
```
millems, line 118: Should these take different request objects, and return different response objects? Using the same requests is potentially confusing since the type system doesn't give the caller any hints about what fields are required. Using the same response is okay for now, but if these methods ever need to return something different, we'll be trapped.

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
```diff
@@ -0,0 +1,264 @@
...
+                                                       Instant expirationDate,
+                                                       String limitToIpAddressCidr) {
+        if (expirationDate == null) {
+            throw SdkClientException.create("Expiration date must be provided to sign CloudFront URLs");
+        }
+        if (resourceUrl == null) {
+            resourceUrl = "*";
+        }
+        return SigningUtils.buildCustomPolicy(resourceUrl, activeDate, expirationDate, limitToIpAddressCidr);
+    }
```
millems, line 240: Should this be private?

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
```diff
@@ -0,0 +1,264 @@
...
+
+    /**
+     * Creates a private key from the file given, either in RSA private key
+     * (.pem) or pkcs8 (.der) format. Other formats will cause an exception to
+     * be thrown.
+     */
+    public static PrivateKey loadPrivateKey(File privateKeyFile) throws InvalidKeySpecException, IOException {
```
millems, line 247: Should this be private? Our API could take both privateKey(PrivateKey) and privateKey(Path) so that they don't have to find this loadPrivateKey method themselves when all they have is a file.

Also:
1. We shouldn't throw checked exceptions.
2. We should accept Path, not File
3. We should be clear that we're choosing pem/der based on the file name. Not anything more intelligent.

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/cookie/CookiesForCannedPolicy.java
```diff
@@ -0,0 +1,63 @@
...
+
+    /**
+     * Returns the expires key
+     */
+    String expiresKey();
+
+    /**
+     * Returns the expires value
+     */
+    String expiresValue();
```
millems, line 37: What are these? How should they be used?

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/cookie/CookiesForCannedPolicy.java
```diff
@@ -0,0 +1,63 @@
...
+
+        /**
+         * Configure the resource URL
+         */
+        Builder resourceUrl(String resourceUrl);
+
+        /**
+         * Configure the expiration value
+         */
+        Builder expires(String expires);
```
millems, line 60: These are different from the getters. Is this not a data object? That surprises me. Our input/output objects are usually just data.

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/cookie/SignedCookie.java
```diff
@@ -0,0 +1,79 @@
...
+@ThreadSafe
+@SdkPublicApi
+public interface SignedCookie {
+
+    String KEY_PAIR_ID_KEY = "CloudFront-Key-Pair-Id";
+    String SIGNATURE_KEY = "CloudFront-Signature";
+
+    /**
+     * Returns the key pair ID key
+     */
```
millems, line 36: These comments aren't useful, they just tell the caller what they can already tell from the method name. What is a key pair ID key? How should they use it? Anything special about it? Its encoding, perhaps?

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/cookie/SignedCookie.java
```diff
@@ -0,0 +1,79 @@
...
+
+    /**
+     * Returns the resource URL
+     */
+    String resourceUrl();
+
+    /**
+     * Generates an HTTP GET request that can be executed by an HTTP client to access the resource
+     */
+    SdkHttpRequest generateHttpGetRequest();
```
millems, line 62: We use "create" or "resolve" elsehwere as a method prefix when we're trying to hint that the getter is doing non-trivial work. Is one of those appropriate?

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/cookie/SignedCookie.java
```diff
@@ -0,0 +1,79 @@
...
+     * @return The cookie header value to pass into an HTTP request
+     */
+    String cookieHeaderValue(CookieType cookieType);
+
+    /**
+     * The types of CloudFront cookies
+     */
+    enum CookieType {
+        KEY_PAIR_ID, SIGNATURE, EXPIRES, POLICY
+    }
```
millems, line 77: Why this pattern instead of a getter for each type?

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
davidh44: Went with private constructor since there are no class fields (region, etc. not needed)
Currently just call the methods as such: CloudFrontUtilities.getSignedUrlWithCannedPolicy(request)
Open to changing

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
davidh44: not needed anymore, updated to URI as per below

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
davidh44: not needed anymore, updated to URI as per below

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
davidh44: agreed, updated to using URI

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
davidh44: Expiration keeps consistent with the CloudFront docs: https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-creating-signed-url-canned-policy.html
Open to changing

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
davidh44: The same request object makes more sense to me since a request for custom policy is just an extension of a request for canned policy, with the option to add active date and IP range (you could add either, both, or none).

Agree with having different response objects in case of future modifications, will split into two classes for URLs for canned and custom policies

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
davidh44: moved to internal SigningUtils class

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
davidh44: moved to internal SigningUtils class.
updated to throw Exception.
updated to Path.
updated javadoc.

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/cookie/CookiesForCannedPolicy.java
davidh44: these were in case user wanted to manually add/create headers, probably not needed since cookieHeaderValue does the job, will remove

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/cookie/CookiesForCannedPolicy.java
davidh44: these are essentially the getters (the values, not the keys)

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/cookie/SignedCookie.java
davidh44: will create getters for each type

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/cookie/SignedCookie.java
davidh44: updated to createHttpGetRequest()

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/cookie/SignedCookie.java
davidh44: updated with the more descriptive javadocs from CloudFrontSignerRequest

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
millems: My concern is that this differs from the way we do it elsewhere, and if we moved to doing it the normal way in the future, customers would get warnings when they call these methods (because you're invoking a static method as an instance method).

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
millems: I think it's fine to be inconsistent here. Some people will want duration, and some will want expiration. We should probably support both eventually, tbh, but that's not in scope here.

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
millems: In the builder, we could just have separate methods for each type?

```
public Builder privateKey(PrivateKey key) {
    this.privateKey = key;
    return this;
}

public Builder privateKey(Path keyFile) {
    this.privateKey = loadPrivateKey(keyFile);
    return this;
}
```

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
millems: More than anything, a different request object makes more sense to me. There can be a common base class, but you don't want people passing an activeDate and ipRange to `getSignedUrlWithCannedPolicy`

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
davidh44: adding create() method as discussed

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
davidh44: Will keep expiration/active

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
davidh44: I hear you, keeping one single response object, creating separate request objects for canned/custom

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
davidh44: added separate methods in request builder

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
```diff
@@ -0,0 +1,189 @@
...
+     * "https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-creating-signed-url-custom-policy.html"
+     * >Creating a signed URL using a custom policy</a>.
+     *
+     * @param request
+     *            A CloudFrontSignerRequest configured with the following values:
+     *            resourceUrl, privateKey, keyPairId, expirationDate, activeDate, ipRange
+     * @return A signed URL that will permit access to distribution and S3
+     *         objects as specified in the policy document.
+     */
+    public SignedUrl getSignedUrlWithCustomPolicy(CustomSignerRequest request) {
```
millems, line 112: Might be worth adding a `Consumer<CustomerSignerRequest.Builder>` overload as well. Our usability tests show that people like them: https://github.com/aws/aws-sdk-java-v2/blob/master/services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Utilities.java#L205

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/cookie/CookiesForCannedPolicy.java
millems: I more meant: why do the getters not match the setters? That's the standard for data objects from https://github.com/aws/aws-sdk-java-v2/blob/master/docs/design/ClientConfiguration.md

These input/output objects shouldn't have any logic in them.

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/cookie/SignedCookie.java
```diff
@@ -0,0 +1,55 @@
...
+    String resourceUrl();
+
+    /**
+     * Generates an HTTP GET request that can be executed by an HTTP client to access the resource
+     */
+    SdkHttpRequest createHttpGetRequest();
+
+    /**
+     * Returns the cookie signature header value
+     */
```
millems, line 47: Can we include more information in all of our getter docs that could be helpful. How should they use these values? Is there any specific format to these values that they should be aware of?

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
```diff
@@ -0,0 +1,189 @@
...
+@ThreadSafe
+@SdkPublicApi
+public final class CloudFrontUtilities {
+
+    private CloudFrontUtilities() {
+    }
+
+    // Used by low-level client
+    @SdkInternalApi
+    static CloudFrontUtilities create() {
```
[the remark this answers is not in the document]
davidh44: updated to public

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
```diff
@@ -0,0 +1,189 @@
...
+    /**
+     * Generate signed cookies that allows access to a specific distribution and
+     * resource path by applying access restrictions from a "canned" (simplified)
+     * policy document.
+     * For more information, see <a href=
+     * "https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-setting-signed-cookie-canned-policy.html"
+     * >Setting signed cookies using a canned policy</a>.
+     *
+     * @param request
+     *            A CloudFrontSignerRequest configured with the following values:
```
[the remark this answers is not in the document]
davidh44: updated and linked to CannedSignerRequest and CustomSignerRequest

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
davidh44: Added overloading methods that takes Consumer to all 4 methods

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/cookie/SignedCookie.java
davidh44: Added more details to the getters docs

millems (approved):
Great work!

PR #3540, opened 2022-11-08
Used retry hint from CRT before converting HTTPException to IOException.

JonathanHenson:
Used retry hint from CRT before converting HTTPException to IOException.

## Motivation and Context
Rather than blindly converting all CRT thrown Http exceptions to IO Exceptions, call the isRetryable() function to check that it should actually be retried.

## Testing
Added unit test to verify that the "it's not retryable" case correctly gets hit. Haven't run it yet as it relies on a CRT version update in maven central.

## Checklist

- [x ] I have read the [CONTRIBUTING](https://github.com/aws/aws-sdk-java-v2/blob/master/CONTRIBUTING.md) document

## License


http-clients/aws-crt-client/src/main/java/software/amazon/awssdk/http/crt/internal/response/CrtResponseAdapter.java
```diff
@@ -129,9 +129,16 @@ private void onSuccessfulResponseComplete(HttpStream stream) {
...
+
+        if (HttpClientConnection.isErrorRetryable(error)) {
+            // IOExceptions get retried, and if the CRT says this error is retryable,
+            // it's semantically an IOException anyway.
+            toThrow = new IOException(error);
+        }
+
+        responsePublisher.error(toThrow);
+        failResponseHandlerAndFuture(stream, toThrow);
     }
```
zoewangg, line 142: Do we need to apply the same logic here? https://github.com/aws/aws-sdk-java-v2/blob/master/http-clients/aws-crt-client/src/main/java/software/amazon/awssdk/http/crt/internal/CrtRequestExecutor.java#L75

http-clients/aws-crt-client/src/test/java/software/amazon/awssdk/http/crt/AwsCrtHttpClientSpiVerificationTest.java
```diff
@@ -132,6 +133,27 @@ public void requestFailed_connectionTimeout_shouldWrapException() {
         }
     }
 
+    @Test
+    public void requestFailed_notRetryable_shouldNotWrapException() {
```
zoewangg, line 137: Looks like this test failed on CodeBuild

```
java.lang.NullPointerException: RequestContentPublisher must not be null.
--
at software.amazon.awssdk.http.crt.AwsCrtHttpClientSpiVerificationTest.requestFailed_notRetryable_shouldNotWrapException(AwsCrtHttpClientSpiVerificationTest.java:151)

```

http-clients/aws-crt-client/src/main/java/software/amazon/awssdk/http/crt/internal/response/CrtResponseAdapter.java
JonathanHenson: I think this should be fine. Whatever exception gets thrown in connection startup time is going to be an IOException.

Line 93 of that same file, however, needs an update. Good catch!

http-clients/aws-crt-client/src/test/java/software/amazon/awssdk/http/crt/AwsCrtHttpClientSpiVerificationTest.java
JonathanHenson: Fixed with a mock class.

zoewangg:
Seems like there are some checkstyle errors. I'll merge it to a feature branch and will fix them

zoewangg:
The change has been merged to mainline via #3575

PR #3558, opened 2022-11-22
ProfileFileSupplier API changes, aggregating

dave-fn:
## Motivation and Context
Update the `ProfileFileSupplier` API to include methods for aggregating multiple `ProfileFile` objects into a single `ProfileFileSupplier`. This will mimic how multiple `ProfileFile` can be aggregated into another `ProfileFile` object.
This change is in preparation for adding reloading functionality to `DefaultCredentialsProvider`.

## Modifications
Revised 2 methods and added 4 methods to `ProfileFileSupplier`.

## Testing
Added unit tests.

## Types of changes

## Checklist

## License

### DIFF
--- a/.changes/next-release/feature-AWSSDKforJavav2-8c45016.json
+++ b/.changes/next-release/feature-AWSSDKforJavav2-8c45016.json
@@ -1,0 +1,6 @@
+{
+    "category": "AWS SDK for Java v2", 
+    "contributor": "", 
+    "type": "feature", 
+    "description": "Add methods to ProfileFileSupplier for aggregating"
+}
--- a/.changes/next-release/feature-AWSSDKforJavav2-8c45016.json
+++ b/.changes/next-release/feature-AWSSDKforJavav2-8c45016.json
@@ -1,6 +1,0 @@
-{
-    "category": "AWS SDK for Java v2", 
-    "contributor": "", 
-    "type": "feature", 
-    "description": "Add methods to ProfileFileSupplier for aggregating"
-}
--- a/core/profiles/src/main/java/software/amazon/awssdk/profiles/ProfileFileSupplier.java
+++ b/core/profiles/src/main/java/software/amazon/awssdk/profiles/ProfileFileSupplier.java
@@ -16,9 +16,9 @@
 package software.amazon.awssdk.profiles;
 
 import java.nio.file.Path;
+import java.util.Arrays;
 import java.util.Objects;
 import java.util.Optional;
-import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Supplier;
 import software.amazon.awssdk.annotations.SdkPublicApi;
 import software.amazon.awssdk.profiles.internal.ProfileFileRefresher;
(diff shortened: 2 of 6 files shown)


core/profiles/src/main/java/software/amazon/awssdk/profiles/ProfileFileSupplier.java
```diff
@@ -105,4 +177,64 @@ static ProfileFileSupplier wrapIntoNullableSupplier(ProfileFile profileFile) {
...
+                return currentValuesBySupplier.get(this);
+            }
+
+            @Override
+            public void close() {
+                currentValuesBySupplier.remove(this);
+                currentValuesBySupplier.keySet().forEach(ProfileFileSupplier::close);
+            }
+
+            private boolean updateCurrentValue(ProfileFileSupplier supplier, Consumer<ProfileFile> action) {
```
L-Applin, line 218: maybe `updateCurrentValue` might not be the most accurate name for this method? It seems to also check if the supplier provided a modified `ProfileFile`.

.changes/next-release/feature-AWSSDKforJavav2-8c45016.json
```diff
@@ -0,0 +1,6 @@
+{
+    "category": "AWS SDK for Java v2", 
+    "contributor": "", 
+    "type": "feature", 
+    "description": "Add methods to ProfileFileSupplier for aggregating"
```
dagnir, line 5: Since we haven't merged the APIs to mainline yet, you don't need a changelog entry for this

core/profiles/src/main/java/software/amazon/awssdk/profiles/ProfileFileSupplier.java
dave-fn: Agree. I'll have to think hard to try to come up with an appropriate name.

core/profiles/src/main/java/software/amazon/awssdk/profiles/ProfileFileSupplier.java
```diff
@@ -105,4 +177,64 @@ static ProfileFileSupplier wrapIntoNullableSupplier(ProfileFile profileFile) {
...
+                ProfileFile next = supplier.get();
+                ProfileFile current = currentValuesBySupplier.put(supplier, next);
+                action.accept(next);
+
+                return !Objects.equals(next, current);
```
dagnir, line 223: These checks seem redundant. The contract of the API is that `ProfileFileSupplier#get` should return the same object if it's not updated

core/profiles/src/main/java/software/amazon/awssdk/profiles/ProfileFileSupplier.java
dave-fn: Method has been removed as it was not needed as pointed out in other comment.

core/profiles/src/main/java/software/amazon/awssdk/profiles/ProfileFileSupplier.java
dave-fn: Agree. Entire method has been removed.

.changes/next-release/feature-AWSSDKforJavav2-8c45016.json
dave-fn: Removed.

core/profiles/src/main/java/software/amazon/awssdk/profiles/ProfileFileSupplier.java
```diff
@@ -95,14 +114,49 @@ static ProfileFileSupplier fixedProfileFile(ProfileFile profileFile) {
...
+
+            @Override
+            public void close() {
+                Arrays.stream(suppliers).forEach(ProfileFileSupplier::close);
+            }
+
+            private ProfileFile refreshAndGetCurrentAggregate(ProfileFile.Aggregator aggregator) {
+                ProfileFile next = aggregator.build();
+                if (isNewAggregateProfileFile(next)) {
+                    currentAggregateProfileFile = next;
```
L-Applin, line 149: Does that need to be thread safe? Do we need to use compare and set?

dave-fn: Used same pattern in other classes, for example `ProfileFileRefresher`. Interested to hear what @dagnir has to say, though.

dagnir: Yeah I think we do; we want want to ensure that `next` does not change between lines 148 and 149. Good eye @L-Applin!

If the CAS fails, I think it's fine since that would mean a different thread already updated it.

dave-fn: Changed type of reference. Does the `Object.equals()` check break atomicity? I wonder if this needs a synchronized codeblock.

dagnir (approved):

L-Applin (approved):

PR #3563, opened 2022-11-23
Added AsyncResponseBody.fromInputStream and AsyncRequestBody.forBlockingInputStream.

millems:
This allows streaming operation requests to be written to like an input stream.

Example usage with retry support, without blocking calling thread:

```java
String data = "Hello, world!";
ExecutorService executor = Executors.newCachedThreadPool();
InputStream data = new StringInputStream("Hello, world!");

s3.putObject(r -> r.bucket("foo").key("bar"),
             AsyncRequestBody.fromInputStream(data, data.length(), executor))
  .join();
```

Example usage without retry support, with blocking calling thread:
```java
String data = "Hello, world!";
BlockingInputStreamAsyncRequestBody body = AsyncRequestBody.forBlockingInputStream(data.length());
CompletableFuture<?> responseFuture = s3.putObject(r -> r.bucket("foo").key("bar"), body);
body.writeInputStream(new StringInputStream(data));
responseFuture.join();
```

Other changes made to support this change:
1. Added `InputStreamConsumingPublisher`, a `Publisher<ByteBuffer>` that can read from an `InputStream`.

These supporting changes were also made, and are also being reviewed in https://github.com/aws/aws-sdk-java-v2/pull/3565:
1. Updated `SimplePublisher` to return more useful and predictable error messages when invoked after `onComplete`, `onError` or `cancel` has been called.
1. Added `CompletableFutureUtils.joinInterruptibly`, which behaves equivalent to executing `join()`, but allows the calling thread to still respond to interrupts.
1. Added `CompletableFutureUtils.joinInterruptiblyIgnoringFailures`, which behaves equivalent to executing `join()`, but allows the calling thread to still respond to interrupts, and ignores exceptions that are thrown.

### DIFF
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/InputStreamWithExecutorAsyncRequestBody.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/InputStreamWithExecutorAsyncRequestBody.java
@@ -1,0 +1,127 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.core.internal.async;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.ByteBuffer;
+import java.util.Optional;
+import java.util.concurrent.CancellationException;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import org.reactivestreams.Subscriber;
+import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.annotations.SdkTestInternalApi;
+import software.amazon.awssdk.core.async.AsyncRequestBody;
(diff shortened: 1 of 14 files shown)


utils/src/main/java/software/amazon/awssdk/utils/async/SimplePublisher.java
```diff
@@ -289,31 +281,47 @@ private void doProcessQueue() {
...
+        }
+    }
+
+    private void trySetFailureMessage(Supplier<Throwable> supplier) {
+        if (failureMessageSupplier == null) {
+            failureMessageSupplier = supplier;
         }
     }
 
+    private boolean hasFailureMessage() {
```
joviegas, line 294: This method is  bit confusing since we are checking         ` failureMessageSupplier != null;` while we have failureMessage field

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/InputStreamWithExecutorAsyncRequestBody.java
```diff
@@ -0,0 +1,126 @@
...
+                s.onError(t);
+            }
+        }
+    }
+
+    private void tryReset(InputStream inputStream) {
+        try {
+            inputStream.reset();
+        } catch (IOException e) {
+            throw new UncheckedIOException("Request cannot be retried, because the request stream could not be reset.", e);
```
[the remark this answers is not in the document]
millems: The contract of AsyncRequestBody is related to handling during retries, so it's appropriate to talk about retrying in the error message. As for the exception type, I suppose we could use NonRetryableException to indicate that we shouldn't try again. 

I was trying to be safe here, since I didn't know if the reset failed for a retryable or non-retryable reason, but it's probably not retryable so it would probably be okay to short-circuit future attempts here.

utils/src/main/java/software/amazon/awssdk/utils/async/SimplePublisher.java
millems: I can see why that would be confusing. All of the methods relate to setting and checking the failure message. The implementation uses a supplier (failureMessageSupplier) and cached result (failureMessage), but that's an implementation detail of those methods that the caller shouldn't have to worry about, so I named the methods accordingly.

I can move those fields and methods into an inner class so that it's clear those fields are encapsulated to the failure-related methods.

joviegas (approved):

joviegas (approved):

PR #3565, opened 2022-11-23
Added AsyncRequestBody.forBlockingOutputStream

millems:
This allows streaming operation requests to be written to like an output stream.

Example usage:
```java
BlockingOutputStreamAsyncRequestBody body = AsyncRequestBody.forBlockingOutputStream(5L);
CompletableFuture<?> responseFuture =
    client.streamingInputOperation(StreamingInputOperationRequest.builder().build(), body);

try (OutputStream stream = body.outputStream()) { // TODO: Test writing too much data and too little data
    stream.write("Hello".getBytes(StandardCharsets.UTF_8));
}
responseFuture.join();
```

Other changes made to support this change:
1. Added `OutputStreamPublisher` to utils. This is a `Publisher<ByteBuffer>` that can be written to like it's an output stream. Writes will block when there is no available downstream demand.
1. Added `CancellableOutputStream` to utils. This is an output stream that can be cancelled. This is useful for cancelling the upload via the output stream.

The following changes appear in this PR as well as https://github.com/aws/aws-sdk-java-v2/pull/3563:
1. Updated `SimplePublisher` to return more useful and predictable error messages when invoked after `onComplete`, `onError` or `cancel` has been called.
1. Added `CompletableFutureUtils.joinInterruptibly`, which behaves equivalent to executing `join()`, but allows the calling thread to still respond to interrupts.
1. Added `CompletableFutureUtils.joinInterruptiblyIgnoringFailures`, which behaves equivalent to executing `join()`, but allows the calling thread to still respond to interrupts, and ignores exceptions that are thrown.

### DIFF
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/async/AsyncRequestBody.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/async/AsyncRequestBody.java
@@ -169,9 +169,9 @@ public interface AsyncRequestBody extends SdkPublisher<ByteBuffer> {
      * complete.
      *
      * <p><b>Example Usage</b>
-     *
-     * <pre>
-     *     S3Client s3 = ...;
+     * <p>
+     * {@snippet :
+     *     S3AsyncClient s3 = S3AsyncClient.create(); // Use one client for your whole application!
      *
      *     byte[] dataToSend = "Hello".getBytes(StandardCharsets.UTF_8);
      *     long lengthOfDataToSend = dataToSend.length();
@@ -189,7 +189,7 @@ public interface AsyncRequestBody extends SdkPublisher<ByteBuffer> {
      *
      *     // Wait for the service to respond.
      *     PutObjectResponse response = responseFuture.join();
-     * </pre>
+     * }
      */
     static BlockingOutputStreamAsyncRequestBody forBlockingOutputStream(Long contentLength) {
         return new BlockingOutputStreamAsyncRequestBody(contentLength);
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/async/AsyncRequestBody.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/async/AsyncRequestBody.java
@@ -16,6 +16,7 @@
 package software.amazon.awssdk.core.async;
 
 import java.io.File;
+import java.io.OutputStream;
 import java.nio.ByteBuffer;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
@@ -158,6 +159,40 @@ public interface AsyncRequestBody extends SdkPublisher<ByteBuffer> {
      */
     static AsyncRequestBody fromByteBuffer(ByteBuffer byteBuffer) {
         return fromBytes(BinaryUtils.copyAllBytesFrom(byteBuffer));
+    }
+
+    /**
+     * Creates a {@link BlockingOutputStreamAsyncRequestBody} to use for writing to the downstream service as if it's an output
+     * stream. Retries are not supported for this request body.
+     *
+     * <p>The caller is responsible for calling {@link OutputStream#close()} on the {@link #outputStream()} when writing is
+     * complete.
+     *
+     * <p><b>Example Usage</b>
+     *
+     * <pre>
+     *     S3Client s3 = ...;
+     *
+     *     byte[] dataToSend = "Hello".getBytes(StandardCharsets.UTF_8);
+     *     long lengthOfDataToSend = dataToSend.length();
+     *
+     *     // Start the operation
+     *     BlockingInputStreamAsyncRequestBody body =
+     *         AsyncRequestBody.forBlockingOutputStream(lengthOfDataToSend);
+     *     CompletableFuture<PutObjectResponse> responseFuture =
+     *         s3.putObject(r -> r.bucket("bucketName").key("key"), body);
+     *
+     *     // Write the input stream to the running operation
+     *     try (CancellableOutputStream outputStream = body.outputStream()) {
+     *         outputStream.write(dataToSend);
+     *     }
+     *
+     *     // Wait for the service to respond.
+     *     PutObjectResponse response = responseFuture.join();
+     * </pre>
+     */
+    static BlockingOutputStreamAsyncRequestBody forBlockingOutputStream(Long contentLength) {
+        return new BlockingOutputStreamAsyncRequestBody(contentLength);
     }
 
     /**
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/async/BlockingOutputStreamAsyncRequestBody.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/async/BlockingOutputStreamAsyncRequestBody.java
@@ -1,0 +1,102 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.core.async;
+
+import java.io.OutputStream;
+import java.nio.ByteBuffer;
+import java.time.Duration;
+import java.util.Optional;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import org.reactivestreams.Subscriber;
+import software.amazon.awssdk.annotations.SdkPublicApi;
+import software.amazon.awssdk.core.exception.NonRetryableException;
+import software.amazon.awssdk.core.internal.util.NoopSubscription;
+import software.amazon.awssdk.utils.CancellableOutputStream;
+import software.amazon.awssdk.utils.async.OutputStreamPublisher;
+
+/**
+ * An implementation of {@link AsyncRequestBody} that allows performing a blocking write of an output stream to a downstream
+ * service.
+ *
+ * <p>The caller is responsible for calling {@link OutputStream#close()} on the {@link #outputStream()} when writing is
+ * complete.
+ *
+ * @see AsyncRequestBody#forBlockingOutputStream(Long)
+ */
+@SdkPublicApi
+public final class BlockingOutputStreamAsyncRequestBody implements AsyncRequestBody {
+    private final OutputStreamPublisher delegate = new OutputStreamPublisher();
+    private final CountDownLatch subscribedLatch = new CountDownLatch(1);
+    private final AtomicBoolean subscribeCalled = new AtomicBoolean(false);
+    private final Long contentLength;
+    private final Duration subscribeTimeout;
+
+    BlockingOutputStreamAsyncRequestBody(Long contentLength) {
+        this(contentLength, Duration.ofSeconds(10));
+    }
+
+    BlockingOutputStreamAsyncRequestBody(Long contentLength, Duration subscribeTimeout) {
+        this.contentLength = contentLength;
+        this.subscribeTimeout = subscribeTimeout;
+    }
+
+    /**
+     * Return an output stream to which blocking writes can be made to the downstream service.
+     *
+     * <p>This method will block the calling thread until the SDK is connected to the service. This means that this request body
+     * should usually be passed to the SDK before this method is called.
+     *
+     * <p>You can invoke {@link CancellableOutputStream#cancel()} to cancel any blocked write calls to the downstream service
+     * (and mark the stream as failed).
+     */
+    public CancellableOutputStream outputStream() {
+        waitForSubscriptionIfNeeded();
+        return delegate;
+    }
+
+    @Override
+    public Optional<Long> contentLength() {
+        return Optional.ofNullable(contentLength);
+    }
+
+    @Override
+    public void subscribe(Subscriber<? super ByteBuffer> s) {
+        if (subscribeCalled.compareAndSet(false, true)) {
+            delegate.subscribe(s);
+            subscribedLatch.countDown();
+        } else {
+            s.onSubscribe(new NoopSubscription(s));
+            s.onError(NonRetryableException.create("A retry was attempted, but AsyncRequestBody.forBlockingOutputStream does not "
+                                                   + "support retries."));
+        }
+    }
+
+    private void waitForSubscriptionIfNeeded() {
+        try {
+            long timeoutSeconds = subscribeTimeout.getSeconds();
+            if (!subscribedLatch.await(timeoutSeconds, TimeUnit.SECONDS)) {
+                throw new IllegalStateException("The service request was not made within " + timeoutSeconds + " seconds of "
+                                                + "outputStream being invoked. Make sure to invoke the service request "
+                                                + "BEFORE invoking outputStream if your caller is single-threaded.");
+            }
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            throw new RuntimeException("Interrupted while waiting for subscription.", e);
+        }
+    }
+}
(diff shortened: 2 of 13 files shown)


core/sdk-core/src/main/java/software/amazon/awssdk/core/async/AsyncRequestBody.java
```diff
@@ -160,6 +160,37 @@ static AsyncRequestBody fromByteBuffer(ByteBuffer byteBuffer) {
         return fromBytes(BinaryUtils.copyAllBytesFrom(byteBuffer));
     }
 
+    /**
+     * Creates a {@link BlockingOutputStreamAsyncRequestBody} to use for writing to the downstream service as if it's an output
+     * stream.
+     *
+     * <p><b>Example Usage</b>
+     *
+     * <pre>
```
zoewangg, line 169: Nit, can we use snippet? :) 
```
{@snippet :
  
}
```

core/sdk-core/src/main/java/software/amazon/awssdk/core/async/AsyncRequestBody.java
```diff
@@ -160,6 +160,37 @@ static AsyncRequestBody fromByteBuffer(ByteBuffer byteBuffer) {
...
+     *     S3Client s3 = ...;
```
zoewangg, line 170: `S3Client.create()`?

core/sdk-core/src/main/java/software/amazon/awssdk/core/async/BlockingOutputStreamAsyncRequestBody.java
```diff
@@ -0,0 +1,90 @@
...
+import software.amazon.awssdk.utils.async.OutputStreamPublisher;
+
+/**
+ * An implementation of {@link AsyncRequestBody} that allows performing a blocking write of an output stream to a downstream
+ * service.
+ *
+ * <p>See {@link AsyncRequestBody#forBlockingOutputStream(Long)}.
+ */
+@SdkPublicApi
+public class BlockingOutputStreamAsyncRequestBody implements AsyncRequestBody {
```
zoewangg, line 37: `final`

core/sdk-core/src/main/java/software/amazon/awssdk/core/async/BlockingOutputStreamAsyncRequestBody.java
```diff
@@ -0,0 +1,90 @@
...
+import software.amazon.awssdk.core.exception.NonRetryableException;
+import software.amazon.awssdk.core.internal.util.NoopSubscription;
+import software.amazon.awssdk.utils.CancellableOutputStream;
+import software.amazon.awssdk.utils.async.OutputStreamPublisher;
+
+/**
+ * An implementation of {@link AsyncRequestBody} that allows performing a blocking write of an output stream to a downstream
+ * service.
+ *
+ * <p>See {@link AsyncRequestBody#forBlockingOutputStream(Long)}.
```
zoewangg, line 34: nit: `@see AsyncRequestBody#forBlockingOutputStream(Long)`

utils/src/main/java/software/amazon/awssdk/utils/async/OutputStreamPublisher.java
```diff
@@ -0,0 +1,120 @@
...
+import software.amazon.awssdk.utils.CancellableOutputStream;
+import software.amazon.awssdk.utils.Validate;
+
+/**
+ * Adapts a {@link Publisher} to an {@link OutputStream}.
+ * <p>
+ * Writes to the stream will block until demand is available in the downstream subscriber.
+ */
+@SdkProtectedApi
+public class OutputStreamPublisher extends CancellableOutputStream implements Publisher<ByteBuffer> {
```
zoewangg, line 38: `final`

core/sdk-core/src/main/java/software/amazon/awssdk/core/async/AsyncRequestBody.java
```diff
@@ -160,6 +160,37 @@ static AsyncRequestBody fromByteBuffer(ByteBuffer byteBuffer) {
...
+     *     // Write the input stream to the running operation
+     *     try (CancellableOutputStream outputStream = body.outputStream()) {
+     *         outputStream.write(dataToSend);
+     *     }
+     *
+     *     // Wait for the service to respond.
+     *     PutObjectResponse response = responseFuture.join();
+     * </pre>
+     */
+    static BlockingOutputStreamAsyncRequestBody forBlockingOutputStream(Long contentLength) {
```
zoewangg, line 190: `toBlockingOutputStream`?

core/sdk-core/src/main/java/software/amazon/awssdk/core/async/BlockingOutputStreamAsyncRequestBody.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    @Override
+    public void subscribe(Subscriber<? super ByteBuffer> s) {
+        if (subscribeCalled.compareAndSet(false, true)) {
+            delegate.subscribe(s);
+            subscribedLatch.countDown();
+        } else {
+            s.onSubscribe(new NoopSubscription(s));
+            s.onError(NonRetryableException.create("A retry was attempted, but AsyncRequestBody.forBlockingOutputStream does not "
```
zoewangg, line 73: Can we mention retry is not supported in the Javadoc?

test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/BlockingAsyncRequestResponseBodyTest.java
```diff
@@ -86,4 +98,107 @@ public void blockingResponseTransformer_closeDoesNotThrow() throws IOException {
         ResponseInputStream<StreamingOutputOperationResponse> responseStream = responseFuture.join();
         responseStream.close();
     }
+
+    public void blockingOutputStreamWithoutExecutor_sendsRightValues() throws IOException {
```
zoewangg, line 102: Missing `@Test`?

core/sdk-core/src/main/java/software/amazon/awssdk/core/async/BlockingOutputStreamAsyncRequestBody.java
```diff
@@ -0,0 +1,90 @@
...
+/**
+ * An implementation of {@link AsyncRequestBody} that allows performing a blocking write of an output stream to a downstream
+ * service.
+ *
+ * <p>See {@link AsyncRequestBody#forBlockingOutputStream(Long)}.
+ */
+@SdkPublicApi
+public class BlockingOutputStreamAsyncRequestBody implements AsyncRequestBody {
+    private final OutputStreamPublisher delegate = new OutputStreamPublisher();
+    private final CountDownLatch subscribedLatch = new CountDownLatch(0);
```
zoewangg, line 39: Shouldn't it be `new CountDownLatch(1)`?

utils/src/main/java/software/amazon/awssdk/utils/async/OutputStreamPublisher.java
```diff
@@ -0,0 +1,120 @@
...
+    private final SimplePublisher<ByteBuffer> delegate = new SimplePublisher<>();
+    private final CountDownLatch subscribedLatch = new CountDownLatch(0);
```
zoewangg, line 40: Same here, shouldn't it be `new CountDownLatch(1)`?

zoewangg (commented):
Left some initial comments. Still going through the PR

core/sdk-core/src/main/java/software/amazon/awssdk/core/async/AsyncRequestBody.java
millems: Have you merged the PR that enables snippets yet? I searched around and we weren't using snippets anywhere else in master.

core/sdk-core/src/main/java/software/amazon/awssdk/core/async/AsyncRequestBody.java
millems: I didn't want people to copy and paste, since they would need to close the client. I'll have to fix it with snippets, though.

core/sdk-core/src/main/java/software/amazon/awssdk/core/async/AsyncRequestBody.java
millems: I was trying to differentiate between the normal to* methods that you can fire-and-forget into a client and these methods that required you to interact with the AsyncRequestBody directly. Is there a better way?

test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/BlockingAsyncRequestResponseBodyTest.java
millems: Whoops, good catch. Lost it with all the rebasing involved splitting this into 3 commits.

core/sdk-core/src/main/java/software/amazon/awssdk/core/async/BlockingOutputStreamAsyncRequestBody.java
millems: Oh yeah, guess my testing isn't sufficient. Will fix.

utils/src/main/java/software/amazon/awssdk/utils/async/OutputStreamPublisher.java
millems: Guess my testing isn't sufficient. Will fix. Good catch.

core/sdk-core/src/main/java/software/amazon/awssdk/core/async/BlockingOutputStreamAsyncRequestBody.java
millems: Fixed.

core/sdk-core/src/main/java/software/amazon/awssdk/core/async/AsyncRequestBody.java
zoewangg: Inline snippets should be supported now. I haven't merged the change to enable external snippets.

utils/src/main/java/software/amazon/awssdk/utils/async/SimplePublisher.java
```diff
@@ -185,9 +171,7 @@ public CompletableFuture<Void> error(Throwable error) {
         OnErrorQueueEntry<T> entry = new OnErrorQueueEntry<>(error);
 
         try {
-            validateRejectState();
-            setRejectExceptionOrThrow(() -> new IllegalStateException("error() has been invoked"));
-            eventQueue.add(entry);
+            standardPriorityQueue.add(entry);
```
zoewangg, line 174: Should it be `highPriorityQueue.add(entry)`?

utils/src/main/java/software/amazon/awssdk/utils/async/SimplePublisher.java
```diff
@@ -72,27 +69,19 @@
...
 
     /**
-     * When processing the {@link #eventQueue}, these are the entries that should be skipped (and failed). This is used to
-     * safely drain the queue when there are urgent events needing processing, like a {@link Subscription#cancel()}.
-     */
-    private final Set<QueueEntry.Type> entryTypesToFail = new CopyOnWriteArraySet<>();
-
-    /**
-     * Whether the {@link #eventQueue} is currently being processed. Only one thread may read events from the queue at a time.
+     * Whether the {@link #standardPriorityQueue} is currently being processed. Only one thread may read events from the
```
zoewangg, line 76: Nit: should it be `Whether the {@link #standardPriorityQueue} or {@link #highPriorityQueue} is currently being processed.`?

utils/src/main/java/software/amazon/awssdk/utils/async/SimplePublisher.java
```diff
@@ -401,16 +373,39 @@ public void cancel() {
...
     }
 
     /**
-     * An entry in the {@link #eventQueue}.
+     * A lazily-initialized failure message for future events sent to this publisher after a terminal event has
+     * occurred.
+     */
+    private static final class FailureMessage {
+        private Supplier<Throwable> failureMessageSupplier;
+        private Throwable failureMessage;
```
zoewangg, line 387: Can they be set by different threads? My guess is no, but wanted to double check

core/sdk-core/src/main/java/software/amazon/awssdk/core/async/AsyncRequestBody.java
zoewangg: Naming is hard. I can't think of better alternatives.

zoewangg:
Should we add a couple of integ tests using S3 to transfer a relatively large large object (maybe 4MB)? Existing functional tests do not seem to cover larger payload.
Have we compared the performance between blockingOutputstream using async client and outputStream using sync client?

utils/src/main/java/software/amazon/awssdk/utils/async/SimplePublisher.java
```diff
@@ -289,31 +283,30 @@ private void doProcessQueue() {
...
 
-        if (entry.type() != ON_NEXT) {
-            // This event isn't an on-next event, so we don't need subscriber demand in order to process it.
+        if (failureMessage.isSet()) {
             return true;
         }
 
-        if (entryTypesToFail.contains(ON_NEXT)) {
-            // This is an on-next call (decided above), but we're failing on-next calls. Go ahead and fail it.
+        if (subscriber == null) {
```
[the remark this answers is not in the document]
millems: Those two checks could be in either order, since failureMessage should never be set while we don't have a subscriber.

utils/src/main/java/software/amazon/awssdk/utils/async/SimplePublisher.java
millems: Hmm... it could be, but that's not what the design is. The javadoc implies it should be standardPriorityQueue as well. It's the difference of whether we want to preserve the relative order of send() and error() calls. We could change the semantics of the API to have error() calls skip previous send() calls, but then the caller wouldn't know the order of messages the downstream subscriber receives when errors are involved.

If we really wanted a "no really, die immediately", we could add a high-priority `error()` method. I don't know if we'd use it.

utils/src/main/java/software/amazon/awssdk/utils/async/SimplePublisher.java
millems: This is complicated. 

Yes, different threads read and write this value, but that's okay.

It's guaranteed that every write/read happens while holding the processingQueue flag, except one read we make in processEventQueue when it calls shouldProcessQueueEntry. It's okay that a thread might read a stale value there, because it will only happen when there's another thread currently holding the processingQueue flag, and that thread will re-check the value when it releases the flag.

millems:
> Should we add a couple of integ tests using S3 to transfer a relatively large large object (maybe 4MB)? Existing functional tests do not seem to cover larger payload. 

Sure!

> Have we compared the performance between blockingOutputstream using async client and outputStream using sync client?

No, but sync doesn't support OutputStream. We could compare it against InputStream, but I suspect sync will be faster because the data doesn't have to make the thread jump to the event loop. Do you think we should do that testing?

utils/src/main/java/software/amazon/awssdk/utils/async/SimplePublisher.java
zoewangg: I see, looks like in the existing implementation, it does not skip other data events if there's an error event. I think I missed/forgot that part. Can we add some documentation for `highPriorityQueue`?

zoewangg:
> Do you think we should do that testing?
Yeah, I also expect sync to be faster. It'd be great if we can get some data so that we can provide recommendations to customer accordingly.

zoewangg (approved):

T-TRz879:
```
String data = "Hello, world!";
ExecutorService executor = Executors.newCachedThreadPool();
InputStream data = new StringInputStream("Hello, world!");

s3.putObject(r -> r.bucket("foo").key("bar"),
             AsyncRequestBody.fromInputStream(data, data.length(), executor))
  .join();
```
I would like to inquire about the issue of the flow being able to be reset in the above code.
Now there is a stream coming from somewhere else, and I will read out the content of each segment and put it into a `ByteBuf`. I need to encrypt and other operations on the data in this `ByteBuf`. After the data in ByteBuf is processed, I will write it into the data stream above and release it. Can the data stream be reset at this time
@millems @zoewangg

PR #3579, opened 2022-12-01
Validate signer client options based on authtype

dagnir:
## Motivation and Context
This fixes an issue with the client option validation where the client builder is incorrectly validating that a value for the SIGNER option is present, even when it's not required (because the service does not use AWS auth).

## Modifications
This change only does the appropriate validations within the builder based on the authtypes present in the service model.

## Testing
- New  unit tests

## Types of changes

## Checklist

## License

### DIFF
--- a/codegen/src/main/java/software/amazon/awssdk/codegen/poet/builder/BaseClientBuilderClass.java
+++ b/codegen/src/main/java/software/amazon/awssdk/codegen/poet/builder/BaseClientBuilderClass.java
@@ -48,7 +48,7 @@ import software.amazon.awssdk.codegen.model.service.ClientContextParam;
 import software.amazon.awssdk.codegen.poet.ClassSpec;
 import software.amazon.awssdk.codegen.poet.PoetUtils;
 import software.amazon.awssdk.codegen.poet.rules.EndpointRulesSpecUtils;
-import software.amazon.awssdk.codegen.utils.BearerAuthUtils;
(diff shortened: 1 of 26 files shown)


dave-fn:
No issues at all, Looks good to me.

dave-fn (commented):
No issues at all, Looks good to me.

codegen/src/main/java/software/amazon/awssdk/codegen/poet/builder/BaseClientBuilderClass.java
```diff
@@ -623,4 +625,33 @@ private boolean hasClientContextParams() {
...
+                                               .returns(void.class);
+
+        if (AuthUtils.usesAwsAuth(model)) {
+            builder.addStatement("$T.notNull(c.option($T.SIGNER), $S)",
+                                 Validate.class,
+                                 SdkAdvancedClientOption.class,
+                                 "The 'overrideConfiguration.advancedOption[SIGNER]' must be configured in the client builder.");
+        }
+
+        if (AuthUtils.usesBearerAuth(model)) {
```
zoewangg, line 642: Should it be `else if`? Can they use both aws auth and bearer auth?

dagnir: Yes they can. For example, the service `signatureVersion` has `v4` but also includes one operation that uses `bearer`.

zoewangg (approved):

PR #3581, opened 2022-12-01
Use ByteBufferStoringSubscriber

zoewangg:
## Motivation and Context
Use ByteBufferStoringSubscriber in Transfer Manager

Tested it locally and found it has better performance than the existing subscriber implementation


services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/S3CrtRequestBodyStreamAdapter.java
```diff
@@ -16,323 +16,33 @@
...
-                    throw new RuntimeException(t);
-
-                default:
-                    // In case new event types are introduced that this loop doesn't account for
-                    throw new IllegalStateException("Unknown event type: " + ev.type());
-            }
-        }
-
-        return false;
+        return requestBodySubscriber.blockingTransferTo(outBuffer) == ByteBufferStoringSubscriber.TransferResult.END_OF_STREAM;
```
millems, line 41: What thread is blocking here?

zoewangg: As discussed offline, CRT thread is blocking here; it requires the byte buffer to be completely filled. Added a comment

millems: I don't like that. Seems to scale poorly...

millems (approved):

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/S3CrtRequestBodyStreamAdapter.java
zoewangg: +1 I don't like it either. This will for sure cause issues with larger part size. We are having conversations with CRT team about this

PR #3586, opened 2022-12-03
Throw UnsupportedOperationException if a user tries to pause a upload…

zoewangg:
… with non CRT-based S3 client

## Motivation and Context
Throw UnsupportedOperationException if a user tries to pause a upload with non CRT-based S3 client to instead of NPE


L-Applin:
The [pause](https://github.com/aws/aws-sdk-java-v2/blob/feature/master/transfermanager-GA/services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/model/FileUpload.java#L38) documentation should probably mention in some way that the operation could be not supported by some implementation.

zoewangg:
> The [pause](https://github.com/aws/aws-sdk-java-v2/blob/feature/master/transfermanager-GA/services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/model/FileUpload.java#L38) documentation should probably mention in some way that the operation could be not supported by some implementation.

Good point. I'll add it as part of #3592

cenedhryn (approved):

PR #3487, opened 2022-10-13
ProfileCredentialsProvider can now reload credentials when profile files change

dave-fn:
## Motivation and Context

The ProfileCredentialsProvider loads once and doesn't refresh/reload if the ProfileFile it was generated from got updated, which is unlike V1, in which the ProfileCredentialsProvider has support for this try to reload every Duration X (customizable) if the profile file it was generated from have a newer last modified date.

Should fix [#1754](https://github.com/aws/aws-sdk-java-v2/issues/1754), similar to [PR #1760](https://github.com/aws/aws-sdk-java-v2/pull/1760).
Also addresses [#1327](https://github.com/aws/aws-sdk-java-v2/issues/1327).

## Modifications

Modified `ProfileFile` so that is has both the ability to detect if there are disk updates and to reload itself as a new instance. The included modifications are:
  - Created `Builder.BuildDetails` as a container of build information. `ProfileFile` did not keep track of the `Path` or `InputStream` used to generate it.
  - Add methods for finding out if the instance is stale, as determined by the associated file modification timestamp.
  - Added `reload` method to return a new instance when the file has been modified, or the same instance otherwise.

Modified `ProfileCredentialsProvider` for automatically reloading credentials.
  - Two instance variables are no longer immutable, so the type has been changed to `AtomicReference` objects.
  - A `CachedSupplier` is used to avoid hitting the filesystem whenever `resolveCredentials` is called.
  - Updated the `ProfileCredentialsProvider.Builder` interface for setting credentials refreshing intervals.

## Testing

Unit tested

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

## Checklist

- [ ] I have read the [CONTRIBUTING](https://github.com/aws/aws-sdk-java-v2/blob/master/CONTRIBUTING.md) document
- [x] Local run of `mvn install` succeeds
- [x] My code follows the code style of this project
- [ ] My change requires a change to the Javadoc documentation
- [x] I have updated the Javadoc documentation accordingly
- [x] I have added tests to cover my changes
- [x] All new and existing tests passed
- [x] I have added a changelog entry. Adding a new entry must be accomplished by running the `scripts/new-change` script and following the instructions. Commit the new file created by the script in `.changes/next-release` with your changes.
- [ ] My change is to implement 1.11 parity feature and I have updated [LaunchChangelog](https://github.com/aws/aws-sdk-java-v2/blob/master/docs/LaunchChangelog.md)

## License

[...]

PR #3608, opened 2022-12-12
ProfileTokenProvider Reload ProfileFile

dave-fn:
## Motivation and Context
Update `ProfileTokenProvider` so that the generated token uses the current contents of a `ProfileFile` if desired. This change mimics what was done with [PR 3487](https://github.com/aws/aws-sdk-java-v2/pull/3487).

## Modifications
Mirrored new behavior of `ProfileCredentialsProvider`.

## Testing
Added single test case.

## Types of changes

## Checklist

## License

### DIFF
--- a/core/auth/src/main/java/software/amazon/awssdk/auth/token/credentials/ProfileTokenProvider.java
+++ b/core/auth/src/main/java/software/amazon/awssdk/auth/token/credentials/ProfileTokenProvider.java
@@ -15,6 +15,7 @@
 
 package software.amazon.awssdk.auth.token.credentials;
 
+import java.util.Objects;
 import java.util.Optional;
 import java.util.function.Supplier;
 import software.amazon.awssdk.annotations.SdkPublicApi;
@@ -37,57 +38,35 @@ import software.amazon.awssdk.utils.ToString;
  */
 @SdkPublicApi
 public final class ProfileTokenProvider implements SdkTokenProvider, SdkAutoCloseable {
-    private final SdkTokenProvider tokenProvider;
+    private SdkTokenProvider tokenProvider;
     private final RuntimeException loadException;
-
-    private final ProfileFile profileFile;
+    private final Supplier<ProfileFile> profileFile;
+    private volatile ProfileFile currentProfileFile;
     private final String profileName;
 
     /**
      * @see #builder()
      */
     private ProfileTokenProvider(BuilderImpl builder) {
-        SdkTokenProvider tokenProvider = null;
-        RuntimeException loadException = null;
-        ProfileFile profileFile = null;
-        String profileName = null;
+        RuntimeException thrownException = null;
+        Supplier<ProfileFile> selectedProfileFileSupplier = null;
+        String selectedProfileName = null;
 
         try {
-            profileName = builder.profileName != null ? builder.profileName
-                                                      : ProfileFileSystemSetting.AWS_PROFILE.getStringValueOrThrow();
+            selectedProfileName = Optional.ofNullable(builder.profileName)
+                                          .orElseGet(ProfileFileSystemSetting.AWS_PROFILE::getStringValueOrThrow);
 
-            // Load the profiles file
-            profileFile = Optional.ofNullable(builder.profileFile)
-                                  .orElse(builder.defaultProfileFileLoader)
-                                  .get();
+            selectedProfileFileSupplier = Optional.ofNullable(builder.profileFile)
+                                                  .orElse(builder.defaultProfileFileLoader);
 
-            // Load the profile and token provider
-            String finalProfileName = profileName;
-            ProfileFile finalProfileFile = profileFile;
-            tokenProvider =
-                    profileFile.profile(profileName)
-                               .flatMap(p -> new ProfileTokenProviderLoader(finalProfileFile, p).tokenProvider())
-                               .orElseThrow(() -> {
-                                   String errorMessage = String.format("Profile file contained no information for " +
-                                                                       "profile '%s': %s", finalProfileName, finalProfileFile);
-                                   return SdkClientException.builder().message(errorMessage).build();
-                               });
         } catch (RuntimeException e) {
             // If we couldn't load the provider for some reason, save an exception describing why.
-            loadException = e;
+            thrownException = e;
         }
 
-        if (loadException != null) {
-            this.loadException = loadException;
-            this.tokenProvider = null;
-            this.profileFile = null;
-            this.profileName = null;
-        } else {
-            this.loadException = null;
-            this.tokenProvider = tokenProvider;
-            this.profileFile = profileFile;
-            this.profileName = profileName;
-        }
+        this.loadException = thrownException;
+        this.profileFile = selectedProfileFileSupplier;
+        this.profileName = selectedProfileName;
     }
 
     /**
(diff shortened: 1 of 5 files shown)


core/auth/src/main/java/software/amazon/awssdk/auth/token/credentials/ProfileTokenProvider.java
```diff
@@ -37,57 +38,35 @@
...
-            ProfileFile finalProfileFile = profileFile;
-            tokenProvider =
-                    profileFile.profile(profileName)
-                               .flatMap(p -> new ProfileTokenProviderLoader(finalProfileFile, p).tokenProvider())
-                               .orElseThrow(() -> {
-                                   String errorMessage = String.format("Profile file contained no information for " +
-                                                                       "profile '%s': %s", finalProfileName, finalProfileFile);
-                                   return SdkClientException.builder().message(errorMessage).build();
-                               });
+            selectedProfileName = Optional.ofNullable(builder.profileName)
```
L-Applin, line 56: NIT: I don't know what you prefer, but there is `Validate.getOrDefault(T param, Supplier<T> defaultValue)` in the `utils` module that does exactly the same logic without involving `Optional`.

core/auth/src/main/java/software/amazon/awssdk/auth/token/credentials/ProfileTokenProvider.java
```diff
@@ -137,6 +123,29 @@ public void close() {
...
+                          .flatMap(p -> new ProfileTokenProviderLoader(profileFile, p).tokenProvider())
+                          .orElseThrow(() -> {
+                              String errorMessage = String.format("Profile file contained no information for " +
+                                                                  "profile '%s': %s", profileName, profileFile);
+                              return SdkClientException.builder().message(errorMessage).build();
+                          });
+    }
+
+    private void handleProfileFileReload(ProfileFile profileFile) {
+        tokenProvider = createTokenProvider(profileFile, profileName);
```
L-Applin, line 138: Does this need to be synchronized?

core/auth/src/main/java/software/amazon/awssdk/auth/token/credentials/ProfileTokenProvider.java
```diff
@@ -120,14 +99,21 @@ public SdkToken resolveToken() {
         if (loadException != null) {
             throw loadException;
         }
+
+        ProfileFile cachedOrRefreshedProfileFile = refreshProfileFile();
+        if (isNewProfileFile(cachedOrRefreshedProfileFile)) {
+            currentProfileFile = cachedOrRefreshedProfileFile;
```
L-Applin, line 105: Does this need to be synchronized?

L-Applin (commented):
We lose the immutable nature of the class, which is not the end of the world if it is required, but I was wondering if there was a way to make this change while keeping the immutable nature if this class.

core/auth/src/main/java/software/amazon/awssdk/auth/token/credentials/ProfileTokenProvider.java
dave-fn: I'll leave it as is this time, but thanks for pointing out `getOrDefault` as I was not aware.

core/auth/src/main/java/software/amazon/awssdk/auth/token/credentials/ProfileTokenProvider.java
dave-fn: @zoewangg? If needed, will have to make same change to `ProfileCredentialsProvider`.

core/auth/src/main/java/software/amazon/awssdk/auth/token/credentials/ProfileTokenProvider.java
dave-fn: @zoewangg? If needed, will have to make same change to `ProfileCredentialsProvider`.

zoewangg: Hmm, could `resolveToken()` be invoked by different threads _at the same time_? I'd assume no.

core/auth/src/test/java/software/amazon/awssdk/auth/token/credentials/ProfileTokenProviderTest.java
```diff
@@ -73,6 +74,36 @@ public void compatibleProfileSettings_callsLoader() {
...
+                                       + "sso_session = xyz");
+        Supplier<ProfileFile> supplier = Mockito.mock(Supplier.class);
+
+        ProfileTokenProvider provider =
+            ProfileTokenProvider.builder().profileFile(supplier).profileName("sso").build();
+
+        try {
+            Mockito.when(supplier.get()).thenReturn(file1);
+            provider.resolveToken();
+        } catch (IllegalArgumentException e) {
```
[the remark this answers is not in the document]
dave-fn: Yes, every functional test I reviewed (at least in this class) involving SSO cannot resolve and throw an exception. Since I wanted to verify two invocations, I could only think of swallowing the first exception to continue processing.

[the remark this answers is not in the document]
zoewangg: Gotcha, can we add a comment in catch block?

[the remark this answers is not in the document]
dave-fn: I just realized i could have asserted the exception instead. Think I should make the change?

```
Mockito.when(...);
assertThatThrownBy(provider::resolveToken)....;

Mockito.verify(supplier, ...);
```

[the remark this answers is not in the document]
zoewangg: Yeah, I was going to suggest that. Sounds good to me.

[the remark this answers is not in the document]
dave-fn: Restoring comment deleted by mistake. Will add assertions to better handle the exceptions.

```
Mockito.when(...);
assertThatThrownBy(provider::resolveToken)....;

Mockito.verify(supplier, ...);
```

core/auth/src/main/java/software/amazon/awssdk/auth/token/credentials/ProfileTokenProvider.java
dave-fn: Yeah, not sure what the exact use case is. I'd assume a main thread would be resolving?

zoewangg: Took another look and I think we should use some locking mechanism. Here is an example of how it could be done: https://github.com/aws/aws-sdk-java-v2/blob/master/utils/src/main/java/software/amazon/awssdk/utils/cache/CachedSupplier.java#L185

dave-fn: After offline discussion, decided to use CAS

L-Applin (approved):
lgtm

zoewangg (approved):

PR #3613, opened 2022-12-13
Make crt dependency optional in transfer manager module

zoewangg:
## Motivation and Context
Make crt dependency optional in transfer manager module so that once we have Java based TM, it'll be the default implementation. This means customers will need to explicitly add `aws-crt` dependency for now for parallel feature.

## Modifications
- Make crt dependency optional in transfer manager module
- Use Java based S3AsyncClient if CRT is not on the classpath in `S3TransferManager#create`

### DIFF
--- a/services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/DefaultS3TransferManager.java
+++ b/services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/DefaultS3TransferManager.java
@@ -26,6 +26,7 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
 import java.util.concurrent.Executor;
 import java.util.function.Consumer;
+import java.util.function.Supplier;
 import software.amazon.awssdk.annotations.SdkInternalApi;
 import software.amazon.awssdk.annotations.SdkTestInternalApi;
 import software.amazon.awssdk.arns.Arn;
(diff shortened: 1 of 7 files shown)


services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/DefaultS3TransferManager.java
```diff
@@ -134,6 +142,22 @@ public DefaultS3TransferManager(DefaultBuilder tmBuilder) {
...
+
+    private static boolean crtInClasspath() {
+        try {
+            ClassLoaderHelper.loadClass("software.amazon.awssdk.crt.s3.S3Client", false);
+        } catch (ClassNotFoundException e) {
+            return false;
+        }
+        return true;
+    }
+
```
zoewangg, line 160: It's not very easy to add test for this logic because crt is always in classpath in this module. I'll do a manual one-off test for this

dave-fn (approved):

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/DefaultS3TransferManager.java
joviegas: >crt is always in classpath in this module
I did not understand this part, since its optional why will this be always in class path ?
Also can we add this in test modules like service test module etc ?

Also do we need to update developer guide or any other docs where we need to tell user to take crt dependency explicitly in pom.xml  ?

zoewangg: I meant crt is always available in this module at compilation time, so it's not easy to add a test in this module (we'd need some way to exclude crt from classpath). We could add a test for this in a test module w/o adding crt dependency, but I'm not sure if it's worth the time and it seems flaky and fragile because in the future, we could add crt dependency in that module by accident and that will invalidate the test.

> Also do we need to update developer guide or any other docs where we need to tell user to take crt dependency explicitly in pom.xml ?

For dev guide, I already let Todd know and he will update his draft. Looks like I forgot about README. Will update it.

joviegas (approved):

PR #3630, opened 2022-12-15
Do not prefix copy-source with / when using the sourceBucket and sourceKey APIs

millems:
The main goal of this change is to fix an issue where access point ARNs were prefixed with /, causing them to fail because / is not supported as a prefix. This also stops prefixing buckets with /, but that seems to work fine in the included integration test.

### DIFF
--- a/services/s3/src/it/java/software/amazon/awssdk/services/s3/CopyObjectIntegrationTest.java
+++ b/services/s3/src/it/java/software/amazon/awssdk/services/s3/CopyObjectIntegrationTest.java
@@ -87,9 +87,10 @@ public class CopyObjectIntegrationTest extends S3IntegrationTestBase {
     @Test
     public void copyObject_CopiesObjectToNewKey() throws Exception {
         s3.copyObject(CopyObjectRequest.builder()
-                                       .copySource(BUCKET_NAME + "/" + SOURCE_KEY)
-                                       .bucket(BUCKET_NAME)
-                                       .key(DESTINATION_KEY)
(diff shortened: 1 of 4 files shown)


joviegas (approved):

services/s3/src/it/java/software/amazon/awssdk/services/s3/CopyObjectIntegrationTest.java
```diff
@@ -87,9 +87,10 @@ public static void initializeTestData() throws Exception {
     @Test
     public void copyObject_CopiesObjectToNewKey() throws Exception {
         s3.copyObject(CopyObjectRequest.builder()
-                                       .copySource(BUCKET_NAME + "/" + SOURCE_KEY)
-                                       .bucket(BUCKET_NAME)
-                                       .key(DESTINATION_KEY)
+                                       .sourceBucket(BUCKET_NAME)
```
joviegas, line 90: Why did we remove the ` .copySource(BUCKET_NAME + "/" + SOURCE_KEY)`  , can we keep the test case and assert behavior when user  calls copyObject with ` .copySource(BUCKET_NAME + "/" + SOURCE_KEY)`

millems: The test as it used to be was just testing the service. It's probably not a test we should have. Using sourceBucket and sourceKey is at least testing our creation of the copySource field.

PR #3648, opened 2022-12-20
[DRAFT] IMDS async token

L-Applin:
Still draft (forwarding of future cancellation is still incorrect)
  
## Types of changes

## Checklist

## License


core/imds/src/main/java/software/amazon/awssdk/imds/internal/AsyncTokenCache.java
```diff
@@ -0,0 +1,80 @@
...
+    private void refreshIfNeededAndComplete(CompletableFuture<Token> parentFuture) {
+        if (needsRefresh()) {
+            synchronized (refreshLock) {
+                // the value might have been refreshed while waiting on the lock,
+                // if so, refresh it only if it has already expired agan.
+                if (needsRefresh()) {
+                    try {
+                        CompletableFuture<Token> tokenFuture = supplier.get();
+                        CompletableFutureUtils.forwardExceptionTo(parentFuture, tokenFuture);
+                        cachedToken = tokenFuture.get();
```
millems, line 60: It feels wrong that supplier has its own executor (for returning the future) and we have to have our own executor as well. Could we remove the extra executor? Something like: (disclaimer: I didn't test this)

```java
private List<CompletableFuture<Token>> waitingFutures = new ArrayList<>();
private boolean refreshRunning = false;
private volatile Token cachedToken;

// Guards reads/writes to waitingFutures/refreshRunning and writes to cachedToken
private final Object refreshLock = new Object();

public CompletableFuture<Token> get() {
    // Optimization: read volatile cachedToken once to check needs-refresh and return result instead of
    // reading it twice.
    Token currentValue = cachedToken;
    if (!needsRefresh(currentValue)) {
        return CompletableFuture.completedFuture(currentValue);
    }

    synchronized (refreshLock) {
        // Make sure the value wasn't refreshed while we were waiting for the lock.
        currentValue = cachedToken;
        if (!needsRefresh(currentValue)) {
            return CompletableFuture.completedFuture(currentValue);
        }

        CompletableFuture<Token> result = new CompletableFuture<>();
        waitingFutures.add(result);

        if (!refreshRunning) {
            startRefresh();
        }

        return result;
    }
}

private void startRefresh() {
    supplier.get().whenComplete((token, throwable) -> {
        List<CompletableFuture<Token>> currentWaitingFutures;

        synchronized (refreshLock) {
            // Instead of completing the waiting future while holding the lock, we copy the list reference and
            // release the lock before completing them. This is just in case someone (naughty) is doing something
            // blocking on the complete calls. It's not good that they're doing that, but at least
            // it won't block other threads trying to acquire the lock.
            currentWaitingFutures = waitingFutures;
            waitingFutures = new ArrayList<>();
            refreshRunning = false;
            if (throwable != null) {
                cachedToken = token;
            }
        }

        if (throwable != null) {
            // I assume the supplier has retries built in, so we don't do it here.
            currentWaitingFutures.forEach(f -> f.completeExceptionally(throwable));
        } else {
            currentWaitingFutures.forEach(f -> f.complete(token));
        }
    });

    refreshRunning = true; // After supplier.get(), in case that throws an exception
}
```

millems (commented):
I only reviewed pom.xml and AsyncTokenCache.java. Hope that's okay!

PR #3649, opened 2022-12-20
[IMDS] Token cache for Async client

L-Applin:
## Motivation and Context
Required as part of the async IMDS Client. The async client also needs to cache the token and reuse it if it is not expired on subsequent requests.

## Modifications
Added AsyncTokenCache class, caching the token and returning a CompletableFuture.
Removed TokenCacheStrategy.
Moved some async HTTP request logic to an helper class.
Added log + javadoc.

## Testing
Update async unit test to make them slightly more robust.

## Types of changes

## Checklist

## License


core/imds/src/main/java/software/amazon/awssdk/imds/internal/AsyncTokenCache.java
```diff
@@ -0,0 +1,138 @@
...
+            currentValue = cachedToken;
+            if (!needsRefresh(currentValue)) {
+                return CompletableFuture.completedFuture(currentValue);
+            }
+            CompletableFuture<Token> result = new CompletableFuture<>();
+            if (refreshRunning.get()) {
+                waitingFutures.add(result);
+            } else {
+                CompletableFuture<Token> tokenRequest = startRefresh(result);
+                CompletableFutureUtils.forwardExceptionTo(result, tokenRequest);
```
millems, line 93: We probably don't want this. A cancellation of the result future probably shouldn't stop the refresh that's happening in the background, because other calls might be waiting on that refresh.

core/imds/src/main/java/software/amazon/awssdk/imds/internal/AsyncTokenCache.java
```diff
@@ -0,0 +1,138 @@
...
+     * Indicates if the token refresh request is currently running or not.
+     */
+    private final AtomicBoolean refreshRunning;
+
+    private final Object refreshLock = new Object();
+
+    AsyncTokenCache(Supplier<CompletableFuture<Token>> supplier) {
+        this.supplier = supplier;
+        this.waitingFutures = new ArrayList<>();
+        this.refreshRunning = new AtomicBoolean(false);
```
millems, line 72: Nit: any reason to assign these in the constructor instead of inline? Some other things are assigned inline already.

core/imds/src/main/java/software/amazon/awssdk/imds/internal/AsyncTokenCache.java
```diff
@@ -0,0 +1,138 @@
...
+     * the token once it finishes.
+     */
+    private final Supplier<CompletableFuture<Token>> supplier;
+
+    /**
+     * The collection of future that are waiting for the refresh call to complete. If a call to {@link AsyncTokenCache#get()}
+     * is made while the token request is happening, a future will be added to this collection. All future in this collection
+     * are completed once the token request is done.
+     */
+    private Collection<CompletableFuture<Token>> waitingFutures;
```
millems, line 60: Nit: Should we document that this should only be modified while holding the refreshLock?

core/imds/src/main/java/software/amazon/awssdk/imds/internal/AsyncTokenCache.java
L-Applin: I see what you mean. I think it's fine to remove it because we will still cancel the request made with the token after, it's only the token refresh call that would be not be cancellable by the users.

millems (approved):

PR #3669, opened 2023-01-04
version agnostic README;

h1alexbel:
version agnostic ```README``` improves user experience and you don't need to update version info every time after releasing a new version.


debora-ito:
Hi @h1alexbel,

The version update is done by code generators, and since it actually shows the most recent version of SDK I think it's a better customer experience than simply showing the text "get from maven repository".

Will go ahead and close this. Thank you for reaching out.

PR #3676, opened 2023-01-08
Do not attempt to pad time string if current length is sufficient

belugabehr:
## Motivation and Context
This method is used exclusively to pad the results of `Duration#getNano` up to nine characters.

https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html#getNano--

The range of possible values are: 0 to 999,999,999

Of the set of values with a length less than 9 are: [0,100000000)
Of the set of values with a length greater than 9 are: [100000000,999999999]

So, just by random dart-throwing, 90% of the values are going to have a length of 9 digits and require no padding.  Save on instantiation of `StringBuilder`, and copying data into them, by only padding when required.

## Modifications
Check for the size of the string, and if it is already the same size as the require padded value, simply return the value itself.

## Testing
This is a performance enhancement.  There are no functionality changes and the existing test suite validates.

## Types of changes

## Checklist

## License

### DIFF
--- a/services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/converter/ConverterUtils.java
+++ b/services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/converter/ConverterUtils.java
@@ -53,14 +53,19 @@ public class ConverterUtils {
     }
 
     public static String padLeft(int paddingAmount, int valueToPad) {
-        String value = Integer.toString(valueToPad);
-        int padding = paddingAmount - value.length();
-        StringBuilder result = new StringBuilder(paddingAmount);
-        for (int i = 0; i < padding; i++) {
-            result.append('0');
+        final String result;
+        final String value = Integer.toString(valueToPad);
+        if (value.length() == paddingAmount) {
+            result = value;
+        } else {
+            int padding = paddingAmount - value.length();
+            StringBuilder sb = new StringBuilder(paddingAmount);
+            for (int i = 0; i < padding; i++) {
+                sb.append('0');
+            }
+            result = sb.append(value).toString();
         }
-        result.append(value);
-        return result.toString();
+        return result;
     }
 
     public static String[] splitNumberOnDecimal(String valueToSplit) {


debora-ito:
Hi @belugabehr 

Can you give more context on the motivation behind the change, and how would this impact your use-case?
Also, please send us data showing the performance improvement, if you have.

belugabehr:
No impact, just something I noticed by casual observation of the AWS SDK code.


JMH results:

```
Benchmark                        Mode  Cnt         Score        Error  Units
TestPad.JDkFormat               thrpt   30   1561389.415 ±  49011.241  ops/s
TestPad.padLeft                 thrpt   30  12399672.607 ± 928272.152  ops/s
TestPad.padLeftWithLengthCheck  thrpt   30  17327199.282 ± 502528.561  ops/s
```

* `TestPad.JDkFormat` = `String.format("%09d", value)`
* `TestPad.padLeft` = Current implementation
* `TestPad.padLeftWithLengthCheck` = Proposed implementation

Just randomly generating values and padding them.  Adding the length check is about 40% more operations per second.  Also cuts down on the number of objects that need to be GC'ed.

cenedhryn (approved):

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/converter/ConverterUtils.java
```diff
@@ -53,14 +53,19 @@ public static void validateFloat(Float input) {
     }
 
     public static String padLeft(int paddingAmount, int valueToPad) {
-        String value = Integer.toString(valueToPad);
-        int padding = paddingAmount - value.length();
-        StringBuilder result = new StringBuilder(paddingAmount);
-        for (int i = 0; i < padding; i++) {
-            result.append('0');
+        final String result;
+        final String value = Integer.toString(valueToPad);
```
cenedhryn, line 57: Our checkstyle rules do not allow final on local variables. Can you update?

belugabehr: Done! Thanks for looking at this PR.

PR #3681, opened 2023-01-11
[IMDS] Surface area review comments

L-Applin:
## Modifications
- Add builder consumer style method for retry policy
- Remove getters on Builders
- Add httpClientBuilder method to client builder
- Check and Improve (if needed) error message on missing http client dependency
    - Original message is ok:  `software.amazon.awssdk.core.exception.SdkClientException: Unable to load an HTTP implementation from any provider in the chain. You must declare a dependency on an appropriate HTTP implementation or pass in an SdkHttpClient explicitly to the client builder.`
- Link to URL of the dev guide in the JAVADOC of the get method of the sync and async client
- Rename to Ec2MetadataResponse

## Types of changes

## Checklist

## License


core/imds/src/main/java/software/amazon/awssdk/imds/Ec2MetadataClientBuilder.java
```diff
@@ -33,15 +34,30 @@
     /**
      * Define the retry policy which includes the number of retry attempts for any failed request.
      * <p>
-     * If not specified, defaults to 3 retry attempts and a {@link BackoffStrategy#defaultStrategy()}  backoff strategy} that
-     * uses {@link RetryMode#STANDARD}
+     * If not specified, defaults to 3 retry attempts and a {@link BackoffStrategy#defaultStrategy()} backoff strategy} that
+     * uses {@link RetryMode#STANDARD}. Can be also specified by using the
+     * {@link Ec2MetadataClientBuilder#retryPolicy(Consumer)} method. If both methods are called multiple time, the last call
```
davidh44, line 39: A bit ambiguous here. Do you mean something like, "If either method is called multiple times, or if both methods are called" ?
Same for below

davidh44: Yeah that sounds good to me

core/imds/src/main/java/software/amazon/awssdk/imds/Ec2MetadataAsyncClient.java
```diff
@@ -84,5 +90,15 @@ interface Builder extends Ec2MetadataClientBuilder<Ec2MetadataAsyncClient.Builde
          * @return a reference to this builder
          */
         Builder httpClient(SdkAsyncHttpClient httpClient);
+
+        /**
+         * A http client builder used to retrieve an instance of an {@link SdkAsyncHttpClient}. If specified, the Ec2
```
zoewangg, line 95: nit, `An HTTP client`

core/imds/src/main/java/software/amazon/awssdk/imds/Ec2MetadataAsyncClient.java
```diff
@@ -84,5 +90,15 @@ interface Builder extends Ec2MetadataClientBuilder<Ec2MetadataAsyncClient.Builde
...
+         * Metadata Client will use the instance returned by the builder and manage its lifetime by closing the http client
+         * once the Ec2 Client itself is closed.
+         *
+         * @param builder the builder to used to retrieve an instance.
+         * @return a reference to this builder
+         */
+        Builder httpClient(DefaultSdkAsyncHttpClientBuilder builder);
```
zoewangg, line 102: It should be 

 ```
Builder httpClientBuilder(SdkAsyncHttpClient.Builder builder);
```

https://github.com/aws/aws-sdk-java-v2/blob/master/core/sdk-core/src/main/java/software/amazon/awssdk/core/client/builder/SdkSyncClientBuilder.java#L56

https://github.com/aws/aws-sdk-java-v2/blob/master/core/sdk-core/src/main/java/software/amazon/awssdk/core/client/builder/SdkAsyncClientBuilder.java#L76

File: core/imds/src/main/java/software/amazon/awssdk/imds/Ec2MetadataClient.java [hunk 4]
```diff
@@ -69,6 +74,15 @@ interface Builder extends Ec2MetadataClientBuilder<Ec2MetadataClient.Builder, Ec
...
 
+        /**
+         * A http client builder used to retrieve an instance of an {@link SdkHttpClient}. If specified, the Ec2 Metadata Client
+         * will use the instance returned by the builder and manage its lifetime by closing the http client once the Ec2 Client
+         * itself is closed.
+         *
+         * @param builder the builder to used to retrieve an instance.
+         * @return a reference to this builder
+         */
+        Builder httpClient(DefaultSdkHttpClientBuilder builder);
```
zoewangg, line 85: Same here.

core/imds/src/main/java/software/amazon/awssdk/imds/Ec2MetadataResponse.java
```diff
@@ -30,26 +30,26 @@
 /**
  * The class is used for response handling and parsing the metadata fetched by the get call in the {@link Ec2MetadataClient}
  * interface.
- * The class provides convenience methods to the users to parse the metadata as a String, List and Document.
+ * The class provides convenience methods to the users to parse the metadata as a String, List and Document (json).
```
zoewangg, line 33: What do we mean by "Document (json)"?

core/imds/src/main/java/software/amazon/awssdk/imds/Ec2MetadataRetryPolicy.java
```diff
@@ -122,16 +122,14 @@ public interface Builder extends CopyableBuilder<Ec2MetadataRetryPolicy.Builder,
          */
         Builder numRetries(Integer numRetries);
 
-        @Override
-        Ec2MetadataRetryPolicy build();
     }
 
-    private static final class BuilderImpl implements Builder {
+    public static final class BuilderImpl implements Builder {
```
zoewangg, line 127: Hmm, why do we need it to be public?

core/imds/src/main/java/software/amazon/awssdk/imds/internal/DefaultEc2MetadataAsyncClient.java
```diff
@@ -183,6 +194,15 @@ public Ec2MetadataAsyncBuilder endpointMode(EndpointMode endpointMode) {
...
             this.httpClient = httpClient;
+            this.httpClientIsInternal = false;
+            return this;
+        }
+
+        @Override
+        public Builder httpClient(DefaultSdkAsyncHttpClientBuilder builder) {
+            SdkAsyncHttpClient builtHttpClient = builder.build();
+            this.httpClientIsInternal = true;
+            this.httpClient = builtHttpClient;
```
zoewangg, line 205: Can we move this logic to the ctor of `DefaultEc2MetaAsyncClient`? Builder should be as simple as possible IMO.

core/imds/src/main/java/software/amazon/awssdk/imds/internal/StringResponseHandler.java
```diff
@@ -33,9 +33,9 @@
 public class StringResponseHandler implements HttpResponseHandler<String> {
     private static final Logger log = Logger.loggerFor(StringResponseHandler.class);
 
-    private CompletableFuture<String> future;
+    private CompletableFuture<?> future;
```
zoewangg, line 36: Hmm, is there any reason we change it to wildcard?

core/imds/src/main/java/software/amazon/awssdk/imds/internal/TokenResponseHandler.java
```diff
@@ -16,55 +16,45 @@
...
-            // retryable error
-            future.completeExceptionally(RetryableException.create(responseContent));
-        }
+        String tokenValue = delegateHandler.handle(response, executionAttributes);
         Duration ttl = response.firstMatchingHeader(EC2_METADATA_TOKEN_TTL_HEADER)
                                .map(Long::parseLong)
                                .map(Duration::ofSeconds)
-                               .orElseGet(() -> Duration.ofSeconds(ttlSeconds));
-        return new Token(responseContent, ttl);
+                               .orElseGet(() -> {
```
zoewangg, line 51: Is it possible that the response will not have TTL?

core/imds/src/main/java/software/amazon/awssdk/imds/Ec2MetadataRetryPolicy.java
L-Applin: we need to instantiate a Ec2MetadataRetryPolicy.Builder for the `retryPolicy(Consumer<Ec2MetadataRetryPolicy.Builder> builderConsumer)`

core/imds/src/main/java/software/amazon/awssdk/imds/internal/StringResponseHandler.java
L-Applin: since I'm reusing the `StringResponseHandler` in the `TokenResponseHandler`, this class needs to handle CompletableFutures of String and Token. Since it's only completed exceptionnaly here, a wildcard seemed fine.

core/imds/src/main/java/software/amazon/awssdk/imds/Ec2MetadataResponse.java
L-Applin: what I mean is that the method `asDocument` returns a `Document` interface, which is a representation of json a format.

core/imds/src/main/java/software/amazon/awssdk/imds/internal/TokenResponseHandler.java
L-Applin: The server should always return a response with the ttl header. I was hesitating between this and throwing an exception if none were found.

core/imds/src/main/java/software/amazon/awssdk/imds/Ec2MetadataRetryPolicy.java
zoewangg: It seems we can just use `applyMutation` instead of instantiating the builder class.
```
Ec2MetadataRetryPolicy.builder().applyMutation(builderConsumer)? 
```
https://github.com/aws/aws-sdk-java-v2/blob/master/services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/S3TransferManager.java#L209

core/imds/src/main/java/software/amazon/awssdk/imds/internal/TokenResponseHandler.java
zoewangg: Can we confirm with the SEP author to see what the right behavior it should be?

core/imds/src/main/java/software/amazon/awssdk/imds/Ec2MetadataRetryPolicy.java
L-Applin: oh nice, yeah that works 👍

L-Applin: Oh wait, we dont even need applyMutation, I can just use `Ec2MetadataRetryPolicy.builder()` 🤦

core/imds/src/main/java/software/amazon/awssdk/imds/internal/DefaultEc2MetadataAsyncClient.java
zoewangg: If users call `httpClientBuilder(xxx)` multiple times, we'd create multiple HTTP instances, which could cause memory leak in extreme cases...
```
Ec2MetaAsyncClient.builder()
                  .httpClientBuilder(xxx)
                  .httpClientBuilder(xxx)
                  ...
                  .build()
```

zoewangg: It seems we can achieve the same in the ctor. IMO builder class should be treated as a simple POJO class w/o too much logic. We should probably add validation as well to make sure only one of the setters is called. 
https://github.com/aws/aws-sdk-java-v2/blob/2d27e75f6241913e028170b79c7e7072ce06d008/core/sdk-core/src/main/java/software/amazon/awssdk/core/client/builder/SdkDefaultClientBuilder.java#L343-L349

core/imds/src/main/java/software/amazon/awssdk/imds/Ec2MetadataResponse.java
zoewangg: I'm not sure if Document is tied to JSON. Should we just link to the Document class? `{@link Document}`. Users can read more in the Document class

core/imds/src/main/java/software/amazon/awssdk/imds/internal/DefaultEc2MetadataClient.java
```diff
@@ -266,26 +276,31 @@ public Ec2MetadataBuilder endpointMode(EndpointMode endpointMode) {
...
+            return this;
+        }
+
         public Ec2MetadataRetryPolicy getRetryPolicy() {
             return this.retryPolicy;
         }
 
-        @Override
-        public URI getEndpoint() {
+        URI getEndpoint() {
```
zoewangg, line 295: We should probably make them public to be consistent with other getters used for reflection.

core/imds/src/main/java/software/amazon/awssdk/imds/internal/StringResponseHandler.java
zoewangg: I'm a bit confused of how this future works here. We can probably keep it as is and follow up on this in separate PR.

core/imds/src/main/java/software/amazon/awssdk/imds/internal/TokenResponseHandler.java
L-Applin: Talked to Matt, he pointed me to the ruby sdk which throws raise an error, so I will change it to throw an exception in the sync client, and complete exceptionally the future in the async client.

core/imds/src/main/java/software/amazon/awssdk/imds/Ec2MetadataResponse.java
L-Applin: Maybe I wasn't clear, the implementation of `asDocument` assumes the response is in json format, and returns a Document instance which is the representation of that that json. If it's not in json format, an exception will be thrown

L-Applin: would the following be more clear?

>  The class provides convenience methods to the users to parse the metadata as a String and List. Also provides ways to parse a the metadata as Document type if it is in the json format.

core/imds/src/main/java/software/amazon/awssdk/imds/internal/StringResponseHandler.java
L-Applin: Looking at the AsyncResponseHandler class, we should be able to throw an exception here instead of exceptionally completing the future. That will make things more clear

core/imds/src/main/java/software/amazon/awssdk/imds/internal/DefaultEc2MetadataClient.java
```diff
@@ -62,13 +65,20 @@ public final class DefaultEc2MetadataClient extends BaseEc2MetadataClient implem
 
     private DefaultEc2MetadataClient(Ec2MetadataBuilder builder) {
         super(builder);
-        this.httpClient = Validate.getOrDefault(builder.httpClient,
-                                                () -> new DefaultSdkHttpClientBuilder().buildWithDefaults(IMDS_HTTP_DEFAULTS));
+
+        // http client
```
zoewangg, line 69: nit : do we need this comment?

core/imds/src/main/java/software/amazon/awssdk/imds/internal/DefaultEc2MetadataClient.java
```diff
@@ -211,17 +221,24 @@ private Token getToken() {
...
+                             .firstMatchingHeader(EC2_METADATA_TOKEN_TTL_HEADER)
+                             .orElseThrow(() -> RetryableException
+                                 .builder()
+                                 .message(EC2_METADATA_TOKEN_TTL_HEADER + " header not found in token response")
+                                 .build());
+        Duration ttlDuration;
+        try {
+            ttlDuration = Duration.ofSeconds(Long.parseLong(ttl));
+        } catch (NumberFormatException nfe) {
+            throw RetryableException.create("Invalid token format received from IMDS server", nfe);
```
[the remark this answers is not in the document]
L-Applin: I do think so yes, non-retryable errors are only client-side error

core/imds/src/main/java/software/amazon/awssdk/imds/internal/DefaultEc2MetadataClient.java
L-Applin: removed

core/imds/src/main/java/software/amazon/awssdk/imds/internal/DefaultEc2MetadataClient.java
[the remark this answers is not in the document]
zoewangg: IMO, if a service is sending an invalid format, that means something is seriously wrong with it, and retrying will not likely succeed. Based on the SDK retry specification, we should only retry requests if it's a transient error (IOExceptions or certain error code) or a throttling error

zoewangg (approved):

core/imds/src/main/java/software/amazon/awssdk/imds/internal/DefaultEc2MetadataClient.java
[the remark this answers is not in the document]
L-Applin: I see, switching to SdkClientException

PR #3686, opened 2023-01-12
Changed GeneratedServiceMetadataProvider to lazily load elements. Updated unit tests based to reflect changes.

### LINKED ISSUE #3664: Initializing GeneratedServiceMetadataProvider takes a while
### Describe the feature

When creating a service client (e.g. DynamoDbAsyncClient) there is an expensive one time call to create `GeneratedServiceMetadataProvider` which incurs the penalty of an expensive static initializer. The static initializer creates a map object with  299 entries (currently), each of which also initializes an instance that extends `ServiceMetadata` which *also* all have large static initialization overhead. 

From the looks of it, every single instance of `ServiceMetadata` is initialized but in my use case, only `DynamodbServiceMetadata` was needed. And this call was only needed to acquire the client endpoint.

I profiled just the cold start initialization of `GeneratedServiceMetadataProvider` to confirm my suspicions and get a general idea of the overhead incurred. On my machine the operation took between 350 and 500 ms. I think this could be improved significantly by implementing lazy loading of `ServiceMetadata` objects.

### Benchmark

```java
@State(Scope.Benchmark)
@Warmup(iterations = 0)
@Measurement(iterations = 1)
@Fork(20)
public class GeneratedProviderBenchmark {

    @Benchmark
    @BenchmarkMode(Mode.SingleShotTime)
    @OutputTimeUnit(TimeUnit.MILLISECONDS)
    public void createObject() {
        GeneratedServiceMetadataProvider generated = new GeneratedServiceMetadataProvider();
        ServiceMetadata metadata = generated.serviceMetadata("xray");
    }

    public static void main(String... args) throws RunnerException, CommandLineOptionException {
        Options opt = new OptionsBuilder()
            .parent(new CommandLineOptions())
            .include(GeneratedProviderBenchmark.class.getSimpleName())
            .addProfiler(StackProfiler.class)
            .build();
        Collection<RunResult> run = new Runner(opt).run();
    }
}
```
### Results: 

| Benchmark                                                          | Mode  | Cnt   | Score   |  Error        | Units   |
|----------------------------------------------|-------|------|--------|-----------|--------|
| GeneratedProviderBenchmark.createObject   |     ss    | 20     |  364.140 |  ± 16.565   | ms/op |

### Use Case

The goal is to reduce the overhead as much as possible and make cold start times as performant as they can be. This is an ongoing process that a number of other tickets are related to. In fact the [ticket](https://github.com/aws/aws-sdk-java-v2/issues/748) that lead to the creation of the `GeneratedServiceMetadataProvider` class was done so in an effort to reduce cold start times of the `DynamoDbClient`.

### Proposed Solution

Lazily initialize the `GeneratedServiceMetadataProvider` using a static factory method. I created a proof of concept to test a possible improvement to the object initialization. 

```java
public final class GeneratedServiceMetadataProvider implements ServiceMetadataProvider {
    private static final Map<String, ServiceMetadata> SERVICE_METADATA = new HashMap<>();

    private static ServiceMetadata createServiceMetadata(String endpointPrefix) {
        switch (endpointPrefix) {
            case "a4b":
                return new A4bServiceMetadata();
            ... // other cases      
            default:
                throw new IllegalStateException("Unexpected value: " + endpointPrefix);
        }
    }

    private static ServiceMetadata getServiceMetadata(String endpointPrefix) {
        if (!SERVICE_METADATA.containsKey(endpointPrefix)) {
            SERVICE_METADATA.put(endpointPrefix, createServiceMetadata(endpointPrefix));
        }
        return SERVICE_METADATA.get(endpointPrefix);
    }

    @Override
    public ServiceMetadata serviceMetadata(String endpointPrefix) {
        return getServiceMetadata(endpointPrefix);
    }
}
```
This produced the following results using the same benchmark as above:

| Benchmark                                                          | Mode  | Cnt   | Score   |  Error        | Units   |
|----------------------------------------------|-------|------|--------|-----------|--------|
| GeneratedProviderBenchmark.createObject   |     ss    | 20     |  14.140 |  ± 0.465   | ms/op |

About a 25x improvement in object creation speed, not to mention the memory overhead thats saved by deferring initialization of all the other `ServiceMetadata` instances that aren't needed.

### Other Information

As an aside, I was surprised that providing an `EndpointProvider` to the client builder - one which doesn't go through `GeneratedServiceMetadataProvider` - wasn't used at all when setting the endpoint of the client on creation. Why wouldn't the client builder use the endpoint provider, if present, if it's going to use the provider for every request anyway?

Related issues:
https://github.com/aws/aws-sdk-java-v2/issues/748
https://github.com/aws/aws-sdk-java-v2/issues/6

### Acknowledgements

- [X] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### AWS Java SDK version used

2

### JDK version used

1.8

macOS Monterey 12.6.1

awmcc90:
## Motivation and Context

`GeneratedServiceMetadataProvider` takes a while to initialize. This was benchmarked and shown in [this ticket](https://github.com/aws/aws-sdk-java-v2/issues/3664). Lazily loading metadata significantly reduces the overhead required to initialize this class and reduces the time it takes to load substantially. This is important in the era of serverless infrastructure where cold start times need to be minimized.

Resolves #3664

## Modifications

- All changes scoped to the codegen-lite package
- Changed `ServiceMetadataProviderGenerator`
  - Removed initialization of all `ServiceMetadata` entries in the constructor
  - Added codegen method to create a static `ServiceMetadata` factory method which accepts a `String` value (same as before) and generates the associated `ServiceMetadata` when required.
  - Updated codegen methods to add logic required for calling new factory method
  - Created `ServiceMetadata` objects are cached using the same map as before and this is checked prior to calling the factory
- Updated test resource `service-metadata-provider.java` to reflect new `GeneratedServiceMetadataProvider` structure

## Testing

- Updated unit tests and confirmed they succeeded
- Testing environment:
  - amazon corretto 11 jdk
  - MacOS Monterey 12.6.2

## Types of changes

## Checklist

## License

### DIFF
--- a/codegen-lite/src/main/java/software/amazon/awssdk/codegen/lite/regions/ServiceMetadataProviderGenerator.java
+++ b/codegen-lite/src/main/java/software/amazon/awssdk/codegen/lite/regions/ServiceMetadataProviderGenerator.java
@@ -22,12 +22,12 @@ import static javax.lang.model.element.Modifier.STATIC;
 
 import com.squareup.javapoet.AnnotationSpec;
 import com.squareup.javapoet.ClassName;
-import com.squareup.javapoet.CodeBlock;
 import com.squareup.javapoet.FieldSpec;
 import com.squareup.javapoet.MethodSpec;
 import com.squareup.javapoet.ParameterizedTypeName;
 import com.squareup.javapoet.TypeName;
 import com.squareup.javapoet.TypeSpec;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
@@ -38,7 +38,6 @@ import software.amazon.awssdk.annotations.SdkPublicApi;
 import software.amazon.awssdk.codegen.lite.PoetClass;
 import software.amazon.awssdk.codegen.lite.Utils;
 import software.amazon.awssdk.codegen.lite.regions.model.Partitions;
-import software.amazon.awssdk.utils.ImmutableMap;
 
 public class ServiceMetadataProviderGenerator implements PoetClass {
 
@@ -47,8 +46,8 @@ public class ServiceMetadataProviderGenerator implements PoetClass {
     private final String regionBasePackage;
 
     public ServiceMetadataProviderGenerator(Partitions partitions,
-                                           String basePackage,
-                                           String regionBasePackage) {
+                                            String basePackage,
+                                            String regionBasePackage) {
         this.partitions = partitions;
         this.basePackage = basePackage;
         this.regionBasePackage = regionBasePackage;
@@ -57,8 +56,8 @@ public class ServiceMetadataProviderGenerator implements PoetClass {
     @Override
     public TypeSpec poetClass() {
         TypeName mapOfServiceMetadata = ParameterizedTypeName.get(ClassName.get(Map.class),
-                                                                 ClassName.get(String.class),
-                                                                 ClassName.get(regionBasePackage, "ServiceMetadata"));
+                                                                  ClassName.get(String.class),
+                                                                  ClassName.get(regionBasePackage, "ServiceMetadata"));
         return TypeSpec.classBuilder(className())
                        .addModifiers(PUBLIC)
                        .addSuperinterface(ClassName.get(regionBasePackage, "ServiceMetadataProvider"))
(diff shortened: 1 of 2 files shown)


debora-ito:
Hi @awmcc90 

We are in a process of a major refactor of the endpoint resolution, and ServiceMetadata is one of the key points we are going to tackle. I'm going to close this PR but we appreciate your work. Will update https://github.com/aws/aws-sdk-java-v2/issues/3664 when the refactor is finalized.

Issue #3701, opened 2023-01-18
DynamoDbEnhancedClient :  java.lang.OutOfMemoryError: Metaspace


laulee5135:
### Describe the bug

Use DynamoDbEnhancedClient to perform query operation, java.lang.OutOfMemoryError: Metaspace occurs, after dump file analysis, LambdaToMethodBridgeBuilder overflows

### Expected Behavior

no OOM

### Current Behavior

java.lang.OutOfMemoryError: Metaspace

### Reproduction Steps

```
public class DynamoDBStoreService{
     
     DynamoDbClient ddb = dynamoDbClientBuilder.build();
     DynamoDbEnhancedClient enhancedClient = DynamoDbEnhancedClient.builder()
                    .dynamoDbClient(ddb)
                    .build();

    public void getKlineHistory(){
      //core code
      String partitionVal = "";
      List<KlineVo> klineVos  = new ArrayList<>();
      DynamoDbTable<KlineVo> mappedTable = enhancedClient.table("kline", TableSchema.fromBean(KlineVo.class));
      QueryConditional queryConditional = QueryConditional.keyEqualTo(Key.builder().partitionValue(partitionVal).build());
      QueryEnhancedRequest queryEnhancedRequest = QueryEnhancedRequest.builder()
                        .queryConditional(queryConditional)
                        .scanIndexForward(false)
                        .limit(100)
                        .build();
      PageIterable<KlineVo> query = mappedTable.query(queryEnhancedRequest);
      Iterator<KlineVo> iterator = query.items().iterator();
      while (iterator.hasNext()) {
                    KlineVo klineVo = iterator.next();
                    klineVos.add(klineVo);
      }
      System.out.println(klineVos);
   }
   
}

```

It can be reproduced by pressure testing,  You will find that the number of classes begins to increase significantly, and the metaspace also begins to increase

### AWS Java SDK version used

2.19.17

### JDK version used

openjdk version "17.0.1" 2021-10-19 LTS OpenJDK Runtime Environment Zulu17.30+15-CA (build 17.0.1+12-LTS)

### Operating System and version

Linux

debora-ito:
Hi @laulee5135 thank you for reaching out.

We are aware of the issue and investigating what could be the root cause. 

So far we identified that the issue occurs when using Java 17, and it has to do with `TableSchema.fromClass()` or `TableSchema.fromBean()` calls. Reducing the times those calls are made should help.

We'll post any updates we get from the investigation.

[...]

Issue #2913, opened 2021-12-13
Memory leak for Java 17 - when exclusiveStartKey is set


jmcquaidMetapack:
### Describe the bug

When using Java 17 and querying using exclusiveStartKey (specifying partition and sort values), the JVM metaspace fills up with instances of this class software.amazon.awssdk.enhanced.dynamodb.internal.mapper.LambdaToMethodBridgeBuilder.

GC has no effect. 

### Expected behavior

Metaspace does not fill up. For Java 11, the space is periodically cleared of instances of class software.amazon.awssdk.enhanced.dynamodb.internal.mapper.LambdaToMethodBridgeBuilder

### Current behavior

Metapace fills up incrementally- the more queries that are made the more memory is uses up.

This line of code results in a new class permanently added to metaspace each time it's invoked:
`new HashMap<>(Key.builder()
            .partitionValue("1")
            .sortValue("tom")
            .build()
            .primaryKeyMap(TableSchema.fromClass(TestItem.class)));`

### Steps to Reproduce

Follow the steps on the readme
https://bitbucket.org/jmcquaidlct/aws-enhanced-client-java-17-issue/src/master/

### AWS Java SDK version used

2.17.99

### JDK version used

17.0.1

### Operating System and version

Mac OS 11.4

jmcquaidMetapack:
It turns out the specific piece of code that causes this issue is:
TableSchema.fromClass(TestItem.class);

By creating this only once we've resolved the issue

PR #3710, opened 2023-01-24
Cache BeanTableSchema and ImmutableTableSchema to prevent potential m…

zoewangg:
## Motivation and Context
Related to #3701, #2913 

Starting from JDK 15, `LambdaMetaFactory` creates [hidden class](https://openjdk.org/jeps/371)es instead of VM-anonymous class, which has caused behavior change of how generated classes are loaded and unloaded. When running the application with JDK 15+, if users invokes `TableSchema#fromClass` or `TableSchame#fromBean` repeatably, they will get `java.lang.OutOfMemoryError: Metaspace`

## Modifications

- Cache BeanTableSchema and ImmutableTableSchema to prevent potential metaspace memory leak with JDK 15+
- Update documentation

## Testing
N/A
## Types of changes

## Checklist

## License


cenedhryn (approved):

millems:
Should we use a weak hash map, so that we don't preserve GC'd classes in the cache? Otherwise loading/unloading classloaders would leak resources.

zoewangg:
> Should we use a weak hash map, so that we don't preserve GC'd classes in the cache? Otherwise loading/unloading classloaders would leak resources.

I was only thinking about the case where users were trying to create table schema for the same bean per request, so it probably would not matter too much. But yeah, I guess in extreme cases, if they have a lot of beans, yeah, it could potentially be an issue. I'll make the change. With JDK 15+, those generated hidden classes will not be able to be unloaded anyway though because their defining classloaders are app class loaders, which will never be reclaimed.

millems (approved):

PR #3712, opened 2023-01-24
Merge credentials reloading feature branch to mainline

dave-fn:
## Motivation and Context
Add support for reloading credentials provided by `ProfileCredentialsProvider` if file changes on disk.
In addition, propagate changes to other classes with similar behavior: such as `DefaultCredentialsProvider`, `InstanceProfileCredentialsProvider` and `ProfileTokenProvider`.
Finally, some service classes also implemented this functionality so that configuration options might be changed on the fly.

## Modifications
Multiple PRs

## Testing
Added new test cases
Will run integ tests

## Types of changes

## Checklist

## License


core/auth/src/main/java/software/amazon/awssdk/auth/credentials/ProfileCredentialsProvider.java
```diff
@@ -127,19 +115,45 @@ public AwsCredentials resolveCredentials() {
...
-                       .add("profileFile", profileFile)
+                       .add("profileFile", currentProfileFile)
                        .build();
     }
 
     @Override
     public void close() {
+        if (profileFile instanceof SdkAutoCloseable) {
+            ((SdkAutoCloseable) profileFile).close();
+        }
```
millems, line 156: This is a red flag to me: closing something we didn't necessarily create. What if this is being used by someone else and we just closed it/broke it?

My bigger question is: why are profile file suppliers closeable?

core/profiles/src/main/java/software/amazon/awssdk/profiles/ProfileFileSupplier.java
```diff
@@ -0,0 +1,160 @@
...
+    static ProfileFileSupplier defaultSupplier() {
+        Optional<ProfileFileSupplier> credentialsSupplierOptional
+            = ProfileFileLocation.credentialsFileLocation()
+                                 .map(path -> reloadWhenModified(path, ProfileFile.Type.CREDENTIALS));
+
+        Optional<ProfileFileSupplier> configurationSupplierOptional
+            = ProfileFileLocation.configurationFileLocation()
+                                 .map(path -> reloadWhenModified(path, ProfileFile.Type.CONFIGURATION));
+
+        ProfileFileSupplier supplier = () -> null;
```
millems, line 61: Should we return an empty file instead of null? Or do we handle null everywhere that we handle profile files?

core/profiles/src/main/java/software/amazon/awssdk/profiles/ProfileFileSupplier.java
```diff
@@ -0,0 +1,160 @@
...
+            final AtomicReference<ProfileFile> currentAggregateProfileFile = new AtomicReference<>();
+
+            @Override
+            public ProfileFile get() {
+                ProfileFile.Aggregator aggregator = ProfileFile.aggregator();
+                for (ProfileFileSupplier supplier : suppliers) {
+                    aggregator.addFile(supplier.get());
+                }
+
+                return refreshAndGetCurrentAggregate(aggregator);
```
millems, line 139: Do we really need to re-aggregate with every profile file get()?

core/profiles/src/main/java/software/amazon/awssdk/profiles/ProfileFileSupplierBuilder.java
```diff
@@ -0,0 +1,109 @@
...
+import java.nio.file.Path;
+import java.time.Clock;
+import java.util.Objects;
+import java.util.function.Consumer;
+import java.util.function.Supplier;
+import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.profiles.internal.ProfileFileRefresher;
+
+@SdkInternalApi
+final class ProfileFileSupplierBuilder {
```
millems, line 27: Does anything use this class?

core/profiles/src/main/java/software/amazon/awssdk/profiles/internal/ProfileFileRefresher.java
```diff
@@ -0,0 +1,267 @@
...
+        if (isNewProfileFile(cachedOrRefreshedProfileFile)) {
+            currentRefreshRecord = cachedOrRefreshedRecord;
+        }
+
+        return cachedOrRefreshedProfileFile;
+    }
+
+    @Override
+    public void close() {
+        profileFileCache.close();
```
millems, line 86: We only need to close the cached supplier when it is async. I think it's fine to exclude this close for now so that we don't have to deal with managing the lifecycle of a profile file supplier.

core/profiles/src/main/java/software/amazon/awssdk/profiles/internal/ProfileFileRefresher.java
```diff
@@ -0,0 +1,267 @@
...
+
+    private ProfileFileRefresher(Builder builder) {
+        this.exceptionHandler = builder.exceptionHandler;
+        this.clock = builder.clock;
+        this.profileFile = builder.profileFile;
+        this.profileFilePath = builder.profileFilePath;
+        this.onProfileFileReload = builder.onProfileFileReload;
+        this.profileFileCache = CachedSupplier.builder(this::refreshResult)
+                                              .clock(this.clock)
+                                              .build();
```
millems, line 60: It doesn't look like we're actually using the cached supplier here for any of its caching functionality, just as a proxy for calling refreshResult. Do we actually need it, or can we rewrite this to not require it?

core/profiles/src/main/java/software/amazon/awssdk/profiles/internal/ProfileFileRefresher.java
```diff
@@ -0,0 +1,267 @@
...
+                                                                             .refreshTime(now)
+                                                                             .build();
+
+            return wrapIntoRefreshResult(refreshRecord, staleTime);
+        }
+    }
+
+    private RefreshResult<ProfileFileRefreshRecord> reloadAsRefreshResultIfStale() {
+        Instant now = clock.instant();
+        Instant staleTime = now;
```
millems, line 107: Why have two variables for the same value?

core/profiles/src/main/java/software/amazon/awssdk/profiles/internal/ProfileFileRefresher.java
```diff
@@ -0,0 +1,267 @@
...
+        profileFileCache.close();
+    }
+
+    private RefreshResult<ProfileFileRefreshRecord> refreshResult() {
+        try {
+            return reloadAsRefreshResultIfStale();
+        } catch (RuntimeException exception) {
+            Instant now = Instant.now();
+            Instant staleTime = now;
+            ProfileFile exceptionProfileFile = exceptionHandler.apply(exception);
```
millems, line 95: Why do we not just throw the exception? What's the purpose of an exception profile file?

core/profiles/src/main/java/software/amazon/awssdk/profiles/internal/ProfileFileRefresher.java
dave-fn: Yeah. We can probably discuss this offline, but yes, the original intent of `CachedSupplier` was to abstract away managing individually with the profile files. I may have chipped too much off and now it doesn't really make sense to use it.
One benefit of still using the cached supplier, if I understand correctly is to avoid refreshing a value when the request occurs within a timeframe (jitter time), which in this case means not hitting the disk excessively in a short amount of time.

core/profiles/src/main/java/software/amazon/awssdk/profiles/internal/ProfileFileRefresher.java
dave-fn: Removed

core/profiles/src/main/java/software/amazon/awssdk/profiles/internal/ProfileFileRefresher.java
dave-fn: Was being too verbose. Remove this and other instance.

core/profiles/src/main/java/software/amazon/awssdk/profiles/internal/ProfileFileRefresher.java
dave-fn: I was copying `CachedTokenRefresher`. I guess we could recover from an exception and provide a fallback profile file, maybe the default.

core/profiles/src/main/java/software/amazon/awssdk/profiles/ProfileFileSupplier.java
millems: Got it, yeah if we can avoid re-aggregating when the files haven't changed that's a good fix.

core/profiles/src/main/java/software/amazon/awssdk/profiles/ProfileFileSupplierBuilder.java
millems: Gotcha, I searched this CR to see if things were using it and I must have missed something using it.

core/profiles/src/main/java/software/amazon/awssdk/profiles/internal/ProfileFileRefresher.java
millems: Ah, I didn't think of that benefit. I think excluding jittering for local resources isn't a major risk. The failure mode is pretty different than a remote resource where jittering becomes important.

core/profiles/src/main/java/software/amazon/awssdk/profiles/internal/ProfileFileRefresher.java
millems: Interesting. I was thinking the intuitive behavior would just be to throw the exception (failing whatever profile file refresh is happening) and then let the next get() call try again. Essentially, push the failure handling upstream.

Trying to do something more intelligent might be confusing for the user to understand.

core/profiles/src/main/java/software/amazon/awssdk/profiles/ProfileFileSupplier.java
dave-fn: Reimplemented previous code that was looking at whether the individual files had changes. With this new version, we should not re-aggregate if not needed.

core/profiles/src/main/java/software/amazon/awssdk/profiles/internal/ProfileFileRefresher.java
dave-fn: Agreed this may be too complex. Removevd.

millems (approved):

L-Applin (approved):

Issue #3569, opened 2022-11-28
S3 PutObject with ChecksumAlgorithm Sets Empty Content-Encoding Header


iheffernan:
### Describe the bug

When performing an S3 PutObject operation with the `ChecksumAlgorithm` property set, an empty `Content-Encoding` header is set on the object in S3.

![Screenshot 2022-11-28 at 7 51 52 AM](https://user-images.githubusercontent.com/6098661/204282463-2c5a1758-e895-456a-acb7-6e4480dbfe11.png)

NOTE: If I manually compute and specify the checksum, the `Content-Encoding` header does NOT get set.

### Expected Behavior

`Content-Encoding` header should remain unset when using `ChecksumAlgorithm`.

### Current Behavior

No errors present. Only noticed because it was causing issues downstream when retrieving objects through the CloudFront.

### Reproduction Steps

```
PutObjectRequest.Builder builder = PutObjectRequest.builder().bucket(bucket).key(key);
builder.checksumAlgorithm(ChecksumAlgorithm.SHA256);
s3Client.putObject(builder.build(), RequestBody.fromByteBuffer(payload));
```

results in empty `Content-Encoding` System defined metadata when viewing in S3 console.

```
PutObjectRequest.Builder builder = PutObjectRequest.builder().bucket(bucket).key(key);
builder.checksumSHA256(createSHA256Checksum(payload).get());
s3Client.putObject(builder.build(), RequestBody.fromByteBuffer(payload));
```

results in no `Content-Encoding` System defined metadata in S3 console.

### AWS Java SDK version used

2.18.20

### JDK version used

17.0.3

### Operating System and version

MacOS Ventura 13.0.1

awanrky:
I see the same behavior, version 2.18.22, jdk 11.0.17

In addition, when trying to set the Content-Encoding while using checksumAlgorithm(), the system metadata will have a blank value instead of the value I try to set.

```
PutObjectRequest.Builder builder = PutObjectRequest.builder().bucket(bucket).key(key);
builder.checksumAlgorithm(ChecksumAlgorithm.SHA256);
builder.contentEncoding("gzip");
s3Client.putObject(builder.build(), RequestBody.fromByteBuffer(payload));
```

results in a blank Content-Encoding in the system metadata instead of the value of "gzip".  Leaving out the builder.checksumAlgorithm() line results in the value being set correctly.

debora-ito:
@iheffernan @awanrky thank you for reaching out.

[...]

PR #3720, opened 2023-01-26
Fix for Issue #3569 where the user set Content-Encoding was over writ…

joviegas:
## Motivation and Context

Fix for [Issue 3569](https://github.com/aws/aws-sdk-java-v2/issues/3569)

## Modifications

Checksum related interceptors and signers did a putHeader which over wrote the content-encoding header instead of overwriting it, Thus appendHeader is used in these places.     

-DefaultAwsCrtS3V4aSigner
- AbstractAwsS3V4Signer
- AsyncRequestBodyHttpChecksumTrailerInterceptor
- SyncHttpChecksumInTrailerInterceptor

### Fix
- The sdk will append the aws-chunked content-header value. Thus if user has set the content header as gzip then sdk will send this header as multiple header as mentioned in https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-streaming.html
 ```
Amazon S3 supports multiple content encodings. For example:
Content-Encoding : aws-chunked,gzip
That is, you can specify your custom content-encoding when using Signature Version 4 streaming API.
 ```

## Testing

- Added Junits
- Aded integ test

## Types of changes

## License

### DIFF
--- a/.changes/next-release/bugfix-AWSSDKforJavav2-16b5911.json
+++ b/.changes/next-release/bugfix-AWSSDKforJavav2-16b5911.json
@@ -1,0 +1,6 @@
+{
+    "type": "bugfix",
+    "category": "AWS SDK for Java v2",
+    "contributor": "",
+    "description": "Append Content-encoding header instead of over writing the header when Checksum algorithm is selected along with user set Content-encoding"
+}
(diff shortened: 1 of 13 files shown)


L-Applin:
Code looks good, but there are some things I'm confused about. Is this quote from the issue still correct?

> Expected Behavior
> Content-Encoding header should remain unset when using ChecksumAlgorithm.

Because, maybe Im not understanding correctly, it seems we are fixing an issue that overwrites the `Content-Encoding`, but does not address the fact that it has an empty value? Does the Header needs to be unset or have the `aws-chunked` value.

joviegas:
> Code looks good, but there are some things I'm confused about. Is this quote from the issue still correct?
> 
> > Expected Behavior
> > Content-Encoding header should remain unset when using ChecksumAlgorithm.
> 
> Because, maybe Im not understanding correctly, it seems we are fixing an issue that overwrites the `Content-Encoding`, but does not address the fact that it has an empty value? Does the Header needs to be unset or have the `aws-chunked` value.

The user is trying to say AWS-sdk-java should not overwrite(reset) the User set Cheunked endcoding value. The unset more of means "over write" here.
With the fix we will be appending the "aws-chunked" header and not reset/overwrite while sending the User set value to S3

L-Applin (approved):

PR #3748, opened 2023-02-07
Fix for Default MD5 check getting skipped even when http checksum algorithm is not set

joviegas:
## Motivation and Context

- By default for putRequest Md5 checksum validation should happen. By default as in when the new http checksum is not set to any algorithm.
- Currently it was skipping since the check was directly checking for ResolvedChecksum specs even when request algorithm were null.

- Note that this validation is done in S3 when httpChecksum algorithm is set.

## Modifications

- Added checks to check for requestAlgorithms.

## Testing

- Added new Test cases for putObject

## Types of changes

## License

### DIFF
--- a/.changes/next-release/bugfix-AWSSDKforJavav2-a15138d.json
+++ b/.changes/next-release/bugfix-AWSSDKforJavav2-a15138d.json
@@ -1,6 +1,6 @@
 {
     "type": "bugfix",
-    "category": "AWS SDK for Java v2",
+    "category": "Amazon S3",
     "contributor": "",
     "description": "Bug fix for Default Md5 checksum was getting skipped even when new http checksum was not set to any checksum algorithm."
 }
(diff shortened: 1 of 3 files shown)


.changes/next-release/bugfix-AWSSDKforJavav2-a15138d.json
```diff
@@ -0,0 +1,6 @@
+{
+    "type": "bugfix",
+    "category": "AWS SDK for Java v2",
```
millems, line 3: Is this just S3?

millems (approved):

.changes/next-release/bugfix-AWSSDKforJavav2-a15138d.json
joviegas: Done

PR #3754, opened 2023-02-09
ProfileFileSupplier aggregate files in order

dave-fn:
## Motivation and Context
The implementation of `aggregate` does not guarantee the profile files were submitted to `ProfileFile.Aggregator.addFile()` in the order given by the suppliers.
The current implementation uses a `Map` to keep track of whether a supplier returns a different object. Unfortunately, the `values()` method does not guarantee an ordering.

## Modifications
In order to preserve ordering and thread-safety, a `List` scoped to the method will be used to store the supplied profile files. If these need to be aggregated, the list will be used to feed the aggregator.

## Testing
Added two unit tests.

## Types of changes

## Checklist

## License

### DIFF
--- a/core/profiles/src/main/java/software/amazon/awssdk/profiles/ProfileFileSupplier.java
+++ b/core/profiles/src/main/java/software/amazon/awssdk/profiles/ProfileFileSupplier.java
@@ -16,11 +16,10 @@
 package software.amazon.awssdk.profiles;
 
 import java.nio.file.Path;
-import java.util.ArrayList;
-import java.util.List;
+import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.Objects;
 import java.util.Optional;
-import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Supplier;
 import software.amazon.awssdk.annotations.SdkPublicApi;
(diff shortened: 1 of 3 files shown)


core/profiles/src/main/java/software/amazon/awssdk/profiles/ProfileFileSupplier.java
```diff
@@ -124,29 +126,31 @@ static ProfileFileSupplier aggregate(ProfileFileSupplier... suppliers) {
...
+                nextValues.add(next);
                 ProfileFile current = currentValuesBySupplier.put(supplier, next);
 
                 return !Objects.equals(next, current);
             }
 
-            private void refreshCurrentAggregate() {
+            private void refreshCurrentAggregate(List<ProfileFile> nextValues) {
                 ProfileFile.Aggregator aggregator = ProfileFile.aggregator();
-                currentValuesBySupplier.values().forEach(aggregator::addFile);
```
dave-fn, line 149: This won't necessarily return the values in order since it's backed by a Map.

millems: Could we switch it to a linked hash map?

dave-fn: Yes. The thing is we lose the `Concurrency` from the current map, but we can synchronize manually. Would that be a preferred solution?

core/profiles/src/main/java/software/amazon/awssdk/profiles/ProfileFileSupplier.java
```diff
@@ -137,7 +138,7 @@ public ProfileFile get() {
                 return  currentAggregateProfileFile.get();
             }
 
-            private boolean didSuppliedValueChange(Supplier<ProfileFile> supplier) {
+            private synchronized boolean didSuppliedValueChange(Supplier<ProfileFile> supplier) {
```
millems, line 141: Should we just wrap currentValuesBySupplier in a Collections.synchronizedMap instead of synchronizing ourselves?

If there's some problem with that:

Can we synchronize just around the writes to the currentValuesBySupplier so that we don't hold the lock while calling the supplier?

```java
ProfileFile current;
synchronized (currentValuesBySupplier) {
    current = currentValuesBySupplier.put(supplier, next);
}
```

I think we also need to synchronize our iterating over the values on line 150, since modification of the collection during iteration is undefined.

```java
synchronized (currentValuesBySupplier) {
    currentValuesBySupplier.values().forEach(aggregator::addFile);
}
```

dave-fn: I agree the easiest solution is just to wrap into Collections.synchronizedMap.

Yep, I missed the synchronized keyword on line 150.

millems (approved):

PR #3767, opened 2023-02-17
[Bug bash] Include sso and ssooidc dependencies in maven archetype

L-Applin:
## Motivation and Context
New IAM best practices recommend using Identity Center credentials, which requires the `sso` and `ssooidc` dependencies

## Modifications
Add require property `ssooidc` to the maven archetype, which can take value `true` or `false`

## Testing
Added a project to testing with `ssooidc=true`


archetypes/archetype-app-quickstart/README.md
```diff
@@ -43,12 +43,13 @@ mvn archetype:generate \
...
       
 Parameter Name | Default Value | Description
----|---|---
+---|-----|---
 `service` (required) | n/a | Specifies the service client to be used in the application, eg: s3, dynamodb. Only one service should be provided. You can find available services [here][java-sdk-v2-services]. 
 `groupId`(required) | n/a | Specifies the group ID of the project
 `artifactId`(required) | n/a | Specifies the artifact ID of the project
 `nativeImage`(required)  | n/a | Specifies whether GraalVM Native Image configuration should be included
 `httpClient`(required) | n/a | Specifies the http client to be used by the SDK client. Available options are `url-connection-client` (sync), `apache-client` (sync), `netty-nio-client` (async). See [http clients][sdk-http-clients]
+`ssoIdc` (required) | n/a | Specify if the sso and ssoidc dependency should be include in the project. Include them to enabled the use of [IAM Identity center](https://aws.amazon.com/iam/identity-center/) for credentials.
```
cenedhryn, line 52: The term is `ssooidc` with an extra `o`!

archetypes/archetype-app-quickstart/src/test/resources/projects/ssoidc/reference/README.md
```diff
@@ -0,0 +1,53 @@
+# App
+
+This project contains a maven application with [AWS Java SDK 2.x](https://github.com/aws/aws-sdk-java-v2) dependencies.
```
cenedhryn, line 3: Do we say that it's using SSO/SSOOIDC or are the other projects also generic?

archetypes/archetype-app-quickstart/README.md
cenedhryn: Nit: spelling 
`include` should be `included`
`enabled` should be `enable`

archetypes/archetype-app-quickstart/src/test/resources/projects/ssoidc/reference/README.md
L-Applin: We don't specify which individual dependencies that are included for other property of the archetype, so I think we should leave it as-is

joviegas (commented):
Is there a way to append the data like append dependency to pom.xml rather than duplicating the entire pom.xml ?

L-Applin:
> Is there a way to append the data like append dependency to pom.xml rather than duplicating the entire pom.xml ?
Not sure I understand your point. The new dependency were added (appended?) to the pom.xml template. The other new pom.xml whole file is for testing the execution of the plugin with the new parameters.

L-Applin:
> Is there a way to append the data like append dependency to pom.xml rather than duplicating the entire pom.xml ?

Thanks John, but I am not sure I understand what you mean by that. The new dependency were added (appended?) to the pom.xml template. The other new pom.xml whole file is for testing the execution of the plugin with the new parameters.

joviegas:
> 

Thanks Olivier for the clarification , yeas since its a test case , adding new files to check the final output makes sense.
I was trying to figure out How are these generated files compared while testing?

archetypes/archetype-app-quickstart/README.md
```diff
@@ -43,12 +43,13 @@ mvn archetype:generate \
...
       
 Parameter Name | Default Value | Description
----|---|---
+---|-----|---
 `service` (required) | n/a | Specifies the service client to be used in the application, eg: s3, dynamodb. Only one service should be provided. You can find available services [here][java-sdk-v2-services]. 
 `groupId`(required) | n/a | Specifies the group ID of the project
 `artifactId`(required) | n/a | Specifies the artifact ID of the project
 `nativeImage`(required)  | n/a | Specifies whether GraalVM Native Image configuration should be included
 `httpClient`(required) | n/a | Specifies the http client to be used by the SDK client. Available options are `url-connection-client` (sync), `apache-client` (sync), `netty-nio-client` (async). See [http clients][sdk-http-clients]
+`ssooidc` (required) | n/a | Specify if the sso and ssoidc dependency should be included in the project. Include them to enable the use of [IAM Identity center](https://aws.amazon.com/iam/identity-center/) for credentials.
```
joviegas, line 52: Should we mention the sso and ssoidc or mention someting which is easier for user to undersatnd
"specifc SSO Credetial Provider or SSO TokenProvider related dependencies "

I guess this request is very specifc to SSO Token Provider.
 Note : If we User wants only SSO Credential provider than we donot need sso-oidc dependency , we can clarify this with the requester of this feature.

On other note we can make it generic so that we can add future dependencies related to Credentials providers like STS for example.

Example something like "Press  following option to auto add the depencires related to Credentials Providers 1 for SSO Credentials  2.  SSO Token Providers  3. STS Credentials providers"

when 1 we just add SSO 
when 2 we add SSO + SSO-OIDC
when 3 we add STS (for future not required in this PR)

L-Applin: How about `-DcredentialProvider=(default|identity-center)` for now, with `identity-center` adding both `sso` and `ssooidc` dependency

cenedhryn: In the future we'll need to widen the terminology to reference identity providers, but for now this should be fine. I think the refactoring of the SDK with regards to new modules will eventually change the setup scripts quite a lot.

cenedhryn (approved):

joviegas (approved):

PR #3769, opened 2023-02-17
New API for the retries module

sugmanue:
This new module includes the interfaces and classes that will be used to implement the new retry logic within the SDK.

## Motivation and Context

As part of the Smithy Reference Architecture this change adds a new module for retries. This change adds only the API part.

## Modifications
No modifications to existing code were made. We will make those changes in follow up pull requests.

## Testing
* Added unit tests for all the default methods in the `RetryStrategy.Builder` interface that create lambdas to be used to decide if an exception should be retried.
* Added unit tests for `ExponentialDelayWithJitter` and `FixedDelayWithJitterTest` similar to the tests present for its equivalent class [here](https://github.com/aws/aws-sdk-java-v2/blob/03f2db714bf987b59e8e5b7f6ec2d9b46f3126dc/core/sdk-core/src/test/java/software/amazon/awssdk/core/retry/backoff/FullJitterBackoffStrategyTest.java#L41)

N/A

## Types of changes

- [ X] New feature (non-breaking change which adds functionality)

## Checklist

- [ x] My change requires a change to the Javadoc documentation
- [ x] I have updated the Javadoc documentation accordingly
- [ x] I have added tests to cover my changes

## License


core/retries-api/src/main/java/software/amazon/awssdk/retriesapi/AcquireInitialTokenRequest.java
```diff
@@ -0,0 +1,36 @@
...
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.retriesapi;
```
zoewangg, line 16: Should it be `software.amazon.awssdk.retries.api`?

core/retries-api/src/main/java/software/amazon/awssdk/retriesapi/BackoffStrategy.java
```diff
@@ -0,0 +1,241 @@
...
+import software.amazon.awssdk.utils.ToString;
+import software.amazon.awssdk.utils.Validate;
+
+/**
+ * Determines how long to wait before each execution attempt.
+ */
+@SdkPublicApi
+@ThreadSafe
+public interface BackoffStrategy {
+    Duration BASE_DELAY_CEILING = Duration.ofMillis(Integer.MAX_VALUE); // Around ~24.8 days
```
zoewangg, line 34: These constants seem to be implementation details, should we move them to the classes that use them? If they are shared between classes, should we move them to a constant class?

core/retries-api/src/main/java/software/amazon/awssdk/retriesapi/TokenAcquisitionFailedException.java
```diff
@@ -0,0 +1,38 @@
...
+
+package software.amazon.awssdk.retriesapi;
+
+import software.amazon.awssdk.annotations.SdkPublicApi;
+
+/**
+ * Exception thrown by {@link RetryStrategy} when a new token cannot be acquired.
+ */
+@SdkPublicApi
+public final class TokenAcquisitionFailedException extends RuntimeException {
```
zoewangg, line 24: I wonder if it should extend SdkClientException to make it easier for customers to handle

zoewangg (commented):
Left some initial comments, still going through them.

core/retries-api/src/main/java/software/amazon/awssdk/retriesapi/AcquireInitialTokenRequest.java
sugmanue: I just realized that the original plan was to use instead `software.amazon.awssdk.api.retries`, I can go either way but I like the original idea better, thoughts?

core/retries-api/pom.xml
```diff
@@ -0,0 +1,72 @@
...
+            <artifactId>junit-jupiter</artifactId>
+            <version>${junit5.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
```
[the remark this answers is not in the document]
sugmanue: Sure, I will change it.

core/retries-api/src/main/java/software/amazon/awssdk/retriesapi/BackoffStrategy.java
```diff
@@ -0,0 +1,241 @@
...
+        @Override
+        public String toString() {
+            return "(Immediately)";
+        }
+    }
+
+    /**
+     * Strategy that waits for a period of time equal to the provided delay.
+     */
+    final class FixedDelayWithoutJitter implements BackoffStrategy {
```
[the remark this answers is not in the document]
sugmanue: Sure, I can move those into their own files.

core/retries-api/src/main/java/software/amazon/awssdk/retriesapi/RetryStrategy.java
```diff
@@ -0,0 +1,208 @@
...
+                }
+                return shouldRetry;
+            });
+        }
+
+        /**
+         * Configure the strategy to retry the root cause of a failure (the final cause) a failure exception is an instance of to
+         * the provided class (includes subtypes).
+         */
+        default Builder retryOnRootCauseInstanceOf(Class<? extends Throwable> throwable) {
```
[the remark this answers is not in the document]
sugmanue: I guess that some of these conditions are subtle and error prone, having these spell out  with proper names make the API easier to use and understand, but I can remove some of these. Which ones do you think it will be worth keeping and which ones do you think we can go without?

core/retries-api/src/main/java/software/amazon/awssdk/retriesapi/RetryStrategy.java
```diff
@@ -0,0 +1,208 @@
...
+         */
+        default Builder retryOnException(Class<? extends Throwable> throwable) {
+            return retryOnException(t -> t.getClass() == throwable);
+        }
+
+        /**
+         * Configure the strategy to retry when a failure exception class is an instance of the provided class (includes
+         * subtypes).
+         */
+        default Builder retryOnExceptionInstanceOf(Class<? extends Throwable> throwable) {
```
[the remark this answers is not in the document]
sugmanue: I guess the argument in favor would be that it will make some expressions easier but the same can be accomplish by calling this method for each. Since I'm not quite sure that this will add much I think it would be better to do that when and if the need araises.

core/retries-api/src/main/java/software/amazon/awssdk/retriesapi/TokenAcquisitionFailedException.java
sugmanue: I guess we could depend on `core`, but not sure if we want to complect those. This package for now only depends on `annotations` and `utils` and I feel that adding `sdk-core` will make it a lot less easier to reuse outside the SDK.

core/retries-api/pom.xml
```diff
@@ -0,0 +1,68 @@
...
+    <name>AWS Java SDK :: Retries API</name>
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-jar-plugin</artifactId>
+                <configuration>
+                    <archive>
+                        <manifestEntries>
+                            <Automatic-Module-Name>software.amazon.awssdk.retriesapi</Automatic-Module-Name>
```
zoewangg, line 37: This needs to match with the top-level package name.

zoewangg (approved):

PR #3771, opened 2023-02-17
Improve Javadoc for request-level timeout

zoewangg:
## Motivation and Context
Improve Javadoc for request-level timeout


core/sdk-core/src/main/java/software/amazon/awssdk/core/RequestOverrideConfiguration.java
```diff
@@ -120,6 +124,9 @@ public Optional<Duration> apiCallTimeout() {
      * request
      * (i.e. each retry) and the total time spent on all requests across retries (i.e. the 'api call' time).
      *
+     * <p>
+     * This will override the {@link ClientOverrideConfiguration#apiCallAttemptTimeout()} if it is configured.
+     *
```
millems, line 129: Should this be added to the builder's setters as well? People are more likely to read the javadoc on the method they're calling (on the builder) instead of the method that the SDK is using (on the override configuration).

zoewangg: Ah yeah, I thought I did, will update

millems (approved):

PR #3822, opened 2023-03-09
Update tm builder log messages

davidh44:
## Motivation and Context

Original warn message can be misleading for other clients such as encryption client

## Modifications

Updating logging in Transfer Manager builder so that warn level message is logged only for Java based S3 client. Adding OTHER enum to S3ClientType, and debug level message is logged for other S3 clients such as encryption client.

## Types of changes

## Checklist

## License

### DIFF
--- a/services-custom/s3-transfer-manager/pom.xml
+++ b/services-custom/s3-transfer-manager/pom.xml
@@ -212,6 +212,17 @@
             <version>${commons-codec.verion}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>io.github.hakky54</groupId>
+            <artifactId>logcaptor</artifactId>
+            <version>2.8.0</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+            <version>2.0.6</version>
+        </dependency>
     </dependencies>
 
     <build>
--- a/services-custom/s3-transfer-manager/pom.xml
+++ b/services-custom/s3-transfer-manager/pom.xml
@@ -212,17 +212,6 @@
             <version>${commons-codec.verion}</version>
             <scope>test</scope>
         </dependency>
-        <dependency>
-            <groupId>io.github.hakky54</groupId>
-            <artifactId>logcaptor</artifactId>
-            <version>${logcaptor.version}</version>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>org.slf4j</groupId>
-            <artifactId>slf4j-api</artifactId>
-            <version>2.0.6</version>
-        </dependency>
     </dependencies>
 
     <build>
--- a/services-custom/s3-transfer-manager/pom.xml
+++ b/services-custom/s3-transfer-manager/pom.xml
@@ -215,7 +215,7 @@
         <dependency>
             <groupId>io.github.hakky54</groupId>
             <artifactId>logcaptor</artifactId>
-            <version>2.8.0</version>
+            <version>${logcaptor.version}</version>
             <scope>test</scope>
         </dependency>
         <dependency>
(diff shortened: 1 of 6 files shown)


services-custom/s3-transfer-manager/pom.xml
```diff
@@ -212,6 +212,17 @@
...
+        <dependency>
+            <groupId>io.github.hakky54</groupId>
+            <artifactId>logcaptor</artifactId>
+            <version>2.8.0</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+            <version>2.0.6</version>
```
zoewangg, line 224: Is version needed? Can't it inherit from parent pom?

services-custom/s3-transfer-manager/pom.xml
```diff
@@ -212,6 +212,17 @@
             <version>${commons-codec.verion}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>io.github.hakky54</groupId>
+            <artifactId>logcaptor</artifactId>
+            <version>2.8.0</version>
```
zoewangg, line 218: Let's manage versions in one place https://github.com/aws/aws-sdk-java-v2/blob/master/pom.xml#L119

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/DefaultS3TransferManager.java
```diff
@@ -120,12 +120,15 @@ public DefaultS3TransferManager(DefaultBuilder tmBuilder) {
 
         if (s3AsyncClient instanceof S3CrtAsyncClient) {
             s3ClientType = S3ClientType.CRT_BASED;
-        } else {
+        } else if (s3AsyncClient.getClass().getSimpleName().equals("DefaultS3AsyncClient")){
```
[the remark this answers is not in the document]
davidh44: updated to getName

services-custom/s3-transfer-manager/pom.xml
davidh44: updated

services-custom/s3-transfer-manager/pom.xml
davidh44: logcaptor plugin needs 2.x.x version, while parent uses 1.7.30. Upgrading to 2.x.x causes breaking changes

zoewangg: Hmm, I see, then we can't use it unfortunately, because it will break customers. We could potentially making it a test dependency, but this could be error-prone since we could get confused and just remove the test scope in the future.

services-custom/s3-transfer-manager/src/test/java/software/amazon/awssdk/transfer/s3/progress/LoggingTransferListenerTest.java
```diff
@@ -19,11 +19,11 @@
 import static org.mockito.Mockito.mock;
 
 import java.util.List;
-import org.apache.logging.log4j.Level;
+import nl.altindag.log.LogCaptor;
 import org.apache.logging.log4j.core.LogEvent;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
-import software.amazon.awssdk.testutils.LogCaptor;
```
zoewangg, line 26: Nice, I didn't know we have LogCaptor, can we use it instead?

services-custom/s3-transfer-manager/pom.xml
davidh44: got it, removed plugin and dependencies

services-custom/s3-transfer-manager/src/test/java/software/amazon/awssdk/transfer/s3/progress/LoggingTransferListenerTest.java
davidh44: yep, using LogCaptor from testutils instead

zoewangg (approved):

PR #3829, opened 2023-03-10
Accept and use the new AWS Credentials interfaces

gosar:
## Motivation and Context

As part of SRA, earlier PR added the new interfaces. This PR updates existing code to accept and use the new interfaces.

## Modifications

This PR updates existing interfaces like AwsClientBuilder and others where customers used to be able to provide AwsCredentialsProvider, to be able to provide the new interface - `IdentityProvider<? extends AwsCredentialsIdentity>`.

Pending: 
* ~~Code generator changes for endpoint discovery case - SyncClientClass.~~
* I'm still working on adding tests. 
* Also need to update few more classes like AwsCredentialsProviderChain called out in design doc, and confirm if there are any more places. But to wrap up this PR and not make it bigger, will do that in separate PR.
* Will update code for new Token interfaces later too, in separate PR.
* Potentially move CREDENTIALS_IDENTITY_PROVIDER into SdkClientBuilder to support generic clients using AWS Credentials. Maybe in separate PR.

## Testing

Updated existing tests. Working on adding new tests. Sending PR out as draft to start code review.

## Types of changes

## Checklist

        Ran `./mvnw clean install -pl :auth,:aws-core,:s3,:polly,:rds -am -Pquick` locally

## License


core/aws-core/src/main/java/software/amazon/awssdk/awscore/client/builder/AwsDefaultClientBuilder.java
```diff
@@ -191,7 +193,8 @@ protected final SdkClientConfiguration finalizeChildConfiguration(SdkClientConfi
         configuration = mergeSmartDefaults(configuration);
 
         return configuration.toBuilder()
-                            .option(AwsClientOption.CREDENTIALS_PROVIDER, resolveCredentials(configuration))
```
millems, line 194: Unfortunately we have to keep setting this field, since it's a protected API and old client versions would still rely on it being there. Another thing we can clean up later?

core/aws-core/src/main/java/software/amazon/awssdk/awscore/client/builder/AwsDefaultClientBuilder.java
```diff
@@ -349,9 +352,10 @@ private Boolean resolveUseFipsFromDefaultProvider(SdkClientConfiguration config)
     /**
      * Resolve the credentials that should be used based on the customer's configuration.
      */
-    private AwsCredentialsProvider resolveCredentials(SdkClientConfiguration config) {
-        return config.option(AwsClientOption.CREDENTIALS_PROVIDER) != null
-               ? config.option(AwsClientOption.CREDENTIALS_PROVIDER)
+    private IdentityProvider<? extends AwsCredentialsIdentity> resolveCredentialsIdentityProvider(SdkClientConfiguration config) {
+        // Note, that CREDENTIALS_PROVIDER is never set. It is replaced with CREDENTIALS_IDENTITY_PROVIDER, so just check that
```
millems, line 356: These types of comments are probably better placed on the option itself than the place we're using it.

core/aws-core/src/main/java/software/amazon/awssdk/awscore/client/builder/AwsDefaultClientBuilder.java
```diff
@@ -419,16 +423,34 @@ public final void setFipsEnabled(Boolean fipsEndpointEnabled) {
...
     public final BuilderT credentialsProvider(AwsCredentialsProvider credentialsProvider) {
-        clientConfiguration.option(AwsClientOption.CREDENTIALS_PROVIDER, credentialsProvider);
-        return thisBuilder();
+        // clientConfiguration.option(AwsClientOption.CREDENTIALS_PROVIDER, credentialsProvider);
+        // return thisBuilder();
+        // TODO: by delegating to {@link #credentialsProvider(IdentityProvider)} we longer set
+        //       AwsClientOption.CREDENTIALS_PROVIDER, thus removing need for validating that only one of the 2 is set.
+        //       It may even let us remove the old AwsClientOption (or change the type) - maybe?
+        return credentialsProvider((IdentityProvider<? extends AwsCredentialsIdentity>) credentialsProvider);
     }
```
millems, line 436: Yeah, if the parent interface is doing this with a default method that has the same signature, we can remove this.

core/aws-core/src/main/java/software/amazon/awssdk/awscore/client/builder/AwsDefaultClientBuilder.java
```diff
@@ -419,16 +423,34 @@ public final void setFipsEnabled(Boolean fipsEndpointEnabled) {
...
 
+    // TODO: Not sure where the setter is used. Is it ok to delegate to new type (transitively from
+    //       {@link #credentialsProvider(AwsCredentialsIdentity))}?
     public final void setCredentialsProvider(AwsCredentialsProvider credentialsProvider) {
         credentialsProvider(credentialsProvider);
     }
```
millems, line 442: This is used by spring-XML customers, reflectively. You can just delegate it through, yeah. It's just a bean-style alias for `credentialsProvider(AwsCredentialsProvider)`

core/aws-core/src/main/java/software/amazon/awssdk/awscore/client/config/AwsClientOption.java
```diff
@@ -21,16 +21,29 @@
...
+    // TODO: Should the existing option be removed?
+    //       Or replaced with same name (CREDENTIALS_PROVIDER) but new type below?
+    //       This class is SdkProtectedApi, and it seems customer cannot create an option with this directly (we do it in
+    //       AwsDefaultClientBuilder). Is there risk of old generated code when there are mixed versions of modules?
+    // If it is kept, should it be marked @Deprecated?
     /**
      * @see AwsClientBuilder#credentialsProvider(AwsCredentialsProvider)
      */
     public static final AwsClientOption<AwsCredentialsProvider> CREDENTIALS_PROVIDER =
             new AwsClientOption<>(AwsCredentialsProvider.class);
```
millems, line 39: Discussed this one a bit above. +1 to marking it as deprecated, but like I mentioned I think we need to keep it around and keep setting it. We also need to keep reading it, in case customers are using old client modules with the updated aws-core module.

core/aws-core/src/main/java/software/amazon/awssdk/awscore/internal/authcontext/AuthorizationStrategyFactory.java
```diff
@@ -64,7 +65,8 @@ private TokenAuthorizationStrategy tokenAuthorizationStrategy() {
 
     private AwsCredentialsAuthorizationStrategy awsCredentialsAuthorizationStrategy() {
         Signer defaultSigner = clientConfiguration.option(SdkAdvancedClientOption.SIGNER);
-        AwsCredentialsProvider defaultCredentialsProvider = clientConfiguration.option(AwsClientOption.CREDENTIALS_PROVIDER);
+        IdentityProvider<? extends AwsCredentialsIdentity> defaultCredentialsProvider =
+            clientConfiguration.option(AwsClientOption.CREDENTIALS_IDENTITY_PROVIDER);
```
millems, line 69: If CREDENTIALS_IDENTITY_PROVIDER is null, we should fall back to CREDENTIALS_PROVIDER, in case they're using an old client version with newer core.

core/aws-core/src/main/java/software/amazon/awssdk/awscore/internal/authcontext/AwsCredentialsAuthorizationStrategy.java
```diff
@@ -73,8 +75,15 @@ public Signer resolveSigner() {
...
-        AwsCredentials credentials = resolveCredentials(credentialsProvider, metricCollector);
+        IdentityProvider<? extends AwsCredentialsIdentity> credentialsProvider =
+            resolveCredentialsProvider(request, defaultCredentialsProvider);
+        // TODO: Other option is casting:
+        //       AwsCredentials credentials = (AwsCredentials) resolveCredentials(credentialsProvider, metricCollector);
+        //       This may be unsafe if there is a sub-type of AwsCredentialsIdentity that is not AwsCredentials.
+        //       Like if customer creates an OdinAwsCredentialsIdentity which directly is-a AwsCredentialsIdentity.
+        AwsCredentials credentials = CredentialUtils.toCredentials(resolveCredentials(credentialsProvider, metricCollector));
+
+        // TODO: Should the signer be changed to use AwsCredentialsIdentity? Maybe with Signer SRA work, not now.
```
millems, line 86: The signer definitely should... but I think it's okay to keep it as a TODO here in this PR.

+1 to not downcasting. Hopefully when we refactor the signers, we won't have to do these conversions any more.

core/aws-core/src/main/java/software/amazon/awssdk/awscore/internal/authcontext/AwsCredentialsAuthorizationStrategy.java
```diff
@@ -84,22 +93,27 @@ public void addCredentialsToExecutionAttributes(ExecutionAttributes executionAtt
...
-                                                     MetricCollector metricCollector) {
+    private static AwsCredentialsIdentity resolveCredentials(
+            IdentityProvider<? extends AwsCredentialsIdentity> credentialsProvider,
+            MetricCollector metricCollector) {
         Validate.notNull(credentialsProvider, "No credentials provider exists to resolve credentials from.");
 
-        Pair<AwsCredentials, Duration> measured = MetricUtils.measureDuration(credentialsProvider::resolveCredentials);
+        // TODO: Exception handling for join()?
+        Pair<? extends AwsCredentialsIdentity, Duration> measured =
+            MetricUtils.measureDuration(() -> credentialsProvider.resolveIdentity().join());
```
millems, line 113: Can we avoid joining at all for async? Or is that too much work (a TODO)? 

If we need to join for now, can we use CompletableFutureUtils.joinInterruptibly and then unwrap CompletionExceptions that have a RuntimeException or Error cause?

e.g.

```java
CompletableFutureUtils.joinLikeSync(credentialsProvider::resolveIdentity);

/**
 * Joins (interruptibly) on the future, and re-throws any RuntimeExceptions or Errors just like the async task would have thrown if it was executed synchronously.
 */
<T> T joinLikeSync(CompletableFuture<T> future) {
    try {
        return CompletableFutureUtils.joinInterruptibly(future);
    } catch (CompletionException e) {
        Throwable cause = e.getCause();
        if (cause instanceof RuntimeException) {
            cause.addSuppressed(new CompletionException("Task failed.")); // Make sure we don't lose the context of where the join is in the stack...
            throw (RuntimeException) cause;
        }
    }
}
```

core/aws-core/src/main/java/software/amazon/awssdk/awscore/internal/client/config/AwsClientOptionValidation.java
```diff
@@ -39,8 +39,8 @@ public static void validateSyncClientOptions(SdkClientConfiguration c) {
     }
 
     private static void validateClientOptions(SdkClientConfiguration c) {
-        require("credentialsProvider", c.option(AwsClientOption.CREDENTIALS_PROVIDER));
-
+        // TODO: where does the field name come from? Should it be "credentialsIdentityProvider"?
+        require("credentialsProvider", c.option(AwsClientOption.CREDENTIALS_IDENTITY_PROVIDER));
```
millems, line 43: This is just for helpful error messages. I think `credentialsProvider` is correct, since that's the setter we have on the AWS client.

core/aws-core/src/main/java/software/amazon/awssdk/awscore/internal/client/config/AwsClientOptionValidation.java
```diff
@@ -39,8 +39,8 @@ public static void validateSyncClientOptions(SdkClientConfiguration c) {
     }
 
     private static void validateClientOptions(SdkClientConfiguration c) {
-        require("credentialsProvider", c.option(AwsClientOption.CREDENTIALS_PROVIDER));
```
millems, line 42: Again, I think we need to fall back to this if CREDENTIALS_IDENTITY_PROVIDER is null.

core/aws-core/src/test/java/software/amazon/awssdk/awscore/internal/authcontext/AwsCredentialsAuthorizationStrategyTest.java
```diff
@@ -41,14 +42,17 @@ public class AwsCredentialsAuthorizationStrategyTest {
...
+    AwsCredentialsProvider credentialsProvider;
+    AwsCredentials credentials;
     @Mock MetricCollector metricCollector;
 
     @Before
     public void setUp() throws Exception {
         when(sdkRequest.overrideConfiguration()).thenReturn(Optional.empty());
-        when(credentialsProvider.resolveCredentials()).thenReturn(credentials);
+        // TODO: The below with using @Mock for credentialsProvider and credentials was giving a compile error, so not using mock
+        // when(credentialsProvider.resolveIdentity()).thenReturn(CompletableFuture.completedFuture(credentials));
```
millems, line 53: I think that's fine. Do we need this TODO?

pom.xml
```diff
@@ -634,6 +634,9 @@
                             <!-- Moving the interface methods to new SRA super interfaces causes japicmp to complain -->
                             <exclude>software.amazon.awssdk.auth.credentials.AwsCredentials</exclude>
                             <exclude>software.amazon.awssdk.auth.token.credentials.SdkToken</exclude>
+
+                            <!-- Changing isAnonymous() to use the new SRA super interfaces causes japicmp to complain -->
+                            <exclude>software.amazon.awssdk.auth.credentials.CredentialUtils</exclude>
```
millems, line 639: Yeah, it was right to complain since that's a protected class (see my comment above). Binary backwards-compatibility doesn't allow for changing parameter types to super-types or return types to sub-types, even though intuition would say it's "safe" to do.

services/polly/src/main/java/software/amazon/awssdk/services/polly/internal/presigner/DefaultPollyPresigner.java
```diff
@@ -112,7 +115,10 @@ public Region region() {
         return region;
     }
 
-    public AwsCredentialsProvider credentialsProvider() {
+    // TODO: Why are these 3 getters public? And why methods instead of accessing the members directly - used only inside this
+    //       class. And this class is @SdkInternalApi
+    //       Maybe package scope for unit test
```
millems, line 120: +1 since it's internal, feel free to be merciless.

services/polly/src/main/java/software/amazon/awssdk/services/polly/internal/presigner/DefaultPollyPresigner.java
```diff
@@ -212,11 +218,16 @@ private ExecutionAttributes createExecutionAttributes(PresignRequest presignRequ
...
-        return request.overrideConfiguration().flatMap(AwsRequestOverrideConfiguration::credentialsProvider)
+    private IdentityProvider<? extends AwsCredentialsIdentity> resolveCredentialsProvider(PollyRequest request) {
+        return request.overrideConfiguration().flatMap(AwsRequestOverrideConfiguration::credentialsIdentityProvider)
                 .orElse(credentialsProvider());
     }
 
+    private AwsCredentials resolveCredentials(IdentityProvider<? extends AwsCredentialsIdentity> credentialsProvider) {
+        AwsCredentialsIdentity credentials = credentialsProvider.resolveIdentity().join();
+        return CredentialUtils.toCredentials(credentials);
+    }
```
millems, line 229: Can we return and use AwsCredentialsIdentity directly instead of AwsCredentials?

services/rds/src/main/java/software/amazon/awssdk/services/rds/DefaultRdsUtilities.java
```diff
@@ -105,13 +108,13 @@ private Region resolveRegion(GenerateAuthenticationTokenRequest request) {
...
-    private AwsCredentialsProvider resolveCredentials(GenerateAuthenticationTokenRequest request) {
+    private AwsCredentials resolveCredentials(GenerateAuthenticationTokenRequest request) {
         if (request.credentialsProvider() != null) {
-            return request.credentialsProvider();
+            return request.credentialsProvider().resolveCredentials();
         }
 
         if (this.credentialsProvider != null) {
-            return this.credentialsProvider;
+            return CredentialUtils.toCredentials(this.credentialsProvider.resolveIdentity().join());
```
millems, line 117: TODO: Hopefully we can remove this once we migrate the signers?

services/rds/src/main/java/software/amazon/awssdk/services/rds/DefaultRdsUtilities.java
```diff
@@ -121,13 +124,14 @@ private AwsCredentialsProvider resolveCredentials(GenerateAuthenticationTokenReq
...
         private Region region;
-        private AwsCredentialsProvider credentialsProvider;
+        private IdentityProvider<? extends AwsCredentialsIdentity> credentialsProvider;
 
         DefaultBuilder() {
         }
 
         Builder clientConfiguration(SdkClientConfiguration clientConfiguration) {
-            this.credentialsProvider = clientConfiguration.option(AwsClientOption.CREDENTIALS_PROVIDER);
+            // TODO: update
```
millems, line 133: can you elaborate?

services/rds/src/main/java/software/amazon/awssdk/services/rds/DefaultRdsUtilities.java
```diff
@@ -139,8 +143,9 @@ public Builder region(Region region) {
             return this;
         }
 
+        // TODO: {@link #credentialsProvider(AwsCredentialsProvider)} uses the default implementation which call this
```
millems, line 146: Is this TODO needed?

services/rds/src/main/java/software/amazon/awssdk/services/rds/model/GenerateAuthenticationTokenRequest.java
```diff
@@ -77,6 +80,8 @@ public Region region() {
      * @return The credentials provider to sign the IAM auth request with. If specified, takes precedence over the value
      * specified in {@link RdsUtilities.Builder#credentialsProvider(AwsCredentialsProvider)}}
      */
+    // TODO: should another method returning IdentityProvider<? extends AwsCredentialsIdentity> be added and used in
+    //       DefaultRdsUtilities?
```
millems, line 84: Yes, I think so.

gosar (commented):
There are a few comments from Matt still pending, but responding with some here.

core/aws-core/src/main/java/software/amazon/awssdk/awscore/client/builder/AwsDefaultClientBuilder.java
millems: That is probably sufficient? I don't think we need to set CREDENTIALS_PROVIDER before configuration finalization, but we can check the child classes to be sure. I don't think there are old versions at play for this field that we need to be defensive about.

If it's easier to just set both in the `credentialsProvider` methods than figuring out how to play it safe, I don't think it's a problem.

core/aws-core/src/main/java/software/amazon/awssdk/awscore/internal/authcontext/AwsCredentialsAuthorizationStrategy.java
gosar: Discussed offline, will leave all `join()`s in this PR for now, but will revisit how to convert these to async where possible before we are decide on releasing the Identity changes.

core/aws-core/src/main/java/software/amazon/awssdk/awscore/internal/client/config/AwsClientOptionValidation.java
gosar: Confirmed in offline discussion - fallback not needed.

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/client/test-endpoint-discovery-async.java
```diff
@@ -174,9 +174,9 @@ public CompletableFuture<TestDiscoveryIdentifiersRequiredResponse> testDiscovery
             URI cachedEndpoint = null;
             if (endpointDiscoveryEnabled) {
                 String key = testDiscoveryIdentifiersRequiredRequest.overrideConfiguration()
-                                                                    .flatMap(AwsRequestOverrideConfiguration::credentialsProvider)
-                                                                    .orElseGet(() -> clientConfiguration.option(AwsClientOption.CREDENTIALS_PROVIDER)).resolveCredentials()
-                                                                    .accessKeyId();
+                                                                    .flatMap(AwsRequestOverrideConfiguration::credentialsIdentityProvider)
+                                                                    .orElseGet(() -> clientConfiguration.option(AwsClientOption.CREDENTIALS_IDENTITY_PROVIDER))
+                                                                    .resolveIdentity().join().accessKeyId();
```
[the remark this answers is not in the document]
gosar: Discussed offline, will leave all `join()`s in this PR for now, but will revisit how to convert these to async where possible before we are decide on releasing the Identity changes.

core/aws-core/src/main/java/software/amazon/awssdk/awscore/internal/authcontext/AuthorizationStrategyFactory.java
gosar: Discussed offline - fallback not needed. Don't think child classes change the provider.

gosar: Confirmed after generated all AWS services, that nothing references AwsClientOption.CREDENTIALS_PROVIDER. Only usage of that is in AwsDefaultClientBuilder.finalizeChildConfiguration to set this option for older generated clients. So confirmed that no fallback is needed in the AuthorizationStratergyFactory, so removing it.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/DefaultS3CrtAsyncClient.java
```diff
@@ -110,44 +111,17 @@ private static S3CrtAsyncHttpClient.Builder initializeS3CrtAsyncHttpClient(Defau
...
-        }
-
-        public URI endpointOverride() {
-            return endpointOverride;
-        }
-
-        public Long readBufferSizeInBytes() {
-            return readBufferSizeInBytes;
-        }
-
```
cenedhryn, line 148: Why are these methods removed?

gosar: They are unused methods. The constructor logic of the containing class DefaultS3CrtAsyncClient accesses the private memebers of the builder directly (e.g., `builder.region`). And DefaultS3CrtAsyncClient is `@SdkInternalApi` so thought it is safe to remove them. LMK if I shouldn't.

zoewangg: 🤦🏼‍♀️ I don't know why I added them in the first place(maybe I just accidentally generated them on Intellij) Agreed that we should remove them, but we probably should not do it as part of this PR in the spirit of keeping PRs small.

cenedhryn (approved):

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/CredentialUtils.java
```diff
@@ -28,6 +31,72 @@ private CredentialUtils() {
...
+            AwsSessionCredentialsIdentity awsSessionCredentialsIdentity = (AwsSessionCredentialsIdentity) awsCredentialsIdentity;
+            return AwsSessionCredentials.create(awsSessionCredentialsIdentity.accessKeyId(),
+                                                awsSessionCredentialsIdentity.secretAccessKey(),
+                                                awsSessionCredentialsIdentity.sessionToken());
+        }
+        if (isAnonymous(awsCredentialsIdentity)) {
+            return AwsBasicCredentials.ANONYMOUS_CREDENTIALS;
+        }
+        return AwsBasicCredentials.create(awsCredentialsIdentity.accessKeyId(),
+                                          awsCredentialsIdentity.secretAccessKey());
```
millems, line 77: Should we have `AwsCredentials` extend `AwsCredentialsIdentity` and `AwsSessionCredentials` extend `AwsSessionCredentialsIdentity` (if we don't already)?

If so, then should we do `if (awsCredentialsIdentity instanceof AwsCredentials) { return (AwsCredentials) awsCredentialsIdentity; }` to avoid copying the data for users of the old type?

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/DefaultS3CrtAsyncClient.java
gosar: I needed to change the return type of the `credentialsProvider()` for this PR to the new type, but this method was unused, so removed it, along with the other unused ones. I can move the removals to a separate PR.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/AwsCredentialsProviderChain.java
```diff
@@ -45,6 +45,7 @@
  * <p>This chain implements {@link AutoCloseable}. When closed, it will call the {@link AutoCloseable#close()} on any credential
  * providers in the chain that need to be closed.</p>
  */
+// TODO: deprecate with new IdentityProvider chain in identity-spi
```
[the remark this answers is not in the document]
gosar: Yeah, based on the offline discussion today, not planning to deprecate this. I'll move all changes related to this class in separate PR, so I'll remove this TODO from this PR.

core/aws-core/src/main/java/software/amazon/awssdk/awscore/client/config/AwsClientOption.java
```diff
@@ -21,16 +21,29 @@
...
 @SdkProtectedApi
 public final class AwsClientOption<T> extends ClientOption<T> {
     /**
+     // * This option is deprecated in favor of {@link #CREDENTIALS_IDENTITY_PROVIDER}.
      * @see AwsClientBuilder#credentialsProvider(AwsCredentialsProvider)
      */
+    @Deprecated
+    // smithy codegen TODO: This could be removed when doing a minor version bump where we told customers we'll be breaking
+    //  protected APIs. Postpone this to when we do Smithy code generator migration, where we'll likely have to start
+    //  breaking a lot of protected things.
```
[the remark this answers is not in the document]
gosar: It's in comments, so shouldn't affect javadoc. Either ways, if you think I should do something different here, lmk.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/CredentialUtils.java
gosar: Good call! I was worried about extra conversions, but didn't think of checking here. Thanks for the suggestion. Will update.

gosar: And they do extend already.

core/aws-core/src/main/java/software/amazon/awssdk/awscore/client/config/AwsClientOption.java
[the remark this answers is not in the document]
millems: Ah, I thought that since this comment separated the field from the javadoc that it would mess up the rendering.

millems (approved):

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/CredentialUtils.java
gosar: Oh, adding similar check to `toCredentialsProvider` too.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/DefaultS3CrtAsyncClient.java
gosar: Moved to separate PR https://github.com/aws/aws-sdk-java-v2/pull/3879

Issue #3850, opened 2023-03-21
S3TransferManager got stuck while downloading multiple directories


caroline202210:
### Describe the bug

Hi team,

Our service is using S3TransferManager to download multiple directories from an S3 bucket. The progress stuck on the following stack trace:
```
sdk-async-response-2-7" #345 daemon prio=5 os_prio=0 cpu=1504829.76ms elapsed=1613.05s tid=0x00007ff554188800 nid=0x2851 runnable  [0x00007ff546d76000]
   java.lang.Thread.State: RUNNABLE
        at software.amazon.awssdk.transfer.s3.internal.AsyncBufferingSubscriber.flushBufferIfNeeded(AsyncBufferingSubscriber.java:104)
        at software.amazon.awssdk.transfer.s3.internal.AsyncBufferingSubscriber.onComplete(AsyncBufferingSubscriber.java:150)
        at software.amazon.awssdk.utils.async.DelegatingSubscriber.onComplete(DelegatingSubscriber.java:47)
        at software.amazon.awssdk.core.pagination.async.PaginationSubscription.completeSubscription(PaginationSubscription.java:81)
        at software.amazon.awssdk.core.internal.pagination.async.ItemsSubscription.handleRequests(ItemsSubscription.java:51)
        at software.amazon.awssdk.core.internal.pagination.async.ItemsSubscription.sendNextElement(ItemsSubscription.java:104)
...
```
All 8 CPU cores are running the code above so the CPU usage is about 800%.

The threads above might prevent other threads from running. For example, the following RUNNABLE thread didn't get chance to run since the cpu time is only 0.25ms:
```
"udp-epoll-6" #48 daemon prio=5 os_prio=0 cpu=0.25ms elapsed=2117.01s tid=0x00007ff68e112000 nid=0xfbe runnable  [0x00007ff5d598e000]
   java.lang.Thread.State: RUNNABLE
        at io.micrometer.shaded.io.netty.channel.epoll.Native.epollWait(Native Method)
        at io.micrometer.shaded.io.netty.channel.epoll.Native.epollWait(Native.java:209)
        at io.micrometer.shaded.io.netty.channel.epoll.Native.epollWait(Native.java:202)
        at io.micrometer.shaded.io.netty.channel.epoll.EpollEventLoop.epollWaitNoTimerChange(EpollEventLoop.java:306)
[...]

PR #3867, opened 2023-03-29
TM: downloadDirectory refactor how SDK sends concurrent download file requests

zoewangg:
## Motivation and Context
#3850 

## Modifications
~~Stop polling from the request queue if there are still requests in flight in download directory to prevent potential thread starvation issue.~~

This issue could surface when upstream has finished processing (all S3 objects have been retrieved from paginated ListObjectsV2Requests). If all downloadFile requests are being proceed at the moment, the thread that invoked onComplete would continue to poll from the queue until all downloadFile requests have been finished, which could cause thread starvation and eventually cause the requests to get stuck.

Refactored the publisher to not use StoringSubscriber. 

## Testing
Tested locally

## Types of changes

## Checklist

## License


services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/AsyncBufferingSubscriber.java
```diff
@@ -65,89 +57,40 @@ public void onSubscribe(Subscription subscription) {
...
-
+        numRequestsInFlight.incrementAndGet();
         consumer.apply(item).whenComplete((r, t) -> {
             numRequestsInFlight.decrementAndGet();
-            if (!isStreamingDone) {
-                subscription.request(1);
-            } else {
-                flushBufferIfNeeded();
-            }
+            checkForCompletion();
```
millems, line 69: Potential optimization: pass in numRequestsInFlight to checkForCompletion instead of reading it in that method, to reduce the number of volatile reads.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/AsyncBufferingSubscriber.java
```diff
@@ -65,89 +57,40 @@ public void onSubscribe(Subscription subscription) {
...
+    }
+
+    private void checkForCompletion() {
+        if (!upstreamDone) {
+            subscription.request(1);
+            return;
+        }
+
+        if (upstreamDone && numRequestsInFlight.get() == 0) {
+            returnFuture.complete(null);
```
millems, line 92: Worth adding a comment to this if statement body: This could get invoked multiple times, but it doesn't matter because future.complete is idempotent. 

Worth adding the comment so that other people don't add code here assuming it only happens once.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/AsyncBufferingSubscriber.java
```diff
@@ -65,89 +57,40 @@ public void onSubscribe(Subscription subscription) {
...
-
     @Override
     public void onError(Throwable t) {
-        handleError(t);
-        storingSubscriber.onError(t);
-    }
-
-    private void handleError(Throwable t) {
+        upstreamDone = true;
         returnFuture.completeExceptionally(t);
```
millems, line 76: We should set `upstreamDone = true` _after_ completing exceptionally, to prevent accidental successful completion by a concurrent checkForCompletion.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/AsyncBufferingSubscriber.java
```diff
@@ -65,89 +57,40 @@ public void onSubscribe(Subscription subscription) {
...
-        flushBufferIfNeeded();
+        upstreamDone = true;
+        checkForCompletion();
+    }
+
+    private void checkForCompletion() {
+        if (!upstreamDone) {
+            subscription.request(1);
+            return;
+        }
```
millems, line 89: This volatile read seems unnecessary. I believe `request()` is fine after onComplete or onError is signaled.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/AsyncBufferingSubscriber.java
```diff
@@ -65,89 +57,40 @@ public void onSubscribe(Subscription subscription) {
...
-        isStreamingDone = true;
-        storingSubscriber.onComplete();
-        flushBufferIfNeeded();
+        upstreamDone = true;
+        checkForCompletion();
+    }
+
+    private void checkForCompletion() {
+        if (!upstreamDone) {
+            subscription.request(1);
```
millems, line 87: Although I *think* we have to synchronize the request(1) calls so that they don't happen synchronously:

> A Subscriber MUST ensure that all calls on its Subscription's request and cancel methods are performed [serially](https://github.com/reactive-streams/reactive-streams-jvm#term_serially).

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/AsyncBufferingSubscriber.java
zoewangg: Good catch!

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/AsyncBufferingSubscriber.java
```diff
@@ -64,32 +64,35 @@ public void onSubscribe(Subscription subscription) {
...
+            // because future.complete is idempotent.
+            returnFuture.complete(null);
             return;
         }
 
-        if (upstreamDone && numRequestsInFlight.get() == 0) {
-            returnFuture.complete(null);
+        synchronized (this) {
+            subscription.request(1);
         }
```
millems: Nit: It is weird to request(1) here instead of in onNext. It's not a bug, but it's potentially confusing why checking for completion causes more data to be requested.

millems (approved):

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/AsyncBufferingSubscriber.java
zoewangg: Yeah, it's fair, will update

Issue #272, opened 2017-11-02
S3 URI Parser


millems:
1.11.x provides http://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/s3/AmazonS3URI.html to parse the interesting components out of a S3 URI.

A similar piece of functionality should be made available in 2.0.

In the short term, customers can still use it from 1.11.x.

perihelion1:
Similarly, my project uses [`AmazonS3Client.getUrl(String bucketName, String key)`](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/s3/AmazonS3Client.html#getUrl-java.lang.String-java.lang.String-) to generate and save the URL of an object I've just created. Is there anything equivalent available in 2.0?

millems:
@perihelion1 https://github.com/aws/aws-sdk-java-v2/issues/860 tracks that feature.

ribeirux:
@millems any update on this?

Should we still use the AmazonS3URI from 1.11.x to get the bucket/key/version/region from a S3 URL?

Is there a way of downloading an object from S3 using the URl only with v2 sdk for java?

millems:
@ribeirux There's currently only a function to create a URL, given the bucket/key/etc in 2.x. 

Parsing the URI for bucket/key/etc. is surprisingly challenging, and our functionality in 1.11.x doesn't cover all scenarios. We'd like our 2.x implementation to work for all URLs, but things are in flight with S3's URI patterns right now (see: [path style deprecation](https://aws.amazon.com/blogs/aws/amazon-s3-path-deprecation-plan-the-rest-of-the-story/), among other things).

We'd like to see the dust settle on that chaos before we commit to being able to implement this functionality.

What's the reason you need parsing of the URI? That would help us when we're designing things out.

As for downloading an object using a URI in the SDK: not really. Is there a reason the JDK's URL connection (or something similar like Apache's HTTP client) isn't sufficient? We'd like to create a way to download objects using the SDK's retry policies (see: [the downloading via presigned URLs design discussion](https://github.com/aws/aws-sdk-java-v2/tree/master/docs/design/core/presigners#what-about-execution-of-presigned-requests)), but that's lower in our backlog than removing other 2.x migration blockers.

mrog:
[...]

PR #3874, opened 2023-03-30
S3 URI Parser

davidh44:
## Motivation and Context

V2 Parity - S3 URI Parser

https://github.com/aws/aws-sdk-java-v2/issues/272

## Modifications

Adds parseUri() to S3Utilities, which returns a parsed S3Uri object

## Testing

added unit tests

## Types of changes

## Checklist

## License

### DIFF
--- a/services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Utilities.java
+++ b/services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Utilities.java
@@ -20,7 +20,6 @@ import java.net.MalformedURLException;
 import java.net.URI;
 import java.net.URL;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -69,7 +68,9 @@ import software.amazon.awssdk.services.s3.model.GetObjectRequest;
 import software.amazon.awssdk.services.s3.model.GetUrlRequest;
 import software.amazon.awssdk.services.s3.parsing.S3Uri;
 import software.amazon.awssdk.utils.AttributeMap;
+import software.amazon.awssdk.utils.StringUtils;
 import software.amazon.awssdk.utils.Validate;
+import software.amazon.awssdk.utils.http.SdkHttpUtils;
 
 /**
  * Utilities for working with Amazon S3 objects. An instance of this class can be created by:
@@ -262,8 +263,8 @@ public final class S3Utilities {
      * parameters of the URI. Only path-style and virtual-hosted-style URI parsing is supported, including CLI-style
      * URIs, e.g., "s3://bucket/key". AccessPoints and Outposts URI parsing is not supported. If you work with object keys
      * and/or query parameters with special characters, they must be URL-encoded, e.g., replace " " with "%20". If you work with
-     * bucket names that contain a dot, i.e., ".", the dot must not be URL-encoded. Encoded buckets, keys, and query parameters
-     * will be returned decoded.
+     * virtual-hosted-style URIs with bucket names that contain a dot, i.e., ".", the dot must not be URL-encoded. Encoded
+     * buckets, keys, and query parameters will be returned decoded.
      *
      * <p>
      * For more information on path-style and virtual-hosted-style URIs, see <a href=
@@ -282,11 +283,11 @@ public final class S3Utilities {
      *     URI uri = URI.create(uriString);
      *     S3Uri s3Uri = s3Utilities.parseUri(uri);
      *
-     *     String bucket = s3Uri.bucket().get(); // "myBucket"
-     *     String key = s3Uri.key().get(); // "doc.txt"
-     *     Region region = s3Uri.region().get(); // Region.US_WEST_1
+     *     String bucket = s3Uri.bucket().orElse(null); // "myBucket"
+     *     String key = s3Uri.key().orElse(null); // "doc.txt"
+     *     Region region = s3Uri.region().orElse(null); // Region.US_WEST_1
      *     boolean isPathStyle = s3Uri.isPathStyle(); // false
-     *     String versionId = s3Uri.firstMatchingQueryParamValue("versionId").get(); // "abc123"
+     *     String versionId = s3Uri.firstMatchingRawQueryParameter("versionId").orElse(null); // "abc123"
      *}
      */
     public S3Uri parseUri(URI uri) {
@@ -299,13 +300,47 @@ public final class S3Utilities {
         return parseStandardUri(uri);
     }
 
+    private String[] parsePathStyleUri(URI uri) {
+        String bucket = null;
+        String key = null;
+        String path = uri.getPath();
+
+        if (!StringUtils.isEmpty(path) && !"/".equals(path)) {
+            int index = path.indexOf('/', 1);
+
+            if (index == -1) {
+                // No trailing slash, e.g., "https://s3.amazonaws.com/bucket"
+                bucket = path.substring(1);
+            } else {
+                bucket = path.substring(1, index);
+                if (index != path.length() - 1) {
+                    key = path.substring(index + 1);
+                }
+            }
+        }
+        return new String[]{key, bucket};
+    }
+
+    private String[] parseVirtualHostedStyleUri(URI uri, Matcher matcher) {
+        String bucket = null;
+        String key = null;
+        String path = uri.getPath();
+        String prefix = matcher.group(1);
+
+        bucket = prefix.substring(0, prefix.length() - 1);
+        if (!StringUtils.isEmpty(path) && !"/".equals(path)) {
+            key = path.substring(1);
+        }
+
+        return new String[]{key, bucket};
+    }
+
     private S3Uri parseStandardUri(URI uri) {
         String bucket = null;
         String key = null;
         String region = null;
         boolean isPathStyle = false;
         Map<String, List<String>> queryParams = new HashMap<>();
-        String path = uri.getPath();
 
         if (uri.getHost() == null) {
             throw new IllegalArgumentException("Invalid S3 URI: no hostname: " + uri);
@@ -316,29 +351,16 @@ public final class S3Utilities {
             throw new IllegalArgumentException("Invalid S3 URI: hostname does not appear to be a valid S3 endpoint: " + uri);
         }
 
+        String[] parsed;
         String prefix = matcher.group(1);
         if (prefix == null || prefix.isEmpty()) {
             isPathStyle = true;
-
-            if (!path.isEmpty() && !"/".equals(path)) {
-                int index = path.indexOf('/', 1);
-
-                if (index == -1) {
-                    // No trailing slash, e.g., "https://s3.amazonaws.com/bucket"
-                    bucket = path.substring(1);
-                } else {
-                    bucket = path.substring(1, index);
-                    if (index != path.length() - 1) {
-                        key = path.substring(index + 1);
-                    }
-                }
-            }
+            parsed = parsePathStyleUri(uri);
         } else {
-            bucket = prefix.substring(0, prefix.length() - 1);
-            if (path != null && !path.isEmpty() && !"/".equals(path)) {
-                key = path.substring(1);
-            }
+            parsed = parseVirtualHostedStyleUri(uri, matcher);
         }
+        key = parsed[0];
+        bucket = parsed[1];
 
         if (!"amazonaws".equals(matcher.group(2))) {
             region = matcher.group(2);
@@ -346,7 +368,7 @@ public final class S3Utilities {
 
         String queryPart = uri.getQuery();
         if (queryPart != null) {
-            parseQuery(queryParams, queryPart);
+            queryParams = SdkHttpUtils.uriParams(uri);
         }
 
         Region uriRegion = region != null ? Region.of(region) : null;
@@ -359,6 +381,7 @@ public final class S3Utilities {
                     .isPathStyle(isPathStyle)
                     .queryParams(queryParams)
                     .build();
+
     }
 
     private S3Uri parseAwsCliStyleUri(URI uri) {
@@ -375,11 +398,6 @@ public final class S3Utilities {
 
         if (path.length() > 1) {
             key = path.substring(1);
-        }
-
-        String queryPart = uri.getQuery();
-        if (queryPart != null) {
-            parseQuery(queryParams, queryPart);
         }
 
         return S3Uri.builder()
@@ -401,21 +419,6 @@ public final class S3Utilities {
 
         if (uri.toString().contains(".s3-outposts")) {
             throw new IllegalArgumentException("Outposts URI parsing is not supported: " + uri);
-        }
-    }
-
-    private void parseQuery(Map<String, List<String>> queryParams, String queryPart) {
-        String[] params = queryPart.split("&");
-        for (String param: params) {
-            String[] keyValuePair = param.split("=", 2);
-            String key = keyValuePair[0];
-            if (key.isEmpty()) {
-                continue;
-            }
-            List<String> paramValues = queryParams.containsKey(key) ? queryParams.get(key) : new ArrayList<>();
-            String[] valuesPart = keyValuePair[1].split(",");
-            Collections.addAll(paramValues, valuesPart);
-            queryParams.put(key, paramValues);
         }
     }
 
--- a/services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Utilities.java
+++ b/services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Utilities.java
@@ -20,6 +20,7 @@ import java.net.MalformedURLException;
 import java.net.URI;
 import java.net.URL;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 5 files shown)


services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Utilities.java
```diff
@@ -251,6 +258,103 @@ public URL getUrl(GetUrlRequest getUrlRequest) {
         }
     }
 
+    /**
+     * Returns a parsed {@link S3Uri} with which a user can easily retrieve the the bucket, key, region, style, and query
```
dave-fn, line 262: nit: repeat the

dave-fn (approved):

services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Utilities.java
```diff
@@ -251,6 +258,103 @@ public URL getUrl(GetUrlRequest getUrlRequest) {
...
+            }
+
+            if (!"amazonaws".equals(matcher.group(2))) {
+                region = matcher.group(2);
+            }
+        }
+
+        String queryPart = uri.getQuery();
+        if (queryPart != null) {
+            parseQuery(queryParams, queryPart);
```
zoewangg, line 343: We can probably use `SdkHttpUtils.uriParams` to parse query params

services/s3/src/main/java/software/amazon/awssdk/services/s3/parsing/S3Uri.java
```diff
@@ -0,0 +1,241 @@
...
+     * Returns true if the URI is path-style, false if the URI is virtual-hosted style.
+     */
+    public boolean isPathStyle() {
+        return isPathStyle;
+    }
+
+    /**
+     * Returns a map of the query parameters specified in the URI. Returns an empty map if no queries are specified.
+     */
+    public Map<String, List<String>> queryParams() {
```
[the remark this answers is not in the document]
zoewangg: `rawQueryParameters` is not encoded actually https://github.com/aws/aws-sdk-java-v2/blob/master/http-client-spi/src/main/java/software/amazon/awssdk/http/SdkHttpRequest.java#L92

services/s3/src/main/java/software/amazon/awssdk/services/s3/parsing/S3Uri.java
```diff
@@ -0,0 +1,241 @@
...
+        return Collections.unmodifiableMap(queryParams);
+    }
+
+    /**
+     * Returns the list of values for a specified query parameter. A {@link SdkClientException} is thrown if the URI does not
+     * contain the specified query parameter.
+     */
+    public List<String> queryParamValues(String key) {
```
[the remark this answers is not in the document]
zoewangg: `firstMatchingRawQueryParameters` is what we used in `SdkHttpRequest`. Admittedly, `firstMatching` is a bit confusing, but I'm guessing it is to disambiguate it with `rawQueryParameters`. I'd say consistency should be preferred. 

https://github.com/aws/aws-sdk-java-v2/blob/master/http-client-spi/src/main/java/software/amazon/awssdk/http/SdkHttpRequest.java#L118

services/s3/src/main/java/software/amazon/awssdk/services/s3/parsing/S3Uri.java
```diff
@@ -0,0 +1,253 @@
...
+        List<String> queryValuesCopy = Arrays.asList(new String[queryValues.size()]);
+        Collections.copy(queryValuesCopy, queryValues);
+        return queryValuesCopy;
+    }
+
+    /**
+     * Returns the value for the specified query parameter. If there are multiple values for the query parameter, the first
+     * value is returned. An empty optional is returned if the URI does not contain the specified query parameter.
+     */
+    public Optional<String> firstMatchingQueryParamValue(String key) {
```
zoewangg, line 131: `firstMatchingRawQueryParameter`?

services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Utilities.java
zoewangg: Hmm, doesn't `SdkHttpUtils.uriParams` decode them afterwards?

https://github.com/aws/aws-sdk-java-v2/blob/b11e9b43f2b1473248831a49786623fd13cd8a2d/utils/src/main/java/software/amazon/awssdk/utils/http/SdkHttpUtils.java#L403-L408

services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Utilities.java
```diff
@@ -251,6 +257,168 @@ public URL getUrl(GetUrlRequest getUrlRequest) {
...
+            throw new IllegalArgumentException("Invalid S3 URI: no hostname: " + uri);
+        }
+
+        Matcher matcher = ENDPOINT_PATTERN.matcher(uri.getHost());
+        if (!matcher.find()) {
+            throw new IllegalArgumentException("Invalid S3 URI: hostname does not appear to be a valid S3 endpoint: " + uri);
+        }
+
+        String prefix = matcher.group(1);
+        if (prefix == null || prefix.isEmpty()) {
```
zoewangg, line 320: Nested if-else is a bit hard to read and could be error-prone, can we break it up a bit?
Something like below:
```
private S3Uri parseStandardUri(URI uri) {
   validate(...);

   if (isPathStyle(...)) {
    parsePathStyle(...);
   } else {
     parseVirtualStyle(...);
  }

  parseQueryIfNeeded(...);

} 



```

services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Utilities.java
```diff
@@ -251,6 +257,168 @@ public URL getUrl(GetUrlRequest getUrlRequest) {
...
+     * <p><b>Example Usage</b>
+     * <p>
+     * {@snippet :
+     *     S3Client s3Client = S3Client.create();
+     *     S3Utilities s3Utilities = s3Client.utilities();
+     *     String uriString = "https://myBucket.s3.us-west-1.amazonaws.com/doc.txt?versionId=abc123";
+     *     URI uri = URI.create(uriString);
+     *     S3Uri s3Uri = s3Utilities.parseUri(uri);
+     *
+     *     String bucket = s3Uri.bucket().get(); // "myBucket"
```
zoewangg, line 285: We probably shouldn't just do .get on an Optional since it's a bad practice.

Maybe `s3Uri.bucket.ifPresent(bucket -> System.out.println("bucket: " + bucket)); // "myBucket"`? Same for other properties below

services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Utilities.java
```diff
@@ -251,6 +257,168 @@ public URL getUrl(GetUrlRequest getUrlRequest) {
...
+     *     String key = s3Uri.key().get(); // "doc.txt"
+     *     Region region = s3Uri.region().get(); // Region.US_WEST_1
+     *     boolean isPathStyle = s3Uri.isPathStyle(); // false
+     *     String versionId = s3Uri.firstMatchingQueryParamValue("versionId").get(); // "abc123"
+     *}
+     */
+    public S3Uri parseUri(URI uri) {
```
zoewangg, line 292: This is not a blocker, but we may want to support string, which we will encode for them in the future depending on customer ask.

services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Utilities.java
```diff
@@ -251,6 +257,168 @@ public URL getUrl(GetUrlRequest getUrlRequest) {
...
+            throw new IllegalArgumentException("Invalid S3 URI: bucket not included: " + uri);
+        }
+
+        if (path.length() > 1) {
+            key = path.substring(1);
+        }
+
+        String queryPart = uri.getQuery();
+        if (queryPart != null) {
+            parseQuery(queryParams, queryPart);
```
zoewangg, line 382: I'm not sure if S3 URI supports query params. https://docs.aws.amazon.com/AmazonS3/latest/userguide/access-bucket-intro.html#:~:text=access%20by%20HTTPS.-,Accessing%20a%20bucket%20using%20S3%3A//,-Some%20AWS%20services

services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Utilities.java
```diff
@@ -251,6 +257,168 @@ public URL getUrl(GetUrlRequest getUrlRequest) {
...
+                    bucket = path.substring(1);
+                } else {
+                    bucket = path.substring(1, index);
+                    if (index != path.length() - 1) {
+                        key = path.substring(index + 1);
+                    }
+                }
+            }
+        } else {
+            bucket = prefix.substring(0, prefix.length() - 1);
```
zoewangg, line 337: Isn't this essentially prefix?

services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Utilities.java
```diff
@@ -251,6 +257,168 @@ public URL getUrl(GetUrlRequest getUrlRequest) {
...
+            if (path != null && !path.isEmpty() && !"/".equals(path)) {
```
zoewangg, line 338: nit, `path != null && !path.isEmpty()` can be replaced with `!StringUtils.isEmpty(path)`. Could you search other places where it can be replaced? I think I saw it somewhere else as well.

services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Utilities.java
zoewangg: nvm, end index is exclusive 😅

services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Utilities.java
davidh44: yep, replaced here in virtual style parsing and above in path style parsing

services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Utilities.java
davidh44: Yeah, I got confused since it was failing the tests that had the "=" encoded, but I removed the encoding there.

Also, uriParams() doesn't support multiple values for a query that are separated by a String, e.g., versionId=123,456,789, but only ampersand separators, .e.g, versionId=123&versionId=456&versionId=789. Do we want to support that as well? I changed to uriParams() and removed those tests for now

services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Utilities.java
```diff
@@ -251,6 +258,170 @@ public URL getUrl(GetUrlRequest getUrlRequest) {
...
+                    .bucket(bucket)
+                    .key(key)
+                    .region(uriRegion)
+                    .isPathStyle(isPathStyle)
+                    .queryParams(queryParams)
+                    .build();
+
+    }
+
+    private String[] parsePathStyleUri(URI uri) {
```
zoewangg, line 352: I'd use `Pair<String, String>` instead of `String[]`

services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Utilities.java
zoewangg: I see. There's no standard behavior for multi-value query params. I think we should just support what we currently support, i.e.,`&`.

services/s3/src/main/java/software/amazon/awssdk/services/s3/parsing/S3Uri.java
```diff
@@ -0,0 +1,253 @@
...
+    private final boolean isPathStyle;
+    private final Map<String, List<String>> queryParams;
+
+    private S3Uri(Builder builder) {
+        this.uri = Validate.notNull(builder.uri, "URI must not be null");
+        this.bucket = builder.bucket;
+        this.key = builder.key;
+        this.region = builder.region;
+        this.isPathStyle = Validate.notNull(builder.isPathStyle, "Path style flag must not be null");
+        this.queryParams = Validate.notNull(builder.queryParams, "Query parameters map must not be null");
```
zoewangg, line 59: It may be a better experience if we don't require null here and make it default empty map.
this.queryParams = queryParams == null ? new HashMap<>() : queryParams;

services/s3/src/main/java/software/amazon/awssdk/services/s3/parsing/S3Uri.java
```diff
@@ -0,0 +1,253 @@
...
+     */
+    public boolean isPathStyle() {
+        return isPathStyle;
+    }
+
+    /**
+     * Returns a map of the query parameters specified in the URI. Returns an empty map if no queries are specified.
+     */
+    public Map<String, List<String>> rawQueryParameters() {
+        return Collections.unmodifiableMap(queryParams);
```
zoewangg, line 110: We probably should do a deep copy. https://github.com/aws/aws-sdk-java-v2/blob/master/utils/src/main/java/software/amazon/awssdk/utils/CollectionUtils.java#L83

For performance optimization, can we deep copy it in the ctor instead so that we just do it once?

services/s3/src/main/java/software/amazon/awssdk/services/s3/parsing/S3Uri.java
```diff
@@ -0,0 +1,253 @@
...
+        Collections.copy(queryValuesCopy, queryValues);
+        return queryValuesCopy;
+    }
+
+    /**
+     * Returns the value for the specified query parameter. If there are multiple values for the query parameter, the first
+     * value is returned. An empty optional is returned if the URI does not contain the specified query parameter.
+     */
+    public Optional<String> firstMatchingRawQueryParameter(String key) {
+        return queryParams.get(key) == null ? Optional.empty() : Optional.of(queryParams.get(key).get(0));
```
zoewangg, line 132: nit: `return Optional.ofNullable(queryParams.get(key)).map(q -> q.get(0));`

services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Utilities.java
davidh44: Pair throws an exception when either the key is bucket is null

services/s3/src/main/java/software/amazon/awssdk/services/s3/parsing/S3Uri.java
davidh44: +1

services/s3/src/main/java/software/amazon/awssdk/services/s3/parsing/S3Uri.java
davidh44: +1

services/s3/src/main/java/software/amazon/awssdk/services/s3/parsing/S3Uri.java
davidh44: +1

services/s3/src/main/java/software/amazon/awssdk/services/s3/parsing/S3Uri.java
```diff
@@ -0,0 +1,262 @@
...
+         */
+        public Builder queryParams(Map<String, List<String>> queryParams) {
+            this.queryParams = queryParams;
+            return this;
+        }
+
+        /**
+         * Return the configured URI
+         */
+        public URI uri() {
```
zoewangg, line 252: Seems a bit odd that we just expose getter for one of the property. Can we just pass uri to `addQueryParamsIfNeeded(S3Uri.Builder builder, URI uri)` and `parsePathStyleUri` and `parseVirtualHostedStyleUri`?

davidh44: yup makes sense, updated

services/s3/src/main/java/software/amazon/awssdk/services/s3/parsing/S3Uri.java
```diff
@@ -0,0 +1,255 @@
...
+     */
+    public boolean isPathStyle() {
+        return isPathStyle;
+    }
+
+    /**
+     * Returns a map of the query parameters specified in the URI. Returns an empty map if no queries are specified.
+     */
+    public Map<String, List<String>> rawQueryParameters() {
+        return Collections.unmodifiableMap(queryParams);
```
zoewangg, line 112: Nit: now that we've done deep copy in the ctor, we could just return queryParams as-is I think.

services/s3/src/main/java/software/amazon/awssdk/services/s3/parsing/S3Uri.java
```diff
@@ -0,0 +1,255 @@
...
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.services.s3.parsing;
```
zoewangg, line 16: It's not exactly just related to parsing. `model` subpackage is also not good because people may confuse it with service model POJOs. Maybe just under `software.amazon.awssdk.services.s3`? (sorry, just noticed it now)

zoewangg (approved):
Good job!

PR #3883, opened 2023-04-04
Update AwsCredentialsProviderChain for new Identity type

gosar:
## Motivation and Context

As part of SRA changes, update AwsCredentialsProviderChain for new Identity type.

## Modifications

Added overloaded methods that accept the new Identity type. For Collection input, cannot use the same method name because it results in compile error: `both methods have same erasure`, so created method with different name.

Confirmed that adding overloaded varargs for the 2 types is fine and not ambiguous when when called with zero args, because of https://docs.oracle.com/javase/specs/jls/se8/html/jls-15.html#jls-15.12.2.5 - it chooses the more specific method (old type, which is sub-type of the new type). In any case these methods with varargs when called with zero args, throws RuntimeException because we require at least 1 provider in the chain.

## Testing

Added unit tests and `./mvnw clean install -pl :auth`

## Types of changes

## Checklist

## License

### DIFF
--- a/core/auth/src/main/java/software/amazon/awssdk/auth/credentials/AwsCredentialsProviderChain.java
+++ b/core/auth/src/main/java/software/amazon/awssdk/auth/credentials/AwsCredentialsProviderChain.java
@@ -22,6 +22,8 @@ import java.util.Collections;
 import java.util.List;
 import software.amazon.awssdk.annotations.SdkPublicApi;
 import software.amazon.awssdk.core.exception.SdkClientException;
+import software.amazon.awssdk.identity.spi.AwsCredentialsIdentity;
+import software.amazon.awssdk.identity.spi.IdentityProvider;
 import software.amazon.awssdk.utils.IoUtils;
 import software.amazon.awssdk.utils.Logger;
 import software.amazon.awssdk.utils.SdkAutoCloseable;
@@ -52,11 +54,11 @@ public final class AwsCredentialsProviderChain
                ToCopyableBuilder<AwsCredentialsProviderChain.Builder, AwsCredentialsProviderChain> {
     private static final Logger log = Logger.loggerFor(AwsCredentialsProviderChain.class);
 
-    private final List<AwsCredentialsProvider> credentialsProviders;
+    private final List<IdentityProvider<? extends AwsCredentialsIdentity>> credentialsProviders;
 
     private final boolean reuseLastProviderEnabled;
 
-    private volatile AwsCredentialsProvider lastUsedProvider;
+    private volatile IdentityProvider<? extends AwsCredentialsIdentity> lastUsedProvider;
 
     /**
      * @see #builder()
@@ -84,21 +86,33 @@ public final class AwsCredentialsProviderChain
         return builder().credentialsProviders(awsCredentialsProviders).build();
     }
 
+    /**
+     * Create an AWS credentials provider chain with default configuration that checks the given credential providers.
+     * @param awsCredentialsProviders The credentials providers that should be checked for credentials, in the order they should
+     *                                be checked.
+     * @return A credential provider chain that checks the provided credential providers in order.
+     */
+    public static AwsCredentialsProviderChain of(IdentityProvider<? extends AwsCredentialsIdentity>... awsCredentialsProviders) {
+        return builder().credentialsProviders(awsCredentialsProviders).build();
+    }
+
     @Override
     public AwsCredentials resolveCredentials() {
         if (reuseLastProviderEnabled && lastUsedProvider != null) {
-            return lastUsedProvider.resolveCredentials();
+            // TODO: Exception handling for join?
+            return CredentialUtils.toCredentials(lastUsedProvider.resolveIdentity().join());
         }
 
         List<String> exceptionMessages = null;
-        for (AwsCredentialsProvider provider : credentialsProviders) {
+        for (IdentityProvider<? extends AwsCredentialsIdentity> provider : credentialsProviders) {
             try {
-                AwsCredentials credentials = provider.resolveCredentials();
+                // TODO: Exception handling for join?
+                AwsCredentialsIdentity credentials = provider.resolveIdentity().join();
 
                 log.debug(() -> "Loading credentials from " + provider);
 
                 lastUsedProvider = provider;
-                return credentials;
+                return CredentialUtils.toCredentials(credentials);
             } catch (RuntimeException e) {
                 // Ignore any exceptions and move onto the next provider
                 String message = provider + ": " + e.getMessage();
@@ -156,19 +170,43 @@ public final class AwsCredentialsProviderChain
         /**
          * Configure the credentials providers that should be checked for credentials, in the order they should be checked.
          */
-        Builder credentialsProviders(AwsCredentialsProvider... credentialsProviders);
+        Builder credentialsIdentityProviders(
+            Collection<? extends IdentityProvider<? extends AwsCredentialsIdentity>> credentialsProviders);
+
+        /**
+         * Configure the credentials providers that should be checked for credentials, in the order they should be checked.
+         */
+        default Builder credentialsProviders(AwsCredentialsProvider... credentialsProviders) {
+            return credentialsProviders((IdentityProvider<? extends AwsCredentialsIdentity>[]) credentialsProviders);
+        }
+
+        /**
+         * Configure the credentials providers that should be checked for credentials, in the order they should be checked.
+         */
+        default Builder credentialsProviders(IdentityProvider<? extends AwsCredentialsIdentity>... credentialsProviders) {
+            throw new UnsupportedOperationException();
+        }
 
         /**
          * Add a credential provider to the chain, after the credential providers that have already been configured.
          */
-        Builder addCredentialsProvider(AwsCredentialsProvider credentialsProviders);
+        default Builder addCredentialsProvider(AwsCredentialsProvider credentialsProvider) {
+            return addCredentialsProvider((IdentityProvider<? extends AwsCredentialsIdentity>) credentialsProvider);
+        }
+
+        /**
+         * Add a credential provider to the chain, after the credential providers that have already been configured.
+         */
+        default Builder addCredentialsProvider(IdentityProvider<? extends AwsCredentialsIdentity> credentialsProvider) {
+            throw new UnsupportedOperationException();
+        }
 
         AwsCredentialsProviderChain build();
     }
 
     private static final class BuilderImpl implements Builder {
         private Boolean reuseLastProviderEnabled = true;
-        private List<AwsCredentialsProvider> credentialsProviders = new ArrayList<>();
+        private List<IdentityProvider<? extends AwsCredentialsIdentity>> credentialsProviders = new ArrayList<>();
 
         private BuilderImpl() {
         }
@@ -199,13 +237,25 @@ public final class AwsCredentialsProviderChain
         }
 
         @Override
-        public Builder credentialsProviders(AwsCredentialsProvider... credentialsProviders) {
-            return credentialsProviders(Arrays.asList(credentialsProviders));
+        public Builder credentialsIdentityProviders(
+                Collection<? extends IdentityProvider<? extends AwsCredentialsIdentity>> credentialsProviders) {
+            this.credentialsProviders = new ArrayList<>(credentialsProviders);
+            return this;
+        }
+
+        public void setCredentialsIdentityProviders(
+                Collection<? extends IdentityProvider<? extends AwsCredentialsIdentity>> credentialsProviders) {
+            credentialsIdentityProviders(credentialsProviders);
         }
 
         @Override
-        public Builder addCredentialsProvider(AwsCredentialsProvider credentialsProviders) {
-            this.credentialsProviders.add(credentialsProviders);
+        public Builder credentialsProviders(IdentityProvider<? extends AwsCredentialsIdentity>... credentialsProviders) {
+            return credentialsIdentityProviders(Arrays.asList(credentialsProviders));
+        }
+
+        @Override
+        public Builder addCredentialsProvider(IdentityProvider<? extends AwsCredentialsIdentity> credentialsProvider) {
+            this.credentialsProviders.add(credentialsProvider);
             return this;
         }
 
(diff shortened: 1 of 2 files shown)


core/auth/src/main/java/software/amazon/awssdk/auth/credentials/AwsCredentialsProviderChain.java
```diff
@@ -84,21 +86,33 @@ public static AwsCredentialsProviderChain of(AwsCredentialsProvider... awsCreden
         return builder().credentialsProviders(awsCredentialsProviders).build();
     }
 
+    /**
+     * Create an AWS credentials provider chain with default configuration that checks the given credential providers.
+     * @param awsCredentialsProviders The credentials providers that should be checked for credentials, in the order they should
```
joviegas, line 91: Does this mean ==> The credential providers to be checked in the specified order.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/AwsCredentialsProviderChain.java
```diff
@@ -156,19 +170,43 @@ public interface Builder extends CopyableBuilder<Builder, AwsCredentialsProvider
...
+            throw new UnsupportedOperationException();
+        }
 
         AwsCredentialsProviderChain build();
     }
 
     private static final class BuilderImpl implements Builder {
         private Boolean reuseLastProviderEnabled = true;
-        private List<AwsCredentialsProvider> credentialsProviders = new ArrayList<>();
+        private List<IdentityProvider<? extends AwsCredentialsIdentity>> credentialsProviders = new ArrayList<>();
```
joviegas, line 209: Should we rename credentialsProviders variable to identityProviders ?

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/AwsCredentialsProviderChain.java
```diff
@@ -84,21 +86,33 @@ public static AwsCredentialsProviderChain of(AwsCredentialsProvider... awsCreden
...
+     */
+    public static AwsCredentialsProviderChain of(IdentityProvider<? extends AwsCredentialsIdentity>... awsCredentialsProviders) {
+        return builder().credentialsProviders(awsCredentialsProviders).build();
+    }
+
     @Override
     public AwsCredentials resolveCredentials() {
         if (reuseLastProviderEnabled && lastUsedProvider != null) {
-            return lastUsedProvider.resolveCredentials();
+            // TODO: Exception handling for join?
```
joviegas, line 102: What are the plans to handle this TODO? Will we have a separate PR? Also, I'm just wondering why we can't have a separate synchronous/blocking resolveIdentity API if we want to resolve the identity immediately. That API could internally handle the interruptions for join, instead of caller doing explicit joins and handling exceptions.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/AwsCredentialsProviderChain.java
```diff
@@ -84,21 +86,33 @@ public static AwsCredentialsProviderChain of(AwsCredentialsProvider... awsCreden
...
+            return CredentialUtils.toCredentials(lastUsedProvider.resolveIdentity().join());
         }
 
         List<String> exceptionMessages = null;
-        for (AwsCredentialsProvider provider : credentialsProviders) {
+        for (IdentityProvider<? extends AwsCredentialsIdentity> provider : credentialsProviders) {
             try {
-                AwsCredentials credentials = provider.resolveCredentials();
+                // TODO: Exception handling for join?
+                AwsCredentialsIdentity credentials = provider.resolveIdentity().join();
```
joviegas, line 110: Should we change the variable name `credentials` to reflect something related identity like identityCredentials ?

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/AwsCredentialsProviderChain.java
```diff
@@ -156,19 +170,43 @@ public interface Builder extends CopyableBuilder<Builder, AwsCredentialsProvider
         /**
          * Configure the credentials providers that should be checked for credentials, in the order they should be checked.
          */
-        Builder credentialsProviders(AwsCredentialsProvider... credentialsProviders);
+        Builder credentialsIdentityProviders(
+            Collection<? extends IdentityProvider<? extends AwsCredentialsIdentity>> credentialsProviders);
```
joviegas, line 174: If the order is important shouldn't this be only List , I guess `interface Set<E> extends Collection<E>` and this doesnot mantain the order , please correct me if I am wrong

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/AwsCredentialsProviderChain.java
```diff
@@ -156,19 +170,43 @@ public interface Builder extends CopyableBuilder<Builder, AwsCredentialsProvider
...
+         * Configure the credentials providers that should be checked for credentials, in the order they should be checked.
+         */
+        default Builder credentialsProviders(AwsCredentialsProvider... credentialsProviders) {
+            return credentialsProviders((IdentityProvider<? extends AwsCredentialsIdentity>[]) credentialsProviders);
+        }
+
+        /**
+         * Configure the credentials providers that should be checked for credentials, in the order they should be checked.
+         */
+        default Builder credentialsProviders(IdentityProvider<? extends AwsCredentialsIdentity>... credentialsProviders) {
```
joviegas, line 186: Should this be named as `credentialsIdentityProviders` ?

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/AwsCredentialsProviderChain.java
```diff
@@ -156,19 +170,43 @@ public interface Builder extends CopyableBuilder<Builder, AwsCredentialsProvider
...
+         */
+        default Builder addCredentialsProvider(AwsCredentialsProvider credentialsProvider) {
+            return addCredentialsProvider((IdentityProvider<? extends AwsCredentialsIdentity>) credentialsProvider);
+        }
 
         /**
          * Add a credential provider to the chain, after the credential providers that have already been configured.
          */
-        Builder addCredentialsProvider(AwsCredentialsProvider credentialsProviders);
+        default Builder addCredentialsProvider(IdentityProvider<? extends AwsCredentialsIdentity> credentialsProvider) {
```
joviegas, line 200: should this be addCredentialsIdentityProvider ?

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/AwsCredentialsProviderChain.java
```diff
@@ -199,13 +237,25 @@ public void setCredentialsProviders(Collection<? extends AwsCredentialsProvider>
         }
 
         @Override
-        public Builder credentialsProviders(AwsCredentialsProvider... credentialsProviders) {
-            return credentialsProviders(Arrays.asList(credentialsProviders));
+        public Builder credentialsIdentityProviders(
+                Collection<? extends IdentityProvider<? extends AwsCredentialsIdentity>> credentialsProviders) {
```
joviegas, line 241: General comment on use of Collection , if user is doing Set.of("1","2", "3") the API will accept it but the order will not be maintained.
I think it should be `List` if order is important,

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/AwsCredentialsProviderChain.java
```diff
@@ -199,13 +237,25 @@ public void setCredentialsProviders(Collection<? extends AwsCredentialsProvider>
...
+            return this;
+        }
+
+        public void setCredentialsIdentityProviders(
+                Collection<? extends IdentityProvider<? extends AwsCredentialsIdentity>> credentialsProviders) {
+            credentialsIdentityProviders(credentialsProviders);
+        }
+
+        @Override
+        public Builder credentialsProviders(IdentityProvider<? extends AwsCredentialsIdentity>... credentialsProviders) {
```
joviegas, line 252: credentialsIdentityProviders ? This API input is `IdentityProvider<? extends AwsCredentialsIdentity>... credentialsProviders`
Sorry it was not clear to me how we have named the API as credentialsProviders or credentialsIdentityProviders .

core/auth/src/test/java/software/amazon/awssdk/auth/credentials/AwsCredentialsProviderChainTest.java
```diff
@@ -114,13 +114,57 @@ public void testGetCredentialsException() {
...
-        thrown.expectMessage(provider2.exceptionMessage);
+        SdkClientException e = assertThrows(SdkClientException.class, () -> chain.resolveCredentials());
+        assertThat(e.getMessage()).contains(provider1.exceptionMessage);
+        assertThat(e.getMessage()).contains(provider2.exceptionMessage);
+    }
 
-        chain.resolveCredentials();
+    @Test
+    public void testEmptyChain() {
+        assertThrows(IllegalArgumentException.class, () -> AwsCredentialsProviderChain.of());
```
joviegas, line 124: Can we please Assert the message too ?

core/auth/src/test/java/software/amazon/awssdk/auth/credentials/AwsCredentialsProviderChainTest.java
```diff
@@ -114,13 +114,57 @@ public void testGetCredentialsException() {
...
-        thrown.expectMessage(provider1.exceptionMessage);
-        thrown.expectMessage(provider2.exceptionMessage);
+        SdkClientException e = assertThrows(SdkClientException.class, () -> chain.resolveCredentials());
+        assertThat(e.getMessage()).contains(provider1.exceptionMessage);
+        assertThat(e.getMessage()).contains(provider2.exceptionMessage);
+    }
 
-        chain.resolveCredentials();
+    @Test
+    public void testEmptyChain() {
```
joviegas, line 123: Optional Suggestion
can we make it clear in the test API name like
`emptyProviderChain_ThorwsException`
I feel the test prefix to test case API name is redundant.

core/auth/src/test/java/software/amazon/awssdk/auth/credentials/AwsCredentialsProviderChainTest.java
gosar: done.

core/auth/src/test/java/software/amazon/awssdk/auth/credentials/AwsCredentialsProviderChainTest.java
gosar: I kept the convention of the file so far, to avoid making the change bigger. I'll go ahead and follow the [convention](https://github.com/aws/aws-sdk-java-v2/blob/master/docs/design/NamingConventions.md) across the file.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/AwsCredentialsProviderChain.java
gosar: We are overloading the same method name where possible. This is what we are doing in all other places where new Identity type is accepted (see https://github.com/aws/aws-sdk-java-v2/pull/3829). Here, only in the Collection input case where overloading was not possible, we used `credentialsIdentityProvider()`. Using separate names in all places, could lead to confusion that they are 2 separate _properties_ of the chain. This is what we decided on in surface area review.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/AwsCredentialsProviderChain.java
gosar: Hmm, I kinda see the point. I don't know the history of why List wasn't used. Maybe it was to give more flexibility to the caller. We wouldn't change the signature now anyways. This change is just adding overloaded usage for new type and we wouldn't want it to use List where the other one allows Collection.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/AwsCredentialsProviderChain.java
gosar: No, see other comment about method naming.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/AwsCredentialsProviderChain.java
gosar: No, see other comment about method naming.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/AwsCredentialsProviderChain.java
gosar: See other comment below.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/AwsCredentialsProviderChain.java
gosar: Yes, handling joins and exception handling has been punted for future PR to address it holistically.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/AwsCredentialsProviderChain.java
gosar: These are still aws credentials providers. Just identityProviders could also confused if other identity types are allowed. To remove ambiguity we could use `awsCredentialsIdentityProviders` but I think the names doesn't matter much for private members. If both types were present in the same context, one might want to distinguish, but not needed here. Shorter is more readable. Also, the public methods use `credentialsProvider` still, so using the same name for member is fine.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/AwsCredentialsProviderChain.java
gosar: yes, this is same as existing behavior with old types.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/AwsCredentialsProviderChain.java
gosar: See other similar comment.

joviegas (approved):

PR #3895, opened 2023-04-08
Accept and use the new TokenIdentity interfaces

gosar:
## Motivation and Context

As part of SRA, earlier PR added the new interfaces. This PR updates existing code to accept and use the new Token specific interfaces.

## Modifications

This PR updates existing interfaces where customers used to be able to provide SdkTokenProvider, to be able to provide the new interface - IdentityProvider<? extends TokenIdentity> - in service client builders (via codegen), SdkTokenProviderChain.

## Testing

Added tests and `./mvnw clean install -pl :auth,:aws-core,:codegen,:codegen-generated-classes-test `

## Types of changes

## Checklist

## License


core/auth/src/main/java/software/amazon/awssdk/auth/token/credentials/TokenUtils.java
```diff
@@ -0,0 +1,50 @@
...
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.auth.token.credentials;
+
+import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.identity.spi.TokenIdentity;
+
+@SdkInternalApi
```
gosar, line 21: CredentialsUtils is `@SdkProtectedApi` since it is used in other modules, but this class isn't, so keeping this `@SdkInternalApi`. Once more SRA work is done, these conversion may become unnecessary and I'd like to delete this then.

core/auth/src/main/java/software/amazon/awssdk/auth/token/credentials/SdkTokenProviderChain.java
```diff
@@ -79,21 +81,33 @@ public static SdkTokenProviderChain of(SdkTokenProvider... sdkTokenProviders) {
...
+     */
+    public static SdkTokenProviderChain of(IdentityProvider<? extends TokenIdentity>... sdkTokenProviders) {
+        return builder().tokenProviders(sdkTokenProviders).build();
+    }
+
     @Override
     public SdkToken resolveToken() {
         if (reuseLastProviderEnabled && lastUsedProvider != null) {
-            return lastUsedProvider.resolveToken();
+            // TODO: Exception handling for join?
```
cenedhryn, line 97: Do you have a special format for TODOs? Can't remember how we're tracking them.

core/auth/src/main/java/software/amazon/awssdk/auth/token/credentials/TokenUtils.java
cenedhryn: Can we have a TODO on the class to not forget the intent?
Since the class is internal can you move it to the `internal` package?

core/auth/src/test/java/software/amazon/awssdk/auth/token/credentials/SdkTokenProviderChainTest.java
```diff
@@ -37,7 +38,7 @@ public class SdkTokenProviderChainTest {
      * for any additional calls to getToken.
      */
     @Test
-    public void testReusingLastProvider() {
+    public void resolveToken_reuseEnabled_reusesLastProvider() {
```
cenedhryn, line 41: Thanks for renaming these tests using meaningful names

core/identity-spi/src/main/java/software/amazon/awssdk/identity/spi/TokenIdentity.java
```diff
@@ -33,4 +36,44 @@ public interface TokenIdentity extends Identity {
...
+
+    /**
+     * Constructs a new token object, which can be used to authorize requests to services that use token based auth
+     *
+     * @param token The token used to authorize requests.
+     */
+    static TokenIdentity create(String token) {
+        Validate.paramNotNull(token, "token");
+
+        return new TokenIdentity() {
```
cenedhryn, line 48: Can you refresh my mind why we're doing the identity as an inline definition? Is it because we already have a token elsewhere?

core/auth/src/main/java/software/amazon/awssdk/auth/token/credentials/SdkTokenProviderChain.java
gosar: A special format we was for things that were longer term TODOs like for when we do smithy codegen. This one I'm actively working on now, so no other special format.

core/identity-spi/src/main/java/software/amazon/awssdk/identity/spi/TokenIdentity.java
gosar: Yeah, the token implementations we have are in other `modules` (sso/ssooidc). For AwsCredentials, there are implementations in the `auth` module. Used anonymous inner classes here to avoid creating a named classes like `AwsCredentialsIdentityImpl` or clashing names with `AwsBasicCredentials` and followed the same approach here. I went off a comment from Matt during the surface area review, but I might have misinterpreted it. LMK if you think this should be organized differently.

core/auth/src/main/java/software/amazon/awssdk/auth/token/credentials/SdkTokenProviderChain.java
cenedhryn: I guess we'll find any remaining ones, should they exist, when we diff before merging to master

core/identity-spi/src/main/java/software/amazon/awssdk/identity/spi/TokenIdentity.java
cenedhryn: I'd rather not create more classes right now, this seems easier to change if needed so I think we should keep it.

cenedhryn (approved):

PR #3903, opened 2023-04-11
Add TrustAllCertificates in CRT S3 Client options

joviegas:
## Motivation and Context
- S3 Crtclient did not have a way to disable TrustCertificates in its in-built http client

## Modifications

Based on https://github.com/aws/aws-sdk-java-v2/blob/master/http-clients/aws-crt-client/src/main/java/software/amazon/awssdk/http/crt/AwsCrtAsyncHttpClient.java#L103 added a option in S3CrtHttpConfiguration class to disable TrustAllCertificates options
- Reference [tls_context_options](https://awslabs.github.io/aws-crt-cpp/class_aws_1_1_crt_1_1_io_1_1_tls_context_options.html#a94e652e626b9b6f5b6fcde274244fcb7)

## Types of changes

## License

### DIFF
--- a/services/s3/src/main/java/software/amazon/awssdk/services/s3/crt/S3CrtHttpConfiguration.java
+++ b/services/s3/src/main/java/software/amazon/awssdk/services/s3/crt/S3CrtHttpConfiguration.java
@@ -38,13 +38,13 @@ public final class S3CrtHttpConfiguration implements ToCopyableBuilder<S3CrtHttp
     private final Duration connectionTimeout;
     private final S3CrtProxyConfiguration proxyConfiguration;
     private final S3CrtConnectionHealthConfiguration healthConfiguration;
-    private final Boolean shouldTrustAllCertificates;
+    private final Boolean trustAllCertificatesEnabled;
 
     private S3CrtHttpConfiguration(DefaultBuilder builder) {
         this.connectionTimeout = builder.connectionTimeout;
         this.proxyConfiguration = builder.proxyConfiguration;
         this.healthConfiguration = builder.healthConfiguration;
-        this.shouldTrustAllCertificates = builder.shouldTrustAllCertificates;
+        this.trustAllCertificatesEnabled = builder.trustAllCertificatesEnabled;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


services/s3/src/main/java/software/amazon/awssdk/services/s3/crt/S3CrtHttpConfiguration.java
```diff
@@ -115,6 +128,19 @@ public interface Builder extends CopyableBuilder<S3CrtHttpConfiguration.Builder,
...
         Builder connectionTimeout(Duration connectionTimeout);
 
+
+        /**
+         * <p>
+         *     Option to disable SSL cert validation and SSL host name verification.
+         *     This turns off x.509 validation.
+         *     By default, this option is off.
+         *     Only enable this option for testing purposes.
+         * </p>
```
zoewangg, line 138: nit: `</p>` is not needed;  a single `<p>` should be sufficient

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/S3NativeClientConfiguration.java
```diff
@@ -67,6 +67,11 @@ public S3NativeClientConfiguration(Builder builder) {
         TlsContextOptions clientTlsContextOptions =
             TlsContextOptions.createDefaultClient()
                              .withCipherPreference(TlsCipherPreference.TLS_CIPHER_SYSTEM_DEFAULT);
+
+        if (builder.httpConfiguration != null
+            && builder.httpConfiguration.shouldTrustAllCertificates() != null) {
+            clientTlsContextOptions.withVerifyPeer(!builder.httpConfiguration.shouldTrustAllCertificates());
```
zoewangg, line 73: Should we log a warning message here if it's enabled?
```
SSL Certificate verification is disabled. This is not a safe setting and should only be used for testing.
```

joviegas: I had following concern
-> User had disabled the SSL certificate validation for testing but user doesnot want anyone to know about it. If we are loggin it then we are making this info available to everyone who is accessing the logs.

Because of this this reason I was not sure if we should log this , what do you say about this scenario ?

zoewangg: We log a warning message in other HTTP clients, so it should be fine :) 
https://github.com/aws/aws-sdk-java-v2/blob/master/http-clients/apache-client/src/main/java/software/amazon/awssdk/http/apache/ApacheHttpClient.java#L698-L701

joviegas: Thanks

zoewangg (approved):

Issue #1735, opened 2020-03-25
Redundant byte array copying in AsyncRequestBody


RoKish:
## Describe the bug

It doesn't make sense to copy a byte array in `AsyncRequestBody#fromBytes` and `AsyncRequestBody#fromByteBuffer` since the byte array is copied in the constructor of `ByteArrayAsyncRequestBody` afterwards.

## Possible Solution
I think the best solution will be to copy the byte array in the constructor of `ByteArrayAsyncRequestBody`. Another constructor should be added to accept offset and length in order to support copying from `ByteBuffer`.

## Your Environment
* AWS Java SDK version used: `2.7.15`

debora-ito:
@RoKish seems reasonable, but it's more an enhancement than a bug.

If you can send us a PR, we'll be happy to take a look.

millems:
Fixed by #3925

millems:
@all-contributors please add @StephenFlavin for code

Issue #3193, opened 2022-05-16
S3 getObject combined with AsyncResponseTransformer.toBytes() copies too much data


chibenwa:
### Describe the bug

Given the following piece of code:

```
client.getObject(
                    builder -> builder.bucket(bucket),
                    AsyncResponseTransformer.toBytes())
```

Several data copies do take place:

 - Netty ByteBuf get's copied into a heap ByteBuffer. I understand this copy is done to present a common abstraction level for all transports. Yet maybe `ByteBuf::nioBuffers` could be relied on to minimize copies. I did not expore this option myself.
 - `AsyncResponseTransformer.toBytes()` relies internally on an unsized ByteArrayOutputStream (that thus will expend many times) and the result will be copied when calling `toByteArray()` and another defensive copy is carried other when transforming the future.

### Expected Behavior

Minimize memory copies in such a scenario to minimize heap pressure.

### Current Behavior

![Screenshot from 2022-05-16 10-20-07](https://user-images.githubusercontent.com/6928740/168514448-cdee7863-a5c8-45f0-8171-5f101648982a.png)

### Reproduction Steps

```
client.getObject(
                    builder -> builder.bucket(bucket),
                    AsyncResponseTransformer.toBytes())
```

I used https://github.com/jvm-profiling-tools/async-profiler for the flame graphs.

### Possible Solution

 - Remove needless defensive copies inside `ByteArrayAsyncResponseTransformer`.
 - Rely on `GetResponse::contentLength` to size a byte array and copy incoming buffers to it in place.
 
 I tested this successfully:
 
 ```
 package org.apache.james.blob.objectstorage.aws;

import java.nio.ByteBuffer;
import java.util.concurrent.CompletableFuture;

import org.reactivestreams.Subscriber;
import org.reactivestreams.Subscription;

import software.amazon.awssdk.core.ResponseBytes;
import software.amazon.awssdk.core.async.AsyncResponseTransformer;
import software.amazon.awssdk.core.async.SdkPublisher;
import software.amazon.awssdk.services.s3.model.GetObjectResponse;

/**
 * Class copied for {@link software.amazon.awssdk.core.internal.async.ByteArrayAsyncResponseTransformer}
 *
 * Modified to take advantage of the content length of the get response in order to use a sized array
 * upon content copy. This avoids the usage of a ByteArrayOutputStream that yields additional copies
 * (resizing upon copy, copy of the resulting byte array).
 *
[...]

PR #3925, opened 2023-04-20
Add "unsafe" AsyncRequestBody constructors for better performance when concurrent modification isn't a worry

StephenFlavin:
The static methods in `AsyncRequestBody` do some defensive copying and transforming of types that hamper performance and utilise more memory than needed if the user has already ensured either that concurrent modification won't happen or that it doesn't matter, this pr addresses that with new `AsyncRequestBody` constructors.

This also fixes an unnecessary copy being made in `AsyncRequestBody#fromByteBuffer` (https://github.com/aws/aws-sdk-java-v2/issues/1735) and `AsyncRequestBody#fromString` as a side effect.

## Motivation and Context

Memory consumption and performance.

Related Issues:
 - https://github.com/aws/aws-sdk-java-v2/issues/3193
 - https://github.com/aws/aws-sdk-java-v2/issues/1735
 - https://github.com/aws/aws-sdk-java-v2/issues/3928

## Modifications

Adds `AsyncRequestBody#fromBytesUnsafe(byte[])`, `AsyncRequestBody#fromByteBufferUnsafe(ByteBuffer)` and `AsyncRequestBody#fromByteBuffersUnsafe(ByteBuffer...)` as no copy alternatives to the existing methods.

Adds `AsyncRequestBody#fromByteBuffers(ByteBuffer...)`, `AsyncRequestBody#fromRemainingByteBuffer`, `AsyncRequestBody#fromRemainingByteBufferUnsafe`, `AsyncRequestBody#fromRemainingByteBuffers` and `AsyncRequestBody#fromRemainingByteBuffersUnsafe` methods as implementations that respect the current buffer positions.

Updates `AsyncRequestBodyTest` to junit 5.

Removes `ByteArrayAsyncRequestBody` 

Adds `ByteBuffersAsyncRequestBody` with static builders for `ByteBuffer`, `ByteBuffer[]` and `byte[]`.
 - calls to `ByteBuffersAsyncRequestBody#subscribe` are repeatable (not sure if this is a requirement but would be a breaking change vs `ByteArrayAsyncRequestBody`)
 - `ByteBuffersAsyncRequestBody#subscribe` creates a thread safe subscription.
 - `ByteBuffer[]` allows a single file > `Integer.MAX_VALUE` to be transferred.
 - A benefit of the implementation is `ByteBuffersAsyncRequestBody` doesn't copy an off heap ByteBuffer until it's requested by the subscription so depending on the subscriber less data will be on heap at a time.
     -   _I would like to know why off heap buffers need to be copied at all (see: issue https://github.com/aws/aws-sdk-java-v2/issues/3928)_

Adds `BinaryUtils#immutableCopyOf`, `BinaryUtils#immutableCopyOfRemaining` and `BinaryUtils#toNonDirectBuffer` convenience methods.

## Testing
Unit tests

## Types of changes

## Checklist
- [ ] ~Local run of `mvn install` succeeds~ https://github.com/aws/aws-sdk-java-v2/issues/3926
- [x] All new ~and existing~ tests passed

## License


core/sdk-core/src/main/java/software/amazon/awssdk/core/async/AsyncRequestBody.java
```diff
@@ -143,25 +143,33 @@ static AsyncRequestBody fromString(String string) {
...
-     * original byte array are not reflected in the {@link AsyncRequestBody}.
+     * Creates a {@link AsyncRequestBody} from a byte array.
      *
      * @param bytes The bytes to send to the service.
      * @return AsyncRequestBody instance.
      */
     static AsyncRequestBody fromBytes(byte[] bytes) {
-        return new ByteArrayAsyncRequestBody(bytes, Mimetype.MIMETYPE_OCTET_STREAM);
+        return ByteBuffersAsyncRequestBody.from(bytes);
     }
```
millems, line 153: Similar to `SdkBytes`, can we keep the defensive copy here, but introduce a `fromBytesUnsafe` to create it without the defensive copy, so it's more obvious to the caller that they should not mutate the byte array after it is passed?

core/sdk-core/src/main/java/software/amazon/awssdk/core/async/AsyncRequestBody.java
```diff
@@ -143,25 +143,33 @@ static AsyncRequestBody fromString(String string) {
...
-     * made to the original {@link ByteBuffer} are not reflected in the {@link AsyncRequestBody}.
+     * Creates a {@link AsyncRequestBody} from a {@link ByteBuffer}.
      *
      * @param byteBuffer ByteBuffer to send to the service.
      * @return AsyncRequestBody instance.
      */
     static AsyncRequestBody fromByteBuffer(ByteBuffer byteBuffer) {
-        return fromBytes(BinaryUtils.copyAllBytesFrom(byteBuffer));
+        return ByteBuffersAsyncRequestBody.of(null, byteBuffer);
+    }
```
millems, line 163: Same here. Can we add `fromByteBufferUnsafe` instead of changing the default behavior?

core/sdk-core/src/main/java/software/amazon/awssdk/core/async/AsyncRequestBody.java
```diff
@@ -143,25 +143,33 @@ static AsyncRequestBody fromString(String string) {
...
+
+    /**
+     * Creates a {@link AsyncRequestBody} from an array of {@link ByteBuffer}.
+     *
+     * @param byteBuffers ByteBuffer[] to send to the service.
+     * @return AsyncRequestBody instance.
+     */
+    static AsyncRequestBody fromByteBuffers(ByteBuffer... byteBuffers) {
```
millems, line 171: Can we call this `fromByteBuffersUnsafe`?

A non-blocking observation: This method makes an interesting point, that we don't have a way of joining AsyncRequestBody implementations. This makes it work for byte buffers, but strings, byte arrays, etc. still need to be concatenated.

I almost wonder if we should have a `AsyncRequestBody append(AsyncRequestBody)` method on `AsyncRequestBody` that appends the data from one body to another, so that we don't have to have special treatment of byte buffers here.

`AsyncRequestBody.fromByteBuffer(firstBuffer).append(AsyncRequestBody.fromByteBuffer(secondBuffer));`

I don't think that would obviate the need for a simple way of joining multiple, like we have here, though. If you had 10 byte buffers you wanted to join, that seems like a pain.

Again, I don't expect you to change anything here for this PR... just documenting an observation.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/ByteArrayAsyncRequestBody.java
```diff
@@ -30,8 +30,11 @@
  * @see AsyncRequestBody#fromBytes(byte[])
  * @see AsyncRequestBody#fromByteBuffer(ByteBuffer)
  * @see AsyncRequestBody#fromString(String)
+ *
+ * @deprecated by {@link ByteBuffersAsyncRequestBody#from}.
  */
 @SdkInternalApi
+@Deprecated
 public final class ByteArrayAsyncRequestBody implements AsyncRequestBody {
```
millems, line 38: If we no longer use this at all, we can delete it. Internal APIs don't need to be left for backwards-compatibility.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/ByteBuffersAsyncRequestBody.java
```diff
@@ -0,0 +1,136 @@
...
+                                return;
+                            }
+
+                            long remaining = n;
+
+                            do {
+                                ByteBuffer buffer = buffers[i];
+                                if (!buffer.hasArray()) {
+                                    buffer = ByteBuffer.wrap(BinaryUtils.copyBytesFrom(buffer));
+                                }
```
millems, line 96: Is this because we have assumptions downstream of this publisher that the buffer `hasArray`?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/ByteBuffersAsyncRequestBody.java
```diff
@@ -0,0 +1,136 @@
...
+                                s.onNext(buffer);
+                                remaining--;
+                            } while (remaining > 0 && (i = index.getAndIncrement()) < buffers.length);
+
+                            if (i >= buffers.length - 1 && completed.compareAndSet(false, true)) {
+                                s.onComplete();
+                            }
+                        } else {
+                            s.onError(new IllegalArgumentException("§3.9: non-positive requests are not allowed!"));
```
millems, line 105: `completed.set(true)` to drop future `request` calls?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/ByteBuffersAsyncRequestBody.java
StephenFlavin: I've done this to maintain how the existing code will convert direct buffers (off heap) to non-direct (on heap) buffers see: https://github.com/aws/aws-sdk-java-v2/issues/3928
I've tested with the Netty and CRT clients and neither seem to have any issues handling off-heap buffers but I was attempting to separate out the zero-copy functionality to future PR's

core/sdk-core/src/main/java/software/amazon/awssdk/core/async/AsyncRequestBody.java
StephenFlavin: yes these can all be renamed.
I thought it might be an issue to change the contracts here, although, I'm not sure if `Unsafe` is the right word in this context, as `DirectByteBuffers` are used with `sun.misc.Unsafe` to allow de-allocation of their off heap memory and it may lead to confusion.
I am finding it difficult to come up with another name though, `fromByteBufferDirect` (same issue), `fromMutableByteBuffer` maybe 🤷

One thing to note is that `AsyncRequestBody#fromFile` doesn't provide the same guarantee regarding concurrent modification.

What are your thoughts on the defensive copying?
personally I would say the developer can apply defensive copies themselves before calling these methods as it's not a complex operation, and I'm wondering if this change could be rolled into a major version bump with other changes such as what I've outlined on https://github.com/aws/aws-sdk-java-v2/issues/3928

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/ByteBuffersAsyncRequestBody.java
millems: > I've done this to maintain how the existing code will convert direct buffers (off heap) to non-direct (on heap) buffers see: https://github.com/aws/aws-sdk-java-v2/issues/3928
> I've tested with the Netty and CRT clients and neither seem to have any issues handling off-heap buffers but I was attempting to separate out the zero-copy functionality to future PR's

Got it, I agree it makes sense to keep it separate for now.

core/sdk-core/src/main/java/software/amazon/awssdk/core/async/AsyncRequestBody.java
millems: >  I thought it might be an issue to change the contracts here, although, I'm not sure if Unsafe is the right word in this context,DirectByteBuffers are used with sun.misc.Unsafe to allow de-allocation of their off heap memory and it may lead to confusion.

Yeah, "Unsafe" is meant to illustrate a clear delineation between "the SDK is making no assumptions about your input" and "the SDK is making some critical assumptions about your input". It doesn't have anything to do with `sun.misc.Unsafe`, although I agree that might be one potential confusion. Since we've already committed to the "unsafe" naming elsewhere for this case, though, I'd favor sticking with it.

> One thing to note is that AsyncRequestBody#fromFile doesn't provide the same guarantee regarding concurrent modification.

That's a good point. It is indeed inconsistent. I suppose doing things from disk and doing things from memory were just two different states of mind during our development.

> What are your thoughts on the defensive copying? personally I would say the developer can apply defensive copies themselves before calling these methods as it's not a complex operation, and I'm wondering if this change could be rolled into a major version bump with other changes such as what I've outlined on https://github.com/aws/aws-sdk-java-v2/issues/3928

I think it depends on the audience. If we can assume the developer is thoughtful and experienced, then I agree it's most intuitive to not do a defensive copy. Most people we talk with here on github are the thoughtful and experienced types.

Our assumption about the AWS SDK's customers is that they come from all levels of in-a-hurry and experience. A recently-hired intern, who has never worked with Java, and needs to finish up some work that's leftover from the person that they replaced probably won't be thinking about the mutability of the data they're passing around. They're struggling getting through all of the other AWS complexities that exist. We don't want them to miss their deadlines fighting a data mutability bug and then say "that's dumb, why does Java/AWS/the computer/etc. allow that to happen?" There's already enough places for customer's code to go wrong.

Our solution has been to make the obvious path "safe" and the less-obvious path trade some of that safety for performance. We make them use the method called "unsafe" if they want to prevent the copying.

StephenFlavin: Thank you for the great response, I'll make the changes you suggest as soon as I get a free moment

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/ByteBuffersAsyncRequestBody.java
StephenFlavin: In hindsight I'm not sure if this would be correct, I lifted this logic from `ByteArrayAsyncRequestBody` so there's no logic change here, and I'm not sure if the this subscription would have to be marked as completed if it errors.
What do you think?

core/sdk-core/src/main/java/software/amazon/awssdk/core/async/AsyncRequestBody.java
StephenFlavin: done

core/sdk-core/src/main/java/software/amazon/awssdk/core/async/AsyncRequestBody.java
StephenFlavin: fyi, I added both the safe and unsafe implementation for this in the PR

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/ByteBuffersAsyncRequestBody.java
millems: I think it's correct, but this case probably never happens so I doubt we're testing it. If it's already like this, I'm not really opposed to leaving it like this. We need to move this to an easier-to-use publisher pattern anyway.

core/sdk-core/src/main/java/software/amazon/awssdk/core/async/AsyncRequestBody.java
```diff
@@ -150,18 +151,69 @@ static AsyncRequestBody fromString(String string) {
...
      */
     static AsyncRequestBody fromByteBuffer(ByteBuffer byteBuffer) {
-        return fromBytes(BinaryUtils.copyAllBytesFrom(byteBuffer));
+        ByteBuffer immutableCopy = BinaryUtils.immutableCopyOf(byteBuffer);
+        immutableCopy.rewind();
+        return ByteBuffersAsyncRequestBody.of(null, immutableCopy);
+    }
+
+    /**
+     * Creates a {@link AsyncRequestBody} from a {@link ByteBuffer}.
```
millems, line 184: Nit: Can we add similar wording to what we have in [SdkBytes](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/core/SdkBytes.html)'s unsafe methods? Clarifying why it is called "unsafe", for example.

utils/src/main/java/software/amazon/awssdk/utils/BinaryUtils.java
```diff
@@ -117,6 +117,60 @@ public static ByteArrayInputStream toStream(ByteBuffer byteBuffer) {
...
     }
 
+    /**
+     * Returns an immutable copy of the given {@code ByteBuffer}.
+     * <p>
+     * The new buffer's position will be set to the position of the given {@code ByteBuffer}, but the mark if defined will be
+     * ignored.
+     * <p>
+     * Note: this method intentionally converts direct buffers to non-direct, see
+     * <a href="https://github.com/aws/aws-sdk-java-v2/issues/3928">issue #3928</a>
```
millems, line 127: Nit: I don't know if we should bake this intention into this method, since it will likely be used elsewhere in the future. Can we just say that this is the current behavior, and leave it at that?

core/sdk-core/src/main/java/software/amazon/awssdk/core/async/AsyncRequestBody.java
StephenFlavin: Done, I've updated each methods javadoc to be consistent with this wording

utils/src/main/java/software/amazon/awssdk/utils/BinaryUtils.java
StephenFlavin: should I add a note that this may change in future?

StephenFlavin: done

core/sdk-core/src/main/java/software/amazon/awssdk/core/async/AsyncRequestBody.java
```diff
@@ -143,29 +144,96 @@ static AsyncRequestBody fromString(String string) {
...
+     *
+     * <p>As the method name implies, this is unsafe. Use {@link #fromByteBuffer(ByteBuffer)}} unless you're sure you know the
+     * risks.
+     *
+     * @param byteBuffer ByteBuffer to send to the service.
+     * @return AsyncRequestBody instance.
+     */
+    static AsyncRequestBody fromByteBufferUnsafe(ByteBuffer byteBuffer) {
+        return ByteBuffersAsyncRequestBody.of(null, byteBuffer);
+    }
```
millems, line 201: Sorry for the additional comments, since it felt like we were done, but I've just realized that semantics of these two seem to be different for the position. It seems like fromByteBuffer starts from position 0 (for some reason) and fromByteBufferUnsafe does not (which seems more correct).

Should we do the (probably less correct) thing for `fromByteBufferUnsafe` with `byteBuffer.asReadOnlyBuffer().rewind()`? (same with `fromByteBuffersUnsafe`)

Later on, we should probably deprecate `fromByteBuffer` for a new (more correct) `fromByteBufferPosition` and `fromByteBufferPositionUnsafe` that actually honors the current position of the buffer.

StephenFlavin: Yes I noticed this too, looking at [`RequestBody`](https://github.com/aws/aws-sdk-java-v2/blob/master/core/sdk-core/src/main/java/software/amazon/awssdk/core/sync/RequestBody.java#L174-L198) I think adding the implementation for `fromRemainingByteBuffer`, `fromRemainingByteBufferUnsafe` and `fromRemainingByteBufferUnsafe` makes sense.

StephenFlavin: I think this method ensuring it passes along a read only buffer makes sense too to mitigate
>(2) any users of {@link #fromByteBufferUnsafe(ByteBuffer)} to modify the
     * {@link ByteBuffer} passed into this {@code AsyncRequestBody} implementation.

StephenFlavin: I've made the changes and kept them in a [separate commit](https://github.com/aws/aws-sdk-java-v2/pull/3925/commits/1fd1420bfd97d796d2c90f846e61ff7c6e87be0a) to make it easier to review, if you're happy I'll merge these commits.

millems: LGTM, let's merge!

StephenFlavin: commits merged 🙂

millems (approved):

millems:
Thanks for your contribution! I'm going to update it from the target branch and then run the tests.

millems:
`ChecksumCalculatingAsyncRequestBodyTest.bytesBufferConstructorHasCorrectContentType` in `sdk-core` seems to be failing. Can you try to repro on your end?

StephenFlavin:
@millems the tests were expecting the body to have the `contentLength` set which by default was `Optional.empty()`
I've changed the static `ByteBuffersAsyncRequestBody#of` method to sum the total `ByteBuffer#remaining` which fixes it and added unit tests to validate it.
Apologies I've not been able to get a full build to complete locally, running into issues with [flakey tests](https://github.com/aws/aws-sdk-java-v2/issues/3926) and python.
I've run the sdk-core and utils units successfully at least once.

millems:
@StephenFlavin Awesome, let me kick off the suites again!

StephenFlavin:
It looks like some of the builds have failed, is there a way for me to see the output @millems?
I'm assuming this is because of the flakey tests I mentioned

millems:
@StephenFlavin Sorry, I was off yesterday. I wish there was a way for you to see the output. We're still iterating on the security of our testing solution, trying to figure out how to expose the logs.

It looks like test failures in `:protocol-tests`, and `:codegen-generated-classes-test` this time. Those modules do functional tests of the SDK clients, so they sometimes catch issues that appear in integration of components. Are you able to reproduce those errors yourself by building the modules? Otherwise I can pull the specific test failures.

StephenFlavin:
there does appear to just be some unstable tests on master e.g. https://github.com/aws/aws-sdk-java-v2/commit/c015a19518941d36f3e85ddfc96714d2d95e66c3

StephenFlavin:
I've been able to get them running, `:protocol-tests` the only failing tests I have are integration tests because I don't have IAM credentials I can use re: https://github.com/aws/aws-sdk-java-v2/blob/master/docs/GettingStarted.md#integration-tests
The `:codegen-generated-classes-test` tests do seem to have genuine failures in [`AsyncStreamingCoreMetricsTest`](https://github.com/aws/aws-sdk-java-v2/blob/master/test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/metrics/async/AsyncStreamingCoreMetricsTest.java), I'll have a closer look at this one.
```
Request content was only 0 bytes, but the specified content-length was 10 bytes.

**Comment by StephenFlavin** [comment]
I've not had a chance to spend time on this, will try to get it done this week

**Comment by StephenFlavin** [comment]
@millems test failures fixed.

The issues was `ByteBuffersAsyncRequestBody` didn't support multiple subscribers properly as it was just passing along the same `ByteBuffer` which it was holding a reference to, so if one subscriber read the buffer the other subscriber would attempt to read the same buffer but it had already been consumed, the `Subscriber#onNext` call now does `buffer.asReadOnlyBuffer()` and I've added a test to verify this requirement in future.

This is a change in comparison to how `ByteArrayAsyncRequestBody` worked as it would have been passing along mutable copies but I don't think this is an issue.

To support this I had to fix an issue in `ChunkBuffer.java` where it was expecting `ByteBuffer#array` to work which is an optional operation for `ByteBuffer` implementations.
While fixing this I've identified an issue with `BinaryUtils#copyAllBytesFrom` where it treats implementations that support `ByteBuffer#array` as `BinaryUtils#copyRemainingBytesFrom` because it respects the current position doing 
```java
        if (bb.hasArray()) {
            return Arrays.copyOfRange(
                    bb.array(),
                    bb.arrayOffset(),
                    bb.arrayOffset() + bb.limit());
        }
```
this should just be doing 
```java
        if (bb.hasArray()) {
            return bb.array().clone;
        }
```
I've opted to leave this alone as part of this PR, shall I raise a separate PR for this or shall I open an issue?

millems:
Re-running tests.

millems (approved):

millems:
Thank you for your contribution! It has been merged and will go out with the next regular release.

StephenFlavin:
@millems or @debora-ito,
This PR resolves https://github.com/aws/aws-sdk-java-v2/issues/1735 so that can be closed 🙂 
Also would it be possible to be added to the all-contributors thing 🙏

millems:
Sorry, hadn't realized we didn't add you!

StephenFlavin:
np, thanks 🙂

Issue #52, opened 2017-07-06
Cross-Region S3 Client


millems:
The 1.11.x S3 client automatically routed requests to the correct bucket region in a complicated manner. The 2.0.x client does not perform this automatic re-routing.

We should create a wrapper to pool S3 clients for different regions and automatically route requests to the region in which the bucket was created.

manasshukla:
Is there an update on when this feature would be shipped out ?

millems:
Sorry, no update. 

It's not our #1 priority, because with [access points](https://aws.amazon.com/s3/features/access-points/) and other arn-based ways of accessing buckets/objects plus the [use-arn-region](https://github.com/aws/aws-sdk-java-v2/blob/master/services/s3/src/main/java/software/amazon/awssdk/services/s3/S3Configuration.java#L272-L279) flag, the SDK will automatically route the request to the correct region *without* the overhead of an extra call to AWS to find out where the bucket actually is.

I understand there's still value in this feature request, since the caller may not have a full bucket ARN.

byF:
If anyone thinks using ARNs like `arn:aws:s3:::mybucket` will work, they won't https://github.com/aws/aws-sdk-java-v2/issues/2244

pkgonan:
@millems Hi. Any plan of supporting this feature ?

millems:
At some point, probably, but in the meantime you could try using access points which automatically support cross-region request routing.

pkgonan:
@millems 
Oh. That's good news. When are you expecting?

millems:
@pkgonan Unfortunately we have no expected date to report at this time.

pkgonan:
@millems Thanks for your reply. :)

maciejwalkowiak:
FYI we are working on adding cross-region S3 client so Spring Cloud AWS. It's early stage but looks promising so far: https://github.com/awspring/spring-cloud-aws/blob/3.0.x-s3/spring-cloud-aws-s3-parent/spring-cloud-aws-s3/src/main/java/io/awspring/cloud/s3/CrossRegionS3Client.java

Basic idea is that every region-specific method is wrapped in a try catch that on http status 301 uses client created for the region where the bucket got created. Clients are cached both by region and also by bucket name (to speed up subsequent calls for the same bucket). Implementation is a bit Spring specific as we use LRU cache from `spring-core`, but if needed this can be replaced with something else or abstracted out. 

[...]

PR #3937, opened 2023-04-21
Create design doc for cross-region S3 client

zoewangg:
## Motivation and Context
#52 

## Modifications
Create design doc for cross-region S3 client

### DIFF
--- a/docs/design/services/s3/cross-region-s3-client.md
+++ b/docs/design/services/s3/cross-region-s3-client.md
@@ -1,0 +1,57 @@
+**Design:** New Feature, **Status:** [In Development](../README.md)
+
+# Cross-Region S3 Client Design
+
+## Overview
+
+The [cross-region S3 client](https://github.com/aws/aws-sdk-java-v2/issues/52) is a feature supported in the AWS SDK for Java 1.x that automatically routes requests to the correct bucket region. This is useful for customers who do not know the region of the bucket beforehand. Without this support, customers would need to write their own logic to retrieve the region of a specific bucket for each request and redirect the request accordingly, which could be complex and error prone. This documentation proposes the design to implement it in the AWS SDK for Java 2.x.
+
+## Specification 
+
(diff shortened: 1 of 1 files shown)


docs/design/services/s3/cross-region-s3-client.md
```diff
@@ -0,0 +1,57 @@
...
+
+// AWS CRT-based S3 async client
+S3AsyncClient s3CrtClient = S3AsyncClient.crtBuilder()
+                                         .crossRegionAcccessEnabled(true)
+                                         .build();
+```
+
+### Client Configuration
+
+Users can enable this feature through a client configuration on the client builder, `crossRegionAccessEnabled`. It is default to false.
```
L-Applin, line 39: small typo: `It is default to false`
should be `It is set to false by default` or something like that

docs/design/services/s3/cross-region-s3-client.md
```diff
@@ -0,0 +1,57 @@
...
+## Specification 
+
+The SDK 2.x will support this feature in the S3 sync client, S3 async client and the AWS-CRT based S3 client.
+
+### Usage Examples
+
+#### Example 1: enabling cross-region access in the S3 sync client
+
+```
+S3ClientBuilder s3ClientBuilder = S3Client.builder()
```
joviegas, line 18: The `crossRegionAcccessEnabled` API will be made available to only "S3Client.builder"  and not to `AwsClientBuilder` which will make it available for all the clients even  if its not supported.
Thus we might need to add Customization for S3 only to show this API.

zoewangg: Yes, alternatively, we could consider adding it to `S3Configuration`, I don't have strong opinions. We still need to add `crossRegionAcccessEnabled` to the top level S3CrtAsyncClient builder though.

zoewangg:
Closing because it's obsolete.

PR #3950, opened 2023-04-25
Fixed bug where using profile-based credentials could cause the SDK to read the profile file with each request.

millems:
This bug was introduced with the profile file supplier changes. If the customer was not overriding the profile file via the client override config AND was relying on profile file-based configuration, that configuration would be read with every request. This had a significant performance cost.

```
Benchmarking Results (software.amazon.awssdk.benchmark.apicall.protocol.JsonProtocolBenchmark)
Before:   2858.805 ± 181.491  ops/s
After:    10915.079 ± 677.022  ops/s
```


core/sdk-core/src/main/java/software/amazon/awssdk/core/client/builder/SdkDefaultClientBuilder.java
```diff
@@ -264,9 +264,9 @@ protected SdkClientConfiguration mergeChildDefaults(SdkClientConfiguration confi
      */
     private SdkClientConfiguration mergeGlobalDefaults(SdkClientConfiguration configuration) {
         // Don't load the default profile file if the customer already gave us one.
-        Supplier<ProfileFile> configuredProfileFileSupplier = configuration.option(PROFILE_FILE_SUPPLIER);
-        Supplier<ProfileFile> profileFileSupplier = Optional.ofNullable(configuredProfileFileSupplier)
-                                                            .orElseGet(() -> ProfileFile::defaultProfileFile);
```
cenedhryn, line 269: Where does the potential read-every-request happen?

cenedhryn: Ok, and now we're wrapping it in another layer. Tricky!

cenedhryn (approved):

PR #3964, opened 2023-05-02
Fixed the issue where slash from path parameter got removed with endp…

zoewangg:
## Motivation and Context

Fixed the issue where slash from path parameter got removed with endpoint override that contains a trailing slash.

Suppose this is the endpoint override: `https://override.example.com/` (note the trailing slash), if the path parameter contains a slash, `/foo`, the generated request path currently is `/foo` instead of `//foo`

## Modifications
~~Removing trailing slash from endpoint~~
Bypass path combination if the path resolved by resolver is the same as the endpoint path, i.e., the resolver doesn't modify anything.

## Testing
Added unit tests and ran integ tests locally.

## Types of changes

## Checklist

## License


codegen/src/main/resources/software/amazon/awssdk/codegen/rules/AwsEndpointProviderUtils.java.resource
```diff
@@ -150,19 +150,38 @@ public final class AwsEndpointProviderUtils {
         String clientEndpointPath = clientEndpoint.getRawPath();
 
         // [client endpoint path]/[request path]
-        String requestPath = request.getUri().getRawPath();
+        String requestPath = request.encodedPath();
```
zoewangg, line 153: @dagnir `request.getUri().getRawPath()` should be the same as `encodedPath()` right? This is a micro performance optimization (because there's no need to parse the URI)

dagnir: Yep good catch, it should be: https://github.com/aws/aws-sdk-java-v2/blob/4dd34b1c7ad5a26d0d072c30d3dccda4e1a3aa7b/http-client-spi/src/main/java/software/amazon/awssdk/http/SdkHttpRequest.java#L157

dagnir (approved):

millems (approved):

Issue #3629, opened 2022-12-15
Netty HTTP client blocking on DNS resolution


yanns:
### Describe the bug

Netty threads are being blocked when calling InetAddress.getByName  (https://docs.oracle.com/javase/6/docs/api/java/net/InetAddress.html#getByName(java.lang.String))

<img width="783" alt="Screenshot 2022-12-15 at 14 43 28" src="https://user-images.githubusercontent.com/51669/207884323-3a161223-d5a0-4d90-a363-3029b867729c.png">

This impacts the scalability of the AWS java library.

### Expected Behavior

Netty threads should not be blocked

### Current Behavior

Netty threads are being blocked

### Reproduction Steps

Using the `NettyNioAsyncHttpClient`
```scala
  val builder = NettyNioAsyncHttpClient.builder()
  val httpClient: SdkAsyncHttpClient = builder.build()
```

### Possible Solution

A possible solution is to use netty-resolver-dns like in https://github.com/netty/netty/blob/4.1/example/src/main/java/io/netty/example/dns/tcp/TcpDnsClient.java

### AWS Java SDK version used

2.18.35

### JDK version used

17.0.5_8

### Operating System and version

linux

martinKindall:
Hi @yanns , which tool are you using to show the graphics you attached?

yanns:
The JVM is profiled with https://github.com/async-profiler/async-profiler.
The profile file is then opened by JDK mission control where the screenshot is coming from.

martinKindall:
Thanks @yanns, I installed JDK mission and async-profiler and both work. I already took a flamegraph profile. 

I am having issues reproducing your results though using the .jfr format for the profile output. I can see the stacktrace correctly but the profiler is not saving the "Total Blocked" time, here's my command

`./profiler.sh -d 20 -i 5ms -f result3.jfr <PID>`

Would you mind sharing your command? So I can compare the args you used. Thanks.

yanns:
It's a long time ago, but I think I used the following flags:
`-e cpu,alloc,lock,itimer`

debora-ito:
[...]

PR #3990, opened 2023-05-10
Now it's possible to configure NettyNioAsyncHttpClient for non blocking DNS

martinKindall:
## Motivation and Context
https://github.com/aws/aws-sdk-java-v2/issues/3629

## Modifications
I added a new configuration flag to allow **NettyNioAsyncHttpClient** under the hood to use non blocking DNS resolution.
The default behaviour remains the same, for backward compatibility. In the future it could be considered to make the non blocking resolution the default.

I used https://github.com/lettuce-io/lettuce-core/issues/1498 as reference.

## Testing
The included tests were tested running the unit tests of that module with ./mvnw package
Testing environment

Java version: 1.8.0_361, vendor: Oracle Corporation
OS name: "windows 10", version: "10.0", arch: "amd64"

## Types of changes

## Checklist

## License


http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/BootstrapProvider.java
```diff
@@ -56,8 +61,21 @@ public Bootstrap createBootstrap(String host, int port) {
...
+
         sdkChannelOptions.channelOptions().forEach(bootstrap::option);
 
         return bootstrap;
     }
+
+    private AddressResolverGroup<InetSocketAddress> nonBlockingResolverGroup() {
+        DnsNameResolverBuilder builder = new DnsNameResolverBuilder()
+            .channelType(NioDatagramChannel.class)
+            .resolveCache(NoopDnsCache.INSTANCE);
```
millems, line 77: What's the reasoning for a no-op cache, instead of the default behavior?

http-client-spi/src/main/java/software/amazon/awssdk/http/SdkHttpConfigurationOption.java
```diff
@@ -131,6 +131,16 @@
...
 
+    /**
+     * Configure whether to use a non-blocking dns resolver or not. False by default, as netty's default dns resolver is
+     * blocking; it namely calls java.net.InetAddress.getByName.
+     * <p>
+     * When enabled, a non-blocking dns resolver will be used instead, by modifying netty's bootstrap configuration.
+     * See https://netty.io/news/2016/05/26/4-1-0-Final.html
+     */
+    public static final SdkHttpConfigurationOption<Boolean> USE_NONBLOCKING_DNS_RESOLVER =
+        new SdkHttpConfigurationOption<>("UseNonBlockingDnsResolver", Boolean.class);
```
millems, line 142: Let's leave this off of the default options, since no other HTTP clients use it yet. Prior art: https://github.com/aws/aws-sdk-java-v2/blob/7d302ee48653482b71ab3e650c1b2edb6aa0b4c6/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/NettyNioAsyncHttpClient.java#L662-L666

http-clients/netty-nio-client/pom.xml
```diff
@@ -85,6 +85,14 @@
...
             <artifactId>netty-transport-classes-epoll</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.netty</groupId>
+            <artifactId>netty-resolver</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.netty</groupId>
+            <artifactId>netty-resolver-dns</artifactId>
+        </dependency>
```
millems, line 95: This is an extra 161 KB, for a feature that's off by default. 

It should probably be `optional` for now, with updated javadoc to specify that it requires an extra dependency on `netty-resolver-dns`. Can we also update the error message they get if they try to enable it without the extra dependency to something like "hey, you forgot to depend on io.netty:netty-resolver-dns"? Prior art: https://github.com/aws/aws-sdk-java-v2/blob/7d302ee48653482b71ab3e650c1b2edb6aa0b4c6/core/auth/src/main/java/software/amazon/awssdk/auth/signer/SignerLoader.java#L56-L58

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/BootstrapProvider.java
```diff
@@ -56,8 +61,21 @@ public Bootstrap createBootstrap(String host, int port) {
...
+        }
+
         sdkChannelOptions.channelOptions().forEach(bootstrap::option);
 
         return bootstrap;
     }
+
+    private AddressResolverGroup<InetSocketAddress> nonBlockingResolverGroup() {
+        DnsNameResolverBuilder builder = new DnsNameResolverBuilder()
+            .channelType(NioDatagramChannel.class)
```
millems, line 76: Should we use the event loop group channel factory instead of hard-coding to NioDatagramChannels?

i.e. `.channelFactory(sdkEventLoopGroup.channelFactory())`

martinKindall: But the expected type is different, `.channelType` expects `Class<? extends DatagramChannel>`, and channelFactory returns 
`ChannelFactory<? extends Channel>`. Or do you suggest I should create a channelType() method in `sdkEventLoopGroup`.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/BootstrapProvider.java
millems: Ah, interesting. I think the existing DNS resolver uses the JVM's DNS cache. I don't know the context for that requirement, let me investigate further...

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/BootstrapProvider.java
millems: `channelType` just delegates to `channelFactory`, so I was asking if we could just use `channelFactory` directly from this code. Now I see that `channelFactory` takes a `ChannelFactory<? extends DatagramChannel>` instead of a `ChannelFactory<? extends Channel>`, so my suggestion still won't work.

The concern I'm trying to fix is that in `SocketChannelResolver` chooses KQueue or Oio socket channels in some circumstances, and I'd like to repeat that decision-making logic for the datagram channels. Should we rename `SocketChannelResolver` to `ChannelResolver` and add a `resolveDatagramChannelFactory` so that we use the right datagram channel implementation for the platform?

martinKindall: Makes sense, I'll find a way to implement that.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/BootstrapProvider.java
millems: It looks like that comment is referring to a bug we had in the past that seems unrelated to this.

The bug was that we'd pass a _resolved_ RemoteAddress to netty here: https://github.com/aws/aws-sdk-java-v2/blob/master/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/BootstrapProvider.java#L58

This meant that any requests to the same hostname wouldn't have its DNS resolved by Netty at all. The comment was trying to prevent us from resolving the address ourselves before passing it to Netty, so that Netty would do the resolution.

I think we can still enable the DNS cache in this case, since it's essentially replacing the JVM DNS cache in the old code path, not the "cached" DNS lookup of the original bug.

http-clients/netty-nio-client/pom.xml
martinKindall: While implementing this, I stumped upon a decision: I want to reuse the `ClassLoaderHelper.java` class used in SignerLoader and other classes to dynamically instantiate dependencies. The problem is that this loader resides in the **sdk-core** module. 

On the other hand, **netty-nio-client** does not declare this sdk module. Now I don't now if it's worth this dependency, if it is heavier than 161 KB then it beats the purpouse. I don't know how heavy it is, should I include it? 

Other solution I thought would be just duplicate a small portion of `ClassLoaderHelper.java` inside the **netty-nio-client**, just the necessary to load classes.

millems: Hmm, we certainly can't add the dependency, since it would create a cyclic relationship. Honestly, the `auth` module shouldn't even be depending on `ClassLoaderHelper`, since it's a module-internal API. 

Can you create `ClassLoaderHelper` in the `utils` module and use it from there? 

We should go back and move all uses of `ClassLoaderHelper` from the `sdk-core` copy to the `utils` module, and deprecate-for-removal the one in `sdk-core`. I won't ask you to do the moving-everything-else part of it, since that seems like it should be a separate PR.

martinKindall: Fair enough, I can create create `ClassLoaderHelper` within utils. 

By the way, I will be away for the rest of the week, so I expect to continue on this PR starting next week. Looking forward to it.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/BootstrapProvider.java
martinKindall: I pushed changes that include this solution.

http-clients/netty-nio-client/pom.xml
martinKindall: Done.

http-client-spi/src/main/java/software/amazon/awssdk/http/SdkHttpConfigurationOption.java
martinKindall: Done.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/BootstrapProvider.java
martinKindall: Done.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/DnsResolverLoader.java
```diff
@@ -0,0 +1,62 @@
...
+        try {
+            Class<?> addressResolver = ClassLoaderHelper.loadClass(getAddressResolverGroup(), false, (Class) null);
+            Class<?> dnsNameResolverBuilder = ClassLoaderHelper.loadClass(getDnsNameResolverBuilder(), false, (Class) null);
+
+            Object dnsResolverObj = dnsNameResolverBuilder.newInstance();
+            Method method = dnsResolverObj.getClass().getMethod("channelType", Class.class);
+            method.invoke(dnsResolverObj, channelType);
+
+            Object e = addressResolver.getConstructor(dnsNameResolverBuilder).newInstance(dnsResolverObj);
+            return (AddressResolverGroup<InetSocketAddress>) e;
```
martinKindall, line 45: We don't specify a DNS Caching policy, which means [the default](https://netty.io/4.1/api/io/netty/resolver/dns/DefaultDnsCache.html) will be used.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/utils/ChannelResolver.java
```diff
@@ -21,26 +21,37 @@
...
-        KNOWN_EL_GROUPS.put("io.netty.channel.oio.OioEventLoopGroup", "io.netty.channel.socket.oio.OioSocketChannel");
+        KNOWN_EL_GROUPS_CHANNELS.put("io.netty.channel.kqueue.KQueueEventLoopGroup",
+                                     "io.netty.channel.kqueue.KQueueSocketChannel");
+        KNOWN_EL_GROUPS_CHANNELS.put("io.netty.channel.oio.OioEventLoopGroup",
+                                     "io.netty.channel.socket.oio.OioSocketChannel");
+
+        KNOWN_EL_GROUPS_DATAGRAMS.put("io.netty.channel.kqueue.KQueueEventLoopGroup",
+                                      "io.netty.channel.kqueue.KQueueDatagramChannel");
+        KNOWN_EL_GROUPS_DATAGRAMS.put("io.netty.channel.oio.OioEventLoopGroup",
+                                      "io.netty.channel.socket.oio.OioDatagramChannel");
```
martinKindall, line 51: The styling was complaining for the long lines, so I decided to split them like this.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/SdkEventLoopGroup.java
```diff
@@ -105,6 +130,20 @@ public static SdkEventLoopGroup create(EventLoopGroup eventLoopGroup, ChannelFac
...
+     *
+     * @param eventLoopGroup the EventLoopGroup to be used
+     * @param channelFactory the channel factor to be used
+     * @param channelType the channel type to be used
+     * @return a new instance of SdkEventLoopGroup
+     */
+    public static SdkEventLoopGroup create(EventLoopGroup eventLoopGroup, ChannelFactory<? extends Channel> channelFactory,
+                                           Class<? extends DatagramChannel> channelType) {
+        return new SdkEventLoopGroup(eventLoopGroup, channelFactory, channelType);
+    }
```
millems, line 145: I think we can skip adding this constructor. If people really want to configure the datagram channel type, they can do it using the builder.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/SdkEventLoopGroup.java
```diff
@@ -93,6 +111,13 @@ public ChannelFactory<? extends Channel> channelFactory() {
         return channelFactory;
     }
 
+    /**
+     * @return the {@link ChannelFactory} to be used with Netty Http Client.
+     */
+    public Class<? extends DatagramChannel> channelType() {
+        return channelType;
+    }
```
millems, line 119: Should we do `ChannelFactory<? extends DatagramChannel> datagramChannelFactory()` to be symmetric with the other channel factory? We'd use that type everywhere in here.

It would also avoid the overhead of the `ReflectiveChannelFactory`'s reflection whenever channels are created.

utils/src/main/java/software/amazon/awssdk/utils/internal/ClassLoaderHelper.java
```diff
@@ -0,0 +1,150 @@
...
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.utils.internal;
+
+
+import software.amazon.awssdk.annotations.SdkInternalApi;
+
+@SdkInternalApi
+public final class ClassLoaderHelper {
```
millems, line 22: Nit: Let's move this to `software.amazon.awssdk.utils` and mark it as an `@SdkProtectedApi`.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/SdkEventLoopGroup.java
martinKindall: That wouldn't be possible, I tried but as an example, `NioSocketChannel::new` doesn't satisfy the type `ChannelFactory<? extends DatagramChannel>`.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/SdkEventLoopGroup.java
martinKindall: I see, but then itwouldn't be posible to resolve channelType based on the eventLoopGroup, just like it's being done with the [channelFactory here](https://github.com/aws/aws-sdk-java-v2/blob/master/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/SdkEventLoopGroup.java#LL108C5-L120C6).

millems: I'm trying to avoid the pattern where we add a new constructor each time we add a new builder parameter. We generally only add a constructor if it's particularly common to create that combo. I don't think it's particularly common for people to use the `create(EventLoopGroup, ChannelFactory)` method we have today, and I'd expect even less common for people to use this new one with three parameters. If you think I'm wrong, I am fine adding it.

We could still allow for resolving the channelType based on the eventLoopGroup by keeping things private.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/SdkEventLoopGroup.java
martinKindall: Oh, now I see what you meant. Alright, changes pushed.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/SdkEventLoopGroup.java
martinKindall: You were right, I could resolve the datagramChannelFactory internally using the already existing constructors, so no additional constructor was added. 

I modified the builder to allow configuration of channel factories.

Changed pushed.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/NettyNioAsyncHttpClient.java
```diff
@@ -716,6 +727,16 @@ public void setHttp2Configuration(Http2Configuration http2Configuration) {
...
 
+        @Override
+        public Builder useNonBlockingDnsResolver(boolean useNonBlockingDnsResolver) {
+            this.useNonBlockingDnsResolver = useNonBlockingDnsResolver;
+            return this;
+        }
+
+        public void setUseNonBlockingDnsResolver(boolean useNonBlockingDnsResolver) {
+            useNonBlockingDnsResolver(useNonBlockingDnsResolver);
+        }
```
millems, line 738: Nit: we use boxed types (`Boolean`) for configuration, so that we can differentiate between `null` ("use default") and `false` (disable). It's useful if we ever need to change the defaults in the future.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/DnsResolverLoader.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    public static AddressResolverGroup<InetSocketAddress> init(ChannelFactory<? extends DatagramChannel> datagramChannelFactory) {
+        try {
+            Class<?> addressResolver = ClassLoaderHelper.loadClass(getAddressResolverGroup(), false, (Class) null);
+            Class<?> dnsNameResolverBuilder = ClassLoaderHelper.loadClass(getDnsNameResolverBuilder(), false, (Class) null);
+
+            Object dnsResolverObj = dnsNameResolverBuilder.newInstance();
+            Method method = dnsResolverObj.getClass().getMethod("channelType", Class.class);
+            method.invoke(dnsResolverObj, datagramChannelFactory.newChannel().getClass());
```
millems, line 43: More efficient:
```java
Method method = dnsResolverObj.getClass().getMethod("channelFactory", ChannelFactory.class);
method.invoke(dnsResolverObj, datagramChannelFactory);
```

millems (commented):
Nice work! A few minor non-functional comments. I'll kick off the testing now. Once the tests are passing and we close out those last comments, I'll merge!

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/DnsResolverLoader.java
martinKindall: Done.

I added more tests to increase coverage.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/NettyNioAsyncHttpClient.java
martinKindall: Done.

millems:
I've merged master into your branch and have kicked off the automated tests.

millems (approved):

millems:
Tests are passing! Thanks for your contribution. We'll have to do some testing to make sure this new optional dependency will work with our automation. Once that testing is complete, we'll merge.

martinKindall:
> Can you merge in the `millem/master/import-non-blocking-dns` branch from the upstream repo? It adds import configuration for the new dns dependency. Once you've done that, I'll merge this PR!
> 
> [b35e9ad](https://github.com/aws/aws-sdk-java-v2/commit/b35e9ad77095c96d4d4c9ad901a232360aa3ccb5)

Done.

millems:
Thank you for your contribution! I've merged the change, and it will go out with the next regular SDK release.

PR #3992, opened 2023-05-11
Fix javadocs warnings in the sdk-core package

sugmanue:
## Motivation and Context

This change fixes all the current javadoc warnings in the sdk-core package. This was done to avoid introducing new ones in my upcoming changes.

## Modifications

Removed all the current warnings by fixing some incorrect uses of javadocs, such as

* Use of literal < instead of the `&lt;` entity.
* Add imports for unknown classes referenced in `{@link …}`
* Avoid using `@see` when the doc meant to use `{@link …}`
* Avoid links to private fields, instead rephrase the intent

## Types of changes

## Checklist

## License

### DIFF
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/SdkNumber.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/SdkNumber.java
@@ -27,20 +27,23 @@ import software.amazon.awssdk.utils.Validate;
  * An in-memory representation of Number being given to a service or being returned by a service.
  * This is a SDK representation of a Number. This allows conversion to any desired numeric type by providing constructors
  * as below
- * @see #fromBigDecimal(BigDecimal) to create from a BigDecimal.
- * @see #fromBigInteger(BigInteger) to create from a BigInteger.
- * @see #fromDouble(double) to create from a double
- * @see #fromFloat(float) to create from a float.
- * @see #fromLong(long) to create from a long.
- * @see #fromShort(short) to create from a short.
- * @see #fromInteger(int) to create from an integer.
- * @see #fromString(String)  to create from a Stringl,
- * Thus by doing this, this class is able to preserve arbitrary precison of any given number.
+ * <ul>
+ *   <li>{@link #fromBigDecimal(BigDecimal)} to create from a BigDecimal.</li>
+ *   <li>{@link #fromBigInteger(BigInteger)} to create from a BigInteger.</li>
+ *   <li>{@link #fromDouble(double)} to create from a double</li>
+ *   <li>{@link #fromFloat(float)} to create from a float.</li>
+ *   <li>{@link #fromLong(long)} to create from a long.</li>
+ *   <li>{@link #fromShort(short)} to create from a short.</li>
+ *   <li>{@link #fromInteger(int)} to create from an integer.</li>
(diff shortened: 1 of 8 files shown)


core/sdk-core/src/main/java/software/amazon/awssdk/core/traits/Trait.java
```diff
@@ -16,6 +16,7 @@
 package software.amazon.awssdk.core.traits;
 
 import software.amazon.awssdk.annotations.SdkProtectedApi;
+import software.amazon.awssdk.core.SdkField;
```
L-Applin, line 19: Why are we adding that import?

core/sdk-core/src/main/java/software/amazon/awssdk/core/SdkNumber.java
```diff
@@ -27,20 +27,23 @@
...
- * @see #fromBigDecimal(BigDecimal) to create from a BigDecimal.
- * @see #fromBigInteger(BigInteger) to create from a BigInteger.
- * @see #fromDouble(double) to create from a double
- * @see #fromFloat(float) to create from a float.
- * @see #fromLong(long) to create from a long.
- * @see #fromShort(short) to create from a short.
- * @see #fromInteger(int) to create from an integer.
- * @see #fromString(String)  to create from a Stringl,
- * Thus by doing this, this class is able to preserve arbitrary precison of any given number.
+ * <ul>
```
L-Applin, line 30: What was the problem with using `@see`?

sugmanue: `@see` tags are to generate `See Also` sections generally linking to documents or API outside the current one, not links in general, that's what `{@link …} is designed for usually below the main document, as in 

---
Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.

See Also [Lorem ipsum wikipedia entry](https://en.wikipedia.org/wiki/Lorem_ipsum)

---

Using `@see` here also produces a warning about `warning: Tag {@link}: nested link. If {@link SdkNumber} is expected in a particular number format then its corresponding getter methods can be used.` given that there's leading content below the **See Also** section of the docs.

core/sdk-core/src/main/java/software/amazon/awssdk/core/traits/Trait.java
sugmanue: `SdkField` is being used as a link in the javadocs. Without the import javadoc does not know how to resolve the type and generate the link.

core/sdk-core/src/main/java/software/amazon/awssdk/core/SdkNumber.java
sugmanue: See also the way it is currently rendered [here](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/core/SdkNumber.html).

L-Applin (approved):

Issue #1690, opened 2020-03-06
Overriding default SdkClientOption.SCHEDULED_EXECUTOR_SERVICE in AWS SDK for Java 2


hariharanv01:
Is there a way I can override the default `SdkClientOption.SCHEDULED_EXECUTOR_SERVICE` setting in AWS SDK for Java 2? 

I want to get rid of all the threadpools that are opaque to the application, and ensure that the threadpools are reused across all non-blocking contexts.

The current behavior in SDK is as below (5 being a magic number)

```java
private ScheduledExecutorService resolveScheduledExecutorService() {
        return Executors.newScheduledThreadPool(5, new ThreadFactoryBuilder()
            .threadNamePrefix("sdk-ScheduledExecutor").build());
}
```

debora-ito:
Seems a reasonable ask, marking as a feature request.

Marcus-Rosti:
I'll second that. When using cached, ephemeral clients, I can see that the scheduled thread pool will at times be leaked when the aws client is evicted by the cache

jreznot:
Especially needed for Spring WebFlux and Project Reactor to reuse boundedElastic thread pool instead of spawning separate threads.

RoynniZambrano:
Hello there, has anyone found a workaround for this?
It would be great to be able to configure it in the same way that the **SdkAdvancedAsyncClientOption.FUTURE_COMPLETION_EXECUTOR** is configured.

samarprakash:
Do you have any updates on this issue? This becomes worse if an application has multiple AWS services in use. For example - if app A uses S3/SQS/SNS/DDB, all clients internally create a separate ScheduledExecutorService with a default core thread pool of 5 and obviously with no allowCoreThreadTimeout and its unnecessary waste of system resources. AWS SDK must provide an option to the client to choose the right behaviour for their application, similar to AsyncConfiguration.

debora-ito:
Feature released in the SDK version `2.20.70` via #4002.

PR #4002, opened 2023-05-12
Add client configuration overriding of SCHEDULED_EXECUTOR_SERVICE option

scrocquesel:
## Motivation and Context

Customer should be able to use their own ThreadPool and reuse it accros multiple clients.

https://github.com/aws/aws-sdk-java-v2/issues/1690

## Modifications
Introduce new client config methods to allow to override `SCHEDULED_EXECUTOR_SERVICE` option. It is optional and when not provided will default to the actual code that create a new ThreadPool for each clients

## Testing
Unit tests ensure the override is applied on the client configuration.
I also tested in a private project by setting api call timeout to actually create new thread on the custom thread pool.

## Types of changes

## Checklist

## License

### DIFF
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/client/builder/SdkDefaultClientBuilder.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/client/builder/SdkDefaultClientBuilder.java
@@ -91,6 +91,7 @@ import software.amazon.awssdk.profiles.ProfileFileSupplier;
 import software.amazon.awssdk.profiles.ProfileFileSystemSetting;
 import software.amazon.awssdk.utils.AttributeMap;
 import software.amazon.awssdk.utils.Either;
+import software.amazon.awssdk.utils.ScheduledExecutorUtils;
 import software.amazon.awssdk.utils.ThreadFactoryBuilder;
 import software.amazon.awssdk.utils.Validate;
 
@@ -417,9 +418,10 @@ public abstract class SdkDefaultClientBuilder<B extends SdkClientBuilder<B, C>, 
                 .threadNamePrefix("sdk-ScheduledExecutor").build());
 
             return executor;
-        };            
+        };
 
         return Optional.ofNullable(config.option(SCHEDULED_EXECUTOR_SERVICE))
+            .map(ScheduledExecutorUtils::unmanagedScheduledExecutor)
             .orElseGet(defaultScheduledExecutor);
     }
 
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/client/builder/SdkDefaultClientBuilder.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/client/builder/SdkDefaultClientBuilder.java
@@ -222,6 +222,7 @@ public abstract class SdkDefaultClientBuilder<B extends SdkClientBuilder<B, C>, 
 
         SdkClientConfiguration.Builder builder = configuration.toBuilder();
 
+        builder.option(SCHEDULED_EXECUTOR_SERVICE, clientOverrideConfiguration.scheduledExecutorService().orElse(null));
         builder.option(EXECUTION_INTERCEPTORS, clientOverrideConfiguration.executionInterceptors());
         builder.option(RETRY_POLICY, clientOverrideConfiguration.retryPolicy().orElse(null));
         builder.option(ADDITIONAL_HTTP_HEADERS, clientOverrideConfiguration.headers());
@@ -313,7 +314,7 @@ public abstract class SdkDefaultClientBuilder<B extends SdkClientBuilder<B, C>, 
     private SdkClientConfiguration finalizeConfiguration(SdkClientConfiguration config) {
         RetryPolicy retryPolicy = resolveRetryPolicy(config);
         return config.toBuilder()
-                     .option(SCHEDULED_EXECUTOR_SERVICE, resolveScheduledExecutorService())
+                     .option(SCHEDULED_EXECUTOR_SERVICE, resolveScheduledExecutorService(config))
                      .option(EXECUTION_INTERCEPTORS, resolveExecutionInterceptors(config))
                      .option(RETRY_POLICY, retryPolicy)
                      .option(CLIENT_USER_AGENT, resolveClientUserAgent(config, retryPolicy))
@@ -410,9 +411,16 @@ public abstract class SdkDefaultClientBuilder<B extends SdkClientBuilder<B, C>, 
      * Finalize the internal SDK scheduled executor service that is used for scheduling tasks such
      * as async retry attempts and timeout task.
      */
-    private ScheduledExecutorService resolveScheduledExecutorService() {
-        return Executors.newScheduledThreadPool(5, new ThreadFactoryBuilder()
-            .threadNamePrefix("sdk-ScheduledExecutor").build());
+    private ScheduledExecutorService resolveScheduledExecutorService(SdkClientConfiguration config) {
+        Supplier<ScheduledExecutorService> defaultScheduledExecutor = () -> {
+            ScheduledExecutorService executor = Executors.newScheduledThreadPool(5, new ThreadFactoryBuilder()
+                .threadNamePrefix("sdk-ScheduledExecutor").build());
+
+            return executor;
+        };            
+
+        return Optional.ofNullable(config.option(SCHEDULED_EXECUTOR_SERVICE))
+            .orElseGet(defaultScheduledExecutor);
     }
 
     /**
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/client/config/ClientOverrideConfiguration.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/client/config/ClientOverrideConfiguration.java
@@ -358,6 +358,10 @@ public final class ClientOverrideConfiguration
          * Configure the scheduled executor service that should be used for scheduling tasks such as async retry attempts
          * and timeout task.
          *
+         * <p>
+         * <b>The SDK will not automatically close the executor when the client is closed. It is the responsibility of the
+         * user to manually close the executor once all clients utilizing it have been closed.</b>
+         *
          * @see ClientOverrideConfiguration#scheduledExecutorService()
          */
         Builder scheduledExecutorService(ScheduledExecutorService scheduledExecutorService);
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/client/config/ClientOverrideConfiguration.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/client/config/ClientOverrideConfiguration.java
@@ -23,6 +23,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.TreeMap;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.function.Consumer;
 import software.amazon.awssdk.annotations.SdkPublicApi;
 import software.amazon.awssdk.annotations.ToBuilderIgnoreField;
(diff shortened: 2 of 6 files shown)


core/sdk-core/src/test/java/software/amazon/awssdk/core/client/builder/DefaultClientBuilderTest.java
```diff
@@ -166,6 +171,7 @@ public void overrideConfigurationReturnsSetValues() {
         assertThat(builderOverrideConfig.metricPublishers()).isEqualTo(metricPublishers);
         assertThat(builderOverrideConfig.executionAttributes().getAttributes()).isEqualTo(executionAttributes.getAttributes());
         assertThat(builderOverrideConfig.advancedOption(ENDPOINT_OVERRIDDEN_OVERRIDE)).isEqualTo(Optional.of(Boolean.TRUE));
+        assertThat(builderOverrideConfig.scheduledExecutorService()).isEqualTo(scheduledExecutorService);
```
davidh44, line 174: We need to call `get()` here since `scheduledExecutorService()` returns `Optional`

core/sdk-core/src/main/java/software/amazon/awssdk/core/client/builder/SdkDefaultClientBuilder.java
```diff
@@ -410,9 +411,16 @@ private Executor resolveAsyncFutureCompletionExecutor(SdkClientConfiguration con
...
      */
-    private ScheduledExecutorService resolveScheduledExecutorService() {
-        return Executors.newScheduledThreadPool(5, new ThreadFactoryBuilder()
-            .threadNamePrefix("sdk-ScheduledExecutor").build());
+    private ScheduledExecutorService resolveScheduledExecutorService(SdkClientConfiguration config) {
+        Supplier<ScheduledExecutorService> defaultScheduledExecutor = () -> {
+            ScheduledExecutorService executor = Executors.newScheduledThreadPool(5, new ThreadFactoryBuilder()
+                .threadNamePrefix("sdk-ScheduledExecutor").build());
+
+            return executor;
```
davidh44, line 419: We need to make sure the custom ScheduledExecutorService doesn't get closed when the client is closed. We can create a `NonManagedScheduledExecutorService` wrapper similar to [`NonManagedSdkHttpClient`](https://github.com/aws/aws-sdk-java-v2/blob/master/core/sdk-core/src/main/java/software/amazon/awssdk/core/client/builder/SdkDefaultClientBuilder.java#L526)/`NonManagedSdkAsyncHttpClient`. Currently, the executor is closed [here](https://github.com/aws/aws-sdk-java-v2/blob/master/utils/src/main/java/software/amazon/awssdk/utils/AttributeMap.java#L94) when the client is closed. We can override `shutdown()` to do nothing:

```
    /**
     * Wrapper around {@link ScheduledExecutorService} to prevent it from being closed. Used when the customer provides
     * a custom scheduled executor service in which case they are responsible for the lifecycle of it.
     */
    @SdkTestInternalApi
    public static final class NonManagedScheduledExecutorService implements ScheduledExecutorService {

        private final ScheduledExecutorService delegate;

        NonManagedScheduledExecutorService(ScheduledExecutorService delegate) {
            this.delegate = paramNotNull(delegate, "ScheduledExecutorService");
        }

        public ScheduledExecutorService scheduledExecutorService() {
            return delegate;
        }

        @Override
        public void shutdown() {
            // Do nothing, this executor service is managed by the customer.
        }

        @Override
        public List<Runnable> shutdownNow() {
            return new ArrayList<>();
        }

        @Override
        public boolean isShutdown() {
            return delegate.isShutdown();
        }

        @Override
        public boolean awaitTermination(long timeout, TimeUnit unit) throws InterruptedException {
            return delegate.awaitTermination(timeout, unit);
        }

       // override remaining methods with delegate.method()...
```

```
    private ScheduledExecutorService resolveScheduledExecutorService(SdkClientConfiguration config) {
        ScheduledExecutorService customScheduledExecutorService = config.option(SCHEDULED_EXECUTOR_SERVICE);
        if (customScheduledExecutorService != null) {
            return new NonManagedScheduledExecutorService(customScheduledExecutorService);
        }

        return Executors.newScheduledThreadPool(5, new ThreadFactoryBuilder()
            .threadNamePrefix("sdk-ScheduledExecutor").build());
    }
```

Let's also add a test for this in [`ResourceManagementTest`](https://github.com/aws/aws-sdk-java-v2/blob/master/test/protocol-tests/src/test/java/software/amazon/awssdk/protocol/tests/ResourceManagementTest.java):

```
    @Test
    public void scheduledExecutorFromBuilderNotShutdown() {
        ScheduledExecutorService scheduledExecutorService = mock(ScheduledExecutorService.class);

        asyncClientBuilder().overrideConfiguration(c -> c.scheduledExecutorService(scheduledExecutorService)).build().close();

        verify(scheduledExecutorService, never()).shutdown();
        verify(scheduledExecutorService, never()).shutdownNow();
    }
```

core/sdk-core/src/test/java/software/amazon/awssdk/core/client/builder/DefaultClientBuilderTest.java
```diff
@@ -267,6 +276,7 @@ public void close() {
         assertThat(config.option(METRIC_PUBLISHERS)).contains(metricPublisher);
         assertThat(config.option(EXECUTION_ATTRIBUTES).getAttribute(execAttribute)).isEqualTo("value");
         assertThat(config.option(ENDPOINT_OVERRIDDEN)).isEqualTo(Boolean.TRUE);
+        assertThat(config.option(SCHEDULED_EXECUTOR_SERVICE)).isEqualTo(scheduledExecutorService);
```
davidh44, line 279: Now that we are wrapping with `NonManagedScheduledExecutorService`, we need to retrieve the custom `ScheduledExecutorService`

```
import static software.amazon.awssdk.core.client.builder.SdkDefaultClientBuilder.NonManagedScheduledExecutorService;

        NonManagedScheduledExecutorService customScheduledExecutorService =
            (NonManagedScheduledExecutorService) config.option(SCHEDULED_EXECUTOR_SERVICE);
        assertThat(customScheduledExecutorService.scheduledExecutorService()).isEqualTo(scheduledExecutorService);
```

core/sdk-core/src/main/java/software/amazon/awssdk/core/client/config/ClientOverrideConfiguration.java
```diff
@@ -141,6 +145,14 @@ public List<ExecutionInterceptor> executionInterceptors() {
         return executionInterceptors;
     }
 
+    /**
+     * The optional scheduled executor service that should be used for scheduling tasks such as async retry attempts
+     * and timeout task.
```
davidh44, line 150: Can we add javadocs to remind users that the scheduled executor service will not be closed when the client is closed, and that they are responsible for closing it?

davidh44 (commented):
Thanks for the PR!! Looks great, we just need to make sure the custom ScheduledExecutorService doesn't get closed when the client is closed. Also a few minor comments.

core/sdk-core/src/main/java/software/amazon/awssdk/core/client/builder/SdkDefaultClientBuilder.java
scrocquesel: Would you prefer to create a utility class like https://github.com/aws/aws-sdk-java-v2/blob/master/utils/src/main/java/software/amazon/awssdk/utils/ExecutorUtils.java ? `SdkDefaultClientBuilder` is already big and there is a lot of methods to implements in `ScheduledExecutorService`.

Also, the future completion executor is wrapped in the configuration builder class rather in the client builder, should we keep consistency and wrap the `scheduledExecutorService` in the `ClientOverrideConfiguration:Builder` class ?

scrocquesel:
> Thanks for the PR!! Looks great, we just need to make sure the custom ScheduledExecutorService doesn't get closed when the client is closed. Also a few minor comments.

Should I add more commits to the PR or force pushed ?

core/sdk-core/src/main/java/software/amazon/awssdk/core/client/builder/SdkDefaultClientBuilder.java
davidh44: Yeah that sounds good, maybe `ScheduledExecutorUtils`? 

Makes sense, let's put it in `ClientOverrideConfiguration:Builder` instead.

davidh44:
> 

You can add more commits to the PR

core/sdk-core/src/main/java/software/amazon/awssdk/core/client/builder/SdkDefaultClientBuilder.java
scrocquesel: Make sense, TBH, there is a lot of piece of code everywhere that create thread pool and its not really easy to follow and find out where to override it when possible.
I'll keep it there and maybe raise an issue to refactor the rest ?

davidh44: Agreed, yup sounds good to me

core/sdk-core/src/main/java/software/amazon/awssdk/core/client/builder/SdkDefaultClientBuilder.java
```diff
@@ -222,6 +223,7 @@ private SdkClientConfiguration setOverrides(SdkClientConfiguration configuration
 
         SdkClientConfiguration.Builder builder = configuration.toBuilder();
 
+        builder.option(SCHEDULED_EXECUTOR_SERVICE, clientOverrideConfiguration.scheduledExecutorService().orElse(null));
```
zoewangg, line 226: Do we need to set it here since it's going to be resolved and overriden in `finalizeConfiguration`?

core/sdk-core/src/main/java/software/amazon/awssdk/core/client/config/ClientOverrideConfiguration.java
```diff
@@ -141,6 +145,17 @@ public List<ExecutionInterceptor> executionInterceptors() {
...
+     * <p>
+     * <b>The SDK will not automatically close the executor when the client is closed. It is the responsibility of the
```
zoewangg, line 152: Can we add this note to the builder setter as well since that's where most users will check?

core/sdk-core/src/main/java/software/amazon/awssdk/core/client/builder/SdkDefaultClientBuilder.java
scrocquesel: Seems to me the pattern is to push every params in the attributes map of `SdkClientConfiguration` and finalize with only this attributes map as input/output. So `clientOverrideConfiguration` will never be read after this. Thus, it is needed.

core/sdk-core/src/main/java/software/amazon/awssdk/core/client/config/ClientOverrideConfiguration.java
scrocquesel: done

core/sdk-core/src/main/java/software/amazon/awssdk/core/client/builder/SdkDefaultClientBuilder.java
zoewangg: Ah yeah, I read it wrong.

zoewangg (approved):

davidh44 (approved):

davidh44:
[@all-contributors](https://github.com/all-contributors) please add [@](https://github.com/martinKindall)[scrocquesel ](https://github.com/scrocquesel)for code.

davidh44:
@scrocquesel Great stuff! PR has been merged 👍

debora-ito:
Trying again :) And adding contributors from previous PRs.

@all-contributors 

please add @scrocquesel for code.
please add @dave-fn for code.
please add @L-Applin for code.

PR #4008, opened 2023-05-15
Adds S3CrossRegion clients and codegen for retrieving bucket name

cenedhryn:
## Motivation and Context
Adds a new cross region client that extends DelegatingS3Client and can decorate the s3 client. It needs the value of the request bucket parameter, so this PR provides that value. 

## Modifications
Added codegen for retrieving bucket parameter from any S3 API that supports it.

## Testing
`S3BucketRequestParams` generates to 

~~~
@SdkInternalApi
public final class S3BucketRequestParams {
    public static Optional<String> bucketName(SdkRequest request) {
        if (request instanceof AbortMultipartUploadRequest) {
            return Optional.of(((AbortMultipartUploadRequest) request).bucket());
        }
        if (request instanceof CompleteMultipartUploadRequest) {
            return Optional.of(((CompleteMultipartUploadRequest) request).bucket());
        }
        if (request instanceof CopyObjectRequest) {
            return Optional.of(((CopyObjectRequest) request).destinationBucket());
        }
        ...
        return Optional.empty();
}
~~~

### DIFF
--- a/core/aws-core/src/main/java/software/amazon/awssdk/awscore/AwsServiceClientConfiguration.java
+++ b/core/aws-core/src/main/java/software/amazon/awssdk/awscore/AwsServiceClientConfiguration.java
@@ -77,15 +77,6 @@ public abstract class AwsServiceClientConfiguration extends SdkServiceClientConf
         Builder region(Region region);
 
         @Override
-        Builder overrideConfiguration(ClientOverrideConfiguration clientOverrideConfiguration);
-
-        @Override
-        Builder endpointOverride(URI endpointOverride);
-
-        @Override
-        Builder endpointProvider(EndpointProvider endpointProvider);
-
-        @Override
         AwsServiceClientConfiguration build();
     }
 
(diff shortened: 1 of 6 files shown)


zoewangg:
Can we use `request.getValueForField("bucket", String.class)`?

cenedhryn (commented):
> Can we use request.getValueForField("bucket", String.class)?
I was under the impression that this construct doesn't handle all API variants successfully and needs additional cases anyway and therefore the SIM ticket had another suggestions, but I can take a deeper look and see what the tradeoff is.

core/aws-core/src/main/java/software/amazon/awssdk/awscore/AwsServiceClientConfiguration.java
```diff
@@ -76,15 +76,6 @@ public interface Builder extends SdkServiceClientConfiguration.Builder {
          */
         Builder region(Region region);
 
-        @Override
-        Builder overrideConfiguration(ClientOverrideConfiguration clientOverrideConfiguration);
```
zoewangg, line 80: Why are we removing these?

zoewangg (approved):

core/aws-core/src/main/java/software/amazon/awssdk/awscore/AwsServiceClientConfiguration.java
millems: Don't they refine the return type?

millems: It would be backwards-incompatible (and imo, incorrect) to remove them. They have a different return type.

zoewangg: Yeah, talked with @cenedhryn offline and there were added back

PR #4009, opened 2023-05-15
Fix leaked futures when executor rejects Runnable

dagnir:
## Motivation and Context
This commit fixes  an issue where CompletableFutures are leaked/never completed when the submission to the FUTURE_COMPLETE_EXECUTOR is rejected.

By default, the SDK uses `2 * number of cores` (with a maximum of 64), and uses bounded queue of size 1000. In cases where the throughput to the client exceeds the executor's ability to keep up, it would reject executions, and previously, this would lead to leaked futures.

## Testing
 - Unit tests.
 - Integ tests.

## Types of changes

## Checklist

## License

### DIFF
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/MakeAsyncHttpRequestStage.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/MakeAsyncHttpRequestStage.java
@@ -156,15 +156,26 @@ public final class MakeAsyncHttpRequestStage<OutputT>
             }
         });
 
-        // Offload the completion of the future returned from this stage onto
-        // the future completion executor
-        responseHandlerFuture.whenCompleteAsync((r, t) -> {
-            if (t == null) {
-                responseFuture.complete(r);
-            } else {
-                responseFuture.completeExceptionally(t);
+        // Attempt to offload the completion of the future returned from this
+        // stage onto the future completion executor
+        CompletableFuture<Response<OutputT>> asyncComplete =
+            responseHandlerFuture.whenCompleteAsync((r, t) -> completeResponseFuture(responseFuture, r, t),
+                                                    futureCompletionExecutor);
+
+        // It's possible the async execution above fails. If so, log a warning,
+        // and just complete it synchronously.
+        asyncComplete.whenComplete((ignored, asyncCompleteError) -> {
+            if (asyncCompleteError != null) {
+                log.debug(() -> String.format("Could not complete the service call future on the provided "
+                                              + "FUTURE_COMPLETION_EXECUTOR. The future will be completed synchronously by thread"
+                                              + " %s. This may be an indication that the executor is being overwhelmed by too"
+                                              + " many requests, and it may negatively affect performance. Consider changing "
+                                              + "the configuration of the executor to accommodate the load through the client.",
+                                  Thread.currentThread().getName()),
+                          asyncCompleteError);
+                responseHandlerFuture.whenComplete((r, t) -> completeResponseFuture(responseFuture, r, t));
             }
-        }, futureCompletionExecutor);
+        });
 
         return responseFuture;
     }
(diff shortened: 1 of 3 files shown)


cenedhryn (approved):

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/MakeAsyncHttpRequestStage.java
```diff
@@ -156,15 +156,26 @@ private CompletableFuture<Response<OutputT>> executeHttpRequest(SdkHttpFullReque
...
+        // stage onto the future completion executor
+        CompletableFuture<Response<OutputT>> asyncComplete =
+            responseHandlerFuture.whenCompleteAsync((r, t) -> completeResponseFuture(responseFuture, r, t),
+                                                    futureCompletionExecutor);
+
+        // It's possible the async execution above fails. If so, log a warning,
+        // and just complete it synchronously.
+        asyncComplete.whenComplete((ignored, asyncCompleteError) -> {
+            if (asyncCompleteError != null) {
+                log.debug(() -> String.format("Could not complete the service call future on the provided "
```
zoewangg, line 169: Should it be warn?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/MakeAsyncHttpRequestStage.java
```diff
@@ -156,15 +156,26 @@ private CompletableFuture<Response<OutputT>> executeHttpRequest(SdkHttpFullReque
...
+                                              + "FUTURE_COMPLETION_EXECUTOR. The future will be completed synchronously by thread"
+                                              + " %s. This may be an indication that the executor is being overwhelmed by too"
+                                              + " many requests, and it may negatively affect performance. Consider changing "
+                                              + "the configuration of the executor to accommodate the load through the client.",
+                                  Thread.currentThread().getName()),
+                          asyncCompleteError);
+                responseHandlerFuture.whenComplete((r, t) -> completeResponseFuture(responseFuture, r, t));
```
zoewangg, line 176: Would this potentially block the event loop threads? Should we consider failing the request?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/MakeAsyncHttpRequestStage.java
dagnir: I thought about it, but if they have enough throughput to overwhelm the executor it could pollute the logs. I'm fine with debug for now since execution isn't failing.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/MakeAsyncHttpRequestStage.java
dagnir: It will be the thread that called `whenCompleteAsync`.

I don't like the idea of failing the request because that seems like a waste. At this point the response handler already completed so we have a result that we would just throw away.

zoewangg: My concern is that this may not fix the issue but just delay it. It'd likely be manifested as connection acquire timeout errors, which would be a lot harder to troubleshoot, but I see your point. Playing devil's advocate, I wonder if we should change the queue to be unbounded.

dagnir: > Playing devil's advocate, I wonder if we should change the queue to be unbounded. 

This may be fine too, but could lead to increased memory consumption. Since this is a publicly accessible option, we need to guard against it in any case.

zoewangg: Yeah, it's like pick your poison type of thing. I'd vote for whichever is easier to troubleshoot in the failure case.

zoewangg (approved):

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/MakeAsyncHttpRequestStage.java
dagnir: Closing the loop: discussed with @millems as well, we're all in agreement that completing on the I/O thread is okay, especially since completing it exceptionally would also have to happen on the I/O thread.

PR #4018, opened 2023-05-16
HttpSigner interface changes

gosar:
## Modifications

* Make Identity explicit in HttpSignRequest
* Add consumer builder pattern to HttpSigner methods
* Make explicit sub types for sync and async to make it easier to deal with generics, especially with Publisher<ByteBuffer>. Removes need to use Class<T> directly in methods.

Other fixes:
* Remove <T> for HttpAuthOption.Builder.schemeId
* Fix generics related warning in Signer interfaces

Usage is shown in HttpSignerTest included in this PR.

## Types of changes

## Checklist

## License


core/http-auth-spi/src/main/java/software/amazon/awssdk/http/auth/spi/HttpSigner.java
```diff
@@ -22,24 +22,24 @@
 
 /**
  * Interface for the process of modifying a request destined for a service so that the service can authenticate the SDK
- * customer’s identity
+ * customer’s identity.
  */
 @SdkPublicApi
-public interface HttpSigner {
+public interface HttpSigner<IdentityT> {
```
gosar, line 28: in either option, need to update this to `IdentityT extends Identity`

core/http-auth-spi/src/test/java/software/amazon/awssdk/http/auth/spi/HttpSignerTest.java
```diff
@@ -0,0 +1,113 @@
...
+                                                                                        .putProperty(KEY, VALUE));
+        assertNotNull(signedRequest);
+    }
+
+    // @Test
+    public void signAsync_usingRequest_works() {
+        Publisher<ByteBuffer> payload = subscriber -> {};
+        // Don't know how to make this work....
+        // HttpSignRequest.Builder<Publisher<ByteBuffer>, TokenIdentity> builder =
+        //   HttpSignRequest.builder(Publisher.class, TokenIdentity.class);
```
gosar, line 71: I'm unsure what would this have looked like without the Consumer builder pattern? This would have been a question without making these changes to pull Identity out explicitly.

core/http-auth-spi/src/main/java/software/amazon/awssdk/http/auth/spi/AsyncHttpSignRequest.java
```diff
@@ -0,0 +1,59 @@
...
+import software.amazon.awssdk.annotations.ThreadSafe;
+import software.amazon.awssdk.http.SdkHttpRequest;
+import software.amazon.awssdk.http.auth.spi.internal.DefaultAsyncHttpSignRequest;
+import software.amazon.awssdk.identity.spi.Identity;
+import software.amazon.awssdk.utils.builder.SdkBuilder;
+
+@SdkPublicApi
+@Immutable
+@ThreadSafe
+public interface AsyncHttpSignRequest<IdentityT extends Identity> extends HttpSignRequest<Publisher<ByteBuffer>, IdentityT> {
```
millems, line 31: Nit: Javadoc everywhere

core/http-auth-spi/src/main/java/software/amazon/awssdk/http/auth/spi/AsyncHttpSignRequest.java
```diff
@@ -0,0 +1,59 @@
...
+     */
+    static <IdentityT extends Identity> Builder<IdentityT> builder(Class<IdentityT> ignoredIdentityType) {
+        return new DefaultAsyncHttpSignRequest.BuilderImpl<>();
+    }
+
+    @Override
+    default Class<Publisher<ByteBuffer>> payloadType() {
+        // TODO: Code Review Note: Note this cast
+        return (Class) Publisher.class;
+    }
```
millems, line 43: Do we still need this getter? We could split SignedHttpRequest like we split HttpSignRequest into sync/async so that we don't need this anymore.

core/http-auth-spi/src/main/java/software/amazon/awssdk/http/auth/spi/HttpSignRequest.java
```diff
@@ -20,25 +20,21 @@
...
 @ThreadSafe
-public interface HttpSignRequest<PayloadT> {
-
-    /**
-     * Get a new builder for creating a {@link HttpSignRequest}.
-     */
-    static <T> Builder<T> builder(Class<T> payloadType) {
-        return new DefaultHttpSignRequest.BuilderImpl(payloadType);
-    }
+public interface HttpSignRequest<PayloadT, IdentityT extends Identity> {
```
millems, line 37: Imo, this should implement `ToCopyableBuilder` and its builder should implement `CopyableBuilder`. That prevents some issues here, needing to remember to override every parent method in the child class.

Unfortunately it makes the parameters on this type even crazier.

To make the room, we could change `IdentityT identity()` to `Identity identity()` and `Optional<PayloadT> payload()` to `Optional<? extends Object> payload()`, and then override them in the child types to return the exact types. Even better, we might just remove the getters from here entirely and rely on the subtype. I don't expect it to be a major problem to need to use the subtype to get access to the payload or identity.

We'd also need to remove the `payload` and `identity` setters from this base builder, but that's probably okay - we can just put them in the child types. It should be very rare that someone needs to set the identity or payload using the base builder type...

core/http-auth-spi/src/main/java/software/amazon/awssdk/http/auth/spi/SignedHttpRequest.java
```diff
@@ -58,16 +58,16 @@ static <T> Builder<T> builder(Class<T> payloadType) {
     /**
      * A builder for a {@link SignedHttpRequest}.
      */
-    interface Builder<PayloadT> extends SdkBuilder<Builder<PayloadT>, SignedHttpRequest> {
+    interface Builder<PayloadT> extends SdkBuilder<Builder<PayloadT>, SignedHttpRequest<PayloadT>> {
```
millems, line 61: Same comment here. `CopyableBuilder`?

core/http-auth-spi/src/main/java/software/amazon/awssdk/http/auth/spi/internal/DefaultHttpSignRequest.java
```diff
@@ -22,85 +22,78 @@
...
+    protected final IdentityT identity;
+    protected final Map<SignerProperty<?>, Object> properties;
 
-    DefaultHttpSignRequest(BuilderImpl<PayloadT>  builder) {
-        this.payloadType = Validate.paramNotNull(builder.payloadType, "payloadType");
+    protected DefaultHttpSignRequest(BuilderImpl<?, PayloadT, IdentityT> builder) {
         this.request = Validate.paramNotNull(builder.request, "request");
         this.payload = builder.payload;
+        this.identity = Validate.paramNotNull(builder.identity, "identity");
         this.properties = new HashMap<>(builder.properties);
```
millems, line 40: `unmodifiableMap`, since you're making it protected?

core/http-auth-spi/src/main/java/software/amazon/awssdk/http/auth/spi/AsyncHttpSignRequest.java
gosar: Yeah, tradeoff would be Signer implementations where the logic for sync/async isn't different, will have some code duplication, see HttpSignerTest in this PR. Probably applies to bearer, apiKey, httpBasic/Digest. So I thought this one cast was an okay tradeoff. LMK if you prefer otherwise. Or if you think, if we are splitting the input, the output should also be split anyways, or split neither?

The alternate, without splitting either would be to remove the `Class<T>` payload parameter in the builder() method (and the `payloadType()` accessor). The drawback being that someone using `HttpSignRequest.builder()` would be wondering why they are getting `Object` for `payload()`. I think this might be the better tradeoff, as the compiler helps the customer there. The use of these types is for the sign methods and they expect specific types, so it will be easy enough to understand that `<Publisher<ByteBuffer>>` needed to be added. Moreover, this is ony needed if using the builder directly. We can add the Consumer Builder methods still to alleviate the need to instantiate the builder directly.

core/http-auth-spi/src/main/java/software/amazon/awssdk/http/auth/spi/HttpSignRequest.java
gosar: > That prevents some issues here, like needing to remember to override every parent method in the child class.

Without using To/CopyableBuilder, I was able to remove the overriding of methods in the builder sub-interfaces to return the sub-type by adding a generic type parameter for the builder, in these commits https://github.com/aws/aws-sdk-java-v2/pull/4018/commits/de704e738dbb15530a83ee86f95c9606ccd3cb39 and https://github.com/aws/aws-sdk-java-v2/pull/4018/commits/3e24cc490fcc665f60e9a735e7d2d4a8dcff6ba9.

I'm not sure if we'll really need to mutate these objects, so I think we can defer making these To/CopyableBuilder until needed?

core/http-auth-spi/src/main/java/software/amazon/awssdk/http/auth/spi/HttpSignRequest.java
```diff
@@ -56,28 +47,38 @@ static <T> Builder<T> builder(Class<T> payloadType) {
     Optional<PayloadT> payload();
 
     /**
-     * Returns the property that the {@link HttpSigner} can use during signing.
+     * Returns the identity.
+     */
+    IdentityT identity();
```
millems, line 52: Imo, I'd rather not parameterize this types and just do `Optional<?> payload()` and `Identity identity()` here. Subtypes would override them with the more-specific return type.

This makes it so that code that deal with the base sign request can just deal with a `SignRequest` instead of a `SignRequest<?, ?>`, and the base builder can just deal with a `Builder<?> instead of a Builder<?, ?, ?>`.

WDYT?

core/http-auth-spi/src/main/java/software/amazon/awssdk/http/auth/spi/HttpSignRequest.java
```diff
@@ -56,28 +47,38 @@ static <T> Builder<T> builder(Class<T> payloadType) {
...
         /**
          * Set the body payload of the request. A payload is optional. By default, the payload will be empty.
          */
-        Builder payload(PayloadT payload);
+        B payload(PayloadT payload);
+
+        /**
+         * Set the identity of the request.
+         */
+        B identity(IdentityT identity);
```
millems, line 77: If we go with the suggestion I have above, we'd remove these from the base builder and put it in the subtypes.

core/http-auth-spi/src/main/java/software/amazon/awssdk/http/auth/spi/HttpSigner.java
```diff
@@ -15,31 +15,62 @@
...
-public interface HttpSigner {
+public interface HttpSigner<IdentityT extends Identity> {
+
+    /**
+     * Method that takes in inputs to sign a request with sync payload and returns a signed version of the request.
+     *
+     * @param request The inputs to sign a request.
+     * @return A signed version of the request.
+     */
+    SyncSignedHttpRequest sign(SyncHttpSignRequest<IdentityT> request);
```
millems, line 39: `? extends IdentityT`? (to allow someone passing a `SyncHttpSignRequest<AwsSessionIdentity>` to a `HttpSigner<AwsIdentity>`)

gosar: Oh that's a bad miss! Thanks.

core/http-auth-spi/src/main/java/software/amazon/awssdk/http/auth/spi/HttpSignRequest.java
gosar: I think code that deals with base sign request, might benefit from the type, atleast the IdentityT. So the BearerTokenSigner can extract the token from `HttpSignRequest<?, TokenIdentity identity>`. Though I can see removing the PayloadT (will reduce it by one), but means the subtypes have to (re)define the `payload()` method instead of specifying the generic in type definition. And as an interface, having a HttpSignRequest that has a `payload()` accessor but doesn't know it's type doesn't feel great to me, so I'm in favor of keeping them.

core/http-auth-spi/src/main/java/software/amazon/awssdk/http/auth/spi/HttpSigner.java
```diff
@@ -16,30 +16,65 @@
...
      *
-     * @param request The request to sign, with async payload
-     * @return A signed version of the input request
+     * @param consumer A {@link Consumer} to which an empty {@link HttpSignRequest.Builder} will be given.
+     * @return A signed version of the request.
      */
-    SignedHttpRequest<Publisher<ByteBuffer>> signAsync(HttpSignRequest<? extends Publisher<? extends ByteBuffer>> request);
+    default SignedHttpRequest<Publisher<ByteBuffer>> signAsync(
+        Consumer<HttpSignRequest.Builder<Publisher<ByteBuffer>, IdentityT>> consumer) {
+        return signAsync(new DefaultHttpSignRequest.BuilderImpl<Publisher<ByteBuffer>, IdentityT>((Class)Publisher.class)
```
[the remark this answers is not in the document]
gosar: Fixed by moving to separate sub types for sync and async.

core/http-auth-spi/src/test/java/software/amazon/awssdk/http/auth/spi/HttpSignerTest.java
gosar: Resolved by moving to separate sub types for sync and async.

core/http-auth-spi/src/main/java/software/amazon/awssdk/http/auth/spi/AsyncHttpSignRequest.java
gosar: Resolved by moving to separate sub types for sync and async for SignedHttpRequest too and removing the Class<T> when creating the builder as well as returning it.

core/http-auth-spi/src/main/java/software/amazon/awssdk/http/auth/spi/HttpSignRequest.java
```diff
@@ -17,33 +17,21 @@
...
 /**
- * Represents a request to be signed by {@link HttpSigner}.
+ * Input parameters to sign a request using {@link HttpSigner}.
  *
- * @param <PayloadT> The type of payload of this request.
+ * @param <PayloadT> The type of payload of the request.
+ * @param <IdentityT> The type of the identity.
  */
-@SdkPublicApi
+@SdkProtectedApi
```
millems, line 31: This is still a public api.

core/http-auth-spi/src/main/java/software/amazon/awssdk/http/auth/spi/HttpSignRequest.java
millems: Cool. And we can reassess at the surface area review as well, once we can see the impact on the signers, etc.

millems (approved):
A minor nit, otherwise lgtm.

PR #4025, opened 2023-05-19
Add initial AuthSchemeProvider codegen

gosar:
## Modifications

This includes:
* AuthSchemeParams interface and implementation.
* AuthSchemeProvider interface
* AuthSchemeProvider implementation is stubbed out for now.
* Plumbing to generate authscheme related code

First I'd like to get feedback on whether the generated code itself looks good and the general setup of everything. There are some `TODO: reviewer` questions.

Second, if there are suggestions to improve the code within the `Spec` classes, lmk.

#### Note
This assumes that the service has model-based resolvers, which is most of AWS service, except S3 and EventBridge. Special handling for those 2 services will be added in separate PRs.

## Testing
Added codegen tests for each file.
Ran `./mvnw clean install -pl :codegen,:s3` to confirm generated code builds within real service module.


codegen/src/test/resources/software/amazon/awssdk/codegen/poet/authscheme/auth-scheme-params.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+@Generated("software.amazon.awssdk:codegen")
+@SdkPublicApi
+public interface QueryAuthSchemeParams {
+
+    static Builder builder() {
+        return DefaultQueryAuthSchemeParams.builder();
+    }
+
+    String operation();
```
cenedhryn, line 19: Is it possible to add javadocs for these methods and the builder methods?

codegen/src/main/java/software/amazon/awssdk/codegen/poet/authscheme/AuthSchemeSpecUtils.java
```diff
@@ -0,0 +1,70 @@
...
+    public ClassName providerDefaultImplName() {
+        return ClassName.get(internalPackage(), "Default" + providerInterfaceName().simpleName());
+    }
+
+    public TypeName resolverReturnType() {
+        return ParameterizedTypeName.get(List.class, HttpAuthOption.class);
+    }
+
+    public boolean usesSigV4() {
+        // TODO: reviewer: Does the below method usage change with sigV4a?
```
cenedhryn, line 66: No. The logic of this condition is based on the `signatureVersion` in the service-2.json `metadata` section. When modeling for Sigv4a is introduced, it first of all likely won't affect the `signatureVersion` field, as there are no services that have only Sigv4a today. But in any case, Sigv4a should be added to the internal `AuthType` enum.

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/authscheme/auth-scheme-params.java
gosar: Oh missed some. Thanks for catching. Fixed.

codegen/src/main/java/software/amazon/awssdk/codegen/poet/authscheme/AuthSchemeSpecUtils.java
gosar: Yes. My initial thought was the `region` is an auth scheme param for sigv4 and may not be relevant for sigv4a, so something to think about when we support it. Either ways, it is something that will be evident to handle/adjust (if/as needed) when we do sigv4a.

codegen/src/main/java/software/amazon/awssdk/codegen/poet/authscheme/AuthSchemeParamsSpec.java
```diff
@@ -0,0 +1,131 @@
...
+import com.squareup.javapoet.ParameterizedTypeName;
+import com.squareup.javapoet.TypeSpec;
+import java.util.Optional;
+import javax.lang.model.element.Modifier;
+import software.amazon.awssdk.annotations.SdkPublicApi;
+import software.amazon.awssdk.codegen.model.intermediate.IntermediateModel;
+import software.amazon.awssdk.codegen.poet.ClassSpec;
+import software.amazon.awssdk.codegen.poet.PoetUtils;
+
+// TODO: reviewer: Params v/s Parameters?
```
L-Applin, line 31: Params is fine to me

codegen/src/main/java/software/amazon/awssdk/codegen/internal/Constant.java
```diff
@@ -70,6 +70,9 @@ public final class Constant {
 
     public static final String PACKAGE_NAME_RULES_PATTERN = "%s.endpoints";
 
+    // TODO: reviewer: maybe just auth instead of authscheme? or auth.scheme?
```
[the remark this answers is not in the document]
cenedhryn: I like `auth` as prefix

codegen/src/main/java/software/amazon/awssdk/codegen/poet/authscheme/AuthSchemeParamsSpec.java
```diff
@@ -0,0 +1,131 @@
...
+                              .addModifiers(Modifier.PUBLIC, Modifier.ABSTRACT)
+                              .returns(String.class)
+                              .addJavadoc("Returns the operation for which to resolve the auth scheme.")
+                              .build());
+
+        if (authSchemeSpecUtils.usesSigV4()) {
+            b.addMethod(MethodSpec.methodBuilder("region")
+                                  .addModifiers(Modifier.PUBLIC, Modifier.ABSTRACT)
+                                  // TODO: reviewer: Should region be Regions (Regions.class isn't available here though)
+                                  .returns(ParameterizedTypeName.get(Optional.class, String.class))
```
cenedhryn, line 106: I think it's fine with String, we can make a note to revisit in final review.

codegen/src/main/java/software/amazon/awssdk/codegen/poet/authscheme/AuthSchemeParamsSpec.java
```diff
@@ -0,0 +1,131 @@
...
+                              .addJavadoc("Set the operation for which to resolve the auth scheme.")
+                              .build());
+
+        if (authSchemeSpecUtils.usesSigV4()) {
+            b.addMethod(MethodSpec.methodBuilder("region")
+                                  .addModifiers(Modifier.PUBLIC, Modifier.ABSTRACT)
+                                  .addParameter(ParameterSpec.builder(String.class, "region").build())
+                                  .returns(authSchemeSpecUtils.parametersInterfaceBuilderInterfaceName())
+                                  .addJavadoc("Set the region. The region parameter may be used with  the $S auth scheme.",
+                                              "aws.auth#sigv4") // TODO: Could the known auth schemes be defined somewhere?
```
cenedhryn, line 127: If we're not exposing the auth scheme id (slightly unsure), do we need to reference it? Can we reference the AuthScheme class for Sigv4 instead?

codegen/src/main/java/software/amazon/awssdk/codegen/poet/authscheme/AuthSchemeParamsSpec.java
gosar: Feel the same. Keeping as Params.

codegen/src/main/java/software/amazon/awssdk/codegen/poet/authscheme/AuthSchemeParamsSpec.java
gosar: I think they can be considered public. I used it in the javadoc for [HttpAuthScheme](https://github.com/aws/aws-sdk-java-v2/blob/28ddc6018daf96b8c55e2e8137fca2f86c1d42c4/core/http-auth-spi/src/main/java/software/amazon/awssdk/http/auth/spi/HttpAuthScheme.java#L41). Good point about using the AuthScheme class. It doesn't exist yet. So will keep the TODO to update when that exists.

cenedhryn (approved):

PR #4026, opened 2023-05-22
Adds cross region client logic for decorating endpoint provider

cenedhryn:
## Motivation and Context
Adds the structure and signatures for the new cross region client to be able to successfully decorate the existing S3 client with an override config containing a decorated endpoint provider. 

## Modifications
- Changes the signature of `invokeOperation` (back) to `CompletableFuture`, which is necessary in order to use nonblocking I/O
- Modifies paginated methods that return `Publisher` or `Iterable` to NOT use `invokeOperation` in the abstract delegating clients, i.e. back to originating behavior. The correct place to intercept these calls is when each page is called which will be handled by a later PR. 
- Adds the logic of modifying the request and correctly creating/overriding the RequestOverrideConfig to the cross region client
- Adds a `BucketEndpointProvider` that will contain the cross-region logic (separate PR).

**NOTE**: There is an unchecked cast back to `T` after the request has been modified, due to two contributing factors:
1. The `toBuilder` method returns an S3 request and the type is lost. 
2. The override config method is present on the subtype (T) and on the Aws-level request, but not on the S3 request. 

Adding the override config then returns an Aws request builder. It's not possible to give the compiler enough information to deduce that the request is a T, and we cannot modify the request hierarchy due to backwards compatibility reasons. However, because we control the generation of the subtypes like `GetObjectRequest` etc, we know that the cast is safe. 

## Testing
- Codegen functional testing to verify the delegating client functionality
- S3 functional testing to verify override functionality.

### DIFF
--- a/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionAsyncClient.java
+++ b/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionAsyncClient.java
@@ -17,31 +17,82 @@ package software.amazon.awssdk.services.s3.internal.crossregion;
 
 import java.util.Optional;
 import java.util.function.Function;
+import java.util.concurrent.CompletableFuture;
 import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.awscore.AwsRequestOverrideConfiguration;
+import software.amazon.awssdk.endpoints.Endpoint;
+import software.amazon.awssdk.endpoints.EndpointProvider;
 import software.amazon.awssdk.services.s3.DelegatingS3AsyncClient;
 import software.amazon.awssdk.services.s3.S3AsyncClient;
+import software.amazon.awssdk.services.s3.endpoints.S3EndpointParams;
+import software.amazon.awssdk.services.s3.endpoints.S3EndpointProvider;
 import software.amazon.awssdk.services.s3.model.S3Request;
 
 @SdkInternalApi
 public final class S3CrossRegionAsyncClient extends DelegatingS3AsyncClient {
-
     public S3CrossRegionAsyncClient(S3AsyncClient s3Client) {
         super(s3Client);
     }
 
     @Override
-    protected <T extends S3Request, ReturnT> ReturnT invokeOperation(T request, Function<T, ReturnT> operation) {
-        Optional<String> bucket = request.getValueForField("bucket", String.class);
+    protected <T extends S3Request, ReturnT> CompletableFuture<ReturnT>
+    invokeOperation(T request,
+                    Function<T, CompletableFuture<ReturnT>> operation) {
+
+        Optional<String> bucket = request.getValueForField("Bucket", String.class);
 
         if (!bucket.isPresent()) {
             return operation.apply(request);
         }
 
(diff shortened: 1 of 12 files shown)


services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionAsyncClient.java
```diff
@@ -16,32 +16,79 @@
...
+
+        Optional<String> bucket = request.getValueForField("Bucket", String.class);
 
         if (!bucket.isPresent()) {
             return operation.apply(request);
         }
 
-        //TODO: add modifyRequest logic
-        return operation.apply(request);
+        return operation.apply(requestWithDecoratedEndpointProvider(request, bucket.get()))
```
dagnir, line 46: What does `handleOperationFailure` do? Should we just return the future from `apply` instead?

```
CompletableFuture<T> result = operation.apply(...);
result.whenComplete((r, t) -> handleOperationFailure(...));
return result;
```

The reason for this is that if we return a new future (from `whenComplete`), then we have to handle the case where the caller cancels the future; however "real" S3 client should already have set that up the future returned from `apply`

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionSyncClient.java
```diff
@@ -47,4 +52,43 @@ protected <T extends S3Request, ReturnT> ReturnT invokeOperation(T request, Func
...
     }
+
+    @SuppressWarnings("unchecked")
+    private <T extends S3Request> T requestWithDecoratedEndpointProvider(T request, String bucket) {
+        return (T) request.toBuilder()
+                          .overrideConfiguration(getOrCreateConfigWithEndpointProvider(request, bucket))
+                          .build();
+    }
+
+    //TODO: optimize shared sync/async code
```
dagnir, line 63: Do we have a backlog task in case we forget?

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionAsyncClient.java
dagnir: Another thought, if the `result` gets cancelled by the caller, then we should handle that case in `whenComplete` as well since we probably don't want to remove that cache entry in that case.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionSyncClient.java
cenedhryn: I didn't think we would since it's quite annoying and I'm pretty sure we wouldn't forget, but just in case I added a task.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionAsyncClient.java
cenedhryn: Makes sense. Added this comment to the caching task description.

dagnir (approved):

PR #4040, opened 2023-05-25
Adds accountId to STS credentials providers

cenedhryn:
## Motivation and Context
STS provides accountId in the ARN of the assumedRoleUser included in its response objects. This PR extracts that accountId, if present, and adds it to the credentials identity. 

Resolving credentials end-to-end should work once credentials are resolved before the endpoint.

## Modifications
- Added accountId to `AwsBasicCredentials` and `AwsSessionCredentials` since the credentials providers use those classes. 
- Modified STS providers to find accountId in the response and add it to the returned credentials

## Testing
Added codegen classes test to verify that the accountId is available to the endpoint resolution logic. Because we currently resolve the endpoint before we resolve credentials, this test is not working.

### DIFF
--- a/core/auth/src/main/java/software/amazon/awssdk/auth/credentials/AwsBasicCredentials.java
+++ b/core/auth/src/main/java/software/amazon/awssdk/auth/credentials/AwsBasicCredentials.java
@@ -44,18 +44,17 @@ public final class AwsBasicCredentials implements AwsCredentials {
      * This should be accessed via {@link AnonymousCredentialsProvider#resolveCredentials()}.
      */
     @SdkInternalApi
-    static final AwsBasicCredentials ANONYMOUS_CREDENTIALS = builder().build();
+    static final AwsBasicCredentials ANONYMOUS_CREDENTIALS = builder().validateCredentials(Boolean.FALSE).build();
 
     private final String accessKeyId;
     private final String secretAccessKey;
-
     private final String accountId;
 
     private AwsBasicCredentials(Builder builder) {
         this.accessKeyId = trimToNull(builder.accessKeyId);
         this.secretAccessKey = trimToNull(builder.secretAccessKey);
 
-        if (builder.validateCredentials) {
+        if (Boolean.TRUE.equals(builder.validateCredentials)) {
             Validate.notNull(this.accessKeyId, "Access key ID cannot be blank.");
             Validate.notNull(this.secretAccessKey, "Secret access key cannot be blank.");
         }
@@ -75,7 +74,6 @@ public final class AwsBasicCredentials implements AwsCredentials {
     public static AwsBasicCredentials create(String accessKeyId, String secretAccessKey) {
         return builder().accessKeyId(accessKeyId)
                         .secretAccessKey(secretAccessKey)
-                        .validateCredentials(true)
                         .build();
     }
 
@@ -95,6 +93,9 @@ public final class AwsBasicCredentials implements AwsCredentials {
         return secretAccessKey;
     }
 
+    /**
+     * Retrieve the AWS account id associated with this credentials identity, if found.
+     */
     @Override
     public Optional<String> accountId() {
         return Optional.ofNullable(accountId);
@@ -135,24 +136,36 @@ public final class AwsBasicCredentials implements AwsCredentials {
         private String accessKeyId;
         private String secretAccessKey;
         private String accountId;
-        private boolean validateCredentials;
+        private Boolean validateCredentials = Boolean.TRUE;
 
+        /**
+         * The AWS access key, used to identify the user interacting with services.
+         */
         public Builder accessKeyId(String accessKeyId) {
             this.accessKeyId = accessKeyId;
             return this;
         }
 
+        /**
+         * The AWS secret access key, used to authenticate the user interacting with services.
+         */
         public Builder secretAccessKey(String secretAccessKey) {
             this.secretAccessKey = secretAccessKey;
             return this;
         }
 
+        /**
+         * The AWS account id associated with this credentials identity.
+         */
         public Builder accountId(String accountId) {
             this.accountId = accountId;
             return this;
         }
 
-        public Builder validateCredentials(boolean validateCredentials) {
+        /**
+         * Whether this class should verify that accessKeyId and secretAccessKey are not null.
+         */
+        Builder validateCredentials(Boolean validateCredentials) {
             this.validateCredentials = validateCredentials;
             return this;
         }
--- a/core/auth/src/main/java/software/amazon/awssdk/auth/credentials/AwsBasicCredentials.java
+++ b/core/auth/src/main/java/software/amazon/awssdk/auth/credentials/AwsBasicCredentials.java
@@ -44,7 +44,7 @@ public final class AwsBasicCredentials implements AwsCredentials {
      * This should be accessed via {@link AnonymousCredentialsProvider#resolveCredentials()}.
      */
     @SdkInternalApi
-    static final AwsBasicCredentials ANONYMOUS_CREDENTIALS = builder().validateCredentials(Boolean.FALSE).build();
+    static final AwsBasicCredentials ANONYMOUS_CREDENTIALS = builder().validateCredentials(false).build();
 
     private final String accessKeyId;
     private final String secretAccessKey;
@@ -54,11 +54,23 @@ public final class AwsBasicCredentials implements AwsCredentials {
         this.accessKeyId = trimToNull(builder.accessKeyId);
         this.secretAccessKey = trimToNull(builder.secretAccessKey);
 
-        if (Boolean.TRUE.equals(builder.validateCredentials)) {
+        if (builder.validateCredentials) {
             Validate.notNull(this.accessKeyId, "Access key ID cannot be blank.");
             Validate.notNull(this.secretAccessKey, "Secret access key cannot be blank.");
         }
         this.accountId = builder.accountId;
+    }
+
+    /**
+     * Constructs a new credentials object, with the specified AWS access key and AWS secret key.
+     *
+     * @param accessKeyId The AWS access key, used to identify the user interacting with AWS.
+     * @param secretAccessKey The AWS secret access key, used to authenticate the user interacting with AWS.
+     */
+    protected AwsBasicCredentials(String accessKeyId, String secretAccessKey) {
+        this(builder().accessKeyId(accessKeyId)
+                      .secretAccessKey(secretAccessKey)
+                      .validateCredentials(true));
     }
 
     public static Builder builder() {
@@ -136,7 +148,7 @@ public final class AwsBasicCredentials implements AwsCredentials {
         private String accessKeyId;
         private String secretAccessKey;
         private String accountId;
-        private Boolean validateCredentials = Boolean.TRUE;
+        private boolean validateCredentials = true;
 
         /**
          * The AWS access key, used to identify the user interacting with services.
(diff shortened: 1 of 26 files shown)


services/sts/src/main/java/software/amazon/awssdk/services/sts/auth/StsGetFederationTokenCredentialsProvider.java
```diff
@@ -64,11 +66,23 @@ public static Builder builder() {
...
+    private String accountIdFromArn(FederatedUser federatedUser) {
+        if (federatedUser == null) {
+            return null;
+        }
+        return Arn.parse(federatedUser.arn())
+                  .map(Arn::accountId)
+                  .orElse(null);
     }
 
-    @Override
```
gosar, line 71: `@Override` can remain

services/sts/src/main/java/software/amazon/awssdk/services/sts/auth/SessionCredentialsHolder.java
```diff
@@ -31,18 +31,51 @@ final class SessionCredentialsHolder {
...
+        private Credentials credentials;
+        private Date expiration;
+        private String accountId;
+
+        public Builder credentials(Credentials credentials) {
+            this.credentials = credentials;
+            return this;
+        }
+
+        public Builder expiration(Date expiration) {
```
gosar, line 67: The expiration from `Credentials` is used, so remove this setter?

services/sts/src/main/java/software/amazon/awssdk/services/sts/auth/StsWebIdentityTokenFileCredentialsProvider.java
```diff
@@ -140,10 +141,16 @@ public String toString() {
...
     @Override
-    protected Credentials getUpdatedCredentials(StsClient stsClient) {
+    protected SessionCredentialsHolder getUpdatedCredentials(StsClient stsClient) {
         AssumeRoleWithWebIdentityRequest request = assumeRoleWithWebIdentityRequest.get();
         notNull(request, "AssumeRoleWithWebIdentityRequest can't be null");
-        return stsClient.assumeRoleWithWebIdentity(request).credentials();
+        AssumeRoleWithWebIdentityResponse assumeRoleWithWebIdentityResponse = stsClient.assumeRoleWithWebIdentity(request);
+        return SessionCredentialsHolder.builder()
+                                       .credentials(assumeRoleWithWebIdentityResponse.credentials())
+                                       .accountId(Arn.parse(assumeRoleWithWebIdentityResponse.assumedRoleUser().arn())
```
gosar, line 150: use accountIdFromArn from parent?

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/AwsBasicCredentials.java
```diff
@@ -111,14 +118,47 @@ public boolean equals(Object o) {
...
+
+        public Builder accountId(String accountId) {
+            this.accountId = accountId;
+            return this;
+        }
+
+        public Builder validateCredentials(boolean validateCredentials) {
+            this.validateCredentials = validateCredentials;
+            return this;
+        }
```
millems, line 158: can this be package-protected? There's not a world where customers need to use this.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/AwsBasicCredentials.java
```diff
@@ -111,14 +118,47 @@ public boolean equals(Object o) {
...
         hashCode = 31 * hashCode + Objects.hashCode(secretAccessKey());
+        hashCode = 31 * hashCode + Objects.hashCode(accountId);
         return hashCode;
     }
+
+    public static final class Builder {
+        private String accessKeyId;
+        private String secretAccessKey;
+        private String accountId;
+        private boolean validateCredentials;
```
millems, line 138: default = true?

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/AwsBasicCredentials.java
millems: also, nit: we usually use boxed types, to allow for reverting to the default by setting null. It won't matter here if this is package-protected.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/AwsSessionCredentials.java
```diff
@@ -96,6 +110,38 @@ public int hashCode() {
         hashCode = 31 * hashCode + Objects.hashCode(accessKeyId());
         hashCode = 31 * hashCode + Objects.hashCode(secretAccessKey());
         hashCode = 31 * hashCode + Objects.hashCode(sessionToken());
+        hashCode = 31 * hashCode + Objects.hashCode(accountId);
         return hashCode;
     }
+
+    public static final class Builder {
```
millems, line 117: Should there be any relation between the session credential's builder and the non-session credential's builder?

services/sts/src/main/java/software/amazon/awssdk/services/sts/auth/SessionCredentialsHolder.java
```diff
@@ -31,18 +31,51 @@ final class SessionCredentialsHolder {
...
+        return new Builder();
+    }
+
     public AwsSessionCredentials getSessionCredentials() {
         return sessionCredentials;
     }
 
     public Date getSessionCredentialsExpiration() {
         return sessionCredentialsExpiration;
     }
```
millems, line 55: We should not be using `Date` for anything. `Instant`! (I know you didn't do this, but I'm shocked to see it here).

cenedhryn: Lol, yes I will update this. I kind of noticed it as I was refactoring the class but forgot it

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/AwsSessionCredentials.java
cenedhryn: The two classes are unrelated and old, so I didn't put in the effort. Do you think it's worth being more thorough, and  if so why?

services/sts/src/main/java/software/amazon/awssdk/services/sts/auth/SessionCredentialsHolder.java
cenedhryn: Oh wow it's an instant that we converted to a date and then back to instant again.

services/sts/src/main/java/software/amazon/awssdk/services/sts/auth/StsGetFederationTokenCredentialsProvider.java
cenedhryn: hm, yes that must've happened by mistake

services/sts/src/main/java/software/amazon/awssdk/services/sts/auth/StsWebIdentityTokenFileCredentialsProvider.java
cenedhryn: thanks for finding this, I thought there was one more place but I couldn't find it (should've searched...)

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/AwsSessionCredentials.java
millems: The two classes are related, since AwsSessionCredentials extends AwsCredentials, and the AwsBasicCredentials's builder is essentially the builder for AwsCredentials. It's just in a separate class for *reasons*.

Something being old isn't a great reason not to put in the effort. I guess since this will be deprecated eventually then that's a better reason. In the new types it would be nice to have them share an interface in case someone has generic logic that configures both credential types.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/AwsBasicCredentials.java
```diff
@@ -111,14 +119,59 @@ public boolean equals(Object o) {
...
+        public Builder accountId(String accountId) {
+            this.accountId = accountId;
+            return this;
+        }
+
+        /**
+         * Whether this class should verify that accessKeyId and secretAccessKey are not null.
+         */
+        Builder validateCredentials(Boolean validateCredentials) {
+            this.validateCredentials = validateCredentials;
```
gosar, line 169: nit: I think `this.validateCredentials` can be a `boolean` set to default `true` in declaration and if parameter `Boolean` is passes as null, it sets value to `true`
```
if (validateCredentials == null) {
this.validateCredentials = true;
}
```
allows use of `boolean` everywhere except the setter parameter.

Though, as Matt said, this is package-protected, so not a big deal either ways.

pom.xml
```diff
@@ -635,6 +635,8 @@
                             <!-- Moving the interface methods to new SRA super interfaces causes japicmp to complain -->
                             <exclude>software.amazon.awssdk.auth.credentials.AwsCredentials</exclude>
                             <exclude>software.amazon.awssdk.auth.token.credentials.SdkToken</exclude>
+                            <!-- Removed protected ctor, not used according to internal code usage search -->
+                            <exclude>software.amazon.awssdk.auth.credentials.AwsBasicCredentials</exclude>
```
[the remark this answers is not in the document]
cenedhryn: Adding constructor back.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/AwsBasicCredentials.java
cenedhryn: It got more elegant, adopted this approach

services/sts/src/main/java/software/amazon/awssdk/services/sts/auth/SessionCredentialsHolder.java
```diff
@@ -27,22 +27,48 @@
 @SdkInternalApi
 @ThreadSafe
 final class SessionCredentialsHolder {
```
millems, line 29: Do we even need this anymore, since the expiration is in the session credentials now?

millems (approved):

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/AwsBasicCredentials.java
```diff
@@ -111,14 +131,59 @@ public boolean equals(Object o) {
...
+         */
+        public Builder accountId(String accountId) {
+            this.accountId = accountId;
+            return this;
+        }
+
+        /**
+         * Whether this class should verify that accessKeyId and secretAccessKey are not null.
+         */
+        Builder validateCredentials(Boolean validateCredentials) {
```
gosar, line 180: nit: I think this can be `boolean` too.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/AwsSessionCredentials.java
```diff
@@ -68,10 +78,20 @@ public String sessionToken() {
         return sessionToken;
     }
 
+    @Override
+    public Optional<String> accountId() {
+        return Optional.ofNullable(accountId);
+    }
+
+    public Optional<Instant> expiration() {
```
gosar, line 86: If this makes sense to add to this interface, why not add it to AwsSessionCredentialsIdentity and `@Override` here.

services/sts/src/main/java/software/amazon/awssdk/services/sts/auth/SessionCredentialsHolder.java
cenedhryn: It will be removed in subsequent PRs.

PR #4042, opened 2023-05-26
Fix bug where enum values could not be null

KarolineKarkosch:
This fixes https://github.com/aws/aws-sdk-java-v2/issues/4043

## Motivation and Context
An enum-field can also be null. This works when saving the enum but fails on read, so currently if you set an enum-field to null you create a record in the database which can not be read again. Also the error message is misleading since the problem is not a wrong type.

## Modifications
When an enum is null the object can still be correctly read from the database. Other types still throw an error as would be expected.

## Types of changes

## Checklist

## License


debora-ito:
Closing this for now, as we need more information on the error.

PR #4049, opened 2023-05-29
Fix bug about wrong part size number in CopyObjectHelper

L-Applin:
In `CopyObjectHelper` the method `calculateOptimalPartSizeForCopy(long contentLengthOfSource)` expect the content length of the whole object to copy, but is passed the part size instead. This PR is to fix that issue, now it is passed the content length so that the part size can be automatically adjusted if the total number of parts would exceed 10,000.

### DIFF
--- a/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/CopyObjectHelper.java
+++ b/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/CopyObjectHelper.java
@@ -130,7 +130,7 @@ public final class CopyObjectHelper {
                                Long contentLength,
                                CompletableFuture<CopyObjectResponse> returnFuture,
                                String uploadId) {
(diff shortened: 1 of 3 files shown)


services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/CopyObjectHelper.java
```diff
@@ -130,7 +130,7 @@ private void doCopyInParts(CopyObjectRequest copyObjectRequest,
                                Long contentLength,
                                CompletableFuture<CopyObjectResponse> returnFuture,
                                String uploadId) {
-        long optimalPartSize = calculateOptimalPartSizeForCopy(partSizeInBytes);
+        long optimalPartSize = calculateOptimalPartSizeForCopy(contentLength);
```
zoewangg, line 133: 🤦🏼‍♀️

zoewangg (approved):
Can we add a changelog entry?

PR #4062, opened 2023-06-03
Add new sync and async retryable stages

sugmanue:
This new module includes the interfaces and classes that will be used to implement the new retry logic within the SDK.

### Notes

This change creates a retryable stages for the sync and async pipelines along side with a new helper. To avoid noisy diffs I suffixed the classes names with 2. After this is approved I will move the stages and helper to the original names.

Notice that this change **does not use yet the new retry policies** but instead it will use the adapter, this will allow us to validate that all the current tests cases work fine with the adapter and later on we will change the `SdkDefaultClientBuilder.java` and `AwsSdkDefaultClientBuilder.java` classes to return null if no explicit retry policy is configured and thus pick the retry strategy.

Also, there are several *fixme* comments that will be addressed in a follow up PR and/or when I validate the assumptions made.
 

## Testing
* The current tests are not modified to validate that the wrapper still works on all the existing tests cases.

## Types of changes

## Checklist

## License

### DIFF
--- a/core/aws-core/src/main/java/software/amazon/awssdk/awscore/client/builder/AwsDefaultClientBuilder.java
+++ b/core/aws-core/src/main/java/software/amazon/awssdk/awscore/client/builder/AwsDefaultClientBuilder.java
@@ -378,7 +378,7 @@ public abstract class AwsDefaultClientBuilder<BuilderT extends AwsClientBuilder<
                                        .defaultRetryMode(config.option(SdkClientOption.DEFAULT_RETRY_MODE))
                                        .resolve();
         return AwsRetryPolicy.forRetryMode(retryMode);
-        // fixme This will be changed like this to pick the configured retry strategy
+        // TODO: fixme This will be changed like this to pick the configured retry strategy
         // if no retry policy is configured.
         /*
         RetryPolicy policy = config.option(SdkClientOption.RETRY_POLICY);
--- a/core/aws-core/src/main/java/software/amazon/awssdk/awscore/client/builder/AwsDefaultClientBuilder.java
+++ b/core/aws-core/src/main/java/software/amazon/awssdk/awscore/client/builder/AwsDefaultClientBuilder.java
@@ -39,6 +39,7 @@ import software.amazon.awssdk.awscore.internal.defaultsmode.AutoDefaultsModeDisc
 import software.amazon.awssdk.awscore.internal.defaultsmode.DefaultsModeConfiguration;
 import software.amazon.awssdk.awscore.internal.defaultsmode.DefaultsModeResolver;
 import software.amazon.awssdk.awscore.retry.AwsRetryPolicy;
+import software.amazon.awssdk.awscore.retry.AwsRetryStrategy;
 import software.amazon.awssdk.core.client.builder.SdkDefaultClientBuilder;
 import software.amazon.awssdk.core.client.config.SdkAdvancedClientOption;
 import software.amazon.awssdk.core.client.config.SdkClientConfiguration;
@@ -53,6 +54,7 @@ import software.amazon.awssdk.regions.Region;
 import software.amazon.awssdk.regions.ServiceMetadata;
 import software.amazon.awssdk.regions.ServiceMetadataAdvancedOption;
 import software.amazon.awssdk.regions.providers.DefaultAwsRegionProviderChain;
+import software.amazon.awssdk.retries.api.RetryStrategy;
 import software.amazon.awssdk.utils.AttributeMap;
 import software.amazon.awssdk.utils.CollectionUtils;
 import software.amazon.awssdk.utils.Logger;
@@ -196,6 +198,7 @@ public abstract class AwsDefaultClientBuilder<BuilderT extends AwsClientBuilder<
                             .option(SdkClientOption.EXECUTION_INTERCEPTORS, addAwsInterceptors(configuration))
                             .option(AwsClientOption.SIGNING_REGION, resolveSigningRegion(configuration))
                             .option(SdkClientOption.RETRY_POLICY, resolveAwsRetryPolicy(configuration))
+                            .option(SdkClientOption.RETRY_STRATEGY, resolveAwsRetryStrategy(configuration))
                             .build();
     }
 
@@ -375,6 +378,35 @@ public abstract class AwsDefaultClientBuilder<BuilderT extends AwsClientBuilder<
                                        .defaultRetryMode(config.option(SdkClientOption.DEFAULT_RETRY_MODE))
                                        .resolve();
         return AwsRetryPolicy.forRetryMode(retryMode);
+        // fixme This will be changed like this to pick the configured retry strategy
+        // if no retry policy is configured.
+        /*
+        RetryPolicy policy = config.option(SdkClientOption.RETRY_POLICY);
+
+        if (policy != null) {
+            if (policy.additionalRetryConditionsAllowed()) {
+                return AwsRetryPolicy.addRetryConditions(policy);
+            } else {
+                return policy;
+            }
+        }
+
+        // If we don't have a configured retry policy we will use the configured retry strategy instead.
+        return null;
+         */
+    }
+
+    private RetryStrategy<?, ?> resolveAwsRetryStrategy(SdkClientConfiguration config) {
+        RetryStrategy<?, ?> strategy = config.option(SdkClientOption.RETRY_STRATEGY);
+        if (strategy != null) {
+            return AwsRetryStrategy.addRetryConditions(strategy);
+        }
+        RetryMode retryMode = RetryMode.resolver()
+                                       .profileFile(config.option(SdkClientOption.PROFILE_FILE_SUPPLIER))
+                                       .profileName(config.option(SdkClientOption.PROFILE_NAME))
+                                       .defaultRetryMode(config.option(SdkClientOption.DEFAULT_RETRY_MODE))
+                                       .resolve();
+        return AwsRetryStrategy.forRetryMode(retryMode);
     }
 
     @Override
--- a/core/aws-core/src/main/java/software/amazon/awssdk/awscore/retry/AwsRetryStrategy.java
+++ b/core/aws-core/src/main/java/software/amazon/awssdk/awscore/retry/AwsRetryStrategy.java
@@ -1,0 +1,120 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.awscore.retry;
+
+import software.amazon.awssdk.annotations.SdkPublicApi;
+import software.amazon.awssdk.awscore.exception.AwsServiceException;
+import software.amazon.awssdk.awscore.internal.AwsErrorCode;
+import software.amazon.awssdk.core.internal.retry.SdkDefaultRetryStrategy;
+import software.amazon.awssdk.core.retry.RetryMode;
+import software.amazon.awssdk.retries.AdaptiveRetryStrategy;
+import software.amazon.awssdk.retries.DefaultRetryStrategy;
+import software.amazon.awssdk.retries.LegacyRetryStrategy;
+import software.amazon.awssdk.retries.StandardRetryStrategy;
+import software.amazon.awssdk.retries.api.RetryStrategy;
+
+/**
+ * Retry strategies used by clients when communicating with AWS services.
+ */
+@SdkPublicApi
+public class AwsRetryStrategy {
+
+    private AwsRetryStrategy() {
+    }
+
+    /**
+     * Retrieve the {@link SdkDefaultRetryStrategy#defaultRetryStrategy()} with AWS-specific conditions added.
+     */
+    public static RetryStrategy<?, ?> defaultRetryStrategy() {
+        return forRetryMode(RetryMode.defaultRetryMode());
+    }
+
+    /**
+     * Retrieve the appropriate retry strategy for the retry mode with AWS-specific conditions added.
+     */
+    public static RetryStrategy<?, ?> forRetryMode(RetryMode mode) {
+        switch (mode) {
+            case STANDARD:
+                return standardRetryStrategy();
+            case ADAPTIVE:
+                return adaptiveRetryStrategy();
+            case LEGACY:
+                return legacyRetryStrategy();
+            default:
+                throw new IllegalStateException("unknown retry mode: " + mode);
+        }
+    }
+
+    /**
+     * Update the provided {@link RetryStrategy} to add AWS-specific conditions.
+     */
+    public static RetryStrategy<?, ?> addRetryConditions(RetryStrategy<?, ?> strategy) {
+        return strategy.toBuilder()
+                       .retryOnException(AwsRetryStrategy::retryOnAwsRetryableErrors)
+                       .build();
+    }
+
+    /**
+     * Returns a retry strategy that does not retry.
+     */
+    public static RetryStrategy<?, ?> none() {
+        return DefaultRetryStrategy.none();
+    }
+
+
+    /**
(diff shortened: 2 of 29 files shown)


core/aws-core/src/main/java/software/amazon/awssdk/awscore/client/builder/AwsDefaultClientBuilder.java
```diff
@@ -375,6 +378,35 @@ private RetryPolicy resolveAwsRetryPolicy(SdkClientConfiguration config) {
                                        .defaultRetryMode(config.option(SdkClientOption.DEFAULT_RETRY_MODE))
                                        .resolve();
         return AwsRetryPolicy.forRetryMode(retryMode);
+        // fixme This will be changed like this to pick the configured retry strategy
```
joviegas, line 381: Can we  Tag this as TODO :
Since it is identified in our code checks/checkstyles of release

core/aws-core/src/main/java/software/amazon/awssdk/awscore/client/builder/AwsDefaultClientBuilder.java
```diff
@@ -375,6 +378,35 @@ private RetryPolicy resolveAwsRetryPolicy(SdkClientConfiguration config) {
...
+                return policy;
+            }
+        }
+
+        // If we don't have a configured retry policy we will use the configured retry strategy instead.
+        return null;
+         */
+    }
+
+    private RetryStrategy<?, ?> resolveAwsRetryStrategy(SdkClientConfiguration config) {
```
joviegas, line 399: Can we avoid the Generic wildcard types by referring to the parent classes  ?

core/aws-core/src/main/java/software/amazon/awssdk/awscore/retry/AwsRetryStrategy.java
```diff
@@ -0,0 +1,120 @@
...
+    public static RetryStrategy<?, ?> forRetryMode(RetryMode mode) {
+        switch (mode) {
+            case STANDARD:
+                return standardRetryStrategy();
+            case ADAPTIVE:
+                return adaptiveRetryStrategy();
+            case LEGACY:
+                return legacyRetryStrategy();
+            default:
+                throw new IllegalStateException("unknown retry mode: " + mode);
```
joviegas, line 57: Can we have IllegalArgumentException ?

core/aws-core/src/main/java/software/amazon/awssdk/awscore/retry/AwsRetryStrategy.java
```diff
@@ -0,0 +1,120 @@
...
+        }
+    }
+
+    /**
```
joviegas, line 61: Since this is a public API ,
Can we have Java doc in this format

```java
    /**
     *  Description
     * @param strategy
     * @return
     */
```

core/aws-core/src/main/java/software/amazon/awssdk/awscore/retry/AwsRetryStrategy.java
```diff
@@ -0,0 +1,120 @@
...
+import software.amazon.awssdk.retries.DefaultRetryStrategy;
+import software.amazon.awssdk.retries.LegacyRetryStrategy;
+import software.amazon.awssdk.retries.StandardRetryStrategy;
+import software.amazon.awssdk.retries.api.RetryStrategy;
+
+/**
+ * Retry strategies used by clients when communicating with AWS services.
+ */
+@SdkPublicApi
+public class AwsRetryStrategy {
```
joviegas, line 33: public final class ?

core/sdk-core/src/main/java/software/amazon/awssdk/core/client/config/SdkClientOption.java
```diff
@@ -51,6 +52,12 @@
      */
     public static final SdkClientOption<RetryPolicy> RETRY_POLICY = new SdkClientOption<>(RetryPolicy.class);
 
+    /**
+     * @see ClientOverrideConfiguration#retryStrategy()
+     */
+    @SuppressWarnings("rawtypes")
```
joviegas, line 58: Why is this suppression added?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/RetryableStage2.java
```diff
@@ -0,0 +1,91 @@
...
+import software.amazon.awssdk.core.internal.http.pipeline.RequestPipeline;
+import software.amazon.awssdk.core.internal.http.pipeline.RequestToResponsePipeline;
+import software.amazon.awssdk.core.internal.http.pipeline.stages.utils.RetryableStageHelper2;
+import software.amazon.awssdk.http.SdkHttpFullRequest;
+
+/**
+ * Wrapper around the pipeline for a single request to provide retry, clock-skew and request throttling functionality.
+ */
+@SdkInternalApi
+public final class RetryableStage2<OutputT> implements RequestToResponsePipeline<OutputT> {
```
joviegas, line 36: Can we rename this to more meaningful name like RetryableStageSRA   ? Or any self descriptive name for its b=version 2

core/aws-core/src/main/java/software/amazon/awssdk/awscore/retry/AwsRetryStrategy.java
sugmanue: I was mostly mirroring the [code here](https://github.com/aws/aws-sdk-java-v2/blob/a90b989c77c1cbb7ac0e03b5561c07ef4ea83be7/core/aws-core/src/main/java/software/amazon/awssdk/awscore/retry/AwsRetryPolicy.java#L43) that does not follow this convention, but sure, I can add the `@param` and `@return` tags.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/RetryableStage2.java
sugmanue: This class is temporal to avoid noise in the reviews, it will be renamed to `RetryableStage` and the original class will be removed.

core/sdk-core/src/main/java/software/amazon/awssdk/core/client/config/SdkClientOption.java
sugmanue: I'm using the rawtype `RetryStrategy` instead of `RetryStrategy<B, T>`

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/RetryableStage2.java
sugmanue: (I was going to add a note to the PR description but this is already noted there)

core/aws-core/src/main/java/software/amazon/awssdk/awscore/client/builder/AwsDefaultClientBuilder.java
sugmanue: I can make the change but just to clarify, this is going to be removed in the next PR, and as explained in the description, we kept this like this just to validate that the adapter from retry policy to retry strategy still passes all the unit and integration tests. After that, I will switch to use by default a retry strategy if there's no a user defined retry policy, which will validate that it also does pass all the unit / integration tests already present.

core/aws-core/src/main/java/software/amazon/awssdk/awscore/client/builder/AwsDefaultClientBuilder.java
sugmanue: This type takes the builder and the built types with very specific constraints that allows it to implement the `CopyableBuilder` for the builder and the `ToCopyableBuilder` for the built type and *also* extends it to have separated implementations of this. There's no specific type I can use as we have three different and interchangeable implementations, Legacy, Standard and Adaptive, and the only super type for these is `RetryStrategy`

core/sdk-core/src/main/java/software/amazon/awssdk/core/client/config/SdkClientOption.java
sugmanue: Just to clarify, I cannot use the type params here, even using the generic with wildcards will make the compiler to protest.

core/sdk-core/src/test/java/software/amazon/awssdk/core/internal/http/response/NullErrorResponseHandler.java
```diff
@@ -25,7 +25,9 @@ public class NullErrorResponseHandler implements HttpResponseHandler<SdkServiceE
     @Override
     public SdkServiceException handle(SdkHttpFullResponse response,
                                       ExecutionAttributes executionAttributes) throws Exception {
-        return SdkServiceException.builder().build();
+        return SdkServiceException.builder()
+            .statusCode(response.statusCode())
```
joviegas, line 29: How are we ensuring `SdkHttpFullResponse response` is never null here?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/AsyncRetryableStage2.java
```diff
@@ -0,0 +1,155 @@
...
+            attemptFirstExecute(future);
+            return future;
+        }
+
+        public void attemptFirstExecute(CompletableFuture<Response<OutputT>> future) {
+            Duration backoffDelay = retryableStageHelper.acquireInitialToken();
+            if (backoffDelay.isZero()) {
+                attemptExecute(future);
+            } else {
+                retryableStageHelper.logBackingOff(backoffDelay);
```
joviegas, line 87: I could not find a Junit for this case , how is this tested ?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/RetryableStage2.java
```diff
@@ -0,0 +1,90 @@
...
+    private final RequestPipeline<SdkHttpFullRequest, Response<OutputT>> requestPipeline;
+    private final HttpClientDependencies dependencies;
+
```
joviegas, line 39: Good test coverage for this class 👍

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/retry/RetryPolicyAdapter.java
```diff
@@ -0,0 +1,190 @@
...
+
+    boolean isFastFailRateLimiting() {
+        return Boolean.TRUE.equals(retryPolicy.isFastFailRateLimiting());
+    }
+
+    Duration rateLimitingTokenAcquire() {
+        if (!isRateLimitingEnabled()) {
+            return Duration.ZERO;
+        }
+        OptionalDouble tokenAcquireTimeSeconds = rateLimitingTokenBucket.acquireNonBlocking(1.0, isFastFailRateLimiting());
```
joviegas, line 98: Can we please add coverage for this case in Junit it currently not shown in coevrage

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/retry/SdkDefaultRetryStrategy.java
```diff
@@ -0,0 +1,178 @@
...
+import software.amazon.awssdk.retries.DefaultRetryStrategy;
+import software.amazon.awssdk.retries.LegacyRetryStrategy;
+import software.amazon.awssdk.retries.StandardRetryStrategy;
+import software.amazon.awssdk.retries.api.RetryStrategy;
+
+/**
+ * Retry strategies used by any SDK client.
+ */
+@SdkPublicApi
+public final class SdkDefaultRetryStrategy {
```
joviegas, line 33: Cane we add a Junit test case for this class ?

joviegas:
Can we please take a look at https://github.com/aws/aws-sdk-java-v2/pull/4062#issuecomment-1579672513 and address code smells which ever possible  , or ACK if its already taken care.

core/aws-core/src/main/java/software/amazon/awssdk/awscore/retry/AwsRetryStrategy.java
```diff
@@ -0,0 +1,138 @@
...
+    public static RetryStrategy<?, ?> forRetryMode(RetryMode mode) {
+        switch (mode) {
+            case STANDARD:
+                return standardRetryStrategy();
+            case ADAPTIVE:
+                return adaptiveRetryStrategy();
+            case LEGACY:
+                return legacyRetryStrategy();
+            default:
+                throw new IllegalArgumentException("unknown retry mode: " + mode);
```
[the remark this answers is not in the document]
sugmanue: I will add more tests for these after this is merged in a follow up PR; the idea is to make smaller changes to make it easier to review with self contained changes.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/AsyncRetryableStage2.java
sugmanue: I will add more tests for these after this is merged in a follow up PR; the idea is to make smaller changes to make it easier to review with self-contained minimal changes that build on top of each other.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/retry/RetryPolicyAdapter.java
sugmanue: I will add more tests for these after this is merged in a follow up PR; the idea is to make smaller changes to make it easier to review with self-contained minimal changes that build on top of each other.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/retry/SdkDefaultRetryStrategy.java
sugmanue: I will add more tests for these after this is merged in a follow up PR; the idea is to make smaller changes to make it easier to review with self-contained minimal changes that build on top of each other.

core/sdk-core/src/test/java/software/amazon/awssdk/core/internal/http/response/NullErrorResponseHandler.java
sugmanue: We don't, this class is only used for testing.

sugmanue:
> Can we please take a look at [#4062 (comment)](https://github.com/aws/aws-sdk-java-v2/pull/4062#issuecomment-1579672513) and address code smells which ever possible , or ACK if its already taken care.

I have no idea how to take care of it, the tool is saying that

```
        private static final double SCALE_CONSTANT = 0.4; (1)
```
(1) implies that `SCALE_CONSTANT` is zero which if you look closely is not, `0.4 > 0` and then saying that therefore:

```
   this.timeWindow = Math.pow((this.lastMaxRate * (1 - BETA)) / SCALE_CONSTANT, 1.0 / 3); (2)
```

The use of that constant in (2) will lead to division by zero, which it won't, because is not zero. I think this is a bug in SonarCloud and I don't know how to ACK or silence it, thoughts?

joviegas:
> > Can we please take a look at [#4062 (comment)](https://github.com/aws/aws-sdk-java-v2/pull/4062#issuecomment-1579672513) and address code smells which ever possible , or ACK if its already taken care.
> 
> I have no idea how to take care of it, the tool is saying that
> 
> ```
>         private static final double SCALE_CONSTANT = 0.4; (1)
> ```
> 
> (1) implies that `SCALE_CONSTANT` is zero which if you look closely is not, `0.4 > 0` and then saying that therefore:
> 
> ```
>    this.timeWindow = Math.pow((this.lastMaxRate * (1 - BETA)) / SCALE_CONSTANT, 1.0 / 3); (2)
> ```
> 
> The use of that constant in (2) will lead to division by zero, which it won't, because is not zero. I think this is a bug in SonarCloud and I don't know how to ACK or silence it, thoughts?

Agree , looks like its a bug in SonarCloud Quality Gate.
Did you get a chance to take a look at the Code smells are there any codesmells which we cab remove ?

sugmanue:
> Agree , looks like its a bug in SonarCloud Quality Gate. Did you get a chance to take a look at the Code smells are there any codesmells which we can eliminate ?

Not really, some are well known such as use of wildcards, others are expected such as catching `Throwable` and others will go away soon. Let me know if you have identified any worth taking a look into.

joviegas (approved):

PR #4077, opened 2023-06-08
ddb-enhanced: adds EnhancedType parameters to static builder methods of StaticTableSchema and StaticImmutableTableSchema.

bmaizels:
Adds `EnhancedType` parameters to static builder methods of `StaticTableSchema` and `StaticImmutableTableSchema`.

## Motivation and Context
This is a QoL change that allows a static `TableSchema` to be constructed for a generic class without forcing you to use the raw-type of the class as the type parameter. For use-cases where you may want to dynamically create a static tableschema around a type that is being passed by a caller, this allows you to do so in a type-safe manner.

There are two workarounds I know about this today:

1. You simply use the raw-class and add extra type checking in the methods that get and set the attributes, and any methods that actually use the TableSchema type in its parameters (which is most operations!). This pushes the onus of ensuring type safety on the caller which is not ideal.
2. You create an enclosed dynamic class that implements a generic interface that can be typed with the capture type. It's then safe to use the raw-type of the dynamic class for your TableSchema, but it does not let you export the TableSchema outside that method without losing its type.

Since this change was simple to make and makes good use of the `EnhancedType` anonymous class pattern that is already supported in the same library, I felt that it was worth making this improvement. We already have prior art in other interfaces in this library that accept `EnhancedType<T>` in place of `Class<T>` too.

## Testing
Added unit tests.

## Types of changes

## Checklist

## License

### DIFF
--- a/.changes/next-release/feature-DynamoDBEnhancedClient-270c65a.json
+++ b/.changes/next-release/feature-DynamoDBEnhancedClient-270c65a.json
@@ -1,0 +1,6 @@
+{
+    "category": "DynamoDB Enhanced Client", 
+    "contributor": "bmaizels", 
+    "type": "feature", 
+    "description": "Add EnhancedType parameters to static builder methods of StaticTableSchema and StaticImmitableTableSchema"
+}
--- a/services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/TableSchema.java
+++ b/services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/TableSchema.java
@@ -54,6 +54,18 @@ public interface TableSchema<T> {
     }
 
     /**
+     * Returns a builder for the {@link StaticTableSchema} implementation of this interface which allows all attributes,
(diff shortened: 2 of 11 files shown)


cenedhryn (commented):
Looking good in general, but tests can't compile on JDK8:

~~~
7825 | [ERROR]     cannot access constructor EntityEnvelope()
7826 | [ERROR]       an enclosing instance of type StaticTableSchemaTest is not in scope
7827 | [ERROR] /codebuild/output/src635456198/src/github.com/aws/aws-sdk-java-v2/services-custom/dynamodb-enhanced/src/test/java/software/amazon/awssdk/enhanced/dynamodb/mapper/StaticTableSchemaTest.java:[1517,57] error: invalid method reference
7828 | [ERROR]   non-static method getEntity() cannot be referenced from a static context
7829 | [ERROR]   where T is a type-variable:
7830 | [ERROR]     T extends Object declared in class EntityEnvelope
~~~

bmaizels:
> Looking good in general, but tests can't compile on JDK8:

Fixed and updated PR.

cenedhryn (approved):

PR #4089, opened 2023-06-09
S3CrossRegion Sync and Async Clients Redirect implementation

joviegas:
## Motivation and Context
- Implement redirection in S3CrossRegionSyncClient and S3CrossRegionAsyncClient
- Perform a redirect whenever the first call fails with a redirect exception.
- Pass the supplier to endPointProvider to retrieve the region from cache or headBucket call.

## Modifications

- The initial call is made using the default client.
- If the call fails with a 301 status error, the region is extracted from the error response header.
- The cache is updated, and the call is made by overriding the endpoint provider.
- The overridden endpoint provider checks the cache to obtain the region.
- If the region is not updated in the cache, a HeadBucket call is made to retrieve and update the region.

## Testing

- Junits added
## License

### DIFF
--- a/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionAsyncClient.java
+++ b/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionAsyncClient.java
@@ -20,11 +20,10 @@ import java.util.concurrent.CompletableFuture;
 import java.util.function.Function;
 import software.amazon.awssdk.annotations.SdkInternalApi;
 import software.amazon.awssdk.awscore.AwsRequestOverrideConfiguration;
-import software.amazon.awssdk.endpoints.Endpoint;
 import software.amazon.awssdk.services.s3.DelegatingS3AsyncClient;
 import software.amazon.awssdk.services.s3.S3AsyncClient;
-import software.amazon.awssdk.services.s3.endpoints.S3EndpointParams;
 import software.amazon.awssdk.services.s3.endpoints.S3EndpointProvider;
+import software.amazon.awssdk.services.s3.internal.crossregion.endpointprovider.BucketEndpointProvider;
 import software.amazon.awssdk.services.s3.model.S3Request;
 
 @SdkInternalApi
@@ -67,28 +66,12 @@ public final class S3CrossRegionAsyncClient extends DelegatingS3AsyncClient {
         S3EndpointProvider delegateEndpointProvider = (S3EndpointProvider)
             requestOverrideConfig.endpointProvider().orElseGet(() -> serviceClientConfiguration().endpointProvider().get());
 
+        // TODO : separate PR to provide supplier for Async client
         return requestOverrideConfig.toBuilder()
-                                    .endpointProvider(BucketEndpointProvider.create(delegateEndpointProvider, bucket))
+                                    .endpointProvider(BucketEndpointProvider.create(delegateEndpointProvider, null))
                                     .build();
     }
 
     //TODO: add cross region logic
-    static final class BucketEndpointProvider implements S3EndpointProvider {
-        private final S3EndpointProvider delegate;
-        private final String bucket;
 
-        private BucketEndpointProvider(S3EndpointProvider delegate, String bucket) {
-            this.delegate = delegate;
-            this.bucket = bucket;
-        }
-
-        public static BucketEndpointProvider create(S3EndpointProvider delegate, String bucket) {
-            return new BucketEndpointProvider(delegate, bucket);
-        }
-
-        @Override
-        public CompletableFuture<Endpoint> resolveEndpoint(S3EndpointParams endpointParams) {
-            return delegate.resolveEndpoint(endpointParams);
-        }
-    }
 }
--- a/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionAsyncClient.java
+++ b/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionAsyncClient.java
@@ -18,13 +18,16 @@ package software.amazon.awssdk.services.s3.internal.crossregion;
 import static software.amazon.awssdk.services.s3.internal.crossregion.utils.CrossRegionUtils.getBucketRegionFromException;
 import static software.amazon.awssdk.services.s3.internal.crossregion.utils.CrossRegionUtils.isS3RedirectException;
 import static software.amazon.awssdk.services.s3.internal.crossregion.utils.CrossRegionUtils.requestWithDecoratedEndpointProvider;
+import static software.amazon.awssdk.services.s3.internal.crossregion.utils.CrossRegionUtils.updateUserAgentInConfig;
 
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.BiConsumer;
 import java.util.function.Function;
 import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.awscore.AwsRequestOverrideConfiguration;
 import software.amazon.awssdk.regions.Region;
 import software.amazon.awssdk.services.s3.DelegatingS3AsyncClient;
 import software.amazon.awssdk.services.s3.S3AsyncClient;
@@ -47,44 +50,60 @@ public final class S3CrossRegionAsyncClient extends DelegatingS3AsyncClient {
 
         Optional<String> bucket = request.getValueForField("Bucket", String.class);
 
+        AwsRequestOverrideConfiguration overrideConfiguration = updateUserAgentInConfig(request);
+        T userAgentUpdatedRequest = (T) request.toBuilder().overrideConfiguration(overrideConfiguration).build();
+
         if (!bucket.isPresent()) {
-            return operation.apply(request);
+            return operation.apply(userAgentUpdatedRequest);
         }
         String bucketName = bucket.get();
 
-        if (bucketToRegionCache.containsKey(bucketName)) {
-            return operation.apply(requestWithDecoratedEndpointProvider(request,
-                                                                        () -> bucketToRegionCache.get(bucketName),
-                                                                        serviceClientConfiguration().endpointProvider().get()));
-        }
-
         CompletableFuture<ReturnT> returnFuture = new CompletableFuture<>();
-        operation.apply(request)
-                 .whenComplete((r, t) -> {
-                     if (t != null) {
-                         if (isS3RedirectException(t)) {
-                             bucketToRegionCache.remove(bucketName);
-                             requestWithCrossRegion(request, operation, bucketName, returnFuture, t);
-                             return;
-                         }
-                         returnFuture.completeExceptionally(t);
-                         return;
-                     }
-                     returnFuture.complete(r);
-                 });
+        CompletableFuture<ReturnT> apiOperationFuture = bucketToRegionCache.containsKey(bucketName) ?
+                                                        operation.apply(
+                                                            requestWithDecoratedEndpointProvider(
+                                                                userAgentUpdatedRequest,
+                                                                () -> bucketToRegionCache.get(bucketName),
+                                                                serviceClientConfiguration().endpointProvider().get()
+                                                            )
+                                                        ) :
+                                                        operation.apply(userAgentUpdatedRequest);
+
+        apiOperationFuture.whenComplete(redirectToCrossRegionIfRedirectException(operation,
+                                                                                 userAgentUpdatedRequest,
+                                                                                 bucketName,
+                                                                                 returnFuture));
         return returnFuture;
+    }
+
+    private <T extends S3Request, ReturnT> BiConsumer<ReturnT, Throwable> redirectToCrossRegionIfRedirectException(
+        Function<T, CompletableFuture<ReturnT>> operation,
+        T userAgentUpdatedRequest, String bucketName,
+        CompletableFuture<ReturnT> returnFuture) {
+
+        return (response, throwable) -> {
+            if (throwable != null) {
+                if (isS3RedirectException(throwable)) {
+                    bucketToRegionCache.remove(bucketName);
+                    requestWithCrossRegion(userAgentUpdatedRequest, operation, bucketName, returnFuture, throwable);
+                } else {
+                    returnFuture.completeExceptionally(throwable);
+                }
+            } else {
+                returnFuture.complete(response);
+            }
+        };
     }
 
     private <T extends S3Request, ReturnT> void requestWithCrossRegion(T request,
                                                                        Function<T, CompletableFuture<ReturnT>> operation,
                                                                        String bucketName,
                                                                        CompletableFuture<ReturnT> returnFuture,
-                                                                       Throwable t) {
+                                                                       Throwable throwable) {
 
-        Optional<String> bucketRegionFromException = getBucketRegionFromException((S3Exception) t.getCause());
+        Optional<String> bucketRegionFromException = getBucketRegionFromException((S3Exception) throwable.getCause());
         if (bucketRegionFromException.isPresent()) {
-            sendRequestWithRightRegion(request, operation, bucketName, returnFuture,
-                                       bucketRegionFromException);
+            sendRequestWithRightRegion(request, operation, bucketName, returnFuture, bucketRegionFromException);
         } else {
             fetchRegionAndSendRequest(request, operation, bucketName, returnFuture);
         }
@@ -94,7 +113,6 @@ public final class S3CrossRegionAsyncClient extends DelegatingS3AsyncClient {
                                                                           Function<T, CompletableFuture<ReturnT>> operation,
                                                                           String bucketName,
                                                                           CompletableFuture<ReturnT> returnFuture) {
-
         // // TODO: will fix the casts with separate PR
         ((S3AsyncClient) delegate()).headBucket(b -> b.bucket(bucketName)).whenComplete((response,
                                                                                          throwable) -> {
@@ -102,9 +120,7 @@ public final class S3CrossRegionAsyncClient extends DelegatingS3AsyncClient {
                 if (isS3RedirectException(throwable)) {
                     bucketToRegionCache.remove(bucketName);
                     Optional<String> bucketRegion = getBucketRegionFromException((S3Exception) throwable.getCause());
-
                     if (bucketRegion.isPresent()) {
-                        bucketToRegionCache.put(bucketName, Region.of(bucketRegion.get()));
                         sendRequestWithRightRegion(request, operation, bucketName, returnFuture, bucketRegion);
                     } else {
                         returnFuture.completeExceptionally(throwable);
@@ -112,10 +128,6 @@ public final class S3CrossRegionAsyncClient extends DelegatingS3AsyncClient {
                 } else {
                     returnFuture.completeExceptionally(throwable);
                 }
-            } else {
-                CompletableFuture<ReturnT> newFuture = operation.apply(request);
-                CompletableFutureUtils.forwardResultTo(newFuture, returnFuture);
-                CompletableFutureUtils.forwardExceptionTo(returnFuture, newFuture);
             }
         });
     }
@@ -127,19 +139,11 @@ public final class S3CrossRegionAsyncClient extends DelegatingS3AsyncClient {
                                                                            Optional<String> bucketRegionFromException) {
         String region = bucketRegionFromException.get();
         bucketToRegionCache.put(bucketName, Region.of(region));
-        doSendRequestWithRightRegion(request, operation, returnFuture, region);
-    }
-
-    private <T extends S3Request, ReturnT> void doSendRequestWithRightRegion(T request,
-                                                                             Function<T, CompletableFuture<ReturnT>> operation,
-                                                                             CompletableFuture<ReturnT> returnFuture,
-                                                                             String region) {
         CompletableFuture<ReturnT> newFuture = operation.apply(
             requestWithDecoratedEndpointProvider(request,
                                                  () -> Region.of(region),
                                                  serviceClientConfiguration().endpointProvider().get()));
         CompletableFutureUtils.forwardResultTo(newFuture, returnFuture);
-        // forward exception
         CompletableFutureUtils.forwardExceptionTo(returnFuture, newFuture);
     }
 }
--- a/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionAsyncClient.java
+++ b/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionAsyncClient.java
@@ -26,6 +26,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Function;
 import java.util.function.Supplier;
 import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.awscore.exception.AwsServiceException;
 import software.amazon.awssdk.regions.Region;
 import software.amazon.awssdk.services.s3.DelegatingS3AsyncClient;
 import software.amazon.awssdk.services.s3.S3AsyncClient;
@@ -33,7 +34,6 @@ import software.amazon.awssdk.services.s3.model.HeadBucketRequest;
 import software.amazon.awssdk.services.s3.model.S3Exception;
 import software.amazon.awssdk.services.s3.model.S3Request;
 import software.amazon.awssdk.utils.CompletableFutureUtils;
-import software.amazon.awssdk.utils.StringUtils;
 
 @SdkInternalApi
 public final class S3CrossRegionAsyncClient extends DelegatingS3AsyncClient {
@@ -79,31 +79,24 @@ public final class S3CrossRegionAsyncClient extends DelegatingS3AsyncClient {
 
 
     private Supplier<Region> regionSupplier(String bucket) {
-        return () -> bucketToRegionCache.computeIfAbsent(bucket, this::regionCompletableFuture).join();
+        CompletableFuture<Region> completableFuture = bucketToRegionCache.computeIfAbsent(bucket, this::regionCompletableFuture);
+        return () -> completableFuture.join();
     }
 
     private CompletableFuture<Region> regionCompletableFuture(String bucketName) {
-        StringBuilder stringBuilder = new StringBuilder();
-        return CompletableFuture.supplyAsync(
-                                    () -> ((S3AsyncClient) delegate()).headBucket(HeadBucketRequest.builder()
-                                                                                                   .bucket(bucketName)
-                                                                                                   .build())
-                                                                      .exceptionally(exception -> {
-                                                                          if (isS3RedirectException(exception.getCause())) {
-                                                                              getBucketRegionFromException(
-                                                                                  (S3Exception) exception.getCause()).ifPresent(
-                                                                                  stringBuilder::append);
-                                                                          } else {
-                                                                              CompletableFutureUtils.failedFuture(exception);
-                                                                          }
-                                                                          return null;
-                                                                      }))
-                                .thenApplyAsync(headResponse -> {
-                                    headResponse.join();
-                                    if (headResponse != null && StringUtils.isNotBlank(stringBuilder.toString())) {
-                                        return Region.of(stringBuilder.toString());
-                                    }
-                                    return null;
-                                });
+        return CompletableFuture.supplyAsync(() -> {
+            try {
+                ((S3AsyncClient) delegate()).headBucket(HeadBucketRequest.builder().bucket(bucketName).build()).join();
+            } catch (Exception exception) {
+                if (isS3RedirectException(exception.getCause())) {
+                    String region = getBucketRegionFromException((S3Exception) exception.getCause())
+                        .orElseThrow(() -> AwsServiceException.create("Region name not found in Redirect error",
+                                                                      exception));
+                    return Region.of(region);
+                }
+                throw exception;
+            }
+            return ((S3AsyncClient) delegate()).serviceClientConfiguration().region();
+        });
     }
 }
--- a/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionAsyncClient.java
+++ b/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionAsyncClient.java
@@ -89,9 +89,9 @@ public final class S3CrossRegionAsyncClient extends DelegatingS3AsyncClient {
                                                                                                    .bucket(bucketName)
                                                                                                    .build())
                                                                       .exceptionally(exception -> {
-                                                                          if (isS3RedirectException(exception)) {
+                                                                          if (isS3RedirectException(exception.getCause())) {
                                                                               getBucketRegionFromException(
-                                                                                  (S3Exception) exception).ifPresent(
+                                                                                  (S3Exception) exception.getCause()).ifPresent(
                                                                                   stringBuilder::append);
                                                                           } else {
                                                                               CompletableFutureUtils.failedFuture(exception);
--- a/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionAsyncClient.java
+++ b/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionAsyncClient.java
@@ -15,63 +15,95 @@
 
 package software.amazon.awssdk.services.s3.internal.crossregion;
 
+import static software.amazon.awssdk.services.s3.internal.crossregion.utils.CrossRegionUtils.getBucketRegionFromException;
+import static software.amazon.awssdk.services.s3.internal.crossregion.utils.CrossRegionUtils.isS3RedirectException;
+import static software.amazon.awssdk.services.s3.internal.crossregion.utils.CrossRegionUtils.requestWithDecoratedEndpointProvider;
+
+import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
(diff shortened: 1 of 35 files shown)


services/s3/src/test/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionSyncClientRedirectTest.java
```diff
@@ -0,0 +1,269 @@
...
+        EndpointProvider overridenEndpointProviderPreCache =
+            requestArgumentCaptor.getAllValues().get(1).overrideConfiguration().get().endpointProvider().get();
+        assertThat(overridenEndpointProviderPreCache).isInstanceOf(BucketEndpointProvider.class);
+
+        assertThat(((S3EndpointProvider) overridenEndpointProviderPreCache)
+                       .resolveEndpoint(e -> e.region(Region.US_WEST_2).bucket(CROSS_REGION_BUCKET).build()).get().url().getHost())
+            .isEqualTo("s3." + CROSS_REGION + ".amazonaws.com");
+
+        EndpointProvider overriddenEndpointProviderPostCache =
+            requestArgumentCaptor.getAllValues().get(1).overrideConfiguration().get().endpointProvider().get();
```
cenedhryn, line 137: This call to captor gets the same number `1` as above. 

In general, I am not sure which of these assertions that really show the test case behavior, that the cache is used when the second call is made? I think grouping or removing some of the nice-to-have assertions together will make that easier to see.

services/s3/src/test/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionSyncClientRedirectTest.java
```diff
@@ -0,0 +1,269 @@
...
+        if (region != null) {
+            sdkHttpFullResponseBuilder.appendHeader(X_AMZ_BUCKET_REGION, region);
+        }
+        return S3Exception.builder().
+                          statusCode(statusCode)
+                          .awsErrorDetails(
+                              AwsErrorDetails.builder()
+                                             .sdkHttpResponse(sdkHttpFullResponseBuilder.build())
+                                             .build())
+                          .build();
```
cenedhryn, line 267: weirdly formatted

services/s3/src/test/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionSyncClientRedirectTest.java
```diff
@@ -0,0 +1,269 @@
...
+    void apiCallFailure_afterRedirect_isPropagated() {
+
+        when(mockDelegateClient.serviceClientConfiguration()).thenReturn(ENDPOINT_CONFIGURED);
+
+        when(mockDelegateClient.putObject(any(PutObjectRequest.class), any(RequestBody.class)))
+            .thenThrow(redirectException(301, CROSS_REGION))
+            .thenThrow(SERVICE_EXCEPTION);
+        assertThatExceptionOfType(S3Exception.class).isThrownBy(() -> decoratedS3Client.putObject(p -> p.key("Key").bucket(CROSS_REGION_BUCKET), RequestBody.empty()))
+                                                    .withMessage("Invalid id (Service: S3, Status Code: 400, Request ID: 1, "
+                                                                 + "Extended Request ID: A1)");
```
cenedhryn, line 162: formatting

services/s3/src/test/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionSyncClientRedirectTest.java
```diff
@@ -0,0 +1,269 @@
...
+    }
+
+    @Test
+    void headBucketCallMade_WhenRedirectMessageDoesNotHaveBucketRegion() throws Exception {
+
```
cenedhryn, line 167: Nit: empty lines

services/s3/src/test/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionSyncClientRedirectTest.java
```diff
@@ -0,0 +1,269 @@
...
+        verify(mockDelegateClient, times(2)).putObject(requestArgumentCaptor.capture(), any(RequestBody.class));
+
+        assertThat(requestArgumentCaptor.getAllValues().get(0).overrideConfiguration()).isNotPresent();
+        EndpointProvider overriddenEndpointProviderPreCache =
+            requestArgumentCaptor.getAllValues().get(1).overrideConfiguration().get().endpointProvider().get();
+        assertThat(overriddenEndpointProviderPreCache).isInstanceOf(BucketEndpointProvider.class);
+
+        assertThat(((S3EndpointProvider) overriddenEndpointProviderPreCache)
+                       .resolveEndpoint(e -> e.region(Region.US_WEST_2).bucket(CROSS_REGION_BUCKET).build()).get().url().getHost())
+            .isEqualTo("s3." + CROSS_REGION + ".amazonaws.com");
```
cenedhryn, line 189: maybe extract some common validations to a validation method to make the test cases less verbose

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionSyncClient.java
```diff
@@ -15,80 +15,110 @@
...
-            } catch (Exception e) {
-                handleOperationFailure(e, bucket.get());
+        Optional<String> bucketRequest = bucketNameFromRequest(request);
+        if (!bucketRequest.isPresent()) {
+            return operation.apply(request);
+        }
+        String bucketName = bucketRequest.get();
+        try {
+            if (bucketToRegionCache.containsKey(bucketName)) {
+                return operation.apply(requestWithDecoratedEndpointProvider(request, regionSupplier(bucketName)));
```
cenedhryn, line 63: in actuality, the supplier would always return a bucket name here because we just tested for it, and fetchBucketRegion will never be called, right?

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionSyncClient.java
```diff
@@ -15,80 +15,110 @@
...
+            }
+            return operation.apply(request);
+        } catch (S3Exception exception) {
+            if (exception.statusCode() == REDIRECT_STATUS_CODE) {
+                updateCacheFromRedirectException(exception, bucketName);
+                return operation.apply(requestWithDecoratedEndpointProvider(request, regionSupplier(bucketName)));
```
cenedhryn, line 69: If `fetchBucketRegion` is called because the redirect exception does not contain the region, it seems we will not update the cache?

services/s3/src/test/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionSyncClientRedirectTest.java
```diff
@@ -0,0 +1,269 @@
...
+        when(mockDelegateClient.putObject(any(PutObjectRequest.class), any(RequestBody.class)))
+            .thenThrow(redirectException(301, CROSS_REGION))
+            .thenThrow(SERVICE_EXCEPTION);
+        assertThatExceptionOfType(S3Exception.class).isThrownBy(() -> decoratedS3Client.putObject(p -> p.key("Key").bucket(CROSS_REGION_BUCKET), RequestBody.empty()))
+                                                    .withMessage("Invalid id (Service: S3, Status Code: 400, Request ID: 1, "
+                                                                 + "Extended Request ID: A1)");
+    }
+
+    @Test
+    void headBucketCallMade_WhenRedirectMessageDoesNotHaveBucketRegion() throws Exception {
```
cenedhryn, line 166: the test does not assert what the state of the cache is after the calls

services/s3/src/test/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionSyncClientRedirectTest.java
```diff
@@ -0,0 +1,269 @@
...
+            requestArgumentCaptor.getAllValues().get(1).overrideConfiguration().get().endpointProvider().get();
+        assertThat(overriddenEndpointProviderPreCache).isInstanceOf(BucketEndpointProvider.class);
+
+        assertThat(((S3EndpointProvider) overriddenEndpointProviderPreCache)
+                       .resolveEndpoint(e -> e.region(Region.US_WEST_2).bucket(CROSS_REGION_BUCKET).build()).get().url().getHost())
+            .isEqualTo("s3." + CROSS_REGION + ".amazonaws.com");
+        // Second call to use the local cache
+        ListObjectsResponse afterRegionUpdate = decoratedS3Client.listObjects(i -> i.bucket(CROSS_REGION_BUCKET));
+        assertThat(afterRegionUpdate.contents()).isEqualTo(S3_OBJECTS);
+        requestArgumentCaptor = ArgumentCaptor.forClass(ListObjectsRequest.class);
```
cenedhryn, line 237: Do we need to initialize the captor twice?

services/s3/src/test/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionSyncClientTest.java
```diff
@@ -67,7 +68,7 @@ public void before() {
     public void standardOp_crossRegionClient_noOverrideConfig_SuccessfullyIntercepts() {
         S3Client crossRegionClient = new S3CrossRegionSyncClient(s3Client);
         crossRegionClient.getObject(r -> r.bucket(BUCKET).key(KEY));
-        assertThat(captureInterceptor.endpointProvider).isInstanceOf(S3CrossRegionSyncClient.BucketEndpointProvider.class);
+        assertThat(captureInterceptor.endpointProvider).isInstanceOf(BucketEndpointProvider.class);
     }
```
cenedhryn, line 72: Can we add a couple of tests in this test class too that verifies redirects since this class is more of a functional test that doesn't mock the client?

services/s3/src/test/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionSyncClientRedirectTest.java
joviegas: removed

services/s3/src/test/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionSyncClientTest.java
joviegas: Will continue adding , added some parameterized test cases

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionAsyncClient.java
```diff
@@ -15,80 +15,95 @@
...
             return operation.apply(request);
         }
+        String bucketName = bucket.get();
 
-        return operation.apply(requestWithDecoratedEndpointProvider(request, bucket.get()))
-                        .whenComplete((r, t) -> handleOperationFailure(t, bucket.get()));
+        if (bucketToRegionCache.containsKey(bucketName)) {
+            return operation.apply(requestWithDecoratedEndpointProvider(request,
+                                                                        regionSupplier(bucketName),
+                                                                        serviceClientConfiguration().endpointProvider().get()));
```
joviegas, line 61: Even though this is unlikely should we return DefaultS3EndpointProvider if serviceClientConfiguration().endpointProvider() is null here  ?

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionAsyncClient.java
```diff
@@ -15,80 +15,95 @@
...
+        }
+        return operation.apply(request).thenApply(CompletableFuture::completedFuture)
+                        .exceptionally(exception -> {
+                            if (isS3RedirectException(exception.getCause())) {
```
cenedhryn, line 65: why is it exception.getCause here and exception in the sync case?

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionAsyncClient.java
```diff
@@ -15,80 +15,95 @@
...
+import software.amazon.awssdk.services.s3.model.HeadBucketRequest;
+import software.amazon.awssdk.services.s3.model.S3Exception;
 import software.amazon.awssdk.services.s3.model.S3Request;
+import software.amazon.awssdk.utils.CompletableFutureUtils;
+import software.amazon.awssdk.utils.StringUtils;
 
 @SdkInternalApi
 public final class S3CrossRegionAsyncClient extends DelegatingS3AsyncClient {
+
+    private final Map<String, CompletableFuture<Region>> bucketToRegionCache = new ConcurrentHashMap<>();
```
cenedhryn, line 41: Why do we need to store the region as a `CompletableFuture`?

services/s3/src/test/java/software/amazon/awssdk/services/s3/internal/crossregion/S3DecoratorRedirectBaseTest.java
```diff
@@ -0,0 +1,210 @@
...
+import software.amazon.awssdk.services.s3.model.ListBucketsResponse;
+import software.amazon.awssdk.services.s3.model.ListObjectsRequest;
+import software.amazon.awssdk.services.s3.model.ListObjectsResponse;
+import software.amazon.awssdk.services.s3.model.S3Exception;
+import software.amazon.awssdk.services.s3.model.S3Object;
+
+public abstract class S3DecoratorRedirectBaseTest {
+
+    protected static final String CROSS_REGION_BUCKET = "anyBucket";
+    public static final String X_AMZ_BUCKET_REGION = "x-amz-bucket-region";
```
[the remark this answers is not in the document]
joviegas: Nope this internal package is not available for test

services/s3/src/test/java/software/amazon/awssdk/services/s3/internal/crossregion/S3DecoratorRedirectBaseTest.java
```diff
@@ -0,0 +1,210 @@
...
+    protected static final String CROSS_REGION = Region.EU_CENTRAL_1.toString();
+    protected static final String CHANGED_CROSS_REGION = "us-west-1";
```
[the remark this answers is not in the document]
joviegas: changed

services/s3/src/test/java/software/amazon/awssdk/services/s3/internal/crossregion/S3DecoratorRedirectBaseTest.java
```diff
@@ -0,0 +1,210 @@
...
+
+    protected static final List<S3Object> S3_OBJECTS = Collections.singletonList(S3Object.builder().key("keyObject").build());
+
+    protected static final S3ServiceClientConfiguration ENDPOINT_CONFIGURED =
+        S3ServiceClientConfiguration.builder().endpointProvider(S3EndpointProvider.defaultProvider()).build();
```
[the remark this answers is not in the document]
joviegas: done

services/s3/src/test/java/software/amazon/awssdk/services/s3/internal/crossregion/S3DecoratorRedirectBaseTest.java
```diff
@@ -0,0 +1,210 @@
...
+                                                          .errorMessage(errorMessage != null ? errorMessage : null)
+                                                          .sdkHttpResponse(sdkHttpFullResponseBuilder.build())
+                                                          .errorCode(errorCode)
+                                                          .serviceName("S3")
+                                                          .build())
+                          .build();
+    }
+
+    void verifyTheEndPointProviderOverridden(int attempt, ArgumentCaptor<ListObjectsRequest> requestArgumentCaptor,
+                                             String region) throws Exception {
```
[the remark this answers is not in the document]
joviegas: done

services/s3/src/test/java/software/amazon/awssdk/services/s3/internal/crossregion/S3DecoratorRedirectBaseTest.java
```diff
@@ -0,0 +1,210 @@
...
+        EndpointProvider overridenEndpointProvider =
+            requestArgumentCaptor.getAllValues().get(attempt).overrideConfiguration().get().endpointProvider().get();
+        assertThat(overridenEndpointProvider).isInstanceOf(BucketEndpointProvider.class);
+        assertThat(((S3EndpointProvider) overridenEndpointProvider)
+                       .resolveEndpoint(e -> e.region(Region.US_WEST_2).bucket(CROSS_REGION_BUCKET).build()).get().url().getHost())
```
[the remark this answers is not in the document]
joviegas: Done

services/s3/src/test/java/software/amazon/awssdk/services/s3/internal/crossregion/S3DecoratorRedirectBaseTest.java
```diff
@@ -0,0 +1,210 @@
...
+            .isEqualTo("s3." + region + ".amazonaws.com");
+    }
+
+    protected abstract ListObjectsResponse firstApiCallToService() throws Throwable;
```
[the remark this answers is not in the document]
joviegas: done

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionAsyncClient.java
zoewangg: HeadBucket call is async though, so we probably don't need to use another thread pool to kick it off

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionAsyncClient.java
```diff
@@ -15,80 +15,88 @@
...
-import software.amazon.awssdk.services.s3.endpoints.S3EndpointProvider;
+import software.amazon.awssdk.services.s3.model.HeadBucketRequest;
+import software.amazon.awssdk.services.s3.model.S3Exception;
 import software.amazon.awssdk.services.s3.model.S3Request;
+import software.amazon.awssdk.utils.CompletableFutureUtils;
 
 @SdkInternalApi
 public final class S3CrossRegionAsyncClient extends DelegatingS3AsyncClient {
+
+    private final Map<String, CompletableFuture<Region>> bucketToRegionCache = new ConcurrentHashMap<>();
```
zoewangg, line 41: Should we use LruCache to prevent this map from growing unbounded?

joviegas: LruCache cache doenot provide API to modify the Cache it just caches the content and provides get.
I will create a separate PR to make this bounded cache

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionAsyncClient.java
joviegas: good catch, will modify

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionAsyncClient.java
```diff
@@ -15,80 +15,95 @@
...
-            return new BucketEndpointProvider(delegate, bucket);
-        }
-
-        @Override
-        public CompletableFuture<Endpoint> resolveEndpoint(S3EndpointParams endpointParams) {
-            return delegate.resolveEndpoint(endpointParams);
-        }
+    private CompletableFuture<Region> regionCompletableFuture(String bucketName) {
+        StringBuilder stringBuilder = new StringBuilder();
+        return CompletableFuture.supplyAsync(
```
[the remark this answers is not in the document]
joviegas: done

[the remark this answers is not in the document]
joviegas: done

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionAsyncClient.java
```diff
@@ -15,80 +15,95 @@
...
-        public static BucketEndpointProvider create(S3EndpointProvider delegate, String bucket) {
-            return new BucketEndpointProvider(delegate, bucket);
-        }
-
-        @Override
-        public CompletableFuture<Endpoint> resolveEndpoint(S3EndpointParams endpointParams) {
-            return delegate.resolveEndpoint(endpointParams);
-        }
+    private CompletableFuture<Region> regionCompletableFuture(String bucketName) {
+        StringBuilder stringBuilder = new StringBuilder();
```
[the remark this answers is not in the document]
joviegas: removed

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionSyncClient.java
```diff
@@ -15,80 +15,86 @@
...
+    private Region fetchBucketRegion(String bucketName) {
+        try {
+            ((S3Client) delegate()).headBucket(HeadBucketRequest.builder().bucket(bucketName).build());
+        } catch (S3Exception exception) {
+            if (exception.statusCode() == REDIRECT_STATUS_CODE) {
+                return Region.of(getBucketRegionFromException(exception).orElseThrow(() -> exception));
+            }
+            throw exception;
         }
+        return null;
```
zoewangg, line 96: Should we return current region instead of null if headBucket call doesn't throw exception?

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionAsyncClient.java
```diff
@@ -15,80 +15,124 @@
...
+                                                                        serviceClientConfiguration().endpointProvider().get()));
+        }
 
-    private void handleOperationFailure(Throwable t, String bucket) {
-        //TODO: handle failure case
+        CompletableFuture<ReturnT> returnFuture = new CompletableFuture<>();
+        operation.apply(request)
+                 .whenComplete((r, t) -> {
+                     if (t != null) {
+                         if (isS3RedirectException(t.getCause())) {
```
zoewangg, line 65: Instead of passing t.getCause() here, maybe we can check exception type in the `isS3RedirectException` and check cause there? Something like

```
public boolean isS3RedirectException(Throwable throwable) {
     if (throwable instance of CompletionException) {
             ...check throwable.getCause()
      }

}
```

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionAsyncClient.java
```diff
@@ -15,80 +15,124 @@
...
+                             bucketToRegionCache.remove(bucketName);
+                             Optional<String> bucketRegionFromException =
+                                 getBucketRegionFromException((S3Exception) t.getCause());
+
+                             if (bucketRegionFromException.isPresent()) {
```
zoewangg, line 70: Suggestion: extract this logic (line 67 - line 74) to a separate method to make this method smaller.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionAsyncClient.java
zoewangg: Maybe we can update LruCache to support this? @cenedhryn WDYT?

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionSyncClient.java
joviegas: No its would be not required , since in the BucketEndpointProvider we do  a `crossRegion != null` check. It its null then we use the region passed in the EndpointParams which will be the default params

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionAsyncClient.java
joviegas: done

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionAsyncClient.java
cenedhryn: In any case, I don't think we should add a default here. If it should be needed in the future it should be in the builder/config so that the behavior is consistent.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionAsyncClient.java
```diff
@@ -15,80 +15,95 @@
...
         Optional<String> bucket = request.getValueForField("Bucket", String.class);
 
         if (!bucket.isPresent()) {
             return operation.apply(request);
         }
+        String bucketName = bucket.get();
 
-        return operation.apply(requestWithDecoratedEndpointProvider(request, bucket.get()))
-                        .whenComplete((r, t) -> handleOperationFailure(t, bucket.get()));
+        if (bucketToRegionCache.containsKey(bucketName)) {
```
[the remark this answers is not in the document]
cenedhryn: I meant the code within the if-statement `if (bucketToRegionCache.containsKey(bucketName))`. No other code will execute below that. If the call to `requestWithDecoratedEndpointProvider` gets a redirect, it will not be handled. So the sync and async code will have slightly different behavior in that edge case. Question is if it's likely to occur at all and therefore worth handling - i.e., you get a region from the cache but it's the wrong region. Probably not a huge risk.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionAsyncClient.java
```diff
@@ -15,80 +15,131 @@
...
-
-        return requestOverrideConfig.toBuilder()
-                                    .endpointProvider(BucketEndpointProvider.create(delegateEndpointProvider, bucket))
-                                    .build();
+    private <T extends S3Request, ReturnT> void fetchRegionAndSendRequest(T request,
+                                                                          Function<T, CompletableFuture<ReturnT>> operation,
+                                                                          String bucketName,
+                                                                          CompletableFuture<ReturnT> returnFuture) {
+
+        // // TODO: will fix the casts with separate PR
```
cenedhryn, line 98: Can you also add a `@SuppressWarnings("unchecked")` statement to `requestWithDecoratedEndpointProvider` in `CrossRegionUtils` in that PR?

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionAsyncClient.java
```diff
@@ -15,80 +15,131 @@
...
+        ((S3AsyncClient) delegate()).headBucket(b -> b.bucket(bucketName)).whenComplete((response,
+                                                                                         throwable) -> {
```
cenedhryn, line 100: Can we use the same variables for response/throwable here and further up? I don't mind either r/t or response/throwable, I just think it should be consistent.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionAsyncClient.java
```diff
@@ -15,80 +15,131 @@
...
+    private <T extends S3Request, ReturnT> void doSendRequestWithRightRegion(T request,
+                                                                             Function<T, CompletableFuture<ReturnT>> operation,
+                                                                             CompletableFuture<ReturnT> returnFuture,
+                                                                             String region) {
+        CompletableFuture<ReturnT> newFuture = operation.apply(
+            requestWithDecoratedEndpointProvider(request,
+                                                 () -> Region.of(region),
+                                                 serviceClientConfiguration().endpointProvider().get()));
+        CompletableFutureUtils.forwardResultTo(newFuture, returnFuture);
+        // forward exception
```
cenedhryn, line 142: Nit: unnecessary

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionAsyncClient.java
```diff
@@ -15,80 +15,131 @@
...
-            return new BucketEndpointProvider(delegate, bucket);
-        }
+    private <T extends S3Request, ReturnT> void sendRequestWithRightRegion(T request,
+                                                                           Function<T, CompletableFuture<ReturnT>> operation,
+                                                                           String bucketName,
+                                                                           CompletableFuture<ReturnT> returnFuture,
+                                                                           Optional<String> bucketRegionFromException) {
+        String region = bucketRegionFromException.get();
+        bucketToRegionCache.put(bucketName, Region.of(region));
+        doSendRequestWithRightRegion(request, operation, returnFuture, region);
```
cenedhryn, line 130: In this particular case, I think the code is more readable if the contents of `doSendRequestWithRightRegion` is inlined

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionAsyncClient.java
```diff
@@ -15,80 +15,131 @@
...
+        // // TODO: will fix the casts with separate PR
+        ((S3AsyncClient) delegate()).headBucket(b -> b.bucket(bucketName)).whenComplete((response,
+                                                                                         throwable) -> {
+            if (throwable != null) {
+                if (isS3RedirectException(throwable)) {
+                    bucketToRegionCache.remove(bucketName);
+                    Optional<String> bucketRegion = getBucketRegionFromException((S3Exception) throwable.getCause());
+
+                    if (bucketRegion.isPresent()) {
+                        bucketToRegionCache.put(bucketName, Region.of(bucketRegion.get()));
```
cenedhryn, line 107: Don't need this `put` because there is one inside `sendRequestWithRightRegion`

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionAsyncClient.java
```diff
@@ -15,80 +15,131 @@
...
+                    } else {
+                        returnFuture.completeExceptionally(throwable);
+                    }
+                } else {
+                    returnFuture.completeExceptionally(throwable);
+                }
+            } else {
+                CompletableFuture<ReturnT> newFuture = operation.apply(request);
+                CompletableFutureUtils.forwardResultTo(newFuture, returnFuture);
+                CompletableFutureUtils.forwardExceptionTo(returnFuture, newFuture);
```
cenedhryn, line 118: What use case does this block of code represent? What is the reason to make an unwrapped API call? We already made one further up the chain.

joviegas: This was just a logical block to handle if headObject Passes .
However , in the given code flow it should never pass since this call is made only when we get a redirect error.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionAsyncClient.java
joviegas: done

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionAsyncClient.java
joviegas: done

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionAsyncClient.java
cenedhryn: I usually notice because IntelliJ gets angry and makes lines yellow.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionAsyncClient.java
joviegas: Thanks , Will make the capacity as 300 to remain consistent with V1

https://github.com/aws/aws-sdk-java/blob/71353f994f0a78efac98c50db4da59d1a8a8cdcc/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/AmazonS3Client.java#L480-L490

joviegas: Looks like we cannot implement using ConcurrentHashMap since it internally just does a putAlll and doesnot actually use LinkedHashMap in above

cenedhryn (approved):

PR #4090, opened 2023-06-10
Prep some services modules for easier diff of upcoming changes

gosar:
## Motivation and Context

I'll be working on new AuthSchemeInterceptor and related changes to other code generated classes for auth resolution. To understand what changes are being made to generated code, before working on the codegen itself, I'm thinking of temporarily committing the current generated code into src/main and then making the changes manually, so I can develop the code and we can review the diffs easily. Once those diffs feel ready I can work on the codegen changes to generate that output. And revert this PR then.

## Modifications

* mvn install the module
* copy from target directory to src/main/java
* delete codegen-resources to prevent generation to avoid duplicate class exception
* disabled checkstyle 
* remove these modules from achetype-tools


services/acm/src/main/java/software/amazon/awssdk/services/acm/AcmAsyncClient.java
```diff
@@ -0,0 +1,1805 @@
...
+ * <p>
+ * You can use Certificate Manager (ACM) to manage SSL/TLS certificates for your Amazon Web Services-based websites and
+ * applications. For more information about using ACM, see the <a
+ * href="https://docs.aws.amazon.com/acm/latest/userguide/">Certificate Manager User Guide</a>.
+ * </p>
+ */
+@Generated("software.amazon.awssdk:codegen")
+@SdkPublicApi
+@ThreadSafe
+public interface AcmAsyncClient extends AwsClient {
```
millems, line 68: Will this cause duplicate class errors in IDEs? Should we disable codegen for this module?

services/acm/pom.xml
```diff
@@ -45,6 +45,10 @@
         </plugins>
     </build>
 
+    <properties>
+        <checkstyle.skip>true</checkstyle.skip>
+    </properties>
```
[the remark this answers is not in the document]
gosar: I will entire revert this PR once I get the codegen done for the auth scheme interceptor

services/acm/src/main/java/software/amazon/awssdk/services/acm/AcmAsyncClient.java
gosar: Maybe there way a better way, but I disabled codegen by deleting the `src/main/resources/codegen-resources` from these modules.

services/acm/pom.xml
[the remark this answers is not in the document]
gosar: fyi, I skipped checkstyle because the generated code was failing on it, on the license headers at top of java files. Generated code had a space after asterisk `/* ` that the checkstyle fails on.

services/acm/src/main/java/software/amazon/awssdk/services/acm/AcmAsyncClient.java
millems: Oh, okay. It could be done in the pom.xml, but just deleting it works too since this is temporary.

millems (approved):

PR #4100, opened 2023-06-13
Add IdentityProviderConfiguration implementation

gosar:
## Motivation and Context

Only an interface for IdentityProviderConfiguration was added earlier. Need a default implementation and builder.

## Modifications

Add Builder to the interface and a default internal implementation.

IdentityProviderConfiguration would initially be setup for a client, but we'd want to allow individual request to override its identity provider, so we would need to be able to copy the IdentityProviderConfiguration to mutate. So made it implement ToCopyableBuilder. This meant `@FunctionalInterface` needed to be dropped.

## Testing

Added unit tests.

### DIFF
--- a/core/http-auth-spi/src/main/java/software/amazon/awssdk/http/auth/spi/IdentityProviderConfiguration.java
+++ b/core/http-auth-spi/src/main/java/software/amazon/awssdk/http/auth/spi/IdentityProviderConfiguration.java
@@ -16,8 +16,10 @@
 package software.amazon.awssdk.http.auth.spi;
 
 import software.amazon.awssdk.annotations.SdkPublicApi;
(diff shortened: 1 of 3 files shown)


core/http-auth-spi/src/main/java/software/amazon/awssdk/http/auth/spi/IdentityProviderConfiguration.java
```diff
@@ -26,11 +29,30 @@
  * configured on the SDK.
  */
 @SdkPublicApi
-@FunctionalInterface
```
gosar, line 29: In making it a ToCopyableBuilder the interface also has `toBuilder()` to implement now, so it is no longer a `@FunctionalInterface`.

dagnir (approved):

PR #4102, opened 2023-06-14
MappingSubscriber should halt upon any Exception

polyatail:
## Motivation and Context
When using `SdkPublisher.map`, if the mapping function throws an exception that does not inherit from `RuntimeException` (which often is the case in Kotlin), anything awaiting a result from the Publisher will hang indefinitely, even though the coroutine has ended exceptionally.

## Modifications
This change catches and terminates on all exceptions.

## Types of changes

## Checklist

## License

### DIFF
--- a/utils/src/main/java/software/amazon/awssdk/utils/internal/MappingSubscriber.java
+++ b/utils/src/main/java/software/amazon/awssdk/utils/internal/MappingSubscriber.java
@@ -69,7 +69,7 @@ public class MappingSubscriber<T, U> implements Subscriber<T> {
         if (!isCancelled) {
             try {
                 delegateSubscriber.onNext(mapFunction.apply(t));
-            } catch (RuntimeException e) {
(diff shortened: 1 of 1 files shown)


debora-ito:
@polyatail we are trying to understand which exceptions you want to catch that are not Runtime exceptions. 

Does Kotlin have unchecked exceptions thrown that inherit from Exception but not RuntimeException?
Can you give an example?

debora-ito:
@polyatail let us know if you want to keep this PR alive.

polyatail:
Yes, Kotlin does not built-in exceptions that do not inherit from `RuntimeException`. Even if it didn't, a user could write a piece of code that throws `Exception` and cause a coroutine to hang indefinitely. This has caused us multiple SEVs at our organization already. Is there a reason to only catch `RuntimeException` here and not `Exception` which would be guaranteed to catch everything?

PR #4104, opened 2023-06-15
Implement AwsV4QueryHttpSigner

haydenbaker:
- do NOT merge this PR until the PR which this is dependent on is merged (#4059)

## Motivation and Context
- Adds a default implementation of the AwsV4QueryHttpSigner and tests

## Modifications
- Adds default implementation of the signer
- Adds unit tests to verify behavior
- Refactors some common methods between v4 signers
- Adds copyable builders to sync/async sign requests

## Testing
- Ran tests against new signer
- Ran tests against existing signer which refactor changes impacts

## Types of changes

## Checklist

## License


core/http-auth/src/main/java/software/amazon/awssdk/http/auth/internal/util/SignerConstant.java
```diff
@@ -30,12 +30,28 @@ public final class SignerConstant {
 
     public static final String X_AMZ_SECURITY_TOKEN = "X-Amz-Security-Token";
 
+    public static final String X_AMZ_CREDENTIAL = "X-Amz-Credential";
+
     public static final String X_AMZ_DATE = "X-Amz-Date";
 
+    public static final String X_AMZ_EXPIRES = "X-Amz-Expires";
+
+    public static final String X_AMZ_SIGNED_HEADER = "X-Amz-SignedHeaders";
```
davidh44, line 39: nit: X_AMZ_SIGNED_HEADERS?

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/internal/DefaultAwsV4QueryHttpSigner.java
```diff
@@ -31,11 +55,135 @@ public final class DefaultAwsV4QueryHttpSigner implements AwsV4QueryHttpSigner {
...
+        String scope = buildScope(formattedRequestSigningDate, serviceSigningName, regionName);
+
+        long expirationInSeconds = getSignatureDurationInSeconds(requestSigningInstant, expirationInstant);
+
+        addHostHeader(requestBuilder);
+
+        AwsCredentialsIdentity sanitizedCredentials = sanitizeCredentials(request.identity());
+        if (sanitizedCredentials instanceof AwsSessionCredentialsIdentity) {
+            // For SigV4 query-signing, we need to add "X-Amz-Security-Token"
+            // as a query string parameter, before constructing the canonical
```
gosar, line 96: hmm, the public docs https://docs.aws.amazon.com/IAM/latest/UserGuide/create-signed-request.html#temporary-security-credentials make it seem that this depends on service. So I was thinking this might become a property. But that section in fact seems to apply for when `Authorization` header is used too, i.e., AwsV4HttpSigner, but looks like even there we _always_ do it. So as long as it is carrying over current logic, seems ok.

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/internal/DefaultAwsV4QueryHttpSigner.java
```diff
@@ -31,11 +55,135 @@ public final class DefaultAwsV4QueryHttpSigner implements AwsV4QueryHttpSigner {
...
+        String contentHash = calculateContentHash(request.payload().orElse(null), null);
+
+        SdkHttpRequest signedRequest = doSign(request, contentHash).build();
+        return SyncSignedRequest.builder()
+            .request(signedRequest)
+            .payload(request.payload().orElse(null))
+            .build();
+    }
+
+    private SdkHttpRequest.Builder doSign(SignRequest<?, ? extends AwsCredentialsIdentity> request, String contentHash) {
```
gosar, line 75: I'm still considering whether this can be done via signer properties instead of different interface/impl with lot of duplication.

core/http-auth/src/test/java/software/amazon/awssdk/http/auth/AwsV4QueryHttpSignerTest.java
```diff
@@ -0,0 +1,105 @@
...
+            SdkHttpRequest.builder(),
+            AsyncSignRequest.builder(AwsCredentialsIdentity.create("access", "secret"))
+        );
+
+        AsyncSignedRequest signedRequest = signer.signAsync(request);
+
+        assertEquals(expectedAmzSignature, signedRequest.request().rawQueryParameters().get("X-Amz-Signature").get(0));
+        assertEquals(expectedAmzCredentials, signedRequest.request().rawQueryParameters().get("X-Amz-Credential").get(0));
+        assertEquals(expectedAmzDateHeader, signedRequest.request().rawQueryParameters().get("X-Amz-Date").get(0));
+        assertEquals(expectedAmzExpires, signedRequest.request().rawQueryParameters().get("X-Amz-Expires").get(0));
```
gosar, line 103: seems like we are missing asserts on the Algorithm and SignedHeaders query params

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/internal/DefaultAwsV4QueryHttpSigner.java
```diff
@@ -31,11 +55,135 @@ public final class DefaultAwsV4QueryHttpSigner implements AwsV4QueryHttpSigner {
...
+            serviceSigningName
+        );
+
+        byte[] signature = computeSignature(stringToSign, signingKey);
+
+        String signingCredentials = sanitizedCredentials.accessKeyId() + "/" + scope;
+        requestBuilder.putRawQueryParameter(SignerConstant.X_AMZ_ALGORITHM, SignerConstant.AWS4_SIGNING_ALGORITHM);
+        requestBuilder.putRawQueryParameter(SignerConstant.X_AMZ_DATE, formattedRequestSigningDateTime);
+        requestBuilder.putRawQueryParameter(SignerConstant.X_AMZ_SIGNED_HEADERS, canonicalRequest.signedHeaderString());
+        requestBuilder.putRawQueryParameter(SignerConstant.X_AMZ_EXPIRES, Long.toString(expirationInSeconds));
```
gosar, line 126: hmm, if these are being added later and not part of the canonicalRequestString, that feels wrong.

gosar: yup, the test is failing in the CI.

core/http-auth/src/test/java/software/amazon/awssdk/http/auth/AwsV4QueryHttpSignerTest.java
haydenbaker: I'm pretty sure those are implicit in the signature, but I can add explicit checks

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/AwsV4QueryHttpSigner.java
```diff
@@ -15,18 +15,67 @@
...
+     * This property is required.
+     */
+    SignerProperty<Instant> REQUEST_SIGNING_INSTANT =
+        SignerProperty.create(Instant.class, "RequestSigningInstant");
+
+    /**
+     * The datetime, in milliseconds, for the request.
+     * This property defaults to the max valid duration (7 days) from the signing instant.
+     */
+    SignerProperty<Instant> EXPIRATION_INSTANT =
```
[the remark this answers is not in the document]
haydenbaker: Hmmm.. original code uses an Instant, so I imagine the code to retrieve an Instant is there. Agree with you about using a Duration in general though, must easier to think through as well.

core/http-auth/src/test/java/software/amazon/awssdk/http/auth/AwsV4QueryHttpSignerTest.java
gosar: True for the other query params asserted here too, but for all of these, the asserts become documentation that helps understand the expected behavior better.

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/internal/util/SignerConstant.java
```diff
@@ -30,12 +30,28 @@ public final class SignerConstant {
...
+    public static final String X_AMZ_ALGORITHM = "X-Amz-Algorithm";
+
     public static final String HOST = "Host";
 
     public static final String LINE_SEPARATOR = "\n";
 
+    /**
+     * Seconds in a week, which is the max expiration time Sig-v4 accepts.
+     */
+    public static final long PRESIGN_URL_MAX_EXPIRATION_SECONDS = 60L * 60 * 24 * 7;
```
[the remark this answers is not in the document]
haydenbaker: Agreed.

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/internal/DefaultAwsV4QueryHttpSigner.java
haydenbaker: We can always refactor later if we choose. I don't think the amount of duplication here is harmful, but always open to my mind being changed.

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/internal/DefaultAwsV4QueryHttpSigner.java
haydenbaker: Fixed.

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/AwsV4QueryHttpSigner.java
```diff
@@ -15,18 +15,67 @@
...
+    /**
+     * The datetime, in milliseconds, for the request.
+     * This property is required.
+     */
+    SignerProperty<Instant> REQUEST_SIGNING_INSTANT =
+        SignerProperty.create(Instant.class, "RequestSigningInstant");
+
+    /**
+     * The datetime, in milliseconds, for the request.
+     * This property defaults to the max valid duration (7 days) from the signing instant.
```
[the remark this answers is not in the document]
haydenbaker: I'm not sure I follow what you're saying. Are you saying the expiration shouldn't be a property of the signer?

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/internal/DefaultAwsV4QueryHttpSigner.java
```diff
@@ -31,11 +57,133 @@ public final class DefaultAwsV4QueryHttpSigner implements AwsV4QueryHttpSigner {
...
+        });
+
+        return AsyncSignedRequest.builder()
+            .request(CompletableFutureUtils.joinLikeSync(signedReqFuture))
+            .payload(request.payload().orElse(null))
+            .build();
+    }
+
+    /**
+     * Generates an expiration time, in seconds, for the presigned url. If user has specified
```
gosar, line 173: "Generates" - this method does not need to generate/return anything. It's just a validation method that can return void. The caller already has the Duration and can convert there.

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/internal/DefaultAwsV4QueryHttpSigner.java
```diff
@@ -31,11 +57,133 @@ public final class DefaultAwsV4QueryHttpSigner implements AwsV4QueryHttpSigner {
...
+     * an expiration duration, check if it is in the given limit.
+     */
+    private long getSignatureDurationInSeconds(Instant requestSigningInstant, Duration expirationDuration) {
```
gosar, line 176: method doesn't need requestSigningInstant

core/http-auth/src/test/java/software/amazon/awssdk/http/auth/AwsV4QueryHttpSignerTest.java
```diff
@@ -0,0 +1,110 @@
...
+
+    @Test
+    public void sign_withInvalidExpiration_shouldThrow() {
+        SyncSignRequest<? extends AwsCredentialsIdentity> request = generateBasicRequest(
+            SdkHttpRequest.builder(),
+            SyncSignRequest.builder(AwsCredentialsIdentity.create("access", "secret"))
+                .putProperty(EXPIRATION_DURATION, SignerConstant.PRESIGN_URL_MAX_EXPIRATION_DURATION.plus(Duration.ofSeconds(1)))
+        );
+
+        RuntimeException exception = assertThrows(RuntimeException.class, () -> signer.sign(request));
```
gosar, line 71: assert IllegalArgumentException

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/internal/util/SignerConstant.java
```diff
@@ -30,12 +31,28 @@ public final class SignerConstant {
...
+    public static final String X_AMZ_ALGORITHM = "X-Amz-Algorithm";
+
     public static final String HOST = "Host";
 
     public static final String LINE_SEPARATOR = "\n";
 
+    /**
+     * Seconds in a week, which is the max expiration time Sig-v4 accepts.
+     */
+    public static final Duration PRESIGN_URL_MAX_EXPIRATION_DURATION = Duration.ofSeconds(60L * 60 * 24 * 7);
```
gosar, line 53: Duration.ofDays(7)

core/http-auth-spi/src/main/java/software/amazon/awssdk/http/auth/spi/internal/DefaultSignRequest.java
```diff
@@ -64,10 +64,10 @@ public <T> T property(SignerProperty<T> property) {
...
     protected abstract static class BuilderImpl<B extends Builder<B, PayloadT, IdentityT>, PayloadT,
         IdentityT extends Identity> implements Builder<B, PayloadT, IdentityT> {
-        private SdkHttpRequest request;
-        private PayloadT payload;
-        private IdentityT identity;
-        private final Map<SignerProperty<?>, Object> properties = new HashMap<>();
+        protected SdkHttpRequest request;
+        protected PayloadT payload;
+        protected IdentityT identity;
+        protected Map<SignerProperty<?>, Object> properties = new HashMap<>();
```
gosar, line 70: do these _need_ to change?

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/internal/util/CredentialScope.java
```diff
@@ -0,0 +1,66 @@
...
+import static software.amazon.awssdk.http.auth.internal.util.SignerConstant.AWS4_TERMINATOR;
+import static software.amazon.awssdk.http.auth.internal.util.SignerUtils.formatDateStamp;
+import static software.amazon.awssdk.http.auth.internal.util.SignerUtils.formatTimestamp;
+
+import java.time.Instant;
+import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.identity.spi.AwsCredentialsIdentity;
+
+@SdkInternalApi
+public class CredentialScope {
```
gosar, line 28: final

gosar: `@Immutable`

core/http-auth-spi/src/main/java/software/amazon/awssdk/http/auth/spi/internal/DefaultSyncSignRequest.java
```diff
@@ -38,6 +38,11 @@ public String toString() {
                        .build();
     }
 
+    @Override
+    public SyncSignRequest.Builder<IdentityT> toBuilder() {
+        return new BuilderImpl<>(identity);
```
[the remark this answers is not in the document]
haydenbaker: Fixed.

core/http-auth-spi/src/main/java/software/amazon/awssdk/http/auth/spi/internal/DefaultSignRequest.java
haydenbaker: They do in order to set them in the subclass' `BuilderImpl(...)` method

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/internal/util/CanonicalRequestV2.java
```diff
@@ -0,0 +1,329 @@
...
+        this.canonicalParamsString = getCanonicalQueryString(canonicalParams);
+        this.canonicalHeadersString = getCanonicalHeadersString(canonicalHeaders);
+        this.signedHeadersString = getSignedHeadersString(canonicalHeaders);
+        this.canonicalRequestString = getString();
+    }
+
+    /**
+     * Get the canonical request string for a given {@link SdkHttpRequest}, content-hash, and {@link Options}.
+     */
+    public static String getString(SdkHttpRequest request, String contentHash, Options options) {
```
gosar, line 72: I think remove this

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/internal/util/CanonicalRequestV2.java
```diff
@@ -0,0 +1,329 @@
...
+        return new CanonicalRequestV2(request, contentHash, options).getString();
+    }
+
+    /**
+     * Get the uri-encoded version of the absolute path component URL.
+     * <p>
+     * If the path is empty, a single-forward slash ('/') is returned.
+     */
+    public static String getCanonicalUri(SdkHttpRequest request, Options options) {
```
gosar, line 81: Recommend to make all method private unless they are actually needed to be called by the caller. To make it clear what the real "interface" of this class is.

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/internal/util/CanonicalRequestV2.java
```diff
@@ -0,0 +1,329 @@
...
+    }
+
+    /**
+     * A class for representing options used when creating a {@link CanonicalRequestV2}
+     */
+    public static class Options {
+        final Boolean doubleUrlEncode;
+        final Boolean normalizePath;
+
+        public Options(Boolean doubleUrlEncode, Boolean normalizePath) {
```
gosar, line 324: I think this class is assuming you get non-nulls here. Make them `b`ooleans.

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/internal/util/SignerUtils.java
```diff
@@ -236,10 +191,51 @@ public static byte[] computeSignature(String stringToSign, byte[] signingKey) {
...
+
+        if (sdkChecksum != null && !isUnsignedPayload(contentHashString)) {
+            requestBuilder.putHeader(checksumHeaderName, BinaryUtils.toBase64(sdkChecksum.getChecksumBytes()));
+        }
+    }
+
+    /**
+     * Check if a payload is unsigned based on the content hash
+     */
+    public static Boolean isUnsignedPayload(String contentHashString) {
```
gosar, line 238: `b`oolean

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/internal/DefaultAwsV4HttpSigner.java
```diff
@@ -221,8 +201,72 @@ public AsyncSignedRequest signAsync(AsyncSignRequest<? extends AwsCredentialsIde
...
+        }
+
+        return bodyDigester.digestBytes().thenApply(BinaryUtils::toHex);
+    }
+
+    private CanonicalRequestV2 createCanonicalRequest(SignRequest<?, ? extends AwsCredentialsIdentity> signRequest,
+                                                      SdkHttpRequest request,
+                                                      ContentChecksum contentChecksum) {
+        Boolean doubleUrlEncode = validatedProperty(signRequest, DOUBLE_URL_ENCODE, true);
+        Boolean normalizePath = validatedProperty(signRequest, NORMALIZE_PATH, true);
```
gosar, line 227: I thought we had discussed this in earlier PR, but probably missed, these should be `b`ooleans.

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/internal/DefaultAwsV4HttpSigner.java
```diff
@@ -221,8 +201,72 @@ public AsyncSignedRequest signAsync(AsyncSignRequest<? extends AwsCredentialsIde
...
+    }
+
+    private String createSignature(String stringToSign, byte[] signingKey) {
+        return BinaryUtils.toHex(
+            SignerUtils.computeSignature(stringToSign, signingKey)
+        );
+    }
+
+    // TODO: Rename??
+    public void addSignature(SdkHttpRequest.Builder requestBuilder, String algorithm, AwsCredentialsIdentity credentials,
```
gosar, line 257: parameter `algorithm` seems unused

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/internal/util/CanonicalRequestV2.java
haydenbaker: The static methods are convenience methods for getting certain details about how a request *would* be canonicalized without needing to create a CanonicalRequest object. This is done in the query-signer, because it needs to add a query parameter for the signed-headers string.

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/internal/util/CanonicalRequestV2.java
haydenbaker: Yep. It should use parameters of a canonical request instead of the params of the constructor. Updated.

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/internal/util/CanonicalRequestV2.java
```diff
@@ -0,0 +1,329 @@
...
+            }
+            result.setLength(result.length() - 1);
+            result.append("\n");
+        });
+        return result.toString();
+    }
+
+    /**
+     * Get the string representing which headers are part of the signing process.
+     * Header names are lower-case, sorted in alphabetical order, and separated by a semicolon.
```
gosar, line 229: the lower casing and alphabetical ordering is not part of this method. maybe that part of the comment is meant for getCanonicalHeaders

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/internal/DefaultAwsV4QueryHttpSigner.java
```diff
@@ -15,27 +15,81 @@
...
+        String serviceSigningName = validatedProperty(signRequest, SERVICE_SIGNING_NAME);
+        Duration expirationDuration = validateExpirationDuration(validatedProperty(signRequest, EXPIRATION_DURATION,
+            PRESIGN_URL_MAX_EXPIRATION_DURATION));
+
+        AwsCredentialsIdentity credentials = sanitizeCredentials(signRequest.identity());
+        CredentialScope credentialScope = new CredentialScope(regionName, serviceSigningName, requestSigningInstant);
+
+        addHostHeader(requestBuilder);
+
+        List<Pair<String, List<String>>> canonicalHeaders = getCanonicalHeaders(requestBuilder.build().headers());
```
gosar, line 67: Making the caller get the canonical headers before asking for signed headers `getSignedHeadersString(getCanonicalHeaders())` seems more to ask for the caller, risks bug at call site forgetting to call getCanonicalHeaders. Outside of `getSignedHeadersString()` (overloaded for an already created immutable CanonicalRequest v/s statically for the query case) I don't see the need for CanonicalRequestV2 to have any other non-private method than `getString()`.

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/internal/util/CanonicalRequestV2.java
gosar: I understand, and I know it is an Internal class, but in the spirit of minimizing surface area, we should only make things non-private that callers need. We can always increase visibility if/when needed.

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/internal/DefaultAwsV4QueryHttpSigner.java
```diff
@@ -15,27 +15,81 @@
...
+    public void addSessionCredentials(SdkHttpRequest.Builder requestBuilder,
+                                      AwsSessionCredentialsIdentity credentials) {
+        requestBuilder.putRawQueryParameter(SignerConstant.X_AMZ_SECURITY_TOKEN, credentials.sessionToken());
+    }
+
+    @Override
+    public void addPrerequisites(SdkHttpRequest.Builder requestBuilder,
+                                 SignRequest<?, ? extends AwsCredentialsIdentity> signRequest,
+                                 ContentChecksum contentChecksum) {
+        Instant requestSigningInstant = validatedProperty(signRequest, SIGNING_CLOCK).instant();
```
gosar, line 56: bug alert: Getting a new instant could mean the date used in scope here could be different than the date used in parent's `sign()` if it crosses date boundary in between. I think the scope should be determined

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/internal/DefaultAwsV4HttpSigner.java
```diff
@@ -72,157 +75,225 @@ public SyncSignedRequest sign(SyncSignRequest<? extends AwsCredentialsIdentity>
...
+    /**
+     * Add any prerequisite items to the request using the {@link SdkHttpRequest.Builder}, the ${@link SignRequest},
+     * and the {@link ContentChecksum}
+     * <p>
+     * Such an item could be a header or query parameter that should be included in the signature of the request.
+     */
+    protected void addPrerequisites(SdkHttpRequest.Builder requestBuilder,
+                                    SignRequest<?, ? extends AwsCredentialsIdentity> signRequest,
+                                    ContentChecksum contentChecksum) {
+        Instant requestSigningInstant = validatedProperty(signRequest, SIGNING_CLOCK).instant();
```
gosar, line 198: similar bug alert here too. This could mean the instant used in the x-amz-date header would be different than the one used in the scope.

gosar: I'm thinking how this should have been caught in testing. Not sure if doable, but I think look at if there is a way to setup the Clock in the test such that it uses the fixed time the first time you call instant(), but make the clock tick such that any subsequent call to instant gives a different time, ideally different date, so we explicitly set the test up like that to prevent potential regressions in the future too.

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/internal/util/SignerUtils.java
```diff
@@ -224,22 +151,80 @@ private static byte[] sign(byte[] data, byte[] key, SigningAlgorithm algorithm)
...
+
+        if (sdkChecksum != null && !isUnsignedPayload(contentHashString)) {
+            requestBuilder.putHeader(checksumHeaderName, BinaryUtils.toBase64(sdkChecksum.getChecksumBytes()));
+        }
+    }
+
+    /**
+     * Check if a payload is unsigned based on the content hash
+     */
+    public static boolean isUnsignedPayload(String contentHashString) {
```
gosar, line 227: private

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/internal/DefaultAwsV4HttpSigner.java
```diff
@@ -72,157 +75,225 @@ public SyncSignedRequest sign(SyncSignRequest<? extends AwsCredentialsIdentity>
...
 
-    private SdkHttpRequest.Builder doSign(SyncSignRequest<? extends AwsCredentialsIdentity> request) {
-        String checksumHeaderName = request.property(CHECKSUM_HEADER_NAME);
-        ChecksumAlgorithm checksumAlgorithm = request.property(CHECKSUM_ALGORITHM);
-
-        if (StringUtils.isNotBlank(checksumHeaderName) && checksumAlgorithm == null) {
-            throw new IllegalArgumentException(
-                CHECKSUM_ALGORITHM + " cannot be null when " + CHECKSUM_HEADER_NAME + " is given!");
+    @Override
+    public AsyncSignedRequest signAsync(AsyncSignRequest<? extends AwsCredentialsIdentity> request) {
```
gosar, line 89: In making this class non-final, again to start with strictness and relax later iff needed, make the public sign/signAsync methods final and the methods that don't need to be overridden (except the 3 in query signer) private.

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/internal/DefaultAwsV4HttpSigner.java
```diff
@@ -72,157 +75,225 @@ public SyncSignedRequest sign(SyncSignRequest<? extends AwsCredentialsIdentity>
...
-        String formattedRequestSigningDateTime = formatTimestamp(requestSigningInstant);
-        String regionName = validatedProperty(request, REGION_NAME);
-        String serviceSigningName = validatedProperty(request, SERVICE_SIGNING_NAME);
-        String scope = buildScope(formattedRequestSigningDate, serviceSigningName, regionName);
-
-        AwsCredentialsIdentity sanitizedCredentials = sanitizeCredentials(request.identity());
-        if (sanitizedCredentials instanceof AwsSessionCredentialsIdentity) {
-            addSessionCredentials(requestBuilder, (AwsSessionCredentialsIdentity) sanitizedCredentials);
+        SdkHttpRequest.Builder requestBuilder = signRequest.request().toBuilder();
+        Instant requestSigningInstant = validatedProperty(signRequest, SIGNING_CLOCK).instant();
```
gosar, line 121: nit: move this after required region/signing name, so it can fail fast.

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/internal/DefaultAwsV4HttpSigner.java
```diff
@@ -72,157 +75,225 @@ public SyncSignedRequest sign(SyncSignRequest<? extends AwsCredentialsIdentity>
...
+        String regionName = validatedProperty(signRequest, REGION_NAME);
+        String serviceSigningName = validatedProperty(signRequest, SERVICE_SIGNING_NAME);
```
gosar, line 123: Was going to initially just comment that validating the requiredness of these this late means it possibly wasted spending cycles in generating contentChecksum, so should do this for optimization/fail-fast. But then realized that with signAsync, the method would return with a CompletableFuture first and fail later if these required properties were missing, which is not the desired behavior. Add input validation to the top of the public methods.

Also, please add tests that validate behavior for missing required properties, for both sign and signAsync.

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/internal/DefaultAwsV4QueryHttpSigner.java
gosar: With the refactor to make the query signer a sub-type, and the overrides in the query signer being small, we are close to being able to make "query" a signer property. Will discuss more offline.

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/AwsV4QueryHttpSigner.java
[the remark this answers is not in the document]
gosar: Will discuss offline.

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/AwsV4HttpSigner.java
```diff
@@ -33,11 +33,11 @@
 public interface AwsV4HttpSigner extends HttpSigner<AwsCredentialsIdentity> {
 
     /**
-     * The datetime, as an {@link Instant}, for the request.
+     * A {@link Clock} to be used at the time of signing.
      * This property is required.
```
gosar, line 37: oh, this should be optional.

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/internal/DefaultAwsV4HttpSigner.java
```diff
@@ -72,157 +75,225 @@ public SyncSignedRequest sign(SyncSignRequest<? extends AwsCredentialsIdentity>
...
+    protected String createSignature(String stringToSign, byte[] signingKey) {
+        return BinaryUtils.toHex(
+            SignerUtils.computeSignature(stringToSign, signingKey)
+        );
+    }
+
+    /**
+     * Add the signature to the request in some form (as a header, query-parameter, or otherwise).
+     */
+    protected void addSignature(SdkHttpRequest.Builder requestBuilder, String algorithm, AwsCredentialsIdentity credentials,
```
gosar, line 284: nit: formatting inconsistency. some params are in same line, some are in different lines.

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/internal/DefaultAwsV4HttpSigner.java
```diff
@@ -72,157 +75,225 @@ public SyncSignedRequest sign(SyncSignRequest<? extends AwsCredentialsIdentity>
...
+        String authHeader = algorithm
+            + " Credential="
+            + credentialScope.scope(credentials)
+            + ", SignedHeaders="
+            + canonicalRequest.getSignedHeadersString()
+            + ", Signature="
+            + signature;
+
+        requestBuilder
+            .putHeader(SignerConstant.AUTHORIZATION, authHeader);
```
gosar, line 297: nit: one line

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/internal/DefaultAwsV4HttpSigner.java
```diff
@@ -72,157 +75,225 @@ public SyncSignedRequest sign(SyncSignRequest<? extends AwsCredentialsIdentity>
...
-                .payload(request.payload().orElse(null))
-                .build();
-        }
+    /**
+     * Add any prerequisite items to the request using the {@link SdkHttpRequest.Builder}, the ${@link SignRequest},
+     * and the {@link ContentChecksum}
+     * <p>
+     * Such an item could be a header or query parameter that should be included in the signature of the request.
+     */
+    protected void addPrerequisites(SdkHttpRequest.Builder requestBuilder,
```
gosar, line 195: nit: code layout. these methods in the middle of methods related to checksum made it slightly worse to follow the code while reading. co-locating related methods helps.

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/internal/DefaultAwsV4HttpSigner.java
```diff
@@ -72,157 +75,225 @@ public SyncSignedRequest sign(SyncSignRequest<? extends AwsCredentialsIdentity>
...
+                                    SignRequest<?, ? extends AwsCredentialsIdentity> signRequest,
+                                    ContentChecksum contentChecksum) {
+        Instant requestSigningInstant = validatedProperty(signRequest, SIGNING_CLOCK).instant();
+        String formattedRequestSigningDateTime = formatTimestamp(requestSigningInstant);
```
gosar, line 199: small smell: to use `formatTimestamp` here directly, whereas in subclass, within same method, it derives it from `credentialScope.getDatetime()`. seems like it could be better encapsulated.

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/internal/util/CredentialScope.java
```diff
@@ -0,0 +1,68 @@
...
+    public String getDatetime() {
+        return formatTimestamp(instant);
+    }
+
+    public String scope() {
+        return getDate() + "/" + region + "/" + service + "/" + AWS4_TERMINATOR;
+    }
+
+    public String scope(AwsCredentialsIdentity credentials) {
+        return credentials.accessKeyId() + "/" + getDate() + "/" + region + "/" + service + "/" + AWS4_TERMINATOR;
```
gosar, line 66: nit: credentials.accessKeyId() + "/" + scope();

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/internal/DefaultAwsV4HttpSigner.java
haydenbaker: Yeah, I agree. I don't agree with the naming, so I re-named.

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/internal/DefaultAwsV4HttpSigner.java
haydenbaker: Hmm... I wish check-style would enforce this kind of consistency.

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/internal/DefaultAwsV4HttpSigner.java
haydenbaker: There should probably be some kind of interface method to enforce that. I'll add one.

haydenbaker: Also, in terms of optimization, we should focus on the most frequent case, because otherwise, if we have to validate properties more than once, we're wasting cycles there. If properties are specified from the caller, it should most always add the correct request parameters, right?

If not, then we should add members to the implementation and set those members when `validatedProperty` is called so that we don't need to get and validate a property more than once.

haydenbaker: I would agree with adding members, it eases implementation and re-usability; though, at the cost of reason-ability in some ways. This also can be a behavior or the default signers, not the interface, to keep the flexibility.

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/internal/DefaultAwsV4HttpSigner.java
haydenbaker: I can add a test to verify.

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/internal/DefaultAwsV4QueryHttpSigner.java
haydenbaker: You'd have to try pretty hard to give the wrong headers to `getSignedHeadersString(List<Pair<String, List<String>>> canonicalHeaders)` - I get the spirit of your comment, but I don't necessarily agree in this case.

All the members are set at object construction, so calling `getSignedHeadersString()` is guaranteed to get you the string. 

Based on your other comments about keeping things private before they are needed, sure, I can do that.

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/internal/DefaultAwsV4HttpSigner.java
```diff
@@ -72,157 +94,224 @@ public SyncSignedRequest sign(SyncSignRequest<? extends AwsCredentialsIdentity>
...
-        if (StringUtils.isNotBlank(checksumHeaderName) && checksumAlgorithm == null) {
-            throw new IllegalArgumentException(
-                CHECKSUM_ALGORITHM + " cannot be null when " + CHECKSUM_HEADER_NAME + " is given!");
+    @Override
+    public AsyncSignedRequest signAsync(AsyncSignRequest<? extends AwsCredentialsIdentity> request) {
+        if (CredentialUtils.isAnonymous(request.identity())) {
+            return AsyncSignedRequest.builder()
+                .request(request.request())
+                .payload(request.payload().orElse(null))
+                .build();
```
haydenbaker, line 115: I feel like this shouldn't need to be here. I think the caller should really be determining whether to sign a request or not. Thoughts?

core/http-auth-spi/src/main/java/software/amazon/awssdk/http/auth/spi/internal/DefaultSyncSignRequest.java
```diff
@@ -52,6 +57,13 @@ public BuilderImpl(IdentityT identity) {
             super(identity);
         }
 
+        private BuilderImpl(DefaultSyncSignRequest<IdentityT> request) {
```
haydenbaker, line 60: The other builder methods are causing the spotbugs check to fail because they are public:
```
software.amazon.awssdk.http.auth.spi.internal.DefaultSyncSignRequest$BuilderImpl()V must not be public.

software.amazon.awssdk.http.auth.spi.internal.DefaultSyncSignRequest$BuilderImpl(Lsoftware/amazon/awssdk/identity/spi/Identity;)V must not be public.

[ERROR] Medium: Bad toBuilder implementation. See the SpotBugs logs for problem details. [software.amazon.awssdk.http.auth.spi.internal.DefaultSyncSignRequest$BuilderImpl] At DefaultSyncSignRequest.java:[lines 52-53] BAD_TO_BUILDER

[ERROR] Medium: Bad toBuilder implementation. See the SpotBugs logs for problem details. [software.amazon.awssdk.http.auth.spi.internal.DefaultSyncSignRequest$BuilderImpl] At DefaultSyncSignRequest.java:[lines 57-58] BAD_TO_BUILDER


```

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/internal/DefaultAwsV4HttpSigner.java
```diff
@@ -52,16 +57,33 @@
...
  * A default implementation of {@link AwsV4HttpSigner}.
  */
 @SdkInternalApi
-public final class DefaultAwsV4HttpSigner implements AwsV4HttpSigner {
+public class DefaultAwsV4HttpSigner implements AwsV4HttpSigner {
 
     private static final Logger LOG = Logger.loggerFor(DefaultAwsV4HttpSigner.class);
 
+    // required
+    protected String regionName;
```
gosar, line 70: This makes the signers stateful and not usable across parallel requests. I know the existing code `AuthScheme.signer()` is creating a new instance of signer, but I had been considering a change to use a single instance. I'd want to double check with @millems if there was any requirement of the signers to be thread safe.

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/internal/DefaultAwsV4HttpSigner.java
gosar: I'm less worried about optimization than the behavior, like the signAsync case I mentioned. So definitely add the unit tests.

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/internal/DefaultAwsV4HttpSigner.java
gosar: I didn't see a way to instantiate a Clock with the java.time.Clock methods that'd allow this. Probably good to use a mock Clock with `when(mockClock.instant()).thenReturn(instant).thenReturn(instantPlusADay)`. And possibly also asserting using `verify` with `times(1)` or `verifyNoMore`.

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/internal/DefaultAwsV4HttpSigner.java
gosar: I have wondered about this too. But the caller is going to be identity type agnostic and this is AWSCredentials specific. Also, I'm not 100% sure at this point where _all_ Anonymous credentials can be set. Keeping this here seems fine.

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/internal/DefaultAwsV4HttpSigner.java
haydenbaker: Yep. If we don't want this, then we'll need to refactor all of this to a "parameters" object. The `setParameters` method would then create this object, and every method would pass it along. The signer then becomes re-usable, but that parameters object has to be created on every request.

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/internal/DefaultAwsV4HttpSigner.java
haydenbaker: This is done in the `setParameters()` method.

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/internal/DefaultAwsV4HttpSigner.java
haydenbaker: I've done it in the event-stream test, I simply created an implementation of Clock that allows for setting the time.

core/http-auth-spi/src/main/java/software/amazon/awssdk/http/auth/spi/SyncSignRequest.java
```diff
@@ -44,6 +46,6 @@ static <IdentityT extends Identity> Builder<IdentityT> builder(IdentityT identit
      */
     interface Builder<IdentityT extends Identity>
         extends SignRequest.Builder<Builder<IdentityT>, ContentStreamProvider, IdentityT>,
-                SdkBuilder<Builder<IdentityT>, SyncSignRequest<IdentityT>> {
+            CopyableBuilder<Builder<IdentityT>, SyncSignRequest<IdentityT>> {
```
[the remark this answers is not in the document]
gosar: Instead of `this.properties = request.properties;` in the builder's constructor, it could call `this.properties.putAll(request.properties)`. This would also let you not remove the `final` from the parent builder's properties member.

core/http-auth-spi/src/main/java/software/amazon/awssdk/http/auth/spi/internal/DefaultSignRequest.java
gosar: The setter methods could be used instead of making them protected, but `protected` sounds fine too. Though, I think the map should remain `final`. See another comment later where we discuss that.

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/internal/util/SignerUtils.java
```diff
@@ -224,22 +151,80 @@ private static byte[] sign(byte[] data, byte[] key, SigningAlgorithm algorithm)
...
+        if (sdkChecksum != null && !isUnsignedPayload(contentHashString)) {
+            requestBuilder.putHeader(checksumHeaderName, BinaryUtils.toBase64(sdkChecksum.getChecksumBytes()));
+        }
+    }
+
+    /**
+     * Check if a payload is unsigned based on the content hash
+     */
+    private static boolean isUnsignedPayload(String contentHashString) {
+        return "UNSIGNED_PAYLOAD".equals(contentHashString) || "STREAMING-UNSIGNED-PAYLOAD-TRAILER".equals(contentHashString);
```
gosar, line 228: Seems like this should be "UNSIGNED-PAYLOAD"? dash instead of underscore

core/http-auth-spi/src/main/java/software/amazon/awssdk/http/auth/spi/internal/DefaultSignRequest.java
haydenbaker: changed.

core/http-auth-spi/src/main/java/software/amazon/awssdk/http/auth/spi/internal/DefaultSyncSignRequest.java
haydenbaker: fixed.

core/http-auth-spi/src/main/java/software/amazon/awssdk/http/auth/spi/internal/DefaultAsyncSignRequest.java
```diff
@@ -45,14 +58,21 @@ public static final class BuilderImpl<IdentityT extends Identity>
...
-        public BuilderImpl(IdentityT identity) {
+        private BuilderImpl(IdentityT identity) {
             super(identity);
         }
 
+        private BuilderImpl(DefaultAsyncSignRequest<IdentityT> request) {
+            properties(request.properties)
+                .identity(request.identity)
+                .payload(request.payload)
+                .request(request.request);
```
gosar, line 73: nit: the indentation from the 2nd line makes this odd to read. if that's the default style because of chaining the methods, maybe don't chain them as that works too here.

core/http-auth-spi/src/main/java/software/amazon/awssdk/http/auth/spi/internal/DefaultSyncSignRequest.java
```diff
@@ -44,14 +57,21 @@ public static final class BuilderImpl<IdentityT extends Identity>
...
-        public BuilderImpl(IdentityT identity) {
+        private BuilderImpl(IdentityT identity) {
             super(identity);
         }
 
+        private BuilderImpl(DefaultSyncSignRequest<IdentityT> request) {
+            properties(request.properties)
+                .identity(request.identity)
+                .payload(request.payload)
+                .request(request.request);
```
gosar, line 72: nit: the indentation from the 2nd line makes this odd to read. if that's the default style because of chaining the methods, maybe don't chain them as that works too here.

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/internal/util/CanonicalRequestV2.java
```diff
@@ -0,0 +1,341 @@
...
+            }
+            result.setLength(result.length() - 1);
+            result.append("\n");
+        });
+        return result.toString();
+    }
+
+    /**
+     * Get the string representing which headers are part of the signing process.
+     * Header names are lower-case, in alphabetical order, and separated by a semicolon.
```
gosar, line 229: the lower casing and alphabetical ordering is not part of this method. it is part of the comment for getCanonicalHeaders. remove it from here.

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/internal/util/SignerUtils.java
```diff
@@ -224,22 +151,80 @@ private static byte[] sign(byte[] data, byte[] key, SigningAlgorithm algorithm)
...
+        if (sdkChecksum != null && !isUnsignedPayload(contentHashString)) {
+            requestBuilder.putHeader(checksumHeaderName, BinaryUtils.toBase64(sdkChecksum.getChecksumBytes()));
+        }
+    }
+
+    /**
+     * Check if a payload is unsigned based on the content hash
+     */
+    private static boolean isUnsignedPayload(String contentHashString) {
+        return "UNSIGNED-PAYLOAD".equals(contentHashString) || "STREAMING-UNSIGNED-PAYLOAD-TRAILER".equals(contentHashString);
```
gosar, line 228: I see you fixed from `_` to `-`. Are there tests for both these cases? Or atleast make sure they are covered as part of other signers later.

core/http-auth/src/test/java/software/amazon/awssdk/http/auth/AwsV4HttpSignerTest.java
```diff
@@ -398,4 +377,79 @@ public void sign_withCrc32Checksum_and_withTrailinggHeader_shouldSignWithCrc32Ch
...
+            AwsCredentialsIdentity.create("access", "secret"),
+            (httpRequest -> {
+            }),
+            (signRequest -> signRequest.putProperty(SignerProperty.create(String.class, "ServiceSigningName"), null))
+        );
+
+        RuntimeException exception = assertThrows(RuntimeException.class, () -> signer.sign(request));
+
+        assertThat(exception.getMessage()).contains("must not be null");
+    }
```
gosar, line 454: s/NullPointerException/RuntimeException. Also add these tests for signAsync (more important).

core/http-auth/src/test/java/software/amazon/awssdk/http/auth/AwsV4QueryHttpSignerTest.java
```diff
@@ -0,0 +1,120 @@
...
+import software.amazon.awssdk.http.auth.spi.SyncSignRequest;
+import software.amazon.awssdk.http.auth.spi.SyncSignedRequest;
+import software.amazon.awssdk.identity.spi.AwsCredentialsIdentity;
+
+class AwsV4QueryHttpSignerTest {
+
+    private static final AwsV4QueryHttpSigner signer = AwsV4QueryHttpSigner.create();
+
+    @Test
+    public void sign_withBasicRequest_shouldSign() {
```
gosar, line 39: add tests for when valid EXPIRATION_DURATION is specified and asserts that corresponding value is used (v/s default in the tests here).

core/http-auth/src/test/java/software/amazon/awssdk/http/auth/TestUtils.java
```diff
@@ -1,9 +1,77 @@
...
+                .putHeader("Host", "demo.us-east-1.amazonaws.com")
+                .putHeader("x-amz-archive-description", "test  test")
+                .encodedPath("/")
+                .uri(URI.create("http://demo.us-east-1.amazonaws.com"))
+                .build()
+                .copy(requestOverrides))
+            .payload(() -> new ByteArrayInputStream("{\"TableName\": \"foo\"}".getBytes()))
+            .putProperty(SignerProperty.create(String.class, "RegionName"), "us-east-1")
+            .putProperty(SignerProperty.create(String.class, "ServiceSigningName"), "demo")
+            .putProperty(SignerProperty.create(Clock.class, "SigningClock"), Clock.fixed(Instant.ofEpochMilli(351153000968L),
```
gosar, line 40: Don't use `fixed` in any tests.

gosar (approved):
Mostly LG. Few minor changes needed and tests to add. But giving approval so don't have to wait for another review.

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/internal/util/SignerUtils.java
haydenbaker: Other signers.

core/http-auth/src/test/java/software/amazon/awssdk/http/auth/TestUtils.java
gosar: you fixed one use `fixed` (pun intended ;)) but missed another

PR #4105, opened 2023-06-15
Add User Agent Api name for Cross region API calls

joviegas:
## Motivation and Context

- Add User Agent Api name for Cross region API calls
## Modifications

- Before calling the delgate the request Override config is updated
Final Sample User agent will look like this

```
Optional[aws-sdk-java/2.20.62-SNAPSHOT Mac_OS_X/19.9.1 OpenJDK_128-Bit_Server_VM/11.0.19+7-LTS Java/11.3.19 vendor/Comp_Inc. io/sync http/UNKNOWN cfg/retry-mode/legacy hll/cross-region]

```

## Testing

- Updated Junit

### DIFF
--- a/services/s3/src/test/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionAsyncClientTest.java
+++ b/services/s3/src/test/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionAsyncClientTest.java
@@ -100,9 +100,16 @@ class S3CrossRegionAsyncClientTest {
 
     @Test
     void standardOp_crossRegionClient_containUserAgent() {
-        S3AsyncClient crossRegionClient = new S3CrossRegionAsyncClient(s3Client);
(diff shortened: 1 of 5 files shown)


services/s3/src/test/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionAsyncClientTest.java
```diff
@@ -98,6 +98,13 @@ void crossRegionClient_createdWithWrapping_SuccessfullyIntercepts() {
         assertThat(captureInterceptor.endpointProvider).isInstanceOf(S3CrossRegionAsyncClient.BucketEndpointProvider.class);
     }
 
+    @Test
+    void standardOp_crossRegionClient_containUserAgent() {
```
cenedhryn, line 102: can we also test that the user agent contains something that was there before the change?

joviegas: Done

PR #4109, opened 2023-06-16
Move QueryParametersToBodyInterceptor to front of interceptor chain

davidh44:
## Motivation and Context

Move QueryParametersToBodyInterceptor to the front of interceptor chain so that it comes before other interceptors, e.g., checksum, compression. Some query protocols have custom interceptors, e.g., presigning, that need to come before QueryParametersToBodyInterceptor, so moving custom interceptors to front of interceptor chain for query protocols as well.

## Types of changes

## Checklist

## License


test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/protocolquery/MoveQueryParamsToBodyTest.java
dagnir, line 1: Why is this test removed?

codegen/src/main/java/software/amazon/awssdk/codegen/poet/builder/BaseClientBuilderClass.java
```diff
@@ -288,16 +288,6 @@ private MethodSpec finalizeServiceConfigurationMethod() {
...
 
-        if (model.getMetadata().isQueryProtocol()) {
-            TypeName listType = ParameterizedTypeName.get(List.class, ExecutionInterceptor.class);
-            builder.addStatement("$T protocolInterceptors = $T.singletonList(new $T())",
-                                 listType,
-                                 Collections.class,
-                                 QueryParametersToBodyInterceptor.class);
-            builder.addStatement("interceptors = $T.mergeLists(interceptors, protocolInterceptors)",
-                                 CollectionUtils.class);
-        }
```
dagnir, line 299: Not sure I understand the need for the other changes. Any reason we can't just change the ordering of this statement?

dagnir (commented):
Can you make sure to run integ tests as well?

test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/protocolquery/MoveQueryParamsToBodyTest.java
davidh44: Now that QueryParametersToBodyInterceptor is moved to the front, a custom interceptor like this test interceptor `AdditionalQueryParamInterceptor` wouldn't work, since interceptors set in the client override config are added to the end of the chain after the global interceptors

codegen/src/main/java/software/amazon/awssdk/codegen/poet/builder/BaseClientBuilderClass.java
davidh44: In this `finalizeServiceConfiguration()` the service-specific interceptors are added to the `SdkClientConfiguration` as `SdkClientOption.EXECUTION_INTERCEPTORS` which are then merged with the global interceptors in [SdkDefaultClientBuilder](https://github.com/aws/aws-sdk-java-v2/blob/master/core/sdk-core/src/main/java/software/amazon/awssdk/core/client/builder/SdkDefaultClientBuilder.java#L435). So changing the ordering here would only move the interceptor to the front of the service-specific interceptors; it would still come after the global interceptors. 

Alternatively, we could create another SdkClientOption here and merge it in SdkDefaultClientBuilder, but to keep cross-module dependencies down Matt proposed we add it in the `SyncClientBuilderClass` / `AsyncClientBuilderClass`

davidh44:
> Can you make sure to run integ tests as well?

Integ tests ran successfully

https://us-west-2.console.aws.amazon.com/codesuite/codebuild/131990247566/projects/aws-sdk-java-v2-IntegrationTest-JDK8/build/aws-sdk-java-v2-IntegrationTest-JDK8%3Aeec777b0-d186-43f7-a590-02071520d609/?region=us-west-2

test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/protocolquery/MoveQueryParamsToBodyTest.java
dagnir: If the test is broken, we should rewrite/fix it; we should continue to have a test to ensure this behavior stays in the client.

codegen/src/main/java/software/amazon/awssdk/codegen/poet/builder/SyncClientBuilderClass.java
```diff
@@ -119,26 +124,46 @@ private MethodSpec endpointProviderMethod() {
...
+            return b;
+        }
+
+        TypeName listType = ParameterizedTypeName.get(List.class, ExecutionInterceptor.class);
+        b.addStatement("$T interceptors = clientConfiguration.option($T.EXECUTION_INTERCEPTORS)", listType, SdkClientOption.class)
+            .addStatement("interceptors.add(0, new $T())", QueryParametersToBodyInterceptor.class);
+
+        List<String> customInterceptors = model.getCustomizationConfig().getInterceptors();
+        Collections.reverse(customInterceptors);
+        customInterceptors.forEach(i -> b.addStatement("interceptors.add(0, new $T())", ClassName.bestGuess(i)));
```
dagnir, line 164: This is pretty confusing logic. Can we simplify this to remove the reverse and `add(0`?

dagnir:
> some query protocols have custom interceptors, e.g., presigning, that need to come before QueryParametersToBodyInterceptor, 

What services are currently doing this? Seems like that wouldn't work if the endpoint provider returns a different endpoint.

davidh44:
> > some query protocols have custom interceptors, e.g., presigning, that need to come before QueryParametersToBodyInterceptor,
> 
> What services are currently doing this? Seems like that wouldn't work if the endpoint provider returns a different endpoint.

Checked the (17) query protocols, these services have custom interceptors
- DocDb / Neptune / RDS: Presigning interceptors
- IAM: `GlobalServiceExecutionInterceptor` (wrapper for `HelpfulUnknownHostExceptionInterceptor`, stage is `modifyException`)
- SQS: `MessageMD5ChecksumInterceptor` (stage is `afterExecution`)

Integ tests passed, so looks like its working?

codegen/src/main/java/software/amazon/awssdk/codegen/poet/builder/SyncClientBuilderClass.java
davidh44: Simplified to use `CollectionUtils.mergeLists()` for clarity

test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/protocolquery/MoveQueryParamsToBodyTest.java
davidh44: Discussed offline, we should be okay with this change since technically, you’re allowed to mix and match; you can have some params in the body and some in the URL

codegen/src/main/java/software/amazon/awssdk/codegen/poet/builder/SyncClientBuilderClass.java
```diff
@@ -119,26 +126,53 @@ private MethodSpec endpointProviderMethod() {
...
+    private MethodSpec.Builder addQueryProtocolInterceptors(MethodSpec.Builder b) {
+        if (!model.getMetadata().isQueryProtocol()) {
+            return b;
+        }
+
+        TypeName listType = ParameterizedTypeName.get(List.class, ExecutionInterceptor.class);
+
+        b.addStatement("$T interceptors = clientConfiguration.option($T.EXECUTION_INTERCEPTORS)",
+                       listType, SdkClientOption.class)
+            .addStatement("$T queryParamsToBodyInterceptor = new $T<>($T.asList(new $T()))",
```
dagnir, line 164: nit: `Collections.singletonList`

davidh44: oops, right that was how it was originally in BaseClientBuilderClass

dagnir (approved):

PR #4111, opened 2023-06-16
AuthScheme resolution wiring via hand written Interceptor

gosar:
## Motivation and Context

In https://github.com/aws/aws-sdk-java-v2/pull/4090 I committed the generated clients, so I could make manual edits to get a working SRA auth resolution wiring solution, to facilitate review, before working on code generator changes.

## Modifications

For SRA resolution, we have a service specific resolver and need a service specific entry point. The RequestPipeline is service agnostic. The current approach for service specific logic is an ExecutionInterceptor, similar to for endpoints 2.0.

* Set up client level AuthSchemes and identityProviderConfiguration
* Add a AuthSchemeInterceptor for each service. Interceptor is called from AwsExecutionContextBuilder as part of runInitialInterceptors. Incerpertor does auth resolution and saves the selected auth scheme to use as ExecutionAttribute to be used later.
* SigningStage updated to use this attribute, if present (to retain old code path, at least for now).

There are a lot of TODOs, but this gets the simplest case (1 auth at service level, with basic sigv4 and bearer) working. Additional cases of other signers, will be added in future PRs.

I've put a number of TODOs/review TODOs in code for discussion, future action. These will be cleaned up later, so don't worry about them getting into the branch at this time.

## Testing

* For bearer: Added and ensured a CodeCatalystIntegrationTest passes (after setting up a profile locally for bearer token)
* For sigv4: Ensured AwsCertificateManagerIntegrationTest passes


core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/SigningStage.java
```diff
@@ -52,9 +61,73 @@ public SigningStage(HttpClientDependencies dependencies) {
...
+
+        IdentityProvider<T> identityProvider = selectedAuthScheme.identityProvider();
+        T identity = identityProvider.resolveIdentity(identityRequestBuilder.build()).join();
+
+        // TODO: resolveIdentity(Consumer builder) doesn't get used
+
+        // Consumer<ResolveIdentityRequest.Builder> c = r -> r.putProperty(IdentityProperty.create(String.class, "key"), "value");
+        // identityProvider.resolveIdentity(c);
+        // TODO: Consumer Builder is not working because of ambiguity!!
+        // identityProvider.resolveIdentity(r -> r.putProperty(IdentityProperty.create(String.class, "key"), "value"));
```
gosar, line 91: I think that's because ResolveIdentityRequest has only one abstract method - `putIdentityProperty`. I learnt in a different [PR](https://github.com/aws/aws-sdk-java-v2/pull/4100/files#r1230091719) that making this ToCopyableBuilder will make it not a functional interface and I think making ResolveIdentityRequest copyable sounds good, so I can do that in separate PR. LMK thoughts.

core/sdk-core/src/main/java/software/amazon/awssdk/core/interceptor/SdkExecutionAttribute.java
```diff
@@ -109,6 +114,33 @@ public class SdkExecutionAttribute {
     public static final ExecutionAttribute<ChecksumValidation> HTTP_RESPONSE_CHECKSUM_VALIDATION = new ExecutionAttribute<>(
         "HttpResponseChecksumValidation");
 
+    /**
+     * The auth scheme provider used to resolve the auth scheme for a request.
+     */
+    // TODO: Should this be in SdkInternalExecutionAttribute like ENDPOINT_PROVIDER?
```
cenedhryn, line 120: Can they all be internal @millems ?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/SigningStage.java
```diff
@@ -52,9 +61,73 @@ public SigningStage(HttpClientDependencies dependencies) {
...
+        // SyncSignedRequest signedRequest =
+        //     signer.sign(r -> r.identity(identity)
+        //                       .request(request)
+        //                       .payload(request.contentStreamProvider().orElse(null))
+        //                       .putProperty(SignerProperty.create(String.class, "key"), "value"));
+
+        return toSdkHttpFullRequest(signedRequest);
+    }
+
+    private SdkHttpFullRequest toSdkHttpFullRequest(SyncSignedRequest signedRequest) {
```
cenedhryn, line 112: I don't know this code well but it seems odd that we must do this conversion - guess because of the new signer APIs. Is it the right place to put it?

services/acm/src/main/java/software/amazon/awssdk/services/acm/auth/scheme/internal/DefaultAcmAuthSchemeProvider.java
```diff
@@ -35,6 +41,13 @@ public static DefaultAcmAuthSchemeProvider create() {
...
     @Override
     public List<AuthSchemeOption> resolveAuthScheme(AcmAuthSchemeParams authSchemeParams) {
-        return new ArrayList<>();
+        AwsV4AuthScheme awsV4AuthScheme = AwsV4AuthScheme.create();
+        return Arrays.asList(AuthSchemeOption.builder()
+                                             .schemeId(awsV4AuthScheme.schemeId())
+                                             .putSignerProperty(SERVICE_SIGNING_NAME, "acm")
+                                             .putSignerProperty(REGION_NAME, authSchemeParams.region().id())
+                                             // TODO: Make this signer property optional
+                                             .putSignerProperty(REQUEST_SIGNING_INSTANT, Instant.now())
```
cenedhryn, line 50: What does this signing property do? What is the point of storing an instant at client creation time?

services/acm/src/main/java/software/amazon/awssdk/services/acm/auth/scheme/internal/AcmAuthSchemeInterceptor.java
```diff
@@ -0,0 +1,132 @@
...
+        // @sigv4.
+        String operation = executionAttributes.getAttribute(SdkExecutionAttribute.OPERATION_NAME);
+        // TODO: Do this only if sigv4
+        Region region = executionAttributes.getAttribute(AwsExecutionAttribute.AWS_REGION);
+
+        // Use the operation-level auth scheme resolver if the customer
+        // specified an override, otherwise fall back to the one on the client.
+        // TODO: The fallback should happen while setting in execution attributes
+        AcmAuthSchemeProvider authSchemeProvider =
+            (AcmAuthSchemeProvider) executionAttributes.getAttribute(SdkExecutionAttribute.AUTH_SCHEME_RESOLVER);
```
millems, line 64: Is the `AUTH_SCHEME_RESOLVER` needed outside of this module?

If not, should we have a codegen'd `AcmExecutionAttribute.AUTH_SCHEME_RESOLVER`? That would allow us to make it `AcmAuthSchemeProvider` so that it's type-safe through the whole execution, and it would also let us make it an `@SdkInternalApi` instead of an `@SdkProtectedApi`.

services/acm/src/main/java/software/amazon/awssdk/services/acm/auth/scheme/internal/AcmAuthSchemeInterceptor.java
```diff
@@ -0,0 +1,132 @@
...
+
+        IdentityProviderConfiguration identityResolvers =
+            executionAttributes.getAttribute(SdkExecutionAttribute.IDENTITY_PROVIDER_CONFIGURATION);
+
+        StringBuilder failureReasons = new StringBuilder();
+
+        // Check each option, in the order the auth scheme resolver proposed them.
+        for (AuthSchemeOption authOption : authOptions) {
+            // If we're using no-auth, don't consider which options are enabled.
+            if (authOption.schemeId().equals("smithy.auth#noAuth")) {
```
millems, line 89: Is this synthetic ID definitely approved?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/SigningStage.java
gosar: Conversion needed because output of new signer API needs to be converted to output of SigningStage (SdkHttpFullRequest). I don't want `toSdkHttpFullRequest` in the new http-auth-spi module. A builder/create method inside SdkHttpFullRequest is an option, but that's a SdkPublicApi, so hidden inside this SigningStage is probably better. When the request pipeline is rewritten, it _may_ not need the conversion?

services/acm/src/main/java/software/amazon/awssdk/services/acm/auth/scheme/internal/DefaultAcmAuthSchemeProvider.java
gosar: This value will be the requestTime(x-amz-date) used by the signer.

It won't be storing an instant at client creation time. `resolveAuthScheme` will be called for each request.

I've talked to Hayden about making this optional in the signer. However there is also the use case of accounting for TIME_OFFSET where the REQUEST_SIGNING_INSTANT value will need to be passed by caller. For that case, it can't be set here. I don't know currently where that would get set to solve for the TIME_OFFSET case - that case needs more thinking. But I'll handle that in separate PR. For now, I'll remove setting this from here, once it is made optional.

core/aws-core/src/main/java/software/amazon/awssdk/awscore/client/builder/AwsDefaultClientBuilder.java
```diff
@@ -194,6 +194,10 @@ protected final SdkClientConfiguration finalizeChildConfiguration(SdkClientConfi
         configuration = mergeSmartDefaults(configuration);
 
         IdentityProvider<? extends AwsCredentialsIdentity> identityProvider = resolveCredentialsIdentityProvider(configuration);
+        // add the AwsCredentialsIdentity IdentityProvider to the IdentityProviders configured for the client
+        identityProviderConfigurationBuilder.putIdentityProvider(identityProvider);
```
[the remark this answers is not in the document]
gosar: I followed the pattern similar to `clientContextParams` and based on some discussion in internal thread. Later the fully built IdentityProviderConfiguration is in SdkClientConfiguration, but if you think the `identityProviderConfigurationBuilder` should be too, lmk.

core/aws-core/src/main/java/software/amazon/awssdk/awscore/client/builder/AwsDefaultClientBuilder.java
```diff
@@ -194,6 +194,10 @@ protected final SdkClientConfiguration finalizeChildConfiguration(SdkClientConfi
...
+        // TODO: Does anything have the opportunity to change this clientOption after here?
```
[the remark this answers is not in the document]
gosar: > Opportunity? I suppose so. I don't think anything else will, except for the request-level identity provider overrides.

ok. I won't worry about it now then.

core/aws-core/src/main/java/software/amazon/awssdk/awscore/internal/AwsExecutionContextBuilder.java
```diff
@@ -135,6 +142,35 @@ private AwsExecutionContextBuilder() {
...
+
+        // TODO: this may have to be setup at operation level, as some operations may use different signer for same schemeId,
+        // e.g., EventStreamV4AuthScheme
+
+        // TODO: If request level override is specified, should each operation check that overridden scheme is the
+        //  appropriate type (uses the appropriate Signer) for streaming, etc.
+        Map<String, AuthScheme<?>> authSchemes = clientConfig.option(SdkClientOption.AUTH_SCHEMES);
+
+        // TODO: If request level identity provider is specified, it should override.
+        // review TODO: IDENTITY_PROVIDERS or IDENTITY_PROVIDER_CONFIGURATION
```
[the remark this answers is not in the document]
gosar: Yup. Removing this TODO.

services/acm/src/main/java/software/amazon/awssdk/services/acm/auth/scheme/internal/AcmAuthSchemeInterceptor.java
gosar: Not yet. I still have to work on getting that reviewed.

core/sdk-core/src/main/java/software/amazon/awssdk/core/SelectedAuthScheme.java
```diff
@@ -0,0 +1,56 @@
...
+import software.amazon.awssdk.http.auth.spi.AuthSchemeOption;
+import software.amazon.awssdk.http.auth.spi.HttpSigner;
+import software.amazon.awssdk.identity.spi.Identity;
+import software.amazon.awssdk.identity.spi.IdentityProvider;
+import software.amazon.awssdk.utils.Validate;
+
+/**
+ * A container for the identity resolver, signer and auth option that we selected for use with this service call attempt.
+ */
+@SdkProtectedApi
```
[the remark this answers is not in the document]
gosar: The code generated auth scheme interceptor needs to instantiate it and set it as attribute, which SigningStage (in core module) reads. Since they are different modules I understood it can't be `@SdkInternalApi`.

core/sdk-core/src/main/java/software/amazon/awssdk/core/interceptor/SdkExecutionAttribute.java
```diff
@@ -109,6 +114,33 @@ public class SdkExecutionAttribute {
...
+    public static final ExecutionAttribute<AuthSchemeProvider> AUTH_SCHEME_RESOLVER = new ExecutionAttribute<>(
+        "AuthSchemeProvider");
```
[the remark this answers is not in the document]
gosar: https://github.com/aws/aws-sdk-java-v2/blob/ce8e53b70293074accf2363f3998dfd8e16bed5b/core/sdk-core/src/main/java/software/amazon/awssdk/core/interceptor/SdkInternalExecutionAttribute.java#L27 says
> Only generated code from the SDK clients should set these values.

But these 3 are set by AwsExecutionContextBuilder (and read by generated AuthSchemeInterceptor)
AUTH_SCHEME_RESOLVER
AUTH_SCHEMES
IDENTITY_PROVIDER_CONFIGURATION

And SELECTED_AUTH_SCHEME is set by generated AuthSchemeInterceptor and read by SigningStage.

If moving all of these to `SdkInternalExecutionAttribute` doesn't violate that javadoc, I can move it.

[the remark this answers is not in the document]
gosar: >The customer should be able to override the auth scheme resolver, but they can do that through the request-level override.

They should be able to provide one at the client level too right? Which will set the `SdkClientOption.AUTH_SCHEME_PROVIDER`. In AwsExecutionContextBuilder the clientOption is set as ExecutionAttribute.

core/aws-core/src/main/java/software/amazon/awssdk/awscore/internal/AwsExecutionContextBuilder.java
```diff
@@ -135,6 +142,35 @@ private AwsExecutionContextBuilder() {
                                .build();
     }
 
+    private static void putAuthSchemeResolutionAttributes(ExecutionAttributes executionAttributes,
+                                                          SdkClientConfiguration clientConfig) {
+
+        // review TODO: should this be client v/s advanced option?
```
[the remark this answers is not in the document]
gosar: >I was thinking changing auth schemes and their providers would be on the generated service client builder.

Currently I haven't implemented any of the entry points for changing these. For auth scheme provider, yeah it should be in the generated service client builder so it can be the specific {Service}AuthSchemeProvider. For putAuthScheme, that would be on the AwsClientBuilder as it is not service specific. Your design doc calls out both these. But maybe it is a typo, but your said "and their provider**s**" (plural), so not sure if you are talking about something else there. If so, please clarify.

Either ways, my _review TODO_ question here is about, whether the client option for AUTH_SCHEME_PROVIDER be SdkClientOption.AUTH_SCHEME_PROVIDER or SdkAdvancedClientOption.AUTH_SCHEME_PROVIDER. 
* Today SIGNER is SdkAdvancedClientOption.SIGNER.
* I guess I'm not clear what should be clientOption v/s advancedClientOption. 
* I see making it advanced client option would make the user do `clientBuilder.overrideConfiguration(b->b.putAdvancedOption(AUTH_SCHEME_RESOLVER, resolver))` v/s clientBuilder.authSchemeResolver(resolver). Well, in this case, since we want to use specific {Service}AuthSchemeProvider, using putAdvancedOption would not by type-safe. So makes sense to not use Advanced here.
* But question applies to AUTH_SCHEMES and IDENTITY_PROVIDER_CONFIGURATION options - they are not service specific, so should they be client v/s advnaced?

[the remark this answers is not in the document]
gosar: Hmm, I just noticed that SdkAdvancedClientOption is `@SdkPublicApi` whereas SdkClientOption is `@SdkProtectedApi`! So where something should go still feels a bit confusing, but atleast I know I'll err on the side of not making it public unless I'm sure it needs to be, so I have the answer to my _TODO_ question - will make these all clientOption (not advanced).

services/acm/src/main/java/software/amazon/awssdk/services/acm/auth/scheme/internal/AcmAuthSchemeInterceptor.java
gosar: Like the idea, but AUTH_SCHEME_RESOLVER execution attribute needs to be set by service agnostic AwsExecutionContextBuilder, for the AuthSchemeInterceptor to read, which is needing the cast here. But I'm changing the exeuctionAttribute to be SdkInternalExecutionAttribute and is only set by reading the clientConfig, which will be set via a setter in the generated service specific client builder with {Service}AuthSchemeProvider, so the cast is safe?

This is the same pattern used for EndpointProvider.

core/sdk-core/src/main/java/software/amazon/awssdk/core/client/builder/SdkDefaultClientBuilder.java
```diff
@@ -319,6 +324,8 @@ private SdkClientConfiguration finalizeConfiguration(SdkClientConfiguration conf
                      .option(EXECUTION_INTERCEPTORS, resolveExecutionInterceptors(config))
                      .option(RETRY_POLICY, retryPolicy)
                      .option(CLIENT_USER_AGENT, resolveClientUserAgent(config, retryPolicy))
+                     // review TODO: is finalizeConfiguration the right place for this? is this considered "global configuration"
```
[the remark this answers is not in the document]
gosar: I'm understanding how all the config setup works so much better now ;) 
Updating based on your comments here and internal discussion.

core/sdk-core/src/main/java/software/amazon/awssdk/core/client/builder/SdkDefaultClientBuilder.java
```diff
@@ -322,6 +327,24 @@ private SdkClientConfiguration finalizeConfiguration(SdkClientConfiguration conf
...
+        // Moreover, putIdentityProvider will throw NPE on null, so adding the null check here. Also, validateClientOptions
+        // currently asserts it is not null, which will have to change when we allow unsetting default identity provider.
+        if (identityProvider == null) {
+            return config.option(IDENTITY_PROVIDER_CONFIGURATION);
+        }
+        return config.option(IDENTITY_PROVIDER_CONFIGURATION)
+                     .toBuilder()
+                     .putIdentityProvider(identityProvider)
+                     .build();
+    }
```
millems, line 346: This cross-module API feels like it's going to be a pain to maintain, long-term. Can we just do the null check where we make the call, instead?

services/acm/src/main/java/software/amazon/awssdk/services/acm/auth/scheme/internal/AcmAuthSchemeInterceptor.java
millems: Okay, hopefully we can clean it up with the smithy migration when things get pushed down into the clients.

core/sdk-core/src/main/java/software/amazon/awssdk/core/interceptor/SdkInternalExecutionAttribute.java
```diff
@@ -91,6 +96,26 @@ public final class SdkInternalExecutionAttribute extends SdkExecutionAttribute {
      */
     public static final ExecutionAttribute<Boolean> IS_DISCOVERED_ENDPOINT =
         new ExecutionAttribute<>("IsDiscoveredEndpoint");
+    /**
```
millems, line 99: nit: space before /** (here and below)

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/SigningStage.java
```diff
@@ -52,9 +62,53 @@ public SigningStage(HttpClientDependencies dependencies) {
     @Override
     public SdkHttpFullRequest execute(SdkHttpFullRequest request, RequestExecutionContext context) throws Exception {
         InterruptMonitor.checkInterrupted();
+        if (context.executionAttributes().getAttribute(SdkInternalExecutionAttribute.SELECTED_AUTH_SCHEME) != null) {
+            return sraSignRequest(request, context);
+        }
```
millems, line 67: Good working realizing that this won't always be set.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/SigningStage.java
millems: +1 when the request pipeline is rewritten I would love to do away with the SdkHttpFullRequest.

millems (approved):

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/SigningStage.java
gosar: hehe, maybe you realize more than I did. Once we get the AuthSchemeInterceptor code generated it should be non-null always, right? I did it conditional to leave the old logic in there as fallback, for potentially considering if we release these changes slowly across services and having this check in any case wouldn't hurt. But maybe you see another reason why this won't always be set?

core/sdk-core/src/main/java/software/amazon/awssdk/core/interceptor/SdkInternalExecutionAttribute.java
gosar: it was using IDE to move the constant that didn't add the new line. fixed.

core/sdk-core/src/main/java/software/amazon/awssdk/core/client/builder/SdkDefaultClientBuilder.java
gosar: Move the null check to call site in https://github.com/aws/aws-sdk-java-v2/pull/4111/commits/5b0652e8da4d6fbc4abff7a150483b7568d2832c

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/SigningStage.java
millems: Old service clients + new core is a supported use-case. I hadn't thought of it either until I saw your brilliant foresight here.

We can phase this out as we phase out older protected APIs, e.g. with the smithy codegen migration where we're going to be forced to break protected APIs.

gosar: ah, of course! to be honest, that case wasn't explicitly on my mind :D 

In either case, I have a mental note for the testing plan for SRA needs to include testing old clients with new core.

core/sdk-core/src/main/java/software/amazon/awssdk/core/client/builder/SdkDefaultClientBuilder.java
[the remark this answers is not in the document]
gosar: Changed in https://github.com/aws/aws-sdk-java-v2/pull/4111/commits/b9b1adbce663fcd7089fa73749a02121e18ffc75

core/sdk-core/src/main/java/software/amazon/awssdk/core/client/builder/SdkDefaultClientBuilder.java
gosar: Moved the full logic to call site in https://github.com/aws/aws-sdk-java-v2/pull/4111/commits/5ae9bff1853b993474f2b48e0e9d435a94b7326a

millems (approved):

gosar:
This PR was built on top pf #4090 which I squash merged. Then changed the base branch of this PR to feature/master/sra-identity-auth. Then merged feature/master/sra-identity-auth into this branch (resolving some conflicts). Diff looks similar, my integ tests still passed. Will commit after CI passes.

services/acm/src/main/java/software/amazon/awssdk/services/acm/auth/scheme/internal/AcmAuthSchemeInterceptor.java
```diff
@@ -0,0 +1,133 @@
...
+        IdentityProviderConfiguration identityResolvers =
+            executionAttributes.getAttribute(SdkInternalExecutionAttribute.IDENTITY_PROVIDER_CONFIGURATION);
+
+        StringBuilder failureReasons = new StringBuilder();
+
+        // Check each option, in the order the auth scheme resolver proposed them.
+        for (AuthSchemeOption authOption : authOptions) {
+            // If we're using no-auth, don't consider which options are enabled.
+            if (authOption.schemeId().equals("smithy.auth#noAuth")) {
+                return new SelectedAuthScheme(null, null, authOption);
```
sugmanue, line 91: This bit does not work properly, above we require the params to be not null:

```
        this.identityProvider = Validate.paramNotNull(identityProvider, "identityProvider");
        this.signer = Validate.paramNotNull(signer, "signer");
```

gosar: yes, this was fixed in https://github.com/aws/aws-sdk-java-v2/pull/4182#discussion_r1267367040

PR #4131, opened 2023-06-24
Add support for S3 auth scheme resolution

sugmanue:
## Motivation and Context

Auth scheme provider for S3 and S3control. Uses the same auth type derived from the `signatureVersion` property. Also included in this change is support for the no-auth case.

The generated code adds overrides for the default values for the `DOUBLE_URL_ENCODE` and `NORMALIZE_PATH` properties. 

```java
/*
 * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 * 
 * Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance with
 * the License. A copy of the License is located at
 * 
 * http://aws.amazon.com/apache2.0
 * 
 * or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
 * CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions
 * and limitations under the License.
 */

package software.amazon.awssdk.services.s3.auth.scheme.internal;

import java.util.ArrayList;
import java.util.Collections;
import java.util.List;
import software.amazon.awssdk.annotations.Generated;
import software.amazon.awssdk.http.auth.AwsV4HttpSigner;
import software.amazon.awssdk.http.auth.spi.AuthSchemeOption;
import software.amazon.awssdk.services.s3.auth.scheme.S3AuthSchemeParams;
import software.amazon.awssdk.services.s3.auth.scheme.S3AuthSchemeProvider;

@SdkInternalApi
public final class DefaultS3AuthSchemeProvider implements S3AuthSchemeProvider {
    private static final DefaultS3AuthSchemeProvider DEFAULT = new DefaultS3AuthSchemeProvider();

    private DefaultS3AuthSchemeProvider() {
    }

    public static DefaultS3AuthSchemeProvider create() {
        return DEFAULT;
    }

    @Override
    }
}

```

## Types of changes

## Checklist

## License


codegen/src/main/java/software/amazon/awssdk/codegen/poet/auth/scheme/AuthSchemeCodegenMetadata.java
```diff
@@ -69,8 +101,16 @@ public static AuthSchemeCodegenMetadata fromAuthType(AuthType type) {
         switch (type) {
             case BEARER:
                 return BEARER;
-            default:
+            case NONE:
+                return NO_AUTH;
+            case V4:
+            case V4_UNSIGNED_BODY:
```
gosar, line 107: I think this case may need more changes, but doesn't block this PR.

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/auth/scheme/mini-s3-auth-scheme-default-provider.java
```diff
@@ -0,0 +1,49 @@
...
+import software.amazon.awssdk.annotations.Generated;
+import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.http.auth.AwsV4HttpSigner;
+import software.amazon.awssdk.http.auth.spi.AuthSchemeOption;
+import software.amazon.awssdk.services.minis3.auth.scheme.MiniS3AuthSchemeParams;
+import software.amazon.awssdk.services.minis3.auth.scheme.MiniS3AuthSchemeProvider;
+
+@Generated("software.amazon.awssdk:codegen")
+@SdkInternalApi
+public final class DefaultMiniS3AuthSchemeProvider implements MiniS3AuthSchemeProvider {
```
gosar, line 29: S3's default auth scheme provider has more complexity where it delegates to S3's endpoint resolver first and then fallsback to the resolveAuthScheme logic currently in this class. Were you planning on handling that in separate PR?

codegen/src/main/java/software/amazon/awssdk/codegen/poet/auth/scheme/AuthSchemeCodegenMetadata.java
sugmanue: Sure, can you please elaborate on which other changes are needed?

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/auth/scheme/mini-s3-auth-scheme-default-provider.java
sugmanue: Yes, that's correct, this test is only to test this particular case since it also uses the:

```
        "signatureVersion": "s3",
```

which this code uses to decide as `AuthType` to decide which `AuthSchemeCodegenMetadata` to use.

codegen/src/main/java/software/amazon/awssdk/codegen/poet/auth/scheme/AuthSchemeCodegenMetadata.java
gosar: We will handle in separate PR, when corresponding signer solution is determined, which is being discussed offline. No changes needed for this in this PR.

gosar (approved):

dagnir (approved):

PR #4139, opened 2023-06-27
Implement AwsV4EventStreamHttpSigner

haydenbaker:
- do NOT merge this PR until the PR which this is dependent on is merged (https://github.com/aws/aws-sdk-java-v2/pull/4104)

## Motivation and Context
- Adds a default implementation of the AwsV4EventStreamHttpSigner and tests

## Modifications
- Adds default implementation of the signer
- Adds unit tests to verify behavior
- Adjusts visibility of couple of methods in DefaultAwsV4HttpSigner

## Testing
- Ran tests against new signer and utility classes
- Ran tests against existing signer which refactor changes impacts

## Types of changes

## Checklist

## License


core/http-auth-event-stream/pom.xml
```diff
@@ -44,6 +44,36 @@
             <artifactId>http-auth-spi</artifactId>
             <version>${awsjavasdk.version}</version>
         </dependency>
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>http-auth</artifactId>
```
gosar, line 49: will this be replaced or will we add http-auth-aws when the V4 signers are moved to http-auth-aws

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/internal/DefaultAwsV4HttpSigner.java
```diff
@@ -315,4 +320,13 @@ protected void addSignature(SdkHttpRequest.Builder requestBuilder,
 
         requestBuilder.putHeader(SignerConstant.AUTHORIZATION, authHeader);
     }
+
+    /**
+     * Process a request payload ({@link Publisher<ByteBuffer>}).
+     */
+    protected Publisher<ByteBuffer> processPayload(Publisher<ByteBuffer> payload) {
+        // The default implementation does nothing, as this version of signing does not
+        // modify or updated the payload object
```
gosar, line 329: typo: s/updated/update

core/http-auth-event-stream/src/main/java/software/amazon/awssdk/http/auth/eventstream/internal/SigV4DataFramePublisher.java
```diff
@@ -0,0 +1,231 @@
...
+import software.amazon.awssdk.http.auth.internal.util.SignerConstant;
+import software.amazon.awssdk.identity.spi.AwsCredentialsIdentity;
+import software.amazon.awssdk.utils.BinaryUtils;
+import software.amazon.awssdk.utils.Logger;
+import software.amazon.awssdk.utils.internal.MappingSubscriber;
+import software.amazon.eventstream.HeaderValue;
+import software.amazon.eventstream.Message;
+
+@SdkInternalApi
+public final class SigV4DataFramePublisher implements Publisher<ByteBuffer> {
```
gosar, line 45: not `public`?

core/http-auth-event-stream/src/main/java/software/amazon/awssdk/http/auth/eventstream/internal/TrailingDataFramePublisher.java
```diff
@@ -0,0 +1,111 @@
...
+import java.nio.ByteBuffer;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicLong;
+import org.reactivestreams.Publisher;
+import org.reactivestreams.Subscriber;
+import org.reactivestreams.Subscription;
+import software.amazon.awssdk.annotations.SdkInternalApi;
+
+@SdkInternalApi
+public class TrailingDataFramePublisher implements Publisher<ByteBuffer> {
```
gosar, line 27: not `public`?

core/http-auth-event-stream/src/main/java/software/amazon/awssdk/http/auth/eventstream/internal/DefaultAwsV4EventStreamHttpSigner.java
```diff
@@ -15,27 +15,39 @@
...
+    protected void addPrerequisites(SdkHttpRequest.Builder requestBuilder,
+                                    ContentChecksum contentChecksum) {
+        requestBuilder.putHeader(X_AMZ_CONTENT_SHA256, HTTP_CONTENT_SHA_256);
+        super.addPrerequisites(requestBuilder, contentChecksum);
+    }
 
     @Override
-    public SyncSignedRequest sign(SyncSignRequest<? extends AwsCredentialsIdentity> request) {
-        throw new UnsupportedOperationException();
+    protected Publisher<ByteBuffer> processPayload(Publisher<ByteBuffer> payload) {
```
gosar, line 45: based on the parent calling it, `payload` can be null. Maybe add a `if (null) return null`.

core/http-auth-event-stream/src/main/java/software/amazon/awssdk/http/auth/eventstream/internal/DefaultAwsV4EventStreamHttpSigner.java
```diff
@@ -15,27 +15,39 @@
...
+        return new SigV4DataFramePublisher(payload, credentials, credentialScope, signature, signingClock);
     }
 
     @Override
-    public AsyncSignedRequest signAsync(AsyncSignRequest<? extends AwsCredentialsIdentity> request) {
-        throw new UnsupportedOperationException();
+    protected CompletableFuture<String> createContentHash(Publisher<ByteBuffer> payload) {
+        return CompletableFuture.completedFuture(HTTP_CONTENT_SHA_256);
```
gosar, line 51: based on the parent calling, `payload` can be null. Is it ok to return HTTP_CONTENT_SHA_256 even in that case?

core/http-auth-event-stream/src/test/java/software/amazon/awssdk/http/auth/eventstream/AwsV4EventStreamHttpSignerTest.java
```diff
@@ -0,0 +1,112 @@
...
+import software.amazon.eventstream.HeaderValue;
+import software.amazon.eventstream.Message;
+import software.amazon.eventstream.MessageDecoder;
+
+public class AwsV4EventStreamHttpSignerTest {
+
+    private static final AwsV4EventStreamHttpSigner signer = AwsV4EventStreamHttpSigner.create();
+
+    @Test
+    public void asyncSign_shouldSignWithTransformedPublisher() {
```
gosar, line 38: nit: signAsync

core/http-auth-event-stream/pom.xml
gosar, line 1: Is the idea that this module might contain event-stream signers that are not aws specific? I'm wondering as part of moving the sigv4 signers to an aws specific module, how does that affect this module and whether it is considered aws specific (http-auth-aws-event-stream?)

gosar: Would also affect the java package names

haydenbaker: Yeah, good point. We may want to call this module auth-aws-event-stream instead.

core/http-auth-event-stream/src/main/java/software/amazon/awssdk/http/auth/eventstream/internal/DefaultAwsV4EventStreamHttpSigner.java
```diff
@@ -15,27 +15,39 @@
...
-import software.amazon.awssdk.identity.spi.AwsCredentialsIdentity;
+import software.amazon.awssdk.http.auth.internal.DefaultAwsV4HttpSigner;
+import software.amazon.awssdk.http.auth.internal.checksums.ContentChecksum;
 
 /**
  * A default implementation of {@link AwsV4EventStreamHttpSigner}.
  */
 @SdkInternalApi
-public class DefaultAwsV4EventStreamHttpSigner implements AwsV4EventStreamHttpSigner {
+public class DefaultAwsV4EventStreamHttpSigner extends DefaultAwsV4HttpSigner implements AwsV4EventStreamHttpSigner {
```
gosar, line 33: final

core/http-auth-event-stream/src/main/java/software/amazon/awssdk/http/auth/eventstream/internal/DefaultAwsV4EventStreamHttpSigner.java
```diff
@@ -15,27 +15,39 @@
...
+
+    private static final String HTTP_CONTENT_SHA_256 = "STREAMING-AWS4-HMAC-SHA256-EVENTS";
+
+    @Override
+    protected void addPrerequisites(SdkHttpRequest.Builder requestBuilder,
+                                    ContentChecksum contentChecksum) {
+        requestBuilder.putHeader(X_AMZ_CONTENT_SHA256, HTTP_CONTENT_SHA_256);
```
gosar, line 40: So is the expected behavior of this signer's `sign()` method the same as default v4 signer? Is it considered valid to use this signer without async? Is adding this header here the desired behavior when `eventStreamSIgner.sign()` is called?

core/http-auth-event-stream/pom.xml
haydenbaker: We'll add it once they're moved.

core/http-auth-event-stream/src/main/java/software/amazon/awssdk/http/auth/eventstream/internal/DefaultAwsV4EventStreamHttpSigner.java
haydenbaker: That's the current behavior (old implementation), yes.

core/http-auth-event-stream/src/main/java/software/amazon/awssdk/http/auth/eventstream/internal/DefaultAwsV4EventStreamHttpSigner.java
haydenbaker: Yep done.

core/http-auth-event-stream/src/main/java/software/amazon/awssdk/http/auth/eventstream/internal/DefaultAwsV4EventStreamHttpSigner.java
haydenbaker: Good question. I'd imagine so, because the "content hash" is used to indicate certain things to the service about the request. Perhaps someone with deeper knowledge can help answer here?

core/http-auth-event-stream/pom.xml
haydenbaker: I can refactor the name when I do the other module changes.

core/http-auth-event-stream/src/main/java/software/amazon/awssdk/http/auth/eventstream/internal/DefaultAwsV4EventStreamHttpSigner.java
```diff
@@ -15,27 +15,43 @@
...
-import software.amazon.awssdk.identity.spi.AwsCredentialsIdentity;
+import software.amazon.awssdk.http.auth.internal.DefaultAwsV4HttpSigner;
+import software.amazon.awssdk.http.auth.internal.checksums.ContentChecksum;
 
 /**
  * A default implementation of {@link AwsV4EventStreamHttpSigner}.
  */
 @SdkInternalApi
-public class DefaultAwsV4EventStreamHttpSigner implements AwsV4EventStreamHttpSigner {
+public final class DefaultAwsV4EventStreamHttpSigner extends DefaultAwsV4HttpSigner implements AwsV4EventStreamHttpSigner {
```
millems, line 33: > `extends DefaultAwsV4HttpSigner`

We've found that maintaining cross-module APIs with inheritance creates a lot harder-to-follow code than maintaining cross-module APIs with composition. People don't think very hard about the API contracts for inheritance as much as they think about it for APIs. And absolutely everything is an API contract in inherited types that cross module boundaries - no renaming of protected fields, no changing implicit or explicit behavior around protected methods.

Let's keep going with this approach, but I see a lot of internal coupling that might be tricky to support in the future, since this relationship is cross-module. Once the churn on this stuff has died down, let's see if there's a clean abstraction using composition instead of inheritance.

core/http-auth-event-stream/src/main/java/software/amazon/awssdk/http/auth/eventstream/internal/SigV4DataFramePublisher.java
```diff
@@ -0,0 +1,231 @@
...
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.http.auth.eventstream.internal;
+
+import static software.amazon.awssdk.http.auth.internal.util.HttpChecksumUtils.hash;
+import static software.amazon.awssdk.http.auth.internal.util.SignerConstant.AWS4_CHUNK_SIGNING_ALGORITHM;
+import static software.amazon.awssdk.http.auth.internal.util.SignerUtils.computeSignature;
+import static software.amazon.awssdk.http.auth.internal.util.SignerUtils.deriveSigningKey;
```
millems, line 21: These all look like internal APIs to `http-auth`, so they can't be used here.

haydenbaker: Hmmm... that could be troublesome. Perhaps, we should move all of the SigV4 signers into http-auth-aws.

core/http-auth-event-stream/src/main/java/software/amazon/awssdk/http/auth/eventstream/AwsV4EventStreamHttpSigner.java
```diff
@@ -25,7 +26,7 @@
  * specifically for Event Streams.
  */
 @SdkPublicApi
-public interface AwsV4EventStreamHttpSigner extends HttpSigner<AwsCredentialsIdentity> {
+public interface AwsV4EventStreamHttpSigner extends AwsV4HttpSigner {
```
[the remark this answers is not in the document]
haydenbaker: Yep

core/http-auth-event-stream/src/main/java/software/amazon/awssdk/http/auth/eventstream/AwsV4EventStreamAuthScheme.java
```diff
@@ -28,6 +28,14 @@
 @SdkPublicApi
 public interface AwsV4EventStreamAuthScheme extends AuthScheme<AwsCredentialsIdentity> {
```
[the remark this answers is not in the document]
haydenbaker: Makes sense.

core/http-auth-event-stream/src/main/java/software/amazon/awssdk/http/auth/eventstream/internal/DefaultAwsV4EventStreamHttpSigner.java
haydenbaker: I was going for that initially, so I agree. There was feedback about reducing code duplication across signers, which inheritance naturally solves when lots of commonality (but adds coupling obviously). I'd definitely prefer composition.

gosar: I know we are discussing refactoring, but with this current approach for now, this means we have to change the `DefaultAwsV4HttpSigner` to `@SdkProtectedApi`.

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/internal/DefaultAwsV4HttpSigner.java
```diff
@@ -125,10 +128,12 @@ public AsyncSignedRequest signAsync(AsyncSignRequest<? extends AwsCredentialsIde
                     return builder.build();
                 });
 
+        // process the request payload, if necessary
+        Publisher<ByteBuffer> payload = processPayload(request.payload().orElse(null));
 
         return AsyncSignedRequest.builder()
             .request(CompletableFutureUtils.joinLikeSync(signedReqFuture))
-            .payload(request.payload().orElse(null))
+            .payload(payload)
```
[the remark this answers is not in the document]
haydenbaker: `processPayload(ContentStreamProvider payload)` is used by the `AwsS3V4HttpSigner` for chunk-encoding the payload, yes.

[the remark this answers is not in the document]
haydenbaker: WRT renaming, isn't the fact that it's processing an "async" payload clear in the arguments of that method? IMO, I much prefer when methods that serve the same purpose (processing a payload) are overloaded and distinguished by their parameters instead. Thoughts?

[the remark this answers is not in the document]
gosar: > WRT renaming, isn't the fact that it's processing an "async" payload clear in the arguments of that method? IMO, I much prefer when methods that serve the same purpose (processing a payload) are overloaded and distinguished by their parameters instead. Thoughts?

duh, of course. what was I thinking 🤦

PR #4151, opened 2023-06-29
Attach crossRegionAccessEnabled as CustomClientParams to a client Builder, also mention of Composers are renamed as decorators

joviegas:
Attach crossRegionAccessEnabled as CustomClientParams to a client Builder, also mention of Composers are renamed as decorators

## Motivation and Context

```java
S3Client.builder().crossRegionAccessEnabled(true).build();
```

1. Attach crossRegionAccessEnabled to Client Builder directly.
2. Removed crossRegionAccessEnabled from S3Configuration.
3. Customization to support Custom Client Params in customization.config
4. Rename Composers as Decorators
5. Remove the Type Casting in the Decorators (earlier known as composers)

## Modifications

1. Codegen to support new Client Custom Param context
2. Remove generic type and interface for Composers
3. Replace Composers as decorators
4. Remove crossRegionAccessEnabled API from S3Configuration  

## Testing

1. Junits added .
## License


codegen/src/test/resources/software/amazon/awssdk/codegen/poet/builder/test-composed-sync-default-client-builder.java
```diff
@@ -0,0 +1,130 @@
...
+        ServiceConfiguration.Builder serviceConfigBuilder = ((ServiceConfiguration) config
+            .option(SdkClientOption.SERVICE_CONFIGURATION)).toBuilder();
+        serviceConfigBuilder.profileFile(serviceConfigBuilder.profileFileSupplier() != null ? serviceConfigBuilder
+            .profileFileSupplier() : config.option(SdkClientOption.PROFILE_FILE_SUPPLIER));
+        serviceConfigBuilder.profileName(serviceConfigBuilder.profileName() != null ? serviceConfigBuilder.profileName() : config
+            .option(SdkClientOption.PROFILE_NAME));
+        if (serviceConfigBuilder.crossRegionAccessEnabled() != null) {
+            Validate.validState(
+                clientContextParams.get(JsonClientContextParams.CROSS_REGION_ACCESS_ENABLED) == null,
+                "Cross region access enabled has been configured on both ServiceConfiguration and the client level. Please limit this configuration to one location.");
```
millems, line 75: Is there a reason we support both places? Is it hard to not support it in S3ClientConfiguration?

joviegas: Here's the proofread version of the text:

The reason it is supported in both places is that the existing S3ClientContextParams are available in both locations. To access the values of these context parameters, they are accessed through getters on the S3Configuration.
```java
        if (serviceConfigBuilder.crossRegionAccessEnabled() != null) {
            Validate.validState(Throw Exception in both places);
        } else {
            serviceConfigBuilder.crossRegionAccessEnabled(clientContextParams
                    .get(S3ClientContextParams.CROSS_REGION_ACCESS_ENABLED));
        }
```

What if the user wants to create a Config object solely to set the Client config in that S3 Configuration?"



"Is it difficult to not support it in S3ClientConfiguration?"

"It's not that difficult. We need to pass ContextParams to the decorator along with the service configuration to determine if it's a cross-region request."

"If we do not support this in S3Configuration, we will not be able to group it along with other client settings like useArnRegionEnabled, multiRegionEnabled, etc."

joviegas: Removed we can always add this in future if required

millems (commented):
Can you review the sonar bugs and mark them as a false positive or fix them?

millems (approved):

PR #4154, opened 2023-06-29
Deprecate legacy classes and use new when possible

sugmanue:
## Motivation and Context

* Deprecate all the legacy `RetryPolicy` and adjacent types. 
* Update more places that were still using the legacy types.

## Testing

The same battery of tests that were in place for `RetryPolicy` should now pass with `RetryStrategy`.

## Checklist

## License


core/retries/src/main/java/software/amazon/awssdk/retries/DefaultRetryStrategy.java
```diff
@@ -95,7 +95,7 @@ public static AdaptiveRetryStrategy.Builder adaptiveStrategyBuilder() {
 
     static final class Standard {
         static final int MAX_ATTEMPTS = 3;
-        static final Duration BASE_DELAY = Duration.ofSeconds(1);
+        static final Duration BASE_DELAY = Duration.ofMillis(100);
```
sugmanue, line 98: Bug fix, the value is 100ms [see here](https://github.com/aws/aws-sdk-java-v2/blob/e6e22fd66fd8567968f23aaad304eee3559a893d/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/retry/SdkDefaultRetrySetting.java#L52).

core/aws-core/src/main/java/software/amazon/awssdk/awscore/retry/AwsRetryPolicy.java
```diff
@@ -25,8 +25,11 @@
 
 /**
  * Retry Policy used by clients when communicating with AWS services.
+ *
+ * @deprecated Use instead {@link AwsRetryStrategy}
  */
 @SdkPublicApi
```
joviegas, line 31: Can we remove this @SdkPublicApi since its deprecated i think we can remove this

core/sdk-core/src/main/java/software/amazon/awssdk/core/retry/backoff/BackoffStrategy.java
```diff
@@ -21,8 +21,12 @@
 import software.amazon.awssdk.core.retry.RetryMode;
 import software.amazon.awssdk.core.retry.RetryPolicyContext;
 
+/**
+ * @deprecated Use instead {@link software.amazon.awssdk.retries.api.BackoffStrategy}
+ */
 @SdkPublicApi
```
joviegas, line 27: SdkPublicApi can be removed if its deprecated. Generic comment for other classes

core/sdk-core/src/main/java/software/amazon/awssdk/core/retry/backoff/FixedDelayBackoffStrategy.java
```diff
@@ -24,6 +24,9 @@
 
 /**
  * Simple backoff strategy that always uses a fixed delay for the delay before the next retry attempt.
+ *
+ * @deprecated Use instead {@link software.amazon.awssdk.retries.api.BackoffStrategy} and
+ * {@link software.amazon.awssdk.retries.api.BackoffStrategy#fixedDelay(Duration)}.
  */
 @SdkPublicApi
```
joviegas, line 31: Why dont we habe @Depreacted annotation for this ?

core/sdk-core/src/test/java/software/amazon/awssdk/core/internal/http/timers/HttpClientApiCallTimeoutTest.java
```diff
@@ -60,7 +60,6 @@ public class HttpClientApiCallTimeoutTest {
     @Before
     public void setup() {
         httpClient = testClientBuilder()
-            .retryPolicy(RetryPolicy.none())
```
joviegas, line 63: Why we need to remove this now  ?

core/aws-core/src/main/java/software/amazon/awssdk/awscore/retry/AwsRetryPolicy.java
sugmanue: I'm not sure that we can remove it, if I do the build will bark with:

```
SDK annotation is missing on this class. eg: @SdkProtectedApi [MissingSdkAnnotation]
```

If we do, which one should we use instead?

core/sdk-core/src/main/java/software/amazon/awssdk/core/retry/backoff/BackoffStrategy.java
sugmanue: See above, the build **requires** each class to have one of the API scope annotations. If we remove this one, which one should we use?

core/sdk-core/src/main/java/software/amazon/awssdk/core/retry/backoff/FixedDelayBackoffStrategy.java
sugmanue: I missed this one, I will add it, good catch, thanks!

core/sdk-core/src/test/java/software/amazon/awssdk/core/internal/http/timers/HttpClientApiCallTimeoutTest.java
sugmanue: Well, the class is deprecated so I figured we should also try to remove its uses, technically we don't have to do it now.

joviegas: hmm... I might be over thinking but I was thinking of a case where due to some reason we want to modify older RetryPolicy  (since they are not removed but just deprecated) , we might loose test cases for them

core/aws-core/src/main/java/software/amazon/awssdk/awscore/retry/AwsRetryPolicy.java
joviegas: Deprecated shouldn't change the public/protected/internalness
Confirmed this with the team.
Thanks for the patience while I was waiting to get clarification on this.

joviegas (approved):

core/sdk-core/src/test/java/software/amazon/awssdk/core/internal/http/timers/HttpClientApiCallTimeoutTest.java
sugmanue: Sorry, I missed this comment, yeah, I think that's a fair concern, I will bring this up during review.

PR #4165, opened 2023-07-06
[WIP] AWS SDK Cross Version Compatability: env-var `http_proxy` & `https_proxy` support

### LINKED ISSUE #2958: Use environment variables to configure HTTP proxies
### Describe the feature

The V1 Java SDK and others use the environment variables `PROXY_HTTP` and `PROXY_HTTPs` to configure a proxy. The v2 SDK only allows system properties.

There was already [request to add this](https://github.com/aws/aws-sdk-java-v2/issues/1793), but it was incorrectly closed by the [PR that supported system properties](https://github.com/aws/aws-sdk-java-v2/pull/2771) (and I see that there's an [outstanding issue](https://github.com/aws/aws-sdk-java-v2/issues/1334) that _would_ have been solved by that PR, so perhaps the wrong issue was closed?)

### Is your Feature Request related to a problem?

This means that you can't use a consistent deployment approach for deploying AWS applications in a polyglot environment. Or you have to write explicit code in your Java v2 SDK projects to retrieve the relevant environment variables and set system properties, _before anything else happens in the program._

### Proposed Solution

Update `ProxyConfiguration.java` for all three HTTP client implementations to look for environment variables as well as system properties.

### Describe alternatives you've considered

Not using the v2 SDK is the simplest.

Otherwise, as mentioned above, writing explicit code to set the system properties from the value of the environment variables (because Lambda doesn't let you set system properties when the JVM is first run).

### Acknowledge

- [ ] I may be able to implement this feature request

### AWS Java SDK version used

2.17.101

### JDK version used

openjdk version "11.0.13" 2021-10-19

Ubuntu 20.04

Mythra:
## Motivation and Context

***note: I'm opening this early because as far as I can tell it requires changing the public API, but I'm opening to see if others have ideas, or how we can handle this.***

fixes #2958 

The goal of this is to allow like many other SDKs (including aws-sdk-java v1) support for specifying proxies through environment variables. Using `http_proxy` & `https_proxy` for environment variables. `http_proxy` will be used for connections to HTTP sites, and `https_proxy` for HTTPS sites.

While making this change, I also noticed a bug in how we were selecting the `http.proxyHost` & `https.proxyHost` System Properties. I've gotten them to follow: https://docs.oracle.com/javase/8/docs/technotes/guides/net/proxies.html correctly now. Before we weren't passing the scheme in and only selecting based on explicit configuration (which you can still do, and doesn't break)!

I've changed all the HTTP clients that are in the source tree.

## Modifications

- Every single builder now has support for allowing reading from environment variables (by default this is true like system properties), and is not overriding of anything that exists (not breaking, and also like other SDKs).
  - There is now a way to set whether you want an explicitly configured process to handle HTTP or not HTTPS. By default explicit overrides everything like always. This allows it to play with system properties and environment variables too.
- Every proxy configuration type now takes in the _scheme_ it's connecting too, to change between `http` & `https` proxies without manual configuration, or multiple clients at all.

## Testing

- [x] I've been running tests locally to confirm unit tests continue to pass.
- [ ] When I take this out of WIP I will run integration tests through a proxy to confirm it doesn't break anything.

## Types of changes

- [x] This is also a breaking change -- i cannot see a way to avoid it

## Checklist

## License

### DIFF
--- a/utils/src/main/java/software/amazon/awssdk/utils/http/SdkHttpUtils.java
+++ b/utils/src/main/java/software/amazon/awssdk/utils/http/SdkHttpUtils.java
@@ -22,7 +22,9 @@ import static software.amazon.awssdk.utils.FunctionalUtils.invokeSafely;
 import static software.amazon.awssdk.utils.StringUtils.isEmpty;
 
 import java.io.UnsupportedEncodingException;
+import java.net.MalformedURLException;
 import java.net.URI;
+import java.net.URL;
 import java.net.URLDecoder;
 import java.net.URLEncoder;
 import java.util.ArrayList;
@@ -33,12 +35,14 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.function.UnaryOperator;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import software.amazon.awssdk.annotations.SdkProtectedApi;
+import software.amazon.awssdk.utils.ProxyEnvironmentSetting;
 import software.amazon.awssdk.utils.ProxySystemSetting;
 import software.amazon.awssdk.utils.StringUtils;
 import software.amazon.awssdk.utils.Validate;
@@ -49,6 +53,8 @@ import software.amazon.awssdk.utils.Validate;
 @SdkProtectedApi
 public final class SdkHttpUtils {
     private static final String DEFAULT_ENCODING = "UTF-8";
+    private static final String HTTPS = "https";
+    private static final String HTTP = "http";
 
     /**
      * Characters that we need to fix up after URLEncoder.encode().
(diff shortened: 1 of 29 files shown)


utils/src/main/java/software/amazon/awssdk/utils/http/SdkHttpUtils.java
```diff
@@ -440,5 +446,72 @@ public static Set<String> parseNonProxyHostsProperty() {
...
+            return Optional.empty();
+        }
+        String trimmed = parts[0].trim();
+        if (trimmed.isEmpty()) {
+            return Optional.empty();
+        }
+        return Optional.of(trimmed);
+    }
+
+    public static Optional<String> parsePasswordFromUrl(URL value) {
```
piggjf, line 492: Should this handle passwords that are URLEncoded with special characters?  Maybe after grabbing it out of the parts array, pass it through urlDecode?

debora-ito:
Picking this up after a long time, I'm so sorry about that. 

A lot has changed since 2023, and `http_proxy` & `https_proxy` environment variables are supported now, our Developer Guide describes which client supports which variable [here](https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/http-config-proxy-support.html).

PR #4167, opened 2023-07-07
Allow all endpoint params if enabled and allow list is not defined

sugmanue:
## Motivation and Context

Change the logic about which endpoint params to include if configured. The configuration has to keys, one `"enableEndpointAuthSchemeParams": true` that enables the use of endpoint params to be used for auth scheme resolution and the second `"allowedEndpointAuthSchemeParams": […]` which is an allow list for the endpoint params to be used. 

Before this change, only the params explicitly listed in the allow list were generated in the `*AuthSchemeParams` classes. After this change if `enableEndpointAuthSchemeParams` is set to `true` and no allow list is defined, then **all** the endpoint params are included.

This change also changes the name of the `InternalModeledAuthSchemeProvider` to remove the redundant prefix `Internal`.

## Testing

* Added unit tests

## Types of changes

## Checklist

## License


gosar (approved):

codegen/src/main/java/software/amazon/awssdk/codegen/poet/auth/scheme/AuthSchemeSpecUtils.java
```diff
@@ -102,7 +113,12 @@ public boolean generateEndpointBasedParams() {
     }
 
     public boolean includeParam(String name) {
-        return allowedEndpointAuthSchemeParams.contains(name);
+        if (allowedEndpointAuthSchemeParamsConfigured) {
+            return allowedEndpointAuthSchemeParams.contains(name);
+        }
+        // If no explicit allowed endpoint auth scheme params are configured then by default we include all of them except the
+        // ones already defined by default.
+        return !DEFAULT_AUTH_SCHEME_PARAMS.contains(name.toLowerCase(Locale.US));
```
gosar, line 121: I was playing around with this PR for generating S3 and noticed a bug. DefaultS3AuthSchemeProvider.resolveAuthScheme() when building endpointParameters, is missing adding the `region` parameter because of this logic.

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/auth/scheme/query-endpoint-auth-params-auth-scheme-endpoint-provider-without-allowlist.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    @Override
+    public List<AuthSchemeOption> resolveAuthScheme(QueryAuthSchemeParams params) {
+        QueryEndpointParams endpointParameters = QueryEndpointParams.builder()
+                .useDualStackEndpoint(params.useDualStackEndpoint()).useFipsEndpoint(params.useFipsEndpoint())
+                .endpointId(params.endpointId()).defaultTrueParam(params.defaultTrueParam())
+                .defaultStringParam(params.defaultStringParam()).deprecatedParam(params.deprecatedParam())
+                .booleanContextParam(params.booleanContextParam()).stringContextParam(params.stringContextParam())
+                .operationContextParam(params.operationContextParam()).build();
```
gosar, line 58: I mean `region` is missing here

codegen/src/main/java/software/amazon/awssdk/codegen/poet/auth/scheme/AuthSchemeSpecUtils.java
sugmanue: Good catch, we need separated logic for when to include it for the provider case, I will push a commit to fix this, thanks.

sugmanue: Pushed an update.

millems (approved):

gosar (approved):

PR #4171, opened 2023-07-10
Enable request-level token provider override

chenying-wang:
## Motivation and Context

Currently, AWS SDK for Java 2.0 clients don't support override `SdkTokenProvider` in the request level for the bearer token authorization operations. If using different `SdkTokenProvider`s, it will be required to create one instances of the client per `SdkTokenProvider`, which will undermine the runtime performance.

## Modifications

This change enables request-level `SdkTokenProvider` override.

## Testing

The Unit test is added to cover the change.

## Types of changes

## Checklist

## License

### DIFF
--- a/.changes/next-release/feature-AWSSDKforJavav2-bc8f1f1.json
+++ b/.changes/next-release/feature-AWSSDKforJavav2-bc8f1f1.json
@@ -1,0 +1,6 @@
+{
+    "category": "AWS SDK for Java v2", 
+    "contributor": "chenying-wang", 
+    "type": "feature", 
+    "description": "Enable request-level token provider override"
+}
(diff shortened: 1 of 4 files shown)


debora-ito:
@chenying-wang can you tell us more about your use case?
Are you implementing a custom TokenProvider?

chenying-wang:
@debora-ito

No, we are not implementing a custom `SdkTokenProvider` and we are just using the `StaticTokenProvider`.

We have built the backend service that will vend the custom bearer token to the clients and also authenticate the clients based on the bearer token. We will have a client service using the AWS SDK for Java 2.0 client to interact with the said service with the bearer token authentication. Now, each time the client service will have to create a new instance of the SDK service client for each bearer token, and the performance is not ideal.

Therefore, we want to the SDK client to have capability to override the bearer token at request level so that we don't have to create multiple SDK client instances.

Hope this makes sense.

PR #4144, opened 2023-06-28
Skip signing for smithy.api#noAuth

gosar:
## Motivation and Context

The SRA logic to sign should be skipped when the selected auth scheme after auth resolution is the schemeId that represents no authentication.

## Modifications

Add condition to check for smithy.api#noAuth

PR #4175, opened 2023-07-11
Bugfix and other small changes to SRA SigningStage

gosar:
## Motivation and Context

Fix bug introduced in https://github.com/aws/aws-sdk-java-v2/pull/4144 and other small changes to SRA SigningStage

## Modifications

* Fixed bug missing `!` in shouldSign
* Wrap with updateInterceptorContext
* Added metric for signing duration

## Testing

Double checked that AwsCertficateManagerIntegrationTest succeeds.

### DIFF
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/SigningStage.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/SigningStage.java
@@ -161,7 +161,7 @@ public class SigningStage implements RequestToRequestPipeline {
      */
     private boolean shouldSign(SelectedAuthScheme<?> selectedAuthScheme) {
         // TODO: Should this string be a constant somewhere. Similar logic is used in AuthSchemeInterceptors.
-        return "smithy.api#noAuth".equals(selectedAuthScheme.authSchemeOption().schemeId());
+        return !"smithy.api#noAuth".equals(selectedAuthScheme.authSchemeOption().schemeId());
     }
 
     /**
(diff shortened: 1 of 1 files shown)


core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/SigningStage.java
```diff
@@ -132,6 +141,8 @@ private SdkHttpFullRequest signRequest(SdkHttpFullRequest request, RequestExecut
 
             SdkHttpFullRequest signedRequest = measuredSign.left();
 
+            // TODO: This case does not apply to SigningStage as event stream operations are not supported by SyncClients that
```
cenedhryn, line 144: Why don't we remove it?

gosar: I was leaving old code as-is for sometime, until more testing is done. I might remove it before release.

cenedhryn: Are there any particular test cases that would be written later that we wouldn't remember to write if this were to be removed? I'd think that the tests would show if this piece of code was missing. It seems there are a lot of TODOs still to resolve before release... but if we're planning an activity to just go through all those and resolve them after general code complete, that should work too.

cenedhryn (approved):

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/SigningStage.java
gosar: Yeah, going through all TODOs is in my delivery plan for after general code complete.

PR #4176, opened 2023-07-11
Handle sign/signAsync in AsyncSigningStage

gosar:
## Motivation and Context

Add SRA signing logic to AsyncSigningStage. SigingStage logic was added in [#4111](https://github.com/aws/aws-sdk-java-v2/pull/4111/files#diff-d42322e3fc21746722508844b8868872976da81f4db212d0f60212d36c37b395)

## Modifications

* Call sign v/s signAsync appropriately

## Testing

Manual testing:
* Used Glacier's uploadArchive as example API that has a streaming API. So the AsyncClient would use AsyncRequestBody to test signAsync. Similar to https://github.com/aws/aws-sdk-java-v2/pull/4111, hand edited the code generated Glacier files and added a GlacierAuthSchemeInterceptor. Wrote an integ test based on https://github.com/awsdocs/aws-doc-sdk-examples/blob/main/javav2/example_code/glacier/src/main/java/com/example/glacier/UploadArchive.java and verified via debugger that it uses SRA logic and call succeeds.
* Another test case would be event streaming, but will test that later once event streaming signer is done.

### DIFF
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/AsyncSigningStage.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/AsyncSigningStage.java
@@ -18,19 +18,34 @@ package software.amazon.awssdk.core.internal.http.pipeline.stages;
 import java.time.Duration;
 import java.util.concurrent.CompletableFuture;
 import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.core.SelectedAuthScheme;
 import software.amazon.awssdk.core.async.AsyncRequestBody;
 import software.amazon.awssdk.core.http.ExecutionContext;
 import software.amazon.awssdk.core.interceptor.ExecutionAttributes;
 import software.amazon.awssdk.core.interceptor.SdkExecutionAttribute;
+import software.amazon.awssdk.core.interceptor.SdkInternalExecutionAttribute;
 import software.amazon.awssdk.core.internal.http.HttpClientDependencies;
 import software.amazon.awssdk.core.internal.http.RequestExecutionContext;
 import software.amazon.awssdk.core.internal.http.pipeline.RequestPipeline;
+import software.amazon.awssdk.core.internal.util.MetricUtils;
 import software.amazon.awssdk.core.metrics.CoreMetric;
 import software.amazon.awssdk.core.signer.AsyncRequestBodySigner;
 import software.amazon.awssdk.core.signer.AsyncSigner;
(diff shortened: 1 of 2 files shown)


core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/AsyncSigningStage.java
```diff
@@ -48,9 +63,100 @@ public AsyncSigningStage(HttpClientDependencies dependencies) {
...
+                ResolveIdentityRequest.Builder identityRequestBuilder = ResolveIdentityRequest.builder();
+                authSchemeOption.forEachIdentityProperty(identityRequestBuilder::putProperty);
+
+                IdentityProvider<T> identityProvider = selectedAuthScheme.identityProvider();
+                T identity = identityProvider.resolveIdentity(identityRequestBuilder.build()).join();
+
+                HttpSigner<T> signer = selectedAuthScheme.signer();
+
+                SdkHttpFullRequest result;
+                if (context.requestProvider() == null) {
```
millems, line 100: It's weird we use content stream providers in the async request pipeline. Not your problem, just feels weird we're not consistent about it.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/AsyncSigningStage.java
```diff
@@ -48,9 +63,100 @@ public AsyncSigningStage(HttpClientDependencies dependencies) {
...
+    private SdkHttpFullRequest.Builder toSdkHttpFullRequestBuilder(SignedRequest<?> signedRequest) {
+        SdkHttpRequest request = signedRequest.request();
+        return SdkHttpFullRequest.builder()
+                                 .protocol(request.protocol())
+                                 .method(request.method())
+                                 .host(request.host())
+                                 .port(request.port())
+                                 .encodedPath(request.encodedPath())
+                                 .applyMutation(r -> request.forEachHeader(r::putHeader))
+                                 .applyMutation(r -> request.forEachRawQueryParameter(r::putRawQueryParameter));
```
millems, line 157: This feels inefficient to copy everything. I suppose we have to do this for compatibility with the pipeline, but we should probably move away from it in future refactors.

Can we check to see if signedRequest.request() is already an SdkHttpFullRequest and cast it, if so? I think it should always be true.

gosar: +1

Good call - added the check to return immediately if already SdkHttpFullRequest

millems (approved):

PR #4177, opened 2023-07-11
Iterate SdkFields to convert requests

zoewangg:
Instead of setting each field manually, this change updates RequestConvertionUtils class to iterate SdkFields and set common fields automatically

### DIFF
--- a/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/RequestConversionUtils.java
+++ b/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/RequestConversionUtils.java
@@ -15,12 +15,21 @@
 
 package software.amazon.awssdk.services.s3.internal.crt;
 
+import java.util.Arrays;
+import java.util.HashMap;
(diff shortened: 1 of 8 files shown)


services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/RequestConversionUtils.java
```diff
@@ -33,130 +42,78 @@
 /**
  * Request conversion utility method for POJO classes associated with multipart feature.
  */
-//TODO: iterate over SDK fields to get the data
 @SdkInternalApi
 public final class RequestConversionUtils {
```
cenedhryn, line 46: It's not just a request conversion utility. Can we find a more inclusive name?

zoewangg: Renamed it to `SdkPojoConversionUtils`

cenedhryn (approved):

PR #4180, opened 2023-07-12
Allow null in SelectedAuthScheme

gosar:
## Motivation and Context

The auth scheme interceptors have a use case for `null` IdentityProvider and Signer when schemeId is `smithy.api#noAuth`.

## Modifications

Remove non-null validation for these.
Also fixed `Raw use of parameterized class 'SelectedAuthScheme'` warning with `<>`.


gosar:
This change in now part of https://github.com/aws/aws-sdk-java-v2/pull/4182

PR #4182, opened 2023-07-12
Move identity resolution out of SigningStage

gosar:
## Motivation and Context

Identity resolution needs to move to before endpoint resolution to support accountId based endpoints.

## Modifications

Added identity resolution to the AuthSchemeInterceptor itself and saving it (CompletableFuture) in SelectedAuthScheme.

The SigningStage uses the future. AsyncSigningStage changes for this will be made after https://github.com/aws/aws-sdk-java-v2/pull/4176.

A future change for accountId based endpoints will complete this future earlier than signing, but it would be conditional only when IdentityType is AwsCredentialsIdentity. So SigningStage will still join on the future, instead of potentially receiving the completed Identity directly.

## Testing

Integ tested acm and codecatalyst to confirm they still work.
Updated codegen tests.


core/sdk-core/src/main/java/software/amazon/awssdk/core/interceptor/SdkInternalExecutionAttribute.java
```diff
@@ -110,17 +113,22 @@ public final class SdkInternalExecutionAttribute extends SdkExecutionAttribute {
...
+     * The {@link IdentityProviderConfiguration} for a request.
      */
     public static final ExecutionAttribute<IdentityProviderConfiguration> IDENTITY_PROVIDER_CONFIGURATION =
         new ExecutionAttribute<>("IdentityProviderConfiguration");
 
     /**
      * The selected auth scheme for a request.
      */
-    public static final ExecutionAttribute<SelectedAuthScheme> SELECTED_AUTH_SCHEME = new ExecutionAttribute<>(
+    public static final ExecutionAttribute<SelectedAuthScheme<?>> SELECTED_AUTH_SCHEME = new ExecutionAttribute<>(
```
gosar, line 124: Had missed this warning earlier.

core/sdk-core/src/main/java/software/amazon/awssdk/core/interceptor/SdkInternalExecutionAttribute.java
```diff
@@ -110,17 +113,22 @@ public final class SdkInternalExecutionAttribute extends SdkExecutionAttribute {
         "AuthSchemes");
 
     /**
-     * The auth schemes available for a request.
+     * The {@link IdentityProviderConfiguration} for a request.
```
gosar, line 116: fixed copy-paste error from before.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/SigningStage.java
```diff
@@ -74,7 +74,7 @@ private <T extends Identity> SdkHttpFullRequest sraSignRequest(SdkHttpFullReques
 
         ExecutionAttributes executionAttributes = context.executionAttributes();
         SelectedAuthScheme<T> selectedAuthScheme =
-            executionAttributes.getAttribute(SdkInternalExecutionAttribute.SELECTED_AUTH_SCHEME);
+            (SelectedAuthScheme<T>) executionAttributes.getAttribute(SdkInternalExecutionAttribute.SELECTED_AUTH_SCHEME);
```
gosar, line 77: There are 2 casts here that I need help to find alternatives too.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/SigningStage.java
```diff
@@ -84,16 +84,11 @@ private <T extends Identity> SdkHttpFullRequest sraSignRequest(SdkHttpFullReques
...
-                authSchemeOption.forEachIdentityProperty(identityRequestBuilder::putProperty);
-
-                IdentityProvider<T> identityProvider = selectedAuthScheme.identityProvider();
-                T identity = identityProvider.resolveIdentity(identityRequestBuilder.build()).join();
+                CompletableFuture<T> identityFuture =
+                    (CompletableFuture<T>) executionAttributes.getAttribute(SdkInternalExecutionAttribute.IDENTITY);
 
                 SyncSignRequest.Builder<T> signRequestBuilder = SyncSignRequest
-                    .builder(identity)
+                    .builder(CompletableFutureUtils.joinLikeSync(identityFuture))
```
gosar, line 91: is this fine in SigningStage? AsyncSigningStage would do async logic.

cenedhryn: yes

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/SigningStage.java
cenedhryn: If you haven't come up with anything new since our offline discussion, I think the casts are ok.

cenedhryn (approved):

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/SigningStage.java
```diff
@@ -74,28 +74,22 @@ private <T extends Identity> SdkHttpFullRequest sraSignRequest(SdkHttpFullReques
 
         ExecutionAttributes executionAttributes = context.executionAttributes();
         SelectedAuthScheme<T> selectedAuthScheme =
-            executionAttributes.getAttribute(SdkInternalExecutionAttribute.SELECTED_AUTH_SCHEME);
+            (SelectedAuthScheme<T>) executionAttributes.getAttribute(SdkInternalExecutionAttribute.SELECTED_AUTH_SCHEME);
```
millems, line 77: If you pass the selected auth scheme as a separate parameter to the method, no casting is needed because T will be the selected auth scheme's type parameter:

```java
return sraSignRequest(request, context, context.executionContext().getAttribute(SdkInternalExecutionAttribute.SELECTED_AUTH_SCHEME));
```
```java
 private <T extends Identity> SdkHttpFullRequest sraSignRequest(SdkHttpFullRequest request, RequestExecutionContext context, SelectedAuthScheme<T> selectedAuthScheme);
```

You could also do that within the sraSignRequest method (if you want to keep the sraSignRequest signature simple) by delegating to yet another method (doSraSignRequest?). Essentially, you're trying to assign a name (T) to the ? in SelectedAuthScheme so that you can use it to refer to the type the select auth scheme contains. You can do that with any method call.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/SigningStage.java
```diff
@@ -74,28 +74,22 @@ private <T extends Identity> SdkHttpFullRequest sraSignRequest(SdkHttpFullReques
...
-
-                // TODO: Identity resolution should move to before Endpoint resolution interceptor, to support accountId based
-                //  endpoints and also to logically separate out identity resolution as its own step.
-                ResolveIdentityRequest.Builder identityRequestBuilder = ResolveIdentityRequest.builder();
-                authSchemeOption.forEachIdentityProperty(identityRequestBuilder::putProperty);
-
-                IdentityProvider<T> identityProvider = selectedAuthScheme.identityProvider();
-                T identity = identityProvider.resolveIdentity(identityRequestBuilder.build()).join();
+                CompletableFuture<T> identityFuture =
+                    (CompletableFuture<T>) executionAttributes.getAttribute(SdkInternalExecutionAttribute.IDENTITY);
```
millems, line 86: If you store the identity in the SelectedAuthScheme instead of a separate execution attribute, you can ensure the identity type matches the one in the SelectedAuthScheme. That means you won't have to cast it, because it'll already be T when you retrieve it.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/interceptor/ResolveIdentityInterceptor.java
```diff
@@ -0,0 +1,60 @@
...
+
+    private CompletableFuture<? extends Identity> resolveIdentity(SelectedAuthScheme<?> selectedAuthScheme) {
+        ResolveIdentityRequest.Builder identityRequestBuilder = ResolveIdentityRequest.builder();
+
+        AuthSchemeOption authSchemeOption = selectedAuthScheme.authSchemeOption();
+        authSchemeOption.forEachIdentityProperty(identityRequestBuilder::putProperty);
+
+        IdentityProvider<?> identityProvider = selectedAuthScheme.identityProvider();
+        return identityProvider.resolveIdentity(identityRequestBuilder.build());
+    }
```
millems, line 59: Here we know the result of identityProvider.responseIdentity is the same type as the SelectedAuthScheme's type. Assigning it to a separate IDENTITY attribute loses that association, so you'll need to cast it later on.

Is there any way to store this identity in the select auth scheme so the compiler can make sure we don't end up with an identity that doesn't work with the selected auth scheme?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/interceptor/ResolveIdentityInterceptor.java
```diff
@@ -0,0 +1,60 @@
...
+import software.amazon.awssdk.core.interceptor.ExecutionAttributes;
+import software.amazon.awssdk.core.interceptor.ExecutionInterceptor;
+import software.amazon.awssdk.core.interceptor.SdkInternalExecutionAttribute;
+import software.amazon.awssdk.http.auth.spi.AuthSchemeOption;
+import software.amazon.awssdk.identity.spi.Identity;
+import software.amazon.awssdk.identity.spi.IdentityProvider;
+import software.amazon.awssdk.identity.spi.ResolveIdentityRequest;
+
+@SdkInternalApi
+public class ResolveIdentityInterceptor implements ExecutionInterceptor {
```
millems, line 31: Since this is not service-specific, is there a reason to use an interceptor for this, or can we just add it to the main execution pipeline(s)?

gosar: We need this order: auth scheme resolution -> identity resolution -> endpoint resolution. Currently, auth scheme resolution and endpoint resolution are interceptors (and are service specific) that are run in AwsExecutionContextBuilder before the request pipeline, so this felt the most natural. Anything else seems like it would need big refactoring?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/SigningStage.java
gosar: Yeah, the issue was making that T work for the Identity, which we discuss below.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/interceptor/ResolveIdentityInterceptor.java
gosar: Yeah, at one point I considered doing identityProvider.resolveIdentity in AuthSchemeInterceptor's trySelectAuthSchceme and saving the Identity directly in SelectedAuthScheme instead of the IdentityProvider. This would "work" the same, but doesn't give us the separation of concerns of keeping Identity resolution as a separate step.

The other option would be to start the SelectedAuthScheme with null Identity and then mutate (using copy builder) the SelectedAuthScheme to add the Identity after Identity resolution. I thought of avoiding null/mutating objects, so landed on the approach in this PR. But maybe the better tradeoff is to do this - hardens the matching of the types.

millems: Yeah, I think I would just favor resolving the identity when we select the auth scheme. I don't see a reason we would need to separate those two things in the future, and if we did we can always do it by updating the SelectedAuthScheme.

core/sdk-core/src/main/java/software/amazon/awssdk/core/SelectedAuthScheme.java
```diff
@@ -30,20 +30,20 @@
...
+    private CompletableFuture<? extends T> identity;
     private HttpSigner<T> signer;
     private AuthSchemeOption authSchemeOption;
 
-    public SelectedAuthScheme(IdentityProvider<T> identityProvider,
+    public SelectedAuthScheme(CompletableFuture<? extends T> identity,
                               HttpSigner<T> signer,
                               AuthSchemeOption authSchemeOption) {
-        this.identityProvider = Validate.paramNotNull(identityProvider, "identityProvider");
-        this.signer = Validate.paramNotNull(signer, "signer");
```
gosar, line 41: removing the nonNull check (for smithy.api#noAuth use case) was initially separated out in https://github.com/aws/aws-sdk-java-v2/pull/4180, but folded it into this PR now.

millems (approved):

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/interceptor/ResolveIdentityInterceptor.java
gosar: Changed to this approach.

Issue #39, opened 2017-07-03
IAM Policy Builder


millems:
The IAM policy builder includes some hacks and changes, like removing dashes from account IDs, that could be considered unintuitive.

Review the inherited state of the V1 IAM policy builder and determine which changes are necessary for V2.

(Feel free to comment on this issue with desired changes).

millems:
This shouldn't be a high-effort migration, since there's not a lot of bells and whistles that can be added to the 1.11.x version.

jensentanlo:
This would be helpful and preferable to importing the v1 SDK

leifbennett:
I'm trying to get SNS over SQS working, and these classes would be very helpful, as the subscribe operation no longer sets the policy. I'd really prefer not to have to us the v1 SDK.

millems:
https://github.com/aws/aws-sdk-java-v2/pull/4185 created to add this feature

millems:
This feature was released. Documentation in developer guide is in progress, but you can use it based on the docs here: https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/policybuilder/iam/IamPolicy.html

PR #4185, opened 2023-07-13
Added support for IamPolicy in new module software.amazon.awssdk:iam-…

millems:
…policy-builder, a class to simplify the use of AWS policies.

See IamPolicy for usage examples.

Implements #39

### DIFF
--- a/core/annotations/src/main/java/software/amazon/awssdk/annotations/NotNull.java
+++ b/core/annotations/src/main/java/software/amazon/awssdk/annotations/NotNull.java
@@ -1,0 +1,39 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.annotations;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * The annotated element must not be null. Accepts any type.
+ * <p>
+ * This is useful to tell linting and testing tools that a particular value will never be null. It's not meant to be used on
+ * public interfaces as something that customers should rely on.
+ */
+@Documented
+@Target({ElementType.METHOD,
+         ElementType.FIELD,
+         ElementType.ANNOTATION_TYPE,
+         ElementType.CONSTRUCTOR,
+         ElementType.PARAMETER})
+@Retention(RetentionPolicy.CLASS)
+@SdkProtectedApi
+public @interface NotNull {
+}
--- a/core/json-utils/src/main/java/software/amazon/awssdk/protocols/jsoncore/JsonWriter.java
+++ b/core/json-utils/src/main/java/software/amazon/awssdk/protocols/jsoncore/JsonWriter.java
@@ -20,6 +20,7 @@ import static software.amazon.awssdk.utils.DateUtils.formatUnixTimestampInstant;
 
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
+import java.io.OutputStream;
 import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.nio.ByteBuffer;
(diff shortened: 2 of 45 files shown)


core/annotations/src/main/java/software/amazon/awssdk/annotations/NotNull.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+@Documented
+@Target({ElementType.METHOD,
+         ElementType.FIELD,
+         ElementType.ANNOTATION_TYPE,
+         ElementType.CONSTRUCTOR,
+         ElementType.PARAMETER})
+@Retention(RetentionPolicy.CLASS)
+@SdkProtectedApi
+public @interface NotNull {
```
millems, line 38: Added this because EqualsVerifier will use it to detect which fields are considered not-null. It allows me to test all of the equals/hashcode methods in the package at once from a single test (EqualsHashCodeTest).

core/json-utils/src/main/java/software/amazon/awssdk/protocols/jsoncore/JsonWriter.java
```diff
@@ -179,13 +180,31 @@ private Builder() {
...
+         * configuration, like pretty-printing output.
+         *
+         * <p>It's highly recommended to use a shared {@code JsonFactory} within this generator factory, where possible, so they
+         * should be stored statically: http://wiki.fasterxml.com/JacksonBestPracticesPerformance
+         *
+         * <p>By default, this delegates to {@link JsonNodeParser#DEFAULT_JSON_FACTORY} to create the generator.
+         *
+         * <p>Setting this value will also override any values set via {@link #jsonFactory}.
+         */
+        public JsonWriter.Builder jsonGeneratorFactory(JsonGeneratorFactory jsonGeneratorFactory) {
```
millems, line 206: Added this to allow me to configure the underlying JSON generator to pretty-print the JSON.

services-custom/iam-policy-builder/src/main/java/software/amazon/awssdk/policybuilder/iam/internal/DefaultIamPolicyReader.java
```diff
@@ -0,0 +1,212 @@
...
+        if (node == null) {
+            return null;
+        }
+
+        return expectString(node, key);
+    }
+
+    private String expectString(JsonNode node, String name) {
+        Validate.isTrue(node.isString(), "%s was not a string");
+        return node.asString();
```
davidh44, line 205: name missing as argument:
`Validate.isTrue(node.isString(), "%s was not a string", name);`

services-custom/iam-policy-builder/src/main/java/software/amazon/awssdk/policybuilder/iam/IamStatement.java
```diff
@@ -0,0 +1,1187 @@
...
+         * }
+         * @see <a href="https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_notprincipal.html">
+         * NotPrincipal user guide</a>
+         */
+        Builder addNotPrincipals(IamPrincipalType iamPrincipalType, Collection<String> notPrincipals);
+
+        /**
+         * Append multiple 
+         * <a href="https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_notprincipal.html">{@code
+         * NotPrincipal}s</a> to this statement, denying access to the principal that is specified.
```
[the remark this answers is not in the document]
millems: Good call, I was thinking with an ALLOW statement, but that's not possible with NotPrincipal.

services-custom/iam-policy-builder/src/main/java/software/amazon/awssdk/policybuilder/iam/IamStatement.java
```diff
@@ -0,0 +1,1187 @@
...
+         *                 .sid("GrantReadBookContent")
+         *                 .effect(IamEffect.DENY)
+         *                  // This statement denies access to everyone except the books service and operators:
+         *                 .addNotPrincipals("AWS", Arrays.asList("arn:aws:iam::123456789012:role/books-service",
+         *                                                        "arn:aws:iam::123456789012:role/books-operator"))
+         *                 .addAction("s3:GetObject")
+         *                 .addResource("arn:aws:s3:us-west-2:123456789012:accesspoint/book-content/object/*")
+         *                 .build();
+         * }
+         * @see <a href="https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_notprincipal.html">
```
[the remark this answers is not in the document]
millems: IAM will fail. I generally prefer to let the service do the validation since they can change and we can't (unless the customer upgrades their SDK), but I'm open to disagreement on that.

services-custom/iam-policy-builder/src/main/java/software/amazon/awssdk/policybuilder/iam/internal/DefaultIamPolicyReader.java
millems: Good catch!

services-custom/iam-policy-builder/src/main/java/software/amazon/awssdk/policybuilder/iam/IamStatement.java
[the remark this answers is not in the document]
davidh44: yeah if IAM will fail, then dont think we need to throw error. would be good to mention in javadocs though

davidh44 (approved):

PR #4197, opened 2023-07-18
Fix to handle temporary Redirect 307 for S3 Cross Region client

joviegas:
## Motivation and Context

- S3Cross region client with default region created in us-east-1 could not do cross region access

## Modifications

- Allow redirect access on Temporary Redirect 307 status code.

## Testing

- Added Integ and Junit test case

## License


services/s3/src/it/java/software/amazon/awssdk/services/s3/crossregion/S3CrossRegionCrtIntegrationTest.java
```diff
@@ -20,14 +20,21 @@
...
+import software.amazon.awssdk.services.s3.model.ListObjectsRequest;
+import software.amazon.awssdk.services.s3.model.ListObjectsResponse;
 
 public class S3CrossRegionCrtIntegrationTest extends S3CrossRegionAsyncIntegrationTestBase {
     private static final String BUCKET = temporaryBucketName(S3CrossRegionCrtIntegrationTest.class);
 
     @BeforeAll
     static void setUpClass() {
-        s3 = s3ClientBuilder().build();
+        s3 = s3ClientBuilder().region(Region.AP_SOUTH_1).httpClient(ApacheHttpClient.create()).build();
```
zoewangg, line 37: Question: why do we need to set ApacheHttpClient explicitly? If we have to, can we use `httpClientBuilder(ApacheHttpClient.builder())` so that the http client can be closed?

.changes/next-release/bugfix-AWSSDKforJavav2-4620073.json
```diff
@@ -0,0 +1,6 @@
+{
+    "type": "bugfix",
+    "category": "AWS SDK for Java v2",
```
zoewangg, line 3: nit: category should be S3

joviegas: Done

services/s3/src/it/java/software/amazon/awssdk/services/s3/crossregion/S3CrossRegionCrtIntegrationTest.java
joviegas: Added by mistake just for testing , removed it now

PR #4205, opened 2023-07-18
MappingSubscriber should halt upon any Exception

polyatail:
## Motivation and Context
When using `SdkPublisher.map`, if the mapping function throws an exception that does not inherit from `RuntimeException` (which often is the case in Kotlin), anything awaiting a result from the Publisher will hang indefinitely, even though the coroutine has ended exceptionally.

## Modifications
This change catches and terminates on all exceptions.

## Types of changes

## Checklist

## License

### DIFF
--- a/utils/src/main/java/software/amazon/awssdk/utils/internal/MappingSubscriber.java
+++ b/utils/src/main/java/software/amazon/awssdk/utils/internal/MappingSubscriber.java
@@ -69,7 +69,7 @@ public class MappingSubscriber<T, U> implements Subscriber<T> {
         if (!isCancelled) {
             try {
                 delegateSubscriber.onNext(mapFunction.apply(t));
-            } catch (RuntimeException e) {
+            } catch (Exception e) {
                 // If the map function throws an exception, the subscription should be cancelled as the publisher will
                 // otherwise not be aware it has happened and should have the opportunity to clean up resources.
                 cancelSubscriptions();


polyatail:
Yes, Kotlin does have built-in exceptions that do not inherit from `RuntimeException`. Even if it didn't, a user could write a piece of code that throws `Exception` and cause a coroutine to hang indefinitely. This has caused us multiple SEVs at our organization already. Is there a reason to only catch `RuntimeException` here and not `Exception` which would be guaranteed to catch everything?

utils/src/main/java/software/amazon/awssdk/utils/internal/MappingSubscriber.java
```diff
@@ -69,7 +69,7 @@ public void onNext(T t) {
         if (!isCancelled) {
             try {
                 delegateSubscriber.onNext(mapFunction.apply(t));
-            } catch (RuntimeException e) {
+            } catch (Exception e) {
```
zoewangg, line 72: Should we catch Throwable?

polyatail: Why not? I don't think we can catch _too many_ exceptions here.

debora-ito: @polyatail can you update the PR and add Throwable?

kkywalk2:
is there any update on this PR?

utils/src/main/java/software/amazon/awssdk/utils/internal/MappingSubscriber.java
polyatail: Done.

polyatail:
Please merge this.

debora-ito:
@polyatail So so sorry for overlooking this for so long. Picking it up now.

Will activate the codebuilds.

PR #4211, opened 2023-07-19
Move identity resolution out of AsyncSigningStage

gosar:
## Motivation and Context

Similar change as #4182 , but for AsyncSigningStage.

Identity resolution needs to move to before endpoint resolution to support accountId based endpoints.

## Modifications

* Use Identity from SelectedAuthScheme.
* Account for signAsync returning CompletableFuture - signAsync interface hasn't been updated yet, so faking it here with a temporary private method.
* Addressed some TODOs from #4176 related to updating contexts with results.
* Some refactoring of SigningStage and moved the waiting for Identity to complete out of the SIGNING_DURATION metric.

## Testing

Integ tested Glacier's uploadArchive similar to how I tested signAsync initially in #4176.

### DIFF
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/AsyncSigningStage.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/AsyncSigningStage.java
@@ -89,10 +89,6 @@ public class AsyncSigningStage implements RequestPipeline<SdkHttpFullRequest,
             });
 
             return signedRequestFuture.thenApply(r -> {
-                // TODO: the below updates the result in interceptor context, but if signAsync is called the asyncRequestBody
-                //  could also be different. Should it also be updated in InterceptorContext?
-                //  executionContext.interceptorContext(executionContext.interceptorContext().copy(b -> b.asyncRequestBody(...))).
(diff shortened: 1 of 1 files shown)


millems (approved):

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/AsyncSigningStage.java
```diff
@@ -65,76 +61,91 @@ public CompletableFuture<SdkHttpFullRequest> execute(SdkHttpFullRequest request,
...
+
+            CompletableFuture<SdkHttpFullRequest> signedRequestFuture = doSraSign(request, context, selectedAuthScheme, identity);
+
+            signedRequestFuture.whenComplete((r, t) -> {
+                context.attemptMetricCollector().reportMetric(CoreMetric.SIGNING_DURATION,
+                                                              Duration.ofNanos(System.nanoTime() - signingStart));
+            });
 
+            return signedRequestFuture.thenApply(r -> {
                 // TODO: the below updates the result in interceptor context, but if signAsync is called the asyncRequestBody
```
gosar, line 92: forgot to push the commit that removed this TODO. will do so before merging.

PR #4224, opened 2023-07-25
S3 Multipart API implementation

L-Applin:
Adds `MultipartConfiguration` to the `S3AsyncClientBuilder` to enables Multipart operation on S3AsyncClient.

The following methods were added to `S3AsyncClientBuilder` (codegen):

```java
    /**
     * Configuration for multipart operation of this client. Calling this method with a non null argument will enable
     * automatic conversion of Get, Put and Copy requests to their multipart equivalent.
     */
    default S3AsyncClientBuilder multipartConfiguration(MultipartConfiguration multipartConfiguration) {
    }

    /**
     * Configuration for multipart operation of this client. Calling this method with a non null argument will enable
     * automatic conversion of Get, Put and Copy requests to their multipart equivalent.
     */
    default S3AsyncClientBuilder multipartConfiguration(Consumer<Builder> multipartConfiguration) {
        Builder builder = MultipartConfiguration.builder();
        multipartConfiguration.accept(builder);
        return multipartConfiguration(builder.build());
    }
```

The following method has been added to `DefaultS3AsyncClientBuilder` (codegen):

```java
    @Override
    public DefaultS3AsyncClientBuilder multipartConfiguration(MultipartConfiguration multipartConfig) {
        clientContextParams.put(MultipartConfiguration.MULTIPART_CONFIGURATION_KEY, multipartConfig);
        return this;
    }
```


codegen/src/main/java/software/amazon/awssdk/codegen/model/config/customization/CustomizationConfig.java
```diff
@@ -227,6 +227,18 @@ public class CustomizationConfig {
...
+    /**
+     * For S3AsyncClient only. Fully qualified name of the class to be used for multipart configuration
+     */
+    private String multipartConfigurationClass;
+
+    /**
+     * For S3AsyncClient only. Javadoc associated with the generated method on the S3AsyncClientBuilder interface for multipart
+     * configuration.
+     */
+    private String multipartMethodJavadoc;
```
millems, line 239: What are the pros/cons of this compared to just using customClientContextParams or something similar? https://github.com/aws/aws-sdk-java-v2/blob/master/services/s3/src/main/resources/codegen-resources/customization.config#L255-L260

codegen/src/main/java/software/amazon/awssdk/codegen/poet/builder/AsyncClientBuilderInterface.java
```diff
@@ -17,34 +17,83 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import software.amazon.awssdk.awscore.client.builder.AwsAsyncClientBuilder;
 import software.amazon.awssdk.codegen.model.intermediate.IntermediateModel;
 import software.amazon.awssdk.codegen.poet.ClassSpec;
 import software.amazon.awssdk.codegen.poet.PoetUtils;
+import software.amazon.awssdk.utils.StringUtils;
 
 public class AsyncClientBuilderInterface implements ClassSpec {
+    static Logger log = LoggerFactory.getLogger(AsyncClientBuilderInterface.class);
```
millems, line 35: Nit: Use software.amazon.awssdk.utils.Logger

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/multipart/CopyObjectHelper.java
```diff
@@ -128,6 +126,10 @@ private void doCopyInParts(CopyObjectRequest copyObjectRequest,
                                String uploadId) {
 
         long optimalPartSize = genericMultipartHelper.calculateOptimalPartSizeFor(contentLength, partSizeInBytes);
+        if (optimalPartSize > partSizeInBytes) {
+            log.info(() -> String.format("Configured partSize is %d, but using %d to prevent reaching maximum number of parts "
+                                         + "allowed", partSizeInBytes, optimalPartSize));
```
millems, line 131: Info is a pretty loud log level. Is this necessary?

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/multipart/MultipartUploadHelper.java
```diff
@@ -120,6 +120,10 @@ private void doUploadInParts(Pair<PutObjectRequest, AsyncRequestBody> request,
                                  String uploadId) {
 
         long optimalPartSize = genericMultipartHelper.calculateOptimalPartSizeFor(contentLength, partSizeInBytes);
+        if (optimalPartSize > partSizeInBytes) {
+            log.info(() -> String.format("Configured partSize is %d, but using %d to prevent reaching maximum number of parts "
+                                         + "allowed", partSizeInBytes, optimalPartSize));
```
millems, line 125: Same here

services/s3/src/main/java/software/amazon/awssdk/services/s3/multipart/MultipartDownloadType.java
```diff
@@ -0,0 +1,25 @@
...
+package software.amazon.awssdk.services.s3.multipart;
+
+/**
+ *
+ * todo complete when implementing multipart download
+ */
+public enum MultipartDownloadType {
+    PART,
+    RANGE;
+}
```
millems, line 25: What are the pros/cons for letting the customer pick this instead of just having us pick this for them? Permissions makes sense as one thing.

codegen/src/main/java/software/amazon/awssdk/codegen/model/config/customization/CustomizationConfig.java
L-Applin: Custom context param also adds the method to the sync client builder, but we only want multipart for the async client

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/multipart/CopyObjectHelper.java
L-Applin: This is only happens when `total size / part size > 10,000` which will result in not using the configured part size set by the user (to not hit the 10,000 part limit imposed by s3), I thought being a little loud about it was a good idea.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/multipart/MultipartUploadHelper.java
L-Applin: see other comment

.changes/next-release/feature-AWSSDKforJavav2-5d806ad.json
```diff
@@ -0,0 +1,6 @@
+{
+    "type": "feature",
+    "category": "AWS SDK for Java v2",
+    "contributor": "StephenFlavin",
+    "description": "Add \"unsafe\" and \"fromRemaining\" AsyncRequestBody constructors for byte arrays and ByteBuffers"
+}
```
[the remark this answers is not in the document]
L-Applin: had to merge from master to get ClientComposer/DelegateClient related classes

codegen/src/main/java/software/amazon/awssdk/codegen/model/config/customization/CustomizationConfig.java
millems: Seems fine, since this is just an internal API. We can always refactor it later if we wanted to. E.g. make customClientContextParams have an async-only flag.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/multipart/CopyObjectHelper.java
millems: Should we have a flag that the user opts-in to allow this to happen, so that we don't have to warn when it does? (They've consented to it, after all!)

Logging above debug for anything that isn't an error scares me, since we don't know what the user's throughput is like, and we frequently get people upset about us filling their logs.

L-Applin: I see, I can lower to debug then

PR #4226, opened 2023-07-25
Support streaming with unknown content length

zoewangg:
## Motivation and Context
Support streaming with unknown content length

## Modifications
- Support streaming with unknown content length https://github.com/aws/aws-sdk-java-v2/pull/4226/commits/9b661d8be84d037c19aea8a6bae37c123c22d85e

This is implemented by buffering and storing the first `AsyncRequestBody` received from `onNext` and wait for the next signal. If `onComplete` is invoked next, use single object upload, or if a second `AsyncRequestBody` is received, use MPU.

- Refactoring code
Below is the class structure:
Screen Shot 2023-07-25 at 4 09 36 PM

## Testing
Pending, need to add more tests


core/sdk-core/src/main/java/software/amazon/awssdk/core/async/AsyncRequestBody.java
```diff
@@ -420,24 +420,20 @@ static AsyncRequestBody empty() {
      * @param maxMemoryUsageInBytes the max memory the SDK will use to buffer the content
      * @return SplitAsyncRequestBodyResult
      */
-    default SplitAsyncRequestBodyResponse split(long chunkSizeInBytes, long maxMemoryUsageInBytes) {
+    default SdkPublisher<AsyncRequestBody> split(long chunkSizeInBytes, long maxMemoryUsageInBytes) {
```
zoewangg, line 423: I made this change because returning a future along with a publisher means there are two ways to get notifications, which seems a bit weird.

PR #4227, opened 2023-07-25
SignAsync returns Future AsyncSignedRequest

haydenbaker:
## Motivation and Context
- Makes signAsync return a future signed request, such that blocking before returning is not necessary

## Modifications
- Migrate return type to future
- Update call-sites accordingly
- Update tests

## Testing
- Run tests, make sure all pass

## Types of changes

## Checklist

## License


core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/signer/BaseAwsV4HttpSigner.java
```diff
@@ -106,10 +107,11 @@ default AsyncSignedRequest signAsync(AsyncSignRequest<? extends AwsCredentialsId
 
         Publisher<ByteBuffer> payload = processPayload(request.payload().orElse(null), futureV4RequestContext, properties);
 
-        return AsyncSignedRequest.builder()
-            .request(CompletableFutureUtils.joinLikeSync(futureV4RequestContext).getSignedRequest())
+        return futureV4RequestContext.thenApply(ctx -> AsyncSignedRequest.builder()
```
joviegas, line 110: Can we add a test case where we cancel the future captured from signAsync() ; in that case what happens to futureV4RequestContext ?

joviegas (commented):
Have we tested these signer with any of the real services like S3  ?

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/signer/V4HttpSigner.java
```diff
@@ -100,9 +100,9 @@ public AsyncSignedRequest signAsync(AsyncSignRequest<? extends AwsCredentialsIde
 
         Publisher<ByteBuffer> payload = payloadSigner.sign(request.payload().orElse(null), futureV4Context);
```
gosar, line 101: I was going to say this could be moved inside the thenApply below, but see you made that change in https://github.com/aws/aws-sdk-java-v2/pull/4270/files#diff-e7528bac3b3393210083f8bf3ef4b519b2d5754e7771d9aab68d6a8b82848dc4. I guess it could have be brought into this PR, but ok that it comes in the next PR.

gosar (approved):

PR #4230, opened 2023-07-26
Add support for auth scheme interceptor with endpoint params

sugmanue:
## Motivation and Context

Add support for auth scheme params based on endpoints params. For this use the already existing endpoint param class and logic to be populated and then just read the needed values into the auth scheme params.

## Testing

* Added unit tests

## Types of changes

## Checklist

## License

### DIFF
--- a/codegen/src/main/java/software/amazon/awssdk/codegen/poet/auth/scheme/AuthSchemeInterceptorSpec.java
+++ b/codegen/src/main/java/software/amazon/awssdk/codegen/poet/auth/scheme/AuthSchemeInterceptorSpec.java
@@ -33,8 +33,11 @@ import javax.lang.model.element.Modifier;
 import software.amazon.awssdk.annotations.SdkInternalApi;
 import software.amazon.awssdk.awscore.AwsExecutionAttribute;
 import software.amazon.awssdk.codegen.model.intermediate.IntermediateModel;
+import software.amazon.awssdk.codegen.model.rules.endpoints.ParameterModel;
 import software.amazon.awssdk.codegen.poet.ClassSpec;
 import software.amazon.awssdk.codegen.poet.PoetUtils;
+import software.amazon.awssdk.codegen.poet.rules.EndpointRulesSpecUtils;
+import software.amazon.awssdk.core.SdkRequest;
 import software.amazon.awssdk.core.SelectedAuthScheme;
 import software.amazon.awssdk.core.exception.SdkException;
 import software.amazon.awssdk.core.interceptor.Context;
@@ -53,10 +56,14 @@ import software.amazon.awssdk.utils.Logger;
 import software.amazon.awssdk.utils.Validate;
 
 public final class AuthSchemeInterceptorSpec implements ClassSpec {
+    private final IntermediateModel intermediateModel;
     private final AuthSchemeSpecUtils authSchemeSpecUtils;
+    private final EndpointRulesSpecUtils endpointRulesSpecUtils;
 
     public AuthSchemeInterceptorSpec(IntermediateModel intermediateModel) {
+        this.intermediateModel = intermediateModel;
         this.authSchemeSpecUtils = new AuthSchemeSpecUtils(intermediateModel);
+        this.endpointRulesSpecUtils = new EndpointRulesSpecUtils(intermediateModel);
     }
 
     @Override
@@ -78,6 +85,7 @@ public final class AuthSchemeInterceptorSpec implements ClassSpec {
         builder.addMethod(generateBeforeExecution())
                .addMethod(generateResolveAuthOptions())
                .addMethod(generateSelectAuthScheme())
+               .addMethod(generateAuthSchemeParams())
                .addMethod(generateTrySelectAuthScheme());
         return builder.build();
     }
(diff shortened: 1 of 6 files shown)


codegen/src/test/resources/software/amazon/awssdk/codegen/poet/auth/scheme/query-endpoint-auth-params-with-allowlist-auth-scheme-interceptor.java
```diff
@@ -0,0 +1,130 @@
...
+        }
+        throw SdkException
+                .builder()
+                .message(
+                        "Failed to determine how to authenticate the user: "
+                                + discardedReasons.stream().map(Supplier::get).collect(Collectors.joining(", "))).build();
+    }
+
+    private QueryAuthSchemeParams authSchemeParams(SdkRequest request, ExecutionAttributes executionAttributes) {
+        QueryEndpointParams endpointParams = QueryResolveEndpointInterceptor.ruleParams(request, executionAttributes);
```
gosar, line 99: Just thinking aloud (to confirm my understanding, not actionable for now), so with this, it would look like, for S3:
1. S3AuthSchemeInterceptor creates S3EndpointParams (via calling S3ResolveEndpointInterceptor.ruleParams)
2. converts S3EndpointParams to S3AuthSchemeParams
3. DefaultS3AuthSchemeProvider.resolveAuthScheme converts S3AuthSchemeParams back to S3EndpointParams
4. S3ResolveEndpointInterceptor creates another S3EndpointParams instance (2nd time)

codegen/src/main/java/software/amazon/awssdk/codegen/poet/auth/scheme/AuthSchemeInterceptorSpec.java
```diff
@@ -299,4 +343,8 @@ private TypeName toTypeName(Object valueType) {
         }
         return result;
     }
+
+    private Map<String, ParameterModel> parameters() {
+        return intermediateModel.getEndpointRuleSetModel().getParameters();
```
gosar, line 348: nit: can use `endpointRulesSpecUtils.parameters()` here, or not even have this private method at all.

codegen/src/main/java/software/amazon/awssdk/codegen/poet/auth/scheme/AuthSchemeInterceptorSpec.java
```diff
@@ -109,42 +117,78 @@ private MethodSpec generateResolveAuthOptions() {
...
+            }
+            return builder.build();
         }
+
+        builder.addStatement("$T endpointParams = $T.ruleParams(request, executionAttributes)",
+                             endpointRulesSpecUtils.parametersClassName(),
+                             endpointRulesSpecUtils.resolverInterceptorName());
+        builder.addStatement("$1T.Builder builder = $1T.builder()", authSchemeSpecUtils.parametersInterfaceName());
+        boolean regionIncluded = false;
+        for (Map.Entry<String, ParameterModel> kvp : parameters().entrySet()) {
```
gosar, line 164: nit: seems like keySet might work

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/auth/scheme/query-endpoint-auth-params-with-allowlist-auth-scheme-interceptor.java
sugmanue: Yes, I was going to replicate the logic to generate the params here but the amount of code generated is non-trivial, so this way felt much cleaner, even though all this params translations have to happen.

codegen/src/main/java/software/amazon/awssdk/codegen/poet/auth/scheme/AuthSchemeInterceptorSpec.java
sugmanue: Yeah, let me change it that way.

codegen/src/main/java/software/amazon/awssdk/codegen/poet/auth/scheme/AuthSchemeInterceptorSpec.java
sugmanue: Sounds good, I will remove this method in favor of using the `endpointRulesSpecUtils` defined one.

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/auth/scheme/query-endpoint-auth-params-with-allowlist-auth-scheme-interceptor.java
gosar: yeah, duplicating the code is one option, it avoids one conversion, but still creates S3EndpointParams twice. another option I considered is extracting the `ruleParams()` into a S3CreateEndpointParamsInterceptor that is run first and saves the S3EndpointParams for use by both S3AuthSchemeInterceptor and S3ResolveEndpointInterceptor - doesn't avoid any additional conversion, but avoids creating another instance. I was thinking through it, because there is desire to avoid running resolveEndpoint twice, and this is in the slightly related. But probably not worth doing any optimization from what you have. This is not as expensive and just putting data into appropriate objects. So I think what you have is fine.

gosar (approved):

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/auth/scheme/query-endpoint-auth-params-with-allowlist-auth-scheme-interceptor.java
cenedhryn: There are a couple of things about this that aren't optimal - calling it twice as @gosar points out, but also that we're calling a public method on another interceptor, which feels gross. For the first one, I don't know either that we need to optimize it now. We don't have data on how long it takes to resolve the params or resolve the endpoint yet. For the second one, I'm thinking it'd be nice to extract the rule param creation to it's own component, but I'm not super familiar with the other conversions that are happening that Jay mentioned so I don't have an explicit suggestion for now. It can also wait.

millems: `ruleParams` seems pretty inexpensive, at least for query in this one code review, so I'm not too upset about calling it once for auth and once for endpoints.

If we were hand-writing the code, I think I'd favor putting the creation of the params into a different class. But since this is code generation it means a lot more to maintain in the code generator just so that the generated code is cleaner/more intuitive. I'd favor keeping the code we have to maintain simpler than the code we generate.

millems (commented):
No concerns. Please review the sonarcloud bugs before merge, though. Feel free to mark as false positive if they're all false positives.

millems (approved):

gosar:
> No concerns. Please review the sonarcloud bugs before merge, though. Feel free to mark as false positive if they're all false positives.

The sonalrcloud reports have been worse because of the acm/codecatalyst generated code being copied over. Those have been reverted in https://github.com/aws/aws-sdk-java-v2/pull/4243, the remaining bugs are from unrelated code. Should be good to merge this PR.

PR #4181, opened 2023-07-12
Implements all v4 signers as composable pieces

haydenbaker:
## Modifications
- Implements all v4 signers as composable pieces
- The composition of signers will take place where signer properties decide on how to compose a v4 signer (this will be a fast follow-up to this PR) 

## Testing
- run existing and new tests, make sure they pass

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/aws/aws-sdk-java-v2/blob/master/CONTRIBUTING.md) document
- [x] Local run of `mvn install` succeeds
- [x] My code follows the code style of this project
- [ ] My change requires a change to the Javadoc documentation
- [ ] I have updated the Javadoc documentation accordingly
- [x] I have added tests to cover my changes
- [x] All new and existing tests passed
- [ ] I have added a changelog entry. Adding a new entry must be accomplished by running the `scripts/new-change` script and following the instructions. Commit the new file created by the script in `.changes/next-release` with your changes.
- [ ] My change is to implement 1.11 parity feature and I have updated [LaunchChangelog](https://github.com/aws/aws-sdk-java-v2/blob/master/docs/LaunchChangelog.md)

## License

- [x] I confirm that this pull request can be released under the Apache 2 license

PR #4233, opened 2023-07-27
Add new signer properties to sigv4 signer interface

gosar:
## Motivation and Context

To support different signer use cases. Use of these properties is part of another PR #4181. This just separates out the interface change to unblock other changes based on the interface.

Wanted to separate out the interface change, so we can focus on debating any changes to the interface itself separately from the implementation discussions in #4181.

## Modifications

Added signer properties to the sigv4 signer interface.

## Types of changes

## Checklist

## License

### DIFF
--- a/core/http-auth/src/main/java/software/amazon/awssdk/http/auth/AwsV4HttpSigner.java
+++ b/core/http-auth/src/main/java/software/amazon/awssdk/http/auth/AwsV4HttpSigner.java
@@ -16,6 +16,7 @@
 package software.amazon.awssdk.http.auth;
 
 import java.time.Clock;
+import java.time.Duration;
 import software.amazon.awssdk.annotations.SdkPublicApi;
 import software.amazon.awssdk.http.auth.internal.DefaultAwsV4HttpSigner;
 import software.amazon.awssdk.http.auth.internal.checksums.ChecksumAlgorithm;
@@ -81,6 +82,47 @@ public interface AwsV4HttpSigner extends HttpSigner<AwsCredentialsIdentity> {
      */
     SignerProperty<Boolean> NORMALIZE_PATH =
         SignerProperty.create(Boolean.class, "NormalizePath");
+
+    /**
+     * The location where auth-related data is inserted, as a result of signing.
+     * The valid choices are "Header" and "QueryString", where "Header" indicates that request
+     * headers are added, and "QueryString" indicates query-parameters are added.
+     * This property defaults to "Header".
+     */
+    SignerProperty<String> AUTH_LOCATION =
+        SignerProperty.create(String.class, "AuthLocation");
+
+    /**
+     * The duration for the request to be valid.
+     * This property defaults to the max valid duration (7 days).
+     * This is only used in the case of a pre-signing implementation.
+     */
+    SignerProperty<Duration> EXPIRATION_DURATION =
+        SignerProperty.create(Duration.class, "ExpirationDuration");
+
+    /**
+     * Whether to indicate that a payload is signed or not.
+     * This property defaults to true.
+     * This can be set false to disable payload signing.
+     */
+    SignerProperty<Boolean> PAYLOAD_SIGNING =
+        SignerProperty.create(Boolean.class, "PayloadSigning");
+
+    /**
+     * Whether to handle a request as an event-streaming or not.
+     * This property defaults to false.
+     * This is only used in the case of an implementation that supports event-streams.
+     */
+    SignerProperty<Boolean> EVENT_STREAMING =
+        SignerProperty.create(Boolean.class, "EventStreaming");
+
+    /**
+     * Whether to enable chunked encoding or not.
+     * This property defaults to false.
+     * This is only used in the case of an implementation that supports chunked-encoding.
+     */
+    SignerProperty<Boolean> CHUNKED_ENCODING =
+        SignerProperty.create(Boolean.class, "ChunkedEncoding");
 
     /**
      * Get a default implementation of a {@link AwsV4HttpSigner}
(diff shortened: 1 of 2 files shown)


core/http-auth/src/main/java/software/amazon/awssdk/http/auth/AwsV4HttpSigner.java
```diff
@@ -82,6 +83,47 @@ public interface AwsV4HttpSigner extends HttpSigner<AwsCredentialsIdentity> {
     SignerProperty<Boolean> NORMALIZE_PATH =
         SignerProperty.create(Boolean.class, "NormalizePath");
 
+    /**
+     * The location where auth-related data is inserted, as a result of signing.
+     * The valid choices are "Header" and "QueryString", where "Header" indicates that request
+     * headers are added, and "QueryString" indicates query-parameters are added.
+     * This property defaults to "Header".
+     */
+    SignerProperty<String> AUTH_LOCATION =
```
gosar, line 92: I think this can be an enum instead of String. If folks agree I can make that change.

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/AwsV4HttpSigner.java
```diff
@@ -82,6 +83,47 @@ public interface AwsV4HttpSigner extends HttpSigner<AwsCredentialsIdentity> {
...
+     */
+    SignerProperty<Boolean> PAYLOAD_SIGNING =
+        SignerProperty.create(Boolean.class, "PayloadSigning");
+
+    /**
+     * Whether to handle a request as an event-streaming or not.
+     * This property defaults to false.
+     * This is only used in the case of an implementation that supports event-streams.
+     */
+    SignerProperty<Boolean> EVENT_STREAMING =
```
gosar, line 116: thinking aloud: I was thinking if this is really a Signer property. It feels it is really a property of the request / shape of the request than a configuration property of the signing process? An event streaming request needs special handling to sign the payload. By just saying PAYLOAD_SIGNING=true (and assuming there was another way to know that the request is event streaming) could enable the appropriate signing behavior. 

Then if another signer (e.g., sigv4a, or say another scheme in the future) supports signing event streams, it wouldn't be a new/distinct signer property in that Signer too.

> another way to know the request is event streaming?

To go this route, would need the SdkHttpRequest abstraction of SignRequest to be decorated/extended to include these aspects or separate member of SignRequest directly.

This would be a bigger change to the overall integration, as compared to SignerProperties that an AuthSchemeProvider can add based on `operation` AuthSchemeParameter. I'm not sure if this is worth pursuing, but wanted to share these thoughts here for consideration.

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/AwsV4HttpSigner.java
```diff
@@ -82,6 +83,47 @@ public interface AwsV4HttpSigner extends HttpSigner<AwsCredentialsIdentity> {
...
+        SignerProperty.create(Boolean.class, "EventStreaming");
+
+    /**
+     * Whether to enable chunked encoding or not.
+     * This property defaults to false.
+     * This is only used in the case of an implementation that supports chunked-encoding.
+     */
+    SignerProperty<Boolean> CHUNKED_ENCODING =
```
gosar, line 124: The same thought as above may apply to this. Is chunked encoding considered a property of the request and how chunks are signed is behavior of the signer based on if PayloadSigning=true?

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/AwsV4HttpSigner.java
millems: +1 seems good to me

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/AwsV4HttpSigner.java
millems: Can we just look for the header `Content-Type: application/vnd.amazon.eventstream` to determine it's an event stream request? We can do something similar for chunked-encoding.

gosar: Is that header a protocol level detail or defined by event streaming feature itself?

millems: It's specific to HTTP, but I believe it's true for all protocols. I don't know who "owns" it if we're going to be that abstract, but if it ever becomes not-true in the future, we could reasses then.

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/AwsV4HttpSigner.java
millems: Whether chunked encoding is used is determined in a complex way. Essentially, we use chunked encoding if we need to. We need to depending on a combination of factors determined by the SDK. Those factors include whether we're compressing (therefore, we don't know the length), whether we're doing trailing checksums, etc.

It's surprisingly difficult to tell by the model. It's likely better to not try to couple this one to the model, and just let the signer check the request to see if something upstream decided it should be chunked.

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/AwsV4HttpSigner.java
gosar: Based on offline discussion, clarified that by protocol here I meant the serialization aspect too (e.g., restJson1 v/s awsJson1_0) and not specifically the transport protocol (HTTP or other). But we also clarified that the event streaming sigv4 spec is tied to `application/vnd.amazon.eventstream` so using that as a signal to decide how to sign is fine, as opposed to requiring an input SignerProperty to indicate the behavior. It kinda makes sigv4 support for event streaming tied to the amazon protocols, but sigv4 is also amazon specific, so that _coupling_ is fine. In a hypothetical world where someone uses sigv4 with their own protocol that uses something other than `application/vnd.amazon.eventstream`, they'd just have to provide their own sigv4 signer implementation too that support it, our default sigv4 signer implementation would only support the amazon protocols. This is fine :)

gosar:
Closing this, as these changes are being incorporated in #4181 and subsequent PRs directly.

PR #4235, opened 2023-07-27
S3 Multipart API implementation

L-Applin:
> Duplicate of https://github.com/aws/aws-sdk-java-v2/pull/4224 to try to fix merge form master issues

Adds `MultipartConfiguration` to the `S3AsyncClientBuilder` to enables Multipart operation on S3AsyncClient.

The following methods were added to `S3AsyncClientBuilder` (codegen):

```java
    /**
     * Enables automatic conversion of put and copy method to their equivalent multipart operation.
     */
    default S3AsyncClientBuilder multipartEnabled(Boolean enabled) {
    }

    /**
     * Configuration for multipart operation of this client.
     */
    default S3AsyncClientBuilder multipartConfiguration(MultipartConfiguration multipartConfiguration) {
    }

    /**
     * Configuration for multipart operation of this client.
     */
    default S3AsyncClientBuilder multipartConfiguration(Consumer<Builder> multipartConfiguration) {
        Builder builder = MultipartConfiguration.builder();
        multipartConfiguration.accept(builder);
        return multipartConfiguration(builder.build());
    }
```

The following method has been added to `DefaultS3AsyncClientBuilder` (codegen):

```java
    @Override
    public S3AsyncClientBuilder multipartEnabled(Boolean enabled) {
        clientContextParams.put(MultipartConfiguration.MULTIPART_ENABLED_KEY, enabled);
        return this;
    }

    @Override
    public S3AsyncClientBuilder multipartConfiguration(MultipartConfiguration multipartConfig) {
        clientContextParams.put(MultipartConfiguration.MULTIPART_CONFIGURATION_KEY, multipartConfig);
        return this;
    }
```


services/s3/src/main/java/software/amazon/awssdk/services/s3/multipart/MultipartConfiguration.java
```diff
@@ -0,0 +1,206 @@
...
+ * <em>Note</em>: The multipart operation for {@link S3AsyncClient#getObject(GetObjectRequest, AsyncResponseTransformer)} is
+ * temporarily disabled and will result in throwing a {@link UnsupportedOperationException} if called when configured for
+ * multipart operation.
+ */
+@SdkPublicApi
+public final class MultipartConfiguration implements ToCopyableBuilder<MultipartConfiguration.Builder, MultipartConfiguration> {
+    public static final AttributeMap.Key<MultipartConfiguration> MULTIPART_CONFIGURATION_KEY =
+        new AttributeMap.Key<MultipartConfiguration>(MultipartConfiguration.class){};
+    public static final AttributeMap.Key<Boolean> MULTIPART_ENABLED_KEY =
+        new AttributeMap.Key<Boolean>(Boolean.class){};
```
millems, line 45: These seem internal. Should they be on an internal class? MultipartS3AsyncClient maybe?

services/s3/src/main/java/software/amazon/awssdk/services/s3/multipart/MultipartConfiguration.java
```diff
@@ -0,0 +1,206 @@
...
+        /**
+         * Indicated the value of the part configured size.
+         * @return the value of the part size
+         */
+        Long minimumPartSizeInBytes();
+
+        /**
+         * Configures the maximum amount of memory, in bytes, the SDK will use to buffer content of requests in memory.
+         * Increasing this value may lead to better performance at the cost of using more memory.
+         * <p></p>
```
millems, line 152: Nit: I generally see people just do the (invalid XHTML) `<p>` between paragraphs or wrap the non-leading paragraphs in `<p>` and `</p>` I don't usually see this pattern. It probably works, but it looks weird to me.

services/s3/src/main/java/software/amazon/awssdk/services/s3/multipart/MultipartConfiguration.java
L-Applin: Shouldn't they be public so users can retrieve those values with clientContextParams?

millems: Do we need users to be able to retrieve those values?

services/s3/src/main/java/software/amazon/awssdk/services/s3/multipart/MultipartConfiguration.java
zoewangg: It's in our guideline :) https://github.com/aws/aws-sdk-java-v2/blob/master/docs/design/APIReference.md

`A single <p> MUST be used between paragraphs.`

services/s3/src/main/java/software/amazon/awssdk/services/s3/multipart/MultipartConfiguration.java
L-Applin: We dont *need* them to, no, but they might want to. Is there any good reason we would want to prevent users to access those value? The one I see is it might be confusing if they retrieve the config to change it and it has no effect since these have effect on when the client is build, not for every request. Actually, now that I think of it, that might be reason enough to prevent it actually...

millems: Also that it's a public API we'd have to maintain in perpetuity.

L-Applin: I'll move them to `S3AsyncClientDecorator` it is an `@SdkInternalApi` class and is where they are used outside of codegen. And where it needs to be used by the codegen already knows about it, so that seems to be a cohesive solution.

millems (approved):

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/multipart/MultipartS3AsyncClient.java
```diff
@@ -17,31 +17,59 @@
...
 @SdkInternalApi
-public class MultipartS3AsyncClient extends DelegatingS3AsyncClient {
-    private static final long DEFAULT_PART_SIZE_IN_BYTES = 8L * 1024 * 1024;
+public final class MultipartS3AsyncClient extends DelegatingS3AsyncClient {
+
+    private static final ApiName USER_AGENT_API_NAME = ApiName.builder().name("hll").version("s3Multipart").build();
+
+    private static final long DEFAULT_MIN_PART_SIZE = 8L * 1024 * 1024;
     private static final long DEFAULT_THRESHOLD = 8L * 1024 * 1024;
+    private static final long DEFAULT_API_CALL_BUFFER_SIZE = DEFAULT_MIN_PART_SIZE * 2;
```
zoewangg, line 51: Based on the perf result, should we bump this to `* 4` or keep it as is?

services/s3/src/main/java/software/amazon/awssdk/services/s3/multipart/MultipartConfiguration.java
```diff
@@ -0,0 +1,201 @@
...
+        return builder()
+            .apiCallBufferSizeInBytes(apiCallBufferSizeInBytes)
+            .minimumPartSizeInBytes(minimumPartSizeInBytes)
+            .thresholdInBytes(thresholdInBytes);
+    }
+
+    /**
+     * Indicates the value of the configured threshold, in bytes. Any request whose size is less than the configured value will
+     * not
+     * use multipart operation
```
zoewangg, line 67: nite: formatting is a bit off here

services/s3/src/main/java/software/amazon/awssdk/services/s3/multipart/MultipartConfiguration.java
```diff
@@ -0,0 +1,201 @@
...
+     * @return the value of the configured threshold.
+     */
+    public Long thresholdInBytes() {
+        return this.thresholdInBytes;
+    }
+
+    /**
+     * Indicated the size, in bytes, of each individual part of the part requests. The actual part size used might be bigger to
+     * conforms to
```
zoewangg, line 76: same here

services/s3/src/main/java/software/amazon/awssdk/services/s3/multipart/MultipartConfiguration.java
```diff
@@ -0,0 +1,201 @@
...
+
+        /**
+         * Indicates the value of the configured threshold.
+         * @return the value of the threshold.
+         */
+        Long thresholdInBytes();
+
+        /**
+         * Configures the part size, in bytes, to be used in each individual part requests.
+         * <p></p>
```
zoewangg, line 120: Should be single `<p>` as well. You can check the generated javadoc page locally after a successful build.

services/s3/src/main/java/software/amazon/awssdk/services/s3/multipart/MultipartConfiguration.java
```diff
@@ -0,0 +1,201 @@
...
+        Builder thresholdInBytes(Long thresholdInBytes);
+
+        /**
+         * Indicates the value of the configured threshold.
+         * @return the value of the threshold.
+         */
+        Long thresholdInBytes();
+
+        /**
+         * Configures the part size, in bytes, to be used in each individual part requests.
```
zoewangg, line 119: Can we clarify this only applies to putObject and copyObject?

zoewangg (approved):

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/multipart/MultipartS3AsyncClient.java
L-Applin: Yup, I ran some test and going to 4x seems to be a good balance. It is an improvement that seems to be worth it, but going higher sees diminishing returns:

## 4GiB Upload Result - 8 MiB part size
| api call buffer size | Average latency (s) | Average throughput (Gbps) |
|---|---|---|
| 2x (16 Mib) |  6.359 | 5.032 |
| 3x (24 Mib) |  6.598 | 4.850 |
| 4x (32 Mib) | 5.585 | 5.729 |
| 8x (64 Mib) |  5.760 | 5.556 |
| 3 GiB | 4.700 | 6.808 |

PR #4240, opened 2023-07-29
Fix blocking close method in multiplexed channel

Arooba-git:
Hello! 🙂 
This PR fixes a blocking case similar to https://github.com/aws/aws-sdk-java-v2/issues/2145
While that issue was addressed by adding the blocking method to allowed blocking calls, BlockHound still detects another blocking `close` method in Http2MultiplexedChannelPool class:

aws-blocking
aws-blocking-code

This PR fixes the call. We re-ran the tests and they passed after the fix, except the one related to interruption, which we have updated according to the changes. 

In addition we compared the performance (in terms of memory usage) before and after the fix:

**Before**
aws-before

**After**
aws-after

## Types of changes

## Checklist

## License


debora-ito:
Hi @Arooba-git apologies for the delay in response.

I understand you're trying to change close() to be non-blocking, but we don't think that changing to a single thread executor is the right solution, as it would impact throughput performance.

Arooba-git:
> Hi @Arooba-git apologies for the delay in response.

No worries :) 


> we don't think that changing to a single thread executor is the right solution, as it would impact throughput performance.

I see.. though we can set its property so that its a 'daemon' thread... no? 🤔

debora-ito:
Still, we would like to avoid it. We don't have suggestions for now that can be addressed in this PR. You can create a new Github issue as a feature request if you'd like.

PR #4245, opened 2023-08-01
Add unit tests for (Async)SigningStage

gosar:
## Motivation and Context

When SRA logic was added to SigningStage and AsyncSigningStage, unit tests weren't added. No unit tests existed for these classes prior to SRA either.

## Modifications

Added unit tests for SigningStage and AsyncSigningStage.

## Types of changes

## Checklist

## License

### DIFF
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/AsyncSigningStage.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/AsyncSigningStage.java
@@ -15,8 +15,10 @@
 
 package software.amazon.awssdk.core.internal.http.pipeline.stages;
 
+import java.nio.ByteBuffer;
 import java.time.Duration;
 import java.util.concurrent.CompletableFuture;
+import org.reactivestreams.Publisher;
 import software.amazon.awssdk.annotations.SdkInternalApi;
 import software.amazon.awssdk.core.SelectedAuthScheme;
 import software.amazon.awssdk.core.async.AsyncRequestBody;
@@ -59,7 +61,6 @@ public class AsyncSigningStage implements RequestPipeline<SdkHttpFullRequest,
     @Override
     public CompletableFuture<SdkHttpFullRequest> execute(SdkHttpFullRequest request, RequestExecutionContext context)
             throws Exception {
-        // TODO: Add unit tests for SRA signing logic.
         if (shouldUseSelectedAuthScheme(context)) {
             return sraSignRequest(request,
                                   context,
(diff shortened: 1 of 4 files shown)


core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/AsyncSigningStage.java
```diff
@@ -128,8 +129,17 @@ private <T extends Identity> CompletableFuture<SdkHttpFullRequest> doSraSign(Sdk
...
+        if (signedRequest.payload().isPresent()) {
+            Publisher<ByteBuffer> signedPayload = signedRequest.payload().get();
+            if (signedPayload instanceof AsyncRequestBody) {
+                newAsyncRequestBody = (AsyncRequestBody) signedPayload;
+            } else {
+                newAsyncRequestBody = AsyncRequestBody.fromPublisher(signedPayload);
+            }
+        } else {
+            newAsyncRequestBody = null;
+        }
```
dagnir, line 142: nit: can we just assign `null` in line 132 and remove this branch?

core/sdk-core/src/test/java/software/amazon/awssdk/core/internal/http/pipeline/stages/AsyncSigningStageTest.java
dagnir, line 1: Can we include a test that uses an `AsyncRequestBodySigner`?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/AsyncSigningStage.java
gosar: yeah I had started there, but this variable needs to be final or effectively final in the lambda expression later in
```
        executionContext.interceptorContext(executionContext.interceptorContext()
                                                            .copy(b -> b.asyncRequestBody(newAsyncRequestBody)));
```

core/sdk-core/src/test/java/software/amazon/awssdk/core/internal/http/pipeline/stages/AsyncSigningStageTest.java
gosar: yeah, I was focused on adding tests for the new SRA code, which should become the primary code path going forward, but anyways included tests for the basic non-SRA cases instead of covering all branches for prior code. But I'll go ahead and add test for AsyncRequestBodySigner

dagnir (approved):

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/AsyncSigningStage.java
```diff
@@ -128,8 +129,17 @@ private <T extends Identity> CompletableFuture<SdkHttpFullRequest> doSraSign(Sdk
     }
 
     private static void updateAsyncRequestBodyInContexts(RequestExecutionContext context, AsyncSignedRequest signedRequest) {
-        AsyncRequestBody newAsyncRequestBody = signedRequest.payload().isPresent() ?
-                                               AsyncRequestBody.fromPublisher(signedRequest.payload().get()) : null;
+        AsyncRequestBody newAsyncRequestBody;
+        if (signedRequest.payload().isPresent()) {
+            Publisher<ByteBuffer> signedPayload = signedRequest.payload().get();
```
sugmanue, line 134: We currently have some [sonar cloud](https://sonarcloud.io/project/issues?resolved=false&types=BUG&pullRequest=4251&id=aws_aws-sdk-java-v2&open=AYm4BIRB3hbnKbw3XF2W) bugs for using a pattern similar to this one, in general Java does not ensure that 

```java
signedResquest.payload() == singedRequestPayload()
```

i.e., the second call can return a different instance in a multi-threaded environment. I  think that's not possible for this case but sonar cloud does not seem to be able to figure out that the object is immutable. I think is better to acquire a reference for the optional instance locally and avoid the issue which also avoids calling the method twice, as in

```java
Optional<Publisher<ByteBuffer>> optionalPayload = signedRequest.payload()
if (optionalPayload.isPresent()) {
    Publisher<ByteBuffer> signedPayload = optionalPayload.get();
    ⋯
}
```

gosar: Yeah, I marked the sonar finding as a false positive, because like you say it doesn't know that signedRequest.payload)() is immutable. But your suggestion makes sense, so I'll make that change. Thanks!

PR #4251, opened 2023-08-02
Add sigv4a for services that have enabled endpoint based auth params

sugmanue:
## Motivation and Context

`Sigv4a` is a supported auth scheme for endpoints but not directly modeled. This change adds it if the auth params from endpoints param switch is enabled.

**Note** The scheme currently does not exist but in the CRT package, this change just adds the scheme and the changes to the base client builder, subsequent changes will implement the signer that will dynamically load the required classes as currently done as [`SignerLoader`](https://github.com/aws/aws-sdk-java-v2/blob/efb5655f19cd7dbbbf6f4b5b141c94aaf59abe19/core/auth/src/main/java/software/amazon/awssdk/auth/signer/SignerLoader.java#L30) currently does.

For S3 the generated code in the base client will look like:

```java
    private Map<String, AuthScheme<?>> defaultAuthSchemes() {
        AwsV4AuthScheme awsV4AuthScheme = AwsV4AuthScheme.create();
        AwsV4aAuthScheme awsV4aAuthScheme = AwsV4aAuthScheme.create();
        return MapUtils.of(awsV4AuthScheme.schemeId(), awsV4AuthScheme, awsV4aAuthScheme.schemeId(), awsV4aAuthScheme);
    }

```

## Testing

* Added unit tests

## Types of changes

## Checklist

## License

### DIFF
--- a/codegen/src/main/java/software/amazon/awssdk/codegen/poet/auth/scheme/AuthSchemeSpecUtils.java
+++ b/codegen/src/main/java/software/amazon/awssdk/codegen/poet/auth/scheme/AuthSchemeSpecUtils.java
@@ -182,19 +182,6 @@ public final class AuthSchemeSpecUtils {
         return Collections.singletonList(intermediateModel.getMetadata().getAuthType());
     }
 
-    public List<AuthType> allServiceAuthTypes() {
-        Set<AuthType> result =
-            intermediateModel.getOperations()
-                             .values()
-                             .stream()
-                             .map(OperationModel::getAuth)
-                             .flatMap(List::stream)
-                             .collect(Collectors.toSet());
-        List<AuthType> modeled = intermediateModel.getMetadata().getAuth();
-        result.addAll(modeled);
-        return result.stream().sorted().collect(Collectors.toList());
-    }
-
     public Set<Class<?>> allServiceConcreteAuthSchemeClasses() {
         Set<Class<?>> result =
             Stream.concat(intermediateModel.getOperations()
@@ -209,7 +196,7 @@ public final class AuthSchemeSpecUtils {
                   .collect(Collectors.toCollection(() -> new TreeSet<>(Comparator.comparing(Class::getSimpleName))));
 
         if (generateEndpointBasedParams()) {
-            // sigv4a is modeled but needed for the endpoints based auth-scheme cases.
+            // sigv4a is not modeled but needed for the endpoints based auth-scheme cases.
             result.add(AwsV4aAuthScheme.class);
         }
         return result;
--- a/codegen/src/main/java/software/amazon/awssdk/codegen/poet/auth/scheme/AuthSchemeSpecUtils.java
+++ b/codegen/src/main/java/software/amazon/awssdk/codegen/poet/auth/scheme/AuthSchemeSpecUtils.java
@@ -195,7 +195,7 @@ public final class AuthSchemeSpecUtils {
                   .filter(Objects::nonNull)
                   .collect(Collectors.toCollection(() -> new TreeSet<>(Comparator.comparing(Class::getSimpleName))));
 
-        if (generateEndpointBasedParams()) {
+        if (useEndpointBasedAuthProvider()) {
             // sigv4a is not modeled but needed for the endpoints based auth-scheme cases.
             result.add(AwsV4aAuthScheme.class);
         }
--- a/codegen/src/main/java/software/amazon/awssdk/codegen/poet/auth/scheme/AuthSchemeSpecUtils.java
+++ b/codegen/src/main/java/software/amazon/awssdk/codegen/poet/auth/scheme/AuthSchemeSpecUtils.java
@@ -25,13 +25,17 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
+import java.util.TreeSet;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import software.amazon.awssdk.codegen.model.config.customization.CustomizationConfig;
 import software.amazon.awssdk.codegen.model.intermediate.IntermediateModel;
 import software.amazon.awssdk.codegen.model.intermediate.OperationModel;
 import software.amazon.awssdk.codegen.model.service.AuthType;
 import software.amazon.awssdk.codegen.utils.AuthUtils;
+import software.amazon.awssdk.http.auth.AwsV4aAuthScheme;
 import software.amazon.awssdk.http.auth.spi.AuthSchemeOption;
 
 public final class AuthSchemeSpecUtils {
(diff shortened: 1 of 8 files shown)


codegen/src/main/java/software/amazon/awssdk/codegen/poet/auth/scheme/AuthSchemeSpecUtils.java
```diff
@@ -191,6 +195,26 @@ public List<AuthType> allServiceAuthTypes() {
...
+                                           .map(OperationModel::getAuth)
+                                           .flatMap(List::stream),
+                          intermediateModel.getMetadata().getAuth().stream())
+                  .map(AuthSchemeCodegenMetadata::fromAuthType)
+                  .map(AuthSchemeCodegenMetadata::authSchemeClass)
+                  .filter(Objects::nonNull)
+                  .collect(Collectors.toCollection(() -> new TreeSet<>(Comparator.comparing(Class::getSimpleName))));
+
+        if (generateEndpointBasedParams()) {
+            // sigv4a is modeled but needed for the endpoints based auth-scheme cases.
```
gosar, line 212: did you mean to say `sigv4a is not modeled`?

codegen/src/main/java/software/amazon/awssdk/codegen/poet/auth/scheme/AuthSchemeSpecUtils.java
```diff
@@ -191,6 +195,26 @@ public List<AuthType> allServiceAuthTypes() {
         return result.stream().sorted().collect(Collectors.toList());
     }
 
+    public Set<Class<?>> allServiceConcreteAuthSchemeClasses() {
```
gosar, line 198: This replaces the use of `allServiceAuthTypes`. So I think remove that method. If you wanted to keep it, maybe it's a private method that this method calls and then processes it. I don't like it being left there as a public method, because it seems generically useful, but has sorting behavior and use of List as opposed to Set, which was specifically for how it was being used for codegen'ing the authSchemes Map. In short, I think delete `allServiceAuthTypes`.

sugmanue: Removed.

codegen/src/main/java/software/amazon/awssdk/codegen/poet/auth/scheme/AuthSchemeSpecUtils.java
```diff
@@ -178,17 +182,24 @@ public List<AuthType> serviceDefaultAuthTypes() {
...
+                                           .stream()
+                                           .map(OperationModel::getAuth)
+                                           .flatMap(List::stream),
+                          intermediateModel.getMetadata().getAuth().stream())
+                  .map(AuthSchemeCodegenMetadata::fromAuthType)
+                  .map(AuthSchemeCodegenMetadata::authSchemeClass)
+                  .filter(Objects::nonNull)
+                  .collect(Collectors.toCollection(() -> new TreeSet<>(Comparator.comparing(Class::getSimpleName))));
+
+        if (generateEndpointBasedParams()) {
```
gosar, line 198: readability wise, `useEndpointBasedAuthProvider()` seems better

gosar: Moreover, I think we should rename this customization from `enableEndpointAuthSchemeParams` to something like `enableEndpointAuthSchemeResolution` - the scope is generalized v/s specific to params which is just the first thing we customized. And then maybe we don't need multiple methods in this Utils class that checks that flag.

gosar: Maybe `enableSraEndpointAuthSchemeResolution` to convey this is the SRA part of the feature v/s Endpoint 2.0 itself.

codegen/src/test/java/software/amazon/awssdk/codegen/poet/builder/BuilderClassTest.java
```diff
@@ -58,6 +58,11 @@ public void baseQueryClientBuilderClass() throws Exception {
         validateQueryGeneration(BaseClientBuilderClass::new, "test-query-client-builder-class.java");
     }
 
+    @Test
+    public void baseClientBuilderClassWithEndpointsAuthParams() {
+        assertThat(new BaseClientBuilderClass(ClientTestModels.serviceMiniS3()), generatesTo("test-client-builder-endpoints-auth-params.java"));
+    }
```
gosar: Based on the other comment about naming this customization, probably do
```
    public void baseClientBuilderClassWithSraEndpointsAuthSchemeResolution() {
        assertThat(new BaseClientBuilderClass(ClientTestModels.serviceMiniS3()), generatesTo("test-client-builder-sra-endpoints-auth-resolution.java"));
    }
```

codegen/src/test/java/software/amazon/awssdk/codegen/poet/builder/BuilderClassTest.java
```diff
@@ -58,6 +58,11 @@ public void baseQueryClientBuilderClass() throws Exception {
         validateQueryGeneration(BaseClientBuilderClass::new, "test-query-client-builder-class.java");
     }
 
+    @Test
+    public void baseClientBuilderClassWithEndpointsAuthParams() {
+        assertThat(new BaseClientBuilderClass(ClientTestModels.queryServiceModelsEndpointAuthParamsWithAllowList()), generatesTo("test-client-builder-endpoints-auth-params.java"));
+    }
```
gosar, line 64: Based on the other comment about naming this customization, probably do
```
    public void baseClientBuilderClassWithSraEndpointsAuthSchemeResolution() {
        assertThat(new BaseClientBuilderClass(ClientTestModels.queryServiceModelsEndpointAuthParamsWithAllowList()), generatesTo("test-client-builder-sra-endpoints-auth-resolution.java"));
    }
```

codegen/src/main/java/software/amazon/awssdk/codegen/poet/auth/scheme/AuthSchemeSpecUtils.java
sugmanue: I will switch to use `useEndpointBasedAuthProvider()` but let's discuss the renaming/refactoring in a separated thread, I'm not sure about the proposed name or adding the SRA which might only make sense internally.

gosar: separate PR sounds good

gosar (approved):

codegen/src/main/java/software/amazon/awssdk/codegen/poet/auth/scheme/AuthSchemeSpecUtils.java
```diff
@@ -178,17 +182,24 @@ public List<AuthType> serviceDefaultAuthTypes() {
...
-            intermediateModel.getOperations()
-                             .values()
-                             .stream()
-                             .map(OperationModel::getAuth)
-                             .flatMap(List::stream)
-                             .collect(Collectors.toSet());
-        List<AuthType> modeled = intermediateModel.getMetadata().getAuth();
-        result.addAll(modeled);
-        return result.stream().sorted().collect(Collectors.toList());
+    public Set<Class<?>> allServiceConcreteAuthSchemeClasses() {
```
joviegas, line 185: Optional comment : How about `discoverSupportedAuthSchemeClasses` ?

joviegas (approved):

PR #4258, opened 2023-08-03
InputStream#readAllBytes() should receive no premature EOS

faucct:
A read-call with len == 0 happens here:
https://github.com/AdoptOpenJDK/openjdk-jdk11/blob/master/src/java.base/share/classes/java/io/InputStream.java#L396

## License


mgrundie-r7 (approved):

dagnir:
Thanks @faucct. The change itself looks good; can you please add a test for this here? https://github.com/aws/aws-sdk-java-v2/blob/5aa3ff309ca5698eb452aa077f191d842f0fa55a/utils/src/test/java/software/amazon/awssdk/utils/async/InputStreamSubscriberTest.java#L46

faucct:
I have tried adding the test for it, but don’t know, how to skip its compilation on Java 8. @dagnir , could you add it for me, please, if you know the way?

```
    @EnabledForJreRange(min = JRE.JAVA_9)
    @Test
    public void readAllBytesShouldReceiveNoPrematureEOS() throws IOException {
        publisher.subscribe(subscriber);
        publisher.send(byteBufferOfLength(2 << 14));
        publisher.complete();
        assertThat(subscriber.readAllBytes()).isEqualTo(new byte[2 << 14]);
    }
```

dagnir:
Hey sorry for the delayed response. N worries, I can add a test case.

dagnir (approved):

zoewangg (approved):
Can we add a change log entry for this?

dagnir:
@all-contributors please add @faucct for code

PR #4260, opened 2023-08-03
Fix a race condition where the third upload part request was sent bef…

zoewangg:
Fix a race condition where the third upload part request was sent before the second one

### DIFF
--- a/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/multipart/UploadWithUnknownContentLengthHelper.java
+++ b/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/multipart/UploadWithUnknownContentLengthHelper.java
@@ -174,11 +174,13 @@ public final class UploadWithUnknownContentLengthHelper {
                         subscription.cancel();
                     } else {
                         uploadId = createMultipartUploadResponse.uploadId();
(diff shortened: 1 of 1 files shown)


services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/multipart/UploadWithUnknownContentLengthHelper.java
```diff
@@ -174,11 +174,13 @@ public void onNext(AsyncRequestBody asyncRequestBody) {
...
                     } else {
                         uploadId = createMultipartUploadResponse.uploadId();
-                        uploadIdFuture.complete(uploadId);
                         log.debug(() -> "Initiated a new multipart upload, uploadId: " + uploadId);
 
                         sendUploadPartRequest(uploadId, firstRequestBody);
                         sendUploadPartRequest(uploadId, asyncRequestBody);
+
+                        // We need to complete the uploadIdFuture *after* the first two requests have been sent
+                        uploadIdFuture.complete(uploadId);
```
zoewangg, line 183: The race condition is that the third request on line 189 could be executed before the first and second request, so the fix is completing `uploadIdFuture` *after* the first two requests have been finished.

L-Applin (approved):

PR #4220, opened 2023-07-20
Wire up the auth scheme interceptor in the base client builder

sugmanue:
## Motivation and Context

Wire up the auth scheme interceptor in the base client builder. After this some minor issues were found and fixed, this change include those as otherwise the build will not succeed.

## Testing

* Added unit tests

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/aws/aws-sdk-java-v2/blob/master/CONTRIBUTING.md) document
- [x] Local run of `mvn install` succeeds
- [x] My code follows the code style of this project
- [x] My change requires a change to the Javadoc documentation
- [x] I have updated the Javadoc documentation accordingly
- [ ] I have added tests to cover my changes
- [x] All new and existing tests passed
- [ ] I have added a changelog entry. Adding a new entry must be accomplished by running the `scripts/new-change` script and following the instructions. Commit the new file created by the script in `.changes/next-release` with your changes.
- [ ] My change is to implement 1.11 parity feature and I have updated [LaunchChangelog](https://github.com/aws/aws-sdk-java-v2/blob/master/docs/LaunchChangelog.md)

## License

- [ ] I confirm that this pull request can be released under the Apache 2 license

PR #4264, opened 2023-08-04
Use old SigningStage logic if Signer is overridden

gosar:
## Motivation and Context

Pre SRA, customer can provide their own Signer. This will be deprecated with SRA, where customer can provide their own HttpSigner (new interface) via configuring AuthScheme. But for customers that are still using the old approach, it needs to continue to work and override the signer setup via new SRA code.

## Modifications

To support old signer override, leaving the previous signer logic in (Async)SigningStage as is and branching to it, if signer is overridden.
Currently, signer is considered overridden in 2 cases: 1. by customer specifiying signer explicitly, 2. but also via older codegen that used different signer for certain operations than the client's default signer for cases like event streaming , etc. In subsequent PRs, with SRA, these operations specific signer overrides will go away, so isSignerOverridden is represent only customer overridden signer.

## Testing

This allows us to re-enable some tests that were disabled earlier in https://github.com/aws/aws-sdk-java-v2/pull/4220#discussion_r1272531389.

## Types of changes

## Checklist

## License

### DIFF
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/AsyncSigningStage.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/AsyncSigningStage.java
@@ -21,7 +21,6 @@ import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import org.reactivestreams.Publisher;
 import software.amazon.awssdk.annotations.SdkInternalApi;
-import software.amazon.awssdk.core.RequestOverrideConfiguration;
 import software.amazon.awssdk.core.SelectedAuthScheme;
 import software.amazon.awssdk.core.async.AsyncRequestBody;
 import software.amazon.awssdk.core.http.ExecutionContext;
@@ -31,6 +30,7 @@ import software.amazon.awssdk.core.interceptor.SdkInternalExecutionAttribute;
 import software.amazon.awssdk.core.internal.http.HttpClientDependencies;
 import software.amazon.awssdk.core.internal.http.RequestExecutionContext;
 import software.amazon.awssdk.core.internal.http.pipeline.RequestPipeline;
+import software.amazon.awssdk.core.internal.http.pipeline.stages.utils.SignerOverrideUtils;
 import software.amazon.awssdk.core.metrics.CoreMetric;
 import software.amazon.awssdk.core.signer.AsyncRequestBodySigner;
 import software.amazon.awssdk.core.signer.AsyncSigner;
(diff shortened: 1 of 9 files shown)


core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/AsyncSigningStage.java
```diff
@@ -232,10 +233,23 @@ private boolean shouldSign(Signer signer) {
...
-    private boolean shouldUseSelectedAuthScheme(RequestExecutionContext context) {
-        return context.executionAttributes().getAttribute(SdkInternalExecutionAttribute.SELECTED_AUTH_SCHEME) != null;
+    private boolean shouldDoSraSigning(RequestExecutionContext context) {
+        return !isSignerOverridden(context)
+               && context.executionAttributes().getAttribute(SdkInternalExecutionAttribute.SELECTED_AUTH_SCHEME) != null;
+    }
+
+    // This is copied from SignerOverrideUtils.isSignerOverridden, but that is in aws-core, and this is in sdk-core.
+    // Just this method could be moved to sdk-core somewhere, but the other methods in SignerOverrideUtils depend on AwsRequest
+    // and AwsRequestOverrideConfiguration which are in aws-core. Just duplicating the logic here seemed ok.
```
gosar, line 245: I'm considering moving just this method to a `SignerOverrideUtils` in sdk-core. I would have renamed the one in aws-core to AwsSignerOverrideUtils maybe, but I can't change the name / remove the current method from SdkProtectedApi, so I'll probably duplicate just this method and add a `SdkSignerOverrideUtils` in sdk-core. SignerOverrideUtils in aws-core actually should get deprecated in subsequent PR.

test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/AsyncSignerOverrideTest.java
```diff
@@ -58,4 +58,10 @@ public void test_signerOverriddenForStreamingInput_takesPrecedence() {
 
         verify(mockSigner).sign(any(SdkHttpFullRequest.class), any(ExecutionAttributes.class));
     }
+
+    // TODO(sra-identity-and-auth): Does adding a test with SRA way of overriding signer here makes sense? Probably useful as a
+    //  general test of SRAs support of provider your own signer, though not specific to the overriding of of AsyncAws4Signer.
+    //  Maybe still makes sense, by asserting that signAsync is called for streaming input operation (may belong to a different
+    //  test class)?
```
millems, line 65: +1 makes sense to add one. Here is probably as good a location as any.

millems (approved):

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/AsyncSigningStage.java
gosar: I found an issue with the existing method in SignerOverrideUtils which just checks that the ExecutionAttribute SIGNER_OVERRIDDEN is present and is not checking that it is `true`. It doesn't affect anything since the attribute is never set to false anywhere. So I'm going to pull this out in a new Utils in sdk-core.

test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/AsyncSignerOverrideTest.java
gosar: Here=codegen-generated-classes-test yes, I just thought here=AsyncSignerOverride not so, because of Async in the test class name. I'll rename the class to more general and add the test here.

gosar: Well, can't add that test until another task is done - to allow putAuthScheme on client, as a way to provider your own signer. So leaving the TODO, and resolving this comment for now.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/utils/SignerOverrideUtils.java
```diff
@@ -0,0 +1,48 @@
...
+ * Utility to share across {@link SigningStage} and {@link AsyncSigningStage}.
+ */
+@SdkInternalApi
+public final class SignerOverrideUtils {
+
+    private SignerOverrideUtils() {
+    }
+
+    // This is (mostly) same as software.amazon.awssdk.awscore.util.SignerOverrideUtils.isSignerOverridden, but since that
+    // is in aws-core, and this is in sdk-core, copied here. It is "mostly" because it changes the SIGNER_OVERRIDDEN check to
```
gosar, line 37: I could change software.amazon.awssdk.awscore.util.SignerOverrideUtils.isSignerOverridden to check for `true` too, but hesitated since I didn't want to touch existing code in this branch and become part of this release. The old signer will be deprecated as part of this release, so I didn't feel like touching it now. If you feel I should, I can.

millems (approved):

PR #4269, opened 2023-08-07
Implement chunked encoding

haydenbaker:
- Add a chunked-encoding implementation

- Tests are added, which includes a test of an implementation of AWS-signed (V4) and encoded chunks per [the S3 documentation](https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-streaming-trailers.html)

## Motivation and Context
- Chunked-encoding (with V4 signing) is used in the S3 and S3-CRT signers, so this is an essential component to support those

## Modifications
- The implementation is based on the http chunked-encoding specification, it supports:
  - headers
  - chunk extensions
  - trailers
- It includes a few interfaces for extending the chunked-encoding behaviors

## Testing
- Run tests, verify pass

## Types of changes

## Checklist

## License


core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/io/SdkInputStream.java
haydenbaker, line 1: This is a copied file from sdk-core.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/chunkedencoding/Chunk.java
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+     * Get the underlying stream of data for a chunk.
+     */
+    InputStream stream() throws IOException;
+
+    /**
+     * Whether the logical end of a chunk has been reached.
+     */
+    boolean ended();
```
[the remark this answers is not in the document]
haydenbaker: how about `hasEnded`?

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/chunkedencoding/ChunkEncodedInputStream.java
```diff
@@ -0,0 +1,162 @@
...
+        if (currentChunk != null) {
+            currentChunk.close();
+        }
+        // we *have* to read from the backing stream in order to figure out if it's the end or not
+        byte[] chunkData = new byte[chunkSize];
+        int read = stream.read(chunkData, 0, chunkSize);
+
+        if (read > 0) {
+            // form the new chunk, which includes the header, the extensions, and the chunk data
+            byte[] chunk = Arrays.copyOf(chunkData, read);
```
cenedhryn, line 95: I was a bit confused reading this code, because we create a chunk, send it to the output stream and then assign the result back to the chunk. From a readability POV, can we inline the array copy to the writeChunk call?

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/chunkedencoding/ChunkEncodedInputStream.java
```diff
@@ -0,0 +1,162 @@
...
+            ByteArrayOutputStream chunkStream = new ByteArrayOutputStream();
+            writeChunk(chunk, chunkStream);
+            chunkStream.write(CRLF);
+            chunk = chunkStream.toByteArray();
```
cenedhryn, line 99: maybe `chunkContent`?

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/signer/RollingSigner.java
```diff
@@ -0,0 +1,54 @@
...
+
+import java.util.function.Function;
+import software.amazon.awssdk.annotations.SdkInternalApi;
+
+/**
+ * A class which calculates a rolling signature of arbitrary data using HMAC-SHA256.
+ * Each time a signature is calculated, the prior calculation is incorporated, hence "rolling".
+ */
+@SdkInternalApi
+public final class RollingSigner {
```
cenedhryn, line 29: unit tests?

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/io/ChunkInputStream.java
```diff
@@ -0,0 +1,44 @@
...
+
+import java.io.IOException;
+import java.io.InputStream;
+import software.amazon.awssdk.annotations.SdkInternalApi;
+
+/**
+ * A wrapped stream to represent a "chunk" of data
+ */
+@SdkInternalApi
+public final class ChunkInputStream extends SdkLengthAwareInputStream {
```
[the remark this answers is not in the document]
haydenbaker: Yeah, I suppose it wouldn't hurt.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/chunkedencoding/ChunkEncodedInputStream.java
```diff
@@ -0,0 +1,162 @@
...
+    protected InputStream getWrappedInputStream() {
+        return inputStream;
+    }
+
+    public int getChunkSize() {
+        return chunkSize;
+    }
+
+    @Override
+    public int read() throws IOException {
```
zoewangg, line 76: Don't we need to implement `read(byte b[], int off, int len)` as well? I'm not sure how it would work with the current Chunk implementation though.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/chunkedencoding/Trailer.java
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * A functional interface for defining a trailer, where the trailer is a key-value pair.
+ * <p>
+ * The trailer usually depends on the chunk-data itself (checksum, signature, etc.), but is not required to.
+ * <p>
+ * Trailers are optionally included in the final chunk of a chunk-encoded stream.
+ */
+@SdkProtectedApi
+public interface Trailer {
```
zoewangg, line 29: `ChunkTrailer`?

core/http-auth-aws/src/test/java/software/amazon/awssdk/http/auth/aws/chunkedencoding/ChunkEncodedInputStreamTest.java
```diff
@@ -0,0 +1,225 @@
...
+import java.util.Arrays;
+import java.util.Collections;
+import org.junit.jupiter.api.Test;
+import software.amazon.awssdk.http.auth.aws.internal.signer.RollingSigner;
+import software.amazon.awssdk.http.auth.aws.signer.CredentialScope;
+import software.amazon.awssdk.http.auth.aws.util.SignerConstant;
+import software.amazon.awssdk.identity.spi.AwsCredentialsIdentity;
+import software.amazon.awssdk.utils.Pair;
+
+public class ChunkEncodedInputStreamTest {
```
zoewangg, line 26: Can we add tests to cover the cases where multiple extensions, trailers are provided?

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/chunkedencoding/ChunkEncodedInputStream.java
```diff
@@ -0,0 +1,162 @@
...
+        outputStream.write(chunk);
+    }
+
+    private void writeHeader(byte[] chunk, ByteArrayOutputStream outputStream) throws IOException {
+        byte[] hdr = header.get(chunk);
+        outputStream.write(hdr);
+    }
+
+    private void writeExtensions(byte[] chunk, ByteArrayOutputStream outputStream) throws IOException {
+        List<Pair<byte[], byte[]>> exts = extensions.stream().map(ext -> ext.get(chunk)).collect(Collectors.toList());
```
zoewangg, line 131: If we are going to do for loop below, can't we iterate this in the for loop?

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/chunkedencoding/ChunkEncodedInputStream.java
```diff
@@ -0,0 +1,162 @@
...
+        for (Pair<byte[], byte[]> ext : exts) {
+            outputStream.write((byte) ';');
+            outputStream.write(ext.left());
+            outputStream.write((byte) '=');
+            outputStream.write(ext.right());
+        }
+    }
+
+    private void writeTrailers(byte[] chunk, ByteArrayOutputStream outputStream) throws IOException {
+        List<Pair<byte[], byte[]>> tlrs = trailers.stream().map(tlr -> tlr.get(chunk)).collect(Collectors.toList());
```
zoewangg, line 141: Same here

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/io/ChunkInputStream.java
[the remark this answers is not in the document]
zoewangg: The general guideline is that new code should have 80%+ test coverage.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/chunkedencoding/ChunkEncodedInputStream.java
```diff
@@ -0,0 +1,162 @@
...
+import software.amazon.awssdk.utils.Logger;
+import software.amazon.awssdk.utils.Pair;
+
+
+/**
+ * An implementation of chunked transfer encoding, which wrap an @{link InputStream}.
+ * This implementation supports chunk-headers, chunk-extensions, and trailers.
+ */
+@SdkProtectedApi
+public final class ChunkEncodedInputStream extends SdkInputStream {
```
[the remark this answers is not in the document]
haydenbaker: I'm not sure I agree with that naming. If anything, I feel like `SdkInputStream` should be renamed (`ReleasableInputStream`?).

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/chunkedencoding/ChunkEncodedInputStream.java
haydenbaker: No, that method is implemented in `InputStream`.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/chunkedencoding/Trailer.java
haydenbaker: I didn't include `Chunk` in the naming, because it isn't really part of a chunk (like an extension or header is). It's the final bits of data sent ("footer" or "trailing" data).

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/chunkedencoding/Chunk.java
```diff
@@ -0,0 +1,78 @@
...
+     * Get a default implementation of a chunk, which wraps a stream with a fixed size;
+     */
+    static Chunk create(InputStream data, int size) {
+        return new ChunkImpl(new ChunkInputStream(data, size));
+    }
+
+    /**
+     * Get the underlying stream of data for a chunk.
+     */
+    InputStream stream() throws IOException;
```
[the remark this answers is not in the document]
haydenbaker: This was likely a mistake on my part.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/chunkedencoding/ChunkEncodedInputStream.java
```diff
@@ -0,0 +1,162 @@
...
+        this.extensions.addAll(extensions);
+        this.trailers.addAll(trailers);
+    }
+
+    @Override
+    protected InputStream getWrappedInputStream() {
+        return inputStream;
+    }
+
+    public int getChunkSize() {
```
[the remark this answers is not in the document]
haydenbaker: Removed. Used it previously, but ended up not needing it.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/chunkedencoding/ChunkEncodedInputStream.java
```diff
@@ -0,0 +1,162 @@
...
+        if (currentChunk == null || currentChunk.ended() && !isFinished) {
+            readChunk(inputStream);
+        }
+
+        return currentChunk.stream().read();
+    }
+
+    private void readChunk(InputStream stream) throws IOException {
+        LOG.debug(() -> "Reading next chunk.");
+        if (currentChunk != null) {
```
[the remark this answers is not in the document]
haydenbaker: To close the chunk.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/chunkedencoding/ChunkEncodedInputStream.java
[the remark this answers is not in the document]
haydenbaker: Sync'd offline. Solution is to not rely on `read(byte b[], int off, int len)`, and instead read in a loop until we fill the chunk or we reach end-of-stream.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/chunkedencoding/ChunkEncodedInputStream.java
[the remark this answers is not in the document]
haydenbaker: Removed this since unused.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/io/ChunkInputStream.java
```diff
@@ -0,0 +1,44 @@
...
+
+    public ChunkInputStream(InputStream inputStream, long length) {
+        super(inputStream, length);
+    }
+
+    @Override
+    public void close() throws IOException {
+        // Drain this chunk on close, so the stream is left at the end of the chunk.
+        long remaining = remaining();
```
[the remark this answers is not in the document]
haydenbaker: In the event that this method is used, it ensures that it's backing stream's position is set to the logical end of the chunk (which is why we use SdkLengthAwareInputStream). In `ChunkedEncodedInputStream`, we read until the chunk doesn't have anything remaining, so it's somewhat redundant.

core/http-auth-aws/src/test/java/software/amazon/awssdk/http/auth/aws/chunkedencoding/ChunkEncodedInputStreamTest.java
```diff
@@ -0,0 +1,225 @@
...
+
+        assertEquals(expectedBytesRead, bytesRead);
+        assertArrayEquals(expected, actual);
+    }
+
+    @Test
+    public void ChunkEncodedInputStream_withTrailers_returnsEncodedChunksAndTrailerChunk() throws IOException {
+        byte[] data = "abcdefghij".getBytes();
+        InputStream payload = new ByteArrayInputStream(data);
+        int chunkSize = 3;
```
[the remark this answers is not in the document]
haydenbaker: I can add a few parameterized tests, good idea.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/chunkedencoding/Chunk.java
```diff
@@ -0,0 +1,73 @@
...
+        private final ChunkInputStream data;
+
+        public ChunkImpl(ChunkInputStream data) {
+            this.data = data;
+        }
+
+        @Override
+        public boolean hasEnded() {
+            return data.remaining() <= 0;
+        }
```
millems, line 61: Nit: rename to `hasRemaining` to match `ByteBuffer`?

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/chunkedencoding/Chunk.java
```diff
@@ -0,0 +1,73 @@
...
+    /**
+     * Whether the logical end of a chunk has been reached.
+     */
+    boolean hasEnded();
+
+    /**
+     * An implementation of a chunk, backed by a {@link ChunkInputStream}. This allows it to have awareness of its length and
+     * determine the endedness of the chunk.
+     */
+    final class ChunkImpl implements Chunk {
```
millems, line 51: Can we make this package protected? It's public here, since everything is public in an interface.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/chunkedencoding/ChunkExtension.java
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * A functional interface for defining an extension of a chunk, where the extension is a key-value pair.
+ * <p>
+ * The extension usually depends on the chunk-data itself (checksum, signature, etc.), but is not required to.
+ */
+@FunctionalInterface
+@SdkProtectedApi
+public interface ChunkExtension {
+    Pair<byte[], byte[]> get(byte[] chunk);
+}
```
millems, line 30: What IS an extension of a chunk? It's not entirely clear from the javadoc and/or method signature.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/chunkedencoding/ChunkHeader.java
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * A functional interface for defining a header of a chunk.
+ * <p>
+ * The header usually depends on the chunk-data itself (hex-size), but is not required to.
+ */
+@FunctionalInterface
+@SdkProtectedApi
+public interface ChunkHeader {
+    byte[] get(byte[] chunk);
```
millems, line 28: It's unclear how this method is used without reading other code. Can you give an example in the javadoc and some description?

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/chunkedencoding/ChunkedEncodedInputStream.java
```diff
@@ -0,0 +1,293 @@
...
+        // set the current chunk to the written final chunk
+        return getFinalChunk();
+    }
+
+    /**
+     * Read from an input-stream, up to a max number of bytes, storing them in a byte-array.
+     * The actual number of bytes can be less than the max in the event that we reach the end of the stream.
+     * <p>
+     * This method is necessary because we cannot assume the backing stream uses the default implementation of
+     * {@code read(byte b[], int off, int len)}
```
millems, line 103: It's also less efficient if the source stream is array-backed because it could do a `System.arraycopy`, which is faster than copying byte-by-byte.

Can we do this instead?
```
        int read = 0;
        int offset = 0;
        while (read > 0 && offset < maxBytesToRead) {
            read = inputStream.read(buf, offset, maxBytesToRead - offset);
            assert read != 0;
            offset += read;
        }

        return offset;
``` 

`InputStream.read(byte b[], int off, int len)` is guaranteed to return non-zero (assuming `len` is positive), so it *should* be equivalent, but faster for common source stream types, like `ByteArrayInputStream`.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/chunkedencoding/Chunk.java
```diff
@@ -0,0 +1,73 @@
...
+import java.io.InputStream;
+import software.amazon.awssdk.annotations.SdkProtectedApi;
+import software.amazon.awssdk.http.auth.aws.internal.io.ChunkInputStream;
+import software.amazon.awssdk.utils.SdkAutoCloseable;
+
+/**
+ * An interface which defines a "chunk" of data.
+ */
+@SdkProtectedApi
+public interface Chunk extends SdkAutoCloseable {
```
millems, line 29: This can be internal, because it's not returned anywhere, and it doesn't seem to make a lot of sense to create directly.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/chunkedencoding/ChunkedEncodedInputStream.java
```diff
@@ -0,0 +1,293 @@
...
+         * Set the size of chunks from input stream.
+         * The actual size (in bytes) of an encoded chunk depends on the configuration.
+         */
+        Builder chunkSize(int chunkSize);
+
+        /**
+         * Set the header to be used when creating an encoded chunk.
+         * This header will be the first part of an encoded chunk.
+         */
+        Builder header(ChunkHeader header);
```
millems, line 209: should this be called a ChunkHeaderFactory to make it clear it's not a constant value? Same with extensions and trailers?

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/io/SdkInputStream.java
```diff
@@ -0,0 +1,75 @@
...
+import java.io.InputStream;
+import org.slf4j.LoggerFactory;
+import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.utils.IoUtils;
+
+/**
+ * Base class for AWS Java SDK specific {@link InputStream}.
+ */
+@SdkInternalApi
+public abstract class SdkInputStream extends InputStream implements Releasable {
```
millems, line 28: Do we need this? I don't see anything using it, but I could be missing something.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/signer/RollingSigner.java
millems, line 29: What's this used for?

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/chunkedencoding/ChunkedEncodedInputStream.java
haydenbaker: I think "XyzProvider" would be a better name.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/io/SdkInputStream.java
haydenbaker: The `ChunkedEncodedInputStream` extends it.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/signer/RollingSigner.java
haydenbaker: For calculating signatures of chunks. Each chunk signature depends on the last.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/chunkedencoding/ChunkExtension.java
haydenbaker: I could have sworn I added more, but it must've got lost somewhere in wrangling commits.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/io/SdkInputStream.java
millems: I see that, but is it relying on any specific implementation here? The methods don't see to be called, for example.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/signer/RollingSigner.java
millems: I guess my question was more cosmic. It's only used in tests. What is it used for outside of tests?

haydenbaker: It will be used in the S3 signer

millems: Okay, I'll ignore it for now, then.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/io/SdkInputStream.java
haydenbaker: Yeah, fair. The existing implementation extends `SdkInputStream`, so kept it that way. I may end up calling `abortIfNeeded` in some places. What are your thoughts on that? Do we really need it?

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/signer/RollingSigner.java
haydenbaker: The test is an example of how AWS chunked encoding would be configured.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/io/SdkInputStream.java
millems: I'd leave it out until we need it.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/chunkedencoding/ChunkExtension.java
millems: Cool, LMK when it's updated and I'll re-review.

haydenbaker: It's updated

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/chunkedencoding/ChunkedEncodedInputStream.java
```diff
@@ -0,0 +1,303 @@
...
+ *                      last-chunk
+ *                      trailer-part
+ *                      CRLF
+ *     chunk          = chunk-size [ chunk-ext ] CRLF
+ *                      chunk-data CRLF
+ *     chunk-size     = 1*HEXDIG
+ *     last-chunk     = 1*("0") [ chunk-ext ] CRLF
+ *     chunk-data     = 1*OCTET ; a sequence of chunk-size octets
+ * </pre>
+ */
```
millems, line 50: It feels like this class is doing two things:
1. Mapping an unchunked stream to an chunked stream
2. Allowing setting of the chunk size, extensions and trailers during that transformation.

It feels like it would be better supported with two classes:
1. One to take any InputStream and convert it to a chunked stream
2. One to take a chunked-encoded input stream and transform the individual chunks

Without doing this split, we'd probably still need a separate class for taking a chunked stream and transforming the individual chunks.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/chunkedencoding/ChunkedEncodedInputStream.java
```diff
@@ -0,0 +1,303 @@
...
+public final class ChunkedEncodedInputStream extends InputStream {
+    private static final Logger LOG = Logger.loggerFor(ChunkedEncodedInputStream.class);
+    private static final byte[] CRLF = {'\r', '\n'};
+    private static final byte[] END = {};
+
+    private final InputStream inputStream;
+    private final int chunkSize;
+
+    private final ChunkHeader header;
+    private final List<ChunkExtension> extensions = new ArrayList<>();
```
millems, line 61: It's weird that we don't actually allow transforming the chunk body. We'll need that for compression.

Should we just combine these into a single transformer?
```java
@FunctionalInterface
interface ChunkTransformer {
    Chunk transform(Chunk chunk);
}

interface Chunk {
    byte[] length();
    List<byte[]> extensions();
    InputStream payload();

    static Chunk create(byte[] length, List<byte[]> extensions, InputStream payload) {
        // ...
    }
}
``` 

I keep `payload()` as an `InputStream` so that we don't have to copy the chunk data into a byte array if we're not going to transform it.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/chunkedencoding/ChunkedEncodedInputStream.java
```diff
@@ -0,0 +1,303 @@
...
+
+    /**
+     * Get an encoded chunk from the input-stream, or the final chunk if we've reached the end of the input-stream.
+     */
+    private Chunk getChunk(InputStream stream) throws IOException {
+        LOG.debug(() -> "Reading next chunk.");
+        if (currentChunk != null) {
+            currentChunk.close();
+        }
+        // we *have* to read from the backing stream in order to figure out if it's the end or not
```
millems, line 92: On the other hand, it's a big improvement in performance if we don't have to read the chunk data from the backing stream. It reduces the amount of copying of the data that we need to do.

Instead of reading the whole chunk to figure out if it's the end, can we wrap `inputStream` in a [`LookaheadInputStream`](https://github.com/aws/aws-sdk-java-v2/blob/master/utils/src/main/java/software/amazon/awssdk/utils/LookaheadInputStream.java) and peek() at the next `read()` result to see if it's `-1`? If it is, then do `getFinalChunk()`?

We can save this for a future pre-release task, but I'm being picky since this is a protected API.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/chunkedencoding/ChunkedEncodedInputStream.java
```diff
@@ -0,0 +1,303 @@
...
+        int offset = 0;
+        do {
+            read = inputStream.read(buf, offset, maxBytesToRead - offset);
+            assert read != 0;
+            if (read > 0) {
+                offset += read;
+            }
+        } while (read > 0 && offset < maxBytesToRead);
+
+        return offset;
```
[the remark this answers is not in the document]
haydenbaker: Yeah, it won't. It will return 0 in the event that the backing stream is empty, which is fine, since the caller checks for > 0.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/chunkedencoding/ChunkedEncodedInputStream.java
haydenbaker: It's doing (mostly) what the chunked-encoding spec says. Transformation of the body isn't part of chunk transfer encoding.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/chunkedencoding/ChunkedEncodedInputStream.java
haydenbaker: I'm not convinced we would use this class for compression, I think it should be used for chunk transfer encoding only. If a request is to be compressed and chunk encoded, it would make sense that it must be compressed first and then chunk encoded. The other way doesn't make sense - if it was encoded and then compressed, you'd have to encode it AGAIN in order to use chunk transfer encoding.

Either way, from what I understand, compressed+chunked can be either `Content-Encoding: gzip, Transfer-Encoding: gzip` or `Transfer-Encoding: gzip, chunked` depending on the intentions, but in both cases, the data should be compressed first and then chunked. To accomplish that, we'd wrap the compressed input stream with the chunked-encoding input stream.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/chunkedencoding/ChunkedEncodedInputStream.java
haydenbaker: I don't quite understand. We have to read from the stream to use it when transforming the chunk (the whole chunk of data needs to be there, i.e. when chunk-signature is calculated). It's not clear to me (right now) what benefit a `LookaheadInputStream` would bring in this case?

millems: Maybe I'm not thinking it through enough and I need to see this being used in the real world to understand. Thanks for your patience.

My point was that we shouldn't need the whole chunk in memory to calculate a checksum that is needed in the final trailer. We could theoretically provide each byte to the checksum implementation as it is read out of the input stream.

With this CR, when someone reads data from the ChunkedEncodingInputStream, the data gets copied twice: Once on line 94 into an in-memory buffer specific to the chunked encoding input stream and again into whatever buffer the result of read() is written into.

If we defer that until read() and we use the target buffer that the caller provides (assuming we override the other overloads of read(), which we should for performance), we can just copy the data once: into the result buffer.

We don't necessarily need to implement that right now, since what we have works, but since it's a protected API we need for it to be possible. I don't think it is right now, since everywhere the chunk is provided it's already been read into memory so there's no chance we could fix that later.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/chunkedencoding/ChunkedEncodedInputStream.java
millems: So we'd need a separate class to do any transformation?

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/chunkedencoding/ChunkedEncodedInputStream.java
millems: That makes sense if the last thing we'd do before the payload is frozen is to load it up into a chunked encoded stream. Unfortunately that's not always the case. There may be things upstream of the caller that has already chunked it, and we have to transform the already-chunked stream.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/chunkedencoding/ChunkedEncodedInputStream.java
haydenbaker: Yeah, I would say so. Transforming the body *within* this implementation *could* be troublesome. If the body is transformed, then the header/extensions/trailers need some kind of indication of whether they should use the original data or the transformed data. For example, if a chunk is transformed, but the header uses the original to calculate a size, then it may be incorrect. Or if there are multiple transformations, which to use when writing the header/extension/trailers - sounds complex.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/chunkedencoding/ChunkedEncodedInputStream.java
haydenbaker: That's a good point, but that might need to be a requirement we have.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/chunkedencoding/ChunkedEncodedInputStream.java
haydenbaker: I'll rename the classes to '*Provider'.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/chunkedencoding/ChunkedEncodedInputStream.java
haydenbaker: Discussed offline. We'll table this for now, add a TODO, as we will keep these classes internal so that the implementations can be changed around.

millems (approved):

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/chunkedencoding/ChunkInputStream.java
```diff
@@ -0,0 +1,45 @@
...
+    public void close() throws IOException {
+        // Drain this chunk on close, so the stream is left at the end of the chunk.
+        long remaining = remaining();
+        if (remaining > 0) {
+            if (skip(remaining) < remaining) {
+                throw new IOException("Unable to drain stream for chunk. The underlying stream did not allow skipping the "
+                                      + "whole chunk.");
+            }
+        }
+        super.close();
```
zoewangg, line 43: Do we need to close the wrapped inputstream?

zoewangg (approved):

PR #4270, opened 2023-08-07
Implement payload signing for aws event-streams

haydenbaker:
- Incorporates event-stream payload signer in the delegation in the main AwsV4HttpSigner
- Rename event-stream module from http-auth-event-stream to http-auth-aws-event-stream to better fit its use

## Motivation and Context
- Enables http-signing of event-stream payloads through the use of the AwsV4HttpSigner

## Modifications
- Rename module, update downstream POMs
- Adds an implementation of V4PayloadSigner to support http-signing of event-stream requests
- Adds code to dynamically load payload signer in http-auth-aws module 
- Adds tests

## Testing
- Run both new and old tests, verify passing
- TCK tests are added for the SigV4 Publisher, which directly uses TrailingFrame Publisher so coverage is there

## Types of changes

## Checklist

## License


core/http-auth-aws-event-stream/src/main/java/software/amazon/awssdk/http/auth/aws/eventstream/internal/SigV4DataFramePublisher.java
haydenbaker, line 1: This is mostly copied from previous signer.

core/http-auth-aws-event-stream/src/main/java/software/amazon/awssdk/http/auth/aws/eventstream/internal/TrailingDataFramePublisher.java
haydenbaker, line 1: This is mostly copied from old signer.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/signer/V4PayloadSigner.java
```diff
@@ -40,7 +39,7 @@ static V4PayloadSigner create() {
     ContentStreamProvider sign(ContentStreamProvider payload, V4Context v4Context);
 
     /**
-     * Given a payload and a future containing a v4-context, sign the payload via the SigV4 process.
+     * Given a payload and v4-context, sign the payload via the SigV4 process.
      */
-    Publisher<ByteBuffer> sign(Publisher<ByteBuffer> payload, CompletableFuture<V4Context> futureV4Context);
+    Publisher<ByteBuffer> sign(Publisher<ByteBuffer> payload, V4Context v4Context);
```
gosar, line 44: might call this `signAsync` following pattern in the HttpSigner interface

core/http-auth-aws-event-stream/src/main/java/software/amazon/awssdk/http/auth/aws/eventstream/internal/SigV4DataFramePublisher.java
```diff
@@ -0,0 +1,233 @@
...
+                 * the credential scope with the new instant every time the data-frame signer is called
+                 */
+                CredentialScope updatedCredentialScope = new CredentialScope(credentialScope.getRegion(),
+                                                                             credentialScope.getService(), signingInstant);
+                byte[] signingKey = deriveSigningKey(credentials, updatedCredentialScope);
+
+                /**
+                 * Calculate rolling signature
+                 */
+                byte[] payload = byteBuffer.array();
```
millems, line 107: This feels wrong. What if the provided byte buffer doesn't have an array? What if the sender only wants us to read a portion of the data?

The "safe" and "right" thing is to do a `byte[] payload = copyBytesFrom(byteBuffer)`. Can we test doing that? The upstream might need to be updated to ensure the byte buffer position is at the start and the limit is at the end.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/util/SignerUtils.java
```diff
@@ -201,30 +186,57 @@ public static void addDateHeader(SdkHttpRequest.Builder requestBuilder, String d
...
+        } catch (Exception e) {
+            throw new RuntimeException("Unable to read request payload to sign request: " + e.getMessage());
+        }
+    }
 
-        if (sdkChecksum != null && !isUnsignedPayload(contentHashString)) {
-            requestBuilder.putHeader(checksumHeaderName, BinaryUtils.toBase64(sdkChecksum.getChecksumBytes()));
+    public static byte[] hash(InputStream input) {
+        try {
+            MessageDigest md = getMessageDigestInstance();
```
millems, line 210: Can we add a TODO: performance test to see if we should cache these message digest instances in a thread local, like we do in the sigv4 signer?

core/http-auth-aws-event-stream/src/main/java/software/amazon/awssdk/http/auth/aws/eventstream/internal/SigV4DataFramePublisher.java
haydenbaker: Yeah, that's a good point. This was copied over, but I'll make the change.

core/http-auth-aws-event-stream/src/main/java/software/amazon/awssdk/http/auth/aws/eventstream/internal/SigV4DataFramePublisher.java
```diff
@@ -0,0 +1,233 @@
...
+
+    private final Publisher<ByteBuffer> sigv4Publisher;
+
+    public SigV4DataFramePublisher(Publisher<ByteBuffer> publisher,
+                                   AwsCredentialsIdentity credentials,
+                                   CredentialScope credentialScope,
+                                   String signature,
+                                   Clock signingClock) {
+
+        // Adapt the publisher with a trailing-empty frame publisher
```
zoewangg, line 61: This comment is not needed, same as line 64. We should minimize the use of comments by making the code self-explanatory.

core/http-auth-aws-event-stream/src/main/java/software/amazon/awssdk/http/auth/aws/eventstream/AwsV4EventStreamHttpSigner.java
```diff
@@ -33,6 +33,6 @@ public interface AwsV4EventStreamHttpSigner extends HttpSigner<AwsCredentialsIde
      * @return AwsV4EventStreamHttpSigner
      */
     static AwsV4EventStreamHttpSigner create() {
-        return new DefaultAwsV4EventStreamHttpSigner();
+        return null;
```
zoewangg, line 36: Hmm, do we intend to return null?

core/http-auth-aws-event-stream/src/main/java/software/amazon/awssdk/http/auth/aws/eventstream/internal/SigV4DataFramePublisher.java
zoewangg: NVM, seems like this is copied from existing file.

core/http-auth-aws-event-stream/src/main/java/software/amazon/awssdk/http/auth/aws/eventstream/signer/EventStreamV4PayloadSigner.java
```diff
@@ -0,0 +1,60 @@
...
+ * An implementation which supports async signing of event-stream payloads.
+ */
+@SdkProtectedApi
+public class EventStreamV4PayloadSigner implements V4PayloadSigner {
+
+    private final AwsCredentialsIdentity credentials;
+    private final CredentialScope credentialScope;
+    private final Clock signingClock;
+
+    public EventStreamV4PayloadSigner(AwsCredentialsIdentity credentials, CredentialScope credentialScope, Clock signingClock) {
```
zoewangg, line 39: Can we use Builder so that we can easily add more parameters in the future if needed?  Also, we should validate null here. I'd expect if signingClock is not provided, we should use the system default one.

core/http-auth-aws-event-stream/src/main/java/software/amazon/awssdk/http/auth/aws/eventstream/internal/SigV4DataFramePublisher.java
```diff
@@ -0,0 +1,234 @@
...
+@SdkInternalApi
+public final class SigV4DataFramePublisher implements Publisher<ByteBuffer> {
+
+    private static final Logger LOG = Logger.loggerFor(SigV4DataFramePublisher.class);
+    private static final String CHUNK_SIGNATURE = ":chunk-signature";
+    private static final int PAYLOAD_TRUNCATE_LENGTH = 32;
+
+    private final Publisher<ByteBuffer> sigv4Publisher;
+
+    public SigV4DataFramePublisher(Publisher<ByteBuffer> publisher,
```
zoewangg, line 55: Can we use Builder?

core/http-auth-aws-event-stream/src/main/java/software/amazon/awssdk/http/auth/aws/eventstream/internal/SigV4DataFramePublisher.java
```diff
@@ -0,0 +1,233 @@
...
+        sb.append("}, payload=");
+
+        byte[] payload = m.getPayload();
+        byte[] payloadToLog;
+
+        // We don't actually need to truncate if the payload length is already within the truncate limit
+        truncatePayload = truncatePayload && payload.length > PAYLOAD_TRUNCATE_LENGTH;
+
+        if (truncatePayload) {
+            // Would be nice if BinaryUtils.toHex() could take an array index range instead so we don't need to copy
```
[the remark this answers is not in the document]
zoewangg: NVM if this is copied from existing file.

core/http-auth-aws-event-stream/src/main/java/software/amazon/awssdk/http/auth/aws/eventstream/internal/TrailingDataFramePublisher.java
```diff
@@ -0,0 +1,114 @@
...
+import org.reactivestreams.Publisher;
+import org.reactivestreams.Subscriber;
+import org.reactivestreams.Subscription;
+import software.amazon.awssdk.annotations.SdkInternalApi;
+
+/**
+ * A publisher which wraps a publisher and appends a trailing frame once the wrapped publisher is finished.
+ */
+@SdkInternalApi
+public final class TrailingDataFramePublisher implements Publisher<ByteBuffer> {
```
zoewangg, line 30: I'll need to take a deeper look at this class, but we need tck tests for reactive streams implementations. https://github.com/aws/aws-sdk-java-v2/blob/master/CONTRIBUTING.md#reactive-streams

core/http-auth-aws-event-stream/src/main/java/software/amazon/awssdk/http/auth/aws/eventstream/AwsV4EventStreamHttpSigner.java
haydenbaker: Yes, this interface should go away soon.

core/http-auth-aws-event-stream/src/main/java/software/amazon/awssdk/http/auth/aws/eventstream/signer/EventStreamV4PayloadSigner.java
haydenbaker: We don't want any defaults, the caller should provide the default. The caller should also only ever be the AwsV4Signer. Unfortunately, it's in a different module, so this needs to be sdk-protected.

core/http-auth-aws-event-stream/src/main/java/software/amazon/awssdk/http/auth/aws/eventstream/internal/SigV4DataFramePublisher.java
haydenbaker: Since these are internal, and shouldn't leave this module, I don't see the need.

core/http-auth-aws-event-stream/src/main/java/software/amazon/awssdk/http/auth/aws/eventstream/internal/TrailingDataFramePublisher.java
zoewangg: We need TCK test.https://github.com/reactive-streams/reactive-streams-jvm/tree/master/tck

core/http-auth-aws-event-stream/src/main/java/software/amazon/awssdk/http/auth/aws/eventstream/signer/EventStreamV4PayloadSigner.java
zoewangg: Okay, then we should validate null here.

core/http-auth-aws-event-stream/src/main/java/software/amazon/awssdk/http/auth/aws/eventstream/internal/SigV4DataFramePublisher.java
zoewangg: Methods with 3+ parameters are:
- hard to maintain, and it will worsen if we add more in the future
- error prone. If we add another String parameter, say "serviceName" in the future, the caller could easily incorrectly misplace it
- hard to read and use 
```

        Publisher<ByteBuffer> sigV4Publisher = new SigV4DataFramePublisher(
            testVector.payload(),
            credentials,
            credentialScope,
            testVector.signature(),
            signingClock
        );
```
vs

```
SigV4DataFramePublisher.builder()
                                 .publisher(testVector.payload())
                                 .credentialsIdentity(credentials)
                                 .credentialScope(scope)
                                 .signature(signature)
                                 .signingClock(clock)
                                 .build();
```

core/http-auth-aws-event-stream/src/main/java/software/amazon/awssdk/http/auth/aws/eventstream/internal/SigV4DataFramePublisher.java
```diff
@@ -0,0 +1,233 @@
...
+                Map<String, HeaderValue> headers = new HashMap<>(eventHeaders);
+                //Signature headers
+                headers.put(CHUNK_SIGNATURE, HeaderValue.fromByteArray(signatureBytes));
+
+                /**
+                 * Wrap payload and headers in a Message object and then encode to bytes
+                 */
+                Message signedMessage = new Message(sortHeaders(headers), payload);
+
+                if (LOG.isLoggingLevelEnabled("trace")) {
```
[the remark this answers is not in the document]
haydenbaker: I believe the check is there to conditionally print out the fully formatted message+payload if trace is enabled.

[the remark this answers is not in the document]
zoewangg: The SDK Logger class already does that

https://github.com/aws/aws-sdk-java-v2/blob/13887532e50932bb3a93680884e94bd087a58abb/utils/src/main/java/software/amazon/awssdk/utils/Logger.java#L125-L127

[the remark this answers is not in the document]
haydenbaker: It's used when formatting the message: `toDebugString(signedMessage, false));`

haydenbaker:
Looks like CI is failing due to a few flaky tests.

haydenbaker:
Merging to unblock other changes. Will address new comments (that are to be posted) in follow-up.

core/http-auth-aws-event-stream/src/main/java/software/amazon/awssdk/http/auth/aws/eventstream/internal/SigV4DataFramePublisher.java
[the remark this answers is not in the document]
zoewangg: Right, toDebugString will only be invoked if trace is enabled with SDK logger.

core/http-auth-aws-event-stream/src/main/java/software/amazon/awssdk/http/auth/aws/eventstream/signer/EventStreamV4PayloadSigner.java
```diff
@@ -0,0 +1,61 @@
...
+import software.amazon.awssdk.http.auth.aws.signer.CredentialScope;
+import software.amazon.awssdk.http.auth.aws.signer.V4Context;
+import software.amazon.awssdk.http.auth.aws.signer.V4PayloadSigner;
+import software.amazon.awssdk.identity.spi.AwsCredentialsIdentity;
+import software.amazon.awssdk.utils.Validate;
+
+/**
+ * An implementation which supports async signing of event-stream payloads.
+ */
+@SdkProtectedApi
```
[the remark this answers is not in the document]
zoewangg: I don't think so. If we break backwards compatibility in this class, it may still break customers.

core/http-auth-aws-event-stream/src/main/java/software/amazon/awssdk/http/auth/aws/eventstream/signer/EventStreamV4PayloadSigner.java
zoewangg: Let's use builder since this is a protected API to give us more flexibility to add new parameters in the future.

PR #4280, opened 2023-08-09
Remove SignerOverride related codegen logic

gosar:
## Motivation and Context

Currently, `Client`s are setup with a default signer, which is overridden for certain operations:
* unsigned payload
* event streaming
* async v4 streaming 
* have different authType than service.

With SRA, we do not have separate signers, the AwsV4HttpSigner interface supports all the sigv4 use cases, so this type of signer override is not needed in generated clients. And different authType than service (e.g., bearerToken, as opposed to sigv4) is supported by AuthSchemeProvider logic which is operation aware.

Note, customer can still override the default `Signer` at the client/request level, which will result in SIGNER_OVERRIDDEN=true, and that will continue to be supported at that's a public interface (but will be marked deprecated) and old clients may still use it. This PR just takes out the concept of automatic signer overrides within the client.

Note, old `Signer` itself isn't needed to be configured on the client as the default signer too, as with SRA, signers are configured via AuthSchemes, which is already added to generated clients. But will remove those in a separate PR.

## Modifications

Delete a bunch of unneeded code! IMO, that code was also complex and icky - if an operation needed signer different from client, it was setup by abusing RequestOverrideConfiguration. It's a good outcome to delete that code!

## Testing

Manually integ tested:
* acm - vanilla sigv4 case (no signer override), still works with updated generated client
* glacier.uploadArchive - sigv4 case with signer override for async streaming (AsyncAws4Signer)
Haven't tested unsigned payload/event streaming explicitly, but they fall under similar case as glacier.

### DIFF
--- a/codegen/src/main/java/software/amazon/awssdk/codegen/IntermediateModelBuilder.java
+++ b/codegen/src/main/java/software/amazon/awssdk/codegen/IntermediateModelBuilder.java
@@ -37,7 +37,6 @@ import software.amazon.awssdk.codegen.model.intermediate.MemberModel;
 import software.amazon.awssdk.codegen.model.intermediate.OperationModel;
 import software.amazon.awssdk.codegen.model.intermediate.ShapeModel;
 import software.amazon.awssdk.codegen.model.rules.endpoints.EndpointTestSuiteModel;
-import software.amazon.awssdk.codegen.model.service.AuthType;
 import software.amazon.awssdk.codegen.model.service.EndpointRuleSetModel;
 import software.amazon.awssdk.codegen.model.service.Operation;
 import software.amazon.awssdk.codegen.model.service.Paginators;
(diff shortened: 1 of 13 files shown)


codegen/src/main/java/software/amazon/awssdk/codegen/IntermediateModelBuilder.java
```diff
@@ -204,44 +202,6 @@ private void linkOperationsToInputOutputShapes(IntermediateModel model) {
...
-                 linkAuthorizationToRequestShapeForAwsProtocol(c2jOperation.getAuthtype(), shape);
-             });
-    }
-
-    private void linkAuthorizationToRequestShapeForAwsProtocol(AuthType authType, ShapeModel shape) {
-        if (authType == null) {
-            return;
-        }
-
-        switch (authType) {
```
cenedhryn, line 228: Yeah, this feels good to delete

cenedhryn (approved):

sugmanue (approved):

PR #4285, opened 2023-08-09
Set the correct signer properties for v4-unsigned-body

sugmanue:
## Motivation and Context

`v4-unsigned-body` auth type was not properly setting the correct signer properties, in particular setting `PAYLOAD_SIGNING_ENABLED` to `false`.

For S3 the `ModeledS3AuthSchemeProvider#resolveAuthScheme` looks like this after this change:

```java
    @Override
        switch (params.operation()) {
        case "WriteGetObjectResponse":
                    .putSignerProperty(AwsV4HttpSigner.PAYLOAD_SIGNING_ENABLED, false).build());
            break;
        default:
            break;
        }
    }
```

## Testing

* Added unit tests

## Types of changes

## Checklist

## License

### DIFF
--- a/codegen/src/main/java/software/amazon/awssdk/codegen/poet/auth/scheme/AuthSchemeCodegenMetadata.java
+++ b/codegen/src/main/java/software/amazon/awssdk/codegen/poet/auth/scheme/AuthSchemeCodegenMetadata.java
@@ -15,10 +15,11 @@
 
 package software.amazon.awssdk.codegen.poet.auth.scheme;
 
+import com.squareup.javapoet.MethodSpec;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
-import java.util.function.Function;
+import java.util.function.BiConsumer;
 import software.amazon.awssdk.codegen.model.service.AuthType;
 import software.amazon.awssdk.core.SelectedAuthScheme;
 import software.amazon.awssdk.http.auth.BearerAuthScheme;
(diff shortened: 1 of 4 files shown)


dagnir (approved):

codegen/src/main/java/software/amazon/awssdk/codegen/poet/auth/scheme/AuthSchemeCodegenMetadata.java
```diff
@@ -33,44 +34,38 @@ public final class AuthSchemeCodegenMetadata {
...
+        SIGV4.toBuilder()
+             .addProperty(SignerPropertyValueProvider.builder()
+                                                     .containingClass(AwsV4HttpSigner.class)
+                                                     .fieldName(
+                                                         "PAYLOAD_SIGNING_ENABLED")
+                                                     .valueEmitter((spec, utils) -> spec.addCode("$L", false))
+                                                     .build())
+             .build();
+
+    static final AuthSchemeCodegenMetadata S3 =
```
gosar, line 56: S3 by default needs to actually have PAYLOAD_SIGNING_ENABLED=false. Based on https://github.com/aws/aws-sdk-java-v2/blob/4ea3cc28a0825fa101547fd82cfb2b5767b040d9/core/auth/src/main/java/software/amazon/awssdk/auth/signer/params/AwsS3V4SignerParams.java#L91. This is not model driven, but desired behavior of the S3 SDK.

gosar: Note this is only for `s3`, `s3control` (which maps to authType=s3v4) is PAYLOAD_SIGNING_ENABLED=true (the default, like all Sigv4 services), based on https://github.com/aws/aws-sdk-java-v2/blob/4ea3cc28a0825fa101547fd82cfb2b5767b040d9/services/s3control/src/main/java/software/amazon/awssdk/services/s3control/internal/interceptors/PayloadSigningInterceptor.java#L36

gosar: Note, the s3 model has v4-unsigned-body explicitly in the model for `WriteGetObjectResponse`, which, I'm not sure why, but may mean that this operation _has_ to use unsigned payload, while for other S3 operations, both signed and unsigned work and we just want the SDK to default to unsigned.

sugmanue: > S3 by default needs to actually have PAYLOAD_SIGNING_ENABLED=false. Based on
> 
> https://github.com/aws/aws-sdk-java-v2/blob/4ea3cc28a0825fa101547fd82cfb2b5767b040d9/core/auth/src/main/java/software/amazon/awssdk/auth/signer/params/AwsS3V4SignerParams.java#L91
> 
> . This is not model driven, but desired behavior of the S3 SDK.

Looks like the logic is more involved than that, the javadocs says

> Payload signing is optional when chunked encoding is not used and requests are made against an HTTPS endpoint.  Under these conditions the client will by default opt to not sign payloads to optimize performance.  If this flag is set to true the client will instead always sign payloads.

I don't think we should do that unconditionally but might be wrong about it.

PR #4286, opened 2023-08-09
Remove Signer related codegen logic in Clients

gosar:
## Motivation and Context

Currently, Clients are setup with a default `Signer`. With SRA, the new signer interface HttpSigner is used and these signers are configured via AuthSchemes, which is already added to generated clients. So removing codegen to setup default `Signer.

Related: In #4280 setting up Signer overrides was already removed.

Note, customer can still provide old `Signer` at the client/request level, which will result in SIGNER_OVERRIDDEN=true, and that will continue to be supported at that's a public interface (but will be marked deprecated).

This PR deletes codegen logic to _write_ the default signer into clients. Logic to _read_ the default signer is in aws-core (AwsExecutionContextBuilder, AuthorizationStrategy, SigningStage, etc.) which needs to remain to support old clients as well as customer provided overrides.

I thought more tests would fail, especially related to flexible checksums since it is dependent on AuthorizationStrategy returning the `Signer` which `resolveSigningMethodUsed` uses. But those tests aren't failing because of some endpoints 2.0 related logic, which I have another task for. EndpointAuthSchemeInterceptor is still setting authScheme and thus `Signer` which is causing SigningStage to use pre-SRA path. It may also be related that the tests in codegen-generated-classes-test only test the SigningMethod.UNSIGNED_PAYLOAD case. 

## Modifications

Delete more code!

## Testing

Manually integ tested:
* acm - vanilla sigv4 case still works with updated generated client
* codecatalyst - vanilla bearer token still works with updated generated client

Waiting to see if CI passes.

### DIFF
--- a/codegen/src/main/java/software/amazon/awssdk/codegen/poet/client/traits/NoneAuthTypeRequestTrait.java
+++ b/codegen/src/main/java/software/amazon/awssdk/codegen/poet/client/traits/NoneAuthTypeRequestTrait.java
@@ -26,6 +26,9 @@ import software.amazon.awssdk.core.interceptor.SdkInternalExecutionAttribute;
  * Trait which defines if a given request needs to be authenticated.
  * A request is not authenticated only if it has "auththpe" trait  explicitly marked as "none"
  */
+// TODO(sra-identity-and-auth): I think this class can be deleted. Old clients will still have the writer for this generated
+//  and we have to leave the reader of this attribute around for them (in AwsExecutionContextBuilder), but the writer in new
+//  clients can be deleted.
 @SdkInternalApi
 public class NoneAuthTypeRequestTrait {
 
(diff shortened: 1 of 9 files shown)


codegen/src/main/java/software/amazon/awssdk/codegen/poet/client/traits/NoneAuthTypeRequestTrait.java
```diff
@@ -26,6 +26,9 @@
  * Trait which defines if a given request needs to be authenticated.
  * A request is not authenticated only if it has "auththpe" trait  explicitly marked as "none"
  */
+// TODO(sra-identity-and-auth): I think this class can be deleted. Old clients will still have the writer for this generated
```
gosar, line 29: I was a little hesitant to remove this and wanted someone to review if this end state is ok (code base would only have a reader for this attribute and no write). But I'm going to go ahead and remove this class in next revision. If anyone has concerns, let me know.

sugmanue (approved):

cenedhryn (approved):

PR #4289, opened 2023-08-10
Fixed the issue where thresholdInBytes is not the same as minimalPart…

zoewangg:
## Motivation and Context

Currently, if a custom minimalPartSizeInBytes is provided, the default thresholdInBytes will still be 8MB. This is inconsistent with AWS CRT-based S3 client.

### Modifications

Updated the default `thresholdInBytes` to always be the same as `minimalPartSizeInBytes`

## Testing
Added tests

## Types of changes

## Checklist

## License


services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/multipart/MultipartConfigurationResolver.java
```diff
@@ -0,0 +1,53 @@
...
+import software.amazon.awssdk.utils.Validate;
+
+/**
+ * Internal utility class to resolve {@link MultipartConfiguration}.
+ */
+@SdkInternalApi
+public final class MultipartConfigurationResolver {
+
+    private static final long DEFAULT_MIN_PART_SIZE = 8L * 1024 * 1024;
+    private final MultipartConfiguration configuration;
```
L-Applin, line 29: we don't necessarily need to keep a reference to the configuration here

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/multipart/MultipartConfigurationResolver.java
```diff
@@ -0,0 +1,53 @@
...
+    private final long minimalPartSizeInBytes;
+    private final long apiCallBufferSize;
+    private final long thresholdInBytes;
+
+    public MultipartConfigurationResolver(MultipartConfiguration multipartConfiguration) {
+        this.configuration = multipartConfiguration;
```
L-Applin, line 35: should we add `Validate.notNull(multipartConfiguration)` ?

services/s3/src/main/java/software/amazon/awssdk/services/s3/multipart/MultipartConfiguration.java
```diff
@@ -93,13 +93,19 @@ public Long apiCallBufferSizeInBytes() {
...
-         * Configures the minimum number of bytes of the body of the request required for requests to be converted to their
-         * multipart equivalent. Only taken into account when converting {@code putObject} and {@code copyObject} requests.
-         * Any request whose size is less than the  configured value will not use multipart operation,
-         * even if multipart is enabled via {@link S3AsyncClientBuilder#multipartEnabled(Boolean)}.
+         * Configure the size threshold, in bytes, for when to use multipart upload. Uploads/copies over this size will
+         * automatically use a multipart upload strategy, while uploads/copies smaller than this threshold will use a single
+         * connection to upload/copy the whole object.
+         *
          * <p>
+         * Multipart uploads are easier to recover from and also potentially faster than single part uploads, especially when the
```
L-Applin, line 101: nice!

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/multipart/MultipartS3AsyncClient.java
```diff
@@ -46,32 +46,21 @@ public final class MultipartS3AsyncClient extends DelegatingS3AsyncClient {
...
-        long minPartSizeInBytes = Validate.getOrDefault(validConfiguration.minimumPartSizeInBytes(),
-                                                        () -> DEFAULT_MIN_PART_SIZE);
-        long threshold = Validate.getOrDefault(validConfiguration.thresholdInBytes(),
-                                               () -> DEFAULT_THRESHOLD);
-        long apiCallBufferSizeInBytes = Validate.getOrDefault(validConfiguration.apiCallBufferSizeInBytes(),
-                                                              () -> computeApiCallBufferSize(validConfiguration));
+        MultipartConfigurationResolver resolver = new MultipartConfigurationResolver(validConfiguration);
+        long minPartSizeInBytes = resolver.minimalPartSizeInBytes();
+        long threshold = resolver.thresholdInBytes();
+        long apiCallBufferSizeInBytes = resolver.apiCallBufferSize();
```
millems, line 59: Should we just modify the MultipartConfiguration in-place when the client is created? I'm worried about a future bug where someone ends up using the MultipartConfiguration instead of the MultipartConfigurationResolver.

zoewangg: The issue with that is if we are going to introduce a new way to configure settings, say, environment variables, we'd probably still need another layer to resolve it. This may also make it harder to integrate features like smart defaults because the default values depend on other settings that are not available in `MultipartConfiguration`. 
`MultipartConfiguration` returns boxed type, so hopefully that'd force the implementer to figure out they should use this class?

millems: I'm not so sure, but it seems like you've thought about it more than me. I'm fine with it, then.

millems (approved):

services/s3/src/test/java/software/amazon/awssdk/services/s3/internal/multipart/MultipartConfigurationResolverTest.java
```diff
@@ -0,0 +1,74 @@
...
+ */
+
+package software.amazon.awssdk.services.s3.internal.multipart;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.junit.jupiter.api.Test;
+import software.amazon.awssdk.services.s3.multipart.MultipartConfiguration;
+
+public class MultipartConfigurationResolverTest {
```
L-Applin, line 23: Should we add a test case for all empty values: `MultipartConfiguration.builder().build()`?

zoewangg: Sure, added.

L-Applin (approved):

PR #4292, opened 2023-08-11
Remove setting IS_NONE_AUTH_TYPE_REQUEST

gosar:
## Motivation and Context

Newly generated clients won't have this attribute set, but the reader (AwsExecutionContextBuilder) will still continue reading this attribute for old clients.

I'm not very sure if we want to go this route v/s continue to set this attribute for now. So making this a draft PR for discussion (see PR comment later). If we want to go in this direction, there is some refactoring I'll do to reduce code duplication, before getting this ready to ship.

## Modifications

* Remove codegen related to writing IS_NONE_AUTH_TYPE_REQUEST.
* Adjusted logic that used to read this attribute, to continue to work for old clients, as well as new clients with SRA. Ensures existing tests in codegen and codegen-generated-classes-test pass.

* With the changes, presign interceptor use cases break because they expect AWS_CREDENTIALS in execution attribute, which is not being set for SRA code path. So added a path to read credentials from SelectedAuthScheme for the SRA path. Note, currently it is still using the old Signer interface. Separately, we do want to move those to use the new HttpSigner interface.


core/aws-core/src/main/java/software/amazon/awssdk/awscore/internal/AwsExecutionContextBuilder.java
```diff
@@ -121,13 +123,40 @@ private AwsExecutionContextBuilder() {
...
-            AuthorizationStrategy authorizationStrategy =
-                authorizationStrategyFactory.strategyFor(executionParams.credentialType());
-            authorizationStrategy.addCredentialsToExecutionAttributes(executionAttributes);
-            signer = authorizationStrategy.resolveSigner();
+
+        // Note, not setting IS_NONE_AUTH_TYPE_REQUEST in newly generated clients, makes earlier logic not sufficient, as
+        // NoneAuthTypeRequestTest fails.
+        // It is easiest to keep NoneAuthTypeRequest passing by continuing to set IS_NONE_AUTH_TYPE_REQUEST in newly generated
+        // clients. This might become odd in the future, when authType=none is deprecated (authType itself will be less favored
+        // in lieu of authList, to support multiple auth schemes). But we can cross that bridge as part of those changes?
```
gosar, line 131: I'm not very sure if we want to go the direction of this PR v/s continue to set this attribute for now, i.e., discard this PR. No tests are currently broken, if we continue setting IS_NONE_AUTH_TYPE_REQUEST.

I thought it was natural to try to remove the setter (as explained a bit in the code comments in these lines), so started with that, but needed add the conditional branching logic below to make NoneAuthTypeRequestTest pass. My nervous ness with adding these conditions comes from hoping that I've covered all cases.

Please review this question before reviewing lower level code feedback.

gosar: Also, it seems this branching logic (in aws-core) has some relation with the `shouldDoSraSigning()` in SigningStage (sdk-core). I think that's ok, but wanted to bring attention to that for your consideration too.

cenedhryn: Agreed offline on keeping this attribute for backwards compatibility reasons for now.

cenedhryn: I'm not sure I completely understand 
> It is easiest to keep NoneAuthTypeRequest passing by continuing to set IS_NONE_AUTH_TYPE_REQUEST in newly generated clients.
Can we make a note to add compatibility tests for this feature in the cross-module testing that we're presumably planning?

cenedhryn: Additionally, can we remove these commented lines and if needed, add to one of the quip docs for the project as implementation notes etc.

core/aws-core/src/main/java/software/amazon/awssdk/awscore/internal/AwsExecutionContextBuilder.java
```diff
@@ -124,13 +126,40 @@ private AwsExecutionContextBuilder() {
...
+            // selectedAuthScheme.supportsSigning), in which case we want Signer to end up as `null`.
+            if (selectedAuthScheme.supportsSigning() && SignerOverrideUtils.isSignerOverridden(interceptorContext.request(),
+                                                                                               executionAttributes)) {
+                AuthorizationStrategyFactory authorizationStrategyFactory =
+                    new AuthorizationStrategyFactory(interceptorContext.request(), metricCollector, clientConfig);
+                AuthorizationStrategy authorizationStrategy =
+                    authorizationStrategyFactory.strategyFor(executionParams.credentialType());
+                authorizationStrategy.addCredentialsToExecutionAttributes(executionAttributes);
+                signer = authorizationStrategy.resolveSigner();
+            }
```
cenedhryn, line 162: Can we replace this with 

~~~
        if (usesOverriddenSigner(selectedAuthScheme, interceptorContext.request(), executionAttributes) ||
            authenticatedWithoutAuthSchemeSupport(selectedAuthScheme)) {
            AuthorizationStrategyFactory authorizationStrategyFactory =
                new AuthorizationStrategyFactory(interceptorContext.request(), metricCollector, clientConfig);
            AuthorizationStrategy authorizationStrategy =
                authorizationStrategyFactory.strategyFor(executionParams.credentialType());
            authorizationStrategy.addCredentialsToExecutionAttributes(executionAttributes);
            signer = authorizationStrategy.resolveSigner();
        }
~~~
?
Coalescing the different conditions into boolean methods with names that remove the need for long comments. 
Maybe changing the order of the conditions wasn't great, it could be better to do the null auth scheme first

core/aws-core/src/main/java/software/amazon/awssdk/awscore/internal/AwsExecutionContextBuilder.java
gosar: yup, added notes for testing and in implementation notes.

cenedhryn (approved):

PR #4303, opened 2023-08-15
Fix used and unused dependencies

sugmanue:
## Motivation and Context

Added or removed dependencies to clear all the warning related to use but undeclared, or unused but declared dependencies.

Given that codecatalyst does not use `http-auth-aws` this dependency was added to all the services but that one instead of just adding it to the `services/pom.xml` parent pom.

## Types of changes

## Checklist

## License

### DIFF
--- a/services/accessanalyzer/pom.xml
+++ b/services/accessanalyzer/pom.xml
@@ -56,5 +56,10 @@
             <artifactId>aws-json-protocol</artifactId>
             <version>${awsjavasdk.version}</version>
         </dependency>
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>http-auth-aws</artifactId>
+            <version>${awsjavasdk.version}</version>
+        </dependency>
     </dependencies>
 </project>
--- a/codegen/pom.xml
+++ b/codegen/pom.xml
@@ -79,7 +79,17 @@
         </dependency>
         <dependency>
             <groupId>software.amazon.awssdk</groupId>
+            <artifactId>identity-spi</artifactId>
+            <version>${awsjavasdk.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
             <artifactId>http-auth</artifactId>
+            <version>${awsjavasdk.version}</version>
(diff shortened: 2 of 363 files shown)


services/accessanalyzer/pom.xml
```diff
@@ -56,5 +56,10 @@
             <artifactId>aws-json-protocol</artifactId>
             <version>${awsjavasdk.version}</version>
         </dependency>
+        <dependency>
```
zoewangg, line 59: Instead of adding it to every service module, can we add it to this pom? https://github.com/aws/aws-sdk-java-v2/blob/master/services/pom.xml This way, if we need to change anything, say artifact ID, we can do it just in one place.

sugmanue: I added a note to the description about it

> Given that codecatalyst does not use http-auth-aws this dependency was added to all the services but that one instead of just adding it to the services/pom.xml parent pom.

So, we can add it to the `services/pom.xml` and live with one warning of unused dependency for `codecatalyst`, or we can just add it to all of the services but `codecatalyst` as done here and avoid having any warnings, thoughts?

zoewangg: My bad, should've read the description first 😅 Agreed we should add the dependency to each service module since not all services use them. The last thing we want is adding unused dependency since it'll increase JAR size.

zoewangg: We need to update this template though https://github.com/aws/aws-sdk-java-v2/blob/master/services/new-service-template/pom.xml This is used to create new service modules. We need to figure out how to add this dependency conditionally.

sugmanue: The template is already updated, we do not have at the moment a way to "opt-out" other than editing the resulting pom. I can take a look at that.

zoewangg: Discussed offline, new services will be addressed in another PR

zoewangg (approved):

services/accessanalyzer/pom.xml
sugmanue: Added on #4309

PR #4311, opened 2023-08-15
Remove legacy endpoint auth scheme logic and add putAuthScheme to AWS client builder.

millems:
Changes:
* Move endpoint resolution header addition from EndpointAuthSchemeInterceptor to EndpointResolverInterceptor
* Remove EndpointAuthSchemeInterceptor
* When endpoint provider returns signer parameters, override parameters from the auth scheme provider
* Add the ability to specify new or override existing auth schemes via putAuthScheme on AWS client builder.


codegen/src/main/java/software/amazon/awssdk/codegen/poet/auth/scheme/EndpointBasedAuthSchemeProviderSpec.java
```diff
@@ -123,7 +124,8 @@ private MethodSpec resolveAuthSchemeMethod() {
             }
         });
         spec.addStatement(".build()");
-        spec.addStatement("$T endpoint = DELEGATE.resolveEndpoint(endpointParameters).join()", Endpoint.class);
+        spec.addStatement("$T endpoint = $T.joinLikeSync(DELEGATE.resolveEndpoint(endpointParameters))",
+                          Endpoint.class, CompletableFutureUtils.class);
```
millems, line 128: Needed to fix some test cases in S3. It asserts on the exception type that's thrown, and that changed when we switched to the endpoint-based auth scheme providers without `joinLikeSync`.

utils/src/test/java/software/amazon/awssdk/utils/async/FlatteningSubscriberTest.java
```diff
@@ -205,8 +207,9 @@ public void requestsFromDownstreamDoNothingAfterOnError() {
     @Test
     public void stochastic_dataFlushedBeforeOnComplete() {
         ExecutorService exec = Executors.newSingleThreadExecutor();
+        Instant end = Instant.now().plus(10, SECONDS);
         try {
-            for (int i = 0; i < 30_000_000; ++i) {
+            while (Instant.now().isBefore(end)) {
```
haydenbaker, line 212: 🙌

haydenbaker: Is this reasonably long enough to catch random errors like 30m iterations would?

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/rules/endpoint-resolve-interceptor.java
```diff
@@ -24,22 +29,33 @@
...
+            List<EndpointAuthScheme> authSchemes = endpoint.attribute(AwsEndpointAttribute.AUTH_SCHEMES);
+            SelectedAuthScheme<?> selectedAuthScheme = executionAttributes
+                .getAttribute(SdkInternalExecutionAttribute.SELECTED_AUTH_SCHEME);
+            if (authSchemes != null && selectedAuthScheme != null) {
+                selectedAuthScheme = AwsEndpointProviderUtils.mergeIntoResolvedAuthScheme(authSchemes, selectedAuthScheme);
+                executionAttributes.putAttribute(SdkInternalExecutionAttribute.SELECTED_AUTH_SCHEME, selectedAuthScheme);
+            }
+            if (selectedAuthScheme != null) {
+                AuthSchemeUtils.setSigningParams(executionAttributes, selectedAuthScheme.authSchemeOption());
+            }
```
cenedhryn, line 56: Is this the only place that signing params are copied from the auth scheme to the execution attributes? It looks like it, but not sure.

utils/src/test/java/software/amazon/awssdk/utils/async/FlatteningSubscriberTest.java
millems: It's less likely to catch errors. How much less depends on the hardware. On my mac, I don't recall how long it took to run these tests before, but if it was 2 minutes then we have a 25% chance of catching an error that we would have caught with the previous code. E.g. if the previous code had a 0.01% chance of catching an error, now it's a 0.0025% chance on my mac. That varies based on the hardware, of course, since we switched from iterations to time. Faster hardware will have a higher chance of catching the errors.

Tests that run when a developer is present are expensive, because we pay for both the developer and hardware to run them. We need to make sure these are short. In this case, this test will miss errors that occur with a very low likelihood. That was true before, but now it's even more true.

These tests are good to build a base level of confidence that things won't immediately explode If they go to production, but if we need even more assurance we should move them to be tests that run when a developer is not present. Canaries, for example. Then it's just a hardware cost plus the maintenance cost of the canaries.

I don't think that's necessary right now since we have higher-level canary tests that also exercise this code path and we don't have reason to suspect this is a problematic code path and it's not likely to be changed anytime soon. If we do think it'll be changed more often, then it might be worth creating a dedicated canary for it to raise our chance of catching the bug.

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/builder/test-query-client-builder-class.java
```diff
@@ -116,17 +122,18 @@ private IdentityProvider<? extends TokenIdentity> defaultTokenProvider() {
...
     }
 
-    private Map<String, AuthScheme<?>> defaultAuthSchemes() {
+    private Map<String, AuthScheme<?>> authSchemes() {
         Map<String, AuthScheme<?>> schemes = new HashMap<>(2);
         AwsV4AuthScheme awsV4AuthScheme = AwsV4AuthScheme.create();
         schemes.put(awsV4AuthScheme.schemeId(), awsV4AuthScheme);
         BearerAuthScheme bearerAuthScheme = BearerAuthScheme.create();
         schemes.put(bearerAuthScheme.schemeId(), bearerAuthScheme);
+        schemes.putAll(this.additionalAuthSchemes);
```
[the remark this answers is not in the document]
sugmanue: Good point, I think is OK to leave as-is, the map will grow as needed.

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/rules/endpoint-resolve-interceptor.java
millems: Yes, to my knowledge it is.

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/builder/test-query-client-builder-class.java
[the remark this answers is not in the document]
sugmanue: I don't feel there's a need to split hairs for this but if we wanted to be "precise" we could change the codegen to:

```
-        builder.addStatement("$T schemes = new $T<>($L)", returns, HashMap.class, concreteAuthSchemeClasses.size());
+        builder.addStatement("$T schemes = new $T<>($L + this.additionalAuthSchemes.size())", returns, HashMap.class,
+                             concreteAuthSchemeClasses.size());
```

services/s3/src/test/java/software/amazon/awssdk/services/s3/functionaltests/PutObjectWithChecksumTest.java
```diff
@@ -35,6 +35,7 @@
 import java.util.function.Function;
 import java.util.stream.Collectors;
 import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.Disabled;
```
sugmanue, line 38: Interesting, I seem to remember trying this and checkstyle complaining about it.

sugmanue (approved):

services/s3/src/test/java/software/amazon/awssdk/services/s3/functionaltests/PutObjectWithChecksumTest.java
millems: Checkstyle complains about @Ignore! It doesn't mind @Disabled. That's probably a gap we should fix, but it's convenient right now!

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/builder/test-query-client-builder-class.java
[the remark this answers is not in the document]
millems: Fixed in the latest revision.

cenedhryn (approved):

joviegas (approved):

PR #4314, opened 2023-08-16
Changes default behavior for DynamoDb Enhanced atomic counter extension

cenedhryn:
## Motivation and Context
The initial implementation of the atomic counter extension made the assumption that it was not necessary to modify the item itself while adding an update expression, because users should not include attributes that were tagged as an atomic counter in the item to be updated and if they did, they would get an error back. 

For instance, if attribute `accessCounter` of the class `Balance` was marked with the `AtomicCounter` attribute, this would result in an error:
~~~
Balance balanceToUpdate = new Balance();
balanceToUpdate.accessCounter(4);
table.updateItem(balanceToUpdate);
~~~
Which makes sense since `accessCounter` will be updated anyway in the `AtomicCounterExtension`, and the assumption was that an error is better than silently changing the item when `updateItem` was called. This assumption did not take into account that a very common, and natural, behavior for users is read-modify-update:
~~~
Balance balanceToUpdate = table.getItem(key);
balanceToUpdate.anImportantAttribute(newValue);
table.updateItem(balanceToUpdate);
~~~
Here, the user will actually get the same error as above, because while they did not explicitly set `accessCounter`, the instance returned from DDB will have it set, and it's included when the user calls `updateItem`. 

Therefore, we should change this behavior to filter out any counter attributes in the item map if they're present when we add an update expression for the atomic counter. From a backwards compatibility perspective, this change should have little effect on existing user code. 

## Modifications
AtomicCounterExpression filters any attributes from the item that are marked as atomic counters. If it finds any, it also logs a debug level warning to provide context to users.  

## Testing
Unit and functional testing

## Types of changes

## Checklist

## License

### DIFF
--- a/services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/extensions/AtomicCounterExtension.java
+++ b/services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/extensions/AtomicCounterExtension.java
@@ -19,8 +19,10 @@ import static software.amazon.awssdk.enhanced.dynamodb.internal.EnhancedClientUt
 import static software.amazon.awssdk.enhanced.dynamodb.internal.EnhancedClientUtils.valueRef;
 import static software.amazon.awssdk.enhanced.dynamodb.internal.update.UpdateExpressionUtils.ifNotExists;
 
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
 import software.amazon.awssdk.annotations.SdkPublicApi;
@@ -37,10 +39,11 @@ import software.amazon.awssdk.enhanced.dynamodb.update.SetAction;
 import software.amazon.awssdk.enhanced.dynamodb.update.UpdateExpression;
 import software.amazon.awssdk.services.dynamodb.model.AttributeValue;
 import software.amazon.awssdk.utils.CollectionUtils;
+import software.amazon.awssdk.utils.Logger;
 
 /**
- * This extension enables atomic counter attributes to be written to the database.
- * The extension is loaded by default when you instantiate a
+ * This extension enables atomic counter attributes to be changed in DynamoDb by creating instructions for modifying
+ * an existing value or setting a start value. The extension is loaded by default when you instantiate a
  * {@link DynamoDbEnhancedClient} and only needs to be added to the client if you
  * are adding custom extensions to the client.
  * <p>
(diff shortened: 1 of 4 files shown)


services-custom/dynamodb-enhanced/src/test/java/software/amazon/awssdk/enhanced/dynamodb/functionaltests/AtomicCounterTest.java
```diff
@@ -112,15 +112,31 @@ public void createViaPut_incrementsCorrectly() {
...
+    public void updateItem_retrievedFromDb_shouldNotThrowException() {
+        AtomicCounterRecord record = new AtomicCounterRecord();
+        record.setId(RECORD_ID);
+        record.setAttribute1(STRING_VALUE);
+        mappedTable.updateItem(record);
+
+        AtomicCounterRecord retrievedRecord = mappedTable.getItem(record);
+        retrievedRecord.setAttribute1("ChangingThisAttribute");
+
+        assertThat(mappedTable.updateItem(retrievedRecord)).isNotNull();
```
zoewangg, line 139: Can we add an assertion for the counter value?

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/extensions/AtomicCounterExtension.java
```diff
@@ -37,10 +39,11 @@
 import software.amazon.awssdk.enhanced.dynamodb.update.UpdateExpression;
 import software.amazon.awssdk.services.dynamodb.model.AttributeValue;
 import software.amazon.awssdk.utils.CollectionUtils;
+import software.amazon.awssdk.utils.Logger;
 
 /**
- * This extension enables atomic counter attributes to be written to the database.
- * The extension is loaded by default when you instantiate a
+ * This extension enables atomic counter attributes to be changed in DynamoDb by creating instructions for modifying
```
zoewangg, line 45: Can we update the Javadoc to mention that users updating the fields marked with atomic counter manually has no effect?

services-custom/dynamodb-enhanced/src/test/java/software/amazon/awssdk/enhanced/dynamodb/functionaltests/AtomicCounterTest.java
cenedhryn: yes

zoewangg (approved):

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/extensions/AtomicCounterExtension.java
cenedhryn: yes

PR #4315, opened 2023-08-16
Add NoAuthAuthScheme to have visibility into the scheme id

sugmanue:
## Motivation and Context

Added a new class to contain the scheme id for no auth auth scheme. This is only for visibility of users wanting to refer this specific auth scheme.

## Types of changes

## Checklist

## License

### DIFF
--- a/core/http-auth/src/main/java/software/amazon/awssdk/http/auth/NoAuthAuthScheme.java
+++ b/core/http-auth/src/main/java/software/amazon/awssdk/http/auth/NoAuthAuthScheme.java
@@ -1,0 +1,32 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.http.auth;
+
+import software.amazon.awssdk.annotations.SdkPublicApi;
+
+/**
+ * No-auth auth scheme ID.
+ */
+@SdkPublicApi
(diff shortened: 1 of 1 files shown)


core/http-auth/src/main/java/software/amazon/awssdk/http/auth/NoAuthAuthScheme.java
```diff
@@ -0,0 +1,32 @@
...
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.http.auth;
+
+import software.amazon.awssdk.annotations.SdkPublicApi;
+
+/**
+ * No-auth auth scheme ID.
```
gosar, line 21:
```suggestion
 * An auth scheme trait that represent no authentication.
```

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/NoAuthAuthScheme.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+@SdkPublicApi
+public final class NoAuthAuthScheme {
```
gosar, line 24: I think we make this implement `AuthScheme` like the other schemes. The `signer` and `identityProvider` methods can throw IllegalStateException because they are not expected to be called, or maybe return null.

Since `sdk-core` does not depend on `http-auth` and only `http-auth-spi`, maybe this class could live in `http-auth-spi`.

gosar: For the `T extends Identity`, we could do `public final class NoAuthAuthScheme implements AuthScheme<Identity>`

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/NoAuthAuthScheme.java
sugmanue: This is not a *trait* but rather a function of a synthetic trait and that's why I avoided that wording, even more so if we change it to implement an `AuthScheme`.

gosar: Sorry, didn't mean to say `trait` in there. I suggest,`An auth scheme that represent no authentication.`

sugmanue:
Followed up on #4319

PR #4319, opened 2023-08-17
Promote no-auth to a proper AuthScheme

sugmanue:
## Motivation and Context

Created a new `AuthScheme` class `NoAuthAuthScheme` and substituted all the code paths that were relying on the hardcoded auth-scheme id. This simplifies several paths. This new auth-scheme acts as a no-op.

## Modifications

* `BaseClientBuilder` Added the new no-auth auth scheme to the default auth schemes.
* `AuthSchemeInterceptor` Updated the code to make use of it while skipping the signing needs if not needed
* 

## Types of changes

## Checklist

## License


core/sdk-core/src/main/java/software/amazon/awssdk/core/SelectedAuthScheme.java
```diff
@@ -29,19 +29,32 @@
  */
 @SdkProtectedApi
 public final class SelectedAuthScheme<T extends Identity> {
```
L-Applin, line 31: Should this be 2 classes instead, one that support signing and one that does not? It would avoid the big `if` in the constructor

core/sdk-core/src/main/java/software/amazon/awssdk/core/SelectedAuthScheme.java
```diff
@@ -57,6 +70,6 @@ public AuthSchemeOption authSchemeOption() {
     }
 
     public boolean supportsSigning() {
-        return !authSchemeOption.schemeId().equals(SMITHY_NO_AUTH);
+        return supportsSigning;
```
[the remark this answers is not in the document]
sugmanue: Yes, the idea is that the auth-scheme needs to pass down this information to the signing stages after the identity is resolved, given that this value is copied over from the auth scheme to the selected auth scheme and not computed I think there is little risk that both might end up in disagreement.

core/http-auth-spi/src/main/java/software/amazon/awssdk/http/auth/spi/NoAuthAuthScheme.java
```diff
@@ -0,0 +1,41 @@
...
+    String SCHEME_ID = "smithy.api#noAuth";
+
+    static NoAuthAuthScheme create() {
+        return DefaultNoAuthAuthScheme.create();
+    }
+
+    /**
+     * An anonymous identity used by the no-auth auth scheme.
+     */
+    interface AnonymousIdentity extends Identity {
```
gosar, line 39: I think we could define this directly in identity-spi, but we can leave that for surface area review if needed.

core/http-auth-spi/src/main/java/software/amazon/awssdk/http/auth/spi/internal/DefaultNoAuthAuthScheme.java
```diff
@@ -0,0 +1,134 @@
...
+            }
+        };
+    }
+
+    private static AnonymousIdentity anonymousIdentity() {
+        return new AnonymousIdentity() {
+            @Override
+            public Optional<Instant> expirationTime() {
+                return Optional.empty();
+            }
```
gosar, line 131: shouldn't need this, as that is the default impl in the parent interface.

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/auth/scheme/query-auth-scheme-interceptor.java
```diff
@@ -70,13 +71,6 @@ private SelectedAuthScheme<? extends Identity> selectAuthScheme(List<AuthSchemeO
                 .getAttribute(SdkInternalExecutionAttribute.IDENTITY_PROVIDER_CONFIGURATION);
         List<Supplier<String>> discardedReasons = new ArrayList<>();
         for (AuthSchemeOption authOption : authOptions) {
-            if (authOption.schemeId().equals("smithy.api#noAuth")) {
-                if (!discardedReasons.isEmpty()) {
-                    LOG.debug(() -> String.format("%s auth will be used, discarded: '%s'", authOption.schemeId(),
-                            discardedReasons.stream().map(Supplier::get).collect(Collectors.joining(", "))));
-                }
-                return new SelectedAuthScheme<>(null, null, authOption);
-            }
```
gosar, line 79: with this getting removed, we need to also configure the IdentityProviderConfiguration to include the IdentityProvider for the AnonymousIdentity in the SdkClientOption.IDENTITY_PROVIDER_CONFIGURATION that is configured for the client. Otherwise trySelectAuthScheme will return null and we will throw the "Failed to determine..." exception below.

gosar: I think this might need the IdentityProvider in NoAuthScheme to not be private and moved into identity-spi too.

gosar: see https://github.com/aws/aws-sdk-java-v2/blob/198a9d312be30d6fd0ced0c8afe9f2235ff937eb/core/sdk-core/src/main/java/software/amazon/awssdk/core/client/builder/SdkDefaultClientBuilder.java#L280

gosar: I've been using the `acm` integration test as validation, which would probably fail with this PR. I think, we actually need a test in codegen-generated-classes-test that ensures the right auth scheme gets used. I don't think this PR needs to necessarily block on that though.

sugmanue: The current implementation of the `identityProvider` returns an static one that does not take into account the configured ones, this works fine and it works pretty much the same way it was doing before.

```
    public IdentityProvider<AnonymousIdentity> identityProvider(IdentityProviderConfiguration providers) {
        return DEFAULT_IDENTITY_PROVIDER;
    }
```
FWIW, I just ran the `acm` integration test and it succeeded. Now, I understand that for completeness sake we might want to change this and select the anonymous identity if configured but I think that can be done in a follow up pull request, thoughts?

gosar: Ah nice, I think this approach is good.

gosar (approved):

codegen/src/main/java/software/amazon/awssdk/codegen/poet/auth/scheme/AuthSchemeCodegenMetadata.java
```diff
@@ -87,35 +88,30 @@ public final class AuthSchemeCodegenMetadata {
...
         .build();
 
     private final String schemeId;
     private final List<SignerPropertyValueProvider> properties;
+    private final Class<?> authSchemeClass;
 
     private AuthSchemeCodegenMetadata(Builder builder) {
         this.schemeId = Validate.paramNotNull(builder.schemeId, "schemeId");
         this.properties = Collections.unmodifiableList(Validate.paramNotNull(builder.properties, "properties"));
+        this.authSchemeClass = Validate.paramNotNull(builder.authSchemeClass, "authSchemeClass");
```
gosar, line 106: nit: given this, it seems `.filter(Objects::nonNull)` here https://github.com/aws/aws-sdk-java-v2/blob/2a9dcf28b3158d87b70f03beff06ba95cf2f818a/codegen/src/main/java/software/amazon/awssdk/codegen/poet/auth/scheme/AuthSchemeSpecUtils.java#L195 can be removed

core/http-auth-spi/src/main/java/software/amazon/awssdk/http/auth/spi/NoAuthAuthScheme.java
L-Applin: can we add a TODO comment for this?

core/http-auth-spi/src/main/java/software/amazon/awssdk/http/auth/spi/internal/DefaultNoAuthAuthScheme.java
```diff
@@ -0,0 +1,129 @@
...
+    public IdentityProvider<AnonymousIdentity> identityProvider(IdentityProviderConfiguration providers) {
+        return DEFAULT_IDENTITY_PROVIDER;
+    }
+
+    @Override
+    public HttpSigner<AnonymousIdentity> signer() {
+        return DEFAULT_SIGNER;
+    }
+
+    private static IdentityProvider<AnonymousIdentity> noAuthIdentityProvider() {
```
L-Applin, line 73: **NIT**: it is a bit cumbersome to define a static method that returns an anonymous class... is there a particular reason for that? Why not just use straight up a private static class:

```
private static class NoAuthIdentityProvider implements IdentityProvider<AnonymousIdentity> {
// ...
}
```
and have 
```
private static final IdentityProvider<AnonymousIdentity> DEFAULT_IDENTITY_PROVIDER = new NoAuthIdentityProviderI()
```

core/http-auth-spi/src/main/java/software/amazon/awssdk/http/auth/spi/internal/DefaultNoAuthAuthScheme.java
```diff
@@ -0,0 +1,129 @@
...
+            }
+
+            @Override
+            public CompletableFuture<AnonymousIdentity> resolveIdentity(ResolveIdentityRequest request) {
+                return CompletableFuture.completedFuture(ANONYMOUS_IDENTITY);
+            }
+        };
+    }
+
+    private static HttpSigner<AnonymousIdentity> noAuthSigner() {
```
L-Applin, line 88: **NIT**: see other comment about static methods returning anonymous class

core/http-auth-spi/src/main/java/software/amazon/awssdk/http/auth/spi/internal/DefaultNoAuthAuthScheme.java
```diff
@@ -0,0 +1,129 @@
...
+                    public Optional<Publisher<ByteBuffer>> payload() {
+                        return request.payload();
+                    }
+                };
+                return CompletableFuture.completedFuture(result);
+            }
+        };
+    }
+
+    private static AnonymousIdentity anonymousIdentity() {
```
L-Applin, line 125: **NIT**: same here, we could directly instantiate it for the static field

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/AsyncSigningStage.java
```diff
@@ -75,11 +75,6 @@ private <T extends Identity> CompletableFuture<SdkHttpFullRequest> sraSignReques
                                                                                       RequestExecutionContext context,
                                                                                       SelectedAuthScheme<T> selectedAuthScheme) {
         updateHttpRequestInInterceptorContext(request, context.executionContext());
-
-        if (!selectedAuthScheme.supportsSigning()) {
```
L-Applin, line 79: nice that we were able to remove this! 👍

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/SigningStage.java
```diff
@@ -75,11 +78,6 @@ private <T extends Identity> SdkHttpFullRequest sraSignRequest(SdkHttpFullReques
                                                                    RequestExecutionContext context,
                                                                    SelectedAuthScheme<T> selectedAuthScheme) {
         updateHttpRequestInInterceptorContext(request, context.executionContext());
-
-        if (!selectedAuthScheme.supportsSigning()) {
```
L-Applin, line 79: nice!

L-Applin (approved):

PR #4331, opened 2023-08-21
Optimize file uploads by reading from file in parallel

zoewangg:
## Motivation and Context
Currently, when multipart configuration is enabled, Java based S3 multipart client will read from a file sequentially. This PR optimizes file uploads performance by reading from the file in different offsets in parallel.

## Modifications
- Adding `FileRequestBodyConfiguration` to allow users to configure starting position and number of bytes to read
- Updating multipart upload helper class to use SimplePublisher to create `FileAsyncRequestBody`s from different positions.

## Testing
Added unit tests and integ tests.

## Types of changes

## Checklist

## License


core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/FileAsyncRequestBody.java
```diff
@@ -193,14 +232,17 @@ private static final class FileSubscription implements Subscription {
         private FileSubscription(Path path,
                                  AsynchronousFileChannel inputChannel,
                                  Subscriber<? super ByteBuffer> subscriber,
-                                 int chunkSize) throws IOException {
+                                 int chunkSize,
+                                 long position,
+                                 long numBytesToRead) throws IOException {
```
zoewangg, line 237: This ctor is becoming a bit crazy now, I'll fix this

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/multipart/UploadWithKnownContentLengthHelper.java
```diff
@@ -120,23 +128,68 @@ private void doUploadInParts(Pair<PutObjectRequest, AsyncRequestBody> request,
...
+                    long numBytesToRead;
+                    long position = optimalPartSize * i;
+                    if (i == partCount - 1) {
+                        numBytesToRead = mpuRequestContext.contentLength - position;
+                    } else {
+                        numBytesToRead = optimalPartSize;
+                    }
+                    simplePublisher.send(AsyncRequestBody.fromFile(
+                        config -> config.path(path)
+                                        .chunkSizeInBytes(1024 * 1024) // TODO: perf test this
```
zoewangg, line 163: This number is pending perf test

core/sdk-core/src/main/java/software/amazon/awssdk/core/FileRequestBodyConfiguration.java
```diff
@@ -0,0 +1,209 @@
...
+         * Sets the size of chunks read from the file. Increasing this will cause more data to be buffered into memory but
+         * may yield better latencies. Decreasing this will reduce memory usage but may cause reduced latency. Setting this value
+         * is very dependent on upload speed and requires some performance testing to tune.
+         *
+         * <p>The default chunk size is 16 KiB</p>
+         *
+         * @param chunkSize New chunk size in bytes.
+         * @return This builder for method chaining.
+         */
+        Builder chunkSizeInBytes(Integer chunkSize);
```
zoewangg, line 139: I kept Integer here because we only support reading up to `Integer.MAX_VALUE` number of bytes per chunk for now (ByteBuffer only takes int) and I don't think we should support long for this use case. LMK if you think otherwise.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/multipart/UploadWithKnownContentLengthHelper.java
```diff
@@ -68,17 +71,20 @@ public UploadWithKnownContentLengthHelper(S3AsyncClient s3AsyncClient,
     }
 
     public CompletableFuture<PutObjectResponse> uploadObject(PutObjectRequest putObjectRequest,
-                                                             AsyncRequestBody asyncRequestBody,
+                                                             Either<AsyncRequestBody, Path> requestBodyOrPath,
```
millems, line 74: Can we `@Override` `split` in `AsyncFileRequestBody` so that we don't need to do the splitting here?

zoewangg: Good idea!

core/sdk-core/src/main/java/software/amazon/awssdk/core/FileRequestBodyConfiguration.java
millems: 2 GB per chunk is probably a reasonable maximum for now, if it's not easy to support a higher number.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/FileAsyncRequestBodySplitHelper.java
```diff
@@ -0,0 +1,162 @@
...
+        } finally {
+            isSendingRequestBody.set(false);
+        }
+    }
+
+    private void doSendAsyncRequestBody(SimplePublisher<AsyncRequestBody> simplePublisher) {
+        while (true) {
+            if (!shouldSendMore()) {
+                break;
+            }
```
millems, line 103: `while (shouldSendMore())`?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/FileAsyncRequestBodySplitHelper.java
```diff
@@ -0,0 +1,162 @@
...
+        }
+    }
+
+    private void checkCompletion(SimplePublisher<AsyncRequestBody> simplePublisher, AsyncRequestBody currentAsyncRequestBody) {
+        long remaining = remainingBytes.addAndGet(-currentAsyncRequestBody.contentLength().get());
+
+        if (remaining == 0) {
+            isDone = true;
+            simplePublisher.complete();
+        }
```
millems, line 118: Can we add a branch for < 0 and call error on the publisher? If there's a bug, I'd rather it manifest as an exception than a hung future.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/FileAsyncRequestBodySplitHelper.java
```diff
@@ -0,0 +1,183 @@
...
+        return SdkPublisher.adapt(simplePublisher);
+    }
+
+    private void sendAsyncRequestBody(SimplePublisher<AsyncRequestBody> simplePublisher) {
+        if (!isSendingRequestBody.compareAndSet(false, true)) {
+            return;
+        }
+
+        try {
+            doSendAsyncRequestBody(simplePublisher);
```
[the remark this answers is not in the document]
zoewangg: Good call! Updated

millems (approved):

PR #4337, opened 2023-08-23
Remove dependency on http-auth-aws from codecatalyst.

millems:
Previously (#4311), AwsEndpointProviderUtils and AuthSchemeUtils, which are copied into every client, depend on classes from http-auth-aws. As a result, every client needed a dependency on http-auth-aws. With this change, we move the methods from those classes that depend on http-auth-aws to be code generated within only the clients that need them.

### DIFF
--- a/codegen/src/test/resources/software/amazon/awssdk/codegen/poet/rules/endpoint-resolve-interceptor.java
+++ b/codegen/src/test/resources/software/amazon/awssdk/codegen/poet/rules/endpoint-resolve-interceptor.java
@@ -5,9 +5,12 @@ import java.util.Optional;
 import java.util.concurrent.CompletionException;
 import software.amazon.awssdk.annotations.Generated;
 import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.auth.signer.AwsSignerExecutionAttribute;
 import software.amazon.awssdk.awscore.AwsExecutionAttribute;
 import software.amazon.awssdk.awscore.endpoints.AwsEndpointAttribute;
 import software.amazon.awssdk.awscore.endpoints.authscheme.EndpointAuthScheme;
+import software.amazon.awssdk.awscore.endpoints.authscheme.SigV4AuthScheme;
+import software.amazon.awssdk.awscore.endpoints.authscheme.SigV4aAuthScheme;
 import software.amazon.awssdk.core.SdkRequest;
 import software.amazon.awssdk.core.SelectedAuthScheme;
 import software.amazon.awssdk.core.exception.SdkClientException;
@@ -18,6 +21,14 @@ import software.amazon.awssdk.core.interceptor.SdkExecutionAttribute;
 import software.amazon.awssdk.core.interceptor.SdkInternalExecutionAttribute;
 import software.amazon.awssdk.endpoints.Endpoint;
 import software.amazon.awssdk.http.SdkHttpRequest;
+import software.amazon.awssdk.http.auth.aws.AwsV4AuthScheme;
+import software.amazon.awssdk.http.auth.aws.AwsV4HttpSigner;
+import software.amazon.awssdk.http.auth.aws.AwsV4aAuthScheme;
+import software.amazon.awssdk.http.auth.aws.AwsV4aHttpSigner;
+import software.amazon.awssdk.http.auth.spi.AuthSchemeOption;
+import software.amazon.awssdk.identity.spi.Identity;
+import software.amazon.awssdk.regions.Region;
+import software.amazon.awssdk.regions.RegionScope;
 import software.amazon.awssdk.services.query.endpoints.QueryClientContextParams;
 import software.amazon.awssdk.services.query.endpoints.QueryEndpointParams;
 import software.amazon.awssdk.services.query.endpoints.QueryEndpointProvider;
@@ -44,15 +55,15 @@ public final class QueryResolveEndpointInterceptor implements ExecutionIntercept
                     endpoint = AwsEndpointProviderUtils.addHostPrefix(endpoint, hostPrefix.get());
                 }
             }
-            List<EndpointAuthScheme> authSchemes = endpoint.attribute(AwsEndpointAttribute.AUTH_SCHEMES);
+            List<EndpointAuthScheme> endpointAuthSchemes = endpoint.attribute(AwsEndpointAttribute.AUTH_SCHEMES);
             SelectedAuthScheme<?> selectedAuthScheme = executionAttributes
                 .getAttribute(SdkInternalExecutionAttribute.SELECTED_AUTH_SCHEME);
-            if (authSchemes != null && selectedAuthScheme != null) {
-                selectedAuthScheme = AwsEndpointProviderUtils.mergeIntoResolvedAuthScheme(authSchemes, selectedAuthScheme);
+            if (endpointAuthSchemes != null && selectedAuthScheme != null) {
+                selectedAuthScheme = copyEndpointSignerPropertiesToAuth(endpointAuthSchemes, selectedAuthScheme);
                 executionAttributes.putAttribute(SdkInternalExecutionAttribute.SELECTED_AUTH_SCHEME, selectedAuthScheme);
             }
             if (selectedAuthScheme != null) {
-                AuthSchemeUtils.setSigningParams(executionAttributes, selectedAuthScheme.authSchemeOption());
+                copySignerPropertiesToAttributes(selectedAuthScheme.authSchemeOption(), executionAttributes);
             }
             executionAttributes.putAttribute(SdkInternalExecutionAttribute.RESOLVED_ENDPOINT, endpoint);
             return result;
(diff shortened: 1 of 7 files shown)


codegen/src/test/resources/software/amazon/awssdk/codegen/poet/rules/endpoint-resolve-interceptor.java
```diff
@@ -44,15 +55,15 @@ public SdkRequest modifyRequest(Context.ModifyRequest context, ExecutionAttribut
...
                 .getAttribute(SdkInternalExecutionAttribute.SELECTED_AUTH_SCHEME);
-            if (authSchemes != null && selectedAuthScheme != null) {
-                selectedAuthScheme = AwsEndpointProviderUtils.mergeIntoResolvedAuthScheme(authSchemes, selectedAuthScheme);
+            if (endpointAuthSchemes != null && selectedAuthScheme != null) {
+                selectedAuthScheme = copyEndpointSignerPropertiesToAuth(endpointAuthSchemes, selectedAuthScheme);
                 executionAttributes.putAttribute(SdkInternalExecutionAttribute.SELECTED_AUTH_SCHEME, selectedAuthScheme);
             }
             if (selectedAuthScheme != null) {
-                AuthSchemeUtils.setSigningParams(executionAttributes, selectedAuthScheme.authSchemeOption());
+                copySignerPropertiesToAttributes(selectedAuthScheme.authSchemeOption(), executionAttributes);
```
gosar, line 66: I know the context of this PR is to fix codecatalyst, but the earlier PR where this logic was added to ResolveEndpointInterceptor is already merged, so commenting here.

It seems odd for this to be responsibility of Endpoint related interceptor. It only deals with the 3 properties that endpoint rules can set, but I think we may do this generically for all signer properties, if we leave any code paths where old signer will be used (like in Presigners).

What would be broken if we didn't have this right now? I'm guessing maybe S3 related tests? Could we disable those similar to other tests that were disabled? If it is just presigner, I think we can do the generic conversion for all properties in there?

test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/endpointproviders/EndpointInterceptorTests.java
```diff
@@ -329,7 +332,135 @@ public void async_endpointProviderReturnsSignerProperties_overridesV4AAuthScheme
...
+                                                                           .putAttribute(AwsEndpointAttribute.AUTH_SCHEMES, emptyList())
+                                                                           .build()))
+            .build();
+
+        assertThatThrownBy(() -> client.operationWithHostPrefix(r -> {}).join())
+            .hasMessageContaining("stop");
+
+        assertThat(interceptor.executionAttributes.getAttribute(AwsSignerExecutionAttribute.SIGNING_REGION)).isEqualTo(Region.of("X"));
+        assertThat(interceptor.executionAttributes.getAttribute(AwsSignerExecutionAttribute.SERVICE_SIGNING_NAME)).isEqualTo("Y");
+        assertThat(interceptor.executionAttributes.getAttribute(AwsSignerExecutionAttribute.SIGNER_DOUBLE_URL_ENCODE)).isEqualTo(true);
```
gosar, line 360: Related to my comment above about `copySignerPropertiesToAttributes`, this is surprising behavior for Endpoint Interceptor. In this test case, Endpoint rules isn't playing a role in auth scheme/properties, but is still converting _some_ (just these 3) signer properties to execution attributes. If in this test case, a 4th SignerProperty was set in AuthSchemeOption the corresponding execution attribute would not be set and that is odd behavior.

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/rules/endpoint-resolve-interceptor.java
millems: A lot breaks without this - anything that relies on the old signers. That's a lot to disable when it is already fixed here. I'd also favor leaving it in for now.

In terms of whose responsibility it is, I put it here because it is this interceptor that (currently) finalizes the selected auth scheme's properties, but is that the right place? I don't think so, because what if something else needs to change the selected auth scheme again? Why is it the endpoint interceptor that gets to say that people aren't allowed to change it anymore?

The next idea would be to give ownership to the end of the request pipeline. Why don't we do it right before we call the signer? That way we know for sure the selected auth scheme is properly reflected in the execution attributes when they get to the legacy signer.

There's a problem with that, though. There may be legacy interceptors that rely on the values being set sooner. We want the values to be available as soon as possible to minimize the chance of backwards-incompatibility with older interceptors, so that pushes us to do it every single time we set or change the selected auth scheme. That's a very leaky abstraction.

So what do I think we should do? I think we should create what I've described in the past as "derived execution attributes". That is, ones that are determined **at the time they are read** based on other execution attributes. That way, we can define the derived execution attribute as being based on the current value of the selected auth scheme execution attribute.

For example, `executionAttributes.putDerivedAttribute(SIGNING_REGION, e -> e.getAttribute(SELECTED_AUTH_SCHEME).signingRegion())`. The advantage is that we just have to do this once - when we initially set the `SELECTED_AUTH_SCHEME`. Then we know that even if `SELECTED_AUTH_SCHEME` attribute changes, that we don't have to go and update `SIGNING_REGION` attribute.

It's not a major change. If you think this is an okay idea I can do a follow-up PR to make that change?

test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/endpointproviders/EndpointInterceptorTests.java
millems: See my comment above for the ownership concerns. I agree that it's weird here and we should move it. To where? Let's talk on my reply above.

As for the new 4th signer property concern, I don't see that being a problem. If we added new signer properties, we probably won't bother adapting it to the old execution attributes - I can't think of a reason to do so. If that's right it  means the mapping code probably will never change to add support for new attributes.

cenedhryn: Yes, from a maintainability perspective it's confusing and error prone. I'd expect the attributes that are related to AuthSchemeOption to be set after auth scheme resolution, but there are maybe some blockers I don't know about.

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/rules/endpoint-resolve-interceptor.java
```diff
@@ -44,15 +55,15 @@ public SdkRequest modifyRequest(Context.ModifyRequest context, ExecutionAttribut
...
                 }
             }
-            List<EndpointAuthScheme> authSchemes = endpoint.attribute(AwsEndpointAttribute.AUTH_SCHEMES);
+            List<EndpointAuthScheme> endpointAuthSchemes = endpoint.attribute(AwsEndpointAttribute.AUTH_SCHEMES);
             SelectedAuthScheme<?> selectedAuthScheme = executionAttributes
                 .getAttribute(SdkInternalExecutionAttribute.SELECTED_AUTH_SCHEME);
-            if (authSchemes != null && selectedAuthScheme != null) {
-                selectedAuthScheme = AwsEndpointProviderUtils.mergeIntoResolvedAuthScheme(authSchemes, selectedAuthScheme);
+            if (endpointAuthSchemes != null && selectedAuthScheme != null) {
+                selectedAuthScheme = copyEndpointSignerPropertiesToAuth(endpointAuthSchemes, selectedAuthScheme);
```
cenedhryn, line 62: Seems weird that a method called `copyEndpointSignerPropertiesToAuth` returns an auth scheme. I gather that the reason for it is to maintain some immutability by creating a new scheme. Can we rename the method, maybe `authSchemeWithEndpointSignerProperties`?

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/rules/endpoint-resolve-interceptor.java
cenedhryn: That's a cool idea (and I don't recall the past well). It'd be super nice to have that in the code. It'd be useful to be able to distinguish between read-only and read-write derived attributes

gosar (approved):
good to ship if CI passes after the merge from upstream.

cenedhryn (approved):

test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/endpointproviders/EndpointInterceptorTests.java
gosar: Clarified offline that the 4th signer property case I was thinking was not a new Signer property we'll add later, but other existing signer properties that map to existing execution attributes, like payload signing.

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/rules/endpoint-resolve-interceptor.java
gosar: >What would be broken if we didn't have this right now?

Discussed offline, generically, this would be any cases where Signer is still used. With this project, we expect all newly generated clients that have SRA auth scheme resolution, to also use the new HttpSigner and not Signer. So it is only for when customer overrides to provide their own old interface Signer (isSignerOverridden=true cases). If customer provides their own Signer impl, we don't necessarily expect them to use our signer execution attributes because they are SdkProtectedApi - but it still have practical risk of breaking someone. But more than that, there's the case (I hadn't thought about this before discussion) where customers is explicitly providing Signer and they are using our existing SdkPublicApi (to be deprecated) Signers, say they want to use sigv4a. To support these use case, we do need ability to map new `SignerProperty`s to signer ExecutionAttributes used by old Signer - beyond the 3 endpoint related here. So @millems is going to follow up (in separate PR) on his idea above to support these.

PR #4339, opened 2023-08-24
Add identity fetch metrics for SRA path

gosar:
## Motivation and Context

In pre SRA code path, AuthorizationStrategy would add CREDENTIALS_FETCH_DURATION/TOKEN_FETCH_DURATION metric. But with SRA, AuthorizationStrategy is not used since https://github.com/aws/aws-sdk-java-v2/pull/4292. This breaks some tests that expect the metric, and this PR is meant to fix that, by adding these metrics in the SRA code path.

The change to generated code (in {Service}AuthSchemeInterceptor) will change
```
        CompletableFuture<? extends T> identity = identityProvider.resolveIdentity(identityRequestBuilder.build());
```

to 

```
        MetricCollector metricCollector = executionAttributes.getAttribute(SdkExecutionAttribute.API_CALL_METRIC_COLLECTOR);

        CompletableFuture<? extends T> identity;
        SdkMetric<Duration> metric = getIdentityMetric(identityProvider);
        if (metric == null) {
            identity = identityProvider.resolveIdentity(identityRequestBuilder.build());
        } else {
            identity = MetricUtils.reportDuration(
                () -> identityProvider.resolveIdentity(identityRequestBuilder.build()), metricCollector, metric);
        }

<snip>

    private SdkMetric<Duration> getIdentityMetric(IdentityProvider<?> identityProvider) {
        Class<?> identityType = identityProvider.identityType();
        if (identityType == AwsCredentialsIdentity.class) {
            return CoreMetric.CREDENTIALS_FETCH_DURATION;
        }
        if (identityType == TokenIdentity.class) {
            return CoreMetric.TOKEN_FETCH_DURATION;
        }
        return null;
    }
```

## Modifications

* Adding some utility for metrics reporting with CompletableFuture. and refactoring some places to use it.
* codegen changes and re-enable test
* Re-enable commented tests from https://github.com/aws/aws-sdk-java-v2/pull/4342

## Testing

Re-enabled metrics tests


core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/util/MetricUtils.java
```diff
@@ -56,6 +58,26 @@ public static <T> Pair<T, Duration> measureDuration(Supplier<T> c) {
...
+     */
+    public static <T> CompletableFuture<T> reportDuration(Supplier<CompletableFuture<T>> c,
+                                                          MetricCollector metricCollector,
+                                                          SdkMetric<Duration> metric) {
+        long start = System.nanoTime();
+        CompletableFuture<T> result = c.get();
+        result.whenComplete((r, t) -> {
+            Duration d = Duration.ofNanos(System.nanoTime() - start);
+            metricCollector.reportMetric(metric, d);
+        });
```
gosar, line 77: note, this is publishing the metric whether future completed successfully or exceptionally. Could overload this method if we want behavior where we only want to report on success - is that what we want for the identity / signing metrics? I was thinking it is ok to publish a duration metric even on failure, so did this method like that. LMK thoughts.

dagnir: For consistency, let's report on success only for now. For other auth related metrics, we're only reporting when it's successful 

- https://github.com/aws/aws-sdk-java-v2/blob/05492d2145acbd733c2a8d1af182133c2f39f1ae/core/aws-core/src/main/java/software/amazon/awssdk/awscore/internal/authcontext/TokenAuthorizationStrategy.java#L81
 - https://github.com/aws/aws-sdk-java-v2/blob/05492d2145acbd733c2a8d1af182133c2f39f1ae/core/aws-core/src/main/java/software/amazon/awssdk/awscore/internal/authcontext/AwsCredentialsAuthorizationStrategy.java#L101

codegen/src/main/java/software/amazon/awssdk/codegen/poet/auth/scheme/AuthSchemeInterceptorSpec.java
```diff
@@ -255,13 +262,40 @@ private MethodSpec generateTrySelectAuthScheme() {
                              ResolveIdentityRequest.class,
                              ResolveIdentityRequest.class);
         builder.addStatement("authOption.forEachIdentityProperty(identityRequestBuilder::putProperty)");
-        builder.addStatement("$T identity = identityProvider.resolveIdentity(identityRequestBuilder.build())",
-                             namedIdentityFuture());
+
+        builder.addStatement("$T<? extends T> identity", CompletableFuture.class);
```
dagnir, line 266: nit: May want to use a combination of `ParameterizedTypeName` and `TypeVariableName` to be more idiomatic

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/util/MetricUtils.java
gosar: yeah, I knew this is deviating from current behavior, which is why I called it out :). My thought: for SIGNING_DURATION, we are changing this in SRA to emit the SIGNING_DURATION metric even if authType=none. See https://github.com/aws/aws-sdk-java-v2/pull/4319. I think we don't need to be tied to old behavior here too, so if we think the preferred behavior would be emit it even on failure, we can do that. I would prefer a duration metric to be emitted, so I know how much time is spent, even if it resulted in a failure, since that's how I imagine duration metrics are used - to analyze where time is spent. So my "LMK thoughts" was to primarily see if we prefer that behavior or actually prefer to keep the old behavior. LMK if you still prefer to keep the old behavior and I can change the PR accordingly.

dagnir: If these processes fail, it usually fails the entire API call, so in that case it's not really useful from a debugging purpose usually. I don't have a strong preference in keeping with the old behavior, and in fact we do report other non-auth/signing metrics even in failure. I'd prefer consistency going forward so if others are being changed, then let's report it unconditionally.

codegen/src/main/java/software/amazon/awssdk/codegen/poet/auth/scheme/AuthSchemeInterceptorSpec.java
gosar: ah, good catch. `namedIdentityFuture()` was already there in previous code. I just moved it incorrectly to line 270 below in this change. fixed.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/util/MetricUtils.java
gosar: cool, so I'm keeping this like I have. 

btw,
>it usually fails the entire API call, so in that case it's not really useful from a debugging purpose usually

Sometimes, in failed API call case, knowing where time was spent/not spent, may actually help with debugging and point you somewhere. Maybe. Point being there's not a good reason (imo) to discard the time metric if you could as easily report it.

dagnir (approved):

PR #4341, opened 2023-08-24
Set the signing clock property

sugmanue:
## Motivation and Context
The new signer interface needs to be aware of any detected time offset to use that when signing. We cannot set that property properly since we don't have that information inside the auth interceptor, so unlike the other properties, we set this when we create the request. The way it's currently implemented follows the same logic that the previous signers were using before.

## Modifications

Modified the sync/async signing stages to grab from the dependencies the current timeoffset and use it to construct a `Clock` that the signer can use to compute the singing date.

## Types of changes

## Checklist

## License


core/http-auth-spi/src/main/java/software/amazon/awssdk/http/auth/spi/HttpSignerCommon.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+package software.amazon.awssdk.http.auth.spi;
+
+import java.time.Clock;
+import software.amazon.awssdk.annotations.SdkPublicApi;
+
+/**
+ * Keeps common signing properties used by different signers.
+ */
```
millems, line 23: This makes sense. I was thinking about this a bit when I was working on some of the v4+v4a attributes, but decided we probably shouldn't do it for those attributes. It makes a lot of sense for the signing clock, though.

I think we should be really strict in the javadoc about this. It should be things that CAN NEVER be signer-specific. A clock is universal, it makes sense here. We need to write the javadoc in a way that keeps things like service signing name out of this. In my opinion, clock is probably the only thing that ever makes sense here, but who knows what the future will hold...

core/http-auth-spi/src/main/java/software/amazon/awssdk/http/auth/spi/HttpSignerCommon.java
```diff
@@ -0,0 +1,33 @@
...
+@SdkPublicApi
+public final class HttpSignerCommon {
```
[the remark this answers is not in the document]
sugmanue: I like it better but checkstyle does not:


```
Enum or class name should be singular, not plural (Property.VALUE not Properties.VALUE). [PluralEnumNames]
```

What about `SingerPropertyCommon`?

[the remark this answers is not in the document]
sugmanue: Yet another option is to be **super** specific, such as `ClockSignerProperty`.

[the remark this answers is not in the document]
sugmanue: I went this path, let me know what you think.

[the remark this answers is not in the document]
millems: `SharedSignerProperty`? We even could put it in `SignerProperty`, but that might be a bit much.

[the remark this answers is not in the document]
millems: It definitely would discourage what happened with execution attributes where people aren't careful. I'm worried a bit about the pattern of making multiple classes if we do actually end up needing multiple shared properties. We could do a quick poll in the channel? I think I'd favor a single class, but both approaches have merit.

core/http-auth-spi/src/main/java/software/amazon/awssdk/http/auth/spi/SharedSignerProperty.java
```diff
@@ -0,0 +1,33 @@
...
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.http.auth.spi;
+
+import java.time.Clock;
+import software.amazon.awssdk.annotations.SdkPublicApi;
+
+/**
+ * Container class for the clock signer property that is common to some signing schemes.
```
millems, line 22: How about:
> Signer properties that are common to every possible current and future signer. See the individual signer interfaces for properties that are not universal across signers.

core/http-auth-spi/src/main/java/software/amazon/awssdk/http/auth/spi/HttpSignerCommon.java
gosar: My initial thought was this goes in the `HttpSigner` interface itself and specific signers javadoc whether they use this property or not. What do folks think about that?

core/sdk-core/src/test/java/software/amazon/awssdk/core/internal/http/pipeline/stages/AsyncSigningStageTest.java
```diff
@@ -177,8 +234,13 @@ public void execute_selectedAuthScheme_asyncRequestBody_doesSraSignAsync() throw
         AsyncSignRequest<? extends Identity> signRequest = asyncSignRequestCaptor.getValue();
         assertThat(signRequest.identity()).isSameAs(identity);
         assertThat(signRequest.request()).isSameAs(request);
-        assertThat(signRequest.property(SIGNER_PROPERTY)).isEqualTo("value");
         assertThat(signRequest.payload().get()).isSameAs(asyncPayload);
+        assertThat(signRequest.property(SIGNER_PROPERTY)).isEqualTo("value");
+
+        // Assert that the time offset set was zero
+        assertThat(signRequest.property(SharedSignerProperty.SIGNING_CLOCK)).isNotNull();
```
gosar, line 241: I think rather not set the property if offset is 0

core/http-auth-spi/src/main/java/software/amazon/awssdk/http/auth/spi/HttpSigner.java
```diff
@@ -31,6 +32,12 @@
 @SdkPublicApi
 public interface HttpSigner<IdentityT extends Identity> {
 
+    /**
+     * A {@link Clock} to be used at the time of signing which is shared among several signers. This property defaults to the time
```
gosar, line 36: nit: The "which is shared among several signers" can read like the Clock is shared as opposed to the property being used by multiple signers. I think avoiding "shared" here is better.

Suggest something like:
```
* A {@link Clock} to be used to derive the signing time. This property defaults to the system clock.
*
*<p>Note, signing time may not be relevant to some signers.
```

core/sdk-core/src/test/java/software/amazon/awssdk/core/internal/http/pipeline/stages/AsyncSigningStageTest.java
sugmanue: That will introduce a branch that is not needed the way it's written now, so I think is better just to have a single code path. Can you please help me understand why you think is better not to set it?

core/http-auth-spi/src/main/java/software/amazon/awssdk/http/auth/spi/HttpSigner.java
sugmanue: I just pushed a commit.

core/sdk-core/src/test/java/software/amazon/awssdk/core/internal/http/pipeline/stages/AsyncSigningStageTest.java
gosar: I thought it is better since the test more fragile and can fail sometimes if for some reason Instant.now is actually beyond the `within` value?

gosar: The other test is still fragile like that. I think we can have a package-private constructor that takes in a Clock (instead of Clock.systemUTC), so that unit test can use a mock Clock and test exactly, instead of `isCloseTo`.

gosar: That would address my main concern - test fragility. We could still chose to not set or set the clock property if offset is `0`. 
Outside of that, I was thinking that if there is no need for offset, the SigningStage can let the HttpSigner decide whether it needs a clock at all or not. But I don't feel strongly about that, if the test fragility is addressed.

sugmanue: I'd rather tackle the problem when and if arises, I'm know that what you describing is possible but I don't think is likely, for that to happen it would mean that the machine running the tests will need to park the thread for more than 100ms (~ 300 Million instructions). Were the previous tests brittle like actually failing every so often?

gosar (approved):
Approving this to not block on decision on what you do about test fragility.

millems (approved):

core/sdk-core/src/test/java/software/amazon/awssdk/core/internal/http/pipeline/stages/AsyncSigningStageTest.java
```diff
@@ -67,7 +70,7 @@
 
 @RunWith(MockitoJUnitRunner.class)
 public class AsyncSigningStageTest {
```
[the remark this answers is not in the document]
gosar: good call about that use case!

PR #4345, opened 2023-08-24
Add checksums module

haydenbaker:
- Adds checksums module
- Adds Checksum enum

## Motivation and Context
- This new module is meant to be the place for the new reference architecture checksums work (that will be done some day in the future)
- This will be used in the ongoing signers work in order to map from reference-based checksums (the new enum) to the internal SDK implementations of the checksums (`SdkChecksum`), such that we can someday swap out the old checksum code in the signers easily

## Types of changes

## Checklist

## License

### DIFF
--- a/core/checksums/pom.xml
+++ b/core/checksums/pom.xml
@@ -42,8 +42,7 @@
         <dependency>
             <groupId>software.amazon.awssdk</groupId>
             <artifactId>checksums-spi</artifactId>
-            <version>2.20.131-SNAPSHOT</version>
-            <scope>compile</scope>
+            <version>${awsjavasdk.version}</version>
         </dependency>
 
         <dependency>
--- a/core/checksums/pom.xml
+++ b/core/checksums/pom.xml
@@ -1,0 +1,60 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+  ~
+  ~ Licensed under the Apache License, Version 2.0 (the "License").
+  ~ You may not use this file except in compliance with the License.
+  ~ A copy of the License is located at
+  ~
+  ~  http://aws.amazon.com/apache2.0
+  ~
+  ~ or in the "license" file accompanying this file. This file is distributed
+  ~ on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+  ~ express or implied. See the License for the specific language governing
+  ~ permissions and limitations under the License.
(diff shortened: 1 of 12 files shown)


core/checksums/src/main/java/software/amazon/awssdk/checksums/Checksum.java
```diff
@@ -0,0 +1,30 @@
...
+ * An enumeration of supported checksums.
+ */
+@SdkProtectedApi
+public enum Checksum {
+    CRC32C,
+    CRC32,
+    MD5,
+    SHA256,
+    SHA1,
+}
```
millems, line 30: This is going to sound weird, but can we make this a class with public static final fields for the checksum values?

```
public final class Checksum {
    public final class Checksum CRC32C = Checksum.of("CRC32C");
    public final class Checksum CRC32 = Checksum.of("CRC32");
    public final class Checksum MD5 = Checksum.of("MD5");
    public final class Checksum SHA256 = Checksum.of("SHA256");
    public final class Checksum SHA1 = Checksum.of("SHA1");

    private static final Map<String, Checksum> VALUES = new ConcurrentHashMap<>();

    private final Checksum of(String name) {
        return VALUES.computeIfAbsent(name, n -> new Checksum());
    }
}
```

Making it not an enum allows us to add more in the future from within an SDK plugin. For example, S3 can define `Checksum.of("SHA512")` without needing to have it pushed "upstream" to the `Checksum` type. Making that useful will also require additional methods being added to Checksum so that we can interact with it polymorphically, (e.g. to get an implementation of the actual checksum algorithm), but making it not-an-enum now will allow for that.

millems (commented):
Can we also add this to `bom/pom.xml`?

core/checksums/src/main/java/software/amazon/awssdk/checksums/Checksum.java
haydenbaker: Sounds good to me.

core/checksums/pom.xml
```diff
@@ -0,0 +1,72 @@
...
+    <dependencies>
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>annotations</artifactId>
+            <version>${awsjavasdk.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>checksums-spi</artifactId>
+            <version>2.20.131-SNAPSHOT</version>
```
millems, line 45: ${awsjavasdk.version}

millems (approved):

PR #4358, opened 2023-08-28
Configure client with NoAuthAuthScheme by default

sugmanue:
## Motivation and Context

This change adds the `NoAuthAuthScheme` to the list of available auth schemes to all services.

## Types of changes

## Checklist

## License

### DIFF
--- a/codegen/src/main/java/software/amazon/awssdk/codegen/poet/auth/scheme/AuthSchemeSpecUtils.java
+++ b/codegen/src/main/java/software/amazon/awssdk/codegen/poet/auth/scheme/AuthSchemeSpecUtils.java
@@ -36,6 +36,7 @@ import software.amazon.awssdk.codegen.model.service.AuthType;
 import software.amazon.awssdk.codegen.utils.AuthUtils;
 import software.amazon.awssdk.http.auth.aws.AwsV4aAuthScheme;
 import software.amazon.awssdk.http.auth.spi.AuthSchemeOption;
+import software.amazon.awssdk.http.auth.spi.NoAuthAuthScheme;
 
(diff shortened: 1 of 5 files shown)


cenedhryn (approved):

codegen/src/main/java/software/amazon/awssdk/codegen/poet/auth/scheme/AuthSchemeSpecUtils.java
```diff
@@ -197,6 +198,8 @@ public Set<Class<?>> allServiceConcreteAuthSchemeClasses() {
             // sigv4a is not modeled but needed for the endpoints based auth-scheme cases.
             result.add(AwsV4aAuthScheme.class);
         }
+        // Make the no-auth scheme available.
+        result.add(NoAuthAuthScheme.class);
```
gosar, line 202: not your change, but looking at where this method is used, looks like there is a potential bug in BaseClientBuilderClass.authSchemesMethod() that would have returned emptyMap before your change, ignoring the additionalAuthSchemes. I think this PR makes it irrelevant (as the collection will never be null). I was thinking we are missing a test case for service with zero auth schemes, but looks like C2J just doesn't allow it today. Either ways, with your PR, looks like 
```
        if (concreteAuthSchemeClasses.isEmpty()) {
            builder.addStatement("return $T.emptyMap()", Collections.class);
            return builder.build();
        }
```
can be removed. I have done it locally, will send PR

gosar: https://github.com/aws/aws-sdk-java-v2/pull/4364 for this

PR #4364, opened 2023-08-29
Remove emptyMap case that would miss additionalAuthSchemes

gosar:
## Motivation and Context

context: https://github.com/aws/aws-sdk-java-v2/pull/4358/files#r1308068350

## Modifications

Remove emptyMap case that would miss additionalAuthSchemes

## Testing

Was looking at add a test case for model with zero auth schemes (to showcase the case where concreteAuthSchemeClasses would have been empty before #4358), but the model doesn't seem to support it today.

The closest equivalent in the model is
```
        "signatureVersion": "none",
        "auth": [],
```
which would still not have triggered the emptyMap case, as it would be a list of ["none"]. But since I coded it up, adding that test case anyways. It may not be directly relevant to the current model format, but whenever codegen is moved to Smithy models, this test case may serve as reminder to convert that test case to Smithy, where zero auth types is supported.


gosar:
CI failing pending https://github.com/aws/aws-sdk-java-v2/pull/4363

sugmanue (approved):

PR #4366, opened 2023-08-29
Add AddingTrailingDataSubscriber to allow users to send additional data t…

zoewangg:
…o the downstream subscriber

## Motivation and Context
There are use-cases where we want to send trailing data on top of what the publisher publishes. For example, in async signing code path, we always send a trailing empty frame.

## Modifications
- Add AddingTrailingDataSubscriber to allow users to send trailling to the downstream subscriber
- Expose `SdkPublisher#addTrailingData` 

## Testing
Added TCK tests and unit tests.

## Types of changes

## Checklist

## License


utils/src/main/java/software/amazon/awssdk/utils/async/AdditionalDataSubscriber.java
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    @Override
+    public void onSubscribe(Subscription subscription) {
+
+        if (upstreamSubscription != null) {
+            log.warn(() -> "Received duplicate subscription, cancelling the duplicate.", new IllegalStateException());
+            subscription.cancel();
+            return;
+        }
```
cenedhryn, line 73: Can we test this?

zoewangg: This is covered in TCK tests :) `required_spec205_mustCallSubscriptionCancelIfItAlreadyHasAnSubscriptionAndReceivesAnotherOnSubscribeSignal`

utils/src/main/java/software/amazon/awssdk/utils/async/AddingTrailingDataSubscriber.java
```diff
@@ -0,0 +1,148 @@
...
+import software.amazon.awssdk.annotations.SdkProtectedApi;
+import software.amazon.awssdk.utils.Logger;
+import software.amazon.awssdk.utils.Validate;
+
+/**
+ * Allows to send trailing data before invoking onComplete on the downstream subscriber.
+ * If the trailingDataSupplier returns null, this class will invoke onComplete directly
+ */
+@SdkProtectedApi
+public class AddingTrailingDataSubscriber<T> extends DelegatingSubscriber<T, T> {
```
dagnir, line 32: I think this is great. Would it be much more work to to generalize this to work for N number additional data?

utils/src/main/java/software/amazon/awssdk/utils/async/AddingTrailingDataSubscriber.java
```diff
@@ -0,0 +1,148 @@
...
+        downstreamDemand.decrementAndGet();
+        subscriber.onNext(t);
+    }
+
+    @Override
+    public void onComplete() {
+        onCompleteCalledByUpstream = true;
+
+        trailingData = trailingDataSupplier.get();
+        if (trailingData == null || downstreamDemand.get() > 0) {
```
L-Applin, line 122: We `get()` the trailing data from the supplier, but send it only if the supplier returns `null`? Is this really what we want?

utils/src/main/java/software/amazon/awssdk/utils/async/AddingTrailingDataSubscriber.java
zoewangg: Good idea! Updated

utils/src/main/java/software/amazon/awssdk/utils/async/AddingTrailingDataSubscriber.java
zoewangg: This code has been updated, but to answer your question, the original code allows it to trigger onComplete immediately if trailingData is null

cenedhryn (approved):

L-Applin (approved):

dagnir (approved):

PR #4376, opened 2023-08-31
Add flexible checksum support for non-streaming cases

haydenbaker:
## Motivation and Context
- Flexible checksum support is needed for a few specific cases where a custom checksum can be added to a request
- This PR addresses the non-streaming cases for unsigned and signed payload
- A follow-up PR will be sent to address the streaming (aws-chunked) case, but there is no plan to support flexible checksums for event-streaming case or non-streaming CRT-V4a

## Modifications
- Copies `SdkChecksum` and related code from `sdk-core` to use internally without importing `sdk-core`
- Add new internal classes to handle flexible checksum cases, and remove the old digest-related stream and subscriber code
- Cleans up a couple of APIs that should be internal, not protected

## Testing
- Run new and existing tests, make sure no failures
- Run subscriber TCK test, make sure the implementation is compliant

## Types of changes

## Checklist

- [x]I have added tests to cover my changes

## License


core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/checksums/Md5Checksum.java
haydenbaker, line 1: Copied. Ignore.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/checksums/SdkChecksum.java
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    /**
+     * Return an encoded-string representation of the checksum. By default, this returns a String that is the lowercase, base16
+     * (hex) representation of the checksum.
+     *
+     */
+    default String getChecksum() {
+        return toHex(getChecksumBytes());
+    }
```
haydenbaker, line 120: Added this convenience method - it is important for when we ask for the checksum, because the constant-checksum overrides this to not format in hex.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/checksums/factory/CrtBasedChecksumProvider.java
haydenbaker, line 1: Copied. Ignore.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/checksums/factory/SdkCrc32.java
haydenbaker, line 1: Copied. Ignore.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/checksums/factory/SdkCrc32C.java
haydenbaker, line 1: Copied. Ignore.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/io/SdkDigestInputStream.java
haydenbaker, line 1: Removed this in favor of `ChecksumInputStream`

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/signer/Checksummer.java
```diff
@@ -15,15 +15,28 @@
...
+import software.amazon.awssdk.http.auth.aws.internal.signer.FlexibleChecksummer;
+import software.amazon.awssdk.http.auth.aws.internal.signer.PrecomputedChecksummer;
+import software.amazon.awssdk.utils.ImmutableMap;
 
 /**
  * An interface for defining how a checksum is formed from a payload synchronously and asynchronously.
+ * <p>
+ * The implementation may choose to also manipulate the request with the checksum, such as adding it as a header.
  */
 @SdkProtectedApi
```
haydenbaker, line 41: This can be internal, but we can make that change when we clean up protected surface.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/signer/DefaultAwsV4HttpSigner.java
```diff
@@ -206,11 +204,8 @@ private static CompletableFuture<AsyncSignedRequest> doSign(AsyncSignRequest<? e
         SdkHttpRequest.Builder requestBuilder = request.request().toBuilder();
 
         CompletableFuture<V4Context> futureV4Context =
-            checksummer.checksum(request.payload().orElse(null)).thenApply(
-                checksum -> {
-                    requestBuilder.putHeader(X_AMZ_CONTENT_SHA256, checksum);
-                    return requestSigner.sign(requestBuilder);
-                });
+            checksummer.checksum(request.payload().orElse(null), requestBuilder)
+                       .thenApply(__ -> requestSigner.sign(requestBuilder));
```
haydenbaker, line 208: Is there a better way to do this without having an unused arg?

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/signer/Checksummer.java
cenedhryn: Yes, should we add a TODO to remember?

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/signer/Checksummer.java
```diff
@@ -35,12 +48,49 @@ static Checksummer create() {
     }
 
     /**
-     * Given a payload, calculate a checksum and return it as a string.
+     * Get a flexible checksummer that uses the given checksum-algorithm and the default.
```
cenedhryn, line 51: The default what?

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/checksums/Crc32CChecksum.java
```diff
@@ -0,0 +1,97 @@
...
+        } else {
+            crc32c = cloneChecksum(lastMarkedCrc32C);
+        }
+    }
+
+
+    private Checksum cloneChecksum(Checksum checksum) {
+        if (isCrtBasedChecksum) {
+            try {
+                Method method = checksum.getClass().getDeclaredMethod("clone");
```
zoewangg, line 87: Reflection is expensive. Can we make CRT an optional dependency in this module? This way, it's available at build time. See https://github.com/aws/aws-sdk-java-v2/blob/77e7880fd8d22767f5a16bc3ca80401174b2db07/services/s3/pom.xml#L88

haydenbaker: Sure. This is copied, but I agree with the improvement.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/signer/DefaultAwsV4HttpSigner.java
cenedhryn: Not that I know of.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/util/ChecksumUtil.java
```diff
@@ -0,0 +1,116 @@
...
+     * x-amz-checksum-sha256, x-amz-checksum-sha1, x-amz-checksum-crc32, x-amz-checksum-crc32c, x-amz-checksum-md5
+     * </p>
+     */
+    public static String checksumHeaderName(ChecksumAlgorithm checksumAlgorithm) {
+        return "x-amz-checksum-" + checksumAlgorithm.algorithmId().toLowerCase(Locale.US);
+    }
+
+    /**
+     * Gets the SdkChecksum object based on the given ChecksumAlgorithm. Instances for CRC32C, CRC32 Algorithm will be added from
+     * CRT Java library once they are available in release.
```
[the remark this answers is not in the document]
haydenbaker: Whoops. I think this was a paste error. It should just say the first sentence.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/signer/Checksummer.java
```diff
@@ -35,12 +48,50 @@ static Checksummer create() {
...
+            return new FlexibleChecksummer(checksums);
+        }
+        return create(checksum);
+    }
+
+    /**
+     * Given a payload, calculate a checksum and add it to the request.
      */
-    String checksum(ContentStreamProvider payload);
+    void checksum(ContentStreamProvider payload, SdkHttpRequest.Builder request);
```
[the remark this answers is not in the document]
haydenbaker: Before flexible checksums, the checksummer only ever returned the checksum as a value. Now, the checksummer has to be able to do multiple things (i.e. compute a checksum, add it to the request, compute another checksum, add it to the request somewhere else).

core/http-auth-aws/src/test/java/software/amazon/awssdk/http/auth/aws/internal/signer/DefaultAwsV4HttpSignerTest.java
```diff
@@ -231,4 +233,37 @@ public void sign_WithPayloadSigningFalseAndChunkEncodingTrueWithoutTrailer_Throw
 
         assertThrows(UnsupportedOperationException.class, () -> signer.sign(request));
     }
+
+    @Test
+    public void sign_withFlexibleChecksum_DelegatesToFlexibleChecksummer() {
```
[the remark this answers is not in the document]
haydenbaker: I want to be a little more verbose, and I think these tests should only roughly test the delegation, not necessarily the entire logic E2E (which is why some of the tests only approximate the behavior by checking if a value is present or not).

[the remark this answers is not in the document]
haydenbaker: Updated to `sign_withChecksumAlgorithm_DelegatesToChecksummerWithThatChecksumAlgorithm` and `sign_withChecksumAlgorithmAndPayloadSigningFalse_DelegatesToChecksummerWithThatChecksumAlgorithm`

core/http-auth-aws/src/test/java/software/amazon/awssdk/http/auth/aws/internal/signer/DefaultAwsV4HttpSignerTest.java
```diff
@@ -231,4 +233,37 @@ public void sign_WithPayloadSigningFalseAndChunkEncodingTrueWithoutTrailer_Throw
...
+            AwsCredentialsIdentity.create("access", "secret"),
+            httpRequest -> {
+            },
+            signRequest -> signRequest.putProperty(CHECKSUM_ALGORITHM, CRC32)
+        );
+
+        SyncSignedRequest signedRequest = signer.sign(request);
+
+        assertThat(signedRequest.request().firstMatchingHeader("x-amz-checksum-crc32")).isPresent();
+        assertThat(signedRequest.request().firstMatchingHeader("x-amz-content-sha256")).isPresent();
```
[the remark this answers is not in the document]
haydenbaker: Hmmm.. I'd prefer not to, since I think these tests should try to focus on the delegation part itself instead. I do think more tests should be added to your point for specific behaviors like that.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/checksums/Crc32CChecksum.java
```diff
@@ -0,0 +1,92 @@
...
+    public void reset() {
+        if (lastMarkedCrc32C == null) {
+            crc32c.reset();
+        } else {
+            crc32c = cloneChecksum(lastMarkedCrc32C);
+        }
+    }
+
+
+    private Checksum cloneChecksum(Checksum checksum) {
```
zoewangg, line 84: Suggesting using `instanceof` check. This way we don't need another flag

```
    private Checksum cloneChecksum(Checksum checksum) {
        if (checksum instanceof CRC32C) {
            return (Checksum) ((CRC32C) checksum).clone();
        }

        if (checksum instanceof SdkCrc32C) {
            return (Checksum) ((SdkCrc32C) checksum).clone();
        }

        throw new IllegalStateException("Unsupported checksum");
    }
```

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/checksums/Crc32Checksum.java
```diff
@@ -0,0 +1,89 @@
...
+    @Override
+    public void reset() {
+        if ((lastMarkedCrc32 == null)) {
+            crc32.reset();
+        } else {
+            crc32 = cloneChecksum(lastMarkedCrc32);
+        }
+    }
+
+    private Checksum cloneChecksum(Checksum checksum) {
```
zoewangg, line 82: Same here

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/checksums/factory/CrtBasedChecksumProvider.java
```diff
@@ -0,0 +1,78 @@
...
+
+    public static Checksum createCrc32C() {
+        return createCrtBasedChecksum(CRT_CRC32_C_CLASS_LOADER);
+    }
+
+    private static Checksum createCrtBasedChecksum(Lazy<Optional<Class<?>>> lazyClassLoader) {
+        return lazyClassLoader.getValue().map(
+            checksumClass -> {
+                try {
+                    return (Checksum) checksumClass.getDeclaredConstructor().newInstance();
```
zoewangg, line 56: I don't think we need to instantiate the class using reflection anymore now that crt is an optional dependency. Suggesting exposing something like below and using this method in Crc32CChecksum and Crc32Checksum

```
    public static boolean isCrtAvailable() {
        try {
            ClassLoaderHelper.loadClass(CRT_CLASSPATH_FOR_CRC32, false);
        } catch (ClassNotFoundException e) {
           return false;
        };
        
        return true;
    }
```

```
    public Crc32CChecksum() {
        if (isCrtAvailable()) {
            crc32c = new CRC32C();
        } else {
            crc32c = SdkCrc32C.create();
        }
    }
```

core/http-auth-aws/src/test/java/software/amazon/awssdk/http/auth/aws/internal/io/ChecksumInputStreamTest.java
```diff
@@ -0,0 +1,66 @@
...
+import org.junit.jupiter.api.Test;
+import software.amazon.awssdk.http.auth.aws.internal.checksums.Crc32Checksum;
+import software.amazon.awssdk.http.auth.aws.internal.checksums.SdkChecksum;
+import software.amazon.awssdk.http.auth.aws.internal.checksums.Sha256Checksum;
+import software.amazon.awssdk.utils.BinaryUtils;
+
+public class ChecksumInputStreamTest {
+
+    @Test
+    public void test_computesCorrectSha256() {
```
zoewangg, line 34: nit, we don't really use "test" as part of test name since it's unnecessary. The patterns we've established are  `methodToTest_when_expectedBehavior` or `given_when_then`

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/checksums/Crc32CChecksum.java
zoewangg: Nice, thank you!

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/io/ChecksumSubscriber.java
```diff
@@ -46,18 +50,10 @@ public DigestComputingSubscriber(MessageDigest messageDigest) {
...
-        } catch (NoSuchAlgorithmException e) {
-            throw new RuntimeException("Unable to create SHA-256 computing subscriber" + e.getMessage());
-        }
-    }
-
     @Override
     public void onSubscribe(Subscription subscription) {
         synchronized (this) {
-            if (!canceled) {
+            if (!canceled && this.subscription == null) {
```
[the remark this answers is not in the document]
haydenbaker: TCK test is added, I think this comment became stale after I rebased some changes and pushed.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/io/ChecksumInputStream.java
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * An input-stream that takes a collection of checksums, and updates each checksum when it reads data.
+ */
+@SdkInternalApi
+public class ChecksumInputStream extends FilterInputStream {
+
+    private final Collection<Checksum> checksums = new ArrayList<>();
+
+    public ChecksumInputStream(InputStream stream, Collection<? extends Checksum> checksums) {
```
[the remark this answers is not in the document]
haydenbaker: Yep! It takes multiple because multiple "checksums" are computed when a flexible checksum is specified in the signing request - the "flexible"checksum and the sha-256 content hash (or one of the pre-defined constant values, which is why we need the `ConstantChecksum`)

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/checksums/ConstantChecksum.java
```diff
@@ -0,0 +1,63 @@
...
+package software.amazon.awssdk.http.auth.aws.internal.checksums;
+
+import java.nio.charset.StandardCharsets;
+import software.amazon.awssdk.annotations.SdkInternalApi;
+
+/**
+ * Implementation of {@link SdkChecksum} to provide a constant checksum.
+ */
+@SdkInternalApi
+public class ConstantChecksum implements SdkChecksum {
```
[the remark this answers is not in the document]
haydenbaker: Touched upon here: https://github.com/aws/aws-sdk-java-v2/pull/4376#discussion_r1313298120

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/signer/DefaultAwsV4HttpSigner.java
```diff
@@ -235,10 +230,10 @@ private static boolean isEventStreaming(SdkHttpRequest request) {
...
     }
 
-    private static boolean hasTrailer(SdkHttpRequest request) {
-        // TODO(sra-identity-and-auth): Trailer would be determined by being a flexible-checksum enabled request, we will need
-        //  to update this once flexible checksums is enabled
-        return request.firstMatchingHeader(X_AMZ_TRAILER).isPresent();
+    private static boolean hasTrailer(SignRequest<?, ? extends AwsCredentialsIdentity> request) {
+        // Flexible checksums dictates adding a trailer when signing, but there may be existing trailers on
+        // the request (in the future)
+        return request.hasProperty(CHECKSUM_ALGORITHM) || request.request().firstMatchingHeader(X_AMZ_TRAILER).isPresent();
```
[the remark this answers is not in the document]
haydenbaker: I can see how this would be confusing. Changing.

core/http-auth-aws/src/test/java/software/amazon/awssdk/http/auth/aws/internal/signer/DefaultAwsV4HttpSignerTest.java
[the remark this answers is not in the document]
cenedhryn: Winning with the longest name! 

I do disagree though when the test contains references to constructs that aren't in the test, but I accept your POV :-)

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/signer/DefaultAwsV4HttpSigner.java
cenedhryn: @haydenbaker answered it below, it seems to be because the code will be touching the request in more than one place. Perhaps it's easiest to review once all PRs touching this feature are in place.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/io/ChecksumInputStream.java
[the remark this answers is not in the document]
cenedhryn: Do you have any use cases defined in tests for these variants?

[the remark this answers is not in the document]
haydenbaker: Yep. The `ChecksumInputStreamTest` and `FlexibleChecksummerTest` should test multiple checksums being configured.

cenedhryn (approved):

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/signer/FlexibleChecksummer.java
```diff
@@ -0,0 +1,79 @@
...
+
+    public FlexibleChecksummer(Map<String, ChecksumAlgorithm> algorithmMap) {
+        this.checksumMap = algorithmMap
+            .entrySet()
+            .stream()
+            .collect(Collectors.toMap(Map.Entry::getKey, v -> fromChecksumAlgorithm(v.getValue())));
+    }
+
+    @Override
+    public void checksum(ContentStreamProvider payload, SdkHttpRequest.Builder request) {
```
zoewangg, line 54: Instead of passing request builder and letting each implementation decide what to do with it, can we pass how we want to consume the string to it, i.e, `Consumer<String> checksumConsumer`? This way, `X_AMZ_CONTENT_SHA256` is not leaked in checksum implementations and it gives us flexibility to support a new header or new field.

Something like this:

```java
    public FlexibleChecksummer(Map<Consumer<String>, ChecksumAlgorithm> algorithmMap) {
        this.checksumMap = algorithmMap
            .entrySet()
            .stream()
            .collect(Collectors.toMap(Map.Entry::getKey, v -> fromChecksumAlgorithm(v.getValue())));
    }

    @Override
    public void checksum(ContentStreamProvider payload) {
        InputStream payloadStream = getBinaryRequestPayloadStream(payload);

        ChecksumInputStream computingStream = new ChecksumInputStream(
            payloadStream,
            checksumMap.values()
        );

        readAll(computingStream);

        checksumMap.forEach((consumer, checksum) -> consumer.accept(checksum.getChecksum()));
    }

```

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/signer/DefaultChecksummer.java
```diff
@@ -15,39 +15,59 @@
...
+import software.amazon.awssdk.http.auth.aws.internal.io.ChecksumSubscriber;
 import software.amazon.awssdk.http.auth.aws.signer.Checksummer;
-import software.amazon.awssdk.utils.BinaryUtils;
 
-@SdkProtectedApi
+/**
+ * The default implementation of a checksummer. By default, this will calculate the SHA256 checksum of a payload and add it as the
+ * value for the 'x-amz-content-sha256' header on the request.
+ */
+@SdkInternalApi
```
zoewangg, line 42: Do we need this class? Since technically, this is covered in FlexibleChecksummer by passing just SHA256

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/signer/PrecomputedChecksummer.java
```diff
@@ -15,15 +15,23 @@
...
 import software.amazon.awssdk.http.auth.aws.signer.Checksummer;
 
-@SdkProtectedApi
+/**
+ * An implementation of a checksummer that simply passes along a computed value as a checksum. Specifically, this is used in
+ * the cases where the checksum is a pre-defined value that dictates specific behavior by the signer, and flexible checksums is
+ * not enabled for the request (such as aws-chunked payload signing without trailers, unsigned streaming without trailers, etc.).
+ */
+@SdkInternalApi
 public final class PrecomputedChecksummer implements Checksummer {
```
zoewangg, line 35: Do we need this class? It seems like we can use `FlexibleChecksummer` and pass `ConstantChecksumAlgorithm` right?

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/util/ChecksumUtil.java
```diff
@@ -0,0 +1,118 @@
...
+     * </p>
+     */
+    public static String checksumHeaderName(ChecksumAlgorithm checksumAlgorithm) {
+        return "x-amz-checksum-" + checksumAlgorithm.algorithmId().toLowerCase(Locale.US);
+    }
+
+    /**
+     * Gets the SdkChecksum object based on the given ChecksumAlgorithm.
+     */
+    public static SdkChecksum fromChecksumAlgorithm(ChecksumAlgorithm checksumAlgorithm) {
```
zoewangg, line 58: We may want to create a map for this to make it easier to maintain and may be a micro performance improvement `Map<ChecksumAlgorithm, Supplier<SdkChecksum>>`

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/signer/PrecomputedChecksummer.java
haydenbaker: This is an optimization, using this class means it doesn't need to read the stream like it does with Flexible.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/signer/DefaultChecksummer.java
haydenbaker: Nope, you're right, it can be removed.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/signer/FlexibleChecksummer.java
zoewangg: I'm not suggesting passing request instance to this class during initialization though; I'm suggesting passing how we consume the request builder to this class. 
```
        FlexibleChecksummer checksummer = new FlexibleChecksummer(
            ImmutableMap.of(SHA256, value -> request.putHeader("sha256", value)));
```

IMO, the sole responsibility of checksummer class should be calculating checksum and now it seems it does multiple things and violates SRP, which makes it harder to extend in the future. For example, we may need to save the checksum in a different header for certain operations. It seems with current implementations, we'd need another checksummer class.

zoewangg: Discussed offline, since this is an internal API, having another layer of handling the checksum values seems a bit overkill.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/signer/PrecomputedChecksummer.java
zoewangg: Can we rename it to `PrecomputedSha256Checksummer` to make it more clear?

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/signer/Checksummer.java
```diff
@@ -35,12 +48,50 @@ static Checksummer create() {
...
+     */
+    static Checksummer create(String checksum) {
+        return new PrecomputedChecksummer(() -> checksum);
+    }
+
+    /**
+     * Get a flexible checksummer that performs two checksums: the given checksum-algorithm and a precomputed checksum from the
+     * given checksum string.
+     */
+    static Checksummer create(String checksum, ChecksumAlgorithm checksumAlgorithm) {
```
zoewangg, line 76: Suggesting `static Checksummer forFlexibleChecksum(String preComputedSha256, ChecksumAlgorithm checksumAlgorithm)`

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/signer/FlexibleChecksummer.java
```diff
@@ -0,0 +1,79 @@
...
+import software.amazon.awssdk.http.auth.aws.signer.Checksummer;
+
+/**
+ * A "flexible" implementation of a checksummer. It takes a map of checksums and their header names, computes them efficiently
+ * by updating each checksum while reading the payload (once), and adds the computed checksum strings to the request using the
+ * given header names in the map. This should be used in cases where a (flexible) checksum algorithm is present during signing.
+ */
+@SdkInternalApi
+public final class FlexibleChecksummer implements Checksummer {
+    private final Map<String, SdkChecksum> checksumMap;
```
zoewangg, line 44: Nit: `Map<SdkChecksum, String>` seems to be more intuitive. Can we rename it to `checksumToHeader` or `headerToChecksum` if you want to keep it as is.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/signer/Checksummer.java
```diff
@@ -35,12 +48,50 @@ static Checksummer create() {
...
+        if (checksumAlgorithm != null) {
```
zoewangg, line 77: Can we require chekcsumAlgorithm to be not null

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/signer/Checksummer.java
```diff
@@ -35,12 +48,50 @@ static Checksummer create() {
...
+
+            return new FlexibleChecksummer(checksums);
+        }
+        return create();
+    }
+
+    /**
+     * Get a precomputed checksummer that results the given checksum string.
+     */
+    static Checksummer create(String checksum) {
```
zoewangg, line 68: Suggesting `static Checksummer forPrecomputed256Checksum(String preComputedSha256)`

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/signer/Checksummer.java
```diff
@@ -35,12 +48,50 @@ static Checksummer create() {
     }
 
     /**
-     * Given a payload, calculate a checksum and return it as a string.
+     * Get a flexible checksummer that performs two checksums: the given checksum-algorithm and the default (sha256).
+     */
+    static Checksummer create(ChecksumAlgorithm checksumAlgorithm) {
```
zoewangg, line 53: Suggesting `static Checksummer forFlexibleChecksum(ChecksumAlgorithm checksumAlgorithm)` and requiring checksumAlgorithims to be not null.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/util/ChecksumUtil.java
haydenbaker: We can do this for the most, we can't for ConstantChecksum, since it takes a value as part of construction. Updating for this case.

zoewangg (approved):
Feel free to merge once require chekcsumAlgorithm thing is fixed.

PR #4379, opened 2023-08-31
Performance Enhancement :Enable Passing of requestPath to CRT and monitor the progress using CRTs S3MetaRequestProgress 

joviegas:
## Motivation and Context

- The File IO operation during Upload of the File should not be done in SDK side.
- The requestPath is passed to the CRT and CRT does the IO operation.
 - CRT side changes  https://github.com/awslabs/aws-crt-java/pull/595 and https://github.com/awslabs/aws-c-s3/pull/344 

## Modifications

- Removed the code where S3CrtRequestBodyStreamAdapter is passed to CRTHttp client
- Now we just pass null is request Path is passed to the S3MetaRequest.
- The Progress listener will listen to S3MetaRequestProgress from CRT

## Testing
- No need of new Test since this is internal behaviour test, the existing test cases should cover this.
- Executed Integ tests.

## Types of changes

## License


services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/CrtS3TransferManager.java
```diff
@@ -65,29 +65,25 @@ public FileUpload uploadFile(UploadFileRequest uploadFileRequest) {
...
         progressUpdater.transferInitiated();
-        requestBody = progressUpdater.wrapRequestBody(requestBody);
         progressUpdater.registerCompletion(returnFuture);
 
         try {
             assertNotUnsupportedArn(putObjectRequest.bucket(), "upload");
 
             CompletableFuture<PutObjectResponse> crtFuture =
-                s3AsyncClient.putObject(putObjectRequest, requestBody);
+                s3AsyncClient.putObject(putObjectRequest, AsyncRequestBody.fromFile(uploadFileRequest.source()));
```
millems, line 86: If CRT is reading directly from disk, does this publisher get used?

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/S3CrtProgressListener.java
```diff
@@ -0,0 +1,80 @@
...
+package software.amazon.awssdk.services.s3.internal.crt;
+
+import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.core.async.listener.PublisherListener;
+import software.amazon.awssdk.crt.s3.S3MetaRequestProgress;
+
+/**
+ * S3CrtProgressListener delegates events to the underlying delegateListener if defined, avoiding null checks for API when using
+ * S3MetaRequestProgress for GET calls.
+ */
```
millems, line 25: This feels like a weird application of polymorphism. Instead of S3CrtProgressListener, can we just have a NoOpPublisherListener that we create if delegateListener is null? That just means we need the null check in the setup code of S3CrtResponseHandlerAdapter instead of per-method here.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/S3InternalSdkHttpExecutionAttribute.java
```diff
@@ -41,6 +43,13 @@ public final class S3InternalSdkHttpExecutionAttribute<T> extends SdkHttpExecuti
     public static final S3InternalSdkHttpExecutionAttribute<Region> SIGNING_REGION =
         new S3InternalSdkHttpExecutionAttribute<>(Region.class);
 
+    public static final S3InternalSdkHttpExecutionAttribute<Path> SOURCE_REQ_PATH =
+        new S3InternalSdkHttpExecutionAttribute<>(Path.class);
+
+
+    public static final S3InternalSdkHttpExecutionAttribute<PublisherListener> CRT_PROGRESS_LISTENER =
+        new S3InternalSdkHttpExecutionAttribute<>(PublisherListener.class);
```
millems, line 51: This is marked as @SdkInternalApi, so it can't be used by the transfer manager module, like we're doing above. Do we need to move a subset to a @SdkProtectedApi?

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/S3InternalSdkHttpExecutionAttribute.java
```diff
@@ -41,6 +43,13 @@ public final class S3InternalSdkHttpExecutionAttribute<T> extends SdkHttpExecuti
     public static final S3InternalSdkHttpExecutionAttribute<Region> SIGNING_REGION =
         new S3InternalSdkHttpExecutionAttribute<>(Region.class);
 
+    public static final S3InternalSdkHttpExecutionAttribute<Path> SOURCE_REQ_PATH =
+        new S3InternalSdkHttpExecutionAttribute<>(Path.class);
```
millems, line 47: Naming nitpicks, since this actually seems like a protected API.
* `SOURCE` - Do we ever expect to use this for something other than a PutObject - like a GetObject, where it's actually the `DESTINATION`?
* `REQ` - We should almost always use the full word for something instead of shortening it, so `REQUEST`. It looks like a lot of the existing properties are request properties, though, so I don't think REQUEST is necessary.

If it's always the source, what about `PUT_OBJECT_PATH`? `UPLOAD_PATH`?
If it's potentially also destination, what about `UPLOAD_DOWNLOAD_PATH`?

millems (commented):
* Is it possible to test that what we're intending to fix actually got fixed?
* Is there performance testing to verify that this change didn't hurt performance?
* Do we already have tests that assert the progress listener changes were backwards compatible?

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/S3InternalSdkHttpExecutionAttribute.java
joviegas: Done

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/S3InternalSdkHttpExecutionAttribute.java
joviegas: Thanks Matt..
Named it as OBJECT_FILE_PATH so that it can be used for any file path for get/put

joviegas:
> * Is it possible to test that what we're intending to fix actually got fixed?
> 
Thanks Matt for this suggestion , tested this and realized the CRT doesnot actually fix the issue where upload file size is more than RAM size. I have cut a different ticket to CRT. However , this fix is still needed for performance enhancement. Removed the mention of the orginal issue.

> * Is there performance testing to verify that this change didn't hurt performance?
> 
Yes , I saw that the latencies was reduced to half , I have updated the internal doc.

> * Do we already have tests that assert the progress listener changes were backwards compatible?
> 
Yes , we already have test .

> * Are there any path-processing nuances that may have changed when we switched to the CRT? E.g. handling of the current working directory or expansion of ~/?

No.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/S3CrtResponseHandlerAdapter.java
```diff
@@ -46,9 +48,14 @@ public final class S3CrtResponseHandlerAdapter implements S3MetaRequestResponseH
     private final SdkHttpResponse.Builder respBuilder = SdkHttpResponse.builder();
     private volatile S3MetaRequest metaRequest;
 
-    public S3CrtResponseHandlerAdapter(CompletableFuture<Void> executeFuture, SdkAsyncHttpResponseHandler responseHandler) {
+    private final PublisherListener<S3MetaRequestProgress> progressListener;
```
zoewangg, line 51: Do we need to invoke `this.progressListener.subscriberOnError();` in `onErrorResponseComplete`?

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/CrtContentLengthOnlyAsyncFileRequestBody.java
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    @Override
+    public Optional<Long> contentLength() {
+        return asyncRequestBody.contentLength();
+    }
+
+    @Override
+    public void subscribe(Subscriber<? super ByteBuffer> subscriber) {
+        subscriber.onSubscribe(new Subscription() {
```
zoewangg, line 41: It doesn't seem to comply with the spec
>Subscription.request and Subscription.cancel MUST only be called inside of its Subscriber context.

We can probably use [NoopSubscription](https://github.com/aws/aws-sdk-java-v2/blob/master/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/util/NoopSubscription.java) here. We need to move it outside of the internal package though since it's used cross module

```

            subscriber.onSubscribe(new NoopSubscription(s));
            subscriber.onError(new UnsupportedOperationException());
```

Alternatively, we can follow the same logic, i.e., calling onError in `request` and no-op in cancel.

services/s3/src/it/java/software/amazon/awssdk/services/s3/crt/S3CrtClientPutObjectIntegrationTest.java
```diff
@@ -86,6 +90,25 @@ void putObject_fileRequestBody_objectSentCorrectly() throws Exception {
...
+        s3Crt.putObject(r -> r.bucket(TEST_BUCKET).key(TEST_KEY), testFile.toPath()).join();
+        ResponseInputStream<GetObjectResponse> objContent = S3IntegrationTestBase.s3.getObject(r -> r.bucket(TEST_BUCKET).key(TEST_KEY),
+                                                                                               ResponseTransformer.toInputStream());
+        byte[] expectedSum = ChecksumUtils.computeCheckSum(Files.newInputStream(testFile.toPath()));
+        Assertions.assertThat(ChecksumUtils.computeCheckSum(objContent)).isEqualTo(expectedSum);
+    }
+
+
+    @Test
+    void putObject_ContentOnlyAsynRequestBody_fails()  {
```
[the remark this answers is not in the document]
zoewangg: Nvm, got it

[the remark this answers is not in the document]
joviegas: making sure that when CrtContentLengthOnlyAsyncFileRequestBody is used and if we try to subscribe it then it will fail

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/CrtContentLengthOnlyAsyncFileRequestBody.java
joviegas: Will use `Alternatively` since its better to fail fast in such cases since its error if we try to subscribe CrtContentLengthOnlyAsyncFileRequestBody

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/S3TmInternalSdkHttpExecutionAttribute.java
```diff
@@ -0,0 +1,37 @@
...
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.services.s3.internal.crt;
```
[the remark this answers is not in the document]
joviegas: done

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/S3TmInternalSdkHttpExecutionAttribute.java
```diff
@@ -0,0 +1,37 @@
...
+
+
+import software.amazon.awssdk.annotations.SdkProtectedApi;
+import software.amazon.awssdk.core.async.listener.PublisherListener;
+import software.amazon.awssdk.http.SdkHttpExecutionAttribute;
+
+@SdkProtectedApi
+public class S3TmInternalSdkHttpExecutionAttribute<T> extends SdkHttpExecutionAttribute<T> {
```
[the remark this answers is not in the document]
joviegas: done

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/progress/TransferProgressUpdater.java
```diff
@@ -107,8 +107,7 @@ public void publisherSubscribe(Subscriber<? super S3MetaRequestProgress> subscri
 
             @Override
             public void subscriberOnNext(S3MetaRequestProgress s3MetaRequestProgress) {
-                incrementBytesTransferred(Math.toIntExact(s3MetaRequestProgress.getBytesTransferred()),
-                                          s3MetaRequestProgress.getContentLength());
+                incrementBytesTransferred(Math.toIntExact(s3MetaRequestProgress.getBytesTransferred()));
```
zoewangg, line 110: We probably should not cast it to int. Can we change incrementBytesTransferred to take a long instead?

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/CrtContentLengthOnlyAsyncFileRequestBody.java
zoewangg: Oops, apologies, I mistyped, I meant we can follow the same logic in this class if we don't want to make NoopSubscription protected. We have to call onSubscribe first per reactive streams spec.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/S3CrtResponseHandlerAdapter.java
zoewangg: `failResponseHandlerAndFuture` doesn't seem to be invoked from `onErrorResponseComplete` if it successfully marshalls the error response though. Can we add a test to verify it? [This test](https://github.com/aws/aws-sdk-java-v2/blob/master/services-custom/s3-transfer-manager/src/test/java/software/amazon/awssdk/transfer/s3/internal/S3TransferManagerListenerTest.java#L65) may not suffice now that we are relying on the real CRT client to report progress and it is using a mock s3 client.

```
        responsePublisher.send(ByteBuffer.wrap(errorPayload))
                         .thenRun(responsePublisher::complete)
                         .handle((ignore, throwable) -> {
                             if (throwable != null) {
                                 failResponseHandlerAndFuture(throwable);
                                 return null;
                             }
                             completeFutureAndCloseRequest();
                             return null;
                         });
```

zoewangg (approved):

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/S3CrtResponseHandlerAdapter.java
joviegas: Added new mock tests and removed  explicit onError

PR #4383, opened 2023-08-31
Add flexible checksum support for chunked-encoding cases

haydenbaker:
## Motivation and Context
- Same as #4376, but for chunk-encoding cases

## Modifications
- Updates `AwsChunkedEncodedInputStream` to take a checksum-algorithm parameter
- Flexible checksum is now added to the encoded stream as a trailer
- Pre-existing trailers on the request (which there are none today) will be added as trailers to the stream (and removed from the request) as well

## Types of changes

## Checklist

- [] I have updated the Javadoc documentation accordingly

## License

### DIFF
--- a/core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/signer/AwsChunkedV4PayloadSigner.java
+++ b/core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/signer/AwsChunkedV4PayloadSigner.java
@@ -15,12 +15,16 @@
 
 package software.amazon.awssdk.http.auth.aws.internal.signer;
 
+import static software.amazon.awssdk.http.auth.aws.internal.util.ChecksumUtil.checksumHeaderName;
+import static software.amazon.awssdk.http.auth.aws.internal.util.ChecksumUtil.fromChecksumAlgorithm;
+import static software.amazon.awssdk.http.auth.aws.signer.V4CanonicalRequest.getCanonicalHeaders;
 import static software.amazon.awssdk.http.auth.aws.signer.V4CanonicalRequest.getCanonicalHeadersString;
 import static software.amazon.awssdk.http.auth.aws.util.SignerConstant.STREAMING_SIGNED_PAYLOAD;
 import static software.amazon.awssdk.http.auth.aws.util.SignerConstant.STREAMING_SIGNED_PAYLOAD_TRAILER;
 import static software.amazon.awssdk.http.auth.aws.util.SignerConstant.STREAMING_UNSIGNED_PAYLOAD_TRAILER;
 import static software.amazon.awssdk.http.auth.aws.util.SignerConstant.X_AMZ_CONTENT_SHA256;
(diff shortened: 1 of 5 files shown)


core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/signer/AwsChunkedV4PayloadSigner.java
```diff
@@ -56,10 +64,12 @@ public class AwsChunkedV4PayloadSigner implements V4PayloadSigner {
 
     private final CredentialScope credentialScope;
     private final int chunkSize;
+    private final ChecksumAlgorithm checksumAlgorithm;
 
-    public AwsChunkedV4PayloadSigner(CredentialScope credentialScope, int chunkSize) {
+    public AwsChunkedV4PayloadSigner(CredentialScope credentialScope, int chunkSize, ChecksumAlgorithm checksumAlgorithm) {
```
haydenbaker, line 69: Will change this to a builder.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/signer/AwsChunkedV4PayloadSigner.java
```diff
@@ -169,7 +207,49 @@ private void setupSigTrailer(ChunkedEncodedInputStream.Builder builder, RollingS
...
     }
 
-    private void setupChecksumTrailer(ChunkedEncodedInputStream.Builder builder) {
-        // TODO(sra-identity-and-auth): Set up checksumming of chunks and add as a trailer
+    /**
+     * Add the checksum as a chunk-trailer and add it to the request's trailer header.
+     * <p>
+     * The checksum-algorithm MUST be set if this is called, otherwise it will throw.
+     */
+    private void setupChecksumTrailer(ChunkedEncodedInputStream.Builder builder, SdkHttpRequest.Builder request) {
```
zoewangg, line 215: Maybe we can rename it to `setUpChecksumTrailerIfNeeded` and return if checksumAlgorithm is null? Then we can remove the checksumAlgorithm null check from the callers

haydenbaker: Sure, if that's an established naming pattern, I can do that

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/signer/AwsChunkedV4PayloadSigner.java
```diff
@@ -252,4 +252,49 @@ private void setupPreExistingTrailers(ChunkedEncodedInputStream.Builder builder,
...
     }
+
+    public static Builder builder() {
+        return new BuilderImpl();
+    }
+
+    public interface Builder {
+        Builder credentialScope(CredentialScope credentialScope);
+
+        Builder chunkSize(int chunkSize);
```
zoewangg, line 263: Should probably be boxed type so that we can throw an exception or set a default if the caller doesn't set it.

zoewangg (approved):

Issue #4354, opened 2023-08-26
Kinesis Async Client hangs after java.io.IOException: Response had content-length of 28 bytes, but only received 0 bytes before the connection was closed


MichalZalewskiRASP:
### Describe the bug

I get the same defect as described here: https://github.com/aws/aws-sdk-java-v2/issues/3335 when I use the SubscribeToShard API. It hungs and no other visitor.visit() is executed after the Kinesis Async Client throws 
`java.io.IOException: Response had content-length of 28 bytes, but only received 0 bytes before the connection was closed.`
Basically, stream consumption is halted afterwards regardless of the fact that I use the `onError` handler in the request.

### Expected Behavior

The SubscribeToShard API should be able to resume stream consumption after encountering IOException.

### Current Behavior

The thread hangs and the client does not continue to consume the stream.

### Reproduction Steps

It is difficult to reproduce as it happens once per week randomly when the message length does not pass the validation.
My dependencies:
```
<dependency>
            <groupId>software.amazon.kinesis</groupId>
            <artifactId>amazon-kinesis-client</artifactId>
            <version>2.5.2</version>
</dependency>
 <dependency>
            <groupId>software.amazon.awssdk</groupId>
            <artifactId>url-connection-client</artifactId>
            <version>2.20.123</version>
            <scope>test</scope>
</dependency>
 <dependency>
            <groupId>software.amazon.awssdk</groupId>
            <artifactId>sts</artifactId>
            <version>2.20.123</version>
  </dependency>
```
Here is my setup.

I call (in Kotlin):
```
kinesisClient.subscribeToShard(requestToSubscribe, responseHandler).get()
```
whereby:
* `kinesisClient` is `software.amazon.awssdk.services.kinesis.KinesisAsyncClient`.
* `` is defined as:
```
SubscribeToShardRequest.builder()
        .consumerARN(<some efo consumer arn>)
        .shardId(<some shard name>)
        .startingPosition(<some starting position>)
        .build()
```
* `` is defined as:
```
SubscribeToShardResponseHandler
        .builder()
        .onResponse{
            log.info("Subscription to Kinesis stream started {}", it)
        }
        .onComplete {
            log.info("Subscription to Kinesis stream completed")
        }
        .onError { t: Throwable ->
            when (t) {
                is ResourceInUseException -> log.debug("Other instance is already subscribed to Kinesis: " + t.message)
                else -> log.error("Error when reading data from processor: stack={}", t.stackTrace)
            }
[...]

Issue #3335, opened 2022-08-02
Uncaught exception stops KinesisAsyncClient


MediaMarco:
### Describe the bug

We use the KinesisAsyncClient to read messages from Kinesis like that:

`
GetRecordsResponse response = kinesisClient.getRecords(GetRecordsRequest.builder().shardIterator(id).build()).join();
`

This worked fine for ages, but since SDK version 2.17.115 the reader sometimes stalls and the join() blocks forever.

It seems to have something to do with this change:

> # __2.17.115__ __2022-01-20__
> ## __Netty NIO HTTP Client__
> - ### Bugfixes
> - Verify that the content-length header matches the content returned by the service.

### Expected Behavior

The CompletableFuture returned by GetRecords should complete exceptionally when the netty I/O error occurs to be able to handle it gracefully in application code.

### Current Behavior

`kinesisClient.getRecords(GetRecordsRequest.builder().shardIterator(id).build()).join();`

blocks forever.

When the blocking occurs, we see the following Exception in our logs, nothing else:

Message: `An exceptionCaught() event was fired, and it reached at the tail of the pipeline. It usually means the last handler in the pipeline did not handle the exception.`

Logger: `io.netty.channel.DefaultChannelPipeline`

Thread: `aws-java-sdk-NettyEventLoop-3-7`

Stack Trace:
`software.amazon.awssdk.http.nio.netty.internal.FutureCancelledException: java.io.IOException: Response had content-length of 25 bytes, but only received 0 bytes before the connection was closed.
	at software.amazon.awssdk.http.nio.netty.internal.NettyRequestExecutor.lambda$null$3(NettyRequestExecutor.java:136)
	at io.netty.util.concurrent.PromiseTask.runTask(PromiseTask.java:98)
	at io.netty.util.concurrent.PromiseTask.run(PromiseTask.java:106)
	at io.netty.util.concurrent.AbstractEventExecutor.runTask(AbstractEventExecutor.java:174)
	at io.netty.util.concurrent.AbstractEventExecutor.safeExecute(AbstractEventExecutor.java:167)
	at io.netty.util.concurrent.SingleThreadEventExecutor.runAllTasks(SingleThreadEventExecutor.java:470)
	at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:503)
	at io.netty.util.concurrent.SingleThreadEventExecutor$4.run(SingleThreadEventExecutor.java:997)
[...]

PR #4402, opened 2023-09-07
Fixed an issue in async client causing resposne future to get stuck

zoewangg:
## Motivation and Context
#4354 #3335 
If there is a server error and the server fails to return response body that matches with the content length specified in the response header, the response future will get stuck. This is because we are not invoking onError for errorResponseHandler that is supposed to fail the return future.

## Modifications
Invoke `errorResponseHandler#onError` when onError is invoked.

## Testing
Added functional test

## Types of changes

## Checklist

## License

### DIFF
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/async/CombinedResponseAsyncHttpResponseHandler.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/async/CombinedResponseAsyncHttpResponseHandler.java
@@ -68,6 +68,7 @@ public final class CombinedResponseAsyncHttpResponseHandler<OutputT>
         }
 
         successResponseHandler.onError(error);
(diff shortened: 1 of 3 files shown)


core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/async/CombinedResponseAsyncHttpResponseHandler.java
```diff
@@ -68,6 +68,7 @@ public void onError(Throwable error) {
         }
 
         successResponseHandler.onError(error);
+        errorResponseHandler.onError(error);
```
joviegas, line 71: On line 66, does this comment apply to the errorResponseHandler as well? Should we include a null check here for errorResponseHandler

joviegas (approved):

Issue #4395, opened 2023-09-05
AttributeMap should not close executor service in JDK19


scrocquesel:
### Describe the bug

With JDK19+, the first loop in the `AttributeMap::close` method could lead to deadlock waiting for the executor service to terminate in the same thread that should stop the executor service.

https://github.com/aws/aws-sdk-java-v2/blob/0c5df5e976de731d3639f9e7676f5ee4ca563b33/utils/src/main/java/software/amazon/awssdk/utils/AttributeMap.java#L86-L89

Starting JDK 19+, ExecutorService now implements AutoCloseable.  I know this version is not yet supported but excluding explicitely ExecutorService from the first loop will avoid issues in the future and let users move to newer JDK version

### Expected Behavior

I should be able to Close the Client and terminate the overriden executor service in this order in the same thread.

### Current Behavior

The actual code lead to a deadlock

### Reproduction Steps

```
// provides a custom executor service
ClientOverrideConfiguration.builder().scheduledExecutorService(scheduledExecutorService);

// close the client, then shutdown the given executor
client.close(); // will call scheduledExecutorService.awaitTermination();
scheduledExecutorService.shutdown(); // never reached
```

### Possible Solution

- Explicitely exclude ExecutorService from autocloseable loop in `AttributeMap::close`
- Do not delegate call to the wrapped implementation in `UnmanagedScheduledExecutorService::awaitTermination`

### Additional Information/Context

Stack trace of the origin of the deadlock
```
 java.base@20.0.2/jdk.internal.misc.Unsafe.park(Native Method)
    java.base@20.0.2/java.util.concurrent.locks.LockSupport.parkNanos(LockSupport.java:269)
    org.jboss.threads.EnhancedQueueExecutor.awaitTermination(EnhancedQueueExecutor.java:920)
    software.amazon.awssdk.utils.ScheduledExecutorUtils$UnmanagedScheduledExecutorService.awaitTermination(ScheduledExecutorUtils.java:85)
    java.base@20.0.2/java.util.concurrent.ExecutorService.close(ExecutorService.java:417)
    software.amazon.awssdk.utils.IoUtils.closeQuietly(IoUtils.java:70)
    software.amazon.awssdk.utils.IoUtils.closeIfCloseable(IoUtils.java:87)
    software.amazon.awssdk.utils.AttributeMap.lambda$close$0(AttributeMap.java:87)
[...]

PR #4404, opened 2023-09-07
Update AttributeMap.java

RANJITHROSAN17:
## Motivation and Context
This change is motivated by the need to prevent a deadlock issue that occurs in JDK 19+ when attempting to close an `ExecutorService` included in the `AttributeMap`'s `close` method. The current implementation can lead to a deadlock when closing resources. To resolve this issue, the code has been modified to explicitly exclude `ExecutorService` instances from the `AutoCloseable` loop in `AttributeMap::close`.

## Modifications
- Modified the `AttributeMap::close` method to separate `ExecutorService` instances from other `AutoCloseable` resources, ensuring that they are not closed together. This change prevents potential deadlocks when closing resources.
- Introduced the `shutdownIfExecutorService` method to explicitly shut down `ExecutorService` instances if needed.

## Testing
The changes have been tested thoroughly in a development environment. Testing included:
- Verifying that the deadlock issue no longer occurs when closing the client and `ExecutorService` in the same thread.
- Ensuring that all existing tests continue to pass without issues.
- Validating that the modified code does not introduce new problems or regressions.

## Types of changes

## Checklist
- [ ] My change is to implement 1.11 parity feature, and I have updated [LaunchChangelog](https://github.com/aws/aws-sdk-java-v2/blob/master/docs/LaunchChangelog.md)

## License

### DIFF
--- a/utils/src/main/java/software/amazon/awssdk/utils/AttributeMap.java
+++ b/utils/src/main/java/software/amazon/awssdk/utils/AttributeMap.java
@@ -83,10 +83,14 @@ public final class AttributeMap implements ToCopyableBuilder<AttributeMap.Builde
     }
 
     @Override
-    public void close() {
-        attributes.values().forEach(v -> IoUtils.closeIfCloseable(v, null));
-        attributes.values().forEach(this::shutdownIfExecutorService);
-    }
+    public void close() { 
+     attributes.values().stream()
+             .filter(value -> !(value instanceof ExecutorService))
+             .forEach(v -> IoUtils.closeIfCloseable(v, null)); 
+     attributes.values().stream()
+             .filter(value -> value instanceof ExecutorService)
+             .forEach(this::shutdownIfExecutorService); 
+}
 
     private void shutdownIfExecutorService(Object object) {
         if (object instanceof ExecutorService) {


utils/src/main/java/software/amazon/awssdk/utils/AttributeMap.java
```diff
@@ -83,10 +83,14 @@ public AttributeMap copy() {
     }
 
     @Override
-    public void close() {
-        attributes.values().forEach(v -> IoUtils.closeIfCloseable(v, null));
-        attributes.values().forEach(this::shutdownIfExecutorService);
-    }
+    public void close() { 
+     attributes.values().stream()
```
scrocquesel, line 87: I would comment why we need to explicitely exclude ExecutorService

```suggestion
        // Starting from JDK 19, ExecutorService implements AutoCloseable and will await termination on close
        // When ExecutorService is provided by the user, it can result in thread deadlocks if the ExecutorService
        // is not closed from another thread. ExecutorService are shutdown without waiting for termination
     attributes.values().stream()
```

scrocquesel (commented):
Thanks, I was doing a similar PR to resolve https://github.com/aws/aws-sdk-java-v2/issues/4395

utils/src/main/java/software/amazon/awssdk/utils/AttributeMap.java
```diff
@@ -83,10 +83,14 @@ public AttributeMap copy() {
...
+             .filter(value -> !(value instanceof ExecutorService))
+             .forEach(v -> IoUtils.closeIfCloseable(v, null)); 
+     attributes.values().stream()
+             .filter(value -> value instanceof ExecutorService)
```
tiny-george, line 91: Well, do a filter and later check again in shutdownIfExecutorService method is not a good deal. I would remove this filter or change the method.

dagnir:
Hi, thanks for the PR. This change makes sense to me; I like @ tiny-george's suggestion, can you please implement that?

michaeldimchuk:
> Hi, thanks for the PR. This change makes sense to me; I like @tiny-george's suggestion, can you please implement that?

@dagnir Doesn't look like this PR has been updated in a while, so I created another one that addresses the requested changes and adds some tests: https://github.com/aws/aws-sdk-java-v2/pull/4649. Mind taking a look?

dagnir:
Closing this as #4649 has been merged. Thank you for the PR @RANJITHROSAN17.

PR #4416, opened 2023-09-11
Add auth scheme default impl classes

gosar:
## Motivation and Context

Instead of anonymous inner classes.

## Modifications

Add auth scheme default impl classes instead of anonymous inner classes.
Note https://github.com/aws/aws-sdk-java-v2/pull/4416/commits/5c170777e1a125317afbe9acbb104f5cabf532d8 which creates sigv4a signer instance lazily as that depends on optional dependency http-auth-aws-crt. Whereas other signer instance are created eagerly in static block in AuthScheme impls.

## Testing

s3-transfer-manager was failing earlier because of eager loading of crt signer. fixed with lazy loading.

### DIFF
--- a/core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/AwsV4AuthScheme.java
+++ b/core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/AwsV4AuthScheme.java
@@ -42,14 +42,6 @@ public interface AwsV4AuthScheme extends AuthScheme<AwsCredentialsIdentity> {
     }
 
     /**
-     * Retrieve the scheme ID.
-     */
-    @Override
-    default String schemeId() {
-        return SCHEME_ID;
-    }
-
-    /**
      * Retrieve the {@link AwsCredentialsIdentity} based {@link IdentityProvider} associated with this authentication scheme.
      */
     @Override
--- a/core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/AwsV4AuthScheme.java
+++ b/core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/AwsV4AuthScheme.java
@@ -16,6 +16,7 @@
 package software.amazon.awssdk.http.auth.aws;
 
 import software.amazon.awssdk.annotations.SdkPublicApi;
+import software.amazon.awssdk.http.auth.aws.internal.DefaultAwsV4AuthScheme;
 import software.amazon.awssdk.http.auth.spi.AuthScheme;
 import software.amazon.awssdk.http.auth.spi.IdentityProviderConfiguration;
 import software.amazon.awssdk.identity.spi.AwsCredentialsIdentity;
@@ -37,8 +38,7 @@ public interface AwsV4AuthScheme extends AuthScheme<AwsCredentialsIdentity> {
      * Get a default implementation of a {@link AwsV4AuthScheme}
      */
     static AwsV4AuthScheme create() {
-        return new AwsV4AuthScheme() {
-        };
+        return DefaultAwsV4AuthScheme.create();
     }
 
     /**
@@ -53,15 +53,11 @@ public interface AwsV4AuthScheme extends AuthScheme<AwsCredentialsIdentity> {
      * Retrieve the {@link AwsCredentialsIdentity} based {@link IdentityProvider} associated with this authentication scheme.
      */
     @Override
-    default IdentityProvider<AwsCredentialsIdentity> identityProvider(IdentityProviderConfiguration providers) {
-        return providers.identityProvider(AwsCredentialsIdentity.class);
-    }
+    IdentityProvider<AwsCredentialsIdentity> identityProvider(IdentityProviderConfiguration providers);
 
     /**
      * Retrieve the {@link AwsV4HttpSigner} associated with this authentication scheme.
      */
     @Override
-    default AwsV4HttpSigner signer() {
-        return AwsV4HttpSigner.create();
-    }
+    AwsV4HttpSigner signer();
 }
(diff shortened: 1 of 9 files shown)


core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/AwsV4AuthScheme.java
```diff
@@ -53,15 +53,11 @@ default String schemeId() {
      * Retrieve the {@link AwsCredentialsIdentity} based {@link IdentityProvider} associated with this authentication scheme.
      */
     @Override
-    default IdentityProvider<AwsCredentialsIdentity> identityProvider(IdentityProviderConfiguration providers) {
-        return providers.identityProvider(AwsCredentialsIdentity.class);
-    }
+    IdentityProvider<AwsCredentialsIdentity> identityProvider(IdentityProviderConfiguration providers);
```
gosar, line 56: note (applies to other schemes too), this is redefined in this sub-interface to give a more specific javadoc than the generic one in the parent interface.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/AwsV4AuthScheme.java
```diff
@@ -53,15 +53,11 @@ default String schemeId() {
...
 
     /**
      * Retrieve the {@link AwsV4HttpSigner} associated with this authentication scheme.
      */
     @Override
-    default AwsV4HttpSigner signer() {
-        return AwsV4HttpSigner.create();
-    }
+    AwsV4HttpSigner signer();
```
gosar, line 62: note (applies to other schemes too), this is redefined in this sub-interface to define the more specific HttpSigner in the return type. And to give a more specific javadoc than the generic one in the parent interface.

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/NoAuthAuthScheme.java
```diff
@@ -34,6 +34,14 @@ static NoAuthAuthScheme create() {
         return DefaultNoAuthAuthScheme.create();
     }
 
+    /**
+     * Retrieve the scheme ID.
+     */
+    @Override
+    default String schemeId() {
+        return SCHEME_ID;
+    }
```
gosar, line 43: note the default implementation of `schemeId()` is kept in the interface as opposed to the Default*AuthScheme impl. This is the case in the other schemes already, and changed it in NoAuthAuthScheme to match others.

Q: Should there be no default implementation of `schemeId()` in the interface too?

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/DefaultAwsV4aAuthScheme.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    @Override
+    public IdentityProvider<AwsCredentialsIdentity> identityProvider(IdentityProviderConfiguration providers) {
+        return providers.identityProvider(AwsCredentialsIdentity.class);
+    }
+
+    @Override
+    public AwsV4aHttpSigner signer() {
+        return SignerSingletonHolder.INSTANCE;
```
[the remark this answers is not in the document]
gosar: I had added a note in the PR description but probably should add a code comment too.

The reason is that AwsV4aHttpSigner only has crt implementation and so requires the optional crt dependency `http-auth-aws-crt` to create an instance. So made it lazy only when `v4aScheme.signer()` is called. I guess making it lazy in the other cases is fine too, if that's what you are getting at? But thought it could be avoided to simplify - the instances itself are stateless, so not expensive.

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/NoAuthAuthScheme.java
haydenbaker: I don't see any benefit of either approach over the other, so I suppose for consistency sake, just move it to the default impl and get rid of default here and in the other auth-scheme interfaces

haydenbaker: I guess if `SCHEME_ID`is referenced outside of the module, it makes sense to be defined in this interface.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/DefaultAwsV4aAuthScheme.java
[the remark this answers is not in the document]
cenedhryn: Ok, I though that was why - just wanted to clarify.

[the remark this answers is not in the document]
gosar: 👍 added a code comment in new commit

core/http-auth/src/main/java/software/amazon/awssdk/http/auth/NoAuthAuthScheme.java
gosar: SCHEME_ID will remain defined in the interface.
Based on offline discussions, will remove the default implementation from the interface.

gosar: Removed in https://github.com/aws/aws-sdk-java-v2/pull/4416/commits/a916ca06af5f3d97b4931a57a8a070fc7e619409

cenedhryn (approved):

PR #4423, opened 2023-09-12
Move and rename IdentityProviderConfiguration

gosar:
## Motivation and Context

In surface area review, we decided we wanted to move IdentityProviderConfiguration into identity-spi.
Also, decided to rename it to IdentityProviders, as use of "Configuration" for a class in -spi package doesn't seem appropriate. sdk-core will use it for identity providers _configured_ on a client, but that's one use case. This was something we wanted to revisit from the original PR when this interface was introduced - https://github.com/aws/aws-sdk-java-v2/pull/3999#discussion_r1192804375.

## Modifications

* Move IdentityProviderConfiguration to identity-spi
* Rename IdentityProviderConfiguration to IdentityProviders
* Rename IDENTITY_PROVIDER_CONFIGURATION (in SdkClientOption and SdkInternalExecutionAttrbitutes) to IDENTITY_PROVIDERS


core/http-auth-spi/src/main/java/software/amazon/awssdk/http/auth/spi/AuthScheme.java
```diff
@@ -51,7 +52,7 @@ public interface AuthScheme<T extends Identity> {
      * Note, the returned identity provider may differ from the type of identity provider retrieved from the provided identity
      * provider configuration.
```
gosar, line 53: nit: need to remove configuration from here.

cenedhryn (approved):
Yeah, this change feels right when you see it in the code.

PR #4425, opened 2023-09-12
Fix for Cross-region calls failure due to AuthorizationHeaderMalformed errors while using clients with configured region as us-east-1

joviegas:
## Motivation and Context

- Cross region feature fails to do Cross region calls when US_EAST_1 as client region and trying to access Bucket which were created in other regions.
- Note that we already had Integ test for this , yet this issue was not reproucible in Integ since we are creating New Buckets in cross regions and testing the US_EAST_1 immediately
- This issue is reproducible only if Buckets existed for a long time more than 48 hour.

## Modifications

- Updated the logic to consider as redirect when we Get AuthorizationHeaderMalformed error.
- When we get AuthorizationHeaderMalformed we will do a HeadBucket Call and AuthorizationHeaderMalformed for HeadBucket always contains x-amz-bucket headers and thus we will get the Cross region bucket
- For any Error if we find x-amzn-bucket-region in the response then this Region will be used in doing cross region calls

## Testing

- Added Junits 
- Tested on integ by accessing an Old cross region. bucket from US_EAST_1 and making sure its redirected

## Types of changes

## License

### DIFF
--- a/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/utils/CrossRegionUtils.java
+++ b/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/utils/CrossRegionUtils.java
@@ -16,12 +16,16 @@
 package software.amazon.awssdk.services.s3.internal.crossregion.utils;
 
 
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.CompletionException;
 import java.util.function.Consumer;
 import java.util.function.Supplier;
 import software.amazon.awssdk.annotations.SdkInternalApi;
 import software.amazon.awssdk.awscore.AwsRequestOverrideConfiguration;
+import software.amazon.awssdk.awscore.exception.AwsErrorDetails;
 import software.amazon.awssdk.core.ApiName;
 import software.amazon.awssdk.endpoints.EndpointProvider;
 import software.amazon.awssdk.regions.Region;
@@ -35,6 +39,9 @@ public final class CrossRegionUtils {
     public static final int REDIRECT_STATUS_CODE = 301;
     public static final int TEMPORARY_REDIRECT_STATUS_CODE = 307;
     public static final String AMZ_BUCKET_REGION_HEADER = "x-amz-bucket-region";
+    private static final List<Integer> REDIRECT_STATUS_CODES =
+        Arrays.asList(REDIRECT_STATUS_CODE, TEMPORARY_REDIRECT_STATUS_CODE);
+    private static final List<String> REDIRECT_ERROR_CODES = Collections.singletonList("AuthorizationHeaderMalformed");
     private static final ApiName API_NAME = ApiName.builder().version("cross-region").name("hll").build();
     private static final Consumer<AwsRequestOverrideConfiguration.Builder> USER_AGENT_APPLIER = b -> b.addApiName(API_NAME);
 
@@ -55,10 +62,16 @@ public final class CrossRegionUtils {
     }
 
     private static boolean isRedirectError(S3Exception exceptionToBeChecked) {
-        int statusCode = exceptionToBeChecked.statusCode();
-        return statusCode == REDIRECT_STATUS_CODE || statusCode == TEMPORARY_REDIRECT_STATUS_CODE;
+        if (REDIRECT_STATUS_CODES.stream().anyMatch(status -> status.equals(exceptionToBeChecked.statusCode()))) {
+            return true;
+        }
+        if (getBucketRegionFromException(exceptionToBeChecked).isPresent()) {
+            return true;
+        }
+        AwsErrorDetails awsErrorDetails = exceptionToBeChecked.awsErrorDetails();
+        return awsErrorDetails != null
+               && REDIRECT_ERROR_CODES.stream().anyMatch(code -> code.equals(awsErrorDetails.errorCode()));
     }
-
 
     @SuppressWarnings("unchecked")
     public static <T extends S3Request> T requestWithDecoratedEndpointProvider(T request, Supplier<Region> regionSupplier,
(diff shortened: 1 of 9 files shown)


services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/utils/CrossRegionUtils.java
```diff
@@ -55,11 +62,17 @@ public static boolean isS3RedirectException(Throwable exception) {
     }
 
     private static boolean isRedirectError(S3Exception exceptionToBeChecked) {
-        int statusCode = exceptionToBeChecked.statusCode();
-        return statusCode == REDIRECT_STATUS_CODE || statusCode == TEMPORARY_REDIRECT_STATUS_CODE;
+        if (REDIRECT_STATUS_CODES.stream().anyMatch(status -> status.equals(exceptionToBeChecked.statusCode()))) {
```
joviegas, line 65: For readibilty and debugging using separate ifs instead of clubbing to single statement with ors (||s)

joviegas:
Working on adding a MOCK Wire test case if somehow HeadBucket fails  to respond

cenedhryn (commented):
Do you know why the response from S3 varies depending on whether the bucket is new or established?

cenedhryn (commented):
>Do we make a HeadBucket call? Won't the code do `updateCacheFromRedirectException` and then does not to call `fetchBucketRegion`?


Let's consider a scenario where a bucket has already existed in the US_WEST_2 region for a week. Now, when a user performs the following actions:

1. PutObject call ⇒ In this case, we initially receive an AuthorizationHeaderMalformed error with no 'x-amz-bucket-region'. As a result, we perform a HeadBucket call, which then returns an AuthorizationHeaderMalformed error with 'x-amz-bucket-region'. This region information is subsequently used to execute the PutObject call.
2. HeadBucket call ⇒ In this scenario, we directly receive an AuthorizationHeaderMalformed error with 'x-amz-bucket-region'. This information can be used to make a successful HeadBucket call across regions."


Thus if you see if Response header already has 'x-amz-bucket-region' we need not do another HeadBucket call , we can directly use this as CrossRegion where bucket exists.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/utils/CrossRegionUtils.java
```diff
@@ -55,11 +62,17 @@ public static boolean isS3RedirectException(Throwable exception) {
...
+            return true;
+        }
+        if (getBucketRegionFromException(exceptionToBeChecked).isPresent()) {
+            return true;
+        }
```
cenedhryn, line 70: Which use case does this represent?

The malformed header is the next logical statement

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/utils/CrossRegionUtils.java
```diff
@@ -78,13 +91,13 @@ public static <T extends S3Request> T requestWithDecoratedEndpointProvider(T req
...
-        AwsRequestOverrideConfiguration overrideConfiguration =
+        return
             request.overrideConfiguration().map(c -> c.toBuilder()
                                                       .applyMutation(USER_AGENT_APPLIER)
                                                       .build())
-                   .orElse(AwsRequestOverrideConfiguration.builder()
-                                                          .applyMutation(USER_AGENT_APPLIER)
-                                                          .build());
-        return overrideConfiguration;
+                   .orElseGet(() -> AwsRequestOverrideConfiguration.builder()
```
cenedhryn, line 98: Nice upgrade!

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/utils/CrossRegionUtils.java
joviegas: In this case, when the user directly calls "HeadBucket," we receive the "x-amz-bucket-region" in the error response. We shouldn't make another "HeadBucket" call. Instead, if we find "x-amz-bucket-region" in the S3Exception, we should use it to redirect the call to the region where the bucket actually exists.

joviegas:
> Do you know why the response from S3 varies depending on whether the bucket is new or established?

I'm collaborating with the S3 team, and this behavior is currently observed, although it's not documented anywhere. I'll inform you of the response once I receive it. Regardless of the response, the solution will remain the same.

services/s3/src/test/java/software/amazon/awssdk/services/s3/internal/crossregion/S3DecoratorRedirectTestBase.java
```diff
@@ -162,6 +162,60 @@ void requestsAreNotOverridden_when_NoBucketInRequest() throws Throwable {
...
+        verifyTheEndPointProviderOverridden(1, preCacheCaptor, CROSS_REGION.id());
+        listObjectsResponse = apiCallToService();
+        assertThat(listObjectsResponse.contents()).isEqualTo(S3_OBJECTS);
+        // We need to captor again so that we get the args used in second API Call
+        ArgumentCaptor<ListObjectsRequest> overAllPostCacheCaptor = ArgumentCaptor.forClass(ListObjectsRequest.class);
+        verifyTheApiServiceCall(3, overAllPostCacheCaptor);
+        assertThat(overAllPostCacheCaptor.getAllValues().get(0).overrideConfiguration().get().endpointProvider()).isNotPresent();
+        verifyTheEndPointProviderOverridden(1, overAllPostCacheCaptor, CROSS_REGION.id());
+        verifyTheEndPointProviderOverridden(2, overAllPostCacheCaptor, CROSS_REGION.id());
+        verifyHeadBucketServiceCall(1);
```
cenedhryn, line 209: For readability, can we group the first call statements together and have two different `ListObjectsResponse` objects such as `firstApiCallResponse` etc?

cenedhryn (commented):
Can we rename `S3DecoratorRedirectTestBase` to `S3CrossRegionRedirectTestBase` so it matches the implementing classes?

services/s3/src/test/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionAsyncClientTest.java
```diff
@@ -178,58 +238,67 @@ void crossRegionClient_CallsHeadObject_when_regionNameNotPresentInFallBackCall()
...
         assertThat(captureInterceptor.endpointProvider).isInstanceOf(BucketEndpointProvider.class);
 
     }
 
-    @Test
-    void crossRegionClient_CallsHeadObjectErrors_shouldTerminateTheAPI() {
-        mockAsyncHttpClient.stubResponses(customHttpResponse(301,  null ),
-                                         customHttpResponse(400,  null ),
-                                         successHttpResponse(), successHttpResponse());
+    @ParameterizedTest
```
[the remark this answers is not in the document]
joviegas: added labels

cenedhryn (approved):

PR #4426, opened 2023-09-12
Update AutoScalingIntegrationTest

zoewangg:
Removed integ tests that create resources and only kept a couple of DescribeXXX tests because 1) Autoscaling service does not have any customization, so I don’t think extensive test of CRUD is necessary and they will make our tests less stable. 2) it's hard to make sure the launch configuration always follow best practices

I did not remove the integ test class entirely because autoscaling service is query based and having some smoke tests for query based services is valuable.


joviegas:
Could you please mention a note on why are we removing this now ?

joviegas (approved):

PR #4437, opened 2023-09-15
Update AwsV4aHttpSigner to use RegionSet

haydenbaker:
## Motivation and Context
- V4a signers have a concept of "set of regions" that a signature is valid for, as opposed to a singular region with v4 signers
- The current code could technically support a set of regions in the form of a comma-separated string, but this could error-prone for a caller as there would be no validation of that string
- These changes update the interface to use a stronger typed `RegionSet`, which enables up-front validation on the inputs

## Modifications
- Create a `RegionSet` class, Update interface to use `RegionSet`
- Update codegen to form the correct input, and update codegen tests
- Update attribute-property adapter code to use region-set instead of region-name

## Testing
- Run new and existing tests, make sure pass

## Types of changes

## Checklist

## License

### DIFF
--- a/codegen/src/test/resources/software/amazon/awssdk/codegen/poet/auth/scheme/query-endpoint-auth-params-auth-scheme-endpoint-provider-without-allowlist.java
+++ b/codegen/src/test/resources/software/amazon/awssdk/codegen/poet/auth/scheme/query-endpoint-auth-params-auth-scheme-endpoint-provider-without-allowlist.java
@@ -3,7 +3,6 @@ package software.amazon.awssdk.services.query.auth.scheme.internal;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
-import java.util.stream.Collectors;
 import software.amazon.awssdk.annotations.Generated;
 import software.amazon.awssdk.annotations.SdkInternalApi;
 import software.amazon.awssdk.awscore.endpoints.AwsEndpointAttribute;
@@ -68,8 +67,7 @@ public final class DefaultQueryAuthSchemeProvider implements QueryAuthSchemeProv
                     SigV4aAuthScheme sigv4aAuthScheme = Validate.isInstanceOf(SigV4aAuthScheme.class, authScheme,
                                                                               "Expecting auth scheme of class SigV4AuthScheme, got instead object of class %s", authScheme.getClass()
                                                                                                                                                                           .getName());
-                    List<String> signingRegionSet = sigv4aAuthScheme.signingRegionSet();
-                    RegionSet regionSet = RegionSet.create(signingRegionSet.stream().collect(Collectors.joining(",")));
+                    RegionSet regionSet = RegionSet.create(sigv4aAuthScheme.signingRegionSet());
                     options.add(AuthSchemeOption.builder().schemeId("aws.auth#sigv4a")
                                                 .putSignerProperty(AwsV4aHttpSigner.SERVICE_SIGNING_NAME, sigv4aAuthScheme.signingName())
                                                 .putSignerProperty(AwsV4aHttpSigner.REGION_SET, regionSet)
--- a/codegen/src/test/resources/software/amazon/awssdk/codegen/poet/auth/scheme/query-endpoint-auth-params-auth-scheme-endpoint-provider-without-allowlist.java
+++ b/codegen/src/test/resources/software/amazon/awssdk/codegen/poet/auth/scheme/query-endpoint-auth-params-auth-scheme-endpoint-provider-without-allowlist.java
@@ -10,7 +10,6 @@ import software.amazon.awssdk.awscore.endpoints.AwsEndpointAttribute;
 import software.amazon.awssdk.awscore.endpoints.authscheme.EndpointAuthScheme;
 import software.amazon.awssdk.awscore.endpoints.authscheme.SigV4AuthScheme;
 import software.amazon.awssdk.awscore.endpoints.authscheme.SigV4aAuthScheme;
-import software.amazon.awssdk.core.exception.SdkClientException;
 import software.amazon.awssdk.endpoints.Endpoint;
 import software.amazon.awssdk.http.auth.aws.signer.AwsV4HttpSigner;
 import software.amazon.awssdk.http.auth.aws.signer.AwsV4aHttpSigner;
@@ -71,9 +70,6 @@ public final class DefaultQueryAuthSchemeProvider implements QueryAuthSchemeProv
                                                                                                                                                                           .getName());
                     List<String> signingRegionSet = sigv4aAuthScheme.signingRegionSet();
                     RegionSet regionSet = RegionSet.create(signingRegionSet.stream().collect(Collectors.joining(",")));
-                    if (regionSet.toString().isEmpty()) {
-                        throw SdkClientException.create("Signing region set is empty");
-                    }
                     options.add(AuthSchemeOption.builder().schemeId("aws.auth#sigv4a")
                                                 .putSignerProperty(AwsV4aHttpSigner.SERVICE_SIGNING_NAME, sigv4aAuthScheme.signingName())
                                                 .putSignerProperty(AwsV4aHttpSigner.REGION_SET, regionSet)
--- a/codegen/src/test/resources/software/amazon/awssdk/codegen/poet/auth/scheme/query-endpoint-auth-params-auth-scheme-endpoint-provider-without-allowlist.java
+++ b/codegen/src/test/resources/software/amazon/awssdk/codegen/poet/auth/scheme/query-endpoint-auth-params-auth-scheme-endpoint-provider-without-allowlist.java
@@ -3,6 +3,7 @@ package software.amazon.awssdk.services.query.auth.scheme.internal;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.stream.Collectors;
 import software.amazon.awssdk.annotations.Generated;
 import software.amazon.awssdk.annotations.SdkInternalApi;
 import software.amazon.awssdk.awscore.endpoints.AwsEndpointAttribute;
@@ -13,6 +14,7 @@ import software.amazon.awssdk.core.exception.SdkClientException;
 import software.amazon.awssdk.endpoints.Endpoint;
 import software.amazon.awssdk.http.auth.aws.signer.AwsV4HttpSigner;
 import software.amazon.awssdk.http.auth.aws.signer.AwsV4aHttpSigner;
+import software.amazon.awssdk.http.auth.aws.signer.RegionSet;
 import software.amazon.awssdk.http.auth.spi.scheme.AuthSchemeOption;
 import software.amazon.awssdk.services.query.auth.scheme.QueryAuthSchemeParams;
 import software.amazon.awssdk.services.query.auth.scheme.QueryAuthSchemeProvider;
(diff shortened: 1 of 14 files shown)


core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/AwsV4aHttpSigner.java
```diff
@@ -31,15 +32,15 @@
 @SdkPublicApi
 public interface AwsV4aHttpSigner extends AwsV4FamilyHttpSigner, HttpSigner<AwsCredentialsIdentity> {
     /**
-     * The AWS region name to be used for computing the signature. This property is required.
-     * TODO(sra-identity-and-auth): Should this be a list or rename to SIGNING_SCOPE?
+     * The AWS region-set to be used for computing the signature. This property is required.
```
haydenbaker, line 35: Do we want to have a default? Global (i.e. `*`)?

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/crt/internal/signer/DefaultAwsCrtV4aHttpSigner.java
```diff
@@ -64,9 +68,10 @@ private static V4aProperties v4aProperties(BaseSignRequest<?, ? extends AwsCrede
         Clock signingClock = request.requireProperty(SIGNING_CLOCK, Clock.systemUTC());
         Instant signingInstant = signingClock.instant();
         AwsCredentialsIdentity credentials = sanitizeCredentials(request.identity());
-        String regionName = request.requireProperty(REGION_NAME);
+        Set<RegionScope> regionSet = request.requireProperty(REGION_SET, Collections.singleton(RegionScope.GLOBAL));
```
haydenbaker, line 71: I set the default here of global for example, but if we decide no default, then this should be removed.

core/auth/src/main/java/software/amazon/awssdk/auth/signer/AwsSignerExecutionAttribute.java
```diff
@@ -270,16 +272,24 @@ private static RegionScope signingRegionScopeReadMapping(SelectedAuthScheme<?> a
...
+
+        return RegionScope.create(regionSet.stream().findFirst().get().id());
     }
 
     private static <T extends Identity> SelectedAuthScheme<?> signingRegionScopeWriteMapping(SelectedAuthScheme<T> authScheme,
                                                                                              RegionScope regionScope) {
-        String regionScopeString = regionScope == null ? null : regionScope.id();
+        Set<software.amazon.awssdk.http.auth.aws.RegionScope> regionSet =
+            regionScope != null ?
+            Collections.singleton(software.amazon.awssdk.http.auth.aws.RegionScope.create(regionScope.id())) :
```
haydenbaker, line 291: Unfortunately, `RegionScope`is already imported, so the FQN has to be used in these places.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/signer/AwsV4aHttpSigner.java
```diff
@@ -31,15 +32,15 @@
 @SdkPublicApi
 public interface AwsV4aHttpSigner extends AwsV4FamilyHttpSigner, HttpSigner<AwsCredentialsIdentity> {
     /**
-     * The AWS region name to be used for computing the signature. This property is required.
-     * TODO(sra-identity-and-auth): Should this be a list or rename to SIGNING_SCOPE?
+     * The AWS region-set to be used for computing the signature. This property is required.
      */
-    SignerProperty<String> REGION_NAME = SignerProperty.create(String.class, "SigningScope");
+    SignerProperty<Set<RegionScope>> REGION_SET = SignerProperty.create("RegionSet");
```
millems, line 37: I think RegionScope is already supposed to be a set-ish thing. Right now we only support one value (no commas), but that could change in the future as the signers evolve.

It's also weird to model RegionScope here explicitly, instead of just using String, like we do with region.

Should probably just do `SignerProperty<Set<String>>`? CRT literally just takes a `String`, but I don't understand scope enough to know if that's a good idea or not....

haydenbaker: FWIU, we wanted stronger validation than String, so we use `RegionScope`. We also couldn't tell if `RegionScope` was supposed to represent a set, since it disallows commas.

haydenbaker: I'm personally inclined to use a simple string, and the implementation can do validation upon signing, but it's up to the team. I believe the argument for not doing this is that potential errors in the input can be detected when adding the region-set value, instead of when signing happens.

haydenbaker: Concluded offline that we will re-use RegionScope in a more limited fashion, `RegionSet`, and use it instead.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/signer/RegionSet.java
```diff
@@ -0,0 +1,105 @@
...
+    private RegionSet(String regionSet) {
+        this.regionSet = Validate.paramNotBlank(regionSet, "regionSet");
+        validateFormat(regionSet);
+    }
+
+    /**
+     * Gets the string representation of this RegionSet.
+     */
+    @Override
+    public String toString() {
```
gosar, line 63: I think we should have a separate method, like `value()` to retrieve the string value, which is used in business logic. Relying to toString() is fragile as someone looking at this class may not realize changing the implementation can be breaking, say we add some other property to this class, one may think of updating toString(), and probably not easy to find all places that may be calling toString() to understand impact. Unlikely that attempting to change the impl here would be missed as it would be caught in integ testing, but generally shouldn't rely on toString().

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/auth/scheme/query-endpoint-auth-params-auth-scheme-endpoint-provider-without-allowlist.java
```diff
@@ -68,15 +70,13 @@ public List<AuthSchemeOption> resolveAuthScheme(QueryAuthSchemeParams params) {
                                                                               "Expecting auth scheme of class SigV4AuthScheme, got instead object of class %s", authScheme.getClass()
                                                                                                                                                                           .getName());
                     List<String> signingRegionSet = sigv4aAuthScheme.signingRegionSet();
-                    if (signingRegionSet.size() == 0) {
+                    RegionSet regionSet = RegionSet.create(signingRegionSet.stream().collect(Collectors.joining(",")));
+                    if (regionSet.toString().isEmpty()) {
```
gosar, line 74: odd too see isEmpty check here, I assume we wouldn't allow creating RegionSet with empty string.

gosar: turns out you do not allow it in the constructor, so this can be removed.

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/rules/endpoint-resolve-interceptor.java
```diff
@@ -147,7 +149,9 @@ private <T extends Identity> SelectedAuthScheme<T> authSchemeWithEndpointSignerP
                     option.putSignerProperty(AwsV4aHttpSigner.DOUBLE_URL_ENCODE, !v4aAuthScheme.disableDoubleEncoding());
                 }
                 if (v4aAuthScheme.signingRegionSet() != null) {
-                    option.putSignerProperty(AwsV4aHttpSigner.REGION_NAME, String.join(",", v4aAuthScheme.signingRegionSet()));
+                    RegionSet regionSet = RegionSet.create(v4aAuthScheme.signingRegionSet().stream()
```
gosar, line 152: I think we could add method to SigV4aAuthScheme that returns RegionSet directly

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/signer/RegionSet.java
gosar: Looks like RegionScope used `id()` for this, so probably just use the same here.

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/rules/endpoint-resolve-interceptor.java
haydenbaker: Probably.

test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/endpointproviders/EndpointInterceptorTests.java
```diff
@@ -282,7 +283,8 @@ public void sync_endpointProviderReturnsSignerProperties_overridesV4AAuthSchemeR
             .putAuthScheme(authScheme("aws.auth#sigv4a", signer))
             .authSchemeProvider(p -> singletonList(AuthSchemeOption.builder()
                                                                    .schemeId("aws.auth#sigv4a")
-                                                                   .putSignerProperty(AwsV4aHttpSigner.REGION_NAME, "X")
+                                                                   .putSignerProperty(AwsV4aHttpSigner.REGION_SET,
+                                                                                      RegionSet.create("us-east-1"))
```
cenedhryn, line 287: Nit: seems it could be one line

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/rules/endpoint-resolve-interceptor.java
cenedhryn: Yes, I saw 
~~~
List<String> signingRegionSet = sigv4aAuthScheme.signingRegionSet();
RegionSet regionSet = RegionSet.create(signingRegionSet.stream().collect(Collectors.joining(",")));
~~~
and was wondering if there was a reason the auth scheme had to return a list of string.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/AwsV4aHttpSigner.java
gosar: Looks like having it required works for our use from endpoints based sigv4a. I don't feel strongly about either option. Starting required is a 2-way door and we can make '*' default later too.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/signer/RegionSet.java
```diff
@@ -0,0 +1,102 @@
...
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import software.amazon.awssdk.annotations.SdkPublicApi;
+import software.amazon.awssdk.utils.Validate;
+
+/**
+ * This class represents the concept of a set of regions.
+ * <p>
+ * A Re
```
cenedhryn, line 26: (javadoc incomplete)

What are the reasons we represent the values as a comma separated string instead of a `Set`? We discussed it would not be a plain String externally, but I thought we'd model the values separately. I may have missed something.

haydenbaker: It's not a plain string, but a wrapper of a string with validation upon creation, just as `RegionScope` is. `RegionSet` is just a more constrained version of `RegionScope`

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/rules/endpoint-resolve-interceptor.java
haydenbaker: I think we can add this later if we end up using it in more places. Right now, I wouldn't want to add anything to the auth scheme, since it's protected.

cenedhryn (approved):

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/signer/RegionSet.java
```diff
@@ -0,0 +1,108 @@
...
+import java.util.HashSet;
+import java.util.Set;
+import software.amazon.awssdk.annotations.SdkPublicApi;
+import software.amazon.awssdk.utils.Validate;
+
+/**
+ * This class represents the concept of a set of regions.
+ * <p>
+ * A region-set can contain one or more comma-separated AWS regions, or a single wildcard to represent all regions ("global").
+ * Whitespace is trimmed from entries of the set.
```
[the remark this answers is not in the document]
haydenbaker: It is now.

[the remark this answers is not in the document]
gosar: My comment was about whitespace trimming. is that happening now?

Issue #3839, opened 2023-03-15
`S3TransferManager`: The `TransferListener.transferComplete()` is not invoked when `AsyncRequestBody.fromFile()` is used


artembilan:
### Describe the bug

The `FileAsyncRequestBody` does not call `subscriber.onComplete()` because it is cancelled from the `NettyRequestExecutor.StreamedRequest` via this code:
```
                        if (!shouldContinuePublishing()) {
                            done = true;
                            subscription.cancel();
                            subscriber.onComplete();
                        }
```

### Expected Behavior

The `FileAsyncRequestBody` is properly completed and therefore a `TransferListener.transferComplete()` is signaled.

### Current Behavior

The `TransferListener.transferComplete()` is not called when we upload the file.

Works for plain `byte[]` though.

### Reproduction Steps

```
import java.io.File;
import java.io.FileOutputStream;
import java.io.IOException;
import java.io.OutputStream;
import java.nio.file.Path;
import java.util.concurrent.CountDownLatch;
import java.util.concurrent.TimeUnit;

import org.junit.jupiter.api.BeforeAll;
import org.junit.jupiter.api.Test;
import org.junit.jupiter.api.io.TempDir;
import org.testcontainers.containers.localstack.LocalStackContainer;
import org.testcontainers.utility.DockerImageName;
import software.amazon.awssdk.auth.credentials.AwsBasicCredentials;
import software.amazon.awssdk.auth.credentials.StaticCredentialsProvider;
import software.amazon.awssdk.core.ResponseBytes;
import software.amazon.awssdk.core.async.AsyncRequestBody;
import software.amazon.awssdk.core.async.AsyncResponseTransformer;
import software.amazon.awssdk.regions.Region;
import software.amazon.awssdk.services.s3.S3AsyncClient;
import software.amazon.awssdk.services.s3.model.GetObjectResponse;
import software.amazon.awssdk.transfer.s3.S3TransferManager;
import software.amazon.awssdk.transfer.s3.model.UploadRequest;
import software.amazon.awssdk.transfer.s3.progress.LoggingTransferListener;
import software.amazon.awssdk.transfer.s3.progress.TransferListener;

import static org.assertj.core.api.Assertions.assertThat;

public class S3TransferFileWithProgressTests {

	static final LocalStackContainer LOCAL_STACK_CONTAINER =
			new LocalStackContainer(
					DockerImageName.parse("localstack/localstack:1.4.0"))
[...]

PR #4440, opened 2023-09-15
Fixing Java S3 Client Transfer Manager issue with transferComplete() not called for AsyncRequestBody.fromFile

joviegas:
Fixing Java S3 Client Transfer Manager issue with transferComplete() not called for AsyncRequestBody.fromFile

## Motivation and Context

- Fix for issue https://github.com/aws/aws-sdk-java-v2/issues/3839
-
## Root Cause
- [NettyRequestExecutor](https://github.com/aws/aws-sdk-java-v2/blob/df4753216daeb119408f7378ec68c3cb1d0824d7/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/NettyRequestExecutor.java#L496) prematurely cancels the subscription when the last chunk is sent.
- The logic of `shouldContinuePublishing` after [DoneCheck-OnNext](https://github.com/aws/aws-sdk-java-v2/blob/df4753216daeb119408f7378ec68c3cb1d0824d7/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/NettyRequestExecutor.java#L442) checks `<` instead of `<=` in `shouldContinuePublishing()`.

## Modifications

- The actual fix would be to move `shouldContinuePublishing` before the `isDone` check and then update `<` to `<=` in `shouldContinuePublishing()`.
- However, we had an internal discussion, and the outcome was not to modify this but to update the TransferManager Code to ensure that we send the `transferComplete` once all the bytes are received.
- Added a `done` flag to ensure that `transferComplete` is not called more than once.

## Testing

- Added JUnit tests.
- Included integration tests for the Java client since this issue is primarily observed in the Java S3 client Transfer Manager.

## Types of changes

## License

### DIFF
--- a/services-custom/s3-transfer-manager/src/it/java/software/amazon/awssdk/transfer/s3/S3TransferManagerUploadIntegrationTest.java
+++ b/services-custom/s3-transfer-manager/src/it/java/software/amazon/awssdk/transfer/s3/S3TransferManagerUploadIntegrationTest.java
@@ -111,7 +111,7 @@ public class S3TransferManagerUploadIntegrationTest extends S3IntegrationTestBas
 
     @ParameterizedTest
     @MethodSource("transferManagers")
-    void upload_asyncRequestBody_SentCorrectly(S3TransferManager transferManager) throws IOException {
+    void upload_asyncRequestBodyFromString_SentCorrectly(S3TransferManager transferManager) throws IOException {
         String content = UUID.randomUUID().toString();
         CaptureTransferListener transferListener = new CaptureTransferListener();
 
--- a/services-custom/s3-transfer-manager/src/it/java/software/amazon/awssdk/transfer/s3/S3TransferManagerUploadIntegrationTest.java
+++ b/services-custom/s3-transfer-manager/src/it/java/software/amazon/awssdk/transfer/s3/S3TransferManagerUploadIntegrationTest.java
@@ -16,7 +16,6 @@
 package software.amazon.awssdk.transfer.s3;
 
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
 import static software.amazon.awssdk.testutils.service.S3BucketUtils.temporaryBucketName;
 
 import java.io.IOException;
@@ -26,11 +25,15 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.UUID;
 import java.util.concurrent.CancellationException;
+import java.util.stream.Stream;
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.BeforeAll;
-import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
 import software.amazon.awssdk.core.ResponseInputStream;
 import software.amazon.awssdk.core.async.AsyncRequestBody;
+import software.amazon.awssdk.core.internal.async.FileAsyncRequestBody;
 import software.amazon.awssdk.core.sync.ResponseTransformer;
 import software.amazon.awssdk.services.s3.model.ChecksumAlgorithm;
 import software.amazon.awssdk.services.s3.model.GetObjectResponse;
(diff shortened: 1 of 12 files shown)


services-custom/s3-transfer-manager/src/test/java/software/amazon/awssdk/transfer/s3/internal/TransferProgressUpdaterTest.java
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    @BeforeEach
+    void initiate() {
+        captureTransferListener = new CaptureTransferListener();
+    }
+
+    @ParameterizedTest
+    @MethodSource("contentLength")
+    void transferCompleteBasedOnBytesRatioWhenCompleted(Long adjustedContentLength) throws Exception {
```
cenedhryn, line 99: How does this test prove that the ratio 1 will complete the future? It takes content lengths that are both smaller and larger than the 1.0 ratio, and everything gets completed. I don't see where the difference is compared to before the change.

services-custom/s3-transfer-manager/src/it/java/software/amazon/awssdk/transfer/s3/S3TransferManagerUploadIntegrationTest.java
```diff
@@ -92,25 +102,26 @@ void upload_file_SentCorrectly() throws IOException {
...
         assertThat(transferListener.getRatioTransferredList().contains(0.0));
         assertThat(transferListener.getRatioTransferredList().contains(100.0));
         assertThat(transferListener.getExceptionCaught()).isNull();
     }
 
-    @Test
-    void upload_asyncRequestBody_SentCorrectly() throws IOException {
+    @ParameterizedTest
+    @MethodSource("transferManagers")
+    void upload_asyncRequestBody_SentCorrectly(S3TransferManager transferManager) throws IOException {
```
cenedhryn, line 114: Can we rename this method to `upload_asyncRequestBodyFromString_SentCorrectly` to be consistent with the next?

services-custom/s3-transfer-manager/src/test/java/software/amazon/awssdk/transfer/s3/internal/TransferProgressUpdaterTest.java
cenedhryn: There is a lot to unpack in this test case for the code reader not familiar with the code or the test. I suggest something like the following to improve readability:

~~~
    private static Stream<Arguments> contentLength() {
        return Stream.of(
           Arguments.of(OBJ_SIZE, "Total bytes equals transferred, future complete through subscriberOnNext()"),
           Arguments.of(OBJ_SIZE / 2, "Total bytes less than transferred, future complete through subscriberOnNext()"),
           Arguments.of(OBJ_SIZE * 2, "Total bytes more than transferred, future complete through subscriberOnComplete()"));
    }

@ParameterizedTest(name = "{index} - {1}, total bytes = {0}")
    @MethodSource("contentLength")
    void registerCompletion_differentTransferredByteRatios_alwaysCompletesOnce(Long givenContentLength, String description)
~~~

Which gives a test output of 
~~~
registerCompletion_differentTransferredByteRatios_alwaysCompletesOnce(Long, String)
1 - Total bytes equals transferred, future complete through subscriberOnNext(), total bytes = 16777216
2 - Total bytes less than transferred, future complete through subscriberOnNext(), total bytes = 8388608
3 - Total bytes more than transferred, future complete through subscriberOnComplete(), total bytes = 33554432
~~~
(it may not even be meaningful to output the exact bytes)

- The test name indicates that the complete once is the important metric for the test
- The test case descriptions shows that there is a difference between the two cases

services-custom/s3-transfer-manager/src/it/java/software/amazon/awssdk/transfer/s3/S3TransferManagerUploadIntegrationTest.java
joviegas: Done

services-custom/s3-transfer-manager/src/test/java/software/amazon/awssdk/transfer/s3/internal/TransferProgressUpdaterTest.java
joviegas: Thanks

services-custom/s3-transfer-manager/src/test/java/software/amazon/awssdk/transfer/s3/internal/TransferProgressUpdaterTest.java
```diff
@@ -0,0 +1,186 @@
...
+        TransferObjectRequest transferRequest = Mockito.mock(TransferObjectRequest.class);
+        CaptureTransferListener mockListener = Mockito.mock(CaptureTransferListener.class);
+        when(transferRequest.transferListeners()).thenReturn(Arrays.asList(LoggingTransferListener.create(), mockListener,
+                                                                           captureTransferListener));
+
+        sourceFile = new RandomTempFile(OBJ_SIZE);
+        AsyncRequestBody requestBody = AsyncRequestBody.fromFile(sourceFile);
+        TransferProgressUpdater transferProgressUpdater = new TransferProgressUpdater(transferRequest, adjustedContentLength);
```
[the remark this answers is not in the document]
joviegas: adjustedContentLength as in Modified Content length for testing

cenedhryn (commented):
Can we add a comment to the  NettyRequestExecutor.java `shouldContinuePublishing()` to indicate that the check can cause issues and what the mitigation is?

joviegas:
> Can we add a comment to the NettyRequestExecutor.java `shouldContinuePublishing()` to indicate that the check can cause issues and what the mitigation is?

I doubt a comment would help much. We should conduct a Failure Mode Analysis to see what happens if we make the suggested fixes. Just adding a comment might make it seem like we've fully evaluated NettyRequestExecutor's behavior with no room for improvement

cenedhryn (approved):
A deep dive would be ideal, but likely not happening at this point in time. My concern is that after some time the discussion in the PR will be forgotten and we'll lose context and time figuring out the same thing again, for a different issue and that a comment would let us connect the two parts of the code. 
Personally I don't care how people would interpret a comment (if they even saw it). As developers we make trade-off decisions every day, and it's not a bad thing to be prudent. That being said, will leave it up to you.

Issue #3744, opened 2023-02-03
Enhanced Client doesn't return consumed capacity in all operations


Sathyvs:
### Describe the feature

Enhanced Client has only Update and Put operation with Response that returns Consumer capacity, there is nothing for read or query operations, how to get the consumed capacity for every operation ?
Also is there a way to configure cloudwatch metrics and have the consumed capacity directly published to cloud watch for all the operations ?

### Use Case

This data can really help to understand where the dynamo db is not properly used and improve the performance and also make possible data model changes to effectively use the database and increase the application performance

### Proposed Solution

Every operation in the high level api has methods to return the consumed capacity units
Ability to publish the same as a metric to cloud watch automatically for every DB operation

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### AWS Java SDK version used

2.19.30

### JDK version used

11

### Operating System and version

aws lambda x86_64

debora-ito:
Hi @Sathyvs we've added to the backlog the task to return `consumedCapacity` on the high-level query and scan operations. The solution will probably be the same of putItem and updateItem, we'll introduce new methods like scanWithResponse.

Regarding publishing consumedCapacity to CloudWatch metrics, check the CloudWatchMetricPublisher javadoc to see how you can publish custom metrics -
https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/metrics/publishers/cloudwatch/CloudWatchMetricPublisher.html

Sathyvs:
Thanks for the reply Debora, do we know when it will be available ?

debora-ito:
We don't have a timeline to share.

debora-ito:
Enhanced Client `GetItem` (since 2.20.146)  and `Query` (since 2.20.153) now support `ConsumedCapacity`.

Will mark this issue to auto close soon. Let us know if you have any further questions.

PR #2707, opened 2021-09-10
Add file overload to DownloadRequest and UploadRequest

zoewangg:
## Description
Add file overload to DownloadRequest and UploadRequest to be consistent with `FileRequetBody` which also accepts both `File` and `Path`

## Testing
Added test

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

## Checklist

- [x] I have read the **CONTRIBUTING** document
- [x] Local run of `mvn install` succeeds
- [x] My code follows the code style of this project
- [x] My change requires a change to the Javadoc documentation
- [x] I have updated the Javadoc documentation accordingly
- [x] I have read the **README** document
- [x] I have added tests to cover my changes
- [x] All new and existing tests passed
- [x] A short description of the change has been added to the **CHANGELOG**

## License

- [x] I confirm that this pull request can be released under the Apache 2 license

PR #4444, opened 2023-09-18
Adds count, scanned count and consumed capacity to DDB Enhanced Page

cenedhryn:
## Motivation and Context
This PR addresses the issue of missing metadata fields in the response of DDB Enhanced Scan and Query operations, as described in 
- #3744 *Enhanced Client doesn't return consumed capacity in all operations* and
- [#2707](https://github.com/aws/aws-sdk-java/issues/2707) for Java v1

In addition, this PR adds `Count` and `ScannedCount` to the response to aid users who want to understand the results of their queries. 

Scan and Query are paginated operations where each page is represented by a backend call to DDB. The consumed capacity and count metric metadata is specific per page; therefore this metadata is exposed to the user per page. The new response values are available on both table and index operations.

## Modifications
- Adds the parameter to request returning consumed capacity to the `EnhancedScanRequest` and `EnhancedQueryRequest`
- Adds `ConsumedCapacity` to the `Page` class
- Adds `Count` and `ScannedCount` to the `Page` class
- Refactors the existing integration tests into fewer classes that use less resources and let maintainers scan through the tests quicker. 

## Testing
Functional testing, integration testing

## Types of changes

## Checklist

## License


services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/model/Page.java
```diff
@@ -109,4 +141,57 @@ public String toString() {
                        .add("items", items)
                        .build();
     }
+
+    public Integer count() {
```
zoewangg, line 145: Missing Javadoc, same for other methods below.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/model/Page.java
```diff
@@ -33,30 +36,47 @@
 public final class Page<T> {
     private final List<T> items;
     private final Map<String, AttributeValue> lastEvaluatedKey;
+    private final Integer count;
+    private final Integer scannedCount;
+    private final ConsumedCapacity consumedCapacity;
```
zoewangg, line 41: I guess we can't just include a service response here since it could be QueryResponse and ScanResponse. :(

zoewangg (approved):

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/model/Page.java
cenedhryn: I went through the code to check for javadocs, missed these. Well spotted.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/model/Page.java
cenedhryn: If we really wanted to, we could make it a metadata field, then we'd get
`page.responseMetadata().count()` instead of
`page.count()`. 

I find the current version a more natural, but if DDB would add many more metadata fields we'd be glad we had a containing object. WDYT?

zoewangg: I was thinking about including the whole response`scanResponse`, so `page.sdkResponse().count()`, but that doesn't seem to be possible.

PR #4460, opened 2023-09-21
Fixed an issue in InputStreamSubscriber that could cause NPE to be thrown when close and o…

### LINKED ISSUE #4081: NPE in software.amazon.awssdk.utils.async.InputStreamSubscriber
### Describe the bug

Occasionally seeing the following when downloading from S3
```
java.lang.NullPointerException
	at software.amazon.awssdk.utils.async.InputStreamSubscriber.close(InputStreamSubscriber.java:123)
	at software.amazon.awssdk.core.io.SdkFilterInputStream.close(SdkFilterInputStream.java:83)
```

### Expected Behavior

`InputStreamSubscriber.close` should not raise a NPE.

### Current Behavior

`java.lang.NullPointerException` is being thrown occasionally when closing inout stream.

### Reproduction Steps

It is hard to repro, as this only happens occasionally.

### Possible Solution

I think there might be a race condition causing a variable to be used before it is initialised. Adding a snapshot of the relevant code from [InputStreamSubscriber](https://github.com/aws/aws-sdk-java-v2/blob/2.20.59/utils/src/main/java/software/amazon/awssdk/utils/async/InputStreamSubscriber.java) below:
```java
 38 public final class InputStreamSubscriber extends InputStream implements Subscriber<ByteBuffer>, SdkAutoCloseable {
...
 44     private final AtomicReference<State> inputStreamState = new AtomicReference<>(State.UNINITIALIZED);
...
 48     private Subscription subscription;
...
 57     public void onSubscribe(Subscription s) {
 58         if (!inputStreamState.compareAndSet(State.UNINITIALIZED, State.READABLE)) {
 59             close();
 60             return;
 61         }
 62
 63         this.subscription = new CancelWatcher(s);
 64         delegate.onSubscribe(subscription);
 65     }
...
118     public void close() {
119         if (inputStreamState.compareAndSet(State.UNINITIALIZED, State.CLOSED)) {
120             delegate.onSubscribe(new NoOpSubscription());
121             delegate.onError(new CancellationException());
122         } else if (inputStreamState.compareAndSet(State.READABLE, State.CLOSED)) {
123             subscription.cancel();
124             onError(new CancellationException());
125         }
126     }
```
Considering the case where two threads access the class, one invoking the `onSubscribe` and the other `close`, it seems a race condition might cause the NPE when
  - `Thread1` invokes `onSubscribe` and executes the first statement (line 58), setting the state to `READABLE`
  - `Thread2` invokes `close` after `Thread1` executed line 58, but before it gets to execute line 63
      - Since the state is now `READABLE`, `Thread2` attempts to call `subscribe.cancel()` on an unitialised `subcriber`.

_No response_

### AWS Java SDK version used

2.20.59

### JDK version used

11

MacOS 13

zoewangg:
…nSubscribed get invoked concurrently.

## Motivation and Context
Fix #4081 The root cause is a race condition that onSubscribe and close are invoked concurrently, the subscription could be null, causing NPE to be thrown from close method.

## Modifications
Add a lock in onSubscribe and close

## Testing
Tested it locally and verified the test no longer failed with this change. This test is not checked in because it cannot reproduce this issue reliably. 
```
    @Test
    public void closeOnSubscribe_invokeAtSameTime_shouldNotFail() {
        int count = 0;
        for (int i = 0; i < 10000; i++) {
            CompletableFuture<Void> future = CompletableFuture.runAsync(() -> subscriber.close());
            //publisher.subscribe(subscriber);
            subscriber.onSubscribe(new Subscription() {
                @Override
                public void request(long n) {

                }

                @Override
                public void cancel() {

                }
            });
            future.join();
        }

    }
```

## Types of changes

## Checklist

## License

### DIFF
--- a/utils/src/main/java/software/amazon/awssdk/utils/async/InputStreamSubscriber.java
+++ b/utils/src/main/java/software/amazon/awssdk/utils/async/InputStreamSubscriber.java
@@ -45,6 +45,8 @@ public final class InputStreamSubscriber extends InputStream implements Subscrib
     private final AtomicBoolean drainingCallQueue = new AtomicBoolean(false);
     private final Queue<QueueEntry> callQueue = new ConcurrentLinkedQueue<>();
 
+    private final Object subscribeLock = new Object();
+
     private Subscription subscription;
 
     private boolean done = false;
@@ -55,13 +57,15 @@ public final class InputStreamSubscriber extends InputStream implements Subscrib
 
     @Override
     public void onSubscribe(Subscription s) {
-        if (!inputStreamState.compareAndSet(State.UNINITIALIZED, State.READABLE)) {
-            close();
-            return;
-        }
+        synchronized (subscribeLock) {
+            if (!inputStreamState.compareAndSet(State.UNINITIALIZED, State.READABLE)) {
+                close();
+                return;
+            }
 
-        this.subscription = new CancelWatcher(s);
-        delegate.onSubscribe(subscription);
+            this.subscription = new CancelWatcher(s);
+            delegate.onSubscribe(subscription);
+        }
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


utils/src/main/java/software/amazon/awssdk/utils/async/InputStreamSubscriber.java
```diff
@@ -120,12 +124,14 @@ public int read(byte[] bytes, int off, int len) {
...
-            delegate.onError(new CancellationException());
-        } else if (inputStreamState.compareAndSet(State.READABLE, State.CLOSED)) {
-            subscription.cancel();
-            onError(new CancellationException());
+        synchronized (subscribeLock) {
+            if (inputStreamState.compareAndSet(State.UNINITIALIZED, State.CLOSED)) {
+                delegate.onSubscribe(new NoOpSubscription());
+                delegate.onError(new CancellationException());
+            } else if (inputStreamState.compareAndSet(State.READABLE, State.CLOSED)) {
+                subscription.cancel();
```
L-Applin, line 132: If `close()` is called before `onSubscribe(Subscription s)` it seems that the NPE would still occurs, should that be a concern?

zoewangg: I don't think it would since inputStreamState is UNINITIALIZED and it will execute the first if block

L-Applin (approved):

PR #4438, opened 2023-09-15
Allow ServiceClientConfiguration.Builder to read from SdkClientConfiguration.Builder

sugmanue:
## Motivation and Context

This change allows the service generated instance of `ServiceClientConfiguration.Builder` to read/write its values directly from an `SdkClientConfiguration.Builder` by mapping the required option to the getters and setters. 

Also this change adds two new properties to the `ServiceClientConfiguration` class to support the plugins use case:

* `credentialsProvider` of `IdentityProvider<? extends AwsCredentialsIdentity>`
* `authSchemeProvider` of `<Service>AuthSchemeProvider` (the service specific auth scheme provider)

This change is part of the SRA plugins for Identity and Auth.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)

## Checklist

- [ ] I have read the [CONTRIBUTING](https://github.com/aws/aws-sdk-java-v2/blob/master/CONTRIBUTING.md) document
- [ ] Local run of `mvn install` succeeds
- [ ] My code follows the code style of this project
- [ ] My change requires a change to the Javadoc documentation
- [ ] I have updated the Javadoc documentation accordingly
- [ ] I have added tests to cover my changes
- [ ] All new and existing tests passed
- [ ] I have added a changelog entry. Adding a new entry must be accomplished by running the `scripts/new-change` script and following the instructions. Commit the new file created by the script in `.changes/next-release` with your changes.
- [ ] My change is to implement 1.11 parity feature and I have updated [LaunchChangelog](https://github.com/aws/aws-sdk-java-v2/blob/master/docs/LaunchChangelog.md)

## License

- [ ] I confirm that this pull request can be released under the Apache 2 license

PR #4473, opened 2023-09-24
Fix isClientSignerOverridden to check actual value

gosar:
## Motivation and Context

isSignerOverridden is used in AwsExecutionContextBuilder. 

The SIGNER_OVERRIDDEN ExecutionAttribute is set here
https://github.com/aws/aws-sdk-java-v2/blob/32390a25048bdae15b4cac319af8805759dd1aa6/core/aws-core/src/main/java/software/amazon/awssdk/awscore/internal/AwsExecutionContextBuilder.java#L110

And isSignerOverridden is called here
https://github.com/aws/aws-sdk-java-v2/blob/32390a25048bdae15b4cac319af8805759dd1aa6/core/aws-core/src/main/java/software/amazon/awssdk/awscore/internal/AwsExecutionContextBuilder.java#L156
which is called here
https://github.com/aws/aws-sdk-java-v2/blob/32390a25048bdae15b4cac319af8805759dd1aa6/core/aws-core/src/main/java/software/amazon/awssdk/awscore/internal/AwsExecutionContextBuilder.java#L129-L136

So when the client is configured if SIGNER_OVERRIDDEN is non-null, loadSigner would have been true. In master branch,  SIGNER_OVERRIDDEN was always null or true. In the SRA branch now, it can be null or true or false. See [comment in]( https://github.com/aws/aws-sdk-java-v2/pull/4438).

In SRA branch now, Signer is always null, so even if loadSigner is called, the ExecutionContext.signer [here](https://github.com/aws/aws-sdk-java-v2/blob/32390a25048bdae15b4cac319af8805759dd1aa6/core/aws-core/src/main/java/software/amazon/awssdk/awscore/internal/AwsExecutionContextBuilder.java#L147) will be null, so the SRA HttpSigner gets used, which is all fine. However, the call to `addCredentialsToExecutionAttributes` means identity resolution happens there a second time (1st time happened in AuthSchemeInterceptor, which happens before this point). More than the issue of happening twice (which adds a small risk of that call failing), because of the derived attribute logic from https://github.com/aws/aws-sdk-java-v2/pull/4396, addCredentialsToExecutionAttributes will result in the identity to be updated in SelectedAuthScheme, which would break identity property based auth resolution for an upcoming service.

## Modifications

Check that the value of SIGNER_OVERRIDDEN is present instead of just isPresent.

## Testing

I realized this issue while stepping through the debugger for some other testing I was doing and noticed going into loadSigner branch unexpectedly. Verified after this change, it doesn't go into that branch.

## Types of changes

## Checklist

## License

### DIFF
--- a/core/aws-core/src/main/java/software/amazon/awssdk/awscore/util/SignerOverrideUtils.java
+++ b/core/aws-core/src/main/java/software/amazon/awssdk/awscore/util/SignerOverrideUtils.java
@@ -66,12 +66,14 @@ public final class SignerOverrideUtils {
      * @deprecated No longer used by modern clients after migration to reference architecture
      */
     @Deprecated
+    // TODO(sra-identity-and-auth): These used to be used by EndpointsAuthSchemeInterceptor, which has now been removed, but
+    //  this method is still used  from AwsExecutionContextBuilder. Should @Deprecated be removed?
     public static boolean isSignerOverridden(SdkRequest request, ExecutionAttributes executionAttributes) {
-        Optional<Boolean> isClientSignerOverridden = Optional.ofNullable(
-            executionAttributes.getAttribute(SdkExecutionAttribute.SIGNER_OVERRIDDEN));
+        boolean isClientSignerOverridden =
+            Boolean.TRUE.equals(executionAttributes.getAttribute(SdkExecutionAttribute.SIGNER_OVERRIDDEN));
         Optional<Signer> requestSigner = request.overrideConfiguration()
                                                 .flatMap(RequestOverrideConfiguration::signer);
-        return isClientSignerOverridden.isPresent() || requestSigner.isPresent();
+        return isClientSignerOverridden || requestSigner.isPresent();
     }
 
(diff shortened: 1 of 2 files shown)


core/aws-core/src/main/java/software/amazon/awssdk/awscore/util/SignerOverrideUtils.java
```diff
@@ -66,12 +66,14 @@ public static SdkRequest overrideSignerIfNotOverridden(SdkRequest request,
      * @deprecated No longer used by modern clients after migration to reference architecture
      */
     @Deprecated
+    // TODO(sra-identity-and-auth): These used to be used by EndpointsAuthSchemeInterceptor, which has now been removed, but
+    //  this method is still used  from AwsExecutionContextBuilder. Should @Deprecated be removed?
```
millems, line 70: Yes, if we're still using it, let's not deprecate it.

millems (approved):

PR #4476, opened 2023-09-25
Update RegionSet construction, add more tests

haydenbaker:
## Motivation and Context
- Address missed comments on #4437 

## Modifications
- Add null and emptiness checks when constructing RegionSet
- Add more tests and assertions for RegionSet
- Remove signer default for RegionSet property for CRT signers

## Testing
- Run new and existing tests, make sure pass

## Types of changes

## Checklist

## License

### DIFF
--- a/core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/signer/RegionSet.java
+++ b/core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/signer/RegionSet.java
@@ -19,7 +19,10 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
+import java.util.Objects;
 import java.util.Set;
+import java.util.stream.Collectors;
+import software.amazon.awssdk.annotations.Immutable;
 import software.amazon.awssdk.annotations.SdkPublicApi;
 import software.amazon.awssdk.utils.Validate;
 
@@ -35,11 +38,14 @@ import software.amazon.awssdk.utils.Validate;
  *     <li>'eu-west-1' - Represents a single region, eu-west-1</li>
  *     <li>'us-west-2,us-east-1' - Represents 2 regions, us-west-2 and us-east-1</li>
  * </ul>
- * <p>
  */
 @SdkPublicApi
+@Immutable
 public final class RegionSet {
 
+    /**
+     * The "Global" region, which is represented with a single wildcard character: "*".
+     */
     public static final RegionSet GLOBAL;
 
     static {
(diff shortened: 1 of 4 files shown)


core/http-auth-aws/src/test/java/software/amazon/awssdk/http/auth/aws/RegionSetTest.java
```diff
@@ -24,32 +29,100 @@
...
+        assertIterableEquals(stringCase.asSet, regionSet.asSet());
     }
 
     @ParameterizedTest
-    @MethodSource("collectionInputs")
-    public void create_withCollectionInput_succeeds(Collection<String> input) {
-        RegionSet.create(input);
+    @MethodSource("stringFailures")
+    public void create_withInvalidStringInput_throws(String input) {
+        assertThrows(Exception.class, () -> RegionSet.create(input));
```
gosar, line 96: if this is Exception.class to account for NPE v/s IllegalArgument, might be better to do the `null` case separately than the other cases, so the specific exception can be asserted. at the minimum, make it RuntimeException.class.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/signer/RegionSet.java
```diff
@@ -84,7 +90,10 @@ public static RegionSet create(String value) {
      * @param regions A collection of regions.
      */
     public static RegionSet create(Collection<String> regions) {
-        return new RegionSet(regions);
+        Validate.notEmpty(regions, "regions");
+        return new RegionSet(
+            regions.stream().map(s -> Validate.notEmpty(s, "region").trim()).collect(Collectors.toList())
```
gosar, line 95: The way you have trimming logic, `create("  ")` will throw, but `create(Arrays.asList("  ")`, will not throw, which seems not good. Your tests miss this, because collectionFailList has case for `Arrays.asList(" ", ""),` but not just `Arrays.asList(" "),`.

Overall, when I commented https://github.com/aws/aws-sdk-java-v2/pull/4437#discussion_r1334927535 about adding cases for this, I wasn't thinking you'll make a behavior change. I think here with trimming and notEmpty it's starting to get into "validation" territory that we were avoiding. I think failing on null String/Collection seems fine because you didn't give us a value, but if there is a value, just accepting it as-is is what I had imagined.

haydenbaker: Yes, fair. Though, I meant to use `notBlank`, which checks whitespace and not length. I think we should still validate the very base case of what is a valid string to pass. Checking for blankness should be fine. We're not validating the pattern or whether it's a valid region, simply validating that there is actual content.

gosar (approved):

PR #4448, opened 2023-09-19
Add namespacing to Signer and Identity properties

haydenbaker:
## Motivation and Context
- To ensure uniqueness of properties, we can add the concept on namespacing when properties are defined

## Modifications
- Add namespace to property definitions, which is derived from a class' name when `create()` is called
- Updates all places where properties are defined and tests

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/aws/aws-sdk-java-v2/blob/master/CONTRIBUTING.md) document
- [x] Local run of `mvn install` succeeds
- [x] My code follows the code style of this project
- [ ] My change requires a change to the Javadoc documentation
- [ ] I have updated the Javadoc documentation accordingly
- [ ] I have added tests to cover my changes
- [x] All new and existing tests passed
- [ ] I have added a changelog entry. Adding a new entry must be accomplished by running the `scripts/new-change` script and following the instructions. Commit the new file created by the script in `.changes/next-release` with your changes.
- [ ] My change is to implement 1.11 parity feature and I have updated [LaunchChangelog](https://github.com/aws/aws-sdk-java-v2/blob/master/docs/LaunchChangelog.md)

## License

- [x] I confirm that this pull request can be released under the Apache 2 license

PR #4484, opened 2023-09-26
Validate that SignerProperty and IdentityProperty namespaces/names are unique.

millems:
This ensures that users won't unintentionally define two separate properties that are equal.

### DIFF
--- a/core/http-auth-spi/src/main/java/software/amazon/awssdk/http/auth/spi/signer/SignerProperty.java
+++ b/core/http-auth-spi/src/main/java/software/amazon/awssdk/http/auth/spi/signer/SignerProperty.java
@@ -16,9 +16,12 @@
 package software.amazon.awssdk.http.auth.spi.signer;
 
 import java.util.Objects;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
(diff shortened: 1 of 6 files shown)


core/http-auth-spi/src/main/java/software/amazon/awssdk/http/auth/spi/signer/SignerProperty.java
```diff
@@ -52,6 +58,18 @@ public static <T> SignerProperty<T> create(Class<?> namespace, String name) {
         return new SignerProperty<>(namespace.getName(), name);
     }
 
+    private void ensureUnique() {
```
gosar, line 61: Can we add some javadocs to the `create()` method that tells the reader about the uniqueness constraint? That'll also help clarify more that `Class<?> namespace` is not for the type of property value. See earlier discussion in https://github.com/aws/aws-sdk-java-v2/pull/4448#discussion_r1330557121

gosar (approved):

core/http-auth-spi/src/main/java/software/amazon/awssdk/http/auth/spi/signer/SignerProperty.java
millems: Fixed

PR #4487, opened 2023-09-26
Add check for SRA codepath to HttpChecksumStage

haydenbaker:
## Motivation and Context
- Since the new signers handle flexible-checksums and `aws-chunk` coding for unsigned and signed payload cases, we don't need the duplicate logic in HttpChecksumStage
- However, we still need to support the old path when we _are not_ using new signers

## Modifications
- Add a condition to check if we are using new SRA code, and return early if we are

## Types of changes

## Checklist

## License


core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/HttpChecksumStage.java
```diff
@@ -67,6 +68,11 @@ public SdkHttpFullRequest.Builder execute(SdkHttpFullRequest.Builder request, Re
             return request;
         }
 
+        // Check for SRA code-path
```
zoewangg, line 71: So we'll still need to keep this class to add md5 checksum header (after SRA is fully released)?

zoewangg (dismissed):

zoewangg:
Close in favor of https://github.com/aws/aws-sdk-java-v2/pull/4503/

PR #4493, opened 2023-09-27
Add additional backwards-compatibility testing for legacy clients.

millems:
1. Ensure anonymous credentials function correctly
2. Ensure attributes set in custom execution interceptors make it to the legacy signer

### DIFF
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/interceptor/ExecutionAttribute.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/interceptor/ExecutionAttribute.java
@@ -269,7 +269,10 @@ public final class ExecutionAttribute<T> {
 
         @Override
         public void setIfAbsent(Map<ExecutionAttribute<?>, Object> attributes, T value) {
-            attributes.computeIfAbsent(realAttribute, k -> writeMapping.apply(null, value));
+            T currentValue = get(attributes);
+            if (currentValue == null) {
+                set(attributes, value);
+            }
         }
     }
 
--- a/test/old-client-version-compatibility-test/src/test/java/InterceptorSignerAttributeTest.java
+++ b/test/old-client-version-compatibility-test/src/test/java/InterceptorSignerAttributeTest.java
@@ -62,7 +62,8 @@ import software.amazon.awssdk.testutils.service.http.MockAsyncHttpClient;
 import software.amazon.awssdk.testutils.service.http.MockSyncHttpClient;
 
 /**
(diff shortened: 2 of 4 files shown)


core/sdk-core/src/main/java/software/amazon/awssdk/core/interceptor/ExecutionAttribute.java
```diff
@@ -269,7 +269,10 @@ public void set(Map<ExecutionAttribute<?>, Object> attributes, T value) {
 
         @Override
         public void setIfAbsent(Map<ExecutionAttribute<?>, Object> attributes, T value) {
-            attributes.computeIfAbsent(realAttribute, k -> writeMapping.apply(null, value));
+            T currentValue = get(attributes);
+            if (currentValue == null) {
+                set(attributes, value);
+            }
```
millems, line 275: This is more correct, since the presence of the real attribute does not necessarily indicate that the execution attribute value is present (e.g. a signer property's existence isn't guaranteed even if the selected auth scheme is present).

test/old-client-version-compatibility-test/src/test/java/InterceptorSignerAttributeTest.java
```diff
@@ -0,0 +1,244 @@
...
+import software.amazon.awssdk.services.s3.S3AsyncClient;
+import software.amazon.awssdk.services.s3.S3AsyncClientBuilder;
+import software.amazon.awssdk.services.s3.S3Client;
+import software.amazon.awssdk.services.s3.S3ClientBuilder;
+import software.amazon.awssdk.services.s3.model.ChecksumAlgorithm;
+import software.amazon.awssdk.testutils.service.http.MockAsyncHttpClient;
+import software.amazon.awssdk.testutils.service.http.MockSyncHttpClient;
+
+/**
+ * Ensure that attributes set in execution interceptors are passed to custom signers.
```
gosar, line 65: nit: can you add a note here saying the attributes being tested here are SdkProtectedApi, but we still want to make sure if works if customer set these via interceptors.

gosar (approved):
Were you thinking of testing S3SignerExecutionAttributes in later PR?

core/sdk-core/src/main/java/software/amazon/awssdk/core/interceptor/ExecutionAttribute.java
gosar: This fixed issues we've identified in other testing, even with new clients, since we still have code in core that does putIfAbsent after auth scheme is selected. Thank you!

PR #4495, opened 2023-09-27
Make s3control work with preSRA codegen

gosar:
## Motivation and Context

This is follow up from https://github.com/aws/aws-sdk-java-v2/pull/4491, where I forced s3control to useSraAuth=true path since tests were failing with useSraAuth=false. Fixing it here.

## Modifications

With https://github.com/aws/aws-sdk-java-v2/pull/4396 since we could have a dummy auth scheme (which would definitely be true for useSraAuth=false), we add the SignerProperties to the SelectedAuthScheme, irrespective of the schemeId. For useSraAuth path, where old Signer is used, this effectively adds the ExecutionAttributes that old Signer needs because of https://github.com/aws/aws-sdk-java-v2/pull/4396.

## Testing

` ./mvnw clean install -pl :codegen,:s3control` passes locally.

## Types of changes

## Checklist

## License

### DIFF
--- a/codegen/src/test/resources/software/amazon/awssdk/codegen/poet/rules/endpoint-resolve-interceptor.java
+++ b/codegen/src/test/resources/software/amazon/awssdk/codegen/poet/rules/endpoint-resolve-interceptor.java
@@ -125,9 +125,6 @@ public final class QueryResolveEndpointInterceptor implements ExecutionIntercept
     private <T extends Identity> SelectedAuthScheme<T> authSchemeWithEndpointSignerProperties(
         List<EndpointAuthScheme> endpointAuthSchemes, SelectedAuthScheme<T> selectedAuthScheme) {
         for (EndpointAuthScheme endpointAuthScheme : endpointAuthSchemes) {
-            if (!endpointAuthScheme.schemeId().equals(selectedAuthScheme.authSchemeOption().schemeId())) {
-                continue;
-            }
             AuthSchemeOption.Builder option = selectedAuthScheme.authSchemeOption().toBuilder();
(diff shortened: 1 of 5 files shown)


codegen/src/test/resources/software/amazon/awssdk/codegen/poet/rules/endpoint-resolve-interceptor.java
```diff
@@ -125,9 +125,6 @@ private static void operationWithStaticContextParamsStaticContextParams(QueryEnd
     private <T extends Identity> SelectedAuthScheme<T> authSchemeWithEndpointSignerProperties(
         List<EndpointAuthScheme> endpointAuthSchemes, SelectedAuthScheme<T> selectedAuthScheme) {
         for (EndpointAuthScheme endpointAuthScheme : endpointAuthSchemes) {
-            if (!endpointAuthScheme.schemeId().equals(selectedAuthScheme.authSchemeOption().schemeId())) {
```
gosar, line 128: Thanks @millems for highlighting that not checking this could be an issue for useSraAuth=true client, in a case for example where SelectedAuthScheme=sigv4a, but endpointAuthSchemes=["sigv4","sigv4a"] and the shared signer properties like disableDoubleEncoding have different values from endpoint rules. Without this check, the value from sigv4 EndpointAuthScheme will get used, even though it should use from sigv4a EndpointAuthScheme.

Will address this is separate PR.

gosar: addressing in https://github.com/aws/aws-sdk-java-v2/pull/4505

PR #4499, opened 2023-09-28
Disable S3 payload signing.

millems:
This is currently redundant, but when useSraAuth is made true, it will ensure that payload signing remains disabled.

### DIFF
--- a/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/handlers/DisablePayloadSigningInterceptor.java
+++ b/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/handlers/DisablePayloadSigningInterceptor.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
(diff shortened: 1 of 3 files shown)


gosar (approved):
Pre SRA does customer have a way to enable payload signing? Thought being, that if there's a way, maybe have a test that confirms that still works.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/handlers/DisablePayloadSigningInterceptor.java
```diff
@@ -0,0 +1,33 @@
...
+
+import software.amazon.awssdk.auth.signer.S3SignerExecutionAttribute;
+import software.amazon.awssdk.core.interceptor.Context;
+import software.amazon.awssdk.core.interceptor.ExecutionAttributes;
+import software.amazon.awssdk.core.interceptor.ExecutionInterceptor;
+
+/**
+ * Disables payload signing for all S3 operations.
+ *
+ * <p>TODO: After S3's migration to the SRA, we should use signer properties in the auth scheme resolver.
```
gosar, line 26: nit:
```suggestion
 * <p>TODO(sra-identity-auth): After S3's migration to the SRA, we should use signer properties in the auth scheme resolver.
```

millems: Fixed

millems:
> Pre SRA does customer have a way to enable payload signing? Thought being, that if there's a way, maybe have a test that confirms that still works.

Yes, via interceptors. That's tested in the backwards-compatibility changes I pushed yesterday, but it's not actually tested with the latest changes but useSraAuth = false. Let me add a test for that...

gosar:
> Yes, via interceptors. 

yeah, I was wondering if there's another _supported_ way (sounds like no), since the `S3SignerExecutionAttribute.ENABLE_PAYLOAD_SIGNING` is SdkProtectedApi. But yeah, we still want to make sure it continues to work, so the test is good.

With SRA there will be a publicly supported way via SignerProperty! We should add a test for that (non-blocking right now).

My ship it stands anyways!

PR #4502, opened 2023-09-28
Remove CodeCatalystIntegration test

millems:
It's not possible to automatically test bearer token support for this service, because it requires a user to be present. Updated the test to mock out the service.


gosar (approved):
Nice. I added that test as part of my local testing for initial auth scheme resolver integration. Didn't know of a better way, and needed _some_ test. This is nicer.

PR #4503, opened 2023-09-28
Use HttpChecksumStage for async trailer support

zoewangg:
## Motivation and Context
Async signing is not supported in SRA signer at the moment, so we need to rely on existing HttpChecksumStage. Note that for sync, we still use new SRA signer.

## Modifications
Use HttpChecksumStage for async trailer support

## Testing
Ran `AsyncHttpChecksumIntegrationTest` and verified it worked.

### DIFF
--- a/core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/signer/DefaultAwsV4HttpSigner.java
+++ b/core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/signer/DefaultAwsV4HttpSigner.java
@@ -168,6 +168,33 @@ public final class DefaultAwsV4HttpSigner implements AwsV4HttpSigner {
         return V4PayloadSigner.create();
     }
 
+    private static V4PayloadSigner v4PayloadAsyncSigner(
+        BaseSignRequest<?, ? extends AwsCredentialsIdentity> request,
+        V4Properties properties) {
+
+        boolean isPayloadSigning = request.requireProperty(PAYLOAD_SIGNING_ENABLED, true);
+        boolean isEventStreaming = isEventStreaming(request.request());
+        boolean isChunkEncoding = request.requireProperty(CHUNK_ENCODING_ENABLED, false);
+
+        if (isEventStreaming) {
+            if (isPayloadSigning) {
+                return getEventStreamV4PayloadSigner(
+                    properties.getCredentials(),
+                    properties.getCredentialScope(),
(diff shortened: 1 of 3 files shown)


services/s3/src/it/java/software/amazon/awssdk/services/s3/checksum/AsyncHttpChecksumIntegrationTest.java
```diff
@@ -234,6 +234,11 @@ void asyncValidSignedTrailerChecksumCalculatedBySdkClient() {
         assertThat(response).isEqualTo("Hello world");
     }
 
+    /**
+     * If http is used, payload signing will be enforced, but it's not currently supported in async path
+     * TODO: re-enable it once it's supported
+     */
+    @Disabled("Payload signing is not supported for S3 async client")
```
zoewangg, line 241: This test is not supported today. The reason it passes is because payload signing is bypassed (request.contentStreamProvider() == null)

https://github.com/aws/aws-sdk-java-v2/blob/38e221bd815af31a6c6b91557499af155103c21a/core/auth/src/main/java/software/amazon/awssdk/auth/signer/internal/AbstractAwsS3V4Signer.java#L279-L285

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/HttpChecksumStage.java
```diff
@@ -74,6 +75,11 @@ public SdkHttpFullRequest.Builder execute(SdkHttpFullRequest.Builder request, Re
             return request;
         }
 
+        // If SRA is enabled, we skip flexible checksum in header
```
cenedhryn, line 78: Can we add a TODO to indicate that we should rewrite this codepath to remove flexible checksums in headers once the released code is stable?

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/signer/DefaultAwsV4HttpSigner.java
```diff
@@ -168,6 +168,33 @@ private static V4PayloadSigner v4PayloadSigner(
         return V4PayloadSigner.create();
     }
 
+    private static V4PayloadSigner v4PayloadAsyncSigner(
```
cenedhryn, line 171: Nice, this makes it clear.

cenedhryn (approved):

PR #4504, opened 2023-09-28
Add dependency on http-auth-aws-crt from auth-crt.

millems:
Customers currently depend on auth-crt to get Sigv4a support. In the future, customers should depend on http-auth-aws-crt for Sigv4a support. In order to ensure existing customers still work after SRA, we want users who depend on auth-crt today to automatically depend on http-auth-aws-crt.

### DIFF
--- a/core/auth-crt/pom.xml
+++ b/core/auth-crt/pom.xml
@@ -68,6 +68,13 @@
             <version>${awscrt.version}</version>
         </dependency>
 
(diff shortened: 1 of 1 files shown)


gosar (approved):
`http-auth-aws-crt` is basically a dummy package, so even for customers who pick this up without picking up new SRA clients, it's ok.

PR #4506, opened 2023-09-28
Invoke plugins at client build time

sugmanue:
## Motivation and Context

Adds the wiring to run the plugins at client build time. 

The plugins will be invoked **after** building the configuration and **before** the finalizing hooks, 

The final configuration used for the sync client is [built here](https://github.com/aws/aws-sdk-java-v2/blob/349a0e6918b90478d5704006ef348e5c0bb9f006/core/sdk-core/src/main/java/software/amazon/awssdk/core/client/builder/SdkDefaultClientBuilder.java#L169) for the sync configuration and [built here](https://github.com/aws/aws-sdk-java-v2/blob/349a0e6918b90478d5704006ef348e5c0bb9f006/core/sdk-core/src/main/java/software/amazon/awssdk/core/client/builder/SdkDefaultClientBuilder.java#L197) for the async configuration. 

During the first phase (as demarcated by the comments) the override configuration and defaults creates the initial revision of the configuration.

```
        // Apply overrides
        configuration = setOverrides(configuration);

        // Apply defaults
        configuration = mergeChildDefaults(configuration);
        configuration = mergeGlobalDefaults(configuration);
```

The next part, the finalizers, create additional configuration values from the default applied-configuration.

```
        // Create additional configuration from the default-applied configuration
        configuration = finalizeChildConfiguration(configuration);
        configuration = finalizeSyncConfiguration(configuration);
        configuration = finalizeConfiguration(configuration);
```

For AWS clients we have logic to use the configured values to derive new configuration values, for instance [this is how we create](https://github.com/aws/aws-sdk-java-v2/blob/fa9dbcce47637486e3f7d4d366ab6509b535342a/core/aws-core/src/main/java/software/amazon/awssdk/awscore/client/builder/AwsDefaultClientBuilder.java#L207C1-L219C1) a default `IdentityProviders` class out of a single `IdentityProvider` of `AwsCredentialsProvider`.

```
        // Add the identityProvider to the IdentityProviders configured for the client.
        // Currently, it is not possible for identityProvider to be null as default provider is used while building the client if
        // the clientConfig is null. However, we do want to support ability to unset a identity provider later.
        // Moreover, putIdentityProvider will throw NPE on null, so adding the null check here. Also, validateClientOptions
        // currently asserts it is not null, which will have to change when we allow unsetting default identity provider.
        if (identityProvider != null) {
            configBuilder.option(SdkClientOption.IDENTITY_PROVIDERS,
                                 configuration.option(SdkClientOption.IDENTITY_PROVIDERS)
                                              .toBuilder()
                                              .putIdentityProvider(identityProvider)
                                              .build());
        }

```

This means that the plugins have to be able to update the configuration before the finalizers phase make decisions on it or somehow captures it, otherwise those changes won't be reflected on the changes made by the finalizers.

After this change the configuration will be built like the code below, invoking the plugins after defaults and before the finalizers.

```
        // Apply overrides
        configuration = setOverrides(configuration);
        // Apply defaults
        configuration = mergeChildDefaults(configuration);
        configuration = mergeGlobalDefaults(configuration);

        // Invoke the plugins after defaults and before finalizing the configuration.
        configuration = invokePluginsIfNeeded(configuration);

        // Create additional configuration from the default-applied configuration
        configuration = finalizeChildConfiguration(configuration);
        configuration = finalizeAsyncConfiguration(configuration);
```

## Types of changes

## Checklist

## License

### DIFF
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/client/builder/SdkDefaultClientBuilder.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/client/builder/SdkDefaultClientBuilder.java
@@ -58,11 +58,13 @@ import software.amazon.awssdk.annotations.SdkProtectedApi;
 import software.amazon.awssdk.annotations.SdkTestInternalApi;
 import software.amazon.awssdk.core.CompressionConfiguration;
 import software.amazon.awssdk.core.SdkPlugin;
+import software.amazon.awssdk.core.SdkServiceClientConfiguration;
 import software.amazon.awssdk.core.SdkSystemSetting;
 import software.amazon.awssdk.core.client.config.ClientAsyncConfiguration;
 import software.amazon.awssdk.core.client.config.ClientOverrideConfiguration;
 import software.amazon.awssdk.core.client.config.SdkClientConfiguration;
 import software.amazon.awssdk.core.client.config.SdkClientOption;
+import software.amazon.awssdk.core.client.config.internal.ConfigurationUpdater;
 import software.amazon.awssdk.core.client.config.internal.SdkClientConfigurationUtil;
 import software.amazon.awssdk.core.interceptor.ClasspathInterceptorChainFactory;
 import software.amazon.awssdk.core.interceptor.ExecutionInterceptor;
@@ -177,10 +179,14 @@ public abstract class SdkDefaultClientBuilder<B extends SdkClientBuilder<B, C>, 
         configuration = mergeChildDefaults(configuration);
         configuration = mergeGlobalDefaults(configuration);
 
+        // Invoke the plugins after defaults and before finalizing the configuration.
+        configuration = invokePluginsIfNeeded(configuration);
+
         // Create additional configuration from the default-applied configuration
         configuration = finalizeChildConfiguration(configuration);
         configuration = finalizeSyncConfiguration(configuration);
         configuration = finalizeConfiguration(configuration);
+
 
         return configuration;
     }
@@ -204,6 +210,9 @@ public abstract class SdkDefaultClientBuilder<B extends SdkClientBuilder<B, C>, 
         // Apply defaults
         configuration = mergeChildDefaults(configuration);
         configuration = mergeGlobalDefaults(configuration);
+
+        // Invoke the plugins after defaults and before finalizing the configuration.
+        configuration = invokePluginsIfNeeded(configuration);
 
         // Create additional configuration from the default-applied configuration
         configuration = finalizeChildConfiguration(configuration);
@@ -359,6 +368,14 @@ public abstract class SdkDefaultClientBuilder<B extends SdkClientBuilder<B, C>, 
                      .option(RETRY_POLICY, retryPolicy)
                      .option(CLIENT_USER_AGENT, resolveClientUserAgent(config, retryPolicy))
                      .build();
+    }
+
+    /**
+     * Invoke all the registered plugins and return the configuration.
+     */
+    protected SdkClientConfiguration invokePluginsIfNeeded(SdkClientConfiguration config) {
+        ConfigurationUpdater<SdkServiceClientConfiguration.Builder> configurationUpdater = defaultConfigurationUpdater();
+        return SdkClientConfigurationUtil.invokePlugins(config, registeredPlugins, configurationUpdater);
     }
 
     private String resolveClientUserAgent(SdkClientConfiguration config, RetryPolicy retryPolicy) {
@@ -559,6 +576,12 @@ public abstract class SdkDefaultClientBuilder<B extends SdkClientBuilder<B, C>, 
     public final B addPlugin(SdkPlugin plugin) {
         registeredPlugins.add(Validate.paramNotNull(plugin, "plugin"));
         return thisBuilder();
+    }
+
+    protected ConfigurationUpdater<SdkServiceClientConfiguration.Builder> defaultConfigurationUpdater() {
+        // No-op by default, service specific codegen classes will extend this method will have a proper
+        // implementation of it.
+        return (consumer, configBuilder) -> configBuilder.build();
     }
 
     /**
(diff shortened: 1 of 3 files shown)


test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/SraIdentityResolutionUsingPluginsTest.java
```diff
@@ -0,0 +1,92 @@
...
+import software.amazon.awssdk.utils.Validate;
+
+@RunWith(MockitoJUnitRunner.class)
+public class SraIdentityResolutionUsingPluginsTest {
+
+    @Mock
+    private AwsCredentialsProvider credentialsProvider;
+
+    @Test
+    public void testIdentityBasedPluginsResolutionIsUsedAndNotAnotherIdentityResolution() {
```
sugmanue, line 47: Adapted from this [other test](https://github.com/aws/aws-sdk-java-v2/blob/26d88e253a551cdd306440459c315c6b378e2f23/test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/SraIdentityResolutionTest.java#L43) to do the credentials provider override using plugins.

core/sdk-core/src/main/java/software/amazon/awssdk/core/client/config/internal/SdkClientConfigurationUtil.java
```diff
@@ -55,8 +57,10 @@ public static SdkClientConfiguration.Builder copyOverridesToConfiguration(
 
         // advanced option
         Signer signer = overrides.advancedOption(SdkAdvancedClientOption.SIGNER).orElse(null);
-        builder.option(SdkAdvancedClientOption.SIGNER, signer);
-        builder.option(SdkClientOption.SIGNER_OVERRIDDEN, signer != null);
+        if (signer != null) {
```
sugmanue, line 60: This logic needed to change like this in order to avoid overriding the client config that sets a default signer if there's no override.

core/sdk-core/src/main/java/software/amazon/awssdk/core/client/builder/SdkDefaultClientBuilder.java
```diff
@@ -177,11 +179,15 @@ protected final SdkClientConfiguration syncClientConfiguration() {
         configuration = mergeChildDefaults(configuration);
         configuration = mergeGlobalDefaults(configuration);
 
+        // Invoke the plugins after defaults and before finalizing the configuration.
+        configuration = invokePluginsIfNeeded(configuration);
```
cenedhryn, line 183: Where was it invoked before this, and which `finalizing hooks` did you refer to in the PR description?

test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/SraIdentityResolutionUsingPluginsTest.java
```diff
@@ -0,0 +1,92 @@
...
+            // Below is necessary to create the test case where, addCredentialsToExecutionAttributes was getting called before
+            .overrideConfiguration(ClientOverrideConfiguration.builder().build())
+            .build();
+
+        try {
+            syncClient.allTypes(builder -> {
+            });
+        } catch (Exception expected) {
+            expected.printStackTrace(System.out);
+        }
```
cenedhryn, line 64: Can we adapt this test (and the other one...) to throw an exception before we send the request, and not print a stack trace that just make the logs longer. Searching for `boom` (!) in the codebase gives a few different options.

core/sdk-core/src/main/java/software/amazon/awssdk/core/client/builder/SdkDefaultClientBuilder.java
sugmanue: I have updated the description to add more context and links to the referred code. Please let me know if that helps or if you would like me to expand on anything.

sugmanue: > Where was it invoked before this,

This was not invoked before, now it is at client build time. This is the last step that is used to build final configuration for the client. Some of the supporting code was added before but this just now enabled.

test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/SraIdentityResolutionUsingPluginsTest.java
sugmanue: > Can we adapt this test (and the other one...) to throw an exception before we send the request,

Sure, not sure how to do that, I can look around but if you have any pointers that will be helpful.

> and not print a stack trace that just make the logs longer.

For sure, I will remove it, it was not really meant to part of this change.

core/sdk-core/src/main/java/software/amazon/awssdk/core/client/builder/SdkDefaultClientBuilder.java
cenedhryn: Ah, I see - the finalizing config methods were what you were referring to. Thank you for updating the PR description. Yes, we'd have to call it no later than here.

test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/SraIdentityResolutionUsingPluginsTest.java
cenedhryn: I like this pattern:

```
SdkHttpClient mockClient = mock(SdkHttpClient.class);
when(mockClient.prepareRequest(any())).thenThrow(new RuntimeException("boom"));

ProtocolQueryClient syncClient = ProtocolQueryClient
    .builder()
    .httpClient(mockClient)
    .build();

assertThatThrownBy(() -> syncClient.allTypes(r -> {})).hasMessageContaining("boom");
```

cenedhryn (approved):

test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/SraIdentityResolutionUsingPluginsTest.java
sugmanue: Sounds good, I will create a follow up PR to adjust the tests accordingly.

PR #4514, opened 2023-09-29
Disable SRA in CRT S3 client

zoewangg:
## Motivation and Context
In CRT based S3 client, how checksum is handled depends on the type of the requests. For putObject and getObject, it's done in CRT. For other requests, it's done in `HttpChecksumStage`. Since in the new SRA path, checksum is done as part of the signer, we need to explicitly disable SRA in CRT based S3 client in order to continue to use `HttpChecksumStage` for checksumming for requests other than putObject and getObject.

## Modifications
Disable SRA in CRT S3 client

## Testing
`ChecksumIntegrationTest` passed locally

### DIFF
--- a/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/DefaultS3CrtAsyncClient.java
+++ b/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/DefaultS3CrtAsyncClient.java
@@ -15,6 +15,7 @@
 
 package software.amazon.awssdk.services.s3.internal.crt;
 
+import static software.amazon.awssdk.core.interceptor.SdkInternalExecutionAttribute.AUTH_SCHEMES;
 import static software.amazon.awssdk.core.interceptor.SdkInternalExecutionAttribute.SDK_HTTP_EXECUTION_ATTRIBUTES;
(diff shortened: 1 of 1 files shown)


cenedhryn (approved):

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/DefaultS3CrtAsyncClient.java
```diff
@@ -280,6 +283,15 @@ public S3CrtAsyncClient build() {
     }
 
     private static final class AttachHttpAttributesExecutionInterceptor implements ExecutionInterceptor {
+
+        @Override
+        public void beforeExecution(Context.BeforeExecution context, ExecutionAttributes executionAttributes) {
+            // Hack to disable new SRA path because we still rely on HttpChecksumStage to perform checksum for
+            // non-streaming operation.
+            // TODO: remove this once CRT supports checksum for default requests
```
gosar, line 291: @zoewangg @cenedhryn what does "default requests" here mean? Does it mean non-streaming requests?

cenedhryn: Yes

PR #4450, opened 2023-09-19
Support non-Json String payloads

davidh44:
## Motivation and Context

To support non-Json String payloads

## Modifications

Update Json marshaller/unmarshaller to support explicit String payloads, including non-Json Strings

## Testing

Added unit tests

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

## Checklist

- [ ] I have read the [CONTRIBUTING](https://github.com/aws/aws-sdk-java-v2/blob/master/CONTRIBUTING.md) document
- [ ] Local run of `mvn install` succeeds
- [ ] My code follows the code style of this project
- [ ] My change requires a change to the Javadoc documentation
- [ ] I have updated the Javadoc documentation accordingly
- [ ] I have added tests to cover my changes
- [ ] All new and existing tests passed
- [ ] I have added a changelog entry. Adding a new entry must be accomplished by running the `scripts/new-change` script and following the instructions. Commit the new file created by the script in `.changes/next-release` with your changes.
- [ ] My change is to implement 1.11 parity feature and I have updated [LaunchChangelog](https://github.com/aws/aws-sdk-java-v2/blob/master/docs/LaunchChangelog.md)

## License

- [x] I confirm that this pull request can be released under the Apache 2 license

PR #4553, opened 2023-10-05
Fail codegen for Query protocol operations with explicit String payload

davidh44:
## Motivation and Context

Explicit String payloads are not supported for Query protocols

## Modifications

Update codegen to fail if there is Query protocol operation with explicit String payload

Update Json codegen test classes which were missed in previous [PR](https://github.com/aws/aws-sdk-java-v2/pull/4450) - `string` can be lowercase when calling `shape.getType()`

## Types of changes

## Checklist

## License

### DIFF
--- a/codegen/src/main/java/software/amazon/awssdk/codegen/AddOperations.java
+++ b/codegen/src/main/java/software/amazon/awssdk/codegen/AddOperations.java
@@ -183,7 +183,7 @@ final class AddOperations {
 
                 if (isQueryProtocolWithExplicitStringPayload(c2jShapes, c2jShapes.get(inputShape))) {
                     throw SdkClientException.create("Operations with explicit String payloads are not supported for Query "
-                                                    + "protocols.");
+                                                    + "protocols. Unsupported operation: " + operationName);
                 }
 
                 String documentation = input.getDocumentation() != null ? input.getDocumentation() :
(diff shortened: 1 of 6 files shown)


codegen/src/main/java/software/amazon/awssdk/codegen/AddOperations.java
```diff
@@ -73,7 +74,7 @@ private static boolean isBlobShape(Shape shape) {
      * @return True if shape is a String type. False otherwise
      */
     private static boolean isStringShape(Shape shape) {
-        return shape != null && "String".equals(shape.getType());
+        return shape != null && shape.getType().equalsIgnoreCase("String");
```
dagnir, line 77: Let's keep the original ordering and compare to the constant: `"String".equalsIgnoreCase(shape.getType())`

codegen/src/main/java/software/amazon/awssdk/codegen/AddOperations.java
```diff
@@ -179,6 +180,12 @@ public Map<String, OperationModel> constructOperations() {
             if (input != null) {
                 String originalShapeName = input.getShape();
                 String inputShape = namingStrategy.getRequestClassName(operationName);
+
+                if (isQueryProtocolWithExplicitStringPayload(c2jShapes, c2jShapes.get(inputShape))) {
```
dagnir, line 184: I don't like the idea of doing validation in here since we're not doing it for anything else. Can we do this validation as a preprocessor instead? https://github.com/aws/aws-sdk-java-v2/blob/ea1bb9b7e725c0a1c53fe34336ae7902d3a8924b/codegen/src/main/java/software/amazon/awssdk/codegen/customization/CodegenCustomizationProcessor.java#L21 similar to https://github.com/aws/aws-sdk-java-v2/blob/ea1bb9b7e725c0a1c53fe34336ae7902d3a8924b/codegen/src/main/java/software/amazon/awssdk/codegen/customization/processors/NewAndLegacyEventStreamProcessor.java#L34

codegen/src/main/java/software/amazon/awssdk/codegen/AddOperations.java
davidh44: +1

codegen/src/main/java/software/amazon/awssdk/codegen/AddOperations.java
davidh44: makes sense, moving to preproccesor

dagnir (approved):

PR #4560, opened 2023-10-06
Update presigner so host is set and content-hash is excluded from signed-headers

haydenbaker:
## Motivation and Context
- Presigner only expects a single signed-header (host) for browser-compatible requests

## Modifications
- Update presigner request-signer so that it correctly sets host and excludes

## Types of changes

## Checklist

## License

### DIFF
--- a/core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/signer/V4RequestSigner.java
+++ b/core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/signer/V4RequestSigner.java
@@ -58,8 +58,7 @@ public interface V4RequestSigner {
             addHostHeader(requestBuilder);
             addDateHeader(requestBuilder, formatDateTime(properties.getCredentialScope().getInstant()));
 
-            V4RequestSigningResult result = create(properties).sign(requestBuilder);
-            V4Context ctx = create(properties, getContentHash(requestBuilder)).sign(requestBuilder);
+            V4RequestSigningResult result = create(properties, getContentHash(requestBuilder)).sign(requestBuilder);
 
             // Add the signature within an authorization header
             String authHeader = AWS4_SIGNING_ALGORITHM
@@ -92,8 +91,7 @@ public interface V4RequestSigner {
             requestBuilder.putRawQueryParameter(SignerConstant.X_AMZ_CREDENTIAL,
                                                 properties.getCredentialScope().scope(properties.getCredentials()));
 
-            V4RequestSigningResult result = create(properties).sign(requestBuilder);
-            V4Context ctx = create(properties, getContentHash(requestBuilder)).sign(requestBuilder);
+            V4RequestSigningResult result = create(properties, getContentHash(requestBuilder)).sign(requestBuilder);
 
             // Add the signature
             requestBuilder.putRawQueryParameter(SignerConstant.X_AMZ_SIGNATURE, result.getSignature());
@@ -128,12 +126,12 @@ public interface V4RequestSigner {
                                                 properties.getCredentialScope().scope(properties.getCredentials()));
             requestBuilder.putRawQueryParameter(SignerConstant.X_AMZ_EXPIRES, Long.toString(expirationDuration.getSeconds()));
 
-            V4Context ctx = create(properties, contentHash).sign(requestBuilder);
+            V4RequestSigningResult result = create(properties, contentHash).sign(requestBuilder);
 
             // Add the signature
-            requestBuilder.putRawQueryParameter(SignerConstant.X_AMZ_SIGNATURE, ctx.getSignature());
+            requestBuilder.putRawQueryParameter(SignerConstant.X_AMZ_SIGNATURE, result.getSignature());
 
-            return ctx;
+            return result;
 
         };
     }
--- a/core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/signer/V4RequestSigner.java
+++ b/core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/signer/V4RequestSigner.java
@@ -18,8 +18,11 @@ package software.amazon.awssdk.http.auth.aws.internal.signer;
 import static software.amazon.awssdk.http.auth.aws.internal.signer.V4CanonicalRequest.getCanonicalHeaders;
 import static software.amazon.awssdk.http.auth.aws.internal.signer.V4CanonicalRequest.getSignedHeadersString;
 import static software.amazon.awssdk.http.auth.aws.internal.signer.util.SignerConstant.AWS4_SIGNING_ALGORITHM;
+import static software.amazon.awssdk.http.auth.aws.internal.signer.util.SignerConstant.X_AMZ_CONTENT_SHA256;
 import static software.amazon.awssdk.http.auth.aws.internal.signer.util.SignerUtils.addDateHeader;
+import static software.amazon.awssdk.http.auth.aws.internal.signer.util.SignerUtils.addHostHeader;
 import static software.amazon.awssdk.http.auth.aws.internal.signer.util.SignerUtils.formatDateTime;
+import static software.amazon.awssdk.http.auth.aws.internal.signer.util.SignerUtils.getContentHash;
 
 import java.time.Duration;
 import java.util.List;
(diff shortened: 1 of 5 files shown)


core/http-auth-aws/src/test/java/software/amazon/awssdk/http/auth/aws/internal/signer/V4RequestSignerTest.java
```diff
@@ -106,15 +98,17 @@ public void sign_withPresigned_addsExpirationParam() {
...
         assertEquals("0", ctx.getSignedRequest().rawQueryParameters().get("X-Amz-Expires").get(0));
-        assertEquals("45d00c9c0eb2c05cfe7e42231a572aef30d96d1afb2a205e27d9c6b89867a865", ctx.getSignature());
+        assertEquals("host;x-amz-archive-description",
+                     ctx.getSignedRequest().rawQueryParameters().get("X-Amz-SignedHeaders").get(0));
+        assertEquals("691f39caa2064fe4fb897976dfb4b09df54749c825a5fcd1e2f0b3fcd1bcc600", ctx.getSignature());
     }
 
     @Test
-    public void sign_withNoChecksumHeader_throws() {
+    public void sign_withNoContentHashHeader_throws() {
```
gosar, line 107: I think we'll debate whether this header should always be included or not, but either ways, this test is now sign_withHeader_withNoContentHashHeader_throws()

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/signer/V4RequestSigner.java
```diff
@@ -77,6 +80,8 @@ static V4RequestSigner query(V4Properties properties) {
                 requestBuilder.putRawQueryParameter(SignerConstant.X_AMZ_SECURITY_TOKEN,
                                                     ((AwsSessionCredentialsIdentity) properties.getCredentials()).sessionToken());
             }
+            // We have to add the host-header here explicitly, since query-signed request requires it in the signed-header param
+            addHostHeader(requestBuilder);
```
gosar, line 84: I see addHostHeader logic here in query and presign, and in DefaultV4RequestSigner for header. If host header is needed in all cases, why duplicate the logic v/s move it higher up in single place?

core/http-auth-aws/src/test/java/software/amazon/awssdk/http/auth/aws/internal/signer/DefaultRequestSignerTest.java
```diff
@@ -37,21 +37,20 @@ public class DefaultRequestSignerTest {
                                             .normalizePath(true)
                                             .build();
 
-    V4RequestSigner requestSigner = new DefaultV4RequestSigner(v4Properties);
+    DefaultV4RequestSigner requestSigner = new DefaultV4RequestSigner(v4Properties, "quux");
 
     @Test
     public void requestSigner_sign_shouldReturnSignedContext_butNotAddAnyAuthInfoToRequest() {
```
gosar, line 43: what does "butNotAddAnyAuthInfoToRequest" mean? Don't see any asserts to that effect. Is the idea to assert that `request` before and after is the same? Except the host header? which relates to the other comment I made.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/signer/V4RequestSigner.java
haydenbaker: We can move it out of `DefaultV4RequestSigner`, sure.

core/http-auth-aws/src/test/java/software/amazon/awssdk/http/auth/aws/internal/signer/DefaultRequestSignerTest.java
haydenbaker: Yes, fixed - It asserts that nothing changed on the request.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/signer/V4RequestSigner.java
```diff
@@ -52,9 +55,10 @@ static V4RequestSigner header(V4Properties properties) {
                 requestBuilder.putHeader(SignerConstant.X_AMZ_SECURITY_TOKEN,
                                          ((AwsSessionCredentialsIdentity) properties.getCredentials()).sessionToken());
             }
+            addHostHeader(requestBuilder);
             addDateHeader(requestBuilder, formatDateTime(properties.getCredentialScope().getInstant()));
 
-            V4RequestSigningResult result = create(properties).sign(requestBuilder);
+            V4RequestSigningResult result = create(properties, getContentHash(requestBuilder)).sign(requestBuilder);
```
gosar, line 61: thinking aloud: should the contentHash be a constructor input like properties, or input to `sign()`?

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/signer/V4RequestSigner.java
```diff
@@ -100,11 +106,33 @@ static V4RequestSigner query(V4Properties properties) {
...
-            requestBuilder.putRawQueryParameter(SignerConstant.X_AMZ_EXPIRES,
-                                                Long.toString(expirationDuration.getSeconds())
-            );
+            // Add pre-requisites
+            if (properties.getCredentials() instanceof AwsSessionCredentialsIdentity) {
+                requestBuilder.putRawQueryParameter(SignerConstant.X_AMZ_SECURITY_TOKEN,
+                                                    ((AwsSessionCredentialsIdentity) properties.getCredentials()).sessionToken());
+            }
+            // We have to add the host-header here explicitly, since pre-signed request requires it in the signed-header param
+            addHostHeader(requestBuilder);
```
gosar, line 115: Does the v4a signer already do this for presigned case?
 Does v4a signer always include x-amz-content-sha256?

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/signer/V4RequestSigner.java
gosar: Out of `DefaultV4RequestSigner` and all in the static methods of V4RequestSigner is better than split across both sides. But it's still there multiple times. When I was thinking "move it higher up in single place", I was considering it in DefaultAwsV4HttpSigner before the requestSigner is invoked - did you mean that such manipulation of the requestBuilder to be responsibility of RequestSigner and have DefaultAwsV4HttpSigner just handle the logic flow between these different "processors" - request/payload/checksum.

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/signer/V4RequestSigner.java
haydenbaker: It's easier if it's a constructor param - sign's signature would change, which is a bigger change. If we decide that `x-amz-content-sha256` shouldn't be added as a header unless required (s3), then we'll likely need to refactor it as an input to request-signer and payload-signer

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/signer/V4RequestSigner.java
haydenbaker: CRT would be the thing that manipulates headers - the signer simply delegates to it. However, the signer _does_ configure CRT to always add the hash header, yes: `signingConfig.setSignedBodyHeader(X_AMZ_CONTENT_SHA256)`

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/signer/V4RequestSigner.java
haydenbaker: Yes, that was the thought. DefaultAwsV4HttpSigner should try to be focused on delegating rather than doing things itself.

gosar (approved):
We can close on whether x-amz-content-sha256 always added or not separately.

PR #4267, opened 2023-08-06
Provide a clearer exception message when building a ReadBatch or WriteBatch without setting the mappedTableResource (table)

swar8080:
Currently, a NullPointerException is thrown if you forget to call the `WriteBatch` builder with `mappedTableResource(MappedTableResource<T>)`. `ReadBatch` has the same problem

## Motivation and Context

```
    DynamoDbEnhancedClient client ...
    DynamoDbTable<Customer> customerTable = client.table("customer-table", customerSchema);
    Customer customer = ...

    // This will fail with a NullPointerException, both for addPutItem* and addDeleteItem*
    client.batchWriteItem(builder -> builder.addWriteBatch(
        WriteBatch.builder(Customer.class)
            .addPutItem(customer)
            .build()
    ));

   // This works correctly
   client.batchWriteItem(builder -> builder.addWriteBatch(
        WriteBatch.builder(Customer.class)
            .mappedTableResource(customerTable)
            .addPutItem(customer)
            .build()
    ));
```

## Modifications
Add a clearer error message in the `NullPointerException` letting the person know they forgot to set a `mappedTableResource`

## Testing
Updated unit tests

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist
- [x] I have read the [CONTRIBUTING](https://github.com/aws/aws-sdk-java-v2/blob/master/CONTRIBUTING.md) document
- [x] Local run of `mvn install` succeeds
- [x] My code follows the code style of this project
- [x] My change requires a change to the Javadoc documentation
- [x] I have updated the Javadoc documentation accordingly
- [x] I have added tests to cover my changes
- [x] All new and existing tests passed
- [x] I have added a changelog entry. Adding a new entry must be accomplished by running the `scripts/new-change` script and following the instructions. Commit the new file created by the script in `.changes/next-release` with your changes.

## License
- [x] I confirm that this pull request can be released under the Apache 2 license

PR #4566, opened 2023-10-09
Defer requirement of mappedTableResource until it is required

dagnir:
## Motivation and Context
Fix an issue where we were previously enforcing the requirement for mappedTableResource to be present before it was actually required. Now, we defer the validation utnil we actually need to access it.

## Types of changes

## Checklist

## License

### DIFF
--- a/services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/model/WriteBatch.java
+++ b/services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/model/WriteBatch.java
@@ -188,9 +188,6 @@ public final class WriteBatch {
     }
 
     private static final class BuilderImpl<T> implements Builder<T> {
-        private static final String MAPPED_TABLE_RESOURCE_NOT_NULL_MESSAGE = String.format("A mappedTableResource (table) is "
-                                                                                           + "required when building a %s",
-                                                                                           WriteBatch.class.getSimpleName());
         private Class<? extends T> itemClass;
         private List<Supplier<WriteRequest>> itemSupplierList = new ArrayList<>();
         private MappedTableResource<T> mappedTableResource;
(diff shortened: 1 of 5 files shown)


services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/model/WriteBatch.java
```diff
@@ -226,13 +222,12 @@ public Builder<T> addDeleteItem(Key key) {
 
         @Override
         public Builder<T> addDeleteItem(T keyItem) {
```
joviegas, line 224:
```java
WriteBatch.builder(dbBeanClass)
   .addDeleteItem(putItem)
   .mappedTableResource(table)
   .build())
```

We will see the issue in the above case too with the current fix.

I think the right fix is collect the DeleteItem/PutItemRequest and then at the end do the null check in the build()

Or another option which is not backward compatible is arrange the chaining of Builders such that `mappedTableResource` always comes first

dagnir: This already throws NPE today if `mappedTableResource` is null because we call `mappedTableResource.keyFrom` on the line below. This change basically adds a better message to the exception in this case

joviegas: Yeap I guess https://github.com/aws/aws-sdk-java-v2/pull/4267 which was merges couple of weeks ago.
This added the check , I thing we need to rework or revert that PR

joviegas: Got it , Thanks

joviegas (approved):

Issue #1812, opened 2020-05-04
NPE in TransformingAsyncResponseHandler#onError


ktulinger:
## Describe the bug
From time to time, we get an ERROR log from `TransformingAsyncResponseHandler` telling us that there was an NPE. It does not affect our app in any way, however it feels that it shouldn't throw that.

## Expected Behavior
No log record nor NPE.

## Current Behavior

We get this log statement after NPE has been raised in aws sdk core

```
5352c77fa1b2:2020-05-04 09:36:31,681 [speculation-executor-1] ERROR s.a.a.c.i.h.BaseAsyncClientHandler - Error thrown from TransformingAsyncResponseHandler#onError, ignoring.
5352c77fa1b2:java.lang.NullPointerException: null
5352c77fa1b2:	at software.amazon.awssdk.core.internal.http.async.AsyncResponseHandler.onError(AsyncResponseHandler.java:76)
5352c77fa1b2:	at software.amazon.awssdk.core.internal.http.async.CombinedResponseAsyncHttpResponseHandler.onError(CombinedResponseAsyncHttpResponseHandler.java:70)
5352c77fa1b2:	at software.amazon.awssdk.core.internal.handler.BaseAsyncClientHandler.lambda$doExecute$3(BaseAsyncClientHandler.java:223)
5352c77fa1b2:	at software.amazon.awssdk.utils.FunctionalUtils.runAndLogError(FunctionalUtils.java:40)
5352c77fa1b2:	at software.amazon.awssdk.core.internal.handler.BaseAsyncClientHandler.doExecute(BaseAsyncClientHandler.java:220)
5352c77fa1b2:	at software.amazon.awssdk.core.internal.handler.BaseAsyncClientHandler.execute(BaseAsyncClientHandler.java:86)
5352c77fa1b2:	at software.amazon.awssdk.awscore.client.handler.AwsAsyncClientHandler.execute(AwsAsyncClientHandler.java:52)
5352c77fa1b2:	at software.amazon.awssdk.services.dynamodb.DefaultDynamoDbAsyncClient.query(DefaultDynamoDbAsyncClient.java:2707)
...
(our code)
```

## Steps to Reproduce
I don't know a deterministic way.

## Possible Solution
I realize that the NPE comes from `AsyncResponseHandler`:

```
    @Override
    public void onError(Throwable err) {
        streamFuture.completeExceptionally(err);
    }
```

where streamFuture is null, as `prepare()` has not been called yet.

## Context

[...]

PR #4567, opened 2023-10-09
Fixed bugs in async signing.

millems:
1. Fixed a regression where async request body futures were being subscribed to twice. This prevents failures in request payloads that don't support re-subscribing at the expense of buffering the payload into memory. Added a TODO to stop buffering into memory by adding chunked encoding support.
2. Fixed a bug where if the signer returned a failed future, the SDK would get a NPE and not complete the execution future.


core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/signer/DefaultAwsV4HttpSigner.java
```diff
@@ -257,16 +257,14 @@ private static CompletableFuture<AsyncSignedRequest> doSign(AsyncSignRequest<? e
...
-
-        return resultSigningResultFuture.thenApply(
-            resultSigningResult -> AsyncSignedRequest.builder()
-                                           .request(resultSigningResult.getSignedRequest().build())
-                                           .payload(payloadSigner.signAsync(request.payload().orElse(null), resultSigningResult))
-                                           .build()
-        );
+        return checksummer.checksum(request.payload().orElse(null), requestBuilder)
+                          .thenApply(payload -> {
+                              V4RequestSigningResult resultSigningResultFuture = requestSigner.sign(requestBuilder);
```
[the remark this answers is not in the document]
haydenbaker: or futureRequestSigningResult :)

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/async/AsyncResponseHandler.java
```diff
@@ -73,6 +73,9 @@ public void onStream(Publisher<ByteBuffer> publisher) {
 
     @Override
     public void onError(Throwable err) {
+        if (streamFuture == null) {
+            prepare();
+        }
```
gosar, line 78: For my understanding:
* so onError can possibly get called without prepare()?
* calling prepare here and then immediately calling completeExceptionally on it. is the idea to just make sure streamFuture is non-null or is also setting up the thenComplete important? Wondering about just doing `streamFuture = new CompletableFuture<>();` here instead of `prepare()`, and further why not initialize `streamFuture = new CompletableFuture<>()` in the field declaration itself instead of in `prepare()`.

haydenbaker: FWIU, `prepare()` would be called in the normal flow to get the transformed response (see `MakeAsyncHttpRequestStage`), so if there is an error after `prepare()` is called there, then `streamFuture` wouldn't be null. On the other hand, if there is an error before the normal flow is executed (e.g. `prepare()` wasn't called before an error occured), then `prepare` is called here so that the response-handler is called before exceptionally-completing the stream-future.

haydenbaker (approved):

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/async/AsyncResponseHandler.java
millems: > so onError can possibly get called without prepare()?

Yes. We aren't clear in the docs, but it can happen as it happened in this case. I think in general our use of prepare() should be pushed upstream more, since we have other cases where prepare() isn't getting called soon enough. I wanted to have a light touch here, since I know this is a fragile part of the codebase.

millems: If we did `streamFuture = new CompletableFuture<>()` instead of `prepare()`, the behavior would be the same, assuming `prepare()` is defined as it is today. Which we should use depends on what we think is safer in the future.

Initializing streamFuture in the field declaration is riskier. We don't really document the lifecycle of a handler, so if the same handler is used for retries it wouldn't work. I don't know exactly which we do.

gosar: Thanks Matt for helping me understand better. At least my questions seem ballpark relevant. Agree with the light touch approach - this change seems safest option to me right now.

gosar:
Need to run `new-change` now?

dagnir (approved):

seantsb:
I think this should be labeled as fixing https://github.com/aws/aws-sdk-java-v2/issues/1812 ?

PR #4568, opened 2023-10-09
Treat async-streaming-http as unsigned, add async tests

haydenbaker:
## Motivation and Context
- Today (pre-sra), when signing async+streaming+http requests, we do NOT fall-back to signed-payload like we do for sync requests
- With SRA, aws-chunk content-encoding is not currently supported in the async code path
- For now, HttpChecksumStage handles the (async) unsigned-payload with checksums (i.e. unsigned-streaming-with-trailer) cases, so the signer should not re-encode the payload for this case

## Modifications
- Update the deciding logic for checksum and payload handling in the signer
- Add async tests for async signing cases

## Types of changes

## Checklist

## License


core/http-auth-aws/src/test/java/software/amazon/awssdk/http/auth/aws/internal/signer/DefaultAwsV4HttpSignerTest.java
```diff
@@ -343,4 +344,337 @@ public void sign_WithPayloadSigningNullAndHttp_FallsBackToPayloadSigning() {
...
+
+    @Test
+    public void signAsync_WithHeaderAuthLocationAndExpirationDuration_Throws() {
+        AsyncSignRequest<? extends AwsCredentialsIdentity> request = generateBasicAsyncRequest(
+            AwsCredentialsIdentity.create("access", "secret"),
+            httpRequest -> {
+            },
+            signRequest -> signRequest
+                .putProperty(AUTH_LOCATION, AuthLocation.HEADER)
+                .putProperty(EXPIRATION_DURATION, Duration.ZERO)
```
gosar, line 385: not directly related to this change, should we validate that EXPIRATION_DURATION is non-zero? I see you are using ZERO in even a successful case earlier like sign_WithQueryAuthLocationAndExpiration_DelegatesToPresignedSigner. Maybe we can fix that separately, but for the purpose of this test, use a non zero value.

core/http-auth-aws/src/test/java/software/amazon/awssdk/http/auth/aws/internal/signer/DefaultAwsV4HttpSignerTest.java
```diff
@@ -343,4 +344,337 @@ public void sign_WithPayloadSigningNullAndHttp_FallsBackToPayloadSigning() {
...
+
+    @Test
+    public void signAsync_WithQueryAuthLocationAndExpiration_DelegatesToPresignedSigner() {
+        AsyncSignRequest<? extends AwsCredentialsIdentity> request = generateBasicAsyncRequest(
+            AwsCredentialsIdentity.create("access", "secret"),
+            httpRequest -> {
+            },
+            signRequest -> signRequest
+                .putProperty(AUTH_LOCATION, AuthLocation.QUERY_STRING)
+                .putProperty(EXPIRATION_DURATION, Duration.ZERO)
```
gosar, line 415: use non zero value

core/http-auth-aws/src/test/java/software/amazon/awssdk/http/auth/aws/internal/signer/DefaultAwsV4HttpSignerTest.java
```diff
@@ -343,4 +344,337 @@ public void sign_WithPayloadSigningNullAndHttp_FallsBackToPayloadSigning() {
...
+            utilities.when(() ->ClassLoaderHelper.loadClass(
+                "software.amazon.awssdk.http.auth.aws.eventstream.HttpAuthAwsEventStream",
+                false)
+            ).thenThrow(new ClassNotFoundException("boom!"));
+            Exception e = assertThrows(RuntimeException.class, () -> signer.signAsync(request).join());
+            assertThat(e).hasMessageContaining("http-auth-aws-eventstream");
+        }
+    }
+
+    @Disabled("Chunk-encoding is not currently supported in the Async signing path - it is handled in HttpChecksumStage for now.")
```
gosar, line 458: hmm, with disabled, I can't tell what the current behavior is. Does it throw exception or does something else. would it make sense to have these tests _document_ the current behavior with TODO explaining the not supported.

core/http-auth-aws/src/test/java/software/amazon/awssdk/http/auth/aws/internal/signer/DefaultAwsV4HttpSignerTest.java
```diff
@@ -343,4 +344,337 @@ public void sign_WithPayloadSigningNullAndHttp_FallsBackToPayloadSigning() {
...
+        );
+
+        AsyncSignedRequest signedRequest = signer.signAsync(request).join();
+
+        assertThat(signedRequest.request().firstMatchingHeader("x-amz-checksum-crc32")).isPresent();
+        assertThat(signedRequest.request().firstMatchingHeader("x-amz-content-sha256")).hasValue("UNSIGNED-PAYLOAD");
+    }
+
+    @Test
+    public void signAsync_WithEventStreamContentType_DelegatesToEventStreamPayloadSigner() {
```
gosar, line 587: The existing sign_WithEventStreamContentType_DelegatesToEventStreamPayloadSigner is actually using signAsync. Can we change that one to call sign() and show corresponding current behavior.

core/http-auth-aws/src/test/java/software/amazon/awssdk/http/auth/aws/internal/signer/DefaultAwsV4HttpSignerTest.java
```diff
@@ -343,4 +344,337 @@ public void sign_WithPayloadSigningNullAndHttp_FallsBackToPayloadSigning() {
...
+            }
+        );
+
+        AsyncSignedRequest signedRequest = signer.signAsync(request).join();
+
+        assertThat(signedRequest.payload().get()).isInstanceOf(SigV4DataFramePublisher.class);
+    }
+
+    @Test
+    public void signAsync_WithEventStreamContentTypeAndUnsignedPayload_Throws() {
```
gosar, line 602: same as above.

core/http-auth-aws/src/test/java/software/amazon/awssdk/http/auth/aws/internal/signer/DefaultAwsV4HttpSignerTest.java
```diff
@@ -343,4 +344,337 @@ public void sign_WithPayloadSigningNullAndHttp_FallsBackToPayloadSigning() {
         assertThat(signedRequest.request().firstMatchingHeader("x-amz-content-sha256"))
             .hasValue("a15c8292b1d12abbbbe4148605f7872fbdf645618fee5ab0e8072a7b34f155e2");
     }
+
+    @Test
```
gosar, line 348: for code-organization, I wonder if keeping the sign/signAsync tests for each use case next to each other is better. Easier to compare both tests together, and guide new test writer to add tests for both cases.

core/http-auth-aws/src/test/java/software/amazon/awssdk/http/auth/aws/internal/signer/DefaultAwsV4HttpSignerTest.java
```diff
@@ -343,4 +344,337 @@ public void sign_WithPayloadSigningNullAndHttp_FallsBackToPayloadSigning() {
...
+        );
+
+        AsyncSignedRequest signedRequest = signer.signAsync(request).join();
+
+        assertThat(signedRequest.request().firstMatchingHeader("x-amz-content-sha256"))
+            .hasValue("a15c8292b1d12abbbbe4148605f7872fbdf645618fee5ab0e8072a7b34f155e2");
+    }
+
+    // TODO(sra-identity-and-auth): Once chunk-encoding is implemented in the async path, the assertions this test makes should
+    //  be different - it should asserted that signed-payload (STREAMING-AWS4-HMAC-SHA256-PAYLOAD) is the fallback.
```
gosar, line 645: Thinking about the 2 customers of this library:
* AWS SDKs - since this is existing behavior for AWS SDK, we can figure out what it means for SDKs if/when we change this.
* Direct http-auth-aws customers. We'll get these customers when we release these modules. I'm afraid of them also getting this undesirable behavior and increasing the risk/surface area when we make that change later. Would we just consider this a bugfix or is that reason to consider not making this SdkPublicApi yet or some other guidance to let customer know about issues/limiations.

cc @millems @cenedhryn

core/http-auth-aws/src/test/java/software/amazon/awssdk/http/auth/aws/internal/signer/DefaultAwsV4HttpSignerTest.java
```diff
@@ -343,4 +344,337 @@ public void sign_WithPayloadSigningNullAndHttp_FallsBackToPayloadSigning() {
...
+    @Test
+    public void signAsync_WithPayloadSigningTrueAndChunkEncodingTrueAndHttp_DoesNotFallBackToPayloadSigning() {
```
gosar, line 647: Did you mean for WithPayloadSigning*False* for these 2 tests, given that you say "DoesNotFallBack"?

Well, it might be good to have both cases:
* signAsync_WithPayloadSigningTrueAndChunkEncodingTrueAndHttp_IgnoresPayloadSigning (your current test)
* signAsync_WithPayloadSigningFalseAndChunkEncodingTrueAndHttp_DoesNotFallBackToPayloadSigning

core/http-auth-aws/src/test/java/software/amazon/awssdk/http/auth/aws/internal/signer/DefaultAwsV4HttpSignerTest.java
haydenbaker: The test is disabled because it would fail otherwise. It makes the correct assertions.

core/http-auth-aws/src/test/java/software/amazon/awssdk/http/auth/aws/internal/signer/DefaultAwsV4HttpSignerTest.java
haydenbaker: This is behavior that is in the SDK today, the test just asserts that the signing happens in the same way instead of throwing an exception (for not implemented or unsupported).

core/http-auth-aws/src/test/java/software/amazon/awssdk/http/auth/aws/internal/signer/DefaultAwsV4HttpSignerTest.java
haydenbaker: No, theses tests are asserting that the signer uses unsigned-streamining-payload (as it does today) when payload-signing is true and http is used.

core/http-auth-aws/src/test/java/software/amazon/awssdk/http/auth/aws/internal/signer/DefaultAwsV4HttpSignerTest.java
haydenbaker: Maybe? But there may be slight differences in behavior between sync and async, so you'd end up with a weird mishmash of sync_xyz, async_abc_xyz, async_def_xyz, and so on.

core/http-auth-aws/src/test/java/software/amazon/awssdk/http/auth/aws/internal/signer/DefaultAwsV4HttpSignerTest.java
haydenbaker: Yeah, probably. Added validation for duration > 0

core/http-auth-aws/src/test/java/software/amazon/awssdk/http/auth/aws/internal/signer/DefaultAwsV4HttpSignerTest.java
gosar: Imo, probably still better, as in majority of the cases, they'll be similar. And if there are multiple sub-cases in async compared to sync, seeing them together is also useful.

Either ways, it's a nit, and non-blocking.

core/http-auth-aws/src/test/java/software/amazon/awssdk/http/auth/aws/internal/signer/DefaultAwsV4HttpSignerTest.java
haydenbaker: Added comments.

core/http-auth-aws/src/test/java/software/amazon/awssdk/http/auth/aws/internal/signer/DefaultAwsV4HttpSignerTest.java
haydenbaker: Sure, will re-order

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/signer/DefaultAwsV4HttpSigner.java
```diff
@@ -270,11 +293,11 @@ private static CompletableFuture<AsyncSignedRequest> doSign(AsyncSignRequest<? e
     }
 
     private static Duration validateExpirationDuration(Duration expirationDuration) {
-        if (expirationDuration.compareTo(PRESIGN_URL_MAX_EXPIRATION_DURATION) > 0) {
+        if (expirationDuration.compareTo(Duration.ZERO) < 1 || expirationDuration.compareTo(PRESIGN_URL_MAX_EXPIRATION_DURATION) > 0) {
             throw new IllegalArgumentException(
-                "Requests that are pre-signed by SigV4 algorithm are valid for at most 7" +
+                "Requests that are pre-signed by SigV4 algorithm are valid for at least 1 second and at most 7" +
```
gosar, line 298: can be less than 1 second right? just non-zero. Just `new IllegalArgumentException("EXPIRATION_DURATION must be non-zero and at most 7 days")`?

haydenbaker: Minimum value is 1 second per https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-query-string-auth.html

gosar: Ah ok. Then the check should be `expirationDuration.compareTo(Duration.ofSeconds(1)) < 0`?

Also, what's our rounding behavior? Maybe a test should assert the expected rounding behavior.

core/http-auth-aws/src/test/java/software/amazon/awssdk/http/auth/aws/internal/signer/DefaultAwsV4HttpSignerTest.java
```diff
@@ -175,10 +370,49 @@ public void sign_WithChunkEncodingTrue_DelegatesToAwsChunkedPayloadSigner() {
...
+            AwsCredentialsIdentity.create("access", "secret"),
+            httpRequest -> httpRequest
+                .putHeader(Header.CONTENT_LENGTH, "20"),
+            signRequest -> signRequest
+                .putProperty(CHUNK_ENCODING_ENABLED, true)
+        );
+
+        AsyncSignedRequest signedRequest = signer.signAsync(request).join();
+
+        assertThat(signedRequest.request().firstMatchingHeader("x-amz-content-sha256"))
```
gosar, line 391: assert Content-Encoding? here and in each chunked encoding test case below.

core/http-auth-aws/src/test/java/software/amazon/awssdk/http/auth/aws/internal/signer/DefaultAwsV4HttpSignerTest.java
```diff
@@ -175,10 +370,49 @@ public void sign_WithChunkEncodingTrue_DelegatesToAwsChunkedPayloadSigner() {
...
+            .hasValue("STREAMING-AWS4-HMAC-SHA256-PAYLOAD");
+        assertThat(signedRequest.request().firstMatchingHeader(Header.CONTENT_LENGTH)).hasValue("193");
         assertThat(signedRequest.request().firstMatchingHeader("x-amz-decoded-content-length")).hasValue("20");
     }
 
+    // TODO(sra-identity-and-auth): Replace this test with the above test once chunk-encoding support is added
+    @Test
+    public void signAsync_WithChunkEncodingTrue_DelegatesToAwsChunkedPayloadSigner() {
```
gosar, line 399: so right now, it doesn't get Content-Enoding=aws-chunked?

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/signer/DefaultAwsV4HttpSigner.java
gosar: I have some follow ups on the rounding and edge cases around 0 < duration < 1 second , but let's take that separately, as I don't want to block this PR that fixes broken tests on it.

gosar (approved):

core/http-auth-aws/src/test/java/software/amazon/awssdk/http/auth/aws/internal/signer/DefaultAwsV4HttpSignerTest.java
haydenbaker: No, the signer does not do the encoding in this case (HttpChecksumStage would in a real request).

core/http-auth-aws/src/test/java/software/amazon/awssdk/http/auth/aws/internal/signer/DefaultAwsV4HttpSignerTest.java
gosar: I see you added the assert to "disabled tests". sorry I said "in each chunked encoding test case below". I meant to say all other chunked encoding cases. At least the basic sign_WithChunkEncodingTrue_DelegatesToAwsChunkedPayloadSigner case, but might as well in all chunked encoding cases.

davidh44 (approved):

PR #4569, opened 2023-10-10
Move S3Presigner fully to SRA with HttpSigner

gosar:
## Motivation and Context

S3PresignerTest.accessPointArn_multiRegion_useArnRegionTrue_correctEndpointAndSigner currently fails in feature/master/sra-identity-auth-testing (with useSraAuth=true) with `X-Amz-Algorithm`
```
Expected :"AWS4-ECDSA-P256-SHA256"
Actual   :"AWS4-HMAC-SHA256"
```
The reason is that DefaultS3Presigner is partially on SRA. i.e., it uses S3AuthSchemeResolver, but uses old Signer. And S3ResolveEndpointInterceptor won't have logic to use to old sigv4a Signer because of [this conditional logic](https://github.com/aws/aws-sdk-java-v2/blob/d2a0473a918738718db80f94c4037e0bc91a610d/codegen/src/main/java/software/amazon/awssdk/codegen/poet/rules/EndpointResolverInterceptorSpec.java#L186-L194) in codegen. Instead of solving this via more codegen changes, we might as well move the presigner fully to SRA and use the new HttpSigner from auth scheme resolution, which will select the correct sigv4/sigv4a HttpSigner.

## Modifications

* Move the S3Presigner fully to SRA and use the new HttpSigner
Note, testing this end-to-end, noticed some behaviors of presign implementation for HttpSigner that are being fixed in https://github.com/aws/aws-sdk-java-v2/pull/4560. This PR needs that change for unit and integ tests to pass.

## Testing

Picked up https://github.com/aws/aws-sdk-java-v2/pull/4560 locally and successfully ran unit and integ tests in the s3 module. 

## Types of changes

## Checklist

## License


services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/signing/DefaultS3Presigner.java
```diff
@@ -330,7 +344,15 @@ private <T extends PresignedRequest.Builder, U> T presign(T presignedRequest,
                                                               Function<U, SdkHttpFullRequest> requestMarshaller,
                                                               String operationName) {
 
-        ExecutionContext execCtx = invokeInterceptorsAndCreateExecutionContext(presignRequest, requestToPresign, operationName);
+        // A fixed signingClock is used, so that the current time used by the signing logic, as well as to determine expiration
+        // are the same.
+        Instant signingInstant = Instant.now();
+        Clock signingClock = Clock.fixed(signingInstant, ZoneOffset.UTC);
```
gosar, line 350: An alternative to this would be to let the HttpSigner use current time inside it's logic, and later to calculate `expiration` to set inside presignedRequest in `initializePresignedRequest` we could extract the current time value used by the HttpSigner by reading the value out of `X-Amz-Date` query param inserted by the HttpSigner. `initializePresignedRequest` already has some logic to extract from `X-Amz-SignedHeaders` query param. But it seemed better to not rely more on implementation details of the HttpSigner.

services/s3/src/test/java/software/amazon/awssdk/services/s3/S3PresignerTest.java
```diff
@@ -227,6 +228,7 @@ public void getObject_Sigv4PresignerHonorsSignatureDuration() {
 
         assertThat(presigned.httpRequest().rawQueryParameters().get("X-Amz-Expires").get(0)).satisfies(expires -> {
             assertThat(expires).containsOnlyDigits();
+            // TODO(sra-identity-auth): This should be isEqualTo(1234)?
```
gosar, line 231: I think it's a bug if isCloseTo was needed for some reason earlier instead of isEqualTo check. Not sure why it was so from the beginning, but couldn't change it right now because of derived attributes behavior for PRESIGNER_EXPIRATION. I have an internal discussion thread about it. Will update that separately.

haydenbaker (approved):

services/s3/src/test/java/software/amazon/awssdk/services/s3/S3PresignerTest.java
gosar: Addressing the derived attributes and isCloseTo in https://github.com/aws/aws-sdk-java-v2/pull/4602

PR #4572, opened 2023-10-10
Add tests to override configuration with plugins

sugmanue:
## Motivation and Context

Changes to make sure that all overrides using the client builder can also be applied using plugins.

## Modifications

This PR changes the way we deal with the `ClientOverrideConfiguration`. Instead of rebuilding it from its side effect (i.e., the values in `SdkClientConfiguration`) we use the override configuration object that is present during client building or at request execution time. This will reflect the exact values given for this object without having any "magic" derived values to appear. For instance, the execution interceptors in the override configuration are added to the same client option `EXECUTION_INTERCEPTORS` but this is also filled up with default interceptors [here](https://github.com/aws/aws-sdk-java-v2/blob/9a0d9b896701e30018ae8e3934d7dd7e32997d3d/core/aws-core/src/main/java/software/amazon/awssdk/awscore/client/builder/AwsDefaultClientBuilder.java#L203) but those are not part of the original override configuration but before this change would be present in the `SdkServiceClientConfiguration.Builder` object since those were copied blindly.

We also made changes to the way we set the values to avoid overwriting previously added or computed ones, this means the if a list is given (e.g., with execution interceptors) we will follow a conservative logic to avoid duplicating previously added or removing computed values. For instance:

```java 
    static <T> void setClientListOption(SdkClientConfiguration.Builder builder, ClientOption<List<T>> option, List<T> newValue) {
        if (newValue == null || newValue.isEmpty()) {
            return;
        }
        List<T> oldValue = builder.option(option);
        if (oldValue == null || oldValue.isEmpty()) {
            builder.option(option, newValue);
            return;
        }
        // Make sure that we don't override derived values or duplicate existing ones.
        List<T> result = new ArrayList<>(oldValue);
        Set<T> dedup = new HashSet<>();
        dedup.addAll(oldValue);
        for (T value : newValue) {
            if (!dedup.contains(value)) {
                result.add(value);
            }
        }
        builder.option(option, result);
    }
```   

## Types of changes

## Checklist

## License


codegen/src/main/resources/software/amazon/awssdk/codegen/poet/model/SdkClientConfigurationUtil.resource.java
```diff
@@ -0,0 +1,103 @@
...
+import software.amazon.awssdk.core.client.config.SdkAdvancedClientOption;
+import software.amazon.awssdk.core.client.config.SdkClientConfiguration;
+import software.amazon.awssdk.core.client.config.SdkClientOption;
+import software.amazon.awssdk.core.internal.SdkInternalTestAdvancedClientOption;
+import software.amazon.awssdk.core.signer.Signer;
+import software.amazon.awssdk.profiles.ProfileFile;
+import software.amazon.awssdk.profiles.ProfileFileSupplier;
+
+@SdkInternalApi
+public final class SdkClientConfigurationUtil {
```
sugmanue, line 19: Renamed from `SdkClientConfigurationUtil.java.resource` to `SdkClientConfigurationUtil.resource.java` to make it easier for editors to identify it as a Java file. Also removed the method `copyConfigurationToOverrides` which is not longer used.

test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/CodegenServiceClientConfigurationTest.java
```diff
@@ -82,7 +82,9 @@ <T> void externalInternalTransforms_preserves_propertyValues(TestCase<T> testCas
             ProtocolRestJsonServiceClientConfigurationBuilder.builder(clientConfig.toBuilder());
 
         // Assert that we can retrieve the same value
-        assertThat(testCase.getter.apply(anotherBuilder)).isEqualTo(testCase.value);
+        if (testCase.hasDirectMapping) {
```
sugmanue, line 85: We not longer "mirror" the client override configuration from the client configuration, therefore this test no longer works for those properties.

build-tools/src/main/resources/software/amazon/awssdk/spotbugs-suppressions.xml
```diff
@@ -197,6 +197,12 @@
         <Bug pattern="RCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE"/>
     </Match>
 
+    <!-- See SpotBugs bug: https://github.com/spotbugs/spotbugs/issues/600, https://github.com/spotbugs/spotbugs/issues/756 -->
+    <Match>
+        <Class name="software.amazon.awssdk.core.SdkServiceClientConfiguration$Builder"/>
```
sugmanue, line 202: SpotBugs gets confused about the nullability of the `overrideConfiguration`, probably because is looking at the default method that throws an `UnsupportedOperationException`. Adding this since I'm certain that this value **can** be null.

cenedhryn (approved):

PR #4620, opened 2023-10-20
Use a backing attribute for ResolvedChecksumSpecs

haydenbaker:
## Modifications
- Update `RESOLVED_CHECKSUM_SPECS` execution-attribute to be backed by an internal execution attribute, so that it doesn't need to rely on lossless conversion
- Remove bugged code that relies on `AtomicReference`

## Types of changes

## Checklist

## License

### DIFF
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/interceptor/ExecutionAttribute.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/interceptor/ExecutionAttribute.java
@@ -304,7 +304,7 @@ public final class ExecutionAttribute<T> {
         @SuppressWarnings("unchecked") // Safe because of the implementation of set
         @Override
         public void set(Map<ExecutionAttribute<?>, Object> attributes, T value) {
-            attributes.compute(backingAttributeSupplier.get(), (k, attr) -> value);
+            attributes.put(backingAttributeSupplier.get(), value);
             attributes.compute(attributeSupplier.get(), (k, attr) -> writeMapping.apply((U) attr, value));
         }
 
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/interceptor/ExecutionAttribute.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/interceptor/ExecutionAttribute.java
@@ -19,7 +19,6 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
-import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.BiFunction;
 import java.util.function.Function;
 import java.util.function.Supplier;
@@ -92,21 +91,21 @@ public final class ExecutionAttribute<T> {
     }
 
     /**
-     * Create an execution attribute whose value get mapped to another execution attribute.
+     * Create an execution attribute whose value is backed by another attribute, and gets mapped to another execution attribute.
      *
-     * <p>Whenever this value is read, its value is read from the given attribute, but whenever this value is written its
+     * <p>Whenever this value is read, its value is read from the backing attribute, but whenever this value is written its
      * value is written to the given attribute AND the mapped attribute.
      *
-     * <p>This is useful when you have some attribute that depends on the value of another.
+     * <p>This is useful when you have a complex attribute relationship, where certain attributes may depend on other attributes.
      *
      * @param name The name of the attribute to create
-     * @param attributeType The type of the attribute being created
-     * @param realAttributeSupplier The supplier for the mapped attribute which is mapped from this attribute
+     * @param backingAttributeSupplier The supplier for the backing attribute, which this attribute is backed by
+     * @param attributeSupplier The supplier for the attribute which is mapped from the backing attribute
      */
     protected static <T, U> MappedAttributeBuilder<T, U> mappedBuilder(String name,
-                                                                    @SuppressWarnings("unused") Class<T> attributeType,
-                                                                    Supplier<ExecutionAttribute<U>> realAttributeSupplier) {
-        return new MappedAttributeBuilder<>(name, realAttributeSupplier);
+                                                                       Supplier<ExecutionAttribute<T>> backingAttributeSupplier,
+                                                                       Supplier<ExecutionAttribute<U>> attributeSupplier) {
+        return new MappedAttributeBuilder<>(name, backingAttributeSupplier, attributeSupplier);
     }
 
     private void ensureUnique() {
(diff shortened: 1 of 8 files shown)


core/sdk-core/src/main/java/software/amazon/awssdk/core/interceptor/ExecutionAttribute.java
```diff
@@ -276,61 +275,74 @@ public void setIfAbsent(Map<ExecutionAttribute<?>, Object> attributes, T value)
...
+                (U) attributes.get(attributeSupplier.get())
+            );
         }
 
         @SuppressWarnings("unchecked") // Safe because of the implementation of set
         @Override
         public void set(Map<ExecutionAttribute<?>, Object> attributes, T value) {
-            previousValue.set(value);
-            attributes.compute(realAttributeSupplier.get(), (k, real) -> writeMapping.apply((U) real, value));
+            attributes.compute(backingAttributeSupplier.get(), (k, attr) -> value);
```
millems, line 307: Nit: Is this just `attributes.put(backingAttributeSupplier.get(), value);`

haydenbaker: Yeah.

core/sdk-core/src/main/java/software/amazon/awssdk/core/interceptor/SdkExecutionAttribute.java
```diff
@@ -152,9 +153,19 @@ public class SdkExecutionAttribute {
...
+        Algorithm.CRC32, CRC32,
+        Algorithm.CRC32C, CRC32C
+    );
+
     protected SdkExecutionAttribute() {
     }
 
+    /**
+     * Map from the SelectedAuthScheme and the backing ChecksumSpecs value to a new value for ChecksumSpecs.
+     */
```
haydenbaker, line 168: Right now, with this mapping, updates to the signer property reflect in the `ChecksumSpecs` attribute.  This means that if the `ChecksumAlgorithm` property gets set to null, the `ChecksumSpecs` gets set to null as well. If the property gets set to some algorithm, it gets reflected in `ChecksumSpecs`, but `ChecksumSpecs` also retains whatever other data it had at the time (such as `headerName`, `responseValidationAlgorithms`, etc.)

haydenbaker: Based on offline discussion, we should keep the bi-directional mapping, but update it so that the header-name also gets mapped.

core/sdk-core/src/main/java/software/amazon/awssdk/core/interceptor/SdkExecutionAttribute.java
```diff
@@ -115,6 +139,95 @@ public class SdkExecutionAttribute {
...
     }
+
+    /**
+     * Map from the SelectedAuthScheme and the backing ChecksumSpecs value to a new value for ChecksumSpecs.
+     */
+    private static <T extends Identity> ChecksumSpecs signerChecksumReadMapping(ChecksumSpecs checksumSpecs,
+                                                                                SelectedAuthScheme<T> authScheme) {
+        if (authScheme == null) {
+            return null;
+        }
```
haydenbaker, line 174: Is this the right behavior or should we return the current value (`checksumSpecs`)?

haydenbaker: Discussed offline. Returning the current value is the right behavior, changing.

core/sdk-core/src/main/java/software/amazon/awssdk/core/interceptor/SdkExecutionAttribute.java
haydenbaker: After some more testing, it seems that we can't always re-map the header-name - there are tests that assert that a checksum is placed in a certain header, and it's certainly possible a user could also be relying on this. Ideally, we'd only update the header-name when the signer property gets updated, but there's no way to know that for certain with the current code.

core/sdk-core/src/main/java/software/amazon/awssdk/core/interceptor/SdkExecutionAttribute.java
```diff
@@ -115,6 +139,94 @@ public class SdkExecutionAttribute {
     public static final ExecutionAttribute<ChecksumValidation> HTTP_RESPONSE_CHECKSUM_VALIDATION = new ExecutionAttribute<>(
         "HttpResponseChecksumValidation");
 
+    /**
+     * The backing attribute for RESOLVED_CHECKSUM_SPECS.
+     * This holds the real ChecksumSpecs value, and is used to map to the ChecksumAlgorithm signer property
+     * in the SELECTED_AUTH_SCHEME execution attribute.
+     */
+    private static final ExecutionAttribute<ChecksumSpecs> INTERNAL_RESOLVED_CHECKSUM_SPECS =
+        new ExecutionAttribute<>("InternalResolvedChecksumSpecs");
```
millems, line 148: Should this be in SdkInternalExecutionAttributes?

millems (approved):
A few comments. Nothing blocking merge, so giving the approval.

core/sdk-core/src/main/java/software/amazon/awssdk/core/interceptor/SdkExecutionAttribute.java
haydenbaker: Sure. It will have to be package-private, but that can work.

core/sdk-core/src/main/java/software/amazon/awssdk/core/interceptor/SdkExecutionAttribute.java
```diff
@@ -115,6 +139,94 @@ public class SdkExecutionAttribute {
...
+    }
+
+    /**
+     * Map from ChecksumSpecs to a SelectedAuthScheme with the CHECKSUM_ALGORITHM signer property set.
+     */
+    private static <T extends Identity> SelectedAuthScheme<?> signerChecksumWriteMapping(SelectedAuthScheme<T> authScheme,
+                                                                                         ChecksumSpecs checksumSpecs) {
+        ChecksumAlgorithm checksumAlgorithm =
+            checksumSpecs == null ? null
+                                  : CHECKSUM_ALGORITHM_MAP.getOrDefault(checksumSpecs.algorithm(), null);
```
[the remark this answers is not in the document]
haydenbaker: Well, we need to check that `checksumSpecs` isn't null, but we can just do a `get` instead of `getOrDefault`.

PR #4643, opened 2023-10-27
Add AsyncRequestBody#fromInputStream(AsyncRequestBodyFromInputStreamC…

zoewangg:
…onfiguration) to allow users to specify the max read limit on the InputStream

## Motivation and Context
Transient network error is not uncommon when customers upload objects of a larger size to S3 and those failed requests would normally succeed upon retry. However, if users use `AsyncRequestBodyFromInputStream` as the `AsyncRequestBody` and the stream is larger than the default mark value, the SDK may not be able to retry the request in the event of network issue. This is because the default read limit of the provided inputStream used for mark and reset is 128KB.

```
java.io.IOException: Resetting to invalid mark

	at java.base/java.io.BufferedInputStream.reset(BufferedInputStream.java:454)
```

Related to https://github.com/aws/aws-sdk-java/issues/427 

### Notes:

- This option is exposed in v1 via [RequestClientOptions#setReadLimit](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/RequestClientOptions.html#setReadLimit-int-). I did not include this in the ReqeustOverrideConfiguration because this only applies to `InputStream` only.

- This is technically possible to do in `RequestBody`, (`RequestBody#fromContentProvider`), that's why I didn't make the change in RequestBody

## Modifications
Add `AsyncRequestBody#fromInputStream(AsyncRequestBodyFromInputStreamConfiguration)` that allow users to configure max read limit.

## Testing
Added unit tests

## Types of changes

## Checklist

## License

### DIFF
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/async/AsyncRequestBodyFromInputStreamConfiguration.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/async/AsyncRequestBodyFromInputStreamConfiguration.java
@@ -1,0 +1,201 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.core.async;
+
+import java.io.InputStream;
+import java.util.Objects;
+import java.util.concurrent.ExecutorService;
+import software.amazon.awssdk.annotations.SdkPublicApi;
+import software.amazon.awssdk.utils.Validate;
+import software.amazon.awssdk.utils.builder.CopyableBuilder;
+import software.amazon.awssdk.utils.builder.ToCopyableBuilder;
+
+/**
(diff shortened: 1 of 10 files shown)


core/sdk-core/src/main/java/software/amazon/awssdk/core/async/AsyncRequestBodyFromInputStreamConfiguration.java
```diff
@@ -0,0 +1,200 @@
...
+import software.amazon.awssdk.utils.Validate;
+import software.amazon.awssdk.utils.builder.CopyableBuilder;
+import software.amazon.awssdk.utils.builder.ToCopyableBuilder;
+
+/**
+ * Configuration options for {@link AsyncRequestBody#fromInputStream(AsyncRequestBodyFromInputStreamConfiguration)} to configure how the SDK
+ * should create an {@link AsyncRequestBody} from an {@link InputStream}.
+ */
+@SdkPublicApi
+public final class AsyncRequestBodyFromInputStreamConfiguration implements ToCopyableBuilder<AsyncRequestBodyFromInputStreamConfiguration.Builder,
```
zoewangg, line 31: This is a bit mouthful, lmk if you have better ideas.

core/sdk-core/src/main/java/software/amazon/awssdk/core/async/AsyncRequestBodyFromInputStreamConfiguration.java
```diff
@@ -0,0 +1,200 @@
...
+         * @return This object for method chaining.
+         */
+        Builder executor(ExecutorService executor);
+
+        /**
+         * Configures max read limit used to mark and reset the {@link InputStream}. This will have no
+         * effect if the stream doesn't support mark and reset.
+         *
+         * <p>
+         * By default, it is 128 KiB.
```
zoewangg, line 149: Alternatively, we could make it contentLength + 1 so that users don't have to configure it by themselves, however, this would mean the SDK will buffer all the data, so I'm reluctant to make this the default behavior.

davidh44 (approved):

PR #4651, opened 2023-10-31
Add tests to assert signer and endpoint override behaviors

haydenbaker:
## Types of changes

## Checklist

## License


test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/SignerAndEndpointOverridesTest.java
```diff
@@ -0,0 +1,228 @@
...
+            .credentialsProvider(StaticCredentialsProvider.create(AwsBasicCredentials.create("akid", "skid")))
+            .endpointProvider(v4EndpointProviderOverride())
+            .region(Region.US_WEST_2)
+            .overrideConfiguration(o -> o.putAdvancedOption(SIGNER, mockSigner))
+            .build();
+
+        try {
+            client.streamingInputOperation(r -> {}, RequestBody.fromString("test"));
+        } catch (Exception expected) {
+        }
```
sugmanue, line 76: Can we switch instead to use an interceptor and assert that the exception is the expected one?, See for instance the setup in the [tests here](https://github.com/aws/aws-sdk-java-v2/blob/dbc7909b36acc3e92595f7e01e1d1cdd37da92b6/test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/OverrideConfigurationPluginsTest.java#L38). Rationale, this ensures that the exception thrown is exactly what we expect instead of something else failing in the middle.

For instance, the assertion will be instead

```
    assertThatThrownBy(() -> client.streamingInputOperation(r -> { }, RequestBody.fromString("test")))
            .hasMessageContaining("boom!");
```

test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/SignerAndEndpointOverridesTest.java
```diff
@@ -0,0 +1,228 @@
...
+
+        verify(mockSigner).sign(
+            any(SdkHttpFullRequest.class),
+            argThat(attrs ->
+                        "us-west-2".equals(attrs.getAttribute(AwsSignerExecutionAttribute.SIGNING_REGION).id()) &&
+                        "query-test-v4".equals(attrs.getAttribute(AwsSignerExecutionAttribute.SERVICE_SIGNING_NAME)) &&
+                        !attrs.getAttribute(AwsSignerExecutionAttribute.SIGNER_DOUBLE_URL_ENCODE))
```
sugmanue, line 83: Nit, I think that being explicit with the expected value makes this test easier to read, e.g., 

```
Boolean.FALSE.equals(attrs.getAttribute(AwsSignerExecutionAttribute.SIGNER_DOUBLE_URL_ENCODE)))
```

test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/SignerAndEndpointOverridesTest.java
```diff
@@ -0,0 +1,228 @@
...
+            any(SdkHttpFullRequest.class),
+            argThat(attrs ->
+                        "us-east-1".equals(attrs.getAttribute(AwsSignerExecutionAttribute.SIGNING_REGION).id()) &&
+                        "query-test-v4a".equals(attrs.getAttribute(AwsSignerExecutionAttribute.SERVICE_SIGNING_NAME)) &&
+                        !attrs.getAttribute(AwsSignerExecutionAttribute.SIGNER_DOUBLE_URL_ENCODE))
+        );
+    }
+
+    private static ProtocolQueryAuthSchemeProvider v4AuthSchemeProviderOverride() {
+        return __ -> {
```
sugmanue, line 171: Nit, can we use a letter here instead? (e.g., `x -> {...}`)

test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/SignerAndEndpointOverridesTest.java
```diff
@@ -0,0 +1,228 @@
...
+    public Signer mockSigner = mock(Signer.class);
+
+    @Test
+    public void test_whenV4EndpointAuthSchemeWithSignerOverride_thenEndpointParamsShouldPropagateToSigner() {
+        ProtocolQueryClient client = ProtocolQueryClient
+            .builder()
+            .authSchemeProvider(v4AuthSchemeProviderOverride())
+            .credentialsProvider(StaticCredentialsProvider.create(AwsBasicCredentials.create("akid", "skid")))
+            .endpointProvider(v4EndpointProviderOverride())
+            .region(Region.US_WEST_2)
```
sugmanue, line 69: Nit, let's use here another region to make it clear that the override is working as expected 
```suggestion
            .region(Region.AP_SOUTH_1)
```

test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/SignerAndEndpointOverridesTest.java
```diff
@@ -0,0 +1,228 @@
...
+/**
+ * Tests to ensure that parameters set when endpoint and auth-scheme resolution occurs get propagated to the overriden
+ * signer (i.e. pre-SRA signers).
+ */
+public class SignerAndEndpointOverridesTest {
+
+    public Signer mockSigner = mock(Signer.class);
+
+    @Test
+    public void test_whenV4EndpointAuthSchemeWithSignerOverride_thenEndpointParamsShouldPropagateToSigner() {
```
sugmanue, line 63: Nit, SonarCloud have complained before about the test methods being public

```suggestion
    void test_whenV4EndpointAuthSchemeWithSignerOverride_thenEndpointParamsShouldPropagateToSigner() {
```

test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/SignerAndEndpointOverridesTest.java
haydenbaker: It's unused, and this is a fairly common when the variable is unused.

sugmanue: Not in Java, and I can't find any other test in the project using it (didn't search thoroughly), but up to the team.

sugmanue: (c.f., the commonly used pattern on some other languages `_` is not even possible in Java since a single underscore is reserved).

test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/SignerAndEndpointOverridesTest.java
haydenbaker: Point taken. We don't need the interceptor, we can just use the mocked signer to do that.

test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/SignerAndEndpointOverridesTest.java
haydenbaker: Yep, hence the usage of two underscores. Changed to 'x'.

test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/SignerAndEndpointOverridesTest.java
```diff
@@ -0,0 +1,241 @@
...
+ * Tests to ensure that parameters set when endpoint and auth-scheme resolution occurs get propagated to the overriden
+ * signer (i.e. pre-SRA signers).
+ */
+public class SignerAndEndpointOverridesTest {
+
+    private static Signer mockSigner = mock(Signer.class);
+
+    @BeforeAll
+    static void setup() {
+        when(mockSigner.sign(any(), any())).thenThrow(new RuntimeException("boom!"));
```
sugmanue, line 69: The idea behind using the interceptor instead is two folded, it's the pattern used by the Java team and I been asked before to use it, and, the exceptions is thrown **before transmission**, IOW, you also validate that everything else in the execution pipeline runs as expected and it's not throwing somewhere else due to the setup you have here.

test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/SignerAndEndpointOverridesTest.java
```diff
@@ -0,0 +1,241 @@
...
+    }
+
+    @Test
+    void test_whenV4EndpointAuthSchemeWithSignerOverride_thenEndpointParamsShouldPropagateToSigner() {
+        ProtocolQueryClient client = ProtocolQueryClient
+            .builder()
+            .authSchemeProvider(v4AuthSchemeProviderOverride())
+            .credentialsProvider(StaticCredentialsProvider.create(AwsBasicCredentials.create("akid", "skid")))
+            .endpointProvider(v4EndpointProviderOverride())
+            .region(Region.US_WEST_1)
```
sugmanue, line 79: The idea behind the ask to change the region here was that you can validate that the override takes place, given that the region here is the same as the override we cannot know whether is it coming from the client builder or from the override.

test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/SignerAndEndpointOverridesTest.java
sugmanue: You forgot this one and the other below.

core/aws-core/src/main/java/software/amazon/awssdk/awscore/internal/AwsExecutionContextBuilder.java
```diff
@@ -183,21 +184,32 @@ private static void putAuthSchemeResolutionAttributes(ExecutionAttributes execut
                                                           SdkClientConfiguration clientConfig,
                                                           SdkRequest originalRequest) {
 
-        // TODO(sra-identity-and-auth): When request-level auth scheme provider is added, use the request-level auth scheme
-        //  provider if the customer specified an override, otherwise fall back to the one on the client.
-        AuthSchemeProvider authSchemeProvider = clientConfig.option(SdkClientOption.AUTH_SCHEME_PROVIDER);
+        // Use auth scheme provider that the user specified at the request level with
+        // preference over that which is configured on the client.
+        AuthSchemeProvider authSchemeProvider =
+            executionAttributes.getOptionalAttribute(SdkInternalExecutionAttribute.AUTH_SCHEME_RESOLVER)
```
gosar, line 190: Directly setting AUTH_SCHEME_RESOLVER into ExecutionAttributes is not how users would specify request level override for it. These attributes are Sdk*Internal*ExecutionAttributes (and SdkProtectedApi) which customers are not supposed to use directly. Even if they can set ExecutionAttributes in RequestOverrideConfiguration. To resolve this TODO we need to support a first class setter somewhere in RequestOverrideConfiguration(or sub-class) for each of these properties.

We have other internal tasks tracking these TODOs, let's not try to address those in the context of this PR.

I think you made this change to have the tests below with sigv4a. The supported way to do that is `putAuthScheme` on the clientBuilder. (also note that's also at client level not request level which is what this TODO is about).

sugmanue:
This PR is stale and not is not longer being work on.

PR #4655, opened 2023-10-31
Enable auto reads on idling connections

dagnir:
## Motivation and Context
This change enables auto reads on idling connections, and disables them on in-use connections (no auto read for in-use connections is existing behavior). Auto reads while the channel is idling allows the channel to read and immediately handle events that are triggered by the remote end such as TLS close notifies. In particular this fixes issues where a channel is leased only for the SDK to find that the remote end has initiated a shutdown of the TLS connection.

## Types of changes

## Checklist

## License

### DIFF
--- a/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/AutoReadDisableChannelPoolListener.java
+++ b/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/AutoReadDisableChannelPoolListener.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.http.nio.netty.internal;
+
+import io.netty.channel.Channel;
+
+/**
+ * Disables auto read on in-use channels to allow upper layers to take care of flow control.
+ */
+public class AutoReadDisableChannelPoolListener implements ListenerInvokingChannelPool.ChannelPoolListener {
+    private static final AutoReadDisableChannelPoolListener INSTANCE = new AutoReadDisableChannelPoolListener();
+
+    @Override
+    public void channelAcquired(Channel channel) {
+        channel.config().setAutoRead(false);
+    }
+
+    public static AutoReadDisableChannelPoolListener create() {
+        return INSTANCE;
+    }
+}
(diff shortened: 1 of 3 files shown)


http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/AwaitCloseChannelPoolMap.java
```diff
@@ -239,12 +241,19 @@ private SdkChannelPool wrapBaseChannelPool(Bootstrap bootstrap, ChannelPool chan
...
+
             // Add a listener that ensures acquired channels are marked IN_USE and thus not eligible for certain idle timeouts.
             InUseTrackingChannelPoolListener.create(),
 
             // Add a listener that removes request-specific handlers with each request.
-            HandlerRemovingChannelPoolListener.create()
+            HandlerRemovingChannelPoolListener.create(),
+
+            // Add a listner that enables auto reads on released connections.
+            AutoReadEnableChannelPoolListener.create()
```
dagnir, line 256: I'm not sure breaking up the acquire/release implementations this way is strictly necessary; however it makes sense to me that the the very first thing we need to do is disable auto reads on channel acquire, but auto reads should only be enabled as the very last thing we do before releasing the channel back to the pool.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/AwaitCloseChannelPoolMap.java
```diff
@@ -239,12 +239,19 @@ private SdkChannelPool wrapBaseChannelPool(Bootstrap bootstrap, ChannelPool chan
                                                                    configuration.maxConnections(),
                                                                    configuration);
 
+
         sdkChannelPool = new ListenerInvokingChannelPool(bootstrap.config().group(), sdkChannelPool, Arrays.asList(
+            // Add a listener that disables auto reads on acquired connections.
+            AutoReadDisableChannelPoolListener.create(),
```
zoewangg, line 245: Should we remove https://github.com/aws/aws-sdk-java-v2/blob/f5c0b21231821b991c5678a63984fa612f6a122d/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/NettyRequestExecutor.java#L202 now that we have this new listener?

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/AutoReadDisableChannelPoolListener.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+package software.amazon.awssdk.http.nio.netty.internal;
+
+import io.netty.channel.Channel;
+import software.amazon.awssdk.annotations.SdkInternalApi;
+
+/**
+ * Disables auto read on in-use channels to allow upper layers to take care of flow control.
+ */
```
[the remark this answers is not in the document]
dagnir: +1

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/AwaitCloseChannelPoolMap.java
dagnir: Ah yeah good point. Sure I can remove this

zoewangg (approved):

PR #4649, opened 2023-10-31
Updated AttributeMap to not close ExecutorService on Java 21

michaeldimchuk:
As of Java 21, [ExecutorService](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/concurrent/ExecutorService.html) extends `AutoCloseable`, so `AttributeMap#close` needs to be updated to avoid closing the executor and potentially cause a deadlock.

## Motivation and Context
Starting with Java 21, `ExecutorService` extends the `AutoCloseable` interface, which causes `AttributeMap#close` to pick up executor instances and try and close them, which can create deadlocks and cause the application to hang.

Issue: https://github.com/aws/aws-sdk-java-v2/issues/4395

## Modifications
Updated the `AttributeMap#close` method to first check if the attribute is an ExecutorService, and shut it down if so, otherwise check if the object is closeable and close it if needed.

## Testing
- Added a test case to ensure that ExecutorService instances are explicitly shut down and not closed
- Added a test case that running `AttributeMap#close` on a thread owned by an ExecutorService the `AttributeMap` instance manages does not result in a deadlock
- Ran all the tests to verify that the changes don't break anything, tested on an M1 MacBook Air with Java 17
- Validated that the new code does not introduce any new issues and maintains the same expected behavior as before

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)

## Checklist
- [x] I have read the [CONTRIBUTING](https://github.com/aws/aws-sdk-java-v2/blob/master/CONTRIBUTING.md) document
- [ ] Local run of `mvn install` succeeds
- [x] My code follows the code style of this project
- [ ] My change requires a change to the Javadoc documentation
- [ ] I have updated the Javadoc documentation accordingly
- [x] I have added tests to cover my changes
- [x] All new and existing tests passed
- [x] I have added a changelog entry. Adding a new entry must be accomplished by running the `scripts/new-change` script and following the instructions. Commit the new file created by the script in `.changes/next-release` with your changes.
- [ ] My change is to implement 1.11 parity feature and I have updated [LaunchChangelog](https://github.com/aws/aws-sdk-java-v2/blob/master/docs/LaunchChangelog.md)

## License
- [x] I confirm that this pull request can be released under the Apache 2 license

PR #4659, opened 2023-11-01
(1 of ?) Improve plugin's ability to affect configuration.

millems:
A few limitations exist in plugin's ability to affect the SDK's configuration:
1. When plugins modify the configuration at the request-level, it does not affect configuration that is "derived" from the configuration that was modified (e.g. changing the profile file does not change the region, even if the region was derived from the profile file originally).
2. Plugins do not see the default configuration applied by the client.
3. Plugins do not see and cannot modify configuration applied by customers at the request level.

This PR fixes (1), except for credentials derived from profile files. Future PRs will address the remaining limitations. As a result, this PR includes many disabled and commented-out tests that will be re-enabled as these limitations are corrected.

The particularly interesting/risky changes are in AwsDefaultClientBuilder, SdkDefaultClientBuilder, and AttributeMap. Major new testing of functionality is in AttributeMapTest and SdkPluginTest.

This PR makes the following large changes:
1. Add support for "lazy" SDK client options. These options can be set, but not evaluated until they are read. This functionality was added in AttributeMap, which SdkClientConfiguration uses.

This PR makes the following small changes and bug fixes:
1. Add support for equals and hashCode in the default endpoint provider.
2. Fixed a bug where scheduled executors specified by plugins were not treated as unmanaged.
3. Reduced the number of times that the profile file was read during client creation.
4. Add support for specifying a profile file supplier in the ClientOverrideConfiguration.
5. Changed the way "child" classes configure the AttributeMap used to create the HTTP client to match the way other configuration works.


sugmanue (approved):

utils/src/main/java/software/amazon/awssdk/utils/AttributeMap.java
```diff
@@ -199,17 +266,291 @@ public <T> Builder put(Key<T> key, T value) {
...
+            }
+
+            Queue<Value<?>> unloadQueue = new ArrayDeque<>(dependents);
+            dependents.clear();
+            while (!unloadQueue.isEmpty()) {
+                Value<?> toUnload = unloadQueue.poll();
+
+                Set<Value<?>> toUnloadDependents = dependents(toUnload.key());
+                unloadQueue.addAll(toUnloadDependents);
+                toUnloadDependents.clear();
```
dagnir, line 385: Why does this get cleared here?

utils/src/main/java/software/amazon/awssdk/utils/AttributeMap.java
```diff
@@ -199,17 +266,291 @@ public <T> Builder put(Key<T> key, T value) {
...
+        }
+
+        @Override
+        public void clearCache() {
+        }
+
+        @Override
+        public void close() {
+            IoUtils.closeIfCloseable(value, null);
+            shutdownIfExecutorService(value);
```
dagnir, line 455: Related to #4649, this won't work in Java 21

utils/src/main/java/software/amazon/awssdk/utils/AttributeMap.java
```diff
@@ -199,17 +266,291 @@ public <T> Builder put(Key<T> key, T value) {
...
+    @FunctionalInterface
+    public interface LazyValueSource {
+        <T> T get(Key<T> sourceKey);
+    }
+
+    /**
+     * Tracks which values "depend on" other values, so that when we update one value, when can clear the cache of any other
+     * values that were derived from the value that was updated.
+     */
+    private static final class DependencyGraph {
```
dagnir, line 344: Neat!

utils/src/main/java/software/amazon/awssdk/utils/AttributeMap.java
```diff
@@ -199,17 +266,291 @@ public <T> Builder put(Key<T> key, T value) {
...
+            if (dependents.isEmpty()) {
+                return;
+            }
+
+            Queue<Value<?>> unloadQueue = new ArrayDeque<>(dependents);
+            dependents.clear();
+            while (!unloadQueue.isEmpty()) {
+                Value<?> toUnload = unloadQueue.poll();
+
+                Set<Value<?>> toUnloadDependents = dependents(toUnload.key());
```
dagnir, line 383: Is it worth optimizing here by avoiding unloading values more than once?

core/sdk-core/src/main/java/software/amazon/awssdk/core/client/builder/SdkDefaultClientBuilder.java
```diff
@@ -217,14 +225,14 @@ protected final SdkClientConfiguration asyncClientConfiguration() {
         configuration = mergeChildDefaults(configuration);
         configuration = mergeGlobalDefaults(configuration);
 
-        // Invoke the plugins after defaults and before finalizing the configuration.
-        configuration = invokePlugins(configuration);
-
-        // Create additional configuration from the default-applied configuration
+        // Finalize configuration after plugins are applied
```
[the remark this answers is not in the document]
millems: Fixed in https://github.com/aws/aws-sdk-java-v2/pull/4686

http-client-spi/src/main/java/software/amazon/awssdk/http/SdkHttpClient.java
```diff
@@ -58,7 +58,6 @@ default String clientName() {
     /**
      * Interface for creating an {@link SdkHttpClient} with service specific defaults applied.
      */
-    @FunctionalInterface
```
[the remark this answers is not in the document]
millems: I did it with purpose, and don't remember why anymore. Fixed in https://github.com/aws/aws-sdk-java-v2/pull/4686

utils/src/main/java/software/amazon/awssdk/utils/AttributeMap.java
```diff
@@ -69,9 +88,15 @@ public <T> T get(Key<T> key) {
...
      */
     public AttributeMap merge(AttributeMap lowerPrecedence) {
-        Map<Key<?>, Object> copiedConfiguration = new HashMap<>(attributes);
-        lowerPrecedence.attributes.forEach(copiedConfiguration::putIfAbsent);
-        return new AttributeMap(copiedConfiguration);
+        Builder resultBuilder = new AttributeMap.Builder(this);
+        lowerPrecedence.attributes.forEach((k, v) -> {
+            resultBuilder.internalPutIfAbsent(k, () -> {
+                Value<?> result = v.copy();
+                result.clearCache();
```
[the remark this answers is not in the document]
millems: When we do `attributeMapBuilder.build().toBuilder().someModification(...)`, we `copy()` every value so that the modifications to the value don't affect the `attributes` of the past builder or attributeMap, but we want to preserve the cached values, because they're still valid.

utils/src/main/java/software/amazon/awssdk/utils/AttributeMap.java
```diff
@@ -84,31 +109,30 @@ public AttributeMap copy() {
...
+        private final Function<Object, T> convertMethod;
 
         protected Key(Class<T> valueType) {
             this.valueType = valueType;
+            this.convertMethod = valueType::cast;
         }
 
         protected Key(UnsafeValueType unsafeValueType) {
             this.valueType = unsafeValueType.valueType;
+            this.convertMethod = v -> (T) v; // 🙏
```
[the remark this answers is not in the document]
millems: 🤔

utils/src/main/java/software/amazon/awssdk/utils/AttributeMap.java
millems: After we have cleared the cache for a value, the things that it depends on might change when it reloads. If we didn't clear this, it means that a value's old dependencies still govern its cache, even if it might not depend on those values.

Example scenario: Pretend the retry mode depends on the defaults mode, and because the defaults mode is "in-region" it depends on the region. (RetryMode -> DefaultsMode -> Region)
1. We modify the Region, so we clear the cache for RetryMode and DefaultsMode
2. We change the DefaultsMode from in-region to standard
3. We build the attribute map, so we derive and re-cache the values for RetryMode and DefaultsMode

If we didn't do `toUnloadDependents.clear()` the dependency graph would still say that DefaultsMode depends on Region, even though we changed the defaults mode to standard and it no longer has a dependency on region. That means when we changed region, we'd still reset the cache of DefaultsMode, even though it doesn't need to be reset.

utils/src/main/java/software/amazon/awssdk/utils/AttributeMap.java
millems: Fixed in https://github.com/aws/aws-sdk-java-v2/pull/4686

utils/src/main/java/software/amazon/awssdk/utils/AttributeMap.java
millems: I don't think so. Unloading the value is pretty much free, since we're just clearing the cache and not actually refreshing the value until it's requested again.

utils/src/main/java/software/amazon/awssdk/utils/AttributeMap.java
dagnir: Ah I see. I didn't consider that it could be that dynamic!

PR #4695, opened 2023-11-13
Add EKS container host in ContainerCredentialsProvider , also validate the resolved endpoint for RelativeURI

joviegas:
## Motivation and Context

- Add support for EKS host and auth token file, EKS: IRSAv2 auth
- Also validation is updated to validate the loopBacks even for Relative URI

## Modifications

- EKS hosts IPV4 and IPV6 added in the allow hosts
- Validation of Endpoint is done as soon as final endpoint is resolved

## Testing

- Added new Test cases from specifications

## Types of changes

## License

### DIFF
--- a/core/auth/src/main/java/software/amazon/awssdk/auth/credentials/ContainerCredentialsProvider.java
+++ b/core/auth/src/main/java/software/amazon/awssdk/auth/credentials/ContainerCredentialsProvider.java
@@ -15,16 +15,23 @@
 
 package software.amazon.awssdk.auth.credentials;
 
+import static java.nio.charset.StandardCharsets.UTF_8;
+
 import java.io.IOException;
 import java.net.InetAddress;
 import java.net.URI;
 import java.net.UnknownHostException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.time.Instant;
 import java.time.temporal.ChronoUnit;
 import java.util.Arrays;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Objects;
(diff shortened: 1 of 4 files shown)


core/auth/src/main/java/software/amazon/awssdk/auth/credentials/ContainerCredentialsProvider.java
```diff
@@ -199,19 +215,45 @@ public ResourcesEndpointRetryPolicy retryPolicy() {
...
+                                                                          .map(this::readToken);
+
+        }
+
+        private String readToken(String filePath) {
+            Path path = Paths.get(filePath);
+            if (!Files.exists(path)) {
+                throw SdkClientException.create(
+                    String.format("Failed to read authorization token from '%s': no such file or directory", filePath));
+            }
```
dagnir, line 243: I don't think this check is necessary; it's also not atomic since the file could exist during this check but not exist when we try to run it. Can we just to `Files.readAllBytes` below and catch the exception for file not exists?

joviegas: The Specs mentioned 
```
      "reason": "failed to read authorization token from '/full/path/to/token/file': no such file or directory"
  ```
  
  Thus I had added this specific case.

But , I think we can let user use the sdkException.getCause() which will give NoSuchFileException. Updated the code

dagnir: Gotcha. I think without this change, we will still meet that requirement

dagnir (approved):

Issue #4697, opened 2023-11-14
Signature created with S3Presigner changed in 2.21.16+ when used with endpointOverride and http address


WIStudent:
### Describe the bug

We are using minio on our local dev machines and noticed that the signature of presigned URLs has changed since version 2.21.16 when used with `endpointOverride` and an http address. Minio currently rejects this new signature.

We are using the following code to generate a presigned URL:

```java
Consumer<GetObjectPresignRequest.Builder> buildPresignRequestFn = builder -> builder
  .signatureDuration(Duration.ofHours(1))
  .getObjectRequest(b -> b.bucket("bucket").key("given/key"));

final var awsCredentialsProvider = StaticCredentialsProvider.create(AwsBasicCredentials.create("minioadmin", "minioadmin"));
final var presigner = S3Presigner.builder()
  .credentialsProvider(awsCredentialsProvider)
  .region(Region.US_EAST_1)
  .endpointOverride(new URI("http://localhost:8082"))
  .build();

final var presignedUrl = presigner.presignGetObject(buildPresignRequestFn).url();
```

Forcing the `X-Amz-Date` to `19700101T000000Z` using mocks results in the signature `501491751374ffbf1170c5808aaf827f51d6cc070ac4a6776fbbf87c7ce7a160` when used with 2.21.15 or older and `2683d70a5695b830615bd63e99a3cfa991596998576f0be9a010663493eeb1d0` when used with 2.21.16 or newer.

### Expected Behavior

Versions <= 2.21.15 and >= 2.21.16 should produce the same signature.

### Current Behavior

Versions <= 2.21.15 and >= 2.21.16 produce different signatures for http endpoints.

### Reproduction Steps

```java
import org.junit.jupiter.api.Test;
import org.mockito.Mockito;
import software.amazon.awssdk.auth.credentials.AwsBasicCredentials;
import software.amazon.awssdk.auth.credentials.StaticCredentialsProvider;
import software.amazon.awssdk.regions.Region;
import software.amazon.awssdk.services.s3.presigner.S3Presigner;
import software.amazon.awssdk.services.s3.presigner.model.GetObjectPresignRequest;

import java.net.URI;
import java.net.URISyntaxException;
import java.time.Clock;
import java.time.Duration;
import java.time.Instant;
import java.util.function.Consumer;

import static org.assertj.core.api.Assertions.assertThat;
import static org.mockito.Mockito.*;

class NewSignatureTest {

  @Test
[...]

PR #4701, opened 2023-11-14
Update signer payload-signing conditional to honor PAYLOAD_SIGNING_ENABLED when payload is null

haydenbaker:
## Motivation and Context
- Prior to SRA, certain signers (i.e. `Aws4UnsignedPayloadSigner`) enforce usage of payload-signing when HTTP (not HTTPS) is used
- SRA tries to keep this behavior, but there is no way to force the signer to do unsigned-payload with HTTP (this PR changes this)
- Presigned requests will not have a payload, so `UNSIGNED-PAYLOAD` should always be used in this case.
- Addresses #4697

## Modifications
- Updates the conditionals for checking whether to sign the payload or not
  - If `PAYLOAD_SIGNING_ENABLED` is set to false, we will honor that if the payload is not present (such as in the case of presigned requests), otherwise, we will still sign the payload (for security)

## Testing
- Ran local integration tests for s3, all passed
- Ran unit tests for all signers and s3

## Types of changes

## Checklist

## License

### DIFF
--- a/core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/crt/internal/signer/DefaultAwsCrtV4aHttpSigner.java
+++ b/core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/crt/internal/signer/DefaultAwsCrtV4aHttpSigner.java
@@ -53,6 +53,7 @@ import software.amazon.awssdk.http.auth.spi.signer.SignRequest;
 import software.amazon.awssdk.http.auth.spi.signer.SignedRequest;
 import software.amazon.awssdk.identity.spi.AwsCredentialsIdentity;
 import software.amazon.awssdk.utils.CompletableFutureUtils;
+import software.amazon.awssdk.utils.Logger;
 
 /**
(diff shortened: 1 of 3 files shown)


sugmanue (approved):

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/crt/internal/signer/DefaultAwsCrtV4aHttpSigner.java
```diff
@@ -174,7 +174,11 @@ private static boolean isPayloadSigning(BaseSignRequest<?, ? extends AwsCredenti
         boolean isPayloadSigningEnabled = request.requireProperty(PAYLOAD_SIGNING_ENABLED, true);
         boolean isEncrypted = "https".equals(request.request().protocol());
 
-        return !isAnonymous && (isPayloadSigningEnabled || !isEncrypted);
+        if (isPayloadSigningEnabled) {
+            return !isAnonymous;
+        }
+
+        return !isAnonymous && !isEncrypted;
```
millems, line 181: These look like the same conditions. Under what conditions would they return different results before and after this change?

millems (approved):

PR #4726, opened 2023-11-30
fix ClientOverrideConfiguration.Builder empty headers throwing NPE

L-Applin:
the `headers()` method returns an unmodifiableMapOfLists which assumes the map it delegates to is not null. But `config.option(ADDITIONAL_HTTP_HEADERS)` returns null if the config is not set.

### DIFF
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/client/config/ClientOverrideConfiguration.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/client/config/ClientOverrideConfiguration.java
@@ -36,7 +36,6 @@ import static software.amazon.awssdk.utils.ScheduledExecutorUtils.unwrapUnmanage
 import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/client/config/ClientOverrideConfiguration.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/client/config/ClientOverrideConfiguration.java
@@ -706,7 +706,8 @@ public final class ClientOverrideConfiguration
 
         @Override
         public Map<String, List<String>> headers() {
-            Map<String, List<String>> option = Validate.getOrDefault(config.option(ADDITIONAL_HTTP_HEADERS), Collections::emptyMap);
+            Map<String, List<String>> option = Validate
+                .getOrDefault(config.option(ADDITIONAL_HTTP_HEADERS), Collections::emptyMap);
             return CollectionUtils.unmodifiableMapOfLists(option);
         }
 
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/client/config/ClientOverrideConfiguration.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/client/config/ClientOverrideConfiguration.java
@@ -706,7 +706,7 @@ public final class ClientOverrideConfiguration
 
         @Override
         public Map<String, List<String>> headers() {
-            Map<String, List<String>> option = Validate.getOrDefault(config.option(ADDITIONAL_HTTP_HEADERS), HashMap::new);
+            Map<String, List<String>> option = Validate.getOrDefault(config.option(ADDITIONAL_HTTP_HEADERS), Collections::emptyMap);
             return CollectionUtils.unmodifiableMapOfLists(option);
         }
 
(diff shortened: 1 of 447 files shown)


core/sdk-core/src/main/java/software/amazon/awssdk/core/client/config/ClientOverrideConfiguration.java
```diff
@@ -705,7 +706,8 @@ public void setHeaders(Map<String, List<String>> additionalHttpHeaders) {
 
         @Override
         public Map<String, List<String>> headers() {
-            return CollectionUtils.unmodifiableMapOfLists(config.option(ADDITIONAL_HTTP_HEADERS));
+            Map<String, List<String>> option = Validate.getOrDefault(config.option(ADDITIONAL_HTTP_HEADERS), HashMap::new);
```
sugmanue, line 709:
```suggestion
            Map<String, List<String>> option = Validate.getOrDefault(config.option(ADDITIONAL_HTTP_HEADERS), Collections::emptyMap);
```

No need to create a new object, we can use the statically defined default one. 

Can we also add a `Validate.paramNotNull(⋯)` [here](https://github.com/aws/aws-sdk-java-v2/blob/e5570f717085d451782c0c1ea0006f8ac35a15b3/utils/src/main/java/software/amazon/awssdk/utils/UnmodifiableMapOfLists.java#L43)?

millems (approved):

core/sdk-core/src/test/java/software/amazon/awssdk/core/client/config/ClientOverrideConfigurationTest.java
```diff
@@ -278,4 +278,16 @@ public void executionAttributes_listPreviouslyAdded_appendedToList() {
         ClientOverrideConfiguration overrideConfig = builder.build();
         assertThat(overrideConfig.executionAttributes().getAttribute(addedAttribute)).isEqualTo(addedValue);
     }
+
+    @Test
+    void builder_headersWithEmptyAdditionalHeaders_shouldReturnEmptyMap_shouldNotThrowNullPointerException() {
+        ClientOverrideConfiguration.Builder builder = ClientOverrideConfiguration.builder();
+        Map<String, List<String>> headers = builder.headers();
+        assertThat(headers).isEmpty(); // <--- should not throw NPE
```
sugmanue, line 286:
```suggestion
        assertThat(headers).isNotNull();
```
Feels better to add an explicit assertion of what you expect instead of leaving it to a comment.

core/sdk-core/src/test/java/software/amazon/awssdk/core/client/config/ClientOverrideConfigurationTest.java
```diff
@@ -278,4 +278,16 @@ public void executionAttributes_listPreviouslyAdded_appendedToList() {
...
+        Map<String, List<String>> newHeaders = new HashMap<>();
+        newHeaders.put("my-key", Collections.singletonList("my-value"));
+        newHeaders.putAll(headers); // <--- should not throw NPE
```
sugmanue, line 289:
```suggestion
        newHeaders.putAll(headers);
```
I'd remove this if we assert that `headers` is not null.

joviegas:
I see that `CollectionUtils.unmodifiableMapOfLists(option);`  was added by https://github.com/aws/aws-sdk-java-v2/pull/3178/ and it is pretty old.

Why are we seeing this issue now ?

core/sdk-core/src/test/java/software/amazon/awssdk/core/client/config/ClientOverrideConfigurationTest.java
L-Applin: The thing is, it's not the `headers` returned by `.headers()` that is null and causing a problem, but the map that this map returns. The NPE is thrown when the top level Map interacts with its delegate.

sugmanue: Yeah, that's why the suggestion above.

> Can we also add a `Validate.paramNotNull(⋯)` [here](https://github.com/aws/aws-sdk-java-v2/blob/e5570f717085d451782c0c1ea0006f8ac35a15b3/utils/src/main/java/software/amazon/awssdk/utils/UnmodifiableMapOfLists.java#L43)?

utils/src/main/java/software/amazon/awssdk/utils/UnmodifiableMapOfLists.java
```diff
@@ -40,6 +40,7 @@ class UnmodifiableMapOfLists<T, U> implements Map<T, List<U>>, Serializable {
     private final Map<T, List<U>> delegate;
 
     UnmodifiableMapOfLists(Map<T, List<U>> delegate) {
+        Validate.notNull(delegate, "delegate must not be null");
         this.delegate = delegate;
```
sugmanue, line 44:
```suggestion
        this.delegate = Validate.paramNotNull(delegate, "delegate");
```

sugmanue (approved):

PR #4734, opened 2023-12-02
Support package name customization

zoewangg:
## Motivation and Context
Allow codegen users to customize package prefix and suffix
- prefix: by default it's `software.amazon.awssdk.services`
- suffix: by default it's the same as service ID, eg: `[packagePrefix].s3`
Example:
```

{
    "packageName": {
        "prefix": "foo.bar",
        "suffix": "helloworld"
    }
}
```

## Testing
Added unit tests.

## Types of changes

## Checklist

## License

### DIFF
--- a/.changes/next-release/feature-AWSSDKforJavav2Codegen-1b31942.json
+++ b/.changes/next-release/feature-AWSSDKforJavav2Codegen-1b31942.json
@@ -1,0 +1,6 @@
+{
+    "type": "feature",
+    "category": "AWS SDK for Java v2 - Codegen",
+    "contributor": "",
+    "description": "Allow users to customize package prefix and suffix."
+}
--- a/codegen/src/main/java/software/amazon/awssdk/codegen/AddMetadata.java
+++ b/codegen/src/main/java/software/amazon/awssdk/codegen/AddMetadata.java
@@ -19,6 +19,7 @@ import java.util.Collections;
 import java.util.Optional;
 import java.util.stream.Collectors;
 import software.amazon.awssdk.codegen.internal.Constant;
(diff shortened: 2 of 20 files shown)


mtdowling (commented):
Why prefix and suffix and not just a single value to define the package name?

zoewangg:
>Why prefix and suffix and not just a single value to define the package name?

Sure, there are two reasons: 1) this is how v1 is implemented, and I searched internally and there are packages using only one of them, so implementing it the same way may make it easier for users to migrate to v2. 2) supporting a single value is actually more work because it require refactoring of the existing code

mtdowling:
I’m not understanding the difference between prefix and suffix. Do they make up the package name or is suffix the client class name?

If they make up the package name, then I don’t see that as something you should carry forward from v1. To reduce churn in this PR, you could also just parse out the prefix and suffix from a single package name string. Migrating this configuration from v1 to v2 seems like a minimal change and worthwhile since it’s easier to understand for new clients going forward.

zoewangg:
> I’m not understanding the difference between prefix and suffix. Do they make up the package name or is suffix the client class name?

@mtdowling They make up the root package name, `[prefix].[suffix]`. There are subpackages such as `[prefix].[suffix].model`, and `[prefix].[suffix].paginators`. We can add customization for those if needed in the future.

We can't parse out the prefix and suffix because suffix can contain `.` in it.

For example, the following custom config would generate root pacakge `foo.bar.hello.world`
```
{
    "packageName": {
        "prefix": "foo.bar",
        "suffix": "hello.world"
    }
}
```

zoewangg:
https://github.com/aws/aws-sdk-java-v2/pull/4751 is the new PR

Issue #4703, opened 2023-11-15
Parsing env variable NO_PROXY


a-brunelliere-maif:
### Describe the bug

Hi,

I don't know if the content of the NO_PROXY environment variable is actually described by a standard, but a pattern like this is quite common : "localhost, .example.com"

The code software.amazon.awssdk.utils.http.SdkHttpUtils.extractNonProxyHosts() code relies on a block of code that only expects a localhost|*.example.com format

### Expected Behavior

Format like "localhost, .example.com" is allowed

### Current Behavior

Format like "localhost, .example.com" is misinterpreted

### Reproduction Steps

Create a client with env var NO_PROXY=localhost, .example.com

### AWS Java SDK version used

2.21.22

### JDK version used

11

### Operating System and version

windows 11

debora-ito:
@a-brunelliere-maif thank you for reaching out.

The reason why we use '|' as separators is because it's the standard for Java System Properties - According to the Java SE 8 Networking Properties [documentation](https://docs.oracle.com/javase/8/docs/api/java/net/doc-files/net-properties.html):

> http.nonProxyHosts (default: localhost|127.*|[::1])
> Indicates the hosts that should be accessed without going through the proxy. Typically this defines internal hosts. The value of this property is a list of hosts, separated by the '|' character. In addition the wildcard character '\*' can be used for pattern matching. For example -Dhttp.nonProxyHosts=”*.foo.com|localhost” will indicate that every hosts in the foo.com domain and the localhost should be accessed directly even if a proxy server is specified.

We use the same '|' separator in the Environment Variable for consistency.

Can you use '|' instead of a comma ',' in the environment variable?

a-brunelliere-maif:
Hi @debora-ito , thanks for you your answer. You provide me the right documentation for the system property  "http.nonProxyHosts" and we respect this format for it. 
But the env variable "NO_PROXY" has an another common format.
An exemple of this format can be found here : https://about.gitlab.com/blog/2021/01/27/we-need-to-talk-no-proxy/

> no_proxy is a comma- or space-separated list of machine
or domain names, with optional :port part.  If no :port
part is present, it applies to all ports on that domain.
Example: no_proxy="cern.ch,some.domain:8001"

The amazon client s3 can retrieve the both values, except it use the "http.nonProxyHosts" format for the "NO_PROXY" variable

martinsefcik:
[...]

PR #4735, opened 2023-12-02
Bugfix to support comma separated values in no_proxy environment settings

joviegas:
## Motivation and Context

[4703](https://github.com/aws/aws-sdk-java-v2/issues/4703)

## Root Cause

- The logic to parse no_proxy environment variables for multiple host used logic same as that of System property NO_PROXY setting which treated "|" as separator

- Why was this missed in testing ?
  - [ProxyConfigurationTest](https://github.com/aws/aws-sdk-java-v2/blob/0e3b14d013a698b9cf4d6a1bc208542509fbe2bb/utils/src/test/java/software/amazon/awssdk/utils/proxy/ProxyConfigurationTest.java#L75-L82) had added a test case to test multiple hosts scenario. Unfortunately the test had a typo error where hosts names were added in double quotes instead of separate string.

## Modifications

- Added logic to split string based on  "," instead of "|".
- Added logic to trim extra spaces
- We are not adding greedy to non-greedy since it can cause issues if user already sends a non-greedy string. Wild cards are passed as it is.

## New revision
After offline Discussion with @zoewangg  updated this to be consistent as [V1-here](https://github.com/aws/aws-sdk-java/blob/c69beeca1e263c67f68e1d1a1f5d7f3dd9bf805a/aws-java-sdk-core/src/main/java/com/amazonaws/ClientConfiguration.java#L1109C20-L1116) and [V1-Here](https://github.com/aws/aws-sdk-java/blob/a186eadd5376908b472993353dfc1b115e61dc1b/aws-java-sdk-core/src/main/java/com/amazonaws/http/apache/SdkProxyRoutePlanner.java#L47-L55) supporting both commas and pipes.

## Testing

- Junits added

## Types of changes

## License

### DIFF
--- a/.changes/next-release/bugfix-AWSSDKforJavav2-02e443d.json
+++ b/.changes/next-release/bugfix-AWSSDKforJavav2-02e443d.json
@@ -2,5 +2,5 @@
     "type": "bugfix",
     "category": "AWS SDK for Java v2",
     "contributor": "",
-    "description": "Comma separated multiple hosts specified in no_proxy environment settings were not treated as multiple hosts"
+    "description": "Comma separated multiple hosts specified in `NO_PROXY` environment variable are now treated as multiple hosts."
 }
--- a/.changes/next-release/bugfix-AWSSDKforJavav2-02e443d.json
+++ b/.changes/next-release/bugfix-AWSSDKforJavav2-02e443d.json
@@ -1,0 +1,6 @@
+{
+    "type": "bugfix",
+    "category": "AWS SDK for Java v2",
+    "contributor": "",
+    "description": "Comma separated multiple hosts specified in no_proxy environment settings were not treated as multiple hosts"
+}
--- a/utils/src/main/java/software/amazon/awssdk/utils/http/SdkHttpUtils.java
+++ b/utils/src/main/java/software/amazon/awssdk/utils/http/SdkHttpUtils.java
@@ -444,9 +444,9 @@ public final class SdkHttpUtils {
         return Collections.emptySet();
     }
 
-    private static Set<String> extractNonProxyHostsFromEnvironmentVariable(String systemNonProxyHosts) {
-        if (systemNonProxyHosts != null && !isEmpty(systemNonProxyHosts)) {
-            return Arrays.stream(systemNonProxyHosts.split(","))
+    private static Set<String> extractNonProxyHostsFromEnvironmentVariable(String environmentNonProxyHosts) {
+        if (environmentNonProxyHosts != null && !isEmpty(environmentNonProxyHosts)) {
+            return Arrays.stream(environmentNonProxyHosts.split(","))
                          .map(String::toLowerCase)
                          .map(String::trim)
                          .collect(Collectors.toSet());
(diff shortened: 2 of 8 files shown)


utils/src/test/java/software/amazon/awssdk/utils/SdkHttpUtilsTest.java
```diff
@@ -29,10 +30,25 @@
...
+import software.amazon.awssdk.testutils.EnvironmentVariableHelper;
 import software.amazon.awssdk.utils.http.SdkHttpUtils;
 
 public class SdkHttpUtilsTest {
+
+
+
+    private static final EnvironmentVariableHelper ENVIRONMENT_VARIABLE_HELPER = new EnvironmentVariableHelper();
+
+    @BeforeEach
```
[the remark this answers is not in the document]
joviegas: Done

utils/src/main/java/software/amazon/awssdk/utils/http/SdkHttpUtils.java
```diff
@@ -431,31 +431,23 @@ public static List<String> splitQueryString(String queryString) {
...
-                         .collect(Collectors.toSet());
-        }
-        return Collections.emptySet();
-    }
-
     public static Set<String> parseNonProxyHostsEnvironmentVariable() {
-        String environmentNonProxyHosts = ProxyEnvironmentSetting.NO_PROXY.getStringValue().orElse(null);
-        return extractNonProxyHostsFromEnvironmentVariable(environmentNonProxyHosts);
+        String hosts = ProxyEnvironmentSetting.NO_PROXY.getStringValue()
+                                                       .map(noProxyHost -> noProxyHost.replace(",", "|"))
```
joviegas, line 449: After offline Discussion with @zoewangg  updated this to be consistent as [V1](https://github.com/aws/aws-sdk-java/blob/a186eadd5376908b472993353dfc1b115e61dc1b/aws-java-sdk-core/src/main/java/com/amazonaws/http/apache/SdkProxyRoutePlanner.java#L47-L55) supporting both commas and pipes.

.changes/next-release/bugfix-AWSSDKforJavav2-02e443d.json
```diff
@@ -0,0 +1,6 @@
+{
+    "type": "bugfix",
+    "category": "AWS SDK for Java v2",
+    "contributor": "",
+    "description": "Comma separated multiple hosts specified in no_proxy environment settings were not treated as multiple hosts"
```
zoewangg, line 5: Can we change it to:

```
Comma separated multiple hosts specified in `NO_PROXY` environment variable and `http.nonProxyHosts ` system property are now treated as multiple hosts.
```

zoewangg (approved):
Can we update Javadocs for `ProxyConfiguration.Builder#useEnvironmentVariableValues` to mention this behavior? We have proxy config class for every http client, hopefully it's not too much copy paste

.changes/next-release/bugfix-AWSSDKforJavav2-02e443d.json
joviegas: Done

PR #4743, opened 2023-12-05
Adding interface methods for Progress Listener and state capturing Progress Snapshots

anirudh9391:
Exposing interface methods for Progress Listener and state capturing Progress Snapshots

## Motivation and Context
V2 parity feature to implement Progress Listeners 

## Modifications
Added interface methods for Listener and Snapshots

## Testing
Successful Build. Default implementations will have test classes

## Types of changes

- [ x] New feature (non-breaking change which adds functionality)

## Checklist

## License


core/sdk-core/src/main/java/software/amazon/awssdk/core/progress/listener/ProgressListener.java
```diff
@@ -0,0 +1,395 @@
...
+import software.amazon.awssdk.annotations.SdkProtectedApi;
+import software.amazon.awssdk.annotations.SdkPublicApi;
+import software.amazon.awssdk.annotations.ThreadSafe;
+import software.amazon.awssdk.core.progress.snapshot.ProgressSnapshot;
+import software.amazon.awssdk.http.SdkHttpRequest;
+
+/**
+ * The {@link ProgressListener} interface may be implemented by your application in order to receive event-driven updates on
+ * the progress of a service call. When you construct an {@link PutObjectRequest} or {@link
+ * UploadPartRequest} request submitted to the Sdk, you may provide a variable number of {@link
```
millems, line 29: Maybe we should say "An {@link SdkRequest} like a {@link PutObjectRequest}..."

core/sdk-core/src/main/java/software/amazon/awssdk/core/progress/listener/ProgressListener.java
```diff
@@ -0,0 +1,395 @@
...
+ * ProgressListener}s to be associated with that request. Then, throughout the lifecycle of the request,
+ * the Sdk
+ * will invoke the provided {@link ProgressListener}s when important events occur, like additional bytes being transferred,
+ * allowing you to monitor the ongoing progress of the transaction.
```
millems, line 33: Here, we might want to use wording that differentiates this from just using the ExecutionInterceptor's "read" methods. E.g. While ExecutionInterceptors are focused on the lifecycle of the request within the SDK, ProgressListeners are focused on the lifecycle of the request within the HTTP client

core/sdk-core/src/main/java/software/amazon/awssdk/core/progress/listener/ProgressListener.java
```diff
@@ -0,0 +1,395 @@
...
+ *         <li>{@link Context.RequestPrepared#progressSnapshot()}</li>
+ *     </ul>
+ *     <li>{@link #requestBytesSent(Context.RequestBytesSent)} - Additional bytes have been sent. This
+ *     method may be called many times per request, depending on the request payload size and I/O buffer sizes.
+ *     <li>{@link #responseBytesReceived(Context.ResponseBytesReceived)} - Additional bytes have been received. This
+ *     method may be called many times per request, depending on the response payload size and I/O buffer sizes.
+ *     <li>{@link #executionSuccess(Context.ExecutionSuccess)} - The transfer has completed successfully. This method is called
+ *     exactly once for a successful transfer.</li>
+ * </ol>
+ * For every failed attempt {@link #attemptFailure(Context.AttemptFailure)}  will be called exactly once.
```
millems, line 57: I'm worried people might be hung up on "exactly once" and think that it only gets called once per execution (even though that's not what this actually says). Do we need the "exactly once" here or is "For every failed attempt, {@link #attemptFailure(Context.AttemptFailure)} will be called." sufficient?

core/sdk-core/src/main/java/software/amazon/awssdk/core/progress/listener/ProgressListener.java
```diff
@@ -0,0 +1,395 @@
...
+ * <ol>
+ *     <li>{@link ProgressListener} implementations should not block, sleep, or otherwise delay the calling thread. If you need
+ *     to perform blocking operations, you should schedule them in a separate thread or executor that you control.</li>
+ *     <li>Be mindful that {@link #requestBytesSent(Context.RequestBytesSent)} or
+ *     {@link #responseBytesReceived(Context.ResponseBytesReceived)}
+ *     may be called extremely often for large payloads
+ *     (subject to I/O buffer sizes). Be careful in implementing expensive operations as a side effect. Consider rate-limiting
+ *     your side effect operations, if needed.</li>
+ *     <li>{@link ProgressListener}s may be invoked by different threads. If your {@link ProgressListener} is stateful,
+ *     ensure that it is also thread-safe.</li>
```
millems, line 70: This implies we don't have a happens-before relationship for each of the methods. Is that true? If so, for which methods? Is it true for HTTP/1 as well as HTTP/2?

core/sdk-core/src/main/java/software/amazon/awssdk/core/progress/listener/ProgressListener.java
```diff
@@ -0,0 +1,395 @@
...
+ * |=================   | 87.5%
+ * |====================| 100.0%
+ * Request execution successful!
+ * </pre>
+ */
+@SdkPublicApi
+public interface ProgressListener {
+
+    /**
+     * This method is called right after a request object is marshalled and ready to be sent to the service
```
millems, line 106: And signed?

core/sdk-core/src/main/java/software/amazon/awssdk/core/progress/listener/ProgressListener.java
```diff
@@ -0,0 +1,395 @@
...
+    * Additional bytes received
+    * <p>
+    * Available context attributes:
+    * <ol>
+    *     <li>{@link Context.ResponseBytesReceived#request()}</li>
+    *     <li>{@link Context.ResponseBytesReceived#progressSnapshot()}</li>
+    * </ol>
+    */
+    default void attemptFailureResponseBytesReceived(Context.AttemptFailureResponseBytesReceived context) {
+    }
```
millems, line 176: Can we describe the differences between responseBytesReceived and attemptFailureResponseBytesReceived here?

core/sdk-core/src/main/java/software/amazon/awssdk/core/progress/listener/ProgressListener.java
```diff
@@ -0,0 +1,395 @@
...
+    /**
+     * This method is called after the request transaction is initiated, i.e. request header is sent to the service
+     * <p>
+     * Available context attributes:
+     * <ol>
+     *     <li>{@link Context.RequestHeaderSent#request()}</li>
+     *     <li>{@link Context.RequestHeaderSent#progressSnapshot()}</li>
+     * </ol>
+     */
+    default void requestHeaderSent(Context.RequestHeaderSent context) {
```
millems, line 126: Can we also add to the Javadoc of each what methods might be called next? E.g. after this will always be either requestBytesSent, responseHeaderReceived, or attemptFailed?

core/sdk-core/src/main/java/software/amazon/awssdk/core/progress/listener/ProgressListener.java
```diff
@@ -0,0 +1,395 @@
...
+     * payload size.
+     * <p>
+     * Available context attributes:
+     * <ol>
+     *     <li>{@link Context.RequestBytesSent#request()}</li>
+     *     <li>{@link Context.RequestBytesSent#progressSnapshot()}</li>
+     * </ol>
+     */
+    default void requestBytesSent(Context.RequestBytesSent context) {
+    }
```
millems, line 140: Is this called if the request payload size is 0?

core/sdk-core/src/main/java/software/amazon/awssdk/core/progress/listener/ProgressListener.java
```diff
@@ -0,0 +1,395 @@
...
+        @Immutable
+        @ThreadSafe
+        @SdkPublicApi
+        @SdkPreviewApi
+        public interface RequestPrepared {
+            /**
+             * The {@link SdkHttppRequest} that was submitted to SDK, i.e., the {@link PutObjectRequest} or
+             * {@link GetObjectRequest}.
+             */
+            SdkHttpRequest request();
```
millems, line 248: This would be the _marshalled_ PutObjectRequest or GetObjectRequest. Can we clarify that in the docs? Should we add the non-marshalled SdkRequest as well?

core/sdk-core/src/main/java/software/amazon/awssdk/core/progress/listener/ProgressListener.java
```diff
@@ -0,0 +1,395 @@
...
+
+            /**
+             * The immutable {@link ProgressSnapahot} for this specific update.
+             */
+            ProgressSnapshot progressSnapshot();
```
millems, line 253: Does this create a new snapshot each time it is called, or is it fixed? (I assume fixed, since it says immutable).

core/sdk-core/src/main/java/software/amazon/awssdk/core/progress/listener/ProgressListener.java
```diff
@@ -0,0 +1,395 @@
...
+         *     <li>{@link RequestPrepared#progressSnapshot()}</li>
+         * </ol>
+         */
+        @Immutable
+        @ThreadSafe
+        @SdkPublicApi
+        @SdkPreviewApi
+        public interface RequestPrepared {
+            /**
+             * The {@link SdkHttppRequest} that was submitted to SDK, i.e., the {@link PutObjectRequest} or
```
millems, line 245: Nit: s/SdkHttppRequest/SdkHttpRequest/. I see a few other small typos on these. Hopefully your IDE can find them all?

core/sdk-core/src/main/java/software/amazon/awssdk/core/progress/listener/ProgressListener.java
```diff
@@ -0,0 +1,395 @@
...
+         *     <li>{@link ResponseHeaderReceived#request()}</li>
+         *     <li>{@link ResponseHeaderReceived#progressSnapshot()}</li>
+         * </ol>
+         */
+        @Immutable
+        @ThreadSafe
+        @SdkPublicApi
+        @SdkPreviewApi
+        public interface ResponseHeaderReceived extends RequestBytesSent {
+        }
```
millems, line 302: Should we add the SdkHttpResponse here?

core/sdk-core/src/main/java/software/amazon/awssdk/core/progress/listener/ProgressListener.java
```diff
@@ -0,0 +1,395 @@
...
+     * <p>
+     * Available context attributes:
+     * <ol>
+     *     <li>{@link Context.ExecutionSuccess#request()}</li>
+     *     <li>{@link Context.ExecutionSuccess#progressSnapshot()}</li>
+     *     <li>{@link Context.ExecutionSuccess#executionSuccess()} ()}</li>
+     * </ol>
+     */
+    default void executionSuccess(Context.ExecutionSuccess context) {
+    }
```
millems, line 189: Did we intend to include an executionFailure method as well, so that the user knows when the retries have all failed? (we'd call executionFailure after attemptFailure if we are done retrying). I see it referenced in a few places below.

core/sdk-core/src/main/java/software/amazon/awssdk/core/progress/listener/ProgressListener.java
```diff
@@ -0,0 +1,395 @@
...
+
+    /**
+     * This method is called for every failure of a request attempt.
+     * An ideal implementation would invoke executionFailure for a number of attemptFailures greater than a threshold
```
millems, line 193: Agreed. Can we add a TODO here so we don't accidentally release this implementer note?

core/sdk-core/src/main/java/software/amazon/awssdk/core/progress/listener/ProgressListener.java
```diff
@@ -0,0 +1,395 @@
...
+         *     <li>{@link AttemptFailure#request()}</li>
+         *     <li>{@link AttemptFailure#progressSnapshot()}</li>
+         * </ol>
+         */
+        @Immutable
+        @ThreadSafe
+        @SdkPublicApi
+        @SdkPreviewApi
+        public interface AttemptFailure extends RequestPrepared {
+        }
```
millems, line 372: Should we also include the Throwable exception() here?

core/sdk-core/src/main/java/software/amazon/awssdk/core/progress/listener/ProgressListener.java
```diff
@@ -0,0 +1,395 @@
...
+         *     <li>{@link ExecutionFailure#request()}</li>
+         *     <li>{@link ExecutionFailure#progressSnapshot()}</li>
+         *     <li>{@link ExecutionFailure#exception()}</li>
+         * </ol>
+         */
+        @Immutable
+        @ThreadSafe
+        @SdkPublicApi
+        @SdkPreviewApi
+        public interface ExecutionFailure extends RequestPrepared {
```
millems, line 388: Nit: Not used. Seems like an accident (see earlier comments)

core/sdk-core/src/main/java/software/amazon/awssdk/core/progress/snapshot/DefaultProgressSnapshot.java
```diff
@@ -0,0 +1,2 @@
+package software.amazon.awssdk.core.progress.snapshot;public class DefaultProgressSnapshot {
+}
```
millems, line 2: Accidental commit?

core/sdk-core/src/main/java/software/amazon/awssdk/core/progress/snapshot/ListenerProgress.java
```diff
@@ -0,0 +1,33 @@
...
+@Immutable
+@ThreadSafe
+@SdkPublicApi
+public interface ListenerProgress {
+
+    /**
+    * Takes a snapshot of the request execution progress
+    * represented by an immutable {@link ProgressSnapshot}.
+    */
+
```
millems, line 31: nit: extra space

core/sdk-core/src/main/java/software/amazon/awssdk/core/progress/snapshot/ListenerProgress.java
```diff
@@ -0,0 +1,33 @@
...
+package software.amazon.awssdk.core.progress.snapshot;
+
+import software.amazon.awssdk.annotations.Immutable;
+import software.amazon.awssdk.annotations.SdkPublicApi;
+import software.amazon.awssdk.annotations.ThreadSafe;
+
+@Immutable
+@ThreadSafe
+@SdkPublicApi
+public interface ListenerProgress {
```
millems, line 25: What's this used for?

core/sdk-core/src/main/java/software/amazon/awssdk/core/progress/listener/ProgressListener.java
millems, line 253: Is this the progress of the upload or download?

core/sdk-core/src/main/java/software/amazon/awssdk/core/progress/listener/ProgressListener.java
```diff
@@ -0,0 +1,395 @@
...
+         */
+        @Immutable
+        @ThreadSafe
+        @SdkPublicApi
+        @SdkPreviewApi
+        public interface ExecutionSuccess extends ResponseBytesReceived {
+            /**
+             * The successful completion of a request submitted to the Sdk
+             */
+            ExecutionSuccessObjectRequest executionSuccess();
```
millems, line 338: Should we just "flatten" ExecutionSuccessObjectRequest into this type? E.g. add response() here and remove the ExecutionSuccessObjectRequest?

core/sdk-core/src/main/java/software/amazon/awssdk/core/progress/snapshot/ProgressSnapshot.java
```diff
@@ -0,0 +1,79 @@
...
+@ThreadSafe
+@SdkPublicApi
+public interface ProgressSnapshot {
+    /**
+     * The total number of bytes that have been sent or received so far.
+     */
+    long transferredBytes();
+
+    /**
+     * Time at which transaction started
```
millems, line 39: What time specifically? Execution started? First HTTP request sent? Response headers received?

core/sdk-core/src/main/java/software/amazon/awssdk/core/progress/snapshot/ProgressSnapshot.java
```diff
@@ -0,0 +1,79 @@
...
+    Instant startTime();
+
+    /**
+     * Elapsed time since the start of the transaction
+     */
+    Duration elapsedTime();
+
+    /**
+     * If transaction size is known, estimate time remaining for transaction completion
+     */
```
millems, line 50: TODO: Describe how this is calculated, so that users don't assume we're doing any kind of magic.

core/sdk-core/src/main/java/software/amazon/awssdk/core/progress/snapshot/ProgressSnapshot.java
```diff
@@ -0,0 +1,79 @@
...
+    Optional<Duration> estimatedTimeRemaining();
+
+    /**
+     * The SDK response, or {@link Optional#empty()} if unknown.
+     */
+    Optional<SdkResponse> sdkResponse();
```
millems, line 56: This was on the context objects. Do we need it here as well?

core/sdk-core/src/main/java/software/amazon/awssdk/core/progress/snapshot/ProgressSnapshot.java
```diff
@@ -0,0 +1,79 @@
...
+
+    /**
+     *  Rate of transfer
+     */
+    double averageBytesPer(TimeUnit timeUnit);
+
+    /**
+     * The total size of the transfer, in bytes, or {@link Optional#empty()} if unknown.
```
millems, line 64: TODO: Specify when "unknown" could occur.

core/sdk-core/src/main/java/software/amazon/awssdk/core/progress/listener/ProgressListener.java
anirudh9391: I thought of having that, and went with SdkHttpRequest as the "requestPrepared" method deals with requests that have been marshalled into as SdkHttpRequest already.

 I do see your point that technically, the SdkRequest is what was submitted. I can change this

core/sdk-core/src/main/java/software/amazon/awssdk/core/progress/listener/ProgressListener.java
anirudh9391: RIght. I am making the language tighter

core/sdk-core/src/main/java/software/amazon/awssdk/core/progress/listener/ProgressListener.java
anirudh9391: Yeah do we need happens-before links to progress methods though ? These methods are ideally triggered only when the previous event has completed and there are further progress updates.

anirudh9391: I thought of it being this way for all methods in all protocols. Let me know if we need to re-think that

core/sdk-core/src/main/java/software/amazon/awssdk/core/progress/listener/ProgressListener.java
anirudh9391: Good point. Changing

core/sdk-core/src/main/java/software/amazon/awssdk/core/progress/listener/ProgressListener.java
anirudh9391: Right good point.

core/sdk-core/src/main/java/software/amazon/awssdk/core/progress/listener/ProgressListener.java
anirudh9391: It is not, i can add this

core/sdk-core/src/main/java/software/amazon/awssdk/core/progress/listener/ProgressListener.java
anirudh9391: Done. I have removed this, we dont need the implementer note here

core/sdk-core/src/main/java/software/amazon/awssdk/core/progress/listener/ProgressListener.java
anirudh9391: Yes, I actually meant to put it in but missed that one. Adding it in now

core/sdk-core/src/main/java/software/amazon/awssdk/core/progress/listener/ProgressListener.java
anirudh9391: Yes I have now clarified in the docs that this will be the marshalled SdkHttpRequest object. Having the non-marshalled SdkRequest would correspond to a different point of the request lifecycle, which is perhaps something of a "requestInitiated". I can see that adding value, if thats what you are suggesting.

core/sdk-core/src/main/java/software/amazon/awssdk/core/progress/listener/ProgressListener.java
anirudh9391: This is used for both. Intended usage is to have the implementation class named Download or Upload. I did not create two different interfaces for download and upload with the same method names to avoid duplication. Unless you think that could be messy ?

anirudh9391: We do create a new snapshot for every update. One snapshot corresponds to one state or progress update. For subsequent updates, we create new snapshots. It is immutable once created. Even the updateAndGet method invoked by ProgressUpdater copies the existing snapshot and modifies it.

core/sdk-core/src/main/java/software/amazon/awssdk/core/progress/listener/ProgressListener.java
anirudh9391: Yup I added this. Thanks for catching this

core/sdk-core/src/main/java/software/amazon/awssdk/core/progress/snapshot/ListenerProgress.java
anirudh9391: Transfer Listener uses this as a wrapping class around Snapshot; it ideally defines methods to update (or create copies and update) the snapshot associated with it, like https://github.com/aws/aws-sdk-java-v2/blob/0b9b0735c447af78316e5d7ef960e2b237f2f073/services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/progress/DefaultTransferProgress.java#L38

core/sdk-core/src/main/java/software/amazon/awssdk/core/progress/snapshot/ProgressSnapshot.java
anirudh9391: I will explain this

core/sdk-core/src/main/java/software/amazon/awssdk/core/progress/snapshot/ProgressSnapshot.java
anirudh9391: I thought it makes sense to read it from the snapshot in case we are downloading a larger payload that might perform a download in multiple stages. Felt like it should be associated with a snapshot and used like here, as TM uses it
https://github.com/aws/aws-sdk-java-v2/blob/0b9b0735c447af78316e5d7ef960e2b237f2f073/services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/model/DefaultFileDownload.java#L70

core/sdk-core/src/main/java/software/amazon/awssdk/core/progress/listener/ProgressListener.java
anirudh9391: Done

core/sdk-core/src/main/java/software/amazon/awssdk/core/progress/listener/ProgressListener.java
anirudh9391: Discussed offline. Added SdkHttpResponse

core/sdk-core/src/main/java/software/amazon/awssdk/core/progress/listener/ProgressListener.java
anirudh9391: Discussed offline, done.

core/sdk-core/src/main/java/software/amazon/awssdk/core/progress/snapshot/ListenerProgress.java
anirudh9391: Corrected

core/sdk-core/src/main/java/software/amazon/awssdk/core/progress/listener/ProgressListener.java
anirudh9391: Included an SdkRequest() too

PR #4752, opened 2023-12-06
Add credentialScope to endpoint params

davidh44:
## Modifications

Update codegen to add `credentialScope` to endpoint params
Add credentialScope `BuiltInParameter`

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

## Checklist

- [ ] I have read the [CONTRIBUTING](https://github.com/aws/aws-sdk-java-v2/blob/master/CONTRIBUTING.md) document
- [ ] Local run of `mvn install` succeeds
- [ ] My code follows the code style of this project
- [ ] My change requires a change to the Javadoc documentation
- [ ] I have updated the Javadoc documentation accordingly
- [ ] I have added tests to cover my changes
- [ ] All new and existing tests passed
- [ ] I have added a changelog entry. Adding a new entry must be accomplished by running the `scripts/new-change` script and following the instructions. Commit the new file created by the script in `.changes/next-release` with your changes.
- [ ] My change is to implement 1.11 parity feature and I have updated [LaunchChangelog](https://github.com/aws/aws-sdk-java-v2/blob/master/docs/LaunchChangelog.md)

## License

- [ ] I confirm that this pull request can be released under the Apache 2 license

PR #4770, opened 2023-12-12
Retrieve credentials from IDENTITY_PROVIDERS attribute

davidh44:
## Modifications

When resolving `credentialScope`, retrieve credentials from `IDENTITY_PROVIDERS` attribute instead of `SELECTED_AUTH_SCHEME`

related PR: https://github.com/aws/aws-sdk-java-v2/pull/4752

## Types of changes

## Checklist

## License

### DIFF
--- a/codegen/src/main/java/software/amazon/awssdk/codegen/poet/rules/EndpointResolverInterceptorSpec.java
+++ b/codegen/src/main/java/software/amazon/awssdk/codegen/poet/rules/EndpointResolverInterceptorSpec.java
@@ -74,6 +74,7 @@ import software.amazon.awssdk.http.auth.aws.signer.RegionSet;
 import software.amazon.awssdk.http.auth.spi.scheme.AuthSchemeOption;
 import software.amazon.awssdk.identity.spi.AwsCredentialsIdentity;
 import software.amazon.awssdk.identity.spi.Identity;
+import software.amazon.awssdk.identity.spi.IdentityProviders;
 import software.amazon.awssdk.regions.Region;
 import software.amazon.awssdk.utils.AttributeMap;
 import software.amazon.awssdk.utils.CompletableFutureUtils;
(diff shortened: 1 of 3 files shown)


codegen/src/main/java/software/amazon/awssdk/codegen/poet/rules/EndpointResolverInterceptorSpec.java
```diff
@@ -317,16 +318,14 @@ private MethodSpec getCredentialScopeIfPresent() {
...
-        b.beginControlFlow("if (identity instanceof $T)", AwsCredentialsIdentity.class);
-        b.addStatement("$T awsCredentialsIdentity = ($T) identity", AwsCredentialsIdentity.class, AwsCredentialsIdentity.class);
-        b.addStatement("return awsCredentialsIdentity.credentialScope().map($T::of).orElse(null)", Region.class);
-        b.endControlFlow();
+        b.addStatement("$T identityProviders = executionAttributes.getAttribute($T.IDENTITY_PROVIDERS)",
+                       IdentityProviders.class, SdkInternalExecutionAttribute.class);
 
-        b.addStatement("return null");
+        b.addStatement("$T awsCredentialsIdentity = $T.joinLikeSync(identityProviders.identityProvider($T.class)"
+                       + ".resolveIdentity())",
```
L-Applin, line 325: Is there a way that the call to `identityProviders.identityProvider(...)` can return null or are we sure that is always returns a valid instance?

gosar: It can be null. We should definitely guard against null.

davidh44: good point it would return null for TokenIdentity, adding null check

gosar:
Can you explain the motivation for this change? It is making an additional `resolveIdentity()`. A given request should not resolve identity more than once, as the results could be different.

PR #4776, opened 2023-12-13
Various performance optimization for CRT sync HTTP client

zoewangg:
## Motivation and Context
Various performance optimization for CRT sync HTTP client:
1. `InputStreamSubscriber`:  not throw exception when `close` is invoked if it has buffered all content to avoid exception creation.
2. `InputStreamAdaptingHttpStreamResponseHandler`: remove volatile for simplePublisher variable since it's unncesssary `private volatile SimplePublisher<ByteBuffer> simplePublisher;` 
3.  `InputStreamAdaptingHttpStreamResponseHandler`: complete future as soon as streaming starts. Previously, the future completes after streaming has finished, which means we'll buffer all data in memory first before user starts to read from the inputstream

### DIFF
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/loader/ClasspathSdkHttpServiceProvider.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/loader/ClasspathSdkHttpServiceProvider.java
@@ -67,7 +67,7 @@ final class ClasspathSdkHttpServiceProvider<T> implements SdkHttpServiceProvider
 
     @Override
     public Optional<T> loadService() {
-        Queue<T> impls = new PriorityQueue<>(Comparator.comparingInt(o -> httpServicesPriority.getOrDefault(o.getClass(),
+        Queue<T> impls = new PriorityQueue<>(Comparator.comparingInt(o -> httpServicesPriority.getOrDefault(o.getClass().getName(),
                                                                                                             Integer.MAX_VALUE)));
         Iterable<T> iterable = () -> serviceLoader.loadServices(serviceClass);
         iterable.forEach(impl -> impls.add(impl));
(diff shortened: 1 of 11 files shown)


utils/src/main/java/software/amazon/awssdk/utils/async/InputStreamSubscriber.java
```diff
@@ -125,6 +128,11 @@ public int read(byte[] bytes, int off, int len) {
     @Override
     public void close() {
         synchronized (subscribeLock) {
+            // If it is done, no-op
+            if (inputStreamState.get().equals(State.STREAMING_DONE)) {
```
zoewangg, line 132: The SDK will close the input stream once a request is finished and in this case, if we already have all data buffered and received, there's no need to raise exception (exception creation is expensive)

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/loader/ClasspathSdkHttpServiceProvider.java
```diff
@@ -67,7 +67,7 @@ final class ClasspathSdkHttpServiceProvider<T> implements SdkHttpServiceProvider
 
     @Override
     public Optional<T> loadService() {
-        Queue<T> impls = new PriorityQueue<>(Comparator.comparingInt(o -> httpServicesPriority.getOrDefault(o.getClass(),
+        Queue<T> impls = new PriorityQueue<>(Comparator.comparingInt(o -> httpServicesPriority.getOrDefault(o.getClass().getName(),
```
zoewangg, line 70: This fixes the bug I found during testing that the priority order was not honored. I updated tests accordingly to catch this issue

http-clients/aws-crt-client/src/main/java/software/amazon/awssdk/http/crt/internal/CrtRequestExecutor.java
```diff
@@ -201,87 +194,6 @@ private void executeRequest(CrtAsyncRequestContext executionContext,
         }
     }
 
-    private static final class InputStreamAdaptingHttpStreamResponseHandler implements HttpStreamResponseHandler {
```
zoewangg, line 204: I moved this class to a separate class.

dagnir:
Nice, can you give a brief overview of the optimizations?

zoewangg:
> Nice, can you give a brief overview of the optimizations?

yup, added it in the description

dagnir (approved):

PR #4796, opened 2023-12-20
Use separate XML writer for S3 operations

davidh44:
## Motivation and Context

Currently, when calling `DeleteObjects` with an object key that contains characters that need to be escaped in XML, and is encoded, the decoded key is deleted instead.

e.g., sending `DeleteObjectsRequest` specifying the key `&lt;Key&gt;objectId&lt;/Key&gt;` deletes the object with the key `<Key>objectId</Key>`

This is because the `XmlWriter` [performs unescaping](https://github.com/aws/aws-sdk-java-v2/blob/master/core/protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/internal/marshall/XmlWriter.java#L210).

## Modifications

Created a separate `S3XmlWriter` which does not perform unescaping, which will be used exclusively for S3 operations.

## Testing

Added integ tests

## Types of changes

## Checklist

## License


core/protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/AwsXmlProtocolFactory.java
```diff
@@ -163,10 +163,15 @@ Optional<XmlElement> getErrorRoot(XmlElement document) {
 
     private XmlGenerator createGenerator(OperationInfo operationInfo) {
         return operationInfo.hasPayloadMembers() ?
-               XmlGenerator.create(operationInfo.addtionalMetadata(XML_NAMESPACE_ATTRIBUTE)) :
+               XmlGenerator.create(operationInfo.addtionalMetadata(XML_NAMESPACE_ATTRIBUTE), isS3(operationInfo)) :
                null;
     }
 
+    private boolean isS3(OperationInfo operationInfo) {
+        return "http://s3.amazonaws.com/doc/2006-03-01/"
```
L-Applin, line 171: Is that the only way we can check for s3? Can the namespace change? What happens if it changes?

core/protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/internal/marshall/S3XmlWriter.java
L-Applin, line 1: Is there anyway the `S3XmlWriter` could be in the s3 module, so that we don't include more service specific code in core packages?

core/protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/AwsXmlProtocolFactory.java
davidh44: namespace seems to be the only way as far as I can tell
good point, updated to check for the prefix without the date and extracted to a class field for more visibility

core/protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/internal/marshall/S3XmlWriter.java
davidh44: unfortunately not, we'd have to add a dependency on the s3 module in order to do so

L-Applin: I notice that `AwsXmlProtocolFactory` has a reference to `SdkClientConfiguration`, couldn't we pass in a client option (set from s3 module) to override the writer?

zoewangg: S3 already uses a custom protocol factory https://github.com/aws/aws-sdk-java-v2/blob/master/services/s3/src/main/resources/codegen-resources/customization.config#L201 can we somehow use this xmlwriter in that factory?

core/protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/internal/marshall/S3XmlWriter.java
```diff
@@ -0,0 +1,43 @@
...
+
+import java.io.Writer;
+import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.utils.StringUtils;
+
+/**
+ * Utility for creating easily creating XML documents, one element at a time, for S3 operations.
+ */
+@SdkInternalApi
+public final class S3XmlWriter extends XmlWriter {
```
zoewangg, line 26: Can we add unit tests for this class? Does this need to be public?

core/protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/internal/marshall/S3XmlWriter.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+package software.amazon.awssdk.protocols.xml.internal.marshall;
+
+import java.io.Writer;
+import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.utils.StringUtils;
+
+/**
+ * Utility for creating easily creating XML documents, one element at a time, for S3 operations.
```
zoewangg, line 23: Can we mention the difference from XmlWriter here?

services/s3/src/it/java/software/amazon/awssdk/services/s3/DeleteObjectsIntegrationTest.java
```diff
@@ -0,0 +1,71 @@
...
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import software.amazon.awssdk.core.sync.RequestBody;
+import software.amazon.awssdk.services.s3.model.Delete;
+import software.amazon.awssdk.services.s3.model.DeleteObjectsRequest;
+import software.amazon.awssdk.services.s3.model.DeleteObjectsResponse;
+import software.amazon.awssdk.services.s3.model.NoSuchKeyException;
+import software.amazon.awssdk.services.s3.model.PutObjectRequest;
+
+@RunWith(Parameterized.class)
```
zoewangg, line 21: Can we use Junit5?

core/protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/internal/marshall/XmlWriter.java
```diff
@@ -31,7 +31,16 @@
  * Utility for creating easily creating XML documents, one element at a time.
  */
 @SdkInternalApi
-final class XmlWriter {
+public class XmlWriter {
```
zoewangg, line 34: Does this need to be public? S3XmlWriter is in the same package, right?

core/protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/internal/marshall/S3XmlWriter.java
davidh44: updated javadocs

core/protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/internal/marshall/S3XmlWriter.java
davidh44: added unit tests, removed public

core/protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/internal/marshall/XmlWriter.java
davidh44: removed public

services/s3/src/it/java/software/amazon/awssdk/services/s3/DeleteObjectsIntegrationTest.java
davidh44: S3IntegrationTestBase uses junit4 so not able to use junit5

core/protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/internal/marshall/S3XmlWriter.java
davidh44: overriding `createGenerator()` in `AwsS3ProtocolFactory` to pass in flag, instead of checking for namespace in `AwsXmlProtocolFactory`

core/protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/internal/marshall/XmlGenerator.java
```diff
@@ -33,9 +33,14 @@ private XmlGenerator(StringWriter stringWriter, XmlWriter xmlWriter) {
         this.xmlWriter = xmlWriter;
     }
 
-    public static XmlGenerator create(String xmlns) {
+    public static XmlGenerator create(String xmlnx) {
+        return create(xmlnx, false);
+    }
+
+    public static XmlGenerator create(String xmlns, boolean isS3) {
```
zoewangg, line 40: Do we still need this method?

services/s3/src/it/java/software/amazon/awssdk/services/s3/DeleteObjectsIntegrationTest.java
```diff
@@ -0,0 +1,71 @@
...
+                                     .build(), RequestBody.fromString("contents"));
+
+        // Does not throw exception
+        s3.headObject(r -> r.bucket(BUCKET).key(key));
+
+        Delete delete = Delete.builder().objects(o -> o.key(key)).build();
+        DeleteObjectsRequest request = DeleteObjectsRequest.builder().bucket(BUCKET).delete(delete).build();
+        DeleteObjectsResponse response = s3.deleteObjects(request);
+
+        assertThat(response.deleted().get(0).key()).isEqualTo(key);
```
zoewangg, line 68: I'm a bit concerned that this test may be flaky since we are doing CRUD operations on the same object back to back. Can we write mock tests instead? something like verifying the raw request?

core/protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/internal/marshall/XmlGenerator.java
davidh44: removed

services/s3/src/it/java/software/amazon/awssdk/services/s3/DeleteObjectsIntegrationTest.java
davidh44: updated to functional mock test

zoewangg (approved):

L-Applin (approved):

Issue #4778, opened 2023-12-13
Seeing java.lang.NoSuchMethodError while initiliazing AwsCredentialsProviderChain 


mukund-thakur:
### Describe the bug

Full stack trace. 
`Caused by: java.lang.NoSuchMethodError: software.amazon.awssdk.thirdparty.org.slf4j.impl.StaticLoggerBinder.getLoggerFactoryClassStr()Ljava/lang/String;
        at software.amazon.awssdk.thirdparty.org.slf4j.LoggerFactory.reportActualBinding(LoggerFactory.java:349) ~[bundle-2.21.41.jar:?]
        at software.amazon.awssdk.thirdparty.org.slf4j.LoggerFactory.bind(LoggerFactory.java:152) ~[bundle-2.21.41.jar:?]
        at software.amazon.awssdk.thirdparty.org.slf4j.LoggerFactory.performInitialization(LoggerFactory.java:124) ~[bundle-2.21.41.jar:?]
        at software.amazon.awssdk.thirdparty.org.slf4j.LoggerFactory.getILoggerFactory(LoggerFactory.java:417) ~[bundle-2.21.41.jar:?]
        at software.amazon.awssdk.thirdparty.org.slf4j.LoggerFactory.getLogger(LoggerFactory.java:362) ~[bundle-2.21.41.jar:?]
        at software.amazon.awssdk.thirdparty.org.slf4j.LoggerFactory.getLogger(LoggerFactory.java:388) ~[bundle-2.21.41.jar:?]
        at software.amazon.awssdk.utils.Logger.loggerFor(Logger.java:221) ~[bundle-2.21.41.jar:?]
        at software.amazon.awssdk.auth.credentials.AwsCredentialsProviderChain.<clinit>(AwsCredentialsProviderChain.java:56) ~[bundle-2.21.41.jar:?]`

### Expected Behavior

Expected it to work. 

### Current Behavior

Not working. Getting above error. 

### Reproduction Steps

Unable to reproduce this locally. This is not working in one of our newly created clusters. 

### Possible Solution

[...]

PR #4797, opened 2023-12-20
Change aws-sdk-java to an optional dependency

dagnir:
Having it as a normal dependency causes third dependencies like SLF4J to also be included, unshaded in the JAR.

## Testing

Build the `bundle` JAR using the `publishing` profile:

```
mvn clean install -P publishing  -Dspotbugs.skip -DskipTests -Dcheckstyle.skip -Djapicmp.skip -Dgpg.skip=true
```

Open the generated JAR and verify that there are no unshaded dependencies apart from `org.reactivestreams` which I believe is by design (this is also true for older versions of the Bundle).

## Types of changes

## Checklist

## License

### DIFF
--- a/.changes/next-release/bugfix-AWSSDKforJavav2Bundle-7731e95.json
+++ b/.changes/next-release/bugfix-AWSSDKforJavav2Bundle-7731e95.json
@@ -1,0 +1,6 @@
+{
+    "type": "bugfix",
+    "category": "AWS SDK for Java v2 Bundle",
+    "contributor": "",
+    "description": "Fix an issue where external dependencies are also included unshaded in in the `bundle` JAR."
+}
--- a/bundle-sdk/pom.xml
+++ b/bundle-sdk/pom.xml
@@ -54,6 +54,7 @@
             <groupId>software.amazon.awssdk</groupId>
             <artifactId>aws-sdk-java</artifactId>
             <version>${awsjavasdk.version}</version>
+            <optional>true</optional>
         </dependency>
     </dependencies>
 


joviegas:
> Open the generated JAR and verify that there are no unshaded dependencies apart from org.reactivestreams which I believe is by design (this is also true for older versions of the Bundle).

Could you please elaborate on where is this design exist  ? 

Are there any cases of backward incompatibility issues ?, as in when there is version upgraded can we have class not found runtime issues ? 


Why dont we exclude the  third dependencies like SLF4J explicitly ?

dagnir:
> Could you please elaborate on where is this design exist ?

It's only a hunch, but we can fix this later if needed since `org.reactivestreams` was never shaded in the `bundle`. I think the intention is because if we shade the `org.reactivestreams` classes, they won't be interoperable with external libraries that use them like [RxJava](https://github.com/ReactiveX/RxJava)

> Why dont we exclude the third dependencies like SLF4J explicitly ?

not sure what you mean by this. We can exclude the classes because they're used for logging. They're shaded so that other libraries don't use them.

joviegas:
> > Could you please elaborate on where is this design exist ?
> 
> It's only a hunch, but we can fix this later if needed since `org.reactivestreams` was never shaded in the `bundle`. I think the intention is because if we shade the `org.reactivestreams` classes, they won't be interoperable with external libraries that use them like [RxJava](https://github.com/ReactiveX/RxJava)
> 
> > Why dont we exclude the third dependencies like SLF4J explicitly ?
> 
> not sure what you mean by this. We can exclude the classes because they're used for logging. They're shaded so that other libraries don't use them.

I meant something like this 

```
    <dependencies>
        <dependency>
            <artifactId>aws-sdk-java</artifactId>
            <exclusions>
                <exclusion>
                    <groupId>org.slf4j</groupId>
                    <artifactId>slf4j-api</artifactId>
                </exclusion>
            </exclusions>
        </dependency>
    </dependencies>
```

dagnir:
See #4778 for context. We don't want to exclude SLF4J entirely, we just want to make sure only the shaded classes are in the JAR.

joviegas (approved):

shwethags:
Hi, which SDK release version contains this fix?

PR #4800, opened 2023-12-22
Fix for Signing Error when ampersand character included in HttpQuery when using DefaultAwsCrtV4aSigner

joviegas:
## Motivation and Context

-  Signing Error when ampersand character included in HttpQuery when using DefaultAwsCrtV4aSigner
- If the Query parameters had `&` in its value then Signing for those request used to always fail.

## Modifications

- The `crtRequestToHttp` loops through query returned by fullUri using "&" and "=" for finding Key and Value.
- When the value has "&" or "=" it assumes it as separate parameters thus we need to use Raw non-decoded parameters,
- The looping is done on `getRawQuery` and then its Decoded while doing a `appendRawQueryParameter`

## Testing

- Added Junit test case.
- tested with a service where the issue was seen.

## Types of changes

## License

### DIFF
--- a/core/auth-crt/src/test/java/software/amazon/awssdk/authcrt/signer/internal/CrtHttpRequestConverterTest.java
+++ b/core/auth-crt/src/test/java/software/amazon/awssdk/authcrt/signer/internal/CrtHttpRequestConverterTest.java
@@ -25,6 +25,9 @@ import java.util.Arrays;
 import java.util.List;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
 import software.amazon.awssdk.crt.http.HttpHeader;
 import software.amazon.awssdk.crt.http.HttpRequest;
 import software.amazon.awssdk.crt.http.HttpRequestBodyStream;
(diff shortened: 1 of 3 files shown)


core/auth-crt/src/test/java/software/amazon/awssdk/authcrt/signer/internal/CrtHttpRequestConverterTest.java
```diff
@@ -87,6 +87,55 @@ public void request_withQueryParams_isConvertedToCrtFormat() {
...
+                                                       .uri(URI.create("https://demo.us-east-1.amazonaws.com"))
+                                                       .build();
+        HttpRequest crtHttpRequest = converter.requestToCrt(request);
+        assertThat(crtHttpRequest.getMethod()).isEqualTo("GET");
+        assertThat(crtHttpRequest.getEncodedPath()).isEqualTo("/path?ampersand%26=one%20%26%20two&equals%3D=three%20%3D%20three");
+        assertHttpRequestSame(request, crtHttpRequest);
+    }
+
+    @Test
+    public void request_withQueryParams_ApersandAndEqualsInValue_isConvertedToCrtFormat() {
```
dagnir, line 108: nit: "ampersand"

dagnir (approved):

core/auth-crt/src/test/java/software/amazon/awssdk/authcrt/signer/internal/CrtHttpRequestConverterTest.java
dagnir, line 1: Can you add a test that includes other values besides `=` and `&` that should be URL encoded? Like a newline or something? Just to make sure CRT handles those too

joviegas: done

PR #4802, opened 2023-12-27
Define Progress Listener Invoker methods, Context classes and modified RequestOverride configuration to include Progress Listener

anirudh9391:
## Motivation and Context
Progress Listeners in V2. Defined invoker methods for all progress tracking methods in different stages of the request lifecycle

## Types of changes

- [ x] New feature (non-breaking change which adds functionality)

## Checklist

- [ x] My change requires a change to the Javadoc documentation
- [ x] I have updated the Javadoc documentation accordingly
- [x ] I have added a changelog entry. Adding a new entry must be accomplished by running the `scripts/new-change` script and following the instructions. Commit the new file created by the script in `.changes/next-release` with your changes.
- [x ] My change is to implement 1.11 parity feature and I have updated [LaunchChangelog](https://github.com/aws/aws-sdk-java-v2/blob/master/docs/LaunchChangelog.md)

## License


core/sdk-core/src/main/java/software/amazon/awssdk/core/RequestOverrideConfiguration.java
```diff
@@ -55,6 +56,7 @@ public abstract class RequestOverrideConfiguration {
     private final EndpointProvider endpointProvider;
     private final CompressionConfiguration compressionConfiguration;
     private final List<SdkPlugin> plugins;
+    private final List<ProgressListener> progressListeners;
```
joviegas, line 59: Should we just have ProgressListener and then have ProgressListenerListenerChain something similar to [SdkTokenProviderChain](https://github.com/aws/aws-sdk-java-v2/blob/9cf5f8adda439c4ed1ac7ae0dd68d71931090549/core/auth/src/main/java/software/amazon/awssdk/auth/token/credentials/SdkTokenProviderChain.java#L51) ?

core/sdk-core/src/main/java/software/amazon/awssdk/core/RequestOverrideConfiguration.java
```diff
@@ -103,9 +106,9 @@ public List<ApiName> apiNames() {
      * unmarshalling, etc. This value should always be positive, if present.
      *
      * <p>The api call timeout feature doesn't have strict guarantees on how quickly a request is aborted when the
-     * timeout is breached. The typical case aborts the request within a few milliseconds but there may occasionally be
-     * requests that don't get aborted until several seconds after the timer has been breached. Because of this, the client
-     * execution timeout feature should not be used when absolute precision is needed.
+     * timeout is breached. The typical case aborts the request within a few milliseconds but there may occasionally be requests
```
joviegas, line 109: Can we please revert unwanted changes  ? (Generic comment)
This becomes helpful when we are doing a final merge.

core/sdk-core/src/main/java/software/amazon/awssdk/core/RequestOverrideConfiguration.java
```diff
@@ -506,6 +511,26 @@ default B putRawQueryParameter(String name, String value) {
         @SdkPreviewApi
         List<SdkPlugin> plugins();
 
+        /**
+         * Sets the progress listeners used to perform request prorgess tracking
```
joviegas, line 515: if possible can we add more java doc , also point to  @ProgressListener class

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/progress/ProgressListenerContext.java
```diff
@@ -0,0 +1,162 @@
...
+import software.amazon.awssdk.core.SdkResponse;
+import software.amazon.awssdk.core.progress.listener.ProgressListener;
+import software.amazon.awssdk.core.progress.snapshot.ProgressSnapshot;
+import software.amazon.awssdk.http.SdkHttpRequest;
+import software.amazon.awssdk.http.SdkHttpResponse;
+import software.amazon.awssdk.utils.ToString;
+import software.amazon.awssdk.utils.builder.CopyableBuilder;
+import software.amazon.awssdk.utils.builder.ToCopyableBuilder;
+
+/**
```
joviegas, line 30: Could you please mention what is the purpose of this class  in the java doc?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/progress/ProgressListenerContext.java
```diff
@@ -0,0 +1,162 @@
...
+            this.downloadProgressSnapshot = downloadProgressSnapshot;
+            return this;
+        }
+
+        public Builder httpResponse(SdkHttpResponse sdkHttpResponse) {
+            this.sdkHttpResponse = sdkHttpResponse;
+            return this;
+        }
+
+        public Builder response(SdkResponse executionSuccessfulSdkResponse) {
```
joviegas, line 151: What is difference between response and httpResponse ?
Can we make naming clear so that we able to distinguish between these two ?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/progress/listener/ProgressListenerInvoker.java
```diff
@@ -0,0 +1,104 @@
...
+import java.util.List;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.Consumer;
+import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.core.progress.listener.ProgressListener;
+import software.amazon.awssdk.utils.Logger;
+import software.amazon.awssdk.utils.Validate;
+
+@SdkInternalApi
+public class ProgressListenerInvoker implements ProgressListener {
```
joviegas, line 29: Java doc for why is this class or responsibility of this Class

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/progress/listener/ProgressListenerInvoker.java
```diff
@@ -0,0 +1,104 @@
...
+
+    private static final Logger log = Logger.loggerFor(ProgressListener.class);
```
joviegas, line 31: Logger class name is incorrect

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/progress/listener/ProgressListenerInvoker.java
joviegas, line 29: Is this class @ThreadSafe ?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/progress/listener/ProgressListenerInvoker.java
```diff
@@ -0,0 +1,104 @@
...
+
+    private final List<ProgressListener> listeners;
+    private final AtomicBoolean prepared = new AtomicBoolean();
```
joviegas, line 34: Why do we require these to be Atomic ?

joviegas (commented):
Please add Junit test cases for the internal classes

core/sdk-core/src/main/java/software/amazon/awssdk/core/RequestOverrideConfiguration.java
anirudh9391: Yeah, these are actually some indendation changes that I felt should go in for the sake of better readability and uniformity. Do we want to keep them as they are ?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/progress/ProgressListenerContext.java
anirudh9391: The httpResponse is what we get back from the service, instance of the SdkHttpResponse class. The response() method defines the SdkResponse, after the httpResponse is marshalled. We made the naming consistent with how Execution Interceptors have defined them https://github.com/aws/aws-sdk-java-v2/blob/722f187b0df0782568e51f1c402bc0287b504eb4/core/sdk-core/src/main/java/software/amazon/awssdk/core/interceptor/Context.java#L156

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/progress/listener/ProgressListenerInvoker.java
anirudh9391: Yes, I can add some javadoc for this

anirudh9391:
> Please add Junit test cases for the internal classes

Yes, I intend do. Just that this was split up into two different pull requests, the subsequent PR is going to have another class that helps testing the functionality of adding the ProgressListener. Will be added in the subsequent PR

core/sdk-core/src/main/java/software/amazon/awssdk/core/RequestOverrideConfiguration.java
anirudh9391: hmm.. good point we could do it this way too, but do we see any advantages either way ? V1 does it this way too. But under the hood, both approaches are similar in that we maintain a list of listeners. We are also staying consistent with TransferListeners by maintaining a list of ProgressListeners. The chain is a wrapper around this list, in essence. We could think of taking a detour from TransferListener's way of implementation if we see definite advantages to the approach.

core/sdk-core/src/main/java/software/amazon/awssdk/core/RequestOverrideConfiguration.java
anirudh9391: Yes sure I will add more javadoc

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/progress/ProgressListenerContext.java
anirudh9391: I can do that

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/progress/listener/ProgressListenerInvoker.java
anirudh9391: I will correct this

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/progress/listener/ProgressListenerInvoker.java
anirudh9391: It is. The AtomicBoolean variables ensure idempotence, i.e. protects from race conditions. In essence the class is threadsafe.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/progress/listener/ProgressListenerInvoker.java
anirudh9391: Yeah this is to make the progress changes idempotent and protect them from race conditions

core/sdk-core/src/main/java/software/amazon/awssdk/core/RequestOverrideConfiguration.java
joviegas: Lets discuss this in SurfaceAPI review.
Could please add a TODO : comment here saying will 
```
// TODO : need to discuss this in SurfaceAPI review
```

core/sdk-core/src/main/java/software/amazon/awssdk/core/RequestOverrideConfiguration.java
joviegas: Lets keep them as they are

core/sdk-core/src/main/java/software/amazon/awssdk/core/RequestOverrideConfiguration.java
anirudh9391: done

core/sdk-core/src/main/java/software/amazon/awssdk/core/RequestOverrideConfiguration.java
anirudh9391: done

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/progress/listener/ProgressListenerInvoker.java
```diff
@@ -0,0 +1,110 @@
+/*
```
anirudh9391, line 1: We dont have defined testing for this class on its own. In a subsequent PR defining a ProgressUpdater class through which this class is accessed, I will define Unit Tests to test the E2E functioning

anirudh9391:
> Please add Junit test cases for the internal classes

Discussed offline. The classes defined here are helper methods that will eventually be invoked by an updater class. The subsequent PR with a ProgressUpdater class will have testing that would test out the code path of the Invoker method in this PR

core/sdk-core/src/main/java/software/amazon/awssdk/core/RequestOverrideConfiguration.java
```diff
@@ -159,6 +163,14 @@ public List<SdkPlugin> plugins() {
         return plugins;
     }
 
+    /**
+     * Return a list of progress listeners that will be used to track request progress.
+     */
+    @SdkPreviewApi
```
joviegas, line 169: We mark as SdkPreviewApi only if the feature is released in Preview mode , are we doing a preview mode release of this feature?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/progress/ProgressListenerContext.java
```diff
@@ -0,0 +1,162 @@
...
+
+    @Override
+    public SdkResponse response() {
+        return response;
+    }
+
+    @Override
+    public String toString() {
+        return ToString.builder("ProgressListenerContext")
+                       .add("request", request)
```
joviegas, line 99: Can we double-check if logging of request doesnot print any sensitive information related to the request , like authorization header etc

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/progress/listener/DefaultSdkRequestProgress.java
```diff
@@ -27,7 +27,7 @@
 /**
  * An SDK-internal implementation of {@link SdkRequestProgress}. This implementation acts as a thin wrapper around {@link
  * AtomicReference}, where calls to get the latest {@link #progressSnapshot()} simply return the latest reference, while {@link
- * SdkRequestProgressUpdater} is responsible for continuously updating the latest reference.
+ * ProgressUpdater} is responsible for continuously updating the latest reference.
```
joviegas, line 30: Typo ?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/progress/listener/ProgressListenerInvoker.java
```diff
@@ -0,0 +1,110 @@
...
+    @Override
+    public void requestHeaderSent(Context.RequestHeaderSent context) {
+        if (!headerSent.getAndSet(true)) {
+            forEach(listener -> listener.requestHeaderSent(context));
+        }
+    }
+
+    @Override
+    public void requestBytesSent(Context.RequestBytesSent context) {
+        forEach(listener -> listener.requestBytesSent(context));
```
joviegas, line 65: We do ThreadSafety checks for few api and not for others

Since "public interface ProgressListener" is public API should and If user is implementing the listeners is it a "MUST" that they need to handle the ThreadSafety  ? something we can discuss offline , just curious

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/progress/listener/DefaultSdkRequestProgress.java
anirudh9391: SdkRequestProgressUpdater is incorrect. I had iniitally named it so, then changed the class name to ProgressUpdater (will be the next pr)

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/progress/listener/ProgressListenerInvoker.java
anirudh9391: Operations like RequestHeaderSent, RequestPrepared are expected to be called exactly once for every request. Thread safety plus idempotence, we dont want repeated calling of the same API. But in the case of BytesSent, they are expected to be called multiple times per operation (with different snapshots). As discussed on Friday, Thread safety would not be an issue, technically, as this is stateless

anirudh9391: I am unsure if, this is technically a thread safe class. I guess since a few of the APIs are, we could call it so ?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/progress/ProgressListenerContext.java
anirudh9391: Okay, I checked out the docs for SdkPreviewApi 

```
Marker interface for preview and experimental APIs. Breaking changes may be
 * introduced to elements marked as {@link SdkPreviewApi}. Users of the SDK
 * should assume that anything annotated as preview will change or break, and
 * <b>should not</b> use them in production.
 ```

I thought we would get this into pre-prod, test it and then remove it.

File: core/sdk-core/src/main/java/software/amazon/awssdk/core/RequestOverrideConfiguration.java (hunk 10 above)
Comment (anirudh9391) [reply core/sdk-core/src/main/java/software/amazon/awssdk/core/RequestOverrideConfiguration.java:169]: Okay, I checked out the docs for SdkPreviewApi 

```
Marker interface for preview and experimental APIs. Breaking changes may be
 * introduced to elements marked as {@link SdkPreviewApi}. Users of the SDK
 * should assume that anything annotated as preview will change or break, and
 * <b>should not</b> use them in production.
 ```

I thought we would get this into pre-prod, test it and then remove it.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/progress/ProgressListenerContext.java
joviegas: Once we merge it to Master its as good as Prod.
The project branches merge for projects that will be fully released as released as @SdkPublicApi
So SdkPreviewApi is added only if we plan to release it in master just for users to try it out in preview mode

anirudh9391: Got it. Incorrect reply for this comment my bad. 

Discussed offline regarding security implications of providing the request object as part of a String. The toString() takes care of omitting private/sensitive information like auth headers, etc.

joviegas: Thanks

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/progress/listener/ProgressListenerInvoker.java
```diff
@@ -0,0 +1,110 @@
...
+    @Override
+    public void executionSuccess(Context.ExecutionSuccess context) {
+        if (!complete.getAndSet(true)) {
+            forEach(listener -> listener.executionSuccess(context));
+        }
+    }
+
+    private void forEach(Consumer<ProgressListener> action) {
+        for (ProgressListener listener : listeners) {
+            runAndLogError(log.logger(), "Exception thrown in TransferListener, ignoring",
```
joviegas, line 106: should this be
```
"Exception thrown in ProgressListener, ignoring"
```

core/sdk-core/src/main/java/software/amazon/awssdk/core/RequestOverrideConfiguration.java
anirudh9391: Discussed offline with Matt. Our guideline around using a Chain abstraction, i.e. V1's progress listener or the SdkTokeProvider, where we update one entity. In this case we have multiple listener methods to be invoked and we maintain a list, as done by plugins https://github.com/aws/aws-sdk-java-v2/blob/5c2868f4512f936000cf084f69bb0707e029a71e/core/sdk-core/src/main/java/software/amazon/awssdk/core/RequestOverrideConfiguration.java#L70

joviegas (commented):
Minor comment related to runErrorLog type

joviegas (approved):

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/progress/listener/ProgressListenerInvoker.java
anirudh9391: Oh yes Typo

PR #4814, opened 2024-01-05
Improved speed and reliability of CloudFrontUtilitiesIntegrationTest.

millems:
Resources are now reused between runs of the integration test, so they should not build up on the account anymore and no longer need to be re-created each time the test is run.

### DIFF
--- a/services/cloudfront/pom.xml
+++ b/services/cloudfront/pom.xml
@@ -63,6 +63,12 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>secretsmanager</artifactId>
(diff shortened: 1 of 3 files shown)


services/cloudfront/pom.xml
```diff
@@ -62,6 +62,11 @@
             <version>${awsjavasdk.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>secretsmanager</artifactId>
+            <version>${awsjavasdk.version}</version>
```
zoewangg, line 68: Test dependency?

millems: Oops, Thanks. Crisis averted.

cenedhryn (approved):

PR #4823, opened 2024-01-10
Create new module for migration tool with basic recipe

zoewangg:
## Motivation and Context, Modifications
- Set up migration-tool module that supports basic OpenRewrite recipes.
- Generate upgrade-sdk-dependencies recipe
- Set up testing infrastructure

## Testing
Tested it locally to convert dependencies for another repo


migration-tool/src/main/resources/scripts/__init__.py
```diff
@@ -0,0 +1 @@
+
```
joviegas, line 1: Is this intentionally blank ?

migration-tool/src/main/resources/generate-recipes
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env python
+
+from scripts.generate_upgrade_sdk_dependencies_recipe import generate_upgrade_sdk_dependencies_recipe
```
joviegas, line 3: Id it possible to write this in Java instead of python ?

migration-tool/src/main/resources/scripts/__init__.py
zoewangg: Yes, it's used to mark this directory as a Python package.

migration-tool/src/main/resources/generate-recipes
zoewangg: It's certainly possible, but since Java is quite verbose, I find it easier to use python for this use-case. We use python elsewhere in this repo for other scripts, so this is not a new dependency.

PR #4825, opened 2024-01-10
Fixed the issue in AWS CRT HTTP clients where the connection is shut down unnecessarily

zoewangg:
## Motivation and Context
Currently, connection may get closed, i.e., not reused if the publisher fails to send the byte buffer to the subscriber, which is not correct since it's client issue not connection issue. 

In addition, CRT currently crashes if connection.shutDown gets invoked after connection.close. This change makes sure we don't invoke connection.shutDown after connection.close

## Modifications
- Ensure the connection only gets closed if there are 5xx server error OR IO errors from reading the response.
- Make sure we don't call connection.shutdown after connection is closed 
- Refactor code to reuse common logic shared in sync and async response handler.

## Testing
Added unit tests

## Types of changes

## Checklist

## License


http-clients/aws-crt-client/src/main/java/software/amazon/awssdk/http/crt/internal/response/InputStreamAdaptingHttpStreamResponseHandler.java
```diff
@@ -84,7 +88,7 @@ public int onResponseBody(HttpStream stream, byte[] bodyBytesIn) {
 
         writeFuture.whenComplete((result, failure) -> {
             if (failure != null) {
-                failFutureAndCloseConnection(stream, failure);
+                failFutureAndReleaseConnection(stream, failure);
```
TingDaoK, line 91: couple lines down:
```
the bodyBytesIn have not cleared the queues yet, so do let backpressure do its thing.
```

Should it be  "do **not** let backpressure do its thing"?

http-clients/aws-crt-client/src/main/java/software/amazon/awssdk/http/crt/internal/response/ResponseHandlerHelper.java
```diff
@@ -0,0 +1,85 @@
...
+        if (connectionClosed.compareAndSet(false, true)) {
+            connection.close();
+            stream.close();
+        }
+    }
+
+    /**
+     * Close the connection completely
+     */
+    public void closeCrtConnection(HttpStream stream) {
```
TingDaoK, line 69: I would suggest to add another bool to the args for shutdown the connection or not, instead of have two different method, which feels more natural to me.

Also, I would rename it to releaseCrtConnectionAndStream/releaseCrtHttpResources, since the passed-in stream also gets released.

http-clients/aws-crt-client/src/main/java/software/amazon/awssdk/http/crt/internal/response/InputStreamAdaptingHttpStreamResponseHandler.java
```diff
@@ -121,16 +128,15 @@ private void onFailedResponseComplete(HttpStream stream, int errorCode) {
...
-
         // For response without a payload, for example, S3 PutObjectResponse, we need to complete the future
         // in onResponseComplete callback since onResponseBody will never be invoked.
         requestCompletionFuture.complete(responseBuilder.build());
-        simplePublisher.complete();
-        crtConn.close();
-        stream.close();
+        simplePublisher.complete().whenComplete((result, failure) -> {
+            if (failure != null) {
+                failFutureAndReleaseConnection(stream, failure);
```
TingDaoK, line 136: Just confused: requestCompletionFuture.complete was invoked at line 133, and this line will call completeExceptionally again, [here](https://github.com/aws/aws-sdk-java-v2/pull/4825/files#diff-b2bbbb71a4e2ac299b2395fd7a94f7511e542dee488d6c2a5a505c13b46fbeb7R113). What behavior are we expecting?

zoewangg: Good catch, will fix it.

http-clients/aws-crt-client/src/main/java/software/amazon/awssdk/http/crt/internal/response/ResponseHandlerHelper.java
zoewangg: I actually like having two methods with meaningful names because it makes it more readable IMO. I'll definitely forget what passing true to `closeCrtResource(stream, true)` implies after a couple of days if I don't go to the definition of the method.

http-clients/aws-crt-client/src/main/java/software/amazon/awssdk/http/crt/internal/response/InputStreamAdaptingHttpStreamResponseHandler.java
zoewangg: Maybe? I'll update it

TingDaoK: Oh, maybe I am wrong. It means to let backpress to fill up the window...

It's just confusing, maybe something like:
```
Return zero to not update the flow control window, as the window will be updated once the data completes writing into publisher.
```

http-clients/aws-crt-client/src/main/java/software/amazon/awssdk/http/crt/internal/response/ResponseHandlerHelper.java
zoewangg: I considered `closeCrtResource`, but don't feel it's as clear as just `closeConnection` since crt stream and crt connection are just implementation details and they are part of a connection.

http-clients/aws-crt-client/src/main/java/software/amazon/awssdk/http/crt/internal/response/InputStreamAdaptingHttpStreamResponseHandler.java
```diff
@@ -121,16 +128,12 @@ private void onFailedResponseComplete(HttpStream stream, int errorCode) {
...
-        // always close the connection on a 5XX response code.
-        if (HttpStatusFamily.of(responseBuilder.statusCode()) == HttpStatusFamily.SERVER_ERROR) {
-            crtConn.shutdown();
-        }
-
         // For response without a payload, for example, S3 PutObjectResponse, we need to complete the future
         // in onResponseComplete callback since onResponseBody will never be invoked.
         requestCompletionFuture.complete(responseBuilder.build());
+
+        // requestCompletionFuture has been completed at this point, no need to notify hte future
```
TingDaoK, line 135: nitpick: typo
```suggestion
        // requestCompletionFuture has been completed at this point, no need to notify the future
```

TingDaoK (approved):

L-Applin (approved):

PR #4826, opened 2024-01-11
Splitting transformer

L-Applin:
Implementation of the Splitting transformer.

Split an AsyncResponseTranformer into multiple ones, retreived via a Publisher of AsyncResponseTranformer. When requested, the publisher will publish a new AsyncResponseTranformer to its subscriber, but the data received in this individual AsyncResponseTranformer will be buffered and then send to the upstream AsyncResponseTranformer (the one that was split).

Also includes a draft of the subscriber that will use the individually published AsyncResponseTransformer, but is still in a draft state.

The intent is to use the SplittingTransformer to perform multipart get request for s3 objects. The `SplittingTranformer` would publish an `AsyncResponseTranformer` instance when requested by the subscriber, which would perform the call to s3 using that `AsyncResponseTranformer`. By using the `split()` method on an instance of an existing  `AsyncResponseTranformer` to create a `SplittingTranformer`, the behaviour of the splitting tranformer can be fine tuned to the different implementations of the AsyncResponseTranformer interface if required.

## Diagram
Splitting Transformer


services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/multipart/MultipartDownloaderSubscriber.java
```diff
@@ -0,0 +1,107 @@
...
+import org.reactivestreams.Subscriber;
+import org.reactivestreams.Subscription;
+import software.amazon.awssdk.core.async.AsyncResponseTransformer;
+import software.amazon.awssdk.services.s3.S3AsyncClient;
+import software.amazon.awssdk.services.s3.model.GetObjectRequest;
+import software.amazon.awssdk.services.s3.model.GetObjectResponse;
+
+// [WIP]
+// Still work in progress, currently only used to help manual testing, please ignore
+public class MultipartDownloaderSubscriber implements Subscriber<AsyncResponseTransformer<GetObjectResponse, GetObjectResponse>> {
```
L-Applin, line 30: Ignore this class for now, this is only a stub that was used for manually testing with s3.

services/s3/src/test/java/software/amazon/awssdk/services/s3/internal/multipart/MultipartDownloadIntegrationTest.java
```diff
@@ -0,0 +1,156 @@
...
+import software.amazon.awssdk.core.async.ResponsePublisher;
+import software.amazon.awssdk.core.async.SplitAsyncResponseTransformer;
+import software.amazon.awssdk.http.nio.netty.NettyNioAsyncHttpClient;
+import software.amazon.awssdk.regions.Region;
+import software.amazon.awssdk.services.s3.S3AsyncClient;
+import software.amazon.awssdk.services.s3.model.GetObjectRequest;
+import software.amazon.awssdk.services.s3.model.GetObjectResponse;
+
+// WIP - please ignore for now, only used in manually testing
+class MultipartDownloadIntegrationTest {
```
L-Applin, line 41: Ignore for now, only used for manually testing with s3 objects.

core/sdk-core/src/main/java/software/amazon/awssdk/core/async/AsyncResponseTransformer.java
```diff
@@ -114,6 +115,21 @@ public interface AsyncResponseTransformer<ResponseT, ResultT> {
      */
     void exceptionOccurred(Throwable error);
 
+    /**
+     * todo
+     *
+     * @return
+     */
+    default SplitAsyncResponseTransformer<ResponseT, ResultT> split(long bufferSize) {
+        CompletableFuture<ResultT> future = new CompletableFuture<>();
```
cenedhryn, line 124: Why can't the `split()` method return a `SplittingTransformer` instead of a `SplitAsyncResponseTransformer`? Can't the splitting transformer create the future and make it available to the user in the same way?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/SplittingTransformer.java
```diff
@@ -0,0 +1,232 @@
...
+                    publisherToUpstream.complete();
+                    downstreamSubscriber.onComplete();
+                }
+            }
+        });
+    }
+
+    // The AsyncResponseTransformer for each of the individual requests that is sent back to the downstreamSubscriber when
+    // requested. A future is created per request that is completed when onComplete is called on the subscriber for that request
+    // body publisher.
```
cenedhryn, line 142: Nit: reformat the comment using `*`

core/sdk-core/src/test/java/software/amazon/awssdk/core/internal/async/SplittingTransformerTest.java
```diff
@@ -0,0 +1,300 @@
...
+            @Override
+            public void onComplete() {
+                // do nothing, test only
+            }
+        });
+        assertThatThrownBy(future::join).hasMessageContaining("TEST ERROR 2");
+    }
+
+    @Test
+    void whenDataExceedsBufferSize_UpstreamShouldReceiveAllData() {
```
cenedhryn, line 134: Can you refactor this test to be more descriptive? The title is clear, but the code is filled with numbers that don't explain the role each play - as a reader, taking the time to go through and do the math is a bit much
~~~
int bodyLength = 7*1024; <- why this choice and how does it relate to 16*1024?
int amount = 9; <- why this number
SplittingTransformer<TestResultObject, Object> split =
        new SplittingTransformer<>(upstreamTestTransformer, 16*1024, future); <- break out into it's own variable

//Can you break out this function to it's own method? And call it contentPerTransformer (or a better name)
String content =
   IntStream.range(0, bodyLength).mapToObj(i -> String.valueOf(total)).collect(Collectors.joining());

// Again this could reuse that function and make it clear what the expected result is
for (int i = 0; i < amount; i++) {
        int value = i;
        expected.append(IntStream.range(0, bodyLength).mapToObj(j -> String.valueOf(value)).collect(Collectors.joining()));
    }
        
~~~

core/sdk-core/src/test/java/software/amazon/awssdk/core/internal/async/SplittingTransformerTest.java
```diff
@@ -0,0 +1,300 @@
...
+        }
+
+        public String contentAsString() {
+            return content.toString();
+        }
+    }
+
+    private static class TestResultObject {
+        String msg;
+        public TestResultObject(String msg) {
```
cenedhryn, line 291: Nit: newline

core/sdk-core/src/test/java/software/amazon/awssdk/core/internal/async/SplittingTransformerTest.java
```diff
@@ -0,0 +1,300 @@
...
+    @Test
+    void whenRequestingMany_allDemandGetsFulfilled() {
+        UpstreamTestTransformer upstreamTestTransformer = new UpstreamTestTransformer();
+        CompletableFuture<Object> future = new CompletableFuture<>();
+        SplittingTransformer<TestResultObject, Object> split =
+            new SplittingTransformer<>(upstreamTestTransformer, 1024*1024*32, future);
+        split.subscribe(new Subscriber<AsyncResponseTransformer<TestResultObject, TestResultObject>>() {
+            private Subscription subscription;
+            private int received = 0;
+            @Override
```
cenedhryn, line 194: Nit: newline

core/sdk-core/src/test/java/software/amazon/awssdk/core/internal/async/SplittingTransformerTest.java
```diff
@@ -0,0 +1,300 @@
...
+        int bodyLength = 7*1024;
+        int amount = 9;
+        UpstreamTestTransformer upstreamTestTransformer = new UpstreamTestTransformer();
+        CompletableFuture<Object> future = new CompletableFuture<>();
+        SplittingTransformer<TestResultObject, Object> split =
+            new SplittingTransformer<>(upstreamTestTransformer, 16*1024, future);
+        split.subscribe(new Subscriber<AsyncResponseTransformer<TestResultObject, TestResultObject>>() {
+            private Subscription subscription;
+            private int total = 0;
+            @Override
```
cenedhryn, line 144: Nit: newline

utils/src/main/java/software/amazon/awssdk/utils/async/DelegatingBufferingSubscriber.java
```diff
@@ -0,0 +1,137 @@
...
+
+import java.nio.ByteBuffer;
+import java.util.Optional;
+import java.util.concurrent.atomic.AtomicLong;
+import org.reactivestreams.Subscriber;
+import org.reactivestreams.Subscription;
+import software.amazon.awssdk.annotations.SdkInternalApi;
+
+/**
+ * This Subscriber buffers the {@link ByteBuffer} it receives until the specififed maxiimumBuffer is reached, at which point
```
cenedhryn, line 28: Nit: `specififed maxiimumBuffer`

utils/src/main/java/software/amazon/awssdk/utils/async/DelegatingBufferingSubscriber.java
```diff
@@ -0,0 +1,137 @@
...
+    public void onNext(ByteBuffer byteBuffer) {
+        // First, if the incoming ByteBuffer would make the total buffered amount exceed the maximum buffer, send what is
+        // currently buffered in the storage to the delegate.
+        if (currentlyBuffered.get() + byteBuffer.remaining() > maximumBuffer) {
+            flushStorageToDelegate();
+        }
+
+        // Then, if the incoming ByteBuffer would still alone bust the buffer size, remove one chunk of it (of maximumBuffer
+        // size) then try again
+        if (currentlyBuffered.get() + byteBuffer.remaining() > maximumBuffer) {
```
cenedhryn, line 83: This is the same condition as in the clause above.

utils/src/main/java/software/amazon/awssdk/utils/async/DelegatingBufferingSubscriber.java
```diff
@@ -0,0 +1,137 @@
...
+            ByteBuffer chunk = ByteBuffer.allocate(maximumBuffer);
+            while (chunk.hasRemaining()) {
+                chunk.put(byteBuffer.get());
+            }
+            chunk.position(0);
+            sendBytesToStorage(chunk);
+            onNext(byteBuffer);
```
cenedhryn, line 90: Are you supposed to call onNext on yourself?

utils/src/main/java/software/amazon/awssdk/utils/async/DelegatingBufferingSubscriber.java
```diff
@@ -0,0 +1,137 @@
...
+            return;
+        }
+
+        sendBytesToStorage(byteBuffer);
+        if (currentlyBuffered.get() >= maximumBuffer) {
+            flushStorageToDelegate();
+        }
```
cenedhryn, line 97: When would this be needed? Aren't the previous checks sufficient, or could there be extra data added on somehow?

core/sdk-core/src/main/java/software/amazon/awssdk/core/async/AsyncResponseTransformer.java
L-Applin: I was trying to avoid returning the concrete type `SplittingTransformer` and instead return the interface type `SdkPublisher<AsyncResponseTransformer<T>>`. If we return a `SplittingTransformer` here, we would need to make it `@SdkPublicAPI`

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/SplittingTransformer.java
```diff
@@ -0,0 +1,232 @@
...
+        }
+
+        @Override
+        public void exceptionOccurred(Throwable error) {
+            upstreamResponseTransformer.exceptionOccurred(error);
+        }
+    }
+
+    // the Subscriber for each of the individual request ByteBuffer publisher
+    private class IndividualPartSubscriber implements Subscriber<ByteBuffer> {
```
zoewangg, line 192: Same here, can we add TCK tests?

core/sdk-core/src/test/java/software/amazon/awssdk/core/internal/async/SplittingTransformerTest.java
```diff
@@ -0,0 +1,300 @@
...
+class SplittingTransformerTest {
+    private static final Logger log = Logger.loggerFor(SplittingTransformerTest.class);
+
+    @Test
+    void whenSubscriberCancelSubscription_AllDataSentToTransformer() {
+        UpstreamTestTransformer upstreamTestTransformer = new UpstreamTestTransformer();
+        CompletableFuture<Object> future = new CompletableFuture<>();
+        SplittingTransformer<TestResultObject, Object> split =
+            new SplittingTransformer<>(upstreamTestTransformer, 1024*1024*32, future);
+        split.subscribe(new Subscriber<AsyncResponseTransformer<TestResultObject, TestResultObject>>() {
```
zoewangg, line 43: Let's create inner class for this.

core/sdk-core/src/test/java/software/amazon/awssdk/core/internal/async/SplittingTransformerTest.java
```diff
@@ -0,0 +1,300 @@
...
+        assertThat(upstreamTestTransformer.contentAsString()).isEqualTo(expected);
+    }
+
+    @Test
+    void whenSubscriberFailsAttempt_UpstreamTransformerCompletesExceptionally() {
+        UpstreamTestTransformer upstreamTestTransformer = new UpstreamTestTransformer();
+        CompletableFuture<Object> future = new CompletableFuture<>();
+        SplittingTransformer<TestResultObject, Object> split =
+            new SplittingTransformer<>(upstreamTestTransformer, 1024*1024*32, future);
+        split.subscribe(new Subscriber<AsyncResponseTransformer<TestResultObject, TestResultObject>>() {
```
zoewangg, line 92: Same here

zoewangg (commented):
Left some initial comments. Still going through the PR

core/sdk-core/src/main/java/software/amazon/awssdk/core/async/AsyncResponseTransformer.java
```diff
@@ -114,6 +115,21 @@ public interface AsyncResponseTransformer<ResponseT, ResultT> {
      */
     void exceptionOccurred(Throwable error);
 
+    /**
+     * todo
```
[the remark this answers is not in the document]
L-Applin: Some details might change between now and actual release

core/sdk-core/src/test/java/software/amazon/awssdk/core/internal/async/SplittingTransformerTest.java
L-Applin: Added comments and renamed some of the varaible. The idea was to test uneven body size compared to the buffer size

utils/src/main/java/software/amazon/awssdk/utils/async/DelegatingBufferingSubscriber.java
L-Applin: yes, but `currentlyBuffered.get()` and `byteBuffer.remaining()` would have changed

utils/src/main/java/software/amazon/awssdk/utils/async/DelegatingBufferingSubscriber.java
L-Applin: removed recursive call and made it iterative instead

utils/src/main/java/software/amazon/awssdk/utils/async/DelegatingBufferingSubscriber.java
L-Applin: It is needed in the case where the ByteBuffer is itself larger than the maximumBuffer, we need to send chunks of it to the delegate

core/sdk-core/src/main/java/software/amazon/awssdk/core/async/AsyncResponseTransformer.java
zoewangg: Is there any reason`SplitAsyncResponseTransformer` doesn't implement `AsyncResponseTransformer` interface?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/SplittingTransformer.java
```diff
@@ -0,0 +1,335 @@
...
+    }
+
+    /**
+     * Creates an unbounded publisher.
+     */
+    public SplittingTransformer(AsyncResponseTransformer<ResponseT, ResultT> upstreamResponseTransformer,
+                                long bufferSize, CompletableFuture<ResultT> returnFuture) {
+        this.upstreamResponseTransformer = Validate.paramNotNull(upstreamResponseTransformer, "asyncRequestBody");
+        this.returnFuture = Validate.paramNotNull(returnFuture, "returnFuture");
+        this.maximumBufferSize = NumericUtils.saturatedCast(bufferSize);
```
zoewangg, line 139: Can we support long?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/SplittingTransformer.java
```diff
@@ -0,0 +1,335 @@
...
+    private final long maxElements;
+    private final boolean isBounded;
+    private final AtomicInteger totalEmitted = new AtomicInteger(0);
+    private final AtomicBoolean terminated = new AtomicBoolean(false);
+
+    /**
+     * Creates a publisher bounded by the {@code maxElements} provided. As per reactive stream specification, this publisher will
+     * be considered 'unbounded' if {@code maxElements == LONG.MAX_VALUE}.
+     */
+    public SplittingTransformer(AsyncResponseTransformer<ResponseT, ResultT> upstreamResponseTransformer,
```
zoewangg, line 122: Let's use builder

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/SplittingTransformer.java
```diff
@@ -0,0 +1,335 @@
...
+        this.isBounded = false;
+    }
+
+    /**
+     * @param downstreamSubscriber the {@link Subscriber} to the individual AsyncResponseTransformer
+     */
+    @Override
+    public void subscribe(Subscriber<? super AsyncResponseTransformer<ResponseT, ResponseT>> downstreamSubscriber) {
+        if (downstreamSubscriber == null) {
+            // §1.09 Subscribe throw NPE on null subscriber
```
zoewangg, line 151: Is this comment needed? Same as other reactive streams related comments. Seems to clutter the code IMO

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/SplittingTransformer.java
```diff
@@ -96,51 +99,145 @@ public class SplittingTransformer<ResponseT, ResultT> implements SdkPublisher<As
...
+
+    /**
+     * Creates an unbounded publisher.
+     */
     public SplittingTransformer(AsyncResponseTransformer<ResponseT, ResultT> upstreamResponseTransformer,
                                 long bufferSize, CompletableFuture<ResultT> returnFuture) {
         this.upstreamResponseTransformer = Validate.paramNotNull(upstreamResponseTransformer, "asyncRequestBody");
         this.returnFuture = Validate.paramNotNull(returnFuture, "returnFuture");
         this.maximumBufferSize = NumericUtils.saturatedCast(bufferSize);
+        // unbounded by default
```
cenedhryn, line 140: Can this constructor call the new one with `maxElements = Long.MAX_VALUE`?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/SplittingTransformer.java
```diff
@@ -0,0 +1,335 @@
...
+     */
+    private final AtomicLong outstandingDemand = new AtomicLong(0);
+
+    /**
+     * tracks if the publisher is currently sending AsyncResponseTransformer to the downstreamSubscriber
+     */
+    private final AtomicBoolean emitting = new AtomicBoolean(false);
+
+    // 'boundedness' info
+    private final long maxElements;
```
zoewangg, line 113: Is there any reason we need `maxElements`?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/SplittingTransformer.java
```diff
@@ -0,0 +1,335 @@
...
+            }
+            if (!isCancelled.get()) {
+                long newDemand = outstandingDemand.updateAndGet(current -> {
+                    if (Long.MAX_VALUE - current < n) {
+                        return Long.MAX_VALUE;
+                    }
+                    return current + n;
+                });
+                log.trace(() -> String.format("new outstanding demand: %s", newDemand));
+                if (newDemand > 0) {
```
zoewangg, line 188: This check seems unnecessary since existing demand should always be non-negative and we have a new demand here.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/SplittingTransformer.java
```diff
@@ -0,0 +1,335 @@
...
+                    return;
+                }
+                if (isCancelled.get() || terminated.get()) {
+                    return;
+                }
+                outstandingDemand.decrementAndGet();
+                downstreamSubscriber.onNext(new IndividualTransformer());
+                totalEmitted.incrementAndGet();
+            }
+            emitting.compareAndSet(true, false);
```
zoewangg, line 230: We should set it to false in `finally` block.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/SplittingTransformer.java
```diff
@@ -0,0 +1,335 @@
...
+        if (terminated.compareAndSet(false, true)) {
+            publisherToUpstream.complete();
+            downstreamSubscriber.onComplete();
+            // §3.13 Cancel must make the publisher eventually drop all references to the subscriber
+            downstreamSubscriber = null;
+        }
+    }
+
+    private void emit() {
+        if (emitting.compareAndSet(false, true)) {
```
zoewangg, line 216: We should double check the state before we set emitting to false because new items might have come in since we decided to release the flag.

See https://github.com/aws/aws-sdk-java-v2/blob/master/utils/src/main/java/software/amazon/awssdk/utils/async/SimplePublisher.java#L216-L235

zoewangg (commented):
Looking at DelegatingBufferingSubscriber now

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/SplittingTransformer.java
```diff
@@ -96,51 +99,145 @@ public class SplittingTransformer<ResponseT, ResultT> implements SdkPublisher<As
...
      * The amount requested by the downstream subscriber that is still left to fulfill. Updated. when the
-     * {@link Subscription#request(long) request} method is called on the downstream subscriber's subscription.
-     * Corresponds to the number of {@code AsyncResponseTransformer} that will be published to the downstream subscriber.
+     * {@link Subscription#request(long) request} method is called on the downstream subscriber's subscription. Corresponds to the
+     * number of {@code AsyncResponseTransformer} that will be published to the downstream subscriber.
      */
     private final AtomicLong outstandingDemand = new AtomicLong(0);
 
+    /**
+     * tracks if the publisher is currently sending AsyncResponseTransformer to the downstreamSubscriber
```
cenedhryn, line 108: Maybe something like `This flag stops the publisher from emit a new transformer while another transformer is currently in flight`

utils/src/main/java/software/amazon/awssdk/utils/async/DelegatingBufferingSubscriber.java
```diff
@@ -0,0 +1,144 @@
...
+        super.onSubscribe(subscription);
+        this.subscription = subscription;
+        publisherToStorage.subscribe(storage);
+        subscription.request(maximumBuffer);
+    }
+
+    @Override
+    public void onNext(ByteBuffer byteBuffer) {
+        while (bufferOverflows(byteBuffer)) {
+            flushStorageToDelegate();
```
zoewangg, line 81: We need to track downstream subscription; if there's no demand from downstream, we shouldn't call flush storage to delegate

Example: https://github.com/aws/aws-sdk-java-v2/blob/master/utils/src/main/java/software/amazon/awssdk/utils/async/FlatteningSubscriber.java

It's going to be super complex and hard to maintain if we add that logic... Let me see if we can reuse some of the logic in some way.

utils/src/main/java/software/amazon/awssdk/utils/async/DelegatingBufferingSubscriber.java
```diff
@@ -0,0 +1,144 @@
...
+    public void onSubscribe(Subscription subscription) {
+        // §2.05 must call subscription.cancel() if it already has a subscription and receives another on subscribe signal
+        if (this.subscription != null) {
+            subscription.cancel();
+            return;
+        }
+        super.onSubscribe(subscription);
+        this.subscription = subscription;
+        publisherToStorage.subscribe(storage);
+        subscription.request(maximumBuffer);
```
zoewangg, line 75: The number is actually number of ByteBuffer instead of number of bytes.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/SplittingTransformer.java
L-Applin: `SplitAsyncResponseTransformer` is just a holder for the publisher and a completable future, the class implementing `AsyncResponseTransformer` is  `SplittingTransformer.IndividualTransformer`

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/SplittingTransformer.java
L-Applin: To be compliant with **[spec 3.2](https://github.com/reactive-streams/reactive-streams-jvm/blob/v1.0.4/README.md#3.2)**. It seems publisher need to have some kind of upper bound

utils/src/main/java/software/amazon/awssdk/utils/async/DelegatingBufferingSubscriber.java
L-Applin: Right, let me know if you think of something better

core/sdk-core/src/main/java/software/amazon/awssdk/core/async/AsyncResponseTransformer.java
L-Applin: SplitAsyncResponseTransformer is just a holder for the publisher and a completable future, the class implementing AsyncResponseTransformer is SplittingTransformer.IndividualTransformer

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/SplittingTransformer.java
L-Applin: we might need to allocate a ByteBuffer of that size in the case we receive a bigger `ByteBuffer` from the publisher of the individual request body. See [DelegatingBufferingSubscriber#L89](https://github.com/aws/aws-sdk-java-v2/pull/4826/files#diff-d03cd0cb555b1f4448dd14c8559f557fd9ae1b0a0e74b6e4b640441f7cf93058R89)

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/SplittingTransformer.java
L-Applin: we have an on going discussion about this

utils/src/main/java/software/amazon/awssdk/utils/async/DelegatingBufferingSubscriber.java
L-Applin: let me know if you think of a better approach

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/SplittingTransformer.java
zoewangg: It seems the failed test expects the publisher to publish bounded number, so suggesting the following as a workaround. This way, we can remove maxElements. I'd say the implementatoin is compliant; it's just that the test case is not supported in our case.

```java
    @Override
    public Publisher<AsyncResponseTransformer<Object, Object>> createPublisher(long l) {
        CompletableFuture<ResponseBytes<Object>> future = new CompletableFuture<>();
        AsyncResponseTransformer<Object, ResponseBytes<Object>> upstreamTransformer = AsyncResponseTransformer.toBytes();
        SplittingTransformer.Builder<Object, ResponseBytes<Object>> builder =
            SplittingTransformer.<Object, ResponseBytes<Object>>builder()
                                .upstreamResponseTransformer(upstreamTransformer)
                                .bufferSize(64 * 1024)
                                .returnFuture(future);
        return SdkPublisher.adapt(builder.build()).limit(6);
    }

    @Override
    public long maxElementsFromPublisher() {
        return Long.MAX_VALUE;
    }
```

L-Applin: maxElements removed, thanks

core/sdk-core/src/main/java/software/amazon/awssdk/core/async/AsyncResponseTransformer.java
zoewangg: Should we make it implement AsyncResponseTransformer?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/SplittingTransformer.java
```diff
@@ -0,0 +1,341 @@
...
+                }
+                if (outstandingDemand.get() > 0) {
+                    outstandingDemand.decrementAndGet();
+                    downstreamSubscriber.onNext(new IndividualTransformer());
+                }
+            } finally {
+                emitting.compareAndSet(true, false);
+            }
+        } while (outstandingDemand.get() > 0);
+    }
```
zoewangg, line 180: It seems we are delivering one item at a time, suggesting the following for micro performance optimization. 

```
    private void emit() {
        do {
            if (!emitting.compareAndSet(false, true)) {
                return;
            }
            try {
                if (doEmit()) {
                    return;
                }

            } finally {
                emitting.compareAndSet(true, false);
            }
        } while (outstandingDemand.get() > 0);
    }

    private boolean doEmit() {
        long demand = outstandingDemand.get();

        while (demand > 0) {
            if (isCancelled.get()) {
                return true;
            }
            if (outstandingDemand.get() > 0) {
                demand = outstandingDemand.decrementAndGet();
                downstreamSubscriber.onNext(new IndividualTransformer());
            }
        }
        return false;
    }
```

utils/src/main/java/software/amazon/awssdk/utils/async/DelegatingBufferingSubscriber.java
zoewangg: Well, can't think of another way, we just need to implement it :(

core/sdk-core/src/main/java/software/amazon/awssdk/core/async/AsyncResponseTransformer.java
L-Applin: no, I don't think so. The usage would be:
```
SplitAsyncResponseTransformer<GetObjectResponse, GetObjectResponse> split = transformer.split(...);
split.publisher().subscribe(downloaderSubscriber);
CompletableFuture<GetObjectResponse> future = split.preparedFuture();
// do whatever we want with the future
GetObjectResponse response = future.join();
```

zoewangg: It just seems a bit odd that the static factory method returns a class that doesn't implement the same interface. Can we make it implement it? We did the same for `toPublisher` method

https://github.com/aws/aws-sdk-java-v2/blob/master/core/sdk-core/src/main/java/software/amazon/awssdk/core/async/AsyncResponseTransformer.java#L235

L-Applin: `.split()` is not a static method though, it is an instance method called on a specific instance of an `AsyncResponseTransformer` just like the `split()` method on `AsyncRequestBody`. We could instead return just the SdkPublisher<AsyncResponseTransformer<ResponseT, ResponseT>>` like `AsyncRequestBody.split`. I would need to find another way to deal with the future that needs to be returned from the service call.

L-Applin:
To answer most concerns about `DelegatingBufferingSubscriber`, it was modified to reuse the logic in `FlatteningSubscriber` for its implementation

core/sdk-core/src/main/java/software/amazon/awssdk/core/async/AsyncResponseTransformer.java
zoewangg: I see, I'll need to think a bit more on this. Let's keep this conversation open for now.

zoewangg: Can we add a TODO to revisit the naming? It's a bit odd that this class is suffixed with AsyncResponseTransformer but doesn't implement it

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/SplittingTransformer.java
```diff
@@ -0,0 +1,362 @@
...
+        /**
+         * The amount of data in byte this publisher will buffer into memory before sending it to the upstream transformer. The
+         * data will be sent if chunk of {@code maximumBufferSize} to the upstream transformer unless the subscription is
+         * cancelled while less amount is buffered, in which case a chunk with a size less than {@code maximumBufferSize} will be
+         * sent.
+         *
+         * @param maximumBufferSize the amount of data buffered and the size of the chunk of data
+         * @return an instance of this builder
+         */
+        public Builder<ResponseT, ResultT> maximumBufferSize(long maximumBufferSize) {
```
zoewangg, line 339: Should it be `Long`?

utils/src/main/java/software/amazon/awssdk/utils/async/DelegatingBufferingSubscriber.java
```diff
@@ -0,0 +1,121 @@
...
+    protected void fulfillDownstreamDemand() {
+        flushStorageToDelegate();
+    }
+
+    /**
+     * Returns true if we need to call onNext downstream.
+     */
+    @Override
+    protected boolean onNextNeeded() {
+        return super.onNextNeeded() && storage.peek().isPresent() && storage.peek().get().type() == ON_NEXT;
```
zoewangg, line 85: Instead of doing `super.onNextNeeded()`, suggesting using template methods:

In the parent class:

```java
    private boolean onNextNeeded() {
        return downstreamDemand.get() > 0 && additionalOnNextNeededCheck();
    }

/**
  Can be overridden by subclasses to provide additional checks
*/
    protected boolean additionalOnNextNeededCheck() {
        return true;
    }
```

utils/src/main/java/software/amazon/awssdk/utils/async/DelegatingBufferingSubscriber.java
```diff
@@ -0,0 +1,121 @@
...
+     */
+    private final AtomicLong currentlyBuffered = new AtomicLong(0);
+
+    /**
+     * Stores the bytes received from the upstream publisher, awaiting sending them to the delegate once the buffer size is
+     * reached.
+     */
+    private final StoringSubscriber<ByteBuffer> storage = new StoringSubscriber<>(Integer.MAX_VALUE);
+
+    public DelegatingBufferingSubscriber(long maximumBuffer, Subscriber<? super ByteBuffer> subscriber) {
```
zoewangg, line 46: Since this is a protected API, we should use builder

utils/src/main/java/software/amazon/awssdk/utils/async/DelegatingBufferingSubscriber.java
```diff
@@ -0,0 +1,121 @@
...
+        super(subscriber);
+        this.maximumBuffer = maximumBuffer;
```
zoewangg, line 48: nit: `maximumBufferInBytes`

utils/src/main/java/software/amazon/awssdk/utils/async/DelegatingBufferingSubscriber.java
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    @Override
+    public void onNext(ByteBuffer item) {
+        if (currentlyBuffered.get() + item.remaining() > maximumBuffer) {
```
zoewangg, line 53: I think we should trigger a flush as long as we have data instead of buffer is full.

utils/src/main/java/software/amazon/awssdk/utils/async/AbstractFlatteningSubscriber.java
```diff
@@ -0,0 +1,274 @@
...
+        while (onNextNeeded() && !onErrorNeeded()) {
+            fulfillDownstreamDemand();
+        }
+    }
+
+    /**
+     * Returns true if we need to call onNext downstream. If this is executed outside the handling-state-update condition, the
+     * result is subject to change.
+     */
+    protected boolean onNextNeeded() {
```
zoewangg, line 205: See my previous comment, suggesting template methods to allow subclasses to add additional checks. Can we make those methods that are intended to be overridden by subclasses package private?

utils/src/main/java/software/amazon/awssdk/utils/async/AbstractFlatteningSubscriber.java
```diff
@@ -0,0 +1,274 @@
...
+            subscriber.onComplete();
+        }
+    }
+
+    /**
+     * Returns true if we need to call onComplete downstream. If this is executed outside the handling-state-update condition, the
+     * result is subject to change.
+     */
+    protected boolean onCompleteNeeded() {
+        return onCompleteCalledByUpstream && !terminalCallMadeDownstream;
```
zoewangg, line 240: Same here, suggesting template methods.

```
    private boolean onCompleteNeeded() {
        return onCompleteCalledByUpstream && !terminalCallMadeDownstream && additionalOnCompleteNeededCheck() ;
   } 

   boolean additionalOnCompleteNeededCheck() {
     return true;
 } 
```

utils/src/main/java/software/amazon/awssdk/utils/async/AbstractFlatteningSubscriber.java
```diff
@@ -0,0 +1,274 @@
...
+        super(subscriber);
+    }
+
+    /**
+     * This method is called inside the onNext signal. Implementation should do what is required to store the data
+     * before fulfilling the demand from the downstream subscriber.
+     *
+     * @param item the value with which onNext was called.
+     */
+    protected abstract void doWithItem(T item);
```
zoewangg, line 79: I can't think of a better naming for this one, can we make this package private so that we can change it in the future?

PR #4831, opened 2024-01-15
Propagating client apiCallTimeout values to S3Express createSession

cenedhryn:
## Motivation and Context
ApiCallTimeout and ApiCallAttemptTimeout are not set by default on calls. For a nested call such as createSession, this would create an issue by making S3 calls with S3Express authentication unresponsive in case of delayed responses or errors. Therefore, createSession has a default ApiCallTimeout value of 10s. This however can cause the opposite issue, if the user wishes the SDK to try longer to retrieve the credentials. 

This PR lets the S3Express identity cache check if there is a client level apiCallTimeout set, and if it is, use that instead of the default value. 

Note that request overrides for apiCallTimeout (on the outer call, such as getObject, putObject etc) are deliberately not taken into account since the cache is common to the client, and separate request timeout values would affect createSession calls unpredictably. 

## Modifications
- Changes the apiCallTimeout value on the createSession request override to also check for existing client level override

## Testing
- Adds unit testing for honoring the value change
- Adds a functional test for verifying call timeout, retry and exception propagation behavior for nested calls

## Types of changes

## Checklist

## License

### DIFF
--- a/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/s3express/S3ExpressIdentityCache.java
+++ b/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/s3express/S3ExpressIdentityCache.java
@@ -16,12 +16,13 @@
 package software.amazon.awssdk.services.s3.internal.s3express;
 
 import java.time.Duration;
-import java.util.function.Consumer;
+import java.util.Optional;
 import software.amazon.awssdk.annotations.SdkInternalApi;
 import software.amazon.awssdk.auth.credentials.AwsCredentials;
 import software.amazon.awssdk.auth.credentials.CredentialUtils;
 import software.amazon.awssdk.auth.credentials.StaticCredentialsProvider;
 import software.amazon.awssdk.core.SdkClient;
+import software.amazon.awssdk.core.SdkServiceClientConfiguration;
 import software.amazon.awssdk.identity.spi.AwsCredentialsIdentity;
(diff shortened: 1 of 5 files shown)


L-Applin (approved):

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/s3express/S3ExpressIdentityCache.java
```diff
@@ -75,28 +76,42 @@ private CachedS3ExpressCredentials getCachedCredentials(S3ExpressIdentityKey key
...
 
-    private static Consumer<CreateSessionRequest.Builder>
+    private static CreateSessionRequest
             createSessionRequest(String bucket,
-                                 IdentityProvider<AwsCredentialsIdentity> provider) {
-        return r -> r.bucket(bucket)
+                                 IdentityProvider<AwsCredentialsIdentity> provider,
+                                 SdkServiceClientConfiguration serviceClientConfiguration) {
+
+        Duration requestApiCallTimeout = clientSetTimeoutIfExists(serviceClientConfiguration).orElse(DEFAULT_API_CALL_TIMEOUT);
```
millems, line 103: This is fine for now, but we might want to propagate all expected configuration, not just client config.

millems (approved):

mukund-thakur:
@cenedhryn  When are you planning to release this?

Issue #4720, opened 2023-11-24
Cross region S3 file read is not working without a write or list request


yuvanakannan:
### Describe the bug

We are using AWS SDK V2 and CRT to read, list and write files to S3 buckets across regions (cross region enabled)
We are able to download / read the file from the primary region without any issues.
But when we tried to download from the S3 bucket in other region, it fails with error like 
"s.a.a.s.s.m.S3Exception: null (Service: S3, Status Code: 400, Request ID: *****, Extended Request ID: ****)
s.a.a.s.s.m.S3Exception$BuilderImpl.build(S3Exception.java:104)"

But if we issue a list or a write operation, it does not error out. Interestingly if we execute the read/get after a list or write, read /get operation works fine. 

It fails if we are directly reading (get object) file from cross region S3 bucket with out executing a list or write operation.

### Expected Behavior

Download of the file from a cross region S3 bucket should work fine. 

### Current Behavior

It fails if we are directly reading (get object) file from cross region S3 bucket with out executing a list or write operation.

### Reproduction Steps

// Downloading file
S3TransferManager.builder().s3Client(S3AsyncClient.crtBuilder().region(Region.of("us-east-1")).crossRegionAccessEnabled(true) .build()).build().downloadFile(DownloadFileRequest.builder().getObjectRequest(b -> b.bucket(s3BucketInUsw2).key(s3Key)).destination(file).build()).completionFuture().join();

### Possible Solution

As a work around, we are issuing the list operation before reading the file. Also disabling the CRT works.

### Additional Information/Context

aws-crt version 0.27.7. Also tested with latest AWS SDK versions - 2.21.28 and CRT 0.28.8

### AWS Java SDK version used

2.21.4

### JDK version used

17

### Operating System and version

linux

patricia-valdez:
> @yuvanakannan thank for for reaching out, I can repro the error on my side. We'll investigate.
> 
> In the meantime, you can use the workarounds you described. I also found that if you remove the region from the `S3AsyncClient.crtBuilder()`, the `downloadFile` works fine.

I tried removing the region when building the CRT S3 Async Client as suggested although I got below exception. If I set the region anywhere else, I end up with the same issue with download file

[...]

PR #4849, opened 2024-01-25
Fix for issue #4720 , Cross Region enabled Clients created in US-EAST-1 will by internally disable global endpoint and do a regional endpoint call.

joviegas:
## Motivation and Context

Fix for issue #4720 , Cross Region enabled Clients created in US-EAST-1 will by internally disable global endpoint and do a regional endpoint call.

In Version 2 (V2), when calling from "US-EAST-1" and "AWS_GLOBAL" regions, we directly address the global endpoint. For instance:
`Host: mumbai-crossregion-bucket.s3.amazonaws.com`
Here, the region is global, and the signing region is "us-east-1". 

For other client regions (e.g., us-east-2), we send requests to regional endpoints:
`Host: mumbai-crossregion-bucket.us-east-2.s3.amazonaws.com`
In this case, the signing region is "us-east-2". When HeadObject is called on the global endpoint, the service does not throw redirect errors; it redirects only for regional endpoints.

However , the S3 Cross region relies on the service to send the redirect error when a cross region bucket is addressed.
Because of above reason the Cross region features failed to redirect when called from `us-east-1`
 Thus we had to disable "GlobalEndpoints" while doing Cross region enabled calls.

## Modifications

- Updated the BucketEndpointProvider to disable "GlobalEndpoints"

## Testing

- Added Junits
- And tested integ .
- Note that this issue is recreated only when a Object and bucket is there in cloud for a long time.

## Types of changes

## License

### DIFF
--- a/.changes/next-release/bugfix-AmazonSimpleStorageService-4b6b8dd.json
+++ b/.changes/next-release/bugfix-AmazonSimpleStorageService-4b6b8dd.json
@@ -1,0 +1,6 @@
+{
+    "type": "bugfix",
+    "category": "Amazon Simple Storage Service",
+    "contributor": "",
+    "description": "Fix for issue [#4720](https://github.com/aws/aws-sdk-java-v2/issues/4720) , Cross Region enabled Clients created in US-EAST-1 will by internally disable global endpoint and do a regional endpoint call."
+}
--- a/.changes/next-release/bugfix-AmazonSimpleStorageService-4b6b8dd.json
+++ b/.changes/next-release/bugfix-AmazonSimpleStorageService-4b6b8dd.json
@@ -2,5 +2,5 @@
     "type": "bugfix",
     "category": "Amazon Simple Storage Service",
     "contributor": "",
-    "description": "Fix for issue [#4720](https://github.com/aws/aws-sdk-java-v2/issues/4720) , Cross Region enabled Clients created in US-EAST-1 will by internally disable global endpoint and do a regional endpoint call."
+    "description": "S3 client configured with crossRegionEnabled(true) will now use us-east-1 regional endpoint instead of the global endpoint. See [#4720](https://github.com/aws/aws-sdk-java-v2/issues/4720)."
 }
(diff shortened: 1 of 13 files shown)


.changes/next-release/bugfix-AmazonSimpleStorageService-4b6b8dd.json
```diff
@@ -0,0 +1,6 @@
+{
+    "type": "bugfix",
+    "category": "Amazon Simple Storage Service",
+    "contributor": "",
+    "description": "Fix for issue [#4720](https://github.com/aws/aws-sdk-java-v2/issues/4720) , Cross Region enabled Clients created in US-EAST-1 will by internally disable global endpoint and do a regional endpoint call."
```
zoewangg, line 5: Nit, suggestion:
"S3 client configured with `crossRegionEnabled(true)` will now use `us-east-1` regional endpoint instead of the global endpoint. See [#4720](https://github.com/aws/aws-sdk-java-v2/issues/4720)"

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionAsyncClient.java
```diff
@@ -52,23 +52,17 @@ protected <T extends S3Request, ReturnT> CompletableFuture<ReturnT> invokeOperat
...
-                                                        operation.apply(
-                                                            requestWithDecoratedEndpointProvider(
-                                                                userAgentUpdatedRequest,
-                                                                () -> bucketToRegionCache.get(bucketName),
-                                                                serviceClientConfiguration().endpointProvider().get()
-                                                            )
-                                                        ) :
-                                                        operation.apply(userAgentUpdatedRequest);
-
+        CompletableFuture<ReturnT> apiOperationFuture = operation.apply(
```
zoewangg, line 61: Why are we removing `bucketToRegionCache.containsKey(bucketName)` check?

.changes/next-release/bugfix-AmazonSimpleStorageService-4b6b8dd.json
joviegas: Thanks Zoe

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionAsyncClient.java
joviegas: Earlier, for the first calls, since CachedBucket was null, we used to skip overriding the EndpointProvider because we were making the calls as they were. Now, since we need to update the 'useGlobalEndpoint' even for the first call, we need to override the EndpointProvider

zoewangg (approved):

PR #4838, opened 2024-01-18
Endpoint based auth scheme resolver should honor endpoint overrides

sugmanue:
## Motivation and Context

Currently the endpoint based auth scheme provider uses an static default `EndpointProvider`. This means that if the customer overrides the endpoint provider this value will not be used for auth scheme resolution. 

## Modifications

This change stores the endpoint provider in `AuthSchemeParams`, so it's more likely to pass-through if the customer is using decoration. For instance, for S3 we now create a new interface

```java
@Generated("software.amazon.awssdk:codegen")
@SdkInternalApi
public interface S3EndpointResolverAware {
    S3EndpointProvider endpointProvider();

    interface Builder {
        Builder endpointProvider(S3EndpointProvider endpointProvider);
    }
}
```

And the default params now implement this interface as well:

```java
@Generated("software.amazon.awssdk:codegen")
@SdkInternalApi
public final class DefaultS3AuthSchemeParams implements S3AuthSchemeParams, S3EndpointResolverAware {
    ⋮
}
```

Then in the interceptor we create the auth scheme params to always have the endpoint provider:

```java
    private S3AuthSchemeParams authSchemeParams(SdkRequest request, ExecutionAttributes executionAttributes) {
        S3EndpointParams endpointParams = S3ResolveEndpointInterceptor.ruleParams(request, executionAttributes);
        S3AuthSchemeParams.Builder builder = S3AuthSchemeParams.builder();
            ⋮
        if (builder instanceof S3EndpointResolverAware.Builder) {
            EndpointProvider endpointProvider = executionAttributes.getAttribute(SdkInternalExecutionAttribute.ENDPOINT_PROVIDER);
            if (endpointProvider instanceof S3EndpointProvider) {
                ((S3EndpointResolverAware.Builder) builder).endpointProvider((S3EndpointProvider) endpointProvider);
            }
        }
        return builder.build();
    }
```

Then in our default S3 auth scheme provider, we check to see if the auth scheme params has an endpoint provider:

```java
    private S3EndpointProvider endpointProvider(S3AuthSchemeParams params) {
        if (params instanceof S3EndpointResolverAware) {
            S3EndpointResolverAware endpointAwareParams = (S3EndpointResolverAware) params;
            S3EndpointProvider endpointProvider = endpointAwareParams.endpointProvider();
            if (endpointProvider != null) {
                return endpointProvider;
            }
        }
        return DELEGATE;
    }
```

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
[...]

PR #4854, opened 2024-01-27
Update test case where endpoint resolution is checked for atleast once

joviegas:
## Motivation and Context

After PR #4838 Endpoint Resolution is done twice ,thus updated the test case.

## Modifications

Update test case where endpoint resolution is checked for atleast once

## License

### DIFF
--- a/services/s3/src/test/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionAsyncClientTest.java
+++ b/services/s3/src/test/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionAsyncClientTest.java
@@ -17,6 +17,7 @@ package software.amazon.awssdk.services.s3.internal.crossregion;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
+import static org.mockito.Mockito.atLeastOnce;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 import static software.amazon.awssdk.services.s3.internal.crossregion.S3CrossRegionRedirectTestBase.CHANGED_CROSS_REGION;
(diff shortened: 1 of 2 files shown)


sugmanue (approved):

services/s3/src/test/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionAsyncClientTest.java
```diff
@@ -437,18 +438,18 @@ void given_AnyRegion_Client_Updates_the_useGlobalEndpointFlag_asFalse(String reg
...
+                                                .region(Region.of(region))
+                                                .endpointProvider(mockEndpointProvider).build();
         s3Client.getObject(r -> r.bucket(BUCKET).key(KEY), AsyncResponseTransformer.toBytes()).join();
         assertThat(captureInterceptor.endpointProvider).isInstanceOf(BucketEndpointProvider.class);
         ArgumentCaptor<S3EndpointParams> collectionCaptor = ArgumentCaptor.forClass(S3EndpointParams.class);
-        verify(mockEndpointProvider).resolveEndpoint(collectionCaptor.capture());
-        S3EndpointParams resolvedParams = collectionCaptor.getAllValues().get(0);
-        assertThat(resolvedParams.region()).isEqualTo(Region.of(region));
-        assertThat(resolvedParams.useGlobalEndpoint()).isFalse();
+        verify(mockEndpointProvider, atLeastOnce()).resolveEndpoint(collectionCaptor.capture());
```
zoewangg, line 446: nit: should we just verify twice, non blocking, can be a follow up PR.

zoewangg (approved):

services/s3/src/test/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionAsyncClientTest.java
joviegas: Will push this once now to get in release , updating might cause checkbuild to start and cause delay , will create separate PR for times(2)

PR #4856, opened 2024-01-29
S3 should configure signer properties 

sugmanue:
## Motivation and Context

Currently S3 makes use of three different interceptors to configure some signing properties

* [ConfigureSignerInterceptor](https://github.com/aws/aws-sdk-java-v2/blob/c534df35adee2fe5be5abcebbc286537f67087d9/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/handlers/ConfigureSignerInterceptor.java#L29). Adds the following signing properties:
  * `AwsSignerExecutionAttribute.SIGNER_DOUBLE_URL_ENCODE, Boolean.FALSE`
  * `AwsSignerExecutionAttribute.SIGNER_NORMALIZE_PATH, Boolean.FALSE`
* [DisablePayloadSigningInterceptor](https://github.com/aws/aws-sdk-java-v2/blob/c534df35adee2fe5be5abcebbc286537f67087d9/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/handlers/DisablePayloadSigningInterceptor.java#L30). Adds the following signing properties:
  * `S3SignerExecutionAttribute.ENABLE_PAYLOAD_SIGNING, false`
* [EnableChunkedEncodingInterceptor](https://github.com/aws/aws-sdk-java-v2/blob/c534df35adee2fe5be5abcebbc286537f67087d9/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/handlers/EnableChunkedEncodingInterceptor.java#L39) 
  * `S3SignerExecutionAttribute.ENABLE_CHUNKED_ENCODING, enableChunkedEncoding` only for PutObject and UploadPart ops and where enableChunkedEncoding is a `S3Configuration` setting.

### Current State

The modeled auth scheme resolver takes care of `DOUBLE_URL_ENCODE`, `NORMALIZE_PATH`, and, `PAYLOAD_SIGNING_ENABLED`. But it does not set `ENABLE_CHUNKED_ENCODING` for `PutObject` and `UploadPart`.

The endpoint based auth scheme provider only set standard properties for sigv4 and sigv4a and none of the aforementioned ones for S3.

### Expected State

* The interceptors should continue to work as they do today
* The auth scheme provider should not rely on the interceptors to behave correctly

### Codegen Signer Defaults

The codegen code path was changed to allow an [`AuthType`](https://github.com/aws/aws-sdk-java-v2/blob/50d97ac6623a960d0623d9f9168151efdac2d66f/codegen/src/main/java/software/amazon/awssdk/codegen/model/service/AuthType.java#L21) as present in the model to be able to override the modeled auth options for a known value of the enum. For that we included a new class `SigV4SignerDefaults` which we can use to define the signer defaults, for instance for [`s3`](https://github.com/aws/aws-sdk-java-v2/blob/50d97ac6623a960d0623d9f9168151efdac2d66f/codegen/src/main/java/software/amazon/awssdk/codegen/model/service/AuthType.java#L28) auth-type is defined as:

```java
    private static SigV4SignerDefaults s3Defaults() {
        return sigv4Default()
            .toBuilder()
            .authType("s3")
            .service("S3")
            .doubleUrlEncode(Boolean.FALSE)
            .normalizePath(Boolean.FALSE)
            .payloadSigningEnabled(Boolean.FALSE)
            .putOperation("UploadPart",
                          sigv4Default()
                              .toBuilder()
                              // Default S3 signer properties
                              .doubleUrlEncode(Boolean.FALSE)
                              .normalizePath(Boolean.FALSE)
                              .payloadSigningEnabled(Boolean.FALSE)
                              // Including chunkEncodingEnabled TRUE
                              .chunkEncodingEnabled(Boolean.TRUE)
                              .build())
            .putOperation("PutObject",
                          sigv4Default()
                              .toBuilder()
                              // Default S3 signer properties
                              .doubleUrlEncode(Boolean.FALSE)
                              .normalizePath(Boolean.FALSE)
                              .payloadSigningEnabled(Boolean.FALSE)
                              // Including chunkEncodingEnabled TRUE
                              .chunkEncodingEnabled(Boolean.TRUE)
                              .build())
            .build();
    }
```

Notice the use of `.service("s3")` and `putOperation()` calls. The first one makes this auth-type a service-wide override for the modeled auth-scheme if the service name equals the modeled service name. The seconds provides overrides for specific operations.

The [`v4-unsigned-body`](https://github.com/aws/aws-sdk-java-v2/blob/50d97ac6623a960d0623d9f9168151efdac2d66f/codegen/src/main/java/software/amazon/awssdk/codegen/model/service/AuthType.java#L27) auth-type is used across services to disable payload signing per operation. Its signer defaults are encoded as:

```java
    private static SigV4SignerDefaults sigv4UnsignedPayload() {
        return sigv4Default().toBuilder()
                             .authType("v4-unsigned-body")
                             .payloadSigningEnabled(false)
                             .build();
    }

```

### Codegen Auth Scheme Providers

The auth scheme provider for each service is codegen using the auth data in the model and in code. For instance, for event bridge the generated auth scheme provider codegen from the modeled auth looks like:

```java
    @Override
    public List<AuthSchemeOption> resolveAuthScheme(EventBridgeAuthSchemeParams params) {
                .putSignerProperty(AwsV4HttpSigner.SERVICE_SIGNING_NAME, "events")
                .putSignerProperty(AwsV4HttpSigner.REGION_NAME, params.region().id()).build());
    }
```

and for S3 looks like:

```java
    @Override
        switch (params.operation()) {
        case "UploadPart":
        case "PutObject":
                    .putSignerProperty(AwsV4HttpSigner.PAYLOAD_SIGNING_ENABLED, false)
                    .putSignerProperty(AwsV4HttpSigner.NORMALIZE_PATH, false)
                    .putSignerProperty(AwsV4HttpSigner.CHUNK_ENCODING_ENABLED, true).build());
            break;
        default:
                    .putSignerProperty(AwsV4HttpSigner.PAYLOAD_SIGNING_ENABLED, false)
            break;
        }
    }
```

Currently the [`AuthSchemeCodegenMetadata`](https://github.com/aws/aws-sdk-java-v2/blob/50d97ac6623a960d0623d9f9168151efdac2d66f/codegen/src/main/java/software/amazon/awssdk/codegen/poet/auth/scheme/AuthSchemeCodegenMetadata.java#L30) class contains all the data used to codegen instances of auth-options with the configured defaults and provided values The current values that are provided (i.e., non-static) are:

  * `AwsV4HttpSigner.SERVICE_SIGNING_NAME` defaults to the configured service name
  * `AwsV4HttpSigner.REGION_NAME` to `params.region().id(` where `params` is the instance of the codegen auth scheme params.

This class was slightly changed to support constant values explicitly and to allow new instances to be created out of `SigV4SignerDefaults` instances. This removes the hard-coded mapping between `AuthType` to `AuthSchemeCodegenMetadata` in favor of the harcoded mapping of `SigV4SignerDefaults` defined in the `AuthTypeToSigV4Default`. For instance, the S3 defaults can be found like `AuthTypeToSigV4Default.authTypeToDefaults(AuthType(AuthType.S3.value())`. This value can be null if the mapping is not know (e.g., `AuthType.NONE` or `AuthType.BEARER` or `AuthType.CUSTOM`)

### Codegen Endpoint Based Auth Scheme Providers

There are two services that has special treatment for auth scheme providers, Event Bridge and S3. Those services are configured to generate a endpoint-based auth scheme provider. Those services run the endpoint resolution logic and from there derive the auth scheme options. Those services generate a different auth scheme interceptor.

```java
    @Override
    public List<AuthSchemeOption> resolveAuthScheme(EventBridgeAuthSchemeParams params) {
        EventBridgeEndpointParams endpointParameters = EventBridgeEndpointParams.builder().region(params.region())
                .endpointId(params.endpointId()).build();
        Endpoint endpoint = CompletableFutureUtils.joinLikeSync(endpointProvider(params).resolveEndpoint(endpointParameters));
        List<EndpointAuthScheme> authSchemes = endpoint.attribute(AwsEndpointAttribute.AUTH_SCHEMES);
        if (authSchemes == null) {
            return MODELED_RESOLVER.resolveAuthScheme(params);
        }
        for (EndpointAuthScheme authScheme : authSchemes) {
            String name = authScheme.name();
            switch (name) {
            case "sigv4":
                SigV4AuthScheme sigv4AuthScheme = Validate.isInstanceOf(SigV4AuthScheme.class, authScheme,
                        "Expecting auth scheme of class SigV4AuthScheme, got instead object of class %s", authScheme.getClass()
                                .getName());
                AuthSchemeOption sigv4AuthSchemeOption = AuthSchemeOption.builder().schemeId(AwsV4AuthScheme.SCHEME_ID)
                        .putSignerProperty(AwsV4HttpSigner.SERVICE_SIGNING_NAME, sigv4AuthScheme.signingName())
                        .putSignerProperty(AwsV4HttpSigner.REGION_NAME, sigv4AuthScheme.signingRegion())
                        .putSignerProperty(AwsV4HttpSigner.DOUBLE_URL_ENCODE, !sigv4AuthScheme.disableDoubleEncoding()).build();
                options.add(sigv4AuthSchemeOption);
                break;
            case "sigv4a":
                SigV4aAuthScheme sigv4aAuthScheme = Validate.isInstanceOf(SigV4aAuthScheme.class, authScheme,
                        "Expecting auth scheme of class SigV4AuthScheme, got instead object of class %s", authScheme.getClass()
                                .getName());
                RegionSet regionSet = RegionSet.create(sigv4aAuthScheme.signingRegionSet());
                AuthSchemeOption sigv4aAuthSchemeOption = AuthSchemeOption.builder().schemeId(AwsV4aAuthScheme.SCHEME_ID)
                        .putSignerProperty(AwsV4HttpSigner.SERVICE_SIGNING_NAME, sigv4aAuthScheme.signingName())
                        .putSignerProperty(AwsV4aHttpSigner.REGION_SET, regionSet)
                        .putSignerProperty(AwsV4HttpSigner.DOUBLE_URL_ENCODE, !sigv4aAuthScheme.disableDoubleEncoding()).build();
                options.add(sigv4aAuthSchemeOption);
                break;
            default:
                throw new IllegalArgumentException("Unknown auth scheme: " + name);
            }
        }
    }
```

This class was changed to allow services with non-standard default to "apply" those after the endpoint-based auth scheme has been selected. For S3 we will call the following method

```java
    private static AuthSchemeOption.Builder applySigV4FamilyDefaults(AuthSchemeOption.Builder option, S3AuthSchemeParams params) {
        switch (params.operation()) {
        case "UploadPart":
        case "PutObject":
            return option.putSignerPropertyIfAbsent(AwsV4HttpSigner.PAYLOAD_SIGNING_ENABLED, false)
                    .putSignerPropertyIfAbsent(AwsV4HttpSigner.NORMALIZE_PATH, false)
                    .putSignerPropertyIfAbsent(AwsV4HttpSigner.CHUNK_ENCODING_ENABLED, true);
        default:
            return option.putSignerPropertyIfAbsent(AwsV4HttpSigner.PAYLOAD_SIGNING_ENABLED, false).putSignerPropertyIfAbsent(
                    AwsV4HttpSigner.NORMALIZE_PATH, false);
        }
    }

```

For services that do no have any special Sigv4 signing attributes this method won't be codegen.

## Testing

* Modified existing unit tests
* Added unit tests for new the new logic
* Ran unit and integration testing for S3

## Types of changes

## Checklist

## License


codegen/src/main/java/software/amazon/awssdk/codegen/poet/auth/scheme/SigV4SignerDefaults.java
```diff
@@ -0,0 +1,240 @@
...
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+import software.amazon.awssdk.utils.Validate;
+
+/**
+ * Tracks a set of explicitly enabled signer properties for the family of AWS SigV4 signers. The currently supported attributes
+ * are {@code doubleUrlEncode}, {@code normalizePath}, {@code payloadSigningEnabled}, {@code chunkEncodingEnabled}. If their
+ * value is not null then is taken as-is, otherwise is ignored as in the value is not overridden. An auth type can also
```
cenedhryn, line 27: > value is not null then is taken as-is, otherwise is ignored as in the value is not overridden

Is this another way of saying that if it's null, we won't override? A simpler language would make the statement easier to understand.

codegen/src/main/java/software/amazon/awssdk/codegen/poet/builder/BaseClientBuilderClass.java
```diff
@@ -784,14 +785,32 @@ private MethodSpec internalPluginsMethod() {
         builder.addStatement("$T internalPlugins = new $T<>()", parameterizedTypeName,  ArrayList.class);
 
         for (String internalPlugin : internalPlugins) {
-            ClassName pluginClass = ClassName.bestGuess(internalPlugin);
-            builder.addStatement("internalPlugins.add(new $T())", pluginClass);
+            String arguments = internalPluginNewArguments(internalPlugin);
+            String internalPluginClass = internalPluginClass(internalPlugin);
+            ClassName pluginClass = ClassName.bestGuess(internalPluginClass);
+            builder.addStatement("internalPlugins.add(new $T($L))", pluginClass, arguments);
```
cenedhryn, line 791: What is the purpose of these changes? Can you point me to where it's being used?

sugmanue: The purpose of the change is allow internal plugins to take a dependency on `SdkClientConfiguration`. The internal plugins method was changed as

```java
    private List<SdkPlugin> internalPlugins(SdkClientConfiguration config) {
        ⋮
    }
```
If the internal plugin does not care for the config then it should be fine, but if it does (by adding `(config)`) to the customization then it will pass down the config to the constructor. For S3 this is currently codegen as:

```java
    private List<SdkPlugin> internalPlugins(SdkClientConfiguration config) {
        List<SdkPlugin> internalPlugins = new ArrayList<>();
        internalPlugins.add(new S3DisableChunkEncodingIfConfiguredPlugin((config)));
        internalPlugins.add(new S3ExpressPlugin());
        return internalPlugins;
    }
```

The idea behind this is that we want to avoid the cost of wrapping the auth scheme provider (done by  `S3DisableChunkEncodingIfConfiguredPlugin`) for those customers that don't care about it (i.e., they didn't change `chunkedEncodingEnabled` in the `S3Configuration` instance). For that case the plugin should do nothing (i.e., I won't configure the auth scheme provider), but if the setting is set and is not its default value then we will install the wrapping auth-scheme provider. The plugins themselves don't have access to `SdkClientConfiguration` so in order to get access to it the configuration object we need and interceptor + a plugin. 

The constructor of the plugin does this:

```java
    public S3DisableChunkEncodingIfConfiguredPlugin(SdkClientConfiguration config) {
        S3Configuration serviceConfiguration =
            (S3Configuration) config.option(SdkClientOption.SERVICE_CONFIGURATION);

        boolean isServiceConfigurationPresent = serviceConfiguration != null;
        boolean shouldAddDisableChunkEncoding = false;
        boolean isChunkedEncodingEnabledConfigured = false;
        boolean isChunkedEncodingEnabledDisabled = false;
        boolean configuresDisableChunkEncoding = false;
        if (isServiceConfigurationPresent) {
            isChunkedEncodingEnabledConfigured = serviceConfiguration.toBuilder().chunkedEncodingEnabled() != null;
            isChunkedEncodingEnabledDisabled = !serviceConfiguration.chunkedEncodingEnabled();
            configuresDisableChunkEncoding = isChunkedEncodingEnabledConfigured && isChunkedEncodingEnabledDisabled;
            if (configuresDisableChunkEncoding) {
                shouldAddDisableChunkEncoding = true;
            }
        }
        this.configuresDisableChunkEncoding = shouldAddDisableChunkEncoding;
        this.isChunkedEncodingEnabledConfigured = isChunkedEncodingEnabledConfigured;
        this.isChunkedEncodingEnabledDisabled = isChunkedEncodingEnabledDisabled;
        this.isServiceConfigurationPresent = isServiceConfigurationPresent;
    }
```

And then the `configureClient` checks for the `configuresDisableChunkEncoding` flags and sets the wrapper accordingly, e.g.,

```java
    @Override
    public void configureClient(SdkServiceClientConfiguration.Builder config) {
        if (configuresDisableChunkEncoding) {
            LOG.debug(() -> String.format("chunkedEncodingEnabled was explicitly disabled in the configuration, adding "
                                          + "`S3DisableChunkEncodingAuthSchemeProvider` auth provider wrapper."));
            S3ServiceClientConfiguration.Builder s3Config = (S3ServiceClientConfiguration.Builder) config;

            // We wrap the default S3AuthSchemeProvider with a S3DisableChunkEncodingAuthSchemeProvider
            // instance that sets the AwsV4FamilyHttpSigner.CHUNK_ENCODING_ENABLED signer
            // property to false but only if the operations is `"PutObject" or "UploadPart"`
            // This legacy logic was implemented before using an interceptor but now requires
            // wrapping the S3AuthSchemeProvider for it to work.
            S3AuthSchemeProvider disablingAuthSchemeProvider =
                S3DisableChunkEncodingAuthSchemeProvider.create(s3Config.authSchemeProvider());
            s3Config.authSchemeProvider(disablingAuthSchemeProvider);
        } else {
            LOG.debug(() -> String.format("chunkedEncodingEnabled was not explicitly disabled in the configuration, not adding "
                                          + "the `S3DisableChunkEncodingAuthSchemeProvider` auth provider wrapper."
                                          + "[isServiceConfigurationPresent: %s, isChunkedEncodingEnabledConfigured: %s,  "
                                          + "isChunkedEncodingEnabledDisabled: %s]",
                                          isServiceConfigurationPresent,
                                          isChunkedEncodingEnabledConfigured,
                                          isChunkedEncodingEnabledDisabled));
        }
    }
```

Let me know if you know of a better way to accomplish this task.

codegen/src/main/java/software/amazon/awssdk/codegen/poet/auth/scheme/AuthSchemeSpecUtils.java
```diff
@@ -188,6 +191,133 @@ public Map<List<String>, List<AuthType>> operationsToAuthType() {
...
+        }
+        return operationsToModeledMetadata();
+    }
+
+    /**
+     * Similar to {@link #operationsToModeledMetadata()} computes a map from operations to codegen metadata objects. The
+     * service default list of codegen metadata is keyed with {@link Collections#emptyList()}.
+     *
+     * This map is used to codegen switch statements.
+     */
```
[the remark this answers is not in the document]
sugmanue: This javadoc is for `computeServiceWideDefaults`, `computeServiceWideDefaults` has its own. I can rephrase it if you find it confusing, let me know.

codegen/src/main/java/software/amazon/awssdk/codegen/poet/auth/scheme/EndpointBasedAuthSchemeProviderSpec.java
```diff
@@ -62,18 +63,25 @@ public ClassName className() {
...
-                        .addSuperinterface(authSchemeSpecUtils.providerInterfaceName())
-                        .addMethod(constructor())
-                        .addField(defaultInstance())
-                        .addField(modeledResolverInstance())
-                        .addField(endpointDelegateInstance())
-                        .addMethod(createMethod())
-                        .addMethod(resolveAuthSchemeMethod())
-                        .addMethod(endpointProvider())
-                        .build();
+        Map<List<String>, AuthSchemeCodegenMetadata> operationsToSigv4 =
```
cenedhryn, line 66: I'm surprised there are no codegen tests to cover this class or these changes. Can we add?

codegen/src/main/java/software/amazon/awssdk/codegen/poet/builder/BaseClientBuilderClass.java
cenedhryn: seems fine for now

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/s3express/S3DisableChunkEncodingAuthSchemeProvider.java
```diff
@@ -0,0 +1,67 @@
...
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.services.s3.internal.s3express;
```
cenedhryn, line 16: Why is this class added in the s3express package? Seems it should be in `plugins`

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/plugins/S3DisableChunkEncodingIfConfiguredPlugin.java
```diff
@@ -0,0 +1,93 @@
...
+        if (configuresDisableChunkEncoding) {
+            LOG.debug(() -> String.format("chunkedEncodingEnabled was explicitly disabled in the configuration, adding "
+                                          + "`S3DisableChunkEncodingAuthSchemeProvider` auth provider wrapper."));
+            S3ServiceClientConfiguration.Builder s3Config = (S3ServiceClientConfiguration.Builder) config;
+
+            // We wrap the default S3AuthSchemeProvider with a S3DisableChunkEncodingAuthSchemeProvider
+            // instance that sets the AwsV4FamilyHttpSigner.CHUNK_ENCODING_ENABLED signer
+            // property to false but only if the operations is `"PutObject" or "UploadPart"`
+            // This legacy logic was implemented before using an interceptor but now requires
+            // wrapping the S3AuthSchemeProvider for it to work.
```
cenedhryn, line 79: Can we remove this comment and maybe add the legacy part to the class doc instead?

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/plugins/S3DisableChunkEncodingIfConfiguredPlugin.java
```diff
@@ -0,0 +1,93 @@
...
+            S3AuthSchemeProvider disablingAuthSchemeProvider =
+                S3DisableChunkEncodingAuthSchemeProvider.create(s3Config.authSchemeProvider());
+            s3Config.authSchemeProvider(disablingAuthSchemeProvider);
+        } else {
+            LOG.debug(() -> String.format("chunkedEncodingEnabled was not explicitly disabled in the configuration, not adding "
```
cenedhryn, line 84: Do we really need this logging?

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/plugins/S3OverrideAuthSchemePropertiesPlugin.java
```diff
@@ -0,0 +1,180 @@
...
+ * {@snippet
+ * s3Client.putObject(PutObjectRequest.builder()
+ *                            .overrideConfiguration(c -> c.addPlugin(
+ *                                S3OverrideAuthSchemePropertiesPlugin.enablePayloadSigningPlugin()))
+ *                            .checksumAlgorithm(ChecksumAlgorithm.SHA256)
+ *                            .bucket("test").key("test").build(), RequestBody.fromBytes("abc".getBytes()));
+ * }
+ */
+@SdkProtectedApi
+public final class S3OverrideAuthSchemePropertiesPlugin implements SdkPlugin {
```
cenedhryn, line 59: Can we add a unit test for this class?

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/plugins/S3DisableChunkEncodingIfConfiguredPlugin.java
sugmanue: Probably not, I can remove it.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/s3express/S3DisableChunkEncodingAuthSchemeProvider.java
sugmanue: Correct, my bad. I will move it accordingly.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/plugins/S3OverrideAuthSchemePropertiesPlugin.java
sugmanue: Added.

codegen/src/main/java/software/amazon/awssdk/codegen/poet/auth/scheme/EndpointBasedAuthSchemeProviderSpec.java
cenedhryn: Discussed offline - we can't test in codegen because the auth type / service is hardcoded as "s3"

cenedhryn (approved):

PR #4858, opened 2024-01-29
Reduce memory usage for chunk-encoded streaming uploads, like those used by flexible checksums in S3.

millems:
Before this change, our chunk encoding logic would copy customer data five times:
1. [From the customer's stream into a byte array.](https://github.com/aws/aws-sdk-java-v2/blob/6040b2be6731e4b5ef64e775a2cfffb07d76766c/core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/signer/chunkedencoding/ChunkedEncodedInputStream.java#L106-L107)
2. [From the byte array into a slightly smaller byte array.](https://github.com/aws/aws-sdk-java-v2/blob/6040b2be6731e4b5ef64e775a2cfffb07d76766c/core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/signer/chunkedencoding/ChunkedEncodedInputStream.java#L111)
3. [From the smaller byte array into a byte array output stream.](https://github.com/aws/aws-sdk-java-v2/blob/6040b2be6731e4b5ef64e775a2cfffb07d76766c/core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/signer/chunkedencoding/ChunkedEncodedInputStream.java#L171)
4. [From the byte array output stream into an array.](https://github.com/aws/aws-sdk-java-v2/blob/6040b2be6731e4b5ef64e775a2cfffb07d76766c/core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/signer/chunkedencoding/ChunkedEncodedInputStream.java#L149)
5. [From the array into the output array.](https://github.com/aws/aws-sdk-java-v2/blob/6040b2be6731e4b5ef64e775a2cfffb07d76766c/core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/signer/chunkedencoding/ChunkedEncodedInputStream.java#L85)

After this change, the logic will copy the data twice:
1. From the customer's stream into a byte array.
2. From the byte array into the output array.

There's a path to make it only one copy, but it requires the chunk encoded input stream to know the length of the underlying stream so that it can detect when the last chunk will be encountered. This will require additional piping, so we can do it in a follow-up PR.

Before this change:
Flame graph showing 159.78 MB usage for putObject before change

After this change:
Flame graph showing 81.67 MB usage for putObject after change

### DIFF
--- a/core/http-auth-aws/src/test/java/software/amazon/awssdk/http/auth/aws/internal/signer/chunkedencoding/ChunkedEncodedInputStreamTest.java
+++ b/core/http-auth-aws/src/test/java/software/amazon/awssdk/http/auth/aws/internal/signer/chunkedencoding/ChunkedEncodedInputStreamTest.java
@@ -54,7 +54,7 @@ public class ChunkedEncodedInputStreamTest {
             .builder()
             .inputStream(payload)
             .chunkSize(chunkSize)
-            .header(chunk -> Integer.toHexString(chunk.length).getBytes())
+            .header(chunk -> Integer.toHexString(chunk.remaining()).getBytes())
             .build();
 
         byte[] tmp = new byte[64];
@@ -90,7 +90,7 @@ public class ChunkedEncodedInputStreamTest {
             .builder()
             .inputStream(payload)
             .chunkSize(chunkSize)
-            .header(chunk -> Integer.toHexString(chunk.length).getBytes())
+            .header(chunk -> Integer.toHexString(chunk.remaining()).getBytes())
             .extensions(Collections.singletonList(helloWorldExt))
             .build();
 
@@ -128,7 +128,7 @@ public class ChunkedEncodedInputStreamTest {
             .builder()
             .inputStream(payload)
             .chunkSize(chunkSize)
-            .header(chunk -> Integer.toHexString(chunk.length).getBytes())
+            .header(chunk -> Integer.toHexString(chunk.remaining()).getBytes())
             .trailers(Collections.singletonList(helloWorldTrailer))
             .build();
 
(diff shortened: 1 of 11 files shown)


s3-benchmark-result.txt
```diff
@@ -0,0 +1,20 @@
+12:33:55.087 [main] INFO  S3BenchmarkRunner - Running benchmarks against S3 Express
```
cenedhryn, line 1: unintended addition?

core/http-auth-aws/src/test/java/software/amazon/awssdk/http/auth/aws/internal/signer/chunkedencoding/ChunkedEncodedInputStreamTest.java
```diff
@@ -274,8 +274,11 @@ public void ChunkEncodedInputStream_withAwsParams_returnsAwsSignedAndEncodedChun
                            "\r\n").getBytes(StandardCharsets.UTF_8)
         );
 
-        assertEquals(expectedBytesRead, bytesRead);
+        // System.out.println("Expected: " + Arrays.toString(expected.toByteArray()));
+        // System.out.println("Actual: " + Arrays.toString(actualBytes));
```
cenedhryn, line 278: clean up

s3-benchmark-result.txt
millems: Good catch, will fix!

cenedhryn (approved):

PR #4867, opened 2024-01-31
Performance improvement for sigv4 signing.

millems:
1. When trimming and removing consecutive spaces during sigv4 normalization, copy word-by-word instead of character-by-character. This reduces the overhead of range and encoding checks in string builder.
2. Increase starting string builder size for canonical headers, to limit resizing (2048 worked well for DynamoDB's get-item).
3. Use a switch statement for whitespace checks instead of consecutive if statements. On my compiler, the switch statement compiles to a jump table which runs quicker.


core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/signer/V4CanonicalRequest.java
```diff
@@ -181,7 +181,7 @@ public static List<Pair<String, List<String>>> getCanonicalHeaders(Map<String, L
      * Each header-value pair is separated by a newline.
      */
     public static String getCanonicalHeadersString(List<Pair<String, List<String>>> canonicalHeaders) {
-        StringBuilder result = new StringBuilder(512);
+        StringBuilder result = new StringBuilder(2048);
```
zoewangg, line 184: Can we add a note about why we chose this number?

zoewangg (approved):

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/signer/V4CanonicalRequest.java
millems: Done

PR #4880, opened 2024-02-01
Expose futureCompletionExecutor on S3 CRT client

### LINKED ISSUE #4841: Allow to override FUTURE_COMPLETION_EXECUTOR for S3Crt client
### Describe the feature

`S3AsyncClientBuilder` allows to configure the executor used to complete future. This configuration is not exposed on the `S3CrtAsyncClientBuilder`

### Use Case

In a quarkus application, futures are executed on a thread with a different TCCL and I encounter class not found exception when CDI proxy tries to create the real object from within the default `sdk-async-response` executor.

### Proposed Solution

Expose `advancedOption(SdkAdvancedAsyncClientOption.FUTURE_COMPLETION_EXECUTOR, executor)` somewhere in `S3CrtAsyncClientBuilder`.

### Other Information

I think that `SCHEDULED_EXECUTOR_SERVICE` should also be configured.

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### AWS Java SDK version used

2.23.6

### JDK version used

17

linux

zoewangg:
## Motivation and Context
Close #4841 
Allow users to configure future completion executor on the AWS CRT-based S3 client via `S3CrtAsyncClientBuilder#futureCompletionExecutor`

## Modifications
Expose `futureCompletionExecutor` on S3 CRT client

I considered exposing `ClientAsyncConfiguration` directly as how we do it on the standard Java client and didn't go with that approach is because if we add other advanced async configuration that's not supported in CRT based S3 client in the future, customers may find it surprising that the same option don't work with CRT based S3 client.

## Testing
Added new unit test

## Types of changes

## Checklist

## License


services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/DefaultS3CrtAsyncClient.java
```diff
@@ -120,22 +125,29 @@ private static S3AsyncClient initializeS3AsyncClient(DefaultS3CrtClientBuilder b
...
+                                                                 .accelerate(builder.accelerate)
+                                                                 .forcePathStyle(builder.forcePathStyle)
+                                                                 .crossRegionAccessEnabled(builder.crossRegionAccessEnabled)
+                                                                 .putAuthScheme(new CrtS3ExpressNoOpAuthScheme())
+                                                                 .httpClientBuilder(initializeS3CrtAsyncHttpClient(builder));
+
+
+        if (builder.futureCompletionExecutor != null) {
+            s3AsyncClientBuilder.asyncConfiguration(b -> b.advancedOption(FUTURE_COMPLETION_EXECUTOR,
+                                                                          builder.futureCompletionExecutor).build());
```
L-Applin, line 148: No need to call build on the `ClientAsyncConfiguration.Builder` in the lambda here

zoewangg: Good catch, fixed

L-Applin (approved):

PR #4899, opened 2024-02-08
Enable setting multipart to true/false by defining it as a custom client context parameter

anirudh9391:
Multipart upload can be set as a client configuration parameter and is accessible at the time of instantiation. This allows classes that instantiate the S3AsyncClient to validate this parameter. 

## Modifications
1)Incorporated multiPartEnabled as a custom client configuration parameter by defining it in customization.config. The codegen module generates getter and setter methods for this parameter. 
2) By default, set multipartEnabled = true in the S3AsyncClient
3) Removed code that sets multiPart Enabled on the client directly

## Testing
Unit Tests to validate that 
1) Disabling multiPart code path logs warnings
2) Default S3AsyncClient has multiPartEnabled = true by default. 
3) Setting MultiPartEnabled = true does not log warnings

## Types of changes

## Checklist

## License

- [ x] I confirm that this pull request can be released under the Apache 2 license


services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/TransferManagerFactory.java
```diff
@@ -68,7 +64,7 @@ private static Supplier<S3AsyncClient> defaultS3AsyncClient() {
         if (crtInClasspath()) {
             return S3AsyncClient::crtCreate;
         }
-        return S3AsyncClient::create;
+        return S3AsyncClient::createWithMultipartEnabled;
```
L-Applin, line 67:
```suggestion
        return () -> S3AsyncClient.builder().multipartEnabled(true).build();
```
Could we avoid creating the new static method `createWithMultipartEnabled ` on the client and just use this instead?

services/s3/src/main/resources/codegen-resources/customization.config
```diff
@@ -229,6 +229,14 @@
...
         "statement": "builder().build()",
         "javaDoc": "Create a CRT-based {@link S3AsyncClient} with default settings"
+      },
+      {
+        "methodName": "createWithMultipartEnabled",
+        "clientType": "async",
+        "instanceType": "software.amazon.awssdk.services.s3.S3AsyncClient",
+        "returnType": "software.amazon.awssdk.services.s3.S3AsyncClient",
+        "statement": "builder().multipartEnabled(true).build()",
+        "javaDoc": "Create a {@link S3AsyncClient} with multi-part operations enabled"
```
L-Applin, line 239: see other comment about static method

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/client/S3AsyncClientDecorator.java
```diff
@@ -33,8 +33,6 @@
 public class S3AsyncClientDecorator {
     public static final AttributeMap.Key<MultipartConfiguration> MULTIPART_CONFIGURATION_KEY =
         new AttributeMap.Key<MultipartConfiguration>(MultipartConfiguration.class){};
-    public static final AttributeMap.Key<Boolean> MULTIPART_ENABLED_KEY =
-        new AttributeMap.Key<Boolean>(Boolean.class){};
```
L-Applin, line 37: If we dont need it anymore, we should also remove it from customization.config: `"contextParamConfigKey": "S3AsyncClientDecorator.MULTIPART_CONFIGURATION_KEY"` **[ln245](https://github.com/aws/aws-sdk-java-v2/pull/4899/files#diff-d517e448d4e73d698fba58683bb9144f65ff5aa70e033cd1b19dbfbd36d101daR253)**

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/TransferManagerFactory.java
anirudh9391: Yeah i contemplated using this, just thought it might be neater to create a new one. Sure, i can change it to this, it makes sense to have one lesser static method to have to test

services/s3/src/main/resources/codegen-resources/customization.config
anirudh9391: Got it. I can change this

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/client/S3AsyncClientDecorator.java
anirudh9391: Ah okay. I missed that one.

anirudh9391: Removed it

PR #4920, opened 2024-02-12
Add content header if the Request doesnot has one for CRT sync Http clients

joviegas:
## Motivation and Context

For request with blob payload
https://github.com/aws/aws-sdk-java-v2/blob/6671c79f763405e245bf96c299061c17e22f49ab/services/sagemakerruntime/src/main/resources/codegen-resources/service-2.json#L300-L302

- The signature mismatch error was seen at the server side since the content-length header was not present

- This issue is only seen for AWS Sync CRT client.
- Other clients like url. Apache added the content length at the end , just before sending the request

## Modifications

- Added code to calculate and add content length for Binary payload and String payload in Marshallers

## Testing

- Added Junits

## Types of changes

## License


http-clients/aws-crt-client/src/main/java/software/amazon/awssdk/http/crt/internal/request/CrtRequestAdapter.java
```diff
@@ -131,11 +134,22 @@ private static List<HttpHeader> createHttpHeaderList(URI uri, HttpExecuteRequest
...
 
-        // Add the rest of the Headers
         sdkRequest.forEachHeader((key, value) -> value.stream().map(val -> new HttpHeader(key, val)).forEach(crtHeaderList::add));
 
         return crtHeaderList;
     }
+
+    public static long calculateContentLength(ContentStreamProvider contentStreamProvider) {
+        try {
+            return  IoUtils.toByteArray(contentStreamProvider.newStream()).length ;
```
L-Applin, line 150: Wont that buffer the whole body content in memory? Do we really want to do that?

joviegas: Need to rework on this , due to internal comment from Matthew miller to handle this in Marshallers.  Will update the PR with that approach.

core/protocols/aws-json-protocol/src/main/java/software/amazon/awssdk/protocols/json/internal/marshall/JsonProtocolMarshaller.java
```diff
@@ -181,12 +181,16 @@ void doMarshall(SdkPojo pojo) {
             Object val = field.getValueOrDefault(pojo);
             if (isExplicitBinaryPayload(field)) {
                 if (val != null) {
-                    request.contentStreamProvider(((SdkBytes) val)::asInputStream);
+                    SdkBytes sdkBytes = (SdkBytes) val;
+                    request.contentStreamProvider(sdkBytes::asInputStream);
+                    updateContentLengthHeader(sdkBytes.asByteArray().length);
```
millems, line 186: `asByteArrayUnsafe()`?

core/protocols/aws-json-protocol/src/main/java/software/amazon/awssdk/protocols/json/internal/marshall/JsonProtocolMarshaller.java
```diff
@@ -196,6 +200,12 @@ void doMarshall(SdkPojo pojo) {
         }
     }
 
+    private void updateContentLengthHeader(int contentLength) {
+        if (!request.firstMatchingHeader(CONTENT_LENGTH).isPresent()) {
+            request.putHeader(CONTENT_LENGTH, Integer.toString(contentLength));
+        }
```
millems, line 206: When will this condition be needed? Why not just set it?

core/protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/internal/marshall/XmlProtocolMarshaller.java
```diff
@@ -91,8 +91,11 @@ void doMarshall(SdkPojo pojo) {
             Object val = field.getValueOrDefault(pojo);
 
             if (isBinary(field, val)) {
-                request.contentStreamProvider(((SdkBytes) val)::asInputStream);
+                SdkBytes sdkBytes = (SdkBytes) val;
+                request.contentStreamProvider(sdkBytes::asInputStream);
                 setContentTypeHeaderIfNeeded("binary/octet-stream");
+                request.putHeader(CONTENT_LENGTH, Integer.toString(sdkBytes.asByteArray().length));
```
millems, line 97: `asByteArrayUnsafe()`?

core/protocols/aws-json-protocol/src/main/java/software/amazon/awssdk/protocols/json/internal/marshall/JsonProtocolMarshaller.java
joviegas: Thank you for catching that oversight. I assumed that "MergeCustomHeadersStage" would override the headers before this. However, this stage comes after Marshalling is done. I removed this check.

core/protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/internal/marshall/XmlProtocolMarshaller.java
joviegas: done

test/protocol-tests/src/test/java/software/amazon/awssdk/protocol/tests/contentlength/MarshallersAddContentLengthTest.java
```diff
@@ -0,0 +1,146 @@
...
+import software.amazon.awssdk.services.protocolrestjson.ProtocolRestJsonClient;
+import software.amazon.awssdk.services.protocolrestjson.model.OperationWithExplicitPayloadStructureResponse;
+import software.amazon.awssdk.services.protocolrestjson.model.SimpleStruct;
+import software.amazon.awssdk.services.protocolrestxml.ProtocolRestXmlClient;
+import software.amazon.awssdk.services.protocolrestxml.model.OperationWithExplicitPayloadStringResponse;
+
+public class MarshallersAddContentLengthTest {
+    public static final String STRING_PAYLOAD = "TEST_STRING_PAYLOAD";
+    @Rule
+    public WireMockRule wireMock = new WireMockRule(0);
```
L-Applin, line 48: Could we use JUnit 5 with `@WireMockTest` instead?

joviegas: done

L-Applin (approved):

Issue #4912, opened 2024-02-09
Cross Region S3 Access using AWS_GLOBAL Endpoint No Longer Works


jjathman:
### Describe the bug

I believe due to the change done for #4720, if a client is configured using the AWS_GLOBAL region and cross region access is enabled, a PUT to any region that is not US-EAST1 will now fail with an authorization error. This broke with BOM version 2.23.13.

### Expected Behavior

This client configuration:

```
        S3AsyncClient.builder()
                .region(AWS_GLOBAL)
                .credentialsProvider(credentialsProvider)
                .crossRegionAccessEnabled(true)
                .build();
```

should be able to write to a bucket in any region. This worked prior to BOM version 2.23.13.

### Current Behavior

The put will fail with an error like this:

```
The authorization header is malformed; the region 'us-east-1' is wrong; expecting 'us-east-2' (Service: S3, Status Code: 400, Request ID: ABC, Extended Request ID: ABC123)
software.amazon.awssdk.services.s3.model.S3Exception: The authorization header is malformed; the region 'us-east-1' is wrong; expecting 'us-east-2' (Service: S3, Status Code: 400, Request ID: ABC, Extended Request ID: ABC123)
	at software.amazon.awssdk.protocols.xml.internal.unmarshall.AwsXmlPredicatedResponseHandler.handleErrorResponse(AwsXmlPredicatedResponseHandler.java:156)
	at software.amazon.awssdk.protocols.xml.internal.unmarshall.AwsXmlPredicatedResponseHandler.handleResponse(AwsXmlPredicatedResponseHandler.java:108)
	at software.amazon.awssdk.protocols.xml.internal.unmarshall.AwsXmlPredicatedResponseHandler.handle(AwsXmlPredicatedResponseHandler.java:85)
	at software.amazon.awssdk.protocols.xml.internal.unmarshall.AwsXmlPredicatedResponseHandler.handle(AwsXmlPredicatedResponseHandler.java:43)
	at software.amazon.awssdk.core.internal.handler.BaseClientHandler.lambda$successTransformationResponseHandler$7(BaseClientHandler.java:279)
	at software.amazon.awssdk.core.internal.http.async.AsyncResponseHandler.lambda$prepare$0(AsyncResponseHandler.java:92)
```

### Reproduction Steps

Create a client as shown above using the current version of the SDK.

### Possible Solution

[...]

PR #4921, opened 2024-02-12
Fix for Issue #4912 where Cross Region S3 Access using AWS_GLOBAL Endpoint failed.

joviegas:
## Motivation and Context

-  Issue #4912
- Note that PR #4849 add fix to enable regional endpoints by setting useGlobalEndpoint , however this doesnot work for AWS_GLOBAL region.
- Also if we try to make direct calls with AWS_GLOBAL without disabling useGlobalEndpoint then we will not get the Redirected region for `Head` Apis like "HeadObject" and "HeadBucket".
- Thus we need to update the AWS_GLOBAL as "US_EAST_1" in order to get the redirected region

## Modifications

- Update Region as  "US_EAST_1" if the call is made from AWS_GLOBAL.
- This makes sure we do a regional endpoint call to get the redirected region

## Types of changes

## License

### DIFF
--- a/services/s3/src/it/java/software/amazon/awssdk/services/s3/crossregion/S3CrossRegionCrtIntegrationTest.java
+++ b/services/s3/src/it/java/software/amazon/awssdk/services/s3/crossregion/S3CrossRegionCrtIntegrationTest.java
@@ -46,7 +46,7 @@ public class S3CrossRegionCrtIntegrationTest extends S3CrossRegionAsyncIntegrati
     @BeforeEach
     public void initialize() {
         crossRegionS3Client = S3AsyncClient.crtBuilder()
-                                           .region(CROSS_REGION)
+                                           .region(Region.AWS_GLOBAL)
                                            .crossRegionAccessEnabled(true)
                                              .build();
     }
--- a/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/endpointprovider/BucketEndpointProvider.java
+++ b/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/endpointprovider/BucketEndpointProvider.java
@@ -43,9 +43,20 @@ public class BucketEndpointProvider implements S3EndpointProvider {
     @Override
     public CompletableFuture<Endpoint> resolveEndpoint(S3EndpointParams endpointParams) {
         Region crossRegion = regionSupplier.get();
-        return delegateEndPointProvider.resolveEndpoint(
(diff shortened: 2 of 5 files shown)


services/s3/src/it/java/software/amazon/awssdk/services/s3/crossregion/S3CrossRegionCrtIntegrationTest.java
```diff
@@ -46,7 +46,7 @@ static void clearClass() {
     @BeforeEach
     public void initialize() {
         crossRegionS3Client = S3AsyncClient.crtBuilder()
-                                           .region(CROSS_REGION)
+                                           .region(Region.AWS_GLOBAL)
```
millems, line 49: Intentional change?

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/endpointprovider/BucketEndpointProvider.java
```diff
@@ -43,9 +43,16 @@ public static BucketEndpointProvider create(S3EndpointProvider delegateEndPointP
...
-            endpointParams.copy(c -> c.region(crossRegion == null ? endpointParams.region() : crossRegion)
-                                      .useGlobalEndpoint(false)));
+        S3EndpointParams.Builder endpointParamsBuilder = endpointParams.toBuilder();
+        if (crossRegion != null) {
+            endpointParamsBuilder.region(crossRegion);
+        } else {
+            if (Region.AWS_GLOBAL.equals(endpointParams.region())) {
+                endpointParamsBuilder.region(Region.US_EAST_1);
+            }
+            endpointParamsBuilder.useGlobalEndpoint(false);
```
millems, line 53: Can we add a comments on why these parts are needed? I'm not sure why we were setting use-global-endpoint to false in the first place, and now I really don't get what all this is for without reading the history.

services/s3/src/it/java/software/amazon/awssdk/services/s3/crossregion/S3CrossRegionCrtIntegrationTest.java
joviegas: yeap , making sure we test this in Integ too since the cross region feature depends mainly on the error headers send by S3 based on the client region configured.
We have different client configured in different regions in https://github.com/aws/aws-sdk-java-v2/tree/b415e16e755d7c95c7bfae4725de4d623fb43c5b/services/s3/src/it/java/software/amazon/awssdk/services/s3/crossregion

millems (approved):

PR #4939, opened 2024-02-16
Add config class for SplittingTransformer

davidh44:
## Modifications

Add `SplitTransformerConfiguration` for when calling `AsyncResponseTransformer.split()`

## Types of changes

## Checklist

## License

### DIFF
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/SplitTransformerConfiguration.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/SplitTransformerConfiguration.java
@@ -1,0 +1,112 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.core;
+
+import java.util.Objects;
+import software.amazon.awssdk.annotations.SdkPublicApi;
+import software.amazon.awssdk.core.async.AsyncResponseTransformer;
+import software.amazon.awssdk.core.internal.async.SplittingTransformer;
+import software.amazon.awssdk.utils.Validate;
+import software.amazon.awssdk.utils.builder.CopyableBuilder;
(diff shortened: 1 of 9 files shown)


core/sdk-core/src/main/java/software/amazon/awssdk/core/SplitTransformerConfiguration.java
```diff
@@ -0,0 +1,112 @@
...
+    SplitTransformerConfiguration> {
+
+    private final Long bufferSize;
+
+    private SplitTransformerConfiguration(DefaultBuilder builder) {
+        this.bufferSize = Validate.paramNotNull(builder.bufferSize, "bufferSize");
+    }
+
+    /**
+     * Create a {@link FileRequestBodyConfiguration.Builder}, used to create a {@link SplitTransformerConfiguration}.
```
dagnir, line 43: Why does this mention `FileRequestBodyConfiguration`?

core/sdk-core/src/main/java/software/amazon/awssdk/core/SplitTransformerConfiguration.java
```diff
@@ -0,0 +1,112 @@
...
+import software.amazon.awssdk.core.async.AsyncResponseTransformer;
+import software.amazon.awssdk.core.internal.async.SplittingTransformer;
+import software.amazon.awssdk.utils.Validate;
+import software.amazon.awssdk.utils.builder.CopyableBuilder;
+import software.amazon.awssdk.utils.builder.ToCopyableBuilder;
+
+/**
+ * Configuration options for {@link AsyncResponseTransformer#split(SplitTransformerConfiguration)} to configure how the SDK
+ * should split the {@link AsyncResponseTransformer}.
+ *
```
dagnir, line 29: This description makes me thing the name is not quite right. "to configure __how__ the SDK should split..." sounds like it's configuration for the splitting transformer (the one we're calling `split` on), but the name makes it sound like it's for the `SplitAsyncResponseTransformer`, which is the result of the `split`.

@zoewangg WDYT?

core/sdk-core/src/main/java/software/amazon/awssdk/core/SplitTransformerConfiguration.java
davidh44: oops missed deleting that

core/sdk-core/src/main/java/software/amazon/awssdk/core/SplitTransformerConfiguration.java
davidh44: right, its used to instantiate a `SplittingTransformer` which is then passed to the `SplitAsyncResponseTransformer`

`SplittingTransformerConfiguration`?

dagnir: Okay thanks for clarifying. In that case then yeah I vote for `SplittingTransformerConfiguration`

dagnir (approved):

PR #4952, opened 2024-02-21
Choose next potential scheme if signer not found

dagnir:
## Motivation and Context
This commit fixes the generated auth scheme selection mechanism by ensuring that auth scheme is able to provide a signer implementation (i.e. does not throw an exception) before selecting the auth scheme. This is useful in situations where the signer implementation relies on components thay may not be available at runtime, such as the AwsV4aAuthScheme, which relies on CRT which is an optional dependency.

Additional changes:
 - introduce crt-unavailable-tests which is a good candidate for testing behavior of components that use CRT when CRT is not present.
 - modify DefaultAwsV4aAuthScheme so that errors thrown during static initialization of the singleton holder for the Sigv4a signer are caught and thrown directly instead of resulting in ExceptionInInitializerError
## Testing
 - New unit tests

## Types of changes

## Checklist

## License

### DIFF
--- a/codegen/src/main/java/software/amazon/awssdk/codegen/poet/auth/scheme/AuthSchemeInterceptorSpec.java
+++ b/codegen/src/main/java/software/amazon/awssdk/codegen/poet/auth/scheme/AuthSchemeInterceptorSpec.java
@@ -52,6 +52,7 @@ import software.amazon.awssdk.core.metrics.CoreMetric;
 import software.amazon.awssdk.endpoints.EndpointProvider;
 import software.amazon.awssdk.http.auth.spi.scheme.AuthScheme;
 import software.amazon.awssdk.http.auth.spi.scheme.AuthSchemeOption;
+import software.amazon.awssdk.http.auth.spi.signer.HttpSigner;
 import software.amazon.awssdk.identity.spi.AwsCredentialsIdentity;
 import software.amazon.awssdk.identity.spi.Identity;
 import software.amazon.awssdk.identity.spi.IdentityProvider;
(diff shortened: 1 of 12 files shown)


codegen/src/main/java/software/amazon/awssdk/codegen/poet/auth/scheme/AuthSchemeInterceptorSpec.java
```diff
@@ -274,6 +275,22 @@ private MethodSpec generateTrySelectAuthScheme() {
...
         }
 
+        builder.addStatement("$T signer",
+                             ParameterizedTypeName.get(ClassName.get(HttpSigner.class), TypeVariableName.get("T")));
+        builder.beginControlFlow("try");
+        {
+            builder.addStatement("signer = authScheme.signer()");
+            builder.endControlFlow();
+        }
+        builder.beginControlFlow("catch (RuntimeException e)");
```
sugmanue, line 285:
```suggestion
        }
        builder.nextControlFlow("catch (RuntimeException e)");
```

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/auth/scheme/query-auth-scheme-interceptor.java
```diff
@@ -100,6 +101,14 @@ private <T extends Identity> SelectedAuthScheme<T> trySelectAuthScheme(AuthSchem
                 .add(() -> String.format("'%s' does not have an identity provider configured.", authOption.schemeId()));
             return null;
         }
+        HttpSigner<T> signer;
+        try {
+            signer = authScheme.signer();
+        } catch (RuntimeException e) {
```
joviegas, line 107: Why are we not catching specific exception like ClassNotFoundException ?

dagnir: - I wanted it to be more generic to catch other issues in the future. I don't think we should only be looking for classnotfound
- The V4a authscheme doesn't throw classnotfound in this case

joviegas: - What exception is thrown in this case?
- What is the actual low-level exception sent in cases when CRT is not in the classpath? 
- Is it possible that the low-level exception is suppressed in the expression and can be retrieved using getCause?
- Why do you think we should catch generic runtime exceptions?"

dagnir: - It throws a `RuntimeException`: https://github.com/aws/aws-sdk-java-v2/blob/b0fd0e65928363573bea404bf1498713c032d538/core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/signer/util/OptionalDependencyLoaderUtil.java#L50

In any case, we're not solving for *just* CRT. This fix is intended to be more broad so that we catch any issues retrieving the signer from the auth scheme. By catching `RuntimeException` we ensure that we can handle a broader set of issues that can happen when trying to retrieve the signer. In all case, we should be moving on to the next available auth scheme.

joviegas (approved):

PR #4955, opened 2024-02-21
Fix request cancellation issue in the AWS CRT-based S3 client that co…

zoewangg:
…uld lead to memory leak

## Motivation and Context
Currently, if a request is cancelled in AWS CRT-based S3 client, CRT resource will not be cancelled and cleaned up properly, leaking to memory leak. 

The root cause is that we are calling `S3MetaRequest.close()` before `S3MetaRequest.cancel()`. 

## Modifications
- Updated the code to invoke cancel before close.
- Created `S3MetaRequestWrapper` class to make sure we don't invoke S3MetaRequest if it's closed. All interactions with `S3MetaRequest` must be through `S3MetaRequestWrapper` going forward. 
- Added checkstyle rule to make sure `S3MetaRequestWrapper` is used instead of `S3MetaRequest`.

## Testing
Tested it locally and added new functional tests.

## Types of changes

## Checklist

## License

### DIFF
--- a/services-custom/s3-transfer-manager/src/test/java/software/amazon/awssdk/transfer/s3/internal/S3CrtTransferProgressListenerTest.java
+++ b/services-custom/s3-transfer-manager/src/test/java/software/amazon/awssdk/transfer/s3/internal/S3CrtTransferProgressListenerTest.java
@@ -160,7 +160,9 @@ public class S3CrtTransferProgressListenerTest {
         assertThat(transferListener.getExceptionCaught()).isInstanceOf(CancellationException.class);
         assertThat(transferListener.isTransferComplete()).isFalse();
         assertThat(transferListener.isTransferInitiated()).isTrue();
-        assertMockOnFailure(transferListenerMock);
+        Mockito.verify(transferListenerMock, times(1)).transferFailed(ArgumentMatchers.any());
+        Mockito.verify(transferListenerMock, times(1)).transferInitiated(ArgumentMatchers.any());
+        Mockito.verify(transferListenerMock, times(0)).transferComplete(ArgumentMatchers.any());
 
     }
 
--- a/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/S3CrtAsyncHttpClient.java
+++ b/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/S3CrtAsyncHttpClient.java
@@ -45,7 +45,6 @@ import software.amazon.awssdk.crt.s3.ChecksumConfig;
 import software.amazon.awssdk.crt.s3.ResumeToken;
 import software.amazon.awssdk.crt.s3.S3Client;
 import software.amazon.awssdk.crt.s3.S3ClientOptions;
-import software.amazon.awssdk.crt.s3.S3MetaRequest;
 import software.amazon.awssdk.crt.s3.S3MetaRequestOptions;
 import software.amazon.awssdk.http.Header;
 import software.amazon.awssdk.http.SdkHttpExecutionAttributes;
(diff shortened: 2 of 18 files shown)


services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/S3MetaRequestPauseObservable.java
```diff
@@ -18,25 +18,24 @@
 import java.util.function.Function;
 import software.amazon.awssdk.annotations.SdkInternalApi;
 import software.amazon.awssdk.crt.s3.ResumeToken;
-import software.amazon.awssdk.crt.s3.S3MetaRequest;
 
 /**
  * An observable that notifies the observer {@link S3CrtAsyncHttpClient} to pause the request.
  */
 @SdkInternalApi
```
zoewangg, line 25: This class should've been a protected API. I will create a separate PR to fix it.

services/s3/src/test/java/software/amazon/awssdk/services/s3/functionaltests/S3ExpressCreateSessionTest.java
```diff
@@ -329,9 +331,8 @@ private static final class CapturingInterceptor implements ExecutionInterceptor
         public void beforeTransmission(Context.BeforeTransmission context, ExecutionAttributes executionAttributes) {
             SdkHttpRequest sdkHttpRequest = context.httpRequest();
             this.headers = sdkHttpRequest.headers();
-            System.out.printf("%s %s%n", sdkHttpRequest.method(), sdkHttpRequest.encodedPath());
-            headers.forEach((k, strings) -> System.out.printf("%s, %s%n", k, strings));
-            System.out.println();
+            log.debug(() -> String.format("%s %s%n", sdkHttpRequest.method(), sdkHttpRequest.encodedPath()));
```
zoewangg, line 334: I changed this because I noticed it produced a lot of (unnecessary) logs.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/S3CrtAsyncHttpClient.java
```diff
@@ -215,23 +219,6 @@ private static S3MetaRequestOptions.MetaRequestType requestType(AsyncExecuteRequ
         return S3MetaRequestOptions.MetaRequestType.DEFAULT;
     }
 
-    private static void closeResourceCallback(CompletableFuture<Void> executeFuture,
```
zoewangg, line 218: I moved this to S3CrtResponseHandlerAdapter because I figured it makes more sense to manage the closure of the S3MetaRequest in one place.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/S3CrtResponseHandlerAdapter.java
```diff
@@ -86,8 +108,7 @@ public int onResponseBody(ByteBuffer bodyBytesIn, long objectRangeStart, long ob
                 failResponseHandlerAndFuture(failure);
                 return;
             }
-
-            metaRequest.incrementReadWindow(bytesReceived);
+            s3MetaRequest().incrementReadWindow(bytesReceived);
```
TingDaoK, line 111: do we need to check for null from the result of `s3MetaRequest()`?

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/S3CrtResponseHandlerAdapter.java
```diff
@@ -86,8 +108,7 @@ public int onResponseBody(ByteBuffer bodyBytesIn, long objectRangeStart, long ob
                 failResponseHandlerAndFuture(failure);
```
TingDaoK, line 108: If the write future failed, we only close on the meta request, which won't stop the meta request keep receiving data, can we also cancel on the meta request before it closes? 

The only result of cancel on the meta request will be the complete callback may invoked with `cancelled` error, but as the result has already been set, it probably doesn't affect anything.

zoewangg: Yeah, makes sense, I'll just call cancel whenever there's an exception

TingDaoK (approved):

services-custom/s3-transfer-manager/src/test/java/software/amazon/awssdk/transfer/s3/internal/S3CrtTransferProgressListenerTest.java
```diff
@@ -160,7 +160,9 @@ void listeners_reports_ErrorsWhenCancelled(WireMockRuntimeInfo wm) throws Interr
         assertThat(transferListener.getExceptionCaught()).isInstanceOf(CancellationException.class);
         assertThat(transferListener.isTransferComplete()).isFalse();
         assertThat(transferListener.isTransferInitiated()).isTrue();
-        assertMockOnFailure(transferListenerMock);
+        Mockito.verify(transferListenerMock, times(1)).transferFailed(ArgumentMatchers.any());
```
zoewangg, line 163: The failure of the test actually verified this fix. Before the fix, the request would never cancelled, so we'd still get data after cancelling future. After the fix, the request would get cancelled properly, so ` Mockito.verify(transferListenerMock, times(1)).bytesTransferred(ArgumentMatchers.any());` may not be invoked at all.

davidh44 (approved):

PR #4956, opened 2024-02-21
Remove the default use of the legacy signer

sugmanue:
## Motivation and Context

Removes the use of the `Aws4Signer` class for the default polly presigner. Instead we fabricate a `SelectedAuthScheme` and delegate to the signer following the new SRA I&A model.

## Testing

There are already existing tests, added another one to validate that the new and old singing strategies works the same by testing using a fixed clock and signature and by using a request level override.

## Types of changes

## Checklist

## License


services/polly/src/main/java/software/amazon/awssdk/services/polly/internal/presigner/DefaultPollyPresigner.java
```diff
@@ -233,10 +316,13 @@ private AwsCredentialsIdentity resolveCredentials(IdentityProvider<? extends Aws
...
-                .orElse(DEFAULT_SIGNER);
-
+                               .orElse(null);
+        if (signer == null) {
+            return null;
+        }
         return Validate.isInstanceOf(Presigner.class, signer,
-                "Signer of type %s given in request override is not a Presigner", signer.getClass().getSimpleName());
+                                     "Signer of type %s given in request override is not a Presigner",
+                                     signer.getClass().getSimpleName());
```
zoewangg, line 325: I understand this is existing code, but getSimpleName is quite expensive, can we change to just use getName() while we are at it?

services/polly/src/main/java/software/amazon/awssdk/services/polly/internal/presigner/DefaultPollyPresigner.java
```diff
@@ -124,22 +136,28 @@ public void close() {
         IoUtils.closeIfCloseable(credentialsProvider, null);
     }
 
+    // For testing
+    static PollyPresigner.Builder builder(Clock signingClock) {
```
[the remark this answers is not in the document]
sugmanue: > Minor: instead of exposing setter method for testing, can we use dependency injection for testing, something like below

Sure, I can change it, but just to clarify, the setter is not ever exposed as is not part of the type interface returned by this method.

services/polly/src/main/java/software/amazon/awssdk/services/polly/internal/presigner/DefaultPollyPresigner.java
sugmanue: I didn't know that, now I do, I will update to use `getName()` instead.

zoewangg (approved):

PR #4962, opened 2024-02-23
Support Non proxy host settings in the ProxyConfiguration for Crt http client.

joviegas:
Support Non proxy host settings in the ProxyConfiguration for Crt http client.

## Motivation and Context

- Need An API in ProxyConfiguration of CRT Http client to specify  
- Also since we are making the Default configuration of Proxy to use Env variables and proxy settings we need to make sure the Non Proxy settings are picked up by CRT which are specified in env variables and ssystem settings 

## Modifications

- Using similar logic as used in Apache HTTP client Non Proxy host
https://github.com/aws/aws-sdk-java-v2/blob/61e0fb1ef16506feccda1900dc00b70f88c9a139/http-clients/apache-client/src/main/java/software/amazon/awssdk/http/apache/internal/SdkProxyRoutePlanner.java#L46-L58 , same is used in V1.

## Testing

- Added Junits
## Types of changes

## License


core/crt-core/src/main/java/software/amazon/awssdk/crtcore/CrtConfigurationUtils.java
```diff
@@ -54,11 +60,19 @@ public static Optional<HttpProxyOptions> resolveProxy(CrtProxyConfiguration prox
         return Optional.of(clientProxyOptions);
     }
 
+    private static boolean doesTargetMatchNonProxyHosts(String target, Set<String> hostPatterns) {
```
L-Applin, line 63: should we extract this same logic that is used by multiple HTTP client into a common utils class?

core/crt-core/src/main/java/software/amazon/awssdk/crtcore/CrtProxyConfiguration.java
```diff
@@ -36,6 +39,7 @@ public abstract class CrtProxyConfiguration {
     private final String password;
     private final Boolean useSystemPropertyValues;
     private final Boolean useEnvironmentVariableValues;
+    private final   Set<String> nonProxyHosts;
```
L-Applin, line 42: nit: extra space

joviegas: Done

http-clients/aws-crt-client/src/test/java/software/amazon/awssdk/http/crt/ProxyWireMockTest.java
```diff
@@ -121,4 +128,47 @@ public void proxyConfigured_httpGet() throws Throwable {
...
+            randomPort = secureRandom.nextInt(65535);
+        } while (randomPort == mockProxy.port());
+
+        proxyCfg = ProxyConfiguration.builder()
+                                     .host("localhost")
+                                     .port(randomPort)
+                                     .nonProxyHosts(Stream.of("localhost").collect(Collectors.toSet()))
+                                     .build();
+
+        client = AwsCrtAsyncHttpClient.builder()
```
zoewangg, line 147: We should close the client.

PR #4971, opened 2024-02-27
Surface area review

L-Applin:
## Surface Area Review Changes
- Remove `SplitAsyncResponseTransformer`, replacing it with `AsyncRersponseTransformer.SplitResult` inner interface with default internal implementation, with builder.
- renamed `returnFuture` to `resultFuture`.
- check and guard against for negative buffer size, renamed to `bufferSizeInBytes`.
- add logic to not perform multipart if part number or range is specified.
- return response from first call to get-object.

## Other changes
- Small refactoring: moved some logic into `DownloadObjectHelper` to be consistent with upload/copy.


core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/SplittingTransformer.java
```diff
@@ -218,7 +222,6 @@ public CompletableFuture<ResponseT> prepare() {
         public void onResponse(ResponseT response) {
             if (onResponseCalled.compareAndSet(false, true)) {
                 log.trace(() -> "calling onResponse on the upstream transformer");
-                // todo: should we send back the first response to the upstreamResponseTransformer as-is?
```
anirudh9391, line 221: Did we decide to not do this ?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/SplittingTransformer.java
```diff
@@ -308,7 +311,7 @@ public static <ResponseT, ResultT> Builder<ResponseT, ResultT> builder() {
 
     public static final class Builder<ResponseT, ResultT> {
 
-        private long maximumBufferSize;
+        private Long maximumBufferSize;
```
anirudh9391, line 314: If it is null, does it defer to default (32MB If i remember correctly ) ?

services/s3/src/test/java/software/amazon/awssdk/services/s3/internal/multipart/MultipartDownloadIntegrationTest.java
```diff
@@ -51,7 +49,7 @@ class MultipartDownloadIntegrationTest {
 
     private S3AsyncClient s3;
     private final SplittingTransformerConfiguration splitConfig = SplittingTransformerConfiguration.builder()
-                                                                                                   .bufferSize(1024 * 1024 * 32L)
+                                                                                                   .bufferSizeInBytes(1024 * 1024 * 32L)
```
anirudh9391, line 52: nit : Would improve readability if we had a variable denoting that this is the default value : 
```
private final int DEFAULT_BUFFERED_SIZE_IN_BYTES = 1024 * 1024 * 32L; 
```
With a comment that the default is 32MB.

services/s3/src/test/java/software/amazon/awssdk/services/s3/internal/multipart/MultipartDownloadIntegrationTest.java
```diff
@@ -63,24 +61,24 @@ void init() {
...
-        ResponseBytes<GetObjectResponse> res = split.preparedFuture().join();
+        ResponseBytes<GetObjectResponse> res = split.resultFuture().join();
         log.info(() -> "complete");
         byte[] bytes = res.asByteArray();
         log.info(() -> String.format("Byte len: %s", bytes.length));
         assertThat(bytes.length).isEqualTo(fileTestSize * 1024 * 1024);
     }
 
-    @Test
+    // @Test
```
anirudh9391, line 81: same as above

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/SplittingTransformer.java
L-Applin: That question was part of the surface area review. In v2, the existing CRT client does return the first response, and in v1 the multipart is done all in transfer-manager so this behaviour does not apply, the GetObjectResponse is never returned to the user. Therefore, returning the first response seems appropriate.

services/s3/src/test/java/software/amazon/awssdk/services/s3/internal/multipart/MultipartDownloadIntegrationTest.java
```diff
@@ -63,24 +61,24 @@ void init() {
                                .build();
     }
 
-    @Test
+    // @Test
```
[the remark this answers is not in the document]
L-Applin: Intentional yes, but please ignore this class, it is only used manually by me to test real world scenario with s3 during development. But I'll add a todo to remove it before release.

services/s3/src/test/java/software/amazon/awssdk/services/s3/internal/multipart/MultipartDownloadIntegrationTest.java
L-Applin: like mentioned in the other comment, please ignore this class

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/SplittingTransformer.java
L-Applin: It's the `MultipartS3AsyncClient` will default to 32MB, but here if it is null, the constructor throws a NPE

anirudh9391 (approved):
Approving

PR #4984, opened 2024-02-29
Adds accountId endpoint mode

cenedhryn:
## Motivation and Context
Account Id endpoint routing has been unblocked by DDB being enabled for SRA identity and auth, thereby resolving credentials before endpoints. Some code changes are required to support this change, because the original code for account Id was written before SRA. 

In addition, it adds AccountId Endpoint Mode as a built-in endpoint parameter. AccountIdMode should be configurable in profile file, system settings/env vars and at the client level

## Modifications
- Adds AccountIdMode as a built-in endpoint parameter
- Adds codegen infrastructure for generating setter method on the client for accountIdMode
  - Introduces a simple class to represent a parameter, `LocalParameter`
  - Introduces a holder class for codegen for endpoint parameters that should be exposed on clients. Calling it `EndpointParamsKnowledgeIndex`, but for now accountIdMode is the only example.
- Removes non-generic endpoint parameters from `AwsEndpointProviderUtils`, because this class is copied to all services. Instead, [EndpointResolverInterceptorSpec.java](https://github.com/aws/aws-sdk-java-v2/pull/4984/files#diff-6c8d7ece5e2b95f56bd0ace36d7de9509e4991bb148508cd36c213abf58e974d) generates logic for those parameters inside the provider itself (accountId, accountIdEndpointMode, s3GlobalEndpoint). 
- AccountIdMode is resolved in the BaseClientBuilder of each service by checking if the value is set on the client, otherwise it will call the resolver to see if any values are set through system properties or profile file


codegen/src/main/java/software/amazon/awssdk/codegen/poet/rules/EndpointParamsKnowledgeIndex.java
```diff
@@ -0,0 +1,190 @@
...
+    private final IntermediateModel intermediateModel;
+    private Map<BuiltInParameter, LocalParameter> parametersToGenerate = new EnumMap<>(BuiltInParameter.class);
+
+    static {
+        BUILT_IN_PARAMS_FOR_CLIENT_BUILDER.put(
+            BuiltInParameter.AWS_AUTH_ACCOUNT_ID_ENDPOINT_MODE,
+            new LocalParameter("accountIdEndpointMode",
+                               AccountIdEndpointMode.class,
+                                     "Sets the behavior when account ID based endpoints are created. "
+                                     + "See {@link AccountIdEndpointMode} for values"));
```
dagnir, line 62: minor: when referencing classes in a javadoc, we should still use `$T` placeholder in case the class is not already imported by something else

core/aws-core/src/main/java/software/amazon/awssdk/awscore/AwsExecutionAttribute.java
```diff
@@ -64,6 +65,12 @@ public final class  AwsExecutionAttribute extends SdkExecutionAttribute {
     public static final ExecutionAttribute<String> AWS_AUTH_ACCOUNT_ID =
         new ExecutionAttribute<>("AwsAuthAccountId");
 
+    /**
+     * The AWS account ID associated with the identity resolved for this request.
+     */
```
dagnir, line 70: Is this accurate? The attribute name is "mode" but this says account ID

codegen/src/main/java/software/amazon/awssdk/codegen/poet/rules/EndpointParamsKnowledgeIndex.java
```diff
@@ -0,0 +1,190 @@
...
+        MethodSpec.Builder builder = MethodSpec.methodBuilder("accountIdFromIdentity")
+                                               .addModifiers(PRIVATE, STATIC)
+                                               .addTypeVariable(TypeVariableName.get("T", Identity.class))
+                                               .addParameter(paramType, "selectedAuthScheme")
+                                               .returns(String.class);
+
+        builder.addStatement("$T identity = $T.joinLikeSync(selectedAuthScheme.identity())", TypeVariableName.get("T"),
+                             CompletableFutureUtils.class);
+        builder.addStatement("$T accountId = null", String.class);
+        builder.beginControlFlow("if (identity instanceof $T)", AwsCredentialsIdentity.class);
```
dagnir, line 184: Just curious, will `AwsCredentialsIdentity` be the only identity type that supports account ID?

core/aws-core/src/main/java/software/amazon/awssdk/awscore/endpoints/AccountIdEndpointMode.java
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * Enum Class for AccountId Endpoint Mode.
+ */
+@SdkPublicApi
+public enum AccountIdEndpointMode {
+
+    PREFERRED,
+    DISABLED,
+    REQUIRED;
```
dagnir, line 28: Can we get javadocs for each value?

codegen/src/main/java/software/amazon/awssdk/codegen/model/internal/LocalParameter.java
```diff
@@ -0,0 +1,44 @@
...
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.codegen.model.internal;
+
+/**
+ * Represents a generic parameter that can be code generated, but isn't tied to a specific model type
```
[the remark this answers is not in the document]
cenedhryn: A model shape that can be a parameter, such as a context parameter. Our existing codegen types are tightly coupled to the model, which didn't work for account id routing mode. Any suggestions for updating the description?

codegen/src/main/java/software/amazon/awssdk/codegen/poet/rules/EndpointParamsKnowledgeIndex.java
cenedhryn: Thanks for the reminder, will fix!

codegen/src/main/java/software/amazon/awssdk/codegen/poet/rules/EndpointParamsKnowledgeIndex.java
cenedhryn: We'd have to look in the crystal ball for that one. Nothing in the pipelines as for now, but it would be good to have another round of discussions here. It's a dilemma to build at the right level - neither adding support that's super generic and never diversified nor being too specific and then getting a broader use case is great.

core/aws-core/src/main/java/software/amazon/awssdk/awscore/endpoints/AccountIdEndpointMode.java
cenedhryn: Yes

codegen/src/main/java/software/amazon/awssdk/codegen/model/internal/LocalParameter.java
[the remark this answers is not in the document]
dagnir: I think "model shape" as you describe sounds better

codegen/src/main/java/software/amazon/awssdk/codegen/poet/rules/EndpointParamsKnowledgeIndex.java
cenedhryn: If we have to add accountId in the future, we can support it through an interface.

dagnir (approved):

PR #4994, opened 2024-03-06
S3 Event Notifications parsers

L-Applin:
Adds a new module that contains the classes used to represent Amazon S3 Event Notifications, and de/serializer from/to json format.

## Motivation and Context
The AWS SDK for Java v1 provides this fucntionality but it had not been ported to v2 yet

## Modifications
New module with POJO classes and de/serialiser classses

## Testing
new unit tests

## Types of changes

## Checklist

## License


services-custom/s3-event-notifications/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+```java
+S3EventNotification event = new S3EventNotification(...);
+String json = event.toJson();
+String jsonPretty = event.toJsonPretty();
+```
+
+`GlacierEventData`, `ReplicationEventData`, `IntelligentTieringEventData` and `LifecycleEventData`
+will be excluded from the json if null. Any other null fields of the object will be 
+serialized in the json as `null`.
```
millems, line 36: Why the difference?

services-custom/s3-event-notifications/pom.xml
```diff
@@ -0,0 +1,114 @@
...
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>utils</artifactId>
+            <version>${awsjavasdk.version}</version>
+        </dependency>
+
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter</artifactId>
+            <version>${junit5.version}</version>
```
millems, line 76: Do we need to specify the versions, or can we rely on the bom-internal to manage the version?

services-custom/s3-event-notifications/src/main/java/software/amazon/awssdk/eventnotifications/s3/internal/DefaultS3EventNotificationReader.java
```diff
@@ -0,0 +1,294 @@
...
+    }
+
+
+    private String expectStringOrNull(JsonNode node, String name) {
+        if (node == null) {
+            return null;
+        }
+        Validate.isTrue(node.isString(), "'%s' was not a string", name);
+        return node.asString();
+    }
```
millems, line 265: Everywhere we're calling this, we're doing a `get` on `node` first with the same string as `name`. Should we do the `get` here, to reduce the chance of copy+paste bugs making these strings different?

services-custom/s3-event-notifications/src/main/java/software/amazon/awssdk/eventnotifications/s3/model/GlacierEventData.java
```diff
@@ -0,0 +1,60 @@
...
+ */
+
+package software.amazon.awssdk.eventnotifications.s3.model;
+
+import java.util.Objects;
+import software.amazon.awssdk.annotations.SdkPublicApi;
+import software.amazon.awssdk.utils.ToString;
+
+@SdkPublicApi
+public class GlacierEventData {
```
millems, line 23: Can we add javadoc here? Maybe link to the S3 docs for this event type?

services-custom/s3-event-notifications/src/main/java/software/amazon/awssdk/eventnotifications/s3/model/S3EventNotificationRecord.java
```diff
@@ -0,0 +1,233 @@
...
+             eventVersion,
+             requestParameters,
+             responseElements,
+             s3,
+             userIdentity,
+             null,
+             null,
+             null,
+             null);
+    }
```
millems, line 92: These constructors are wild. If we wanted to use beans, we should use zero-arg constructors with getters/setters. If we wanted to use v2-style APIs, then we should use builders.

services-custom/s3-event-notifications/README.md
L-Applin: These fields are only present in the event json under specific circumstances. From the **[Event Structure](https://docs.aws.amazon.com/AmazonS3/latest/userguide/notification-content-structure.html)** page:

> The glacierEventData key is only visible for s3:ObjectRestore:Completed events.
> The replicationEventData key is only visible for replication events.
> The intelligentTieringEventData key is only visible for S3 Intelligent-Tiering events.
> The lifecycleEventData key is only visible for S3 Lifecycle transition events.

services-custom/s3-event-notifications/src/main/java/software/amazon/awssdk/eventnotifications/s3/model/S3EventNotificationRecord.java
L-Applin: Yes they are pretty insane... they are the v1 constructors as-is. I kept them to be as consistent as possible with v1.

services-custom/s3-event-notifications/README.md
millems: Do we need the other null fields to be explicitly marshalled as null, or could we exclude them as well?

services-custom/s3-event-notifications/src/main/java/software/amazon/awssdk/eventnotifications/s3/model/S3EventNotificationRecord.java
millems: Imo, we should remove them and let customers call the setters. We don't use constructors anywhere else...

services-custom/s3-event-notifications/README.md
L-Applin: The other fields are marshalled as null in v1, so I think we should keep the same behaviour

services-custom/s3-event-notifications/src/main/java/software/amazon/awssdk/eventnotifications/s3/model/S3EventNotificationRecord.java
```diff
@@ -18,26 +18,57 @@
...
+    private LifecycleEventData lifecycleEventData;
+    private IntelligentTieringEventData intelligentTieringEventData;
+    private ReplicationEventData replicationEventData;
+    private Instant eventTime;
+
+    public S3EventNotificationRecord() {
+    }
+
+    @SdkTestInternalApi
+    protected S3EventNotificationRecord(
```
millems, line 45: Should this be package-protected (no modifier) so that people can't get access to the method through extension?

PR #4908, opened 2024-02-09
Pause/Resume Upload for Transfer Manager with Java S3Client

davidh44:
## Motivation and Context

Currently pausing/resuming a file upload with TransferManager is only supported with CRT S3 Client.

## Modifications

Adding support when using Java S3 Client.
building off PR https://github.com/aws/aws-sdk-java-v2/pull/4886

## Testing

added integ tests
to add unit tests

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

## Checklist

- [ ] I have read the [CONTRIBUTING](https://github.com/aws/aws-sdk-java-v2/blob/master/CONTRIBUTING.md) document
- [ ] Local run of `mvn install` succeeds
- [ ] My code follows the code style of this project
- [ ] My change requires a change to the Javadoc documentation
- [ ] I have updated the Javadoc documentation accordingly
- [ ] I have added tests to cover my changes
- [ ] All new and existing tests passed
- [ ] I have added a changelog entry. Adding a new entry must be accomplished by running the `scripts/new-change` script and following the instructions. Commit the new file created by the script in `.changes/next-release` with your changes.
- [ ] My change is to implement 1.11 parity feature and I have updated [LaunchChangelog](https://github.com/aws/aws-sdk-java-v2/blob/master/docs/LaunchChangelog.md)

## License

- [x] I confirm that this pull request can be released under the Apache 2 license

PR #4998, opened 2024-03-07
Fixed the issue in S3 multipart client where the list of parts could …

zoewangg:
…

## Motivation and Context
Fixed the issue in S3 multipart client where the list of parts could be out of order in `CompleteMultipartRequest`, causing `The list of parts was not in ascending order` error to be thrown.

This was a regression introduced in #4908. The root cause is that the type of `completedParts` was changed from `AtomicReferenceArray` to `ConcurrentHashMap`, which is not sorted when the SDK sends the request.
## Modifications

- Change back to `AtomicReferenceArray` since it should provide better performance than ConcurrentHashMap
- Ensure completeMultipart is only invoked once.

## Testing
Added unit tests

## Types of changes

## Checklist

## License

### DIFF
--- a/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/multipart/KnownContentLengthAsyncRequestBodySubscriber.java
+++ b/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/multipart/KnownContentLengthAsyncRequestBodySubscriber.java
@@ -18,13 +18,15 @@ package software.amazon.awssdk.services.s3.internal.multipart;
 import static software.amazon.awssdk.services.s3.multipart.S3MultipartExecutionAttribute.JAVA_PROGRESS_LISTENER;
 
 import java.util.Collection;
+import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReferenceArray;
 import java.util.function.Consumer;
+import java.util.stream.IntStream;
 import org.reactivestreams.Subscriber;
 import org.reactivestreams.Subscription;
 import software.amazon.awssdk.annotations.SdkInternalApi;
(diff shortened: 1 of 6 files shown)


services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/multipart/MpuRequestContext.java
```diff
@@ -94,7 +94,7 @@ public String uploadId() {
     }
 
     public Map<Integer, CompletedPart> existingParts() {
-        return existingParts != null ? Collections.unmodifiableMap(existingParts) : null;
```
zoewangg, line 97: I removed this since this seems a bit overkill for an internal API.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/multipart/KnownContentLengthAsyncRequestBodySubscriber.java
```diff
@@ -193,15 +199,16 @@ public void onComplete() {
     }
 
     private void completeMultipartUploadIfFinished(int requestsInFlight) {
-        if (isDone && requestsInFlight == 0) {
+        if (isDone && requestsInFlight == 0 && completedMultipartInitiated.compareAndSet(false, true)) {
```
zoewangg, line 202: Added this check back. It was gone as part of refactoring for some reason.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/multipart/KnownContentLengthAsyncRequestBodySubscriber.java
```diff
@@ -193,15 +199,16 @@ public void onComplete() {
...
-            } else if (!completedParts.isEmpty()) {
+                parts =
+                    IntStream.range(0, completedParts.length())
+                             .mapToObj(completedParts::get)
+                             .toArray(CompletedPart[]::new);
+            } else {
                 // List of CompletedParts needs to be in ascending order
                 parts = mergeCompletedParts();
-            } else {
-                parts = existingParts.values().toArray(new CompletedPart[0]);
```
zoewangg, line 204: This else block is not needed since mergeCompletedParts should handle it

cenedhryn (approved):

PR #5003, opened 2024-03-07
Update s3-benchmarks to support java-based multipart download.

L-Applin:
The s3-benchmarks needed some update to complete perf testing.

**IMPORTANT**: this PR also removes the transfer manager workaround for multipart, since it is required to test transfer manager with s3 multipart client.

## Motivation and Context
While testing S3MultipartClient, noticed some problems/missing features.

## Modifications
- **remove TM workaround so that perf test run with it multipart client**
- Add config to specify java client to TM benchmarks
- Fix JavaBased download benchmarks for multiple iterations with file
- Fix JavaBased download benchmarks for large objects with in-memory

## testing
These changes have been used to run perf tests iterations already.


test/s3-benchmarks/src/main/java/software/amazon/awssdk/s3benchmarks/BaseTransferManagerBenchmark.java
```diff
@@ -171,6 +160,40 @@ private void warmUpDownloadBatch() {
...
+                    builder.maxConcurrency(config.maxConcurrency());
+                }
+                return builder.build();
+            }
+            case JAVA: {
+                logger.info(() -> "Using Java-based S3 Async client");
+                return S3AsyncClient.builder()
+                                    .multipartEnabled(true)
+                                    .multipartConfiguration(c -> c.minimumPartSizeInBytes(partSizeInMb * MB)
+                                                                  .thresholdInBytes(partSizeInMb * 2 * MB)
```
zoewangg, line 186: Is there any reason we are setting threshold here but not crt client?

L-Applin: No you are lets use the same for both (default, which is `minimumPartSizeInBytes`)

zoewangg (approved):

PR #5006, opened 2024-03-08
Link documentation to the Immutables library's guide for generating immutable DynamoDB Enhanced client classes

swar8080:
Hello, the Immutables java library's website has a [guide](https://immutables.github.io/dynamodb.html) (that I wrote) on generating DynamoDB Enhanced Client classes using the immutables library. It also has a recipe for doing partial updates and projections with immutable classes.

The README has a lombok example, so maybe others would find this useful!

### DIFF
--- a/services-custom/dynamodb-enhanced/README.md
+++ b/services-custom/dynamodb-enhanced/README.md
@@ -257,6 +257,8 @@ how Lombok's 'onMethod' feature is leveraged to copy the attribute based DynamoD
     }
 ```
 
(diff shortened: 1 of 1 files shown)


debora-ito:
@swar8080 I didn't know about this Immutables guide, good job on the DynamoDB integration page.

We usually don't reference third-party pages in our docs, because (1) the referenced page can get out-of-date or have incorrect content, and we won't have the control to change; (2) we can start getting requests to add other 3p libs.

We appreciate the work you've put into the guide, though! It helps in spreading the knowledge about DynamoDB Enhanced Client.

PR #5008, opened 2024-03-08
Adding source to identity (provider name)

cenedhryn:
## Motivation and Context
After identity has been resolved in the request pipeline, it's not known what identity provider created the identity. In order to capture that information in user agent based metrics, we need to carry that information in a context or in the identity itself. The latter is more reasonable to support with the current architecture. 

## Modifications
- Adds a ProviderSource interface that Identity implements. This interface contains a default implementation of `provider` that implementing classes can override. 
- Adds `provider` to default credential classes in the `auth` and `identity-spi` packages. 
- Modifies the `auth` credentials classes with builder and copyable builder where needed. Copy is necessary for StaticCredentialsProvider (later PR) 
- Excludes `provider` from equals hashcode (TBD)

### DIFF
--- a/core/auth/src/main/java/software/amazon/awssdk/auth/credentials/AwsBasicCredentials.java
+++ b/core/auth/src/main/java/software/amazon/awssdk/auth/credentials/AwsBasicCredentials.java
@@ -52,8 +52,8 @@ public final class AwsBasicCredentials implements AwsCredentials,
 
     private final String accessKeyId;
     private final String secretAccessKey;
-    private boolean validateCredentials;
-    private String provider;
+    private final boolean validateCredentials;
+    private final String provider;
 
     private AwsBasicCredentials(Builder builder) {
         this.accessKeyId = trimToNull(builder.accessKeyId);
@@ -76,6 +76,8 @@ public final class AwsBasicCredentials implements AwsCredentials,
     protected AwsBasicCredentials(String accessKeyId, String secretAccessKey) {
         this.accessKeyId = trimToNull(accessKeyId);
         this.secretAccessKey = trimToNull(secretAccessKey);
+        this.validateCredentials = false;
+        this.provider = null;
     }
 
     public static Builder builder() {
--- a/core/identity-spi/src/main/java/software/amazon/awssdk/identity/spi/Identity.java
+++ b/core/identity-spi/src/main/java/software/amazon/awssdk/identity/spi/Identity.java
@@ -19,7 +19,7 @@ import java.time.Instant;
 import java.util.Optional;
 import software.amazon.awssdk.annotations.SdkPublicApi;
 import software.amazon.awssdk.annotations.ThreadSafe;
-import software.amazon.awssdk.identity.spi.internal.ProviderSource;
+import software.amazon.awssdk.identity.spi.internal.ProviderNameAware;
 
 /**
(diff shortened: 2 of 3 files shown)


core/auth/src/main/java/software/amazon/awssdk/auth/credentials/AwsBasicCredentials.java
```diff
@@ -36,17 +40,32 @@
...
      * This should be accessed via {@link AnonymousCredentialsProvider#resolveCredentials()}.
      */
     @SdkInternalApi
-    static final AwsBasicCredentials ANONYMOUS_CREDENTIALS = new AwsBasicCredentials(null, null, false);
+    static final AwsBasicCredentials ANONYMOUS_CREDENTIALS = builder().validateCredentials(false).build();
 
     private final String accessKeyId;
     private final String secretAccessKey;
+    private boolean validateCredentials;
+    private String provider;
```
millems, line 56: Can these be final?

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/AwsBasicCredentials.java
```diff
@@ -121,4 +146,67 @@ public int hashCode() {
...
+        /**
+         * The AWS secret access key, used to authenticate the user interacting with services.
+         */
+        public Builder secretAccessKey(String secretAccessKey) {
+            this.secretAccessKey = secretAccessKey;
+            return this;
+        }
+
+        /**
+         * The name of the identity provider that created this credential identity.
```
millems, line 193: Are there any rules for what these strings should be?

core/identity-spi/src/main/java/software/amazon/awssdk/identity/spi/internal/ProviderSource.java
```diff
@@ -0,0 +1,35 @@
...
+import java.util.Optional;
+import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.annotations.ThreadSafe;
+
+/**
+ * Interface to represent the provider that is the source of this identity.
+ */
+@SdkInternalApi
+@ThreadSafe
+public interface ProviderSource {
```
millems, line 27: 1. This cannot be internal, since it's implemented by public classes.
2. The name is weird. It sounds like it provides providers? Maybe "Provided" or "ProviderAware"?

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/AwsBasicCredentials.java
cenedhryn: yes

core/identity-spi/src/main/java/software/amazon/awssdk/identity/spi/internal/ProviderSource.java
cenedhryn: updated

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/AwsBasicCredentials.java
millems: Okay, we should update the javadoc at that time.

millems (approved):

PR #5016, opened 2024-03-14
Add Warnings in the async client interface and CrtClient against passing Runnable::run

anirudh9391:
Adding warnings to discourage users from configuring main I/O threads to run completion future objects as it may lead to deadlocks


codegen/src/main/java/software/amazon/awssdk/codegen/poet/client/AsyncClientInterface.java
```diff
@@ -169,7 +169,10 @@ public ClassName className() {
 
     private String getJavadoc() {
         return "Service client for accessing " + model.getMetadata().getDescriptiveServiceName() + " asynchronously. This can be "
-               + "created using the static {@link #builder()} method.\n\n" + model.getMetadata().getDocumentation();
+               + "created using the static {@link #builder()} method.\nWe strongly discourage "
+               + "configuring main I/O threads to run future-completion objects. This could "
+               + "lead to deadlock as the SDK may perform blocking calls in some cases."
+               + "\n\n" + model.getMetadata().getDocumentation();
```
zoewangg, line 175: This may be a bit confusing for customers who are not aware of future completion config. Suggesting the following:
```
... This can be created using the static `builder()` method.  
<p>
The async client performs non-blocking I/O when configured with any {@link SdkAsyncHttpClient} 
supported in the SDK. However, full non-blocking is not guaranteed as the async client may perform 
blocking calls in some cases such as credentials fetching, endpoint discovery as part of the async API call.
```

services/s3/src/main/java/software/amazon/awssdk/services/s3/S3CrtAsyncClientBuilder.java
```diff
@@ -322,11 +322,13 @@ default S3CrtAsyncClientBuilder retryConfiguration(Consumer<S3CrtRetryConfigurat
      *     <li>You want more fine-grained control over the {@link ThreadPoolExecutor} used, such as configuring the pool size
      *     or sharing a single pool between multiple clients.
      *     <li>You want to add instrumentation (i.e., metrics) around how the {@link Executor} is used.
-     *     <li>You know, for certain, that all of your {@link CompletableFuture} usage is strictly non-blocking, and you wish to
-     *     remove the minor overhead incurred by using a separate thread. In this case, you can use
-     *     {@code Runnable::run} to execute the future-completion directly from within the I/O thread.
      * </ol>
      *
```
zoewangg, line 326: nit: can we add `<p>` tag? https://github.com/aws/aws-sdk-java-v2/blob/master/docs/design/APIReference.md

services/s3/src/main/java/software/amazon/awssdk/services/s3/S3CrtAsyncClientBuilder.java
```diff
@@ -323,12 +323,12 @@ default S3CrtAsyncClientBuilder retryConfiguration(Consumer<S3CrtRetryConfigurat
...
      *     <li>You want to add instrumentation (i.e., metrics) around how the {@link Executor} is used.
      * </ol>
-     *
+     * <p>
      * <b>WARNING</b>
      * We strongly <strong>discourage</strong> using {@code Runnable::run}, which executes the future-completion directly from
      * within the I/O thread because it may block the I/O thread and cause deadlock, especially if you are sending
      * another SDK request in the {@link CompletableFuture} chain since the SDK may perform blocking calls in some cases.
-     *
+     * </p>
```
zoewangg, line 331: nit, only a single tag is needed.
> A single `<p>` MUST be used between paragraphs.

zoewangg (approved):

PR #5024, opened 2024-03-19
normalize key bugfix in TM

L-Applin:
Fix normalize key method to not strip extra character when prefix is not immediately followed by separator in the key

## Motivation and Context
Bug raised by internal team.

The sdk strip an extra character from the prefix when the prefix used is not the full name of a folder. For example: given a request with `prefix="top-"` , `delimiter="/"` and `key="top-level/sub-folder/1.txt"`, transfer-manager would incorrectly use the `evel/sub-folder/1.txt` path to save that object.

## Modifications
We now only strip an extra character if the key still starts with the separator after being stripped of the prefix.

## Testing
Added unit test validating previous behaviour is the same and that the incorrect behaviour is fixed.

## Types of changes


services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/DownloadDirectoryHelper.java
```diff
@@ -148,7 +148,7 @@ private Path determineDestinationPath(DownloadDirectoryRequest downloadDirectory
                                           S3Object s3Object) {
         FileSystem fileSystem = downloadDirectoryRequest.destination().getFileSystem();
         String delimiter = listRequest.delimiter() == null ? DEFAULT_DELIMITER : listRequest.delimiter();
```
cenedhryn, line 150: You can remove this line, it's superfluous

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/DirectoryHelperUtils.java
```diff
@@ -0,0 +1,64 @@
...
+
+    private DirectoryHelperUtils() {
+    }
+
+    /**
+     * If the prefix is not empty AND the key contains the delimiter, normalize the key by stripping the prefix from the key. If a
+     * delimiter is null (not provided by user), use "/" by default. For example: given a request with prefix = "notes/2021"  or
+     * "notes/2021/", delimiter = "/" and key = "notes/2021/1.txt", the normalized key should be "1.txt".
+     */
+    static String normalizeKey(ListObjectsV2Request listObjectsRequest,
```
cenedhryn, line 36: Can the parameter be `prefix` instead of `listObjectsRequest`, since this logic doesn't really need to know about the listObjectsRequest, and move the validation up to the calling code?

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/DirectoryHelperUtils.java
```diff
@@ -0,0 +1,64 @@
...
+                                      String key,
+                                      String delimiter) {
+        Validate.paramNotNull(listObjectsRequest, "listObjectsRequest must not be null");
+        String delimiterToUse = Validate.getOrDefault(delimiter, () -> DEFAULT_DELIMITER);
+
+        if (StringUtils.isEmpty(key) || StringUtils.isEmpty(listObjectsRequest.prefix())) {
+            return key;
+        }
+
+        String prefix = listObjectsRequest.prefix();
```
cenedhryn, line 46: Can we move this up to after delimiterToUse

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/DirectoryHelperUtils.java
```diff
@@ -0,0 +1,64 @@
...
+@SdkInternalApi
+final class DirectoryHelperUtils {
+
+    private DirectoryHelperUtils() {
+    }
+
+    /**
+     * If the prefix is not empty AND the key contains the delimiter, normalize the key by stripping the prefix from the key. If a
+     * delimiter is null (not provided by user), use "/" by default. For example: given a request with prefix = "notes/2021"  or
+     * "notes/2021/", delimiter = "/" and key = "notes/2021/1.txt", the normalized key should be "1.txt".
```
cenedhryn, line 34: Can we include an example for a prefix that is not aligned with a delimiter in the path (ie the error case).

services-custom/s3-transfer-manager/src/test/java/software/amazon/awssdk/transfer/s3/internal/DirectoryHelperUtilsTest.java
```diff
@@ -0,0 +1,69 @@
...
+            arg("", "delim/with/separator", "delim/with/separator"),
+            arg("no-delim", "", ""),
+            arg("no-delim", "no-delim", "no-delim"),
+            arg("delim", "delim/", ""),
+            arg("prefix", "not-in-key", "not-in-key"),
+            arg("notes/2021", "notes/2021/1.txt", "1.txt"),
+            arg("notes/2021/", "notes/2021/1.txt", "1.txt"),
+            arg("someInner", "someInnerFolder/another/file1.txt", "Folder/another/file1.txt"),
+            arg("someInner", "someInnerF/another/file1.txt", "F/another/file1.txt"),
+            arg("someInner", "someInner/another/file1.txt", "another/file1.txt")
```
cenedhryn, line 61: Can we also add
`arg("someInner/a", "someInner/another/file1.txt", "nother/file1.txt")`

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/DownloadDirectoryHelper.java
L-Applin: the delimiter is also used for the call to `getRelativePath`

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/DirectoryHelperUtils.java
L-Applin: good point, but instead I moved `String delimiterToUse = Validate.getOrDefault(delimiter, () -> DEFAULT_DELIMITER);` down after "StringUtils.isEmpty" validations.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/DownloadDirectoryHelper.java
```diff
@@ -197,37 +197,6 @@ private CompletableFuture<CompletedFileDownload> doDownloadSingleFile(DownloadDi
...
 
-    /**
-     * If the prefix is not empty AND the key contains the delimiter, normalize the key by stripping the prefix from the key.
-     *
-     * If a delimiter is null (not provided by user), use "/" by default.
-     *
-     * For example: given a request with prefix = "notes/2021"  or "notes/2021/", delimiter = "/" and key = "notes/2021/1.txt",
-     * the normalized key should be "1.txt".
-     */
-    private static String normalizeKey(ListObjectsV2Request listObjectsRequest,
```
zoewangg, line 208: Minor: it'd be great if we can create a separate commit for refactoring in the future since it's hard to tell the diff and easy to miss things if we do refactoring (extracting this method out of this class) and fixing in the same commit. It'd also make reviews easier.

zoewangg:
Can we add a changelog entry? Can we also update `Motivation and Context` to provide more details? An example of what this code change trying to fix would be great.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/DownloadDirectoryHelper.java
cenedhryn: Ah, didn't see that

cenedhryn (approved):

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/DownloadDirectoryHelper.java
L-Applin: will keep that in mind for future pr

PR #5029, opened 2024-03-20
Adding identity provider source to user agent string

cenedhryn:
## Motivation and Context
This PR adds the identity source (which credentials provider that was used to resolve the identity / credentials) to the user agent string. 

The `provider` of the identity is added to the `cfg` part of the user agent as an `auth-source`. Auth source is a new section and not ratified in any specification, but should be parseable by the ua metrics system. 

We choose to add a short form of the credentials provider class as an identifier for now. This can easily be changed as the requirements for the user agent string format and contents are changed, and is backwards compatible.

In this PR, credential providers are represented by the`StaticCredentialsProvider`. A follow up PR will have implementations for all other credentials providers. 

## Modifications
- Adds a provider name to `StaticCredentialsProvider`
- Adds this name to the identity returned by the provider. Because this credentials provider returns externally created objects, it will make a copy of the object and add the provider string to the builder. 
- In the `ApplyUserAgentStage`, the SDK adds the identity source to the user agent string if present. 
- The value of the identity source is validated and in most cases transformed.
  - It's mapped to known values in order to get a shorter form of the name. `StaticCredentialsProvider` <-> `STAT`
  - If the value is not known, it will be added as-is if it 
    - contains known characters
    - is shorter than 62 characters (our longest credentials provider has a class name of 47 characters)
    - ends with `provider` (debatable)
 
#### Q. Why not add the short name to the provider class
Most providers contain a `toString` that prints the name. There's a value in getting the complete name of the provider. In addition, in the future we may not use this short form as the usage of the UA string evolves

#### Q. How is this addition compatible with the proposed changes to the UA string?
When the SDK adds feature ID to the UA string, the value written to the identity by each provider can be changed to a different value, such as a string digit.


core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/useragent/IdentityProviderNameMapping.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    public static Optional<String> fromValue(String value) {
+        if (value == null) {
+            return Optional.empty();
+        }
+
+        for (IdentityProviderNameMapping provider : values()) {
+            if (provider.value().equals(value)) {
+                return Optional.of(provider.name());
```
cenedhryn, line 59: Leaning towards using the lower case variant of the enum, `stat` instead of `STAT`.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/ApplyUserAgentStage.java
```diff
@@ -140,6 +154,32 @@ private String getUserAgent(SdkClientConfiguration config, List<ApiName> request
...
+            .getAttribute(SdkInternalExecutionAttribute.SELECTED_AUTH_SCHEME);
+        if (selectedAuthScheme == null) {
+            return Optional.empty();
+        }
+        return providerNameFromIdentity(selectedAuthScheme);
+    }
+
+    private static <T extends Identity> Optional<String> providerNameFromIdentity(SelectedAuthScheme<T> selectedAuthScheme) {
+        CompletableFuture<? extends T> identityFuture = selectedAuthScheme.identity();
+        T identity = CompletableFutureUtils.joinLikeSync(identityFuture);
```
anirudh9391, line 168: Looks like this is another place where we are making it wait for a future. Do we just want to add a comment indicating that we are blocking here and we eventually would need to remove this ?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/ApplyUserAgentStage.java
```diff
@@ -16,43 +16,56 @@
...
 import software.amazon.awssdk.http.SdkHttpFullRequest;
 import software.amazon.awssdk.http.async.SdkAsyncHttpClient;
+import software.amazon.awssdk.identity.spi.Identity;
+import software.amazon.awssdk.utils.CompletableFutureUtils;
 import software.amazon.awssdk.utils.Logger;
 import software.amazon.awssdk.utils.StringUtils;
 import software.amazon.awssdk.utils.http.SdkHttpUtils;
 
-/**
- * Apply any custom user agent supplied, otherwise instrument the user agent with info about the SDK and environment.
```
anirudh9391, line 39: Would help to have some Javadoc

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/useragent/IdentityProviderNameMapping.java
```diff
@@ -0,0 +1,75 @@
...
+import java.util.Locale;
+import java.util.Optional;
+import java.util.regex.Pattern;
+
+/**
+ * A enum class representing a short form of identity providers to record in the UA string.
+ */
+public enum IdentityProviderNameMapping {
+
+    SYS("SystemPropertyCredentialsProvider"),
```
anirudh9391, line 27: As discussed, would be great to expose these abbreviations to our internal users to avoid any confusions as to which abbreviation maps to which provider (probably in a separate action item)

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/useragent/IdentityProviderNameMapping.java
```diff
@@ -0,0 +1,75 @@
...
+        for (IdentityProviderNameMapping provider : values()) {
+            if (provider.value().equals(value)) {
+                return Optional.of(provider.name());
+            }
+        }
+        return Optional.ofNullable(sanitizedProviderOrNull(value));
+    }
+
+    private static String sanitizedProviderOrNull(String value) {
+        if (containsAllowedCharacters(value) && value.toLowerCase(Locale.US).endsWith("provider")) {
```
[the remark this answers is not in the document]
cenedhryn: We decided offline to remove this criterion

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/ApplyUserAgentStage.java
cenedhryn: I don't think it's necessary; it's easy to search for this in the code and that's something we'd have to do anyway when we've unblocked async identity handling.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/ApplyUserAgentStage.java
cenedhryn: I think I removed it because I had started a refactoring. Added a new text back.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/useragent/IdentityProviderNameMapping.java
cenedhryn: This will be addressed when parsing the metrics.

anirudh9391 (approved):

PR #5033, opened 2024-03-22
Process credentials list of strings args

davidh44:
## Motivation and Context

Currently, setting an executable command on the `ProcessCredentialsProvider` is done by passing a single String, which will be executed in a shell.

## Modifications

Add a setter to pass in an executable command using a list-of-Strings, which will be executed not-in-a-shell.
Deprecating the use of the single-String setter.

## Testing

Added unit test

## Types of changes

## Checklist

## License


core/auth/src/main/java/software/amazon/awssdk/auth/credentials/ProcessCredentialsProvider.java
```diff
@@ -271,13 +282,23 @@ public Builder asyncCredentialUpdateEnabled(Boolean asyncCredentialUpdateEnabled
...
          */
         public Builder command(String command) {
             this.command = command;
             return this;
         }
 
+        /**
+         * Configure the commands that should be executed to retrieve credentials.
+         */
+        public Builder commands(List<String> commands) {
```
millems, line 297: It's still a single command. Maybe still just "command"?

joviegas:
Can we please mention the motivation and some details for this PR

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/ProcessCredentialsProvider.java
```diff
@@ -76,30 +76,19 @@ public final class ProcessCredentialsProvider
 
     private final String commandFromBuilder;
 
+    private final List<String> commandAsListOfStringsFromBuilder;
```
joviegas, line 79: Why are we not naming it as 
`commandsFromBuilder`

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/ProcessCredentialsProvider.java
```diff
@@ -272,12 +283,24 @@ public Builder asyncCredentialUpdateEnabled(Boolean asyncCredentialUpdateEnabled
...
+        @Deprecated
         public Builder command(String command) {
             this.command = command;
             return this;
         }
 
+        /**
+         * Configure the command that should be executed to retrieve credentials, as a list of strings.
+         */
+        public Builder command(List<String> commandAsListOfStrings) {
```
joviegas, line 299: Why is it not names as `commands` since it takes args as a list of objects ?

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/ProcessCredentialsProvider.java
davidh44: Its still executed as a single command, the user is just specifying it in a list, so that it can be executed not-in-a-shell

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/ProcessCredentialsProvider.java
```diff
@@ -111,6 +100,26 @@ private ProcessCredentialsProvider(Builder builder) {
         this.processCredentialCache = cacheBuilder.build();
     }
 
+    private List<String> executableCommand(Builder builder) {
```
gosar, line 103: This means that if someone, with say a command that needed no args, migrates from
```
            ProcessCredentialsProvider.builder().command("script")).build()
```
to
```
            ProcessCredentialsProvider.builder().command(Arrays.asList("script"))).build()
```
because they see the deprecation warning on the previous method, there will be a behavior change where it used to be run in a shell and now it won't? I'm not too sure, so just asking to double-check if there are risks with that behavior change?

davidh44: Yeah thats correct, that is the intent of the change, in order to remove risk of command injections. I don't think there is a risk with changing to not-in-a-shell. I've added [description in the javadocs for the deprecated method](https://github.com/aws/aws-sdk-java-v2/blob/hdavidh/process-creds-list-of-strings/core/auth/src/main/java/software/amazon/awssdk/auth/credentials/ProcessCredentialsProvider.java#L288)  so users should be aware, they can continue to use it if they need the command executed in a shell

millems (approved):

PR #4897, opened 2024-02-07
Handle header lists in URLConnection client

dagnir:
## Motivation and Context
Fix a bug where headers with multiple values don't have all values for that header sent on the wire. This leads to signature mismatch exceptions.

Fixes #4746.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)

## Checklist

- [ ] I have read the [CONTRIBUTING](https://github.com/aws/aws-sdk-java-v2/blob/master/CONTRIBUTING.md) document
- [ ] Local run of `mvn install` succeeds
- [ ] My code follows the code style of this project
- [ ] My change requires a change to the Javadoc documentation
- [ ] I have updated the Javadoc documentation accordingly
- [ ] I have added tests to cover my changes
- [ ] All new and existing tests passed
- [ ] I have added a changelog entry. Adding a new entry must be accomplished by running the `scripts/new-change` script and following the instructions. Commit the new file created by the script in `.changes/next-release` with your changes.
- [ ] My change is to implement 1.11 parity feature and I have updated [LaunchChangelog](https://github.com/aws/aws-sdk-java-v2/blob/master/docs/LaunchChangelog.md)

## License

- [ ] I confirm that this pull request can be released under the Apache 2 license

Issue #4746, opened 2023-12-05
S3 PUT http with url-connection-client contentEncoding is always "aws-chunked"


youngm:
### Describe the bug

When putting an S3Object connecting to S3 via http and url-connection-client, the contentEncoding for that object is always "aws-chunked".

### Expected Behavior

The contentEncoding should be whatever is set in the `PutObjectRequest`.

### Current Behavior

contentEncoding is always `aws-chunked` when connecting to s3 via http and url-connection-client.

### Reproduction Steps

This test creates an object using http s3 and sets a contentEncoding. Note that the contentEncoding is changed to "aws-chunked" by the sdk.

S3Test.java
```
public class S3Test {

  @Test
  public void test() throws Exception {
    try (var s3Client =
        S3Client.builder()
            .endpointOverride(URI.create("http://s3.amazonaws.com"))
            .build()) {
    var putRequest =
        PutObjectRequest.builder()
            .bucket("bucket")
            .key("file")
            .contentEncoding("bogus")
            .build();

    s3Client.putObject(putRequest, RequestBody.empty());
    try (var response =
        s3Client.getObject(GetObjectRequest.builder().bucket("bucket").key("file").build())) {
    assertThat(response.response().contentEncoding()).isEqualTo("bogus");
    }
    }
  }
}
```
pom.xml
```
<project xmlns="http://maven.apache.org/POM/4.0.0"
	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
	<parent>
		<groupId>org.springframework.boot</groupId>
		<artifactId>spring-boot-starter-parent</artifactId>
		<version>3.1.5</version>
	</parent>
	<modelVersion>4.0.0</modelVersion>
	<groupId>s3test</groupId>
	<artifactId>s3test</artifactId>
	<version>1.0</version>
	<properties>
		<java.version>17</java.version>
	</properties>
	<dependencyManagement>
		<dependencies>
			<dependency>
				<groupId>software.amazon.awssdk</groupId>
				<artifactId>bom</artifactId>
				<version>2.21.36</version>
				<type>pom</type>
				<scope>import</scope>
			</dependency>
		</dependencies>
	</dependencyManagement>
	<dependencies>
		<dependency>
			<groupId>software.amazon.awssdk</groupId>
[...]

PR #5043, opened 2024-03-25
Revert "Revert "Handle header lists in URLConnection client (#4897)" (#4960)"

millems:
This reverts commit e715bb6c9554de9ee39a8e8c33f5c7a2626f4615.

X-ray has fixed a risk of customers double-adding the x-ray interceptor, and SQS will no longer reject requests with "invalid" x-ray trace IDs. It should be safe to re-release this URL connection fix for multi-value headers.


dagnir (approved):

ngudbhav:
Was this code change reverted? We are facing this issue again here https://github.com/aws/aws-sdk-java-v2/issues/4746#issuecomment-2276069405

PR #5056, opened 2024-03-29
Modify SDK to not fail on invalid Expires header

anirudh9391:
Implementing a SEP to modify behavior while parsing S3 Expires header. 
## Motivation and Context
- SEP to implement an alternate way of unmarshalling exception handling for the "Expires" header. 
## Modifications
- SDKs will not throw an exception if an invalid expires header is present. 
- "Expires" header will be defined as a timestamp regardless of any changes to its model definition. 

## Testing
Defined a custom unit test and validated by running it once. Source code 
```
    @Test
    void test() throws IOException {

        String bucketName = "anirudkr-bucket-2";
        String key = "key1.txt";

        //File file = new RandomTempFile(10_000);
        S3Client client = S3Client.builder().build();

        //PutObjectRequest put = PutObjectRequest.builder().bucket(bucketName).key(key).expires(Instant.now()).build();
        //client.putObject(put, file.toPath());

        LogCaptor logCaptor = LogCaptor.create();

        HeadObjectRequest request = HeadObjectRequest.builder().bucket(bucketName).key(key).build();
        Assertions.assertThatCode(() -> client.headObject(request))
                  .doesNotThrowAnyException();
        assertTrue(client.headObject(request).expires() == null);
        assertTrue(client.headObject(request).expiresString() != null);

        List<LogEvent> events = logCaptor.loggedEvents();
        String expected = "Invalid datetime format provided in the Expires field 2034-01-01T00:00:00Z";
    }
```

Runs successfully. 

The Expires header on this bucket is an invalid format 
```2034-01-01T00:00:00Z```

Test validates 
1) exception is not logged
2) warning is thrown

## Types of changes

- [x ] Bug fix (non-breaking change which fixes an issue)

### CODE REVIEW

**Review by Crissttianreyes — COMMENTED** [review]
payments2_api1.wix.com

**Comment by anirudh9391** [comment]
> payments2_api1.wix.com

Hello, is this unintentional ?

**Comment by L-Applin** [comment]
I don't see the mentioned unit test in the added code? Also, should we add unit-test the `HeaderUnmarshaller .INSTANT` un-marshaller?

**Comment by L-Applin** [comment]
I think we should include a codegen test for the added trait

File: core/protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/internal/unmarshall/HeaderUnmarshaller.java [hunk 1]
```diff
@@ -69,9 +73,17 @@ private SimpleHeaderUnmarshaller(StringToValueConverter.StringToValue<T> stringT
...
         public T unmarshall(XmlUnmarshallerContext context, List<XmlElement> content, SdkField<T> field) {
-            return context.response().firstMatchingHeader(field.locationName())
-                          .map(s -> stringToValue.convert(s, field))
-                          .orElse(null);
+            try {
+                return context.response().firstMatchingHeader(field.locationName())
+                              .map(s -> stringToValue.convert(s, field))
+                              .orElse(null);
+            } catch (RuntimeException e) {
+                log.warn("Exception found while parsing response header {}", e.getMessage());
```
L-Applin, line 81: I wonder if this might end up being too noisy for users. Maybe not, I'm not sure... Could there be use cases where their logs get swarmed with this warning? Like a user accessing the same object which has this bad header a lot?

services/s3/src/main/resources/codegen-resources/customization.config
L-Applin, line 1: For my own knowledge and understanding, I have a few questions about these customization changes:
- Why do we need to update the Requests objects? I thought we needed to change the un-marshalling, which would be for `PutObjectResponse` (for example), right? Why do we need to change `PutObjectRequest` as well?
-  I thought these request object's `.expire(...)` method already accept Instant, the `emitAsType: timestamp` is still required?

core/protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/internal/unmarshall/HeaderUnmarshaller.java
anirudh9391: I do not think this is a very common use case. Moreover, this is an explicit SEP requirement 
```
The SDKs MUST log a warning message and return null for that field.
```
https://code.amazon.com/packages/AwsDrSeps/blobs/main/--/seps/accepted/shared/s3-expires-shape-customization.md

services/s3/src/main/resources/codegen-resources/customization.config
anirudh9391: - This is to ensure backwards compatibility
```
Model Expires as timestamp
C2J and Smithy will change Expires from a timestamp to a string in the future  Existing SDKs that currrently model Expires as timestamp MUST add customization to ensure it stays as timestamp regardless of model definition for backwards compatibility.
```
This should apply to all shapes that model Expires. My understanding from this is all shapes with Expires will be modeled as a String going forward; for backwards compatibility, we need to ensure models stay timestamps.

anirudh9391:
> I don't see the mentioned unit test in the added code? Also, should we add unit-test the `HeaderUnmarshaller.INSTANT` un-marshaller?

Im working on the wiremock tests. I think the wiremock should cover the code path

> I don't see the mentioned unit test in the added code? Also, should we add unit-test the `HeaderUnmarshaller.INSTANT` un-marshaller?

Im working on the wiremock tests, should publish that along with the next revision.

services/s3/src/main/resources/codegen-resources/customization.config
L-Applin: Ah thanks! Yes of course, when the type will change in the model, we don't want the method type to change.

core/protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/internal/unmarshall/HeaderUnmarshaller.java
```diff
@@ -61,6 +63,8 @@ private HeaderUnmarshaller() {
 
     private static class SimpleHeaderUnmarshaller<T> implements XmlUnmarshaller<T> {
 
+        private static final Logger log = LoggerFactory.getLogger(SimpleHeaderUnmarshaller.class);
```
zoewangg, line 66: We should use SDK Logger https://github.com/aws/aws-sdk-java-v2/blob/master/utils/src/main/java/software/amazon/awssdk/utils/Logger.java#L26

services/s3/src/test/java/software/amazon/awssdk/services/s3/ExpiresHeaderDataTypeErrorTest.java
L-Applin, line 1: Could we use junit 5 instead of junit 4?

core/protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/internal/unmarshall/HeaderUnmarshaller.java
anirudh9391: Done

services/s3/src/test/java/software/amazon/awssdk/services/s3/ExpiresHeaderDataTypeErrorTest.java
```diff
@@ -0,0 +1,91 @@
...
+
+    @Test
+    public void headObjectRequestWithInvalidDate_doesNotThrowException(WireMockRuntimeInfo wm) throws IOException {
+
+        stubFor(any(anyUrl())
+                    .willReturn(aResponse()
+                                    .withHeader("Expires", TEST_DATE)
+                                    .withBody("Hello world!")));
+
+        S3Client s3Client = getS3Client(wm);
```
L-Applin, line 52: for Wiremock tests, I think you can use `@BeforeEach` with `WireMockRuntimeInfo`:

```java
private S3Client s3Client;

@BeforeEach
void init(WireMockRuntimeInfo wm) {
   this.s3Client = ...;
}
```

This would avoid having to call `getS3Client(wm)` manually in every test case.

core/protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/internal/unmarshall/HeaderUnmarshaller.java
```diff
@@ -69,9 +71,17 @@ private SimpleHeaderUnmarshaller(StringToValueConverter.StringToValue<T> stringT
...
         public T unmarshall(XmlUnmarshallerContext context, List<XmlElement> content, SdkField<T> field) {
-            return context.response().firstMatchingHeader(field.locationName())
-                          .map(s -> stringToValue.convert(s, field))
-                          .orElse(null);
+            try {
+                return context.response().firstMatchingHeader(field.locationName())
+                              .map(s -> stringToValue.convert(s, field))
+                              .orElse(null);
+            } catch (RuntimeException e) {
+                log.warn(() -> (String.format("Exception found while parsing response header {}", e.getMessage())));
```
zoewangg, line 79: nit, we should probably log the entire stacktrace. 
`log.warn(() -> "Exception found while parsing response header", e));`

services/s3/src/test/java/software/amazon/awssdk/services/s3/ExpiresHeaderDataTypeErrorTest.java
```diff
@@ -0,0 +1,91 @@
...
+        Assertions.assertThatCode(() -> s3Client.headObject(r -> {
+                      r.bucket("s3_expires_test_dummy_bucket")
+                       .key("s3_expires_test_dummy_key");
+                  }))
+                  .doesNotThrowAnyException();
+
+        assertTrue(s3Client.headObject(r -> {r.bucket("s3_expires_test_dummy_bucket")
+                                              .key("s3_expires_test_dummy_key");}).expires() == null);
+
+        assertEquals(s3Client.headObject(r -> {r.bucket("s3_expires_test_dummy_bucket")
```
zoewangg, line 65: nit, assertJ is preferred because it provides better error message.
`assertThat(s3Client.headObject(r -> r.bucket("s3_expires_test_dummy_bucket")).isEqualTo(xxx)`

core/protocols/aws-xml-protocol/src/main/java/software/amazon/awssdk/protocols/xml/internal/unmarshall/HeaderUnmarshaller.java
anirudh9391: hmm.. I thoght about that, but felt that this would provide the user with sufficient info. The message would be something like 'Unable to parse Expires header at index 1'. Do you think providing a trace gives more useful info ?

services/s3/src/test/java/software/amazon/awssdk/services/s3/ExpiresHeaderDataTypeErrorTest.java
anirudh9391: Good siggestion. changed that

L-Applin (approved):

PR #5059, opened 2024-04-01
Implement TransferListener for COPY for TransferManager with Multipar…

davidh44:
…tS3AsyncClient

## Motivation and Context

TransferListener does not update progress for Copy operations

## Modifications

Add progress listener to `CopyObjectRequest` and report progress in `UploadPartCopyRequestIterable`

## Testing

added unit and integ tests

## Types of changes

## Checklist

## License


services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/GenericS3TransferManager.java
```diff
@@ -457,9 +472,17 @@ public final Copy copy(CopyRequest copyRequest) {
 
         CompletableFuture<CompletedCopy> returnFuture = new CompletableFuture<>();
 
-        TransferProgressUpdater progressUpdater = new TransferProgressUpdater(copyRequest, null);
-        progressUpdater.transferInitiated();
-        progressUpdater.registerCompletion(returnFuture);
+        // set length to 10000 as reference value, since we don't make HeadObject call yet
+        TransferProgressUpdater progressUpdater = new TransferProgressUpdater(copyRequest, 10000L);
```
anirudh9391, line 476: Is this an accurate value ? The TransferProgressUpdater uses the total content size to calculate what fraction of the request body has been sent, etc. If this value is inaccurate, could lead to incorrect progress states being relayed to users.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/GenericS3TransferManager.java
```diff
@@ -457,9 +472,17 @@ public final Copy copy(CopyRequest copyRequest) {
...
+        if (isS3ClientMultipartEnabled()) {
```
anirudh9391, line 477: Do we only want to enable progress listening if it is a multi-part request ? Looks like the behaviour previously called progress updates regardless

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/UploadPartCopyRequestIterable.java
```diff
@@ -32,18 +35,30 @@ public final class UploadPartCopyRequestIterable implements SdkIterable<UploadPa
...
     public UploadPartCopyRequestIterable(String uploadId,
                                          long partSize,
                                          CopyObjectRequest copyObjectRequest,
                                          long remainingBytes) {
         this.uploadId = uploadId;
         this.optimalPartSize = partSize;
         this.copyObjectRequest = copyObjectRequest;
+        this.numberOfParts = (long) Math.ceil((double) remainingBytes / partSize);
+        double partPercentage = (double) 1 / numberOfParts;
+        this.scaledPartValue = (long) (partPercentage * 10000);
```
anirudh9391, line 57: I see we are scaling it to 10000 here. Got it.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/GenericS3TransferManager.java
davidh44: This is basically a check for Java-based vs CRT-based client. Supporting this for CRT would require additional work on the CRT side. Currently INFO logs print `"TransferInitiated"` and nothing after that which isn't ideal. Think its better to just not print anything at all, wdyt?

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/UploadPartCopyRequestIterable.java
```diff
@@ -70,12 +85,23 @@ public UploadPartCopyRequest next() {
...
                                                                uploadId,
                                                                range);
+            updateProgressListener();
             partNumber++;
             offset += partSize;
             remainingBytes -= partSize;
             return uploadPartCopyRequest;
         }
 
+        private void updateProgressListener() {
```
anirudh9391, line 95: WOuld be great if we could see some comments to describe this logic, with some number examples. 

What I understand is that 10000 is the total size of all parts put together. If say we need to transfer 100 bytes, and the size of a part is 10 bytes, it gives us 10 parts. Scaled to 10000, it is (1/10) * 10000 which is 1000 parts. Is this called eveyrtime we complete the upload of a part ?

anirudh9391: Can we make byte updates when the upload of a part is in progress ? Like lets say we have 10 parts of 10 bytes each, could we make a progress update in the middle of the 6th parts upload with the number of bytes remaining ? Im guessing it is possible for streaming operations but unsure if TransferProgressUpdate supports this.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/GenericS3TransferManager.java
anirudh9391: nvm found the answer below

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/GenericS3TransferManager.java
anirudh9391: I see, so it is nothing meaningful. Got it, this way is better. Would be great to see that in a comment though

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/UploadPartCopyRequestIterable.java
davidh44: Added comments to describe logic. This isn't a public method/class so we might not need to be too detailed.

We send the CopyPart request to S3 which handles the operation, so we don't have access to the progress of each individual CopyPart.

anirudh9391 (approved):

PR #5063, opened 2024-04-03
S3 Event Notification javadoc improvement

L-Applin:
Improvement to the java doc of the S3 Event Notification module, including links to S3 User Guide.

## Motivation and Context
Was missing from original release.

## Modifications
Added javadoc on multiple classes and getters.


cenedhryn (approved):
This update should be very helpful to the users, I appreciate going the extra mile and explaining some important context as well as linking to the guides. 

The caveat is that these links may not be there always, but it's probably worth the risk.

PR #5089, opened 2024-04-09
Update javadoc regarding exception handling in async clients

zoewangg:
## Motivation and Context
The exception returned by the SDK client is wrapped with `CompletionException` see below. We can't change it due to backwards compatibility reasons, so we are updating this doc to mention this behavior.
```
    var sqs = SqsAsyncClient.create();
    sqs.getQueueUrl(GetQueueUrlRequest.builder().queueName("invalidqueuename").build())
        .whenComplete((url, e) -> {
            System.out.println("Exception: " + e);
        })
        .join();
}
```

prints out
```
Exception: java.util.concurrent.CompletionException: software.amazon.awssdk.services.sqs.model.QueueDoesNotExistException: 
The specified queue does not exist for this wsdl version. (Service: Sqs, Status Code: 400, Request ID: xxxx)
```

## Modifications
Update documentation

### DIFF
--- a/codegen/src/main/java/software/amazon/awssdk/codegen/docs/DocumentationBuilder.java
+++ b/codegen/src/main/java/software/amazon/awssdk/codegen/docs/DocumentationBuilder.java
@@ -33,7 +33,9 @@ public final class DocumentationBuilder {
     // TODO This prefix is not suitable for paginated operations. Either remove it for paginated operations
     // or change the statement to something generic
     private static final String ASYNC_THROWS_PREFIX = "The CompletableFuture returned by this method can be completed " +
-                                                      "exceptionally with the following exceptions.";
+                                                      "exceptionally with the following exceptions. The exception returned is "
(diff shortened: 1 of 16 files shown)


codegen/src/main/java/software/amazon/awssdk/codegen/docs/DocumentationBuilder.java
```diff
@@ -33,7 +33,9 @@ public final class DocumentationBuilder {
     // TODO This prefix is not suitable for paginated operations. Either remove it for paginated operations
     // or change the statement to something generic
     private static final String ASYNC_THROWS_PREFIX = "The CompletableFuture returned by this method can be completed " +
-                                                      "exceptionally with the following exceptions.";
+                                                      "exceptionally with the following exceptions. The exception returned is "
```
zoewangg, line 36: I'll update the tests accordingly after this doc change is approved.

dagnir (approved):

PR #5101, opened 2024-04-11
Use Content-Range in multipart download for totalBytes

L-Applin:
Change `TransferProgressUpdater` so it uses Content-Range header to determine the maximum number of bytes it tracks as part of the transfer progress listener.

## Motivation and Context
For downloads, `TransferProgressUpdater` was expecting the totalBytes of the whole s3 object to be the value of the `Content-Length` header, when in fact the `Content-Length` header only contains the size of the current part. This cause the value of transferred bytes to exceed to the number total byte expected as soon as a 2nd part was dowloaded.

## Modifications
Add a `ContentRangeParser` class to extract the total object size from the `Content-Range` header, following the **[MDN refgerence](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Range)**. Only support `bytes` unit.

## Testing
- Added unit tests.
- Manual testing with s3 object.

### DIFF
--- a/services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/progress/ContentRangeParser.java
+++ b/services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/progress/ContentRangeParser.java
@@ -28,9 +28,12 @@ import software.amazon.awssdk.utils.StringUtils;
  * The only supported {@code <unit>} is the {@code bytes} value.
  */
 @SdkInternalApi
-public class ContentRangeParser {
+public final class ContentRangeParser {
 
     private static final Logger log = Logger.loggerFor(ContentRangeParser.class);
+
+    private ContentRangeParser() {
+    }
 
     /**
      * Parse the Content-Range to extract the total number of byte from the content. Only supports the {@code bytes} unit, any
@@ -41,7 +44,7 @@ public class ContentRangeParser {
      * @return The total number of bytes in the content range or an empty optional if the contentRange is null, empty or if the
      * total length is not a valid long.
      */
-    public OptionalLong totalBytes(String contentRange) {
+    public static OptionalLong totalBytes(String contentRange) {
         if (StringUtils.isEmpty(contentRange)) {
             return OptionalLong.empty();
         }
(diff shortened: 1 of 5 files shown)


services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/progress/ContentRangeParser.java
```diff
@@ -0,0 +1,72 @@
...
+ * The only supported {@code <unit>} is the {@code bytes} value.
+ */
+@SdkInternalApi
+public class ContentRangeParser {
+
+    private static final Logger log = Logger.loggerFor(ContentRangeParser.class);
+
+    /**
+     * Parse the Content-Range to extract the total number of byte from the content. Only supports the {@code bytes} unit, any
+     * other unit will result in an empty OptionalLong. If the total length in unknown, which is represented by a {@code *} symbol
```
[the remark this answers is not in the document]
L-Applin: It is part of the **[reference spec](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Range)**. I'm not sure if S3 uses it.

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/progress/TransferProgressUpdater.java
```diff
@@ -136,41 +136,31 @@ public void subscriberOnComplete() {
...
-
-                @Override
-                public void subscriberOnNext(ByteBuffer byteBuffer) {
-                    incrementBytesTransferred(byteBuffer.limit());
-                }
-
-                @Override
-                public void subscriberOnError(Throwable t) {
-                    transferFailed(t);
+                    ContentRangeParser contentRangeParser = new ContentRangeParser();
```
zoewangg, line 146: Is there any reason it's not a util class?

L-Applin: No specific reasons beside trying to limit the number of "utils" classes. Since this class has one specific and well bounded use case, I didn't see the need of a util class. But I don't have a string feeling about it.

zoewangg: I'm normally not a pro-util person, but this seems like a classic use of util class (parsing and string handling)

zoewangg: Plus, micro optimization to avoid object creation/GC

zoewangg (approved):

PR #5117, opened 2024-04-18
Fix for threads hanging with forwardTransformedResultTo

L-Applin:
When using `CompletableFutureUtils.forwardTransformedResultTo` if the provided function throws an exception, it would not be catch which can prevent the destination future to be completed and hang on a `.join()` call.

## Motivation and Context
Discovered during perf testing multipart s3 async client, as one of the test returned a null value.

## Modifications
Surround the call to `Function.apply()` with a try-catch and complete the destination future with the caught exception.

## Testing
Added unit test to CompletableUtils and TransferManager

### DIFF
--- a/utils/src/main/java/software/amazon/awssdk/utils/CompletableFutureUtils.java
+++ b/utils/src/main/java/software/amazon/awssdk/utils/CompletableFutureUtils.java
@@ -167,7 +167,7 @@ public final class CompletableFutureUtils {
             DestT result = null;
             try {
                 result = function.apply(r);
-            } catch (Exception functionException) {
+            } catch (Throwable functionException) {
                 dst.completeExceptionally(functionException);
                 return;
             }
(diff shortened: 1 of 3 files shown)


utils/src/main/java/software/amazon/awssdk/utils/CompletableFutureUtils.java
```diff
@@ -161,9 +162,16 @@ public static <SourceT, DestT> CompletableFuture<SourceT> forwardTransformedResu
...
             if (e != null) {
                 dst.completeExceptionally(e);
-            } else {
-                dst.complete(function.apply(r));
+                return;
+            }
+            DestT result = null;
+            try {
+                result = function.apply(r);
+            } catch (Exception functionException) {
```
zoewangg, line 170: Should we catch Throwable?

zoewangg (approved):

PR #5123, opened 2024-04-19
Add ADAPTIVE_V2 retry mode to support the legacy behavior

sugmanue:
## Motivation and Context

This change is aimed to prevent breaking existing users of the legacy adaptive mode while introducing the means needed for new users to onboard and use the correct adaptive implementation.

## Modifications

This change introduces a new `RetryMode` constant, `ADAPTIVE_V2` and uses the legacy `ADAPTIVE` retry policy by adapting it to a retry strategy using the existing adapter when the `ADAPTIVE` mode is requested, either by using `AwsRetryStrategy.forRetryMode(RetryMode.ADAPTIVE)` call, or by setting the `retry_mode=adaptive` value in the profile file or by using the `aws.retryMode=ADAPTIVE` system setting.

Customers that need or want the *correct* adaptive mode will need to use instead `RetryMode.ADAPTIVE_V2`. This will reduce the risk of breaking customers that are using the legacy mode.

## Testing

Added unit tests to validate that all the current ways of configuring the retry logic in the SDK will create the expected and correct instance that implements the desired logic.

## Types of changes

## Checklist

## License

### DIFF
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/retry/RetryMode.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/retry/RetryMode.java
@@ -85,9 +85,9 @@ public enum RetryMode {
      *
      * @see RetryPolicy#isFastFailRateLimiting()
      * @deprecated As of 2.25.xx, replaced by {@link #ADAPTIVE_V2}. The ADAPTIVE implementation has a bug that prevents it
-     * from remembering its state across requests which is needed to correctly estimate its sending rate. Given that the
+     * from remembering its state across requests which is needed to correctly estimate its sending rate. Given that
      * this bug has been present since its introduction and that correct version might change the traffic patterns of the SDK we
-     * deemed too risky to fix this implementation..
+     * deemed too risky to fix this implementation.
      */
     @Deprecated
     ADAPTIVE,
@@ -95,13 +95,12 @@ public enum RetryMode {
     /**
      * Adaptive V2 retry mode builds on {@link #STANDARD} mode.
      * <p>
-     * Adaptive retry mode dynamically limits the rate of AWS requests to maximize success rate. This may be at the
+     * Adaptive retry mode qdynamically limits the rate of AWS requests to maximize success rate. This may be at the
      * expense of request latency. Adaptive V2 retry mode is not recommended when predictable latency is important.
      * <p>
-     * Adaptive V2 mode differs from {@link #ADAPTIVE} mode in the computed delays between calls, including the fist attempt
-     * that might be delayed if the algorithm considers that it's needed to increase the odds of a successful response. The
-     * previous Adaptive mode is also designed to work like so but a bug present since its introduction prevents it from
-     * preserving its internal state across requests.
+     * {@code ADAPTIVE_V2} mode differs from {@link #ADAPTIVE} mode in the computed delays between calls, including the first 
+     * attempt
+     * that might be delayed if the algorithm considers that it's needed to increase the odds of a successful response.
      * <p>
      * <b>Warning:</b> Adaptive V2 retry mode assumes that the client is working against a single resource (e.g. one
      * DynamoDB Table or one S3 Bucket). If you use a single client for multiple resources, throttling or outages
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/retry/RetryMode.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/retry/RetryMode.java
@@ -178,6 +178,8 @@ public enum RetryMode {
                     return Optional.of(STANDARD);
                 case "adaptive":
                     return Optional.of(ADAPTIVE);
+                case "adaptive2":
+                    return Optional.of(ADAPTIVE2);
                 default:
                     throw new IllegalStateException("Unsupported retry policy mode configured: " + string);
             }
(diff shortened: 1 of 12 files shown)


Crissttianreyes:
> ## [Quality Gate Passed](https://sonarcloud.io/dashboard?id=aws_aws-sdk-java-v2&pullRequest=5123) **Quality Gate passed**  
> Issues  
>  [61 New issues](https://sonarcloud.io/project/issues?id=aws_aws-sdk-java-v2&pullRequest=5123&resolved=false&inNewCodePeriod=true)  
>  [0 Accepted issues](https://sonarcloud.io/component_measures?id=aws_aws-sdk-java-v2&pullRequest=5123&metric=new_accepted_issues&view=list)
> 
> Measures  
>  [0 Security Hotspots](https://sonarcloud.io/project/security_hotspots?id=aws_aws-sdk-java-v2&pullRequest=5123&resolved=false&inNewCodePeriod=true)  
>  [87.2% Coverage on New Code](https://sonarcloud.io/component_measures?id=aws_aws-sdk-java-v2&pullRequest=5123&metric=new_coverage&view=list)  
>  [1.4% Duplication on New Code](https://sonarcloud.io/component_measures?id=aws_aws-sdk-java-v2&pullRequest=5123&metric=new_duplicated_lines_density&view=list)  
>   
> [See analysis details on SonarCloud](https://sonarcloud.io/dashboard?id=aws_aws-sdk-java-v2&pullRequest=5123)
> 
>

core/sdk-core/src/main/java/software/amazon/awssdk/core/retry/RetryMode.java
```diff
@@ -84,9 +84,32 @@ public enum RetryMode {
      * the same client. When using adaptive retry mode, we recommend using a single client per resource.
      *
      * @see RetryPolicy#isFastFailRateLimiting()
+     * @deprecated As of 2.25.xx, replaced by {@link #ADAPTIVE_V2}. The ADAPTIVE implementation has a bug that prevents it
+     * from remembering its state across requests which is needed to correctly estimate its sending rate. Given that the
+     * this bug has been present since its introduction and that correct version might change the traffic patterns of the SDK we
```
dagnir, line 89: "Given that ~the~"

core/sdk-core/src/main/java/software/amazon/awssdk/core/retry/RetryMode.java
```diff
@@ -84,9 +84,32 @@ public enum RetryMode {
...
+     * deemed too risky to fix this implementation..
```
dagnir, line 90: repeated period

core/sdk-core/src/main/java/software/amazon/awssdk/core/retry/RetryMode.java
```diff
@@ -84,9 +84,32 @@ public enum RetryMode {
...
+     * Adaptive V2 retry mode builds on {@link #STANDARD} mode.
+     * <p>
+     * Adaptive retry mode dynamically limits the rate of AWS requests to maximize success rate. This may be at the
+     * expense of request latency. Adaptive V2 retry mode is not recommended when predictable latency is important.
+     * <p>
+     * Adaptive V2 mode differs from {@link #ADAPTIVE} mode in the computed delays between calls, including the fist attempt
+     * that might be delayed if the algorithm considers that it's needed to increase the odds of a successful response. The
+     * previous Adaptive mode is also designed to work like so but a bug present since its introduction prevents it from
+     * preserving its internal state across requests.
+     * <p>
```
dagnir, line 105: s/fist/first/

Also should "Adaptive" be capitalized?

```
The
     * previous Adaptive mode is also designed to work like so but a bug present since its introduction prevents it from
     * preserving its internal state across requests.
```

I don't think it's necessary to specify that `ADAPTIVE` is also supposed to work this. If we link back to `ADAPTIVE` from here I think that's sufficient

core/sdk-core/src/main/java/software/amazon/awssdk/core/retry/RetryPolicy.java
```diff
@@ -113,6 +113,10 @@ public static Builder builder(RetryMode retryMode) {
         Validate.paramNotNull(retryMode, "The retry mode cannot be set as null. If you don't want to set the retry mode,"
                                     + " please use the other builder method without setting retry mode, and the default retry"
                                     + " mode will be used.");
+        if (retryMode == RetryMode.ADAPTIVE_V2) {
+            throw new UnsupportedOperationException("ADAPTIVE_V2 is not supported by retry policies, use a RetryStrategy "
+                                                    + "instead");
+        }
```
dagnir, line 119: Can we move this check to `Builder` instead?

sugmanue: Yes, we can, but any reason for that? I don't mind either way but seems to me that we don't validate anything for builders as those can be in a non-consistent state up until they are built. Let me know.

sugmanue: I guess you can argue that it will create a better stack trace that points  to where the problem is instead of on a separated place, so makes sense. I will move it there.

dagnir (approved):

PR #5132, opened 2024-04-22
Fixes bugs in mapping of identity provider names for user agent

cenedhryn:
## Motivation and Context
Fixes a couple of bugs that lead to incorrect recording of user agent strings.


joviegas:
- How did we find these bugs? Could you please describe the bug?
- Do we require a changelog update for this one?

cenedhryn:
> * How did we find these bugs? Could you please describe the user-facing symptoms?

I discovered them when adding the same code to v1. There are no user facing symptoms, unless they are scrutinizing the user agent string in the header 

> * Do we require a changelog update for this one?

No. This is strictly internal code and would only affect values in the metrics database.

joviegas (approved):

PR #5137, opened 2024-04-23
Support Customization string array endpoint params for S3 access grants, until all SDKs add support for string array.

joviegas:
Support Customization string array endpoint params for S3 access grants, until all SDKs add support for string array.

## Motivation and Context

- S3 Access will not do model changes until StringArray is supported for all the SDKs.
- Thus we need to add Customization for S3

## Modifications

- Customization Config classes .

## Testing

- Added Junits

## Types of changes

## License

### DIFF
--- a/codegen/src/main/java/software/amazon/awssdk/codegen/model/config/customization/CustomizationConfig.java
+++ b/codegen/src/main/java/software/amazon/awssdk/codegen/model/config/customization/CustomizationConfig.java
@@ -20,6 +20,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import software.amazon.awssdk.codegen.model.rules.endpoints.ParameterModel;
 import software.amazon.awssdk.codegen.model.service.ClientContextParam;
 import software.amazon.awssdk.core.retry.RetryMode;
 import software.amazon.awssdk.core.traits.PayloadTrait;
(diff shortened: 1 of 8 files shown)


codegen/src/test/resources/software/amazon/awssdk/codegen/poet/client/c2j/query/customization-with-duplicate-endpointparameter.config
```diff
@@ -0,0 +1,9 @@
+{
+  "endpointParameters": {
+        "endpointId": {
+            "required": false,
+            "documentation": "Test to make sure exceptiion is thrown when there is already a paramater defined in model.",
```
cenedhryn, line 5: Nit: `exceptiion` an extra i

zoewangg (approved):

codegen/src/main/java/software/amazon/awssdk/codegen/model/config/customization/CustomizationConfig.java
```diff
@@ -843,4 +850,12 @@ public void setUseMultiAuth(boolean useMultiAuth) {
     public boolean useMultiAuth() {
         return useMultiAuth;
     }
+
+    public Map<String, ParameterModel> getEndpointParameters() {
```
cenedhryn, line 854: I'm not sure if this definition is enough as a criteria for triggering codegen of the JMESPath runtime helper. Endpoint parameters could in the future be present in the customization config without the need for that, since it's only operation context params with nested types that needs it. The ParameterModel does not currently contain information that lets you decide that either, but maybe when adding operation context params that can fit into ParameterModel.

PR #5155, opened 2024-04-25
Extracting existing JMESPath runtime to a separate class

cenedhryn:
that can be used by waiters and other components

## Motivation and Context
The new operation context params for auth and endpoint params require the usage of a subset of JMESPath in order to specify a value or set of values in the request. 

We already have a JMESPath parser at the codegen level, and a Java class that represents a JMESPath expression in the `WaitersRuntime.java` class that's copied to services during codegen. This PR extracts that into a separate runtime.

## Modifications
- Creates a new static "runtime" class that only contains generic JMESPath logic, and removes that from `WaitersRuntime.java`
- Changes waiters codegen to call the new runtime for acceptor logic
- Generates the new runtime as part of waiter codegen, or if `customization.config` contains an eligible endpoint parameter (part of this feature in previous PR)

## Considerations
- Originally intended to not copy JMESPath runtime functionality that isn't used by endpoint param config, but it's only a few hundred lines different. 
- We don't need to provide the ability to only call a scoped subset of JMESPath functionality under certain circumstances at this point. This should be gated at the model level.


codegen/src/main/java/software/amazon/awssdk/codegen/AddMetadata.java
```diff
@@ -104,7 +104,8 @@ private static void configurePackageName(Metadata metadata,
                 .withPaginatorsPackageName(namingStrategy.getPaginatorsPackageName(service))
                 .withWaitersPackageName(namingStrategy.getWaitersPackageName(service))
                 .withEndpointRulesPackageName(namingStrategy.getEndpointRulesPackageName(service))
-                .withAuthSchemePackageName(namingStrategy.getAuthSchemePackageName(service));
+                .withAuthSchemePackageName(namingStrategy.getAuthSchemePackageName(service))
+                .withJPathPackageName(namingStrategy.getJPathPackageName(service));
```
zoewangg, line 108: Should we use the full name `JmesPath` to be more clear?

codegen/src/main/java/software/amazon/awssdk/codegen/emitters/tasks/JPathRuntimeGeneratorTask.java
```diff
@@ -0,0 +1,65 @@
...
+import java.io.UncheckedIOException;
+import java.util.Collections;
+import java.util.List;
+import software.amazon.awssdk.codegen.emitters.GeneratorTask;
+import software.amazon.awssdk.codegen.emitters.GeneratorTaskParams;
+import software.amazon.awssdk.codegen.emitters.SimpleGeneratorTask;
+import software.amazon.awssdk.utils.IoUtils;
+
+public final class JPathRuntimeGeneratorTask extends BaseGeneratorTasks {
+    public static final String RUNTIME_CLASS_NAME = "JPathRuntime";
```
zoewangg, line 29: Same here, JmesPathRuntime?

codegen/src/main/resources/software/amazon/awssdk/codegen/jpath/JPathRuntime.java.resource
```diff
@@ -0,0 +1,450 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+import java.util.function.Function;
+import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.core.SdkPojo;
+import software.amazon.awssdk.utils.ToString;
+
+@SdkInternalApi
+public final class JPathRuntime {
```
zoewangg, line 14: Are there any changes between JmesPathRuntime and WaitersRuntime? If so, could you show me the diff?

codegen/src/main/java/software/amazon/awssdk/codegen/emitters/tasks/EndpointProviderTasks.java
```diff
@@ -57,6 +59,9 @@ protected List<GeneratorTask> createTasks() throws Exception {
             tasks.add(generateDefaultProvider());
             tasks.add(new RulesEngineRuntimeGeneratorTask(generatorTaskParams));
         }
+        if (shouldGenerateJPathRuntime()) {
```
joviegas, line 62: We must ensure that the new JMESPathRuntime remains independent of any configurations. During the binding of OperationContextParams, code generation occurs only when OperationContextParams are available.

The new JMESPathRuntime must be designed to be generic, allowing it to handle any JMESPath parsing without reliance on specific features, similar to the flexibility of [JsonNodeParser](https://github.com/aws/aws-sdk-java-v2/blob/master/core/json-utils/src/main/java/software/amazon/awssdk/protocols/jsoncore/JsonNodeParser.java)

joviegas:
```
    private static void setOperationContextParams(S3EndpointParams.Builder params, DeleteObjectsRequest request) {
        JMESPathRuntime.Value input = new JMESPathRuntime.Value(request);
        List<String> resultValues = input.field("Delete").field("Objects").flatten().field("Key").values();
        params.deleteObjectKeys(resultValues);
    }
```

joviegas, line 62: What are the demerits of always loading the JMESPathRuntime class? Additionally, for waiters, what if we create a class that extends JMESPathRuntime and adds methods only specific to waiters?"

codegen/src/main/java/software/amazon/awssdk/codegen/AddMetadata.java
cenedhryn: I explicitly avoided JmesPath since there has been a lot of discussion around minified versions, not calling that subset JmesPath etc. However, after working with the implementation in Java SDK that distinction has faded - I think the subset will be handled at a model level instead. So if consensus is JmesPath I'm fine with that.

codegen/src/main/java/software/amazon/awssdk/codegen/emitters/tasks/EndpointProviderTasks.java
cenedhryn: The demerits of always loading JMESPathRuntime is that there are almost 400 services but only 80 of them use waiters. 

I don't think the approach of adding methods that are only used in waiters is worth it, because it's just a few hundred lines. It's better to keep the language intact and it will be hard to keep track of different subsets used for different purposes.

joviegas: Thanks

codegen/src/main/resources/software/amazon/awssdk/codegen/jpath/JPathRuntime.java.resource
cenedhryn: No, there are no changes in this PR

codegen/src/main/java/software/amazon/awssdk/codegen/AddMetadata.java
cenedhryn: changed

zoewangg (approved):

PR #5167, opened 2024-04-30
Add AddCommentToMethod recipe

zoewangg:
## Motivation and Context
There are some methods that are removed in Java SDK v2 and for those changes, we'll not update the code and instead, just add a comment providing suggestions.

## Modifications
Add AddCommentToMethod recipe

## Testing
Unit tests added

## Types of changes

## Checklist

## License

### DIFF
--- a/migration-tool/src/main/java/software/amazon/awssdk/migration/internal/recipe/AddCommentToMethod.java
+++ b/migration-tool/src/main/java/software/amazon/awssdk/migration/internal/recipe/AddCommentToMethod.java
@@ -1,0 +1,90 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.migration.internal.recipe;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import org.openrewrite.ExecutionContext;
+import org.openrewrite.Option;
+import org.openrewrite.Recipe;
+import org.openrewrite.TreeVisitor;
+import org.openrewrite.java.JavaIsoVisitor;
+import org.openrewrite.java.MethodMatcher;
+import org.openrewrite.java.tree.J;
+import org.openrewrite.marker.SearchResult;
+import software.amazon.awssdk.annotations.SdkInternalApi;
+
+/**
+ * Add a comment to a method
+ */
+@SdkInternalApi
+public class AddCommentToMethod extends Recipe {
+
+    @Option(displayName = "Method pattern",
+        description = "A method pattern that is used to find matching method invocations.",
+        example = "org.mockito.Matchers anyVararg()")
+    private final String methodPattern;
+
+    @Option(displayName = "Comment",
+        description = "A comment to add to this method.",
+        example = "This method is not supported in AWS SDK for Java v2.")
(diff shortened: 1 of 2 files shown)


migration-tool/src/main/java/software/amazon/awssdk/migration/internal/recipe/AddCommentToMethod.java
```diff
@@ -0,0 +1,88 @@
...
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return new Visitor(methodPattern, comment);
+    }
+
+    private static final class Visitor extends JavaIsoVisitor<ExecutionContext> {
+        private final MethodMatcher methodMatcher;
+        private final String comment;
+
+        Visitor(String methodPattern, String comment) {
+            this.methodMatcher = new MethodMatcher(methodPattern, false);
```
davidh44, line 73: Are there any cases we would want the matchOverrides flag to be true?

migration-tool/src/main/java/software/amazon/awssdk/migration/internal/recipe/AddCommentToMethod.java
```diff
@@ -0,0 +1,88 @@
...
+    @Option(displayName = "Comment",
+        description = "A comment to add to this method.",
+        example = "This method is not supported in AWS SDK for Java v2.")
+    private final String comment;
+
+    @JsonCreator
+    public AddCommentToMethod(@JsonProperty("methodPattern") String methodPattern,
+                              @JsonProperty("comment") String comment) {
+        this.methodPattern = methodPattern;
+        this.comment = comment;
```
davidh44, line 50: Should we begin each comment making it clear it is from the migration tool? e.g., have a constant String to distinguish when there are existing comments

migration-tool/src/test/java/software/amazon/awssdk/migration/recipe/AddCommentToMethodTest.java
```diff
@@ -0,0 +1,90 @@
...
+                + "    }\n"
+                + "}\n",
+                "import com.amazonaws.ClientConfiguration;\n"
+                + "\n"
+                + "public class Example {\n"
+                + "    \n"
+                + "    void test() {\n"
+                + "        ClientConfiguration clientConfiguration = new ClientConfiguration();\n"
+                + "        // Existing comment \n"
+                + "        /*existing comment*/ /*~~(a comment)~~>*/clientConfiguration.setCacheResponseMetadata(false);\n"
```
davidh44, line 84: Does the `/* comment */` get rendered on its own line? Don't see a `\n` as above with the `// comment`

zoewangg: It's rendered on the same line of the code. I followed the same style that openrewrite used for their recipes

migration-tool/src/main/java/software/amazon/awssdk/migration/internal/recipe/AddCommentToMethod.java
zoewangg: Yeah, good idea!

migration-tool/src/main/java/software/amazon/awssdk/migration/internal/recipe/AddCommentToMethod.java
zoewangg: Probably not, I can't think of a scenario where we want to match overrides for our use-case.

migration-tool/src/test/java/software/amazon/awssdk/migration/recipe/AddCommentToMethodTest.java
davidh44: Got it, so this is just for test purposes. I assume we'd want the added comment on a separate line than the existing comment?

davidh44 (approved):

migration-tool/src/test/java/software/amazon/awssdk/migration/recipe/AddCommentToMethodTest.java
zoewangg: Yeah, ideally, I tried to do that but it caused some other formatting issues; it's surprising hard to get the indent and formatting right after we add a new comment, so I just used one of openrewrite classes directly to generate the comment, which happens to be on the same line.

PR #5169, opened 2024-04-30
Converts endpoint param list of string to list of value

cenedhryn:
## Motivation and Context
The type of the endpoint param, `List<String>`, and the type required by the rules engine, `List<Value>`, are incompatible. Codegen Generated Classes Test cannot compile. 

## Modifications
- Changed the parameter assignment in the rules engine to a list of Value by streaming the endpoint parameter contents and converting each `String` using `Value.fromStr`
- Also added `stringarray` to `ParameterType` - required for Codegen Generated Classes Test tests to compile

### DIFF
--- a/codegen/src/main/java/software/amazon/awssdk/codegen/poet/rules/EndpointRulesSpecUtils.java
+++ b/codegen/src/main/java/software/amazon/awssdk/codegen/poet/rules/EndpointRulesSpecUtils.java
@@ -147,7 +147,7 @@ public class EndpointRulesSpecUtils {
                 break;
             case "stringarray":
                 methodName = "fromArray";
-                param = CodeBlock.of("$L.stream().map($T::fromStr).collect($T.toList())",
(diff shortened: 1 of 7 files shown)


codegen/src/main/java/software/amazon/awssdk/codegen/poet/rules/EndpointRulesSpecUtils.java
```diff
@@ -147,6 +147,10 @@ public CodeBlock valueCreationCode(String type, CodeBlock param) {
                 break;
             case "stringarray":
                 methodName = "fromArray";
+                param = CodeBlock.of("$L.stream().map($T::fromStr).collect($T.toList())",
```
dagnir, line 150: minor: use `$N.stream()` instead of `$L`, since `$N` is used for identifiers

dagnir (approved):

PR #5172, opened 2024-05-01
Generates JmesPath expressions for operation context parameters

cenedhryn:
## Motivation and Context
Endpoint parameters of type `stringarray` are supported for operation level traits only, normally via the `OperationContextParam` trait, which uses a JmesPath expression to find and potentially transform a part of an `SdkPojo` in order to create a corresponding endpoint parameter

## Modifications
- Modifies endpoint resolver interceptor codegen to interpret the value of the OperationContextParam as a JmesPath expression string
- The result of this JmesPath expression will be evaluated at runtime, and stored in the endpoint param.
- Unlike waiters, the codegen JmesPath must contain the value of the expression. In order to know what type of value is required, we match the operationContextParameter with it's corresponding endpoint ParameterModel (same key) since the type for that is known.
- The JmesPathRuntime now has support for returning the value as any of the three supported types of endpoint parameter. 
- In addition, the JmesPath generator now supports `wildcard` and the `keys` function. This means that the codegen side supports it, but the capability is lacking at runtime (future PR). 

## Testing
`codegen-generated-classes-test` has a new test class that evaluates that the value of the OperationContextParams are set correctly. The test cases contain the minimal subset of JmesPath that is required for OperationContextParams and already exists in the codebase: 
- identifier
- subexpression


test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/jmespath/JmesPathRuntimeValueTest.java
```diff
@@ -0,0 +1,320 @@
...
+
+    @Test
+    void lengthBehaves() {
+        assertThat(new Value(null).length()).isEqualTo(new Value(null));
+        assertThat(new Value("a").length()).isEqualTo(new Value(1));
+        assertThat(sdkPojoValue(Pair.of("a", "b")).length()).isEqualTo(new Value(1));
+        assertThat(new Value(singletonList("a")).length()).isEqualTo(new Value(1));
+    }
+
+    // @Test
```
cenedhryn, line 221: I just saw this - will need it for next PR

test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/jmespath/JmesPathRuntimeValueTest.java
```diff
@@ -0,0 +1,320 @@
...
+import java.util.stream.Stream;
+import org.junit.jupiter.api.Test;
+import software.amazon.awssdk.core.SdkField;
+import software.amazon.awssdk.core.SdkPojo;
+import software.amazon.awssdk.core.protocol.MarshallingType;
+import software.amazon.awssdk.core.traits.LocationTrait;
+import software.amazon.awssdk.services.restjsonwithwaiters.jmespath.internal.JmesPathRuntime.Value;
+import software.amazon.awssdk.utils.Pair;
+
+class JmesPathRuntimeValueTest {
```
cenedhryn, line 37: Probably should not have renamed this class, but here we are. There is nothing removed.

codegen/src/main/java/software/amazon/awssdk/codegen/poet/rules/EndpointResolverInterceptorSpec.java
```diff
@@ -550,24 +554,52 @@ private MethodSpec setOperationContextParamsMethod(OperationModel opModel) {
...
+    private CodeBlock convertValueToParameterType(ParameterModel parameterModel) {
+        switch (parameterModel.getType().toLowerCase(Locale.US)) {
+            case "boolean":
+                return CodeBlock.of(".booleanValue()");
+            case "string":
+                return CodeBlock.of(".stringValue()");
+            case "stringarray":
+                return CodeBlock.of(".stringValues()");
+            default:
+                return CodeBlock.of("");
```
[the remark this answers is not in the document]
cenedhryn: Yes, I think that's better. Good suggestion.

zoewangg (approved):

PR #5173, opened 2024-05-01
Add S3StreamingResponseToV2 Recipe

dagnir:
## Motivation and Context

This recipe transforms usages of `s3.getObject(..)` such as

```
S3Object myObject = s3.getObject(request);
```

to the equivalent in v2:

```
ResponseInputStream<GetObjectResponse> myObject = s3.getOBject(request);
```

In addition, because of the inversion of streaming and non-streaming members when moving to ResponseInputStream, it further transforms usages of myObject so that calls in v1 that access the stream:

```
myObject.getObjectContent().close()
```

become

```
myObject.close()
```

Likewise, methods that access the non-streaming members

```
myObject.getKey();
```

become

```
myObject.response().getKey();
```

## Types of changes

## Checklist

## License

### DIFF
--- a/migration-tool/src/main/java/software/amazon/awssdk/migration/internal/recipe/S3StreamingResponseToV2.java
+++ b/migration-tool/src/main/java/software/amazon/awssdk/migration/internal/recipe/S3StreamingResponseToV2.java
@@ -1,0 +1,182 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.migration.internal.recipe;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.stream.Collectors;
+import org.openrewrite.ExecutionContext;
+import org.openrewrite.Recipe;
+import org.openrewrite.Tree;
+import org.openrewrite.TreeVisitor;
+import org.openrewrite.java.JavaVisitor;
+import org.openrewrite.java.MethodMatcher;
+import org.openrewrite.java.tree.Expression;
+import org.openrewrite.java.tree.J;
+import org.openrewrite.java.tree.JContainer;
+import org.openrewrite.java.tree.JRightPadded;
+import org.openrewrite.java.tree.JavaType;
+import org.openrewrite.java.tree.Space;
+import org.openrewrite.java.tree.TypeUtils;
+import org.openrewrite.marker.Markers;
+import software.amazon.awssdk.annotations.SdkInternalApi;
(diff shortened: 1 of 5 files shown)


migration-tool/src/main/java/software/amazon/awssdk/migration/internal/recipe/S3StreamingResponseToV2.java
```diff
@@ -0,0 +1,182 @@
...
+        // ResponseInputStream<GetObjectResponse> myObject = ...
+        @Override
+        public J.VariableDeclarations visitVariableDeclarations(J.VariableDeclarations multiVariable,
+                                                                ExecutionContext executionContext) {
+            if (TypeUtils.isAssignableTo(S3_OBJECT_TYPE, multiVariable.getType())) {
+                JavaType.Parameterized newType =
+                    new JavaType.Parameterized(null, RESPONSE_INPUT_STREAM_TYPE,
+                                               Collections.singletonList(S3_GET_OBJECT_RESPONSE_TYPE));
+
+                maybeAddS3ResponseImports();
```
anirudh9391, line 82: nit : Do we need the maybe ? I thought we could just do AddS3ResponseImports and possibly specify a comment that explains under what conditions this happens.

migration-tool/src/main/java/software/amazon/awssdk/migration/internal/recipe/S3StreamingResponseToV2.java
```diff
@@ -0,0 +1,182 @@
...
+                multiVariable = multiVariable.withVariables(variables);
+            }
+
+            multiVariable = super.visitVariableDeclarations(multiVariable, executionContext).cast();
+
+            return multiVariable;
+        }
+
+        private void maybeAddS3ResponseImports() {
+            maybeAddImport(RESPONSE_INPUT_STREAM_TYPE);
```
anirudh9391, line 101: Same as above

migration-tool/src/main/java/software/amazon/awssdk/migration/internal/utils/IdentifierUtils.java
```diff
@@ -0,0 +1,73 @@
...
+        );
+    }
+
+    /**
+     * Simple method for creating a simple name for a {@link JavaType.Parameterized} instance. This is a naive implementation
+     * that currently requires all type parameters to be {@link JavaType.FullyQualified}, and does not handle nested
+     * parameterized types.
+     */
+    public static String simpleName(JavaType.Parameterized parameterized) {
+        StringBuffer sb = new StringBuffer(parameterized.getClassName())
```
anirudh9391, line 54: just curious, is there a reason as to why we chose this ahead of StringBuilder ? Does this have to be thread safe ?

migration-tool/src/main/java/software/amazon/awssdk/migration/internal/recipe/S3StreamingResponseToV2.java
dagnir: I prefer to keep the `maybe` since these visitor methods can be called any number of times and it clearly signifies that we only conditionally add them (i.e. if they aren't already present).

migration-tool/src/main/java/software/amazon/awssdk/migration/internal/utils/IdentifierUtils.java
dagnir: Ooops! Good catch, I wasn't paying attention. Yes `StringBuilder` is what I wanted!

anirudh9391 (approved):

PR #5176, opened 2024-05-02
Fix for multipart integ test failure

L-Applin:
The `S3TransferManagerDownloadPauseResumeIntegrationTest` was failing with S3 Multipart client for 2 reasons.
1. The downloads were not immediately stopped when calling `.pause()`
2. When resuming, the range get would have the whole object size as its content-length, making the TransferProgressUpdater report the wrong number of bytes.

This PR is to fix both of those issues.

## Motivation and Context
Errors were raised during integration test execution.

## Modifications
Fix for the errors consist in 
- Forward cancellation from returnFuture in SplittingTransformer
- Use content-length for range-get in TransferProgressUpdater, when multipart is enabled. 

## Testing
Failing integration test run locally.
Added unit test.

### DIFF
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/SplittingTransformer.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/SplittingTransformer.java
@@ -16,7 +16,6 @@
 package software.amazon.awssdk.core.internal.async;
 
 import java.nio.ByteBuffer;
-import java.util.concurrent.CancellationException;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/SplittingTransformer.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/SplittingTransformer.java
@@ -172,7 +172,7 @@ public class SplittingTransformer<ResponseT, ResultT> implements SdkPublisher<As
         public void cancel() {
             if (isCancelled.compareAndSet(false, true)) {
                 log.trace(() -> "Cancelling splitting transformer");
-                handleCancelState();
+                handleSubscriptionCancel();
             }
         }
     }
@@ -207,7 +207,13 @@ public class SplittingTransformer<ResponseT, ResultT> implements SdkPublisher<As
         return false;
     }
 
-    private void handleCancelState() {
+    /**
+     * Handle the {@code .cancel()} signal received from the downstream subscription. Data that is being sent to the upstream
+     * transformer need to finish processing before we complete. One typical use case for this is completing the multipart
+     * download, the subscriber having reached the final part will signal that it doesn't need more parts by calling {@code
+     * .cancel()} on the subscription.
+     */
+    private void handleSubscriptionCancel() {
         synchronized (cancelLock) {
             if (downstreamSubscriber == null) {
                 return;
@@ -231,6 +237,13 @@ public class SplittingTransformer<ResponseT, ResultT> implements SdkPublisher<As
         }
     }
 
+    /**
+     * Handle when the {@link SplittingTransformer#resultFuture} is cancelled or completed exceptionally from the outside. Data
+     * need to stop being sent to the upstream transformer immediately. One typical use case for this is transfer manager
+     * needing to pause download by calling {@code .cancel(true)} on the future.
+     *
+     * @param e The exception the future was complete exceptionally with.
+     */
     private void handleFutureCancel(Throwable e) {
         synchronized (cancelLock) {
             publisherToUpstream.error(e);
@@ -254,6 +267,9 @@ public class SplittingTransformer<ResponseT, ResultT> implements SdkPublisher<As
         public CompletableFuture<ResponseT> prepare() {
             this.individualFuture = new CompletableFuture<>();
             if (preparedCalled.compareAndSet(false, true)) {
+                if (isCancelled.get()) {
+                    return individualFuture;
+                }
                 CompletableFuture<ResultT> upstreamFuture = upstreamResponseTransformer.prepare();
                 if (!resultFuture.isDone()) {
                     CompletableFutureUtils.forwardResultTo(upstreamFuture, resultFuture);
@@ -263,12 +279,11 @@ public class SplittingTransformer<ResponseT, ResultT> implements SdkPublisher<As
                 if (e == null) {
                     return;
                 }
-                isCancelled.set(true);
-                individualFuture.cancel(true);
+                individualFuture.completeExceptionally(e);
             });
             individualFuture.whenComplete((r, e) -> {
                 if (isCancelled.get()) {
-                    handleCancelState();
+                    handleSubscriptionCancel();
                 }
             });
             return this.individualFuture;
@@ -288,14 +303,16 @@ public class SplittingTransformer<ResponseT, ResultT> implements SdkPublisher<As
             if (downstreamSubscriber == null) {
                 return;
             }
-            if (onStreamCalled.compareAndSet(false, true)) {
-                log.trace(() -> "calling onStream on the upstream transformer");
-                upstreamResponseTransformer.onStream(upstreamSubscriber -> publisherToUpstream.subscribe(
-                    DelegatingBufferingSubscriber.builder()
-                                                 .maximumBufferInBytes(maximumBufferInBytes)
-                                                 .delegate(upstreamSubscriber)
-                                                 .build()
-                ));
+            synchronized (cancelLock) {
+                if (onStreamCalled.compareAndSet(false, true)) {
+                    log.trace(() -> "calling onStream on the upstream transformer");
+                    upstreamResponseTransformer.onStream(upstreamSubscriber -> publisherToUpstream.subscribe(
+                        DelegatingBufferingSubscriber.builder()
+                                                     .maximumBufferInBytes(maximumBufferInBytes)
+                                                     .delegate(upstreamSubscriber)
+                                                     .build()
+                    ));
+                }
             }
             publisher.subscribe(new IndividualPartSubscriber<>(this.individualFuture, response));
         }
@@ -335,9 +352,6 @@ public class SplittingTransformer<ResponseT, ResultT> implements SdkPublisher<As
         public void onNext(ByteBuffer byteBuffer) {
             if (byteBuffer == null) {
                 throw new NullPointerException("onNext must not be called with null byteBuffer");
-            }
-            if (isCancelled.get()) {
-                return;
             }
             publisherToUpstream.send(byteBuffer).whenComplete((r, t) -> {
                 if (t != null) {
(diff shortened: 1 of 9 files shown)


core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/SplittingTransformer.java
```diff
@@ -219,6 +231,16 @@ private void handleCancelState() {
         }
     }
 
+    private void handleFutureCancel(Throwable e) {
```
dagnir, line 234: Pardon my ignorance, why do we have two different methods that seem like they should do the same thing? `handleCancelState` and `handleFutureCancel`

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/SplittingTransformer.java
```diff
@@ -196,7 +208,7 @@ private boolean doEmit() {
     }
 
     private void handleCancelState() {
-        synchronized (this) {
+        synchronized (cancelLock) {
```
dagnir, line 211: What if `!onStreanCalled().get()` is no longer true after we make the check?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/SplittingTransformer.java
```diff
@@ -307,12 +336,17 @@ public void onNext(ByteBuffer byteBuffer) {
...
+            if (isCancelled.get()) {
+                return;
+            }
             publisherToUpstream.send(byteBuffer).whenComplete((r, t) -> {
                 if (t != null) {
                     handleError(t);
                     return;
                 }
-                subscription.request(1);
+                if (!isCancelled.get()) {
```
dagnir, line 347: Similar question as above, is it possible this is no longer true after we check it? If so, is it an issue?

dagnir (commented):
Can we kick off an integ test on the PR as well?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/SplittingTransformer.java
L-Applin: The reason is because`handleCancelState` is for when the subscriber calls `cancel()` on the subscription. In that case we must wait until all data currently being processed by the upstream AsyncResponseTransformer is completed. `handleFutureCancel`  is for the case when the `returnFuture` is cancelled, in which case we need to stop sending data to the upstream AsyncResponseTransformer.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/SplittingTransformer.java
L-Applin: `onStreanCalled` will only go from `false -> true` once and never back to false.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/SplittingTransformer.java
L-Applin: same as above, `onStreanCalled` will only go from `false -> true` once and never back to false.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/SplittingTransformer.java
```diff
@@ -70,7 +70,8 @@ public class SplittingTransformer<ResponseT, ResultT> implements SdkPublisher<As
     private final AtomicBoolean onStreamCalled = new AtomicBoolean(false);
 
     /**
-     * Set to true once {@code .concel()} is called in the subscription of the downstream subscriber
+     * Set to true once {@code .concel()} is called in the subscription of the downstream subscriber, or if the
```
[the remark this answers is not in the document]
L-Applin: "concel" 🤣  good catch

L-Applin:
> Can we kick off an integ test on the PR as well?

Yep, already did

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/SplittingTransformer.java
zoewangg: Let's add some documentation for this since it's not straightforward.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/SplittingTransformer.java
```diff
@@ -237,6 +259,13 @@ public CompletableFuture<ResponseT> prepare() {
                     CompletableFutureUtils.forwardResultTo(upstreamFuture, resultFuture);
                 }
             }
+            resultFuture.whenComplete((r, e) -> {
+                if (e == null) {
+                    return;
+                }
+                isCancelled.set(true);
```
zoewangg, line 266: Should we handle cancellation logic on line 125 here?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/SplittingTransformer.java
```diff
@@ -307,12 +336,17 @@ public void onNext(ByteBuffer byteBuffer) {
             if (byteBuffer == null) {
                 throw new NullPointerException("onNext must not be called with null byteBuffer");
             }
+            if (isCancelled.get()) {
```
zoewangg, line 339: Is this needed? If publisherToUpstream.error is invoked, send will be no op

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/SplittingTransformer.java
zoewangg: Do we need the check here assuming subscription is cancelled properly or will be cancelled at this point? 
https://github.com/reactive-streams/reactive-streams-jvm 

> If a Subscription is cancelled its Subscriber MUST eventually stop being signaled.

Note "eventually" not "right away"

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/SplittingTransformer.java
dagnir: Okay that makes sense. Can `handleCancelState` be renamed to something more clear like `handleSubscriberCancel`?

zoewangg:
Rerunning integ tests after [fixing parameterized tests](https://github.com/aws/aws-sdk-java-v2/pull/5176/commits/86cf6e79a5d486e793eb8a345e29afb90a5d2574). They are passing on my machine now at least. 🤞🏼

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/SplittingTransformer.java
L-Applin: Aaah I see what you mean. Looking at it, yep, it might cause a NPE at [line 293](https://github.com/aws/aws-sdk-java-v2/pull/5176/files/059fa2a71a7a1e4d5bbb5eec6523bbc27fc1257b#diff-b9ec282c15a46057d63a0021282180d584a26dc1c229a66e8a95c3c706dfa87dR293). Adding synchronization around the `if (onStreamCalled.compareAndSet(false, true))` on `cancelLock` there at line 293 should prevent it.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/SplittingTransformer.java
L-Applin: we do need to stop right away in the case of the return future being cancelled, for example when pausing with transfer-manager

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/SplittingTransformer.java
L-Applin: actually, setting `isCancelled` to true is redundant with the `whenComplete` in the constructor, we can just propagate the error to `individualFuture` here and let the `whenComplete` in the constructor handle cancellation logic. Will remove this `isCancelled.set(true);`

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/SplittingTransformer.java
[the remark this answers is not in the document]
L-Applin: So that the `upstreamSubscriber` (coming from the `upstreamResponseTransformer`, e.g. the `FileAsyncResponseTransformer.FileSubscriber`) can receive the `onComplete` signal after data has been sent.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/SplittingTransformer.java
L-Applin: Did some testing around this, yep I think  you are right, this is redundant. I will remove it.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/SplittingTransformer.java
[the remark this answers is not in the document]
zoewangg: Ah right! Thanks for adding documentation. I'm sure I'll forget about this after two days.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/SplittingTransformer.java
zoewangg: What would happen if we don't check it here and continue to send request demand? It should be no-op right?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/SplittingTransformer.java
```diff
@@ -232,14 +267,23 @@ private class IndividualTransformer implements AsyncResponseTransformer<Response
         public CompletableFuture<ResponseT> prepare() {
             this.individualFuture = new CompletableFuture<>();
             if (preparedCalled.compareAndSet(false, true)) {
+                if (isCancelled.get()) {
```
zoewangg, line 270: Why do we need this logic here?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/SplittingTransformer.java
```diff
@@ -232,14 +267,23 @@ private class IndividualTransformer implements AsyncResponseTransformer<Response
...
+            resultFuture.whenComplete((r, e) -> {
+                if (e == null) {
+                    return;
+                }
+                individualFuture.completeExceptionally(e);
+            });
             individualFuture.whenComplete((r, e) -> {
                 if (isCancelled.get()) {
-                    handleCancelState();
+                    handleSubscriptionCancel();
```
zoewangg, line 286: Can you remind me why we need to check isCancelled here for every individual future?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/SplittingTransformer.java
```diff
@@ -118,6 +121,15 @@ private SplittingTransformer(AsyncResponseTransformer<ResponseT, ResultT> upstre
         Validate.notNull(maximumBufferSizeInBytes, "maximumBufferSizeInBytes");
         this.maximumBufferInBytes = Validate.isPositive(
             maximumBufferSizeInBytes, "maximumBufferSizeInBytes");
+
+        this.resultFuture.whenComplete((r, e) -> {
```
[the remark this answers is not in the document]
zoewangg: Thanks for adding a unit test for this. Can we add a end-to-end functional test (wiremock test) to verify the cancellation behavior?

PR #5198, opened 2024-05-07
Supporting wildcard and keys fn in JmesPathRuntime

cenedhryn:
## Modifications
- Modifies the `wildcard` function, which has existed in the runtime but never been used in waiters, to accept lists which is according to specs: https://jmespath.org/specification.html#wildcard-expressions
- Adds a `keys` function which can return the names of `SdkField` members in an `SdkPojo`
- [INCORRECT] Catches errors when evaluating OperationContextParams at runtime and returns null instead, logging errors. It's complex functionality especially when sourcing parameters from a customization config. We can re-evaluate this behavior when model traits are available. 
- Fixes a minor bug in `project` to filter null values from the result. Another alternative would be to just filter nulls when returning `stringValues` but I can't think of a Java use-case where we want to preserve the null when evaluating an expression. 
- Also changed to filter nulls for stringValues, because it is correct.

## Testing
Tested in `codegen-generated-classes-test` 

Testing with incomplete objects where the projected field does not exist in a list projection member didn't work with the approximation of and SdkPojo in the test class. Instead, switched to a modeled pojo that correctly returns null for a field instead of not recognizing field.

### DIFF
--- a/codegen/src/main/java/software/amazon/awssdk/codegen/poet/rules/EndpointResolverInterceptorSpec.java
+++ b/codegen/src/main/java/software/amazon/awssdk/codegen/poet/rules/EndpointResolverInterceptorSpec.java
@@ -87,10 +87,13 @@ import software.amazon.awssdk.metrics.MetricCollector;
 import software.amazon.awssdk.utils.AttributeMap;
 import software.amazon.awssdk.utils.CollectionUtils;
 import software.amazon.awssdk.utils.HostnameValidator;
+import software.amazon.awssdk.utils.Logger;
 import software.amazon.awssdk.utils.StringUtils;
 import software.amazon.awssdk.utils.internal.CodegenNamingUtils;
 
 public class EndpointResolverInterceptorSpec implements ClassSpec {
+
+    private static final String LOGGER_FIELD_NAME = "LOG";
     private final IntermediateModel model;
     private final EndpointRulesSpecUtils endpointRulesSpecUtils;
     private final EndpointParamsKnowledgeIndex endpointParamsKnowledgeIndex;
@@ -124,6 +127,8 @@ public class EndpointResolverInterceptorSpec implements ClassSpec {
                                       .addModifiers(Modifier.PUBLIC, Modifier.FINAL)
                                       .addAnnotation(SdkInternalApi.class)
                                       .addSuperinterface(ExecutionInterceptor.class);
+
+        b.addField(logger());
 
         if (!useSraAuth) {
             b.addField(endpointAuthSchemeStrategyFieldSpec);
@@ -559,6 +564,7 @@ public class EndpointResolverInterceptorSpec implements ClassSpec {
         opModel.getOperationContextParams().forEach((key, value) -> {
             if (Objects.requireNonNull(value.getValue().asToken()) == JsonToken.VALUE_STRING) {
                 String setterName = endpointRulesSpecUtils.paramMethodName(key);
+
                 String jmesPathString = ((JrsString) value.getValue()).getValue();
                 CodeBlock addParam = CodeBlock.builder()
                                               .add("params.$N(", setterName)
@@ -566,7 +572,14 @@ public class EndpointResolverInterceptorSpec implements ClassSpec {
                                               .add(matchToParameterType(key))
                                               .add(")")
                                               .build();
+
+                b.beginControlFlow("try");
                 b.addStatement(addParam);
+                b.endControlFlow();
+                b.beginControlFlow("catch ($T e)", Exception.class);
+                b.addStatement("$N.warn(() -> \"Error resolving operation context parameter '$L'; will set param to null."
+                               + " \\nError message: \" + e.getMessage())", LOGGER_FIELD_NAME, setterName);
+                b.endControlFlow();
             } else {
                 throw new RuntimeException("Invalid operation context parameter path for " + opModel.getOperationName() +
                                            ". Expected VALUE_STRING, but got " + value.getValue().asToken());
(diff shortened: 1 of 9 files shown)


codegen/src/main/java/software/amazon/awssdk/codegen/poet/rules/EndpointResolverInterceptorSpec.java
```diff
@@ -125,6 +128,8 @@ public TypeSpec poetSpec() {
                                       .addAnnotation(SdkInternalApi.class)
                                       .addSuperinterface(ExecutionInterceptor.class);
 
+        b.addField(logger());
```
cenedhryn, line 131: In general, wondering if we should add logic to only generate operation context params if they exist, potentially including the logger dependency. Do not have time to check exactly what is generated under which circumstances for this PR.

davidh44 (approved):

test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/endpointproviders/OperationContextParametersTest.java
```diff
@@ -56,9 +64,36 @@ public void setup() {
...
+        return Stream.of(
+            Arguments.of("Single String parameter works as expected", new TestAssertion(
+                p -> assertThat(p.pojoString()).isEqualTo("StringMemberA"))),
+
+            Arguments.of("List of String parameter works as expected", new TestAssertion(
+                p -> assertThat(p.basicListOfString()).isNotEmpty().hasSize(2).containsExactly("StringA", "StringB"))),
+
+            Arguments.of("List of String parameter from wildcard works as expected", new TestAssertion(
+                p -> assertThat(p.wildcardKeyListOfString()).isNotEmpty().hasSize(2)
+                                                            .containsExactly("StringMemberS1", "StringMemberS2"))),
```
joviegas, line 77: Can we write a test case when only one of the List has entry
 
 Source Json
 ```json
{
  "bucket": "someBucket",
  "delete": {
    "objects": [
      {
        "id": "a",
        "key": "One"
      },
      {
        "id": "b"
      }
    ]
  }
}

```
JMESPath Expression : `delete.objects[*].key`

Output
```json
[
  "One"
]
```

joviegas:
Can we please add test cases in Acceptor too.

https://github.com/aws/aws-sdk-java-v2/blob/e46761c8709b06e65394b3716e7caf99a026b414/codegen/src/test/java/software/amazon/awssdk/codegen/poet/waiters/JmesPathAcceptorGeneratorTest.java#L217-L271

Since we are supporting it now can we add corresponding test for above cases .

Or if we only support `[*]` then add test for `[*]` and keep the remaining assertion as it was before

cenedhryn:
> Can we please add test cases in Acceptor too.
> 
> https://github.com/aws/aws-sdk-java-v2/blob/e46761c8709b06e65394b3716e7caf99a026b414/codegen/src/test/java/software/amazon/awssdk/codegen/poet/waiters/JmesPathAcceptorGeneratorTest.java#L217-L271
> 
> Since we are supporting it now can we add corresponding test for above cases .
> 
> Or if we only support `[*]` then add test for `[*]` and keep the remaining assertion as it was before

There a few things I want to clean up in a follow-up PR - some renaming, excluding codegen of empty methods etc. I will add this to that list.

codegen/src/main/java/software/amazon/awssdk/codegen/poet/rules/EndpointResolverInterceptorSpec.java
```diff
@@ -559,14 +564,22 @@ private MethodSpec setOperationContextParamsMethod(OperationModel opModel) {
...
                 String setterName = endpointRulesSpecUtils.paramMethodName(key);
+
                 String jmesPathString = ((JrsString) value.getValue()).getValue();
                 CodeBlock addParam = CodeBlock.builder()
                                               .add("params.$N(", setterName)
                                               .add(jmesPathGenerator.interpret(jmesPathString, "input"))
                                               .add(matchToParameterType(key))
                                               .add(")")
                                               .build();
+
```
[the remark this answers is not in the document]
cenedhryn: Agreed

test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/endpointproviders/OperationContextParametersTest.java
```diff
@@ -67,23 +102,22 @@ void operationContextParams_customization_resolvedCorrectly(TestCase tc) {
...
+        List<LogEvent> jmesRuntimeEvents = loggedEvents().stream()
+                                                         .filter(e -> e.getLoggerName().contains(
+                                                             "RestJsonEndpointProvidersResolveEndpointInterceptor"))
+                                                         .collect(Collectors.toList());
+        assertThat(jmesRuntimeEvents).hasSize(2);
+        assertThat(jmesRuntimeEvents).extracting("level").containsOnly(Level.WARN);
+        assertThat(jmesRuntimeEvents).extracting("message.formattedMessage")
+                                     .usingElementComparator(stringContains())
+                                     .contains("emptyKeyListOfString", "missingFieldListOfString")
+                                     .contains("No such field: NonExistingMember");
```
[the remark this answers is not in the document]
cenedhryn: Removed

test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/endpointproviders/OperationContextParametersTest.java
cenedhryn: Yes, had to update the test harness to handle this case.

codegen/src/main/java/software/amazon/awssdk/codegen/poet/rules/EndpointResolverInterceptorSpec.java
cenedhryn: Added removing empty methods to follow up task

davidh44 (approved):

joviegas (approved):

PR #5200, opened 2024-05-08
Change package name for migration tool and update dependencies

zoewangg:
## Motivation and Context
- Change package name for migration tool `AwsJavaSdk-MigrationTool` -> `AwsJavaSdk-Tooling-MigrationTool` to not cause potential naming conflicts in the future.
- Update dependencies
- Exclude `migration-tool-tests` from javadoc and deployment

### DIFF
--- a/.brazil.json
+++ b/.brazil.json
@@ -43,7 +43,7 @@
 
         "dynamodb": { "packageName": "AwsJavaSdk-DynamoDb" },
         "waf": { "packageName": "AwsJavaSdk-Waf" },
-        "migration-tool": { "packageName":  "AwsJavaSdk-MigrationTool"},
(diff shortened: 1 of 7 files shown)


.brazil.json
```diff
@@ -102,7 +102,8 @@
         "test-utils": { "skipImport": true },
         "tests-coverage-reporting": { "skipImport": true },
         "third-party": { "skipImport": true },
-        "third-party-slf4j-api": { "skipImport": true }
+        "third-party-slf4j-api": { "skipImport": true },
+        "migration-tool-tests": {"migration-tool-tests":  true}
```
dagnir, line 106: is this supposed to be `"skipImport": true`?

zoewangg: Good catch!

dagnir (approved):

PR #5205, opened 2024-05-09
Make calls to endpoint discovery non-blocking for asynchornous clients

anirudh9391:
Make endpoint discovery calls fully asynchronous. 
## Motivation and Context
Conforms to the expectation that asynchronous requests made by non-blocking clients do not block the thread. 

## Modifications
Created a new internal API to be invoked by asynchronous clients. This new API returns a completableFuture as opposed to waiting on request object completion
## Testing
- Existing Unit tests for Endpoint Discovery pass
- Codegen tests pass

E2E test validation 
```
    @Test
    public void test() {
        SdkEventLoopGroup group = SdkEventLoopGroup.builder()
                                                   .numberOfThreads(1)
                                                   .build();

        SqsAsyncClient sqs = SqsAsyncClient.builder()
                                           .asyncConfiguration(o -> o.advancedOption(SdkAdvancedAsyncClientOption.FUTURE_COMPLETION_EXECUTOR, Runnable::run))
                                           .region(Region.US_WEST_2)
                                           .httpClient(NettyNioAsyncHttpClient.builder()
                                                                              .eventLoopGroup(group)
                                                                              .build())
                                           .build();

        TimestreamWriteAsyncClient timestream = TimestreamWriteAsyncClient.builder().region(Region.US_WEST_2)
                                                                          .httpClient(NettyNioAsyncHttpClient.builder()
                                                                                                             .eventLoopGroup(group).build())
                                                                          .build();

        CompletableFuture<ListQueuesResponse> future = sqs.listQueues();
        future.whenComplete((r, e) -> {
                                timestream.listDatabases(req -> req.maxResults(1));
                                assertThat(1).isEqualTo(1);
                            }
        ).join();

        assertThat(1).isEqualTo(1);
    }
```
## Types of changes


core/sdk-core/src/main/java/software/amazon/awssdk/core/endpointdiscovery/EndpointDiscoveryRefreshCache.java
```diff
@@ -87,6 +90,64 @@ public URI get(String accessKey, EndpointDiscoveryRequest request) {
...
+
+        if (request.cacheKey().isPresent()) {
+            key = key + ":" + request.cacheKey().get();
+        }
+
+        EndpointDiscoveryEndpoint endpoint = cache.get(key);
+
+        if (endpoint == null) {
+            if (request.required()) {
+                return discoverEndpoint(request).handle(
```
dagnir, line 109: This method is pretty long and unwieldy. Can we break this branch out to a separate method?

core/sdk-core/src/main/java/software/amazon/awssdk/core/endpointdiscovery/EndpointDiscoveryRefreshCache.java
dagnir, line 1: Can we add tests?

core/sdk-core/src/main/java/software/amazon/awssdk/core/endpointdiscovery/EndpointDiscoveryRefreshCache.java
```diff
@@ -87,6 +90,64 @@ public URI get(String accessKey, EndpointDiscoveryRequest request) {
...
+                        if (throwable != null) {
+                            if (throwable instanceof InterruptedException) {
+                                Thread.currentThread().interrupt();
+                                throw EndpointDiscoveryFailedException.create(throwable);
+                            }
+                            if (throwable instanceof ExecutionException
+                                    || throwable instanceof CompletionException) {
+                                throw EndpointDiscoveryFailedException.create(throwable.getCause());
+                            }
+                            throw new RuntimeException("new exception");
```
dagnir, line 120: Added by mistake? Or does this need to wrap the original error/cause?

core/sdk-core/src/main/java/software/amazon/awssdk/core/endpointdiscovery/EndpointDiscoveryRefreshCache.java
```diff
@@ -87,6 +90,64 @@ public URI get(String accessKey, EndpointDiscoveryRequest request) {
...
+        if (endpoint == null) {
+            if (request.required()) {
+                return discoverEndpoint(request).handle(
+                    (endpointDiscoveryEndpoint, throwable) -> {
+                        if (throwable != null) {
+                            if (throwable instanceof InterruptedException) {
+                                Thread.currentThread().interrupt();
+                                throw EndpointDiscoveryFailedException.create(throwable);
+                            }
+                            if (throwable instanceof ExecutionException
```
dagnir, line 116: It's not necessary to catch `InterruptedException`/or re-interrupt, especially since we have no idea what thread actually got interrupted; it may not be the same as `Thread.currentThread()`

Same for `ExecutionException` since we're not waiting on a future.get() anymore

dagnir (commented):
Can we include the test we discussed offline for testing a client using `Runnable::run` for `FUTURE_COMPLETION_EXECUTOR`?

core/sdk-core/src/main/java/software/amazon/awssdk/core/endpointdiscovery/EndpointDiscoveryRefreshCache.java
```diff
@@ -29,6 +30,8 @@ public final class EndpointDiscoveryRefreshCache {
 
     private final EndpointDiscoveryCacheLoader client;
 
+    private String key;
```
zoewangg, line 33: Is there any reason we need this field? It doesn't seem to be thread safe

core/sdk-core/src/main/java/software/amazon/awssdk/core/endpointdiscovery/EndpointDiscoveryRefreshCache.java
```diff
@@ -87,6 +90,64 @@ public URI get(String accessKey, EndpointDiscoveryRequest request) {
...
+        key = accessKey;
+
+        // Support null (anonymous credentials) by mapping to empty-string. The backing cache does not support null.
+        if (key == null) {
+            key = "";
+        }
+
+        if (request.cacheKey().isPresent()) {
+            key = key + ":" + request.cacheKey().get();
+        }
```
zoewangg, line 103: Can we reuse the same logic in `get`? maybe we can create a separate method, `getKey(request.cacheKey())`

core/sdk-core/src/main/java/software/amazon/awssdk/core/endpointdiscovery/EndpointDiscoveryRefreshCache.java
zoewangg: Yeah, suggesting the following:
```
if (endpoint != null) {
   return endpointFromCache(xxx);
}

if (request.isRequired()) {
  return discoverEndpoint(xxxx);
}

return temEndpoint(xxx);
```

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/client/test-endpoint-discovery-async.java
```diff
@@ -193,12 +193,12 @@ public CompletableFuture<TestDiscoveryIdentifiersRequiredResponse> testDiscovery
                     .overrideConfiguration().flatMap(AwsRequestOverrideConfiguration::credentialsIdentityProvider)
                     .orElseGet(() -> clientConfiguration.option(AwsClientOption.CREDENTIALS_IDENTITY_PROVIDER))
                     .resolveIdentity();
-                endpointFuture = identityFuture.thenApply(credentials -> {
+                endpointFuture = identityFuture.thenCompose(credentials -> {
```
zoewangg, line 196: Should we propagate exception cancellation? What happens if endpoint discovery takes longer than the configured client execution timeout? Would the request fail properly?

core/sdk-core/src/main/java/software/amazon/awssdk/core/endpointdiscovery/EndpointDiscoveryRefreshCache.java
```diff
@@ -87,6 +90,64 @@ public URI get(String accessKey, EndpointDiscoveryRequest request) {
...
+            return CompletableFuture.completedFuture(tempEndpoint.endpoint());
+        }
+
+        if (endpoint.expirationTime().isBefore(Instant.now())) {
+            cache.put(key, endpoint.toBuilder().expirationTime(Instant.now().plusSeconds(60)).build());
+            refreshCacheAsync(request, key);
+        }
+
+        return CompletableFuture.completedFuture(endpoint.endpoint());
+    }
```
zoewangg, line 149: Same here, can we share the code used in sync?

core/sdk-core/src/main/java/software/amazon/awssdk/core/endpointdiscovery/EndpointDiscoveryRefreshCache.java
anirudh9391: This is so that we can reference this field from within the .handle method defined for the future

core/sdk-core/src/main/java/software/amazon/awssdk/core/endpointdiscovery/EndpointDiscoveryRefreshCache.java
anirudh9391: Sure, i can add a test that validates that we get back a future. We do not have a test suit that runs against this class at the moment, though.

core/sdk-core/src/main/java/software/amazon/awssdk/core/endpointdiscovery/EndpointDiscoveryRefreshCache.java
anirudh9391: oh this is a mistake. Will change this

core/sdk-core/src/main/java/software/amazon/awssdk/core/endpointdiscovery/EndpointDiscoveryRefreshCache.java
anirudh9391: Ah right. We only need to catch the CompletableFuture exception in that case. I will change it

core/sdk-core/src/main/java/software/amazon/awssdk/core/endpointdiscovery/EndpointDiscoveryRefreshCache.java
anirudh9391: Yeah good sugesstion, I will skim through parts of code reusable and also make it more readable

core/sdk-core/src/main/java/software/amazon/awssdk/core/endpointdiscovery/EndpointDiscoveryRefreshCache.java
zoewangg: Can we derive the key in the handle method? Making this field thread safe is more error prone and may impact performance.

anirudh9391: Yeah, thats what Im going to do.

core/sdk-core/src/main/java/software/amazon/awssdk/core/endpointdiscovery/EndpointDiscoveryRefreshCache.java
anirudh9391: Sure, I will make this more readable

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/client/test-endpoint-discovery-async.java
anirudh9391: hmm.. good catch. I will propogate cancellation exceptions and terminate this in the event of a client exception / interrupt

core/sdk-core/src/main/java/software/amazon/awssdk/core/endpointdiscovery/EndpointDiscoveryRefreshCache.java
anirudh9391: done. I have modified this block

core/sdk-core/src/main/java/software/amazon/awssdk/core/endpointdiscovery/EndpointDiscoveryRefreshCache.java
anirudh9391: I have modified this to reuse major chunks of the existing getSync method.

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/client/test-endpoint-discovery-async.java
anirudh9391: I will add an interrupted exception

zoewangg: I meant if users have client execution timeout enabled and if endpoint discovery takes longer than the configured timeout, the returned future will be cancelled, but will endpoint discovery future also be cancelled?

core/sdk-core/src/main/java/software/amazon/awssdk/core/endpointdiscovery/EndpointDiscoveryRefreshCache.java
```diff
@@ -87,6 +78,20 @@ public URI get(String accessKey, EndpointDiscoveryRequest request) {
...
+        EndpointDiscoveryEndpoint endpoint = cache.get(key);
+
+        // If a service call needs to be made to discover endpoint
+        // a completable future for the service call is returned, unblocking I/O
+        // and then completed asynchronously
+        if (endpoint == null && request.required()) {
+            return discoverEndpointHandler(key, request);
+        }
+        // In the event of a cache hit, i.e. service call not required, defer to the synchronous code path method.
+        return CompletableFuture.completedFuture(get(accessKey, request));
```
zoewangg, line 92: If we call get here, it'll go through computing key and checking cache again. Can we extract the part where we refresh cache to a separate method and just call that here?

core/sdk-core/src/main/java/software/amazon/awssdk/core/endpointdiscovery/EndpointDiscoveryRefreshCache.java
```diff
@@ -109,4 +114,30 @@ public CompletableFuture<EndpointDiscoveryEndpoint> discoverEndpoint(EndpointDis
     public void evict(String key) {
         cache.remove(key);
     }
+
+    String getKey(String accessKey, EndpointDiscoveryRequest request) {
```
zoewangg, line 118: Should it be private? It's a bit code smell to expose a method just for testing

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/client/test-endpoint-discovery-async.java
zoewangg: I think we just need ` CompletableFutureUtils.forwardExceptionTo(executeFuture, endpointFuture );` on line 277. 

Can we write a test to verify that?

Example: https://github.com/aws/aws-sdk-java-v2/blob/master/services/s3/src/test/java/software/amazon/awssdk/services/s3/internal/crt/CopyObjectHelperTest.java#L314

core/sdk-core/src/main/java/software/amazon/awssdk/core/endpointdiscovery/EndpointDiscoveryRefreshCache.java
anirudh9391: Hmm.. this codepath is already getting tested out through other tests here. I can probably remove these tests

core/sdk-core/src/main/java/software/amazon/awssdk/core/endpointdiscovery/EndpointDiscoveryRefreshCache.java
anirudh9391: good callout. I will do that

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/client/test-endpoint-discovery-async.java
anirudh9391: Discussed offline. Doing this as a follow up item

zoewangg (approved):

zoewangg (approved):

core/sdk-core/src/test/java/software/amazon/awssdk/core/endpointdiscovery/EndpointDiscoveryRefreshCacheTest.java
```diff
@@ -0,0 +1,91 @@
...
+        EndpointDiscoveryRequest request = EndpointDiscoveryRequest.builder()
+                                                                   .required(true)
+                                                                   .defaultEndpoint(testURI)
+                                                                   .build();
+        CompletableFuture<URI> future = endpointDiscoveryRefreshCache.getAsync("key", request);
+        assertThat(future.isDone()).isEqualTo(false);
+
+        future.cancel(true);
+        assertThat(future.isCancelled()).isEqualTo(true);
+        assertThrows(CancellationException.class, () -> future.get());
```
dagnir, line 87: nit: prefer to use `assertThatThrownBy` since it's also part of assertj which is used for everything else.

dagnir (approved):

PR #5224, opened 2024-05-15
Prevent validating crc32 checksum on incomplete response stream

L-Applin:
This change is to enhance customer experience when dealing with `IOException`. It will prevent validating checksum when an error is encountered during reading the response body input stream and return to the user the actual error that cause a read to fail instead of the erroneous `Crc32MismatchException`.

## Motivation and Context
When an error is encountered during reading the response body, the un-marshalling of the json response is aborted. The un-marshalling abortion logic includes a call to `InputStream.close()` which in turns will trigger validating the checksum (checksum validation is performed upon closing the `Crc32ChecksumValidatingInputStream`). Currently, this checksum validation will then throw a `Crc32MismatchException` since only part of the body was read. The error returned to the user will be this `Crc32MismatchException` and the underlying actual cause for the failure to read from the input stream will be ignored and difficult to troubleshoot.

## Modifications
Add a flag to `Crc32ChecksumValidatingInputStream` class to not validate checksum if an error is encounter during reading from the response input stream.

## Testing
Added unit test.
Manual test against dynamodb.

## Types of changes


core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/util/Crc32ChecksumValidatingInputStream.java
```diff
@@ -53,7 +55,30 @@ public void close() throws IOException {
...
+            return super.read(b, off, len);
+        } catch (Throwable e) {
+            shouldValidateChecksum = false;
+            throw e;
+        }
+    }
+
     private void validateChecksum() throws Crc32MismatchException {
+        if (!shouldValidateChecksum) {
+            return;
```
joviegas, line 80: Should we log the skipping of validateChecksum   ?

joviegas (approved):

PR #5227, opened 2024-05-16
Decode encoded keys in the DeleteMarker field of ListObjectVersionsResponse

L-Applin:
As per the [ListObjectVersions API reference](https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectVersions.html) mentions, any `Key` field on the `ListObjectVersionResponse` will be encoded if the `EncodingType` field is specified in the request: 

> If you specify the encoding-type request parameter, Amazon S3 includes this element in the response, and returns encoded key name values in the following response elements:
> `KeyMarker`, `NextKeyMarker`, `Prefix`, `Key`, and `Delimiter`.

The SDK was not decoding the `DeleteMarker.Key` field. This PR is to fix that and decode the Key field of the `DeleteMarker` when required.

## Motivation and Context
Customer noticed the Key fields of the `ListObjectVersionResponse.DeleteMarker` was not decoded.

## Modifications
Add decoding logic for `DeleteMarker.Key` to the `DecodeUrlEncodedResponseInterceptor` class, an interceptor that already deals with decoding of field for various s3 response objects.

## Testing
Added unit test
Manual verification

## Types of changes


services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/handlers/DecodeUrlEncodedResponseInterceptor.java
```diff
@@ -169,4 +171,13 @@ private static List<MultipartUpload> decodeMultipartUpload(List<MultipartUpload>
                                                             .map(u -> u.toBuilder().key(urlDecode(u.key())).build())
                                                             .collect(Collectors.toList()));
     }
+
+    private static List<DeleteMarkerEntry> decodeDeleteMarkers(List<DeleteMarkerEntry> entries) {
+        if (entries == null) {
```
anirudh9391, line 176: nit : want to check for empty list too ?

L-Applin: other list method don't (`decodeCommonPrefixes(...)`, `decodeMultipartUpload(...)`) I'd rather keep similar behaviour between all of them.

anirudh9391: Sounds reasonable. Looks like we would return an empty list if it is empty. Is that intended behaviour ?

L-Applin: yep

anirudh9391 (approved):

PR #5244, opened 2024-05-24
fix(CopySourceRequestInS3): Updated codegen to customize the request before calling  service

joviegas:
Updated codegen to customize the request before sending, instead of using interceptors. This change ensures that customized parameters are available for EndpointResolveInterceptors during execution.

## Motivation and Context

- In AWS SDK Java 2.x, the “copySource” API provided by model in CopyObjectRequest is deprecated.
- It has been replaced with "sourceBucket," "sourceKey," and "sourceVersionId" via customization.
- Recently, S3 wants  “copySource” to be included in the Endpoint Params by adding in endpoint-ruleset.json
- Since "copySource" is deprecated, the new parameters "sourceBucket," "sourceKey," and "sourceVersionId" are used to update the request.
- The current binding of Endpoint Parameters from Request Parameters via the EndpointResolverInterceptors does not copy the customized values.
- The current CopySourceInterceptor copies over CopySource during the modifyRequest , but we need these to be updated in. 'beforeExecution` of EndpointResolverInterceptors

## Modifications

- Add customization to transform Custom Request

```json
  "preClientExecutionRequestCustomizer": {
    "OperationWithCustomMember": {
      "methodName": "dummyRequestModifier",
      "className": "software.amazon.awssdk.codegen.internal.UtilsTest"
    }
  }
```

Will result in
```java
    @Override
    public CompletableFuture<OperationWithCustomMemberResponse> operationWithCustomMember(
        OperationWithCustomMemberRequest operationWithCustomMemberRequest) {
        // New single line generated from above customization
        operationWithCustomMemberRequest = UtilsTest.dummyRequestModifier(operationWithCustomMemberRequest);
        SdkClientConfiguration clientConfiguration = updateSdkClientConfiguration(operationWithCustomMemberRequest,
        // Existing Code
  

}
```

## Testing

- Added Junits
- Already Integ test were present and these passed
## Types of changes

## License


services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/RequestModifierUtils.java
```diff
@@ -30,32 +26,22 @@
...
- * <a href="https://docs.aws.amazon.com/AmazonS3/latest/API/API_CopyObject.html#API_CopyObject_RequestParameters">API_CopyObject_RequestParameters</a>
+ * For more information, see:
+ * <a href="https://docs.aws.amazon.com/AmazonS3/latest/API/API_CopyObject.html#API_CopyObject_RequestParameters">CopyObject Request Parameters</a>
  * <p>
- * <a href="https://docs.aws.amazon.com/AmazonS3/latest/API/API_UploadPartCopy.html#API_UploadPartCopy_RequestParameters">API_UploadPartCopy_RequestParameters</a>
+ * <a href="https://docs.aws.amazon.com/AmazonS3/latest/API/API_UploadPartCopy.html#API_UploadPartCopy_RequestParameters">UploadPartCopy Request Parameters</a>
  */
 @SdkInternalApi
-public final class CopySourceInterceptor implements ExecutionInterceptor {
+public final class RequestModifierUtils  {
```
L-Applin, line 39: It would be good to keep a reference to `CopySource` in this class name, `CopySourceModifierUtils` maybe?

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/client/c2j/query/customization.config
```diff
@@ -21,5 +21,11 @@
         }
       }
     }
-  ]
+  ],
+  "requestCustomizerMap": {
```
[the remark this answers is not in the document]
joviegas: Any suggestions ?

codegen/src/main/java/software/amazon/awssdk/codegen/poet/client/SyncClientClass.java
```diff
@@ -341,6 +347,26 @@ private MethodSpec traditionalMethod(OperationModel opModel) {
         return method.build();
     }
 
+    public static Optional<CodeBlock> addRequestModifierCode(OperationModel opModel, IntermediateModel model) {
+        Map<String, RequestCustomizer> requestCustomizerMap = model.getCustomizationConfig().getRequestCustomizerMap();
+        if (!CollectionUtils.isNullOrEmpty(requestCustomizerMap)) {
```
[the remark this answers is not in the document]
joviegas: In this particular function, we have two cases where we return Optional.empty():

1. CollectionUtils.isNullOrEmpty(requestCustomizerMap)
2.  customRequestTransformerMap.get(opModel.getOperationName()) == null

Following the reverse if logic will result in duplicate return Optional.empty();. The code will look like this:
```java
    public static Optional<CodeBlock> addRequestModifierCode(OperationModel opModel, IntermediateModel model) {
        Map<String, CustomRequestTransformer> customRequestTransformerMap =
            model.getCustomizationConfig().getCustomRequestTransformerMap();

        if (CollectionUtils.isNullOrEmpty(customRequestTransformerMap)) {
            return Optional.empty();
        }
        CustomRequestTransformer customRequestTransformer = customRequestTransformerMap.get(opModel.getOperationName());
        if (customRequestTransformer == null) {
            return Optional.empty();
        }
        CodeBlock.Builder builder = CodeBlock.builder();
        ClassName instanceType = classNameFromFqcn(customRequestTransformer.getClassName());
        builder.addStatement("$L = $T.$N($L)",
                             opModel.getInput().getVariableName(),
                             instanceType,
                             customRequestTransformer.getMethodName(),
                             opModel.getInput().getVariableName());
        return Optional.of(builder.build());
    }
```

Since there are two cases where we need to return Optional.empty(), avoiding code duplication is preferable for better readability.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/RequestModifierUtils.java
joviegas: I updated the name as CustomRequestTransformerUtils since in future if we need to do similar customization to any API then we can just add an API in this class.
Thus this class will have broader scope of all the RequestTransformer rather than just `CopySource`

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/client/c2j/query/customization.config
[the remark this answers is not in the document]
joviegas: Updated the name as `CustomRequestTransformer`
Will also have internal discussion
Let me know if you have any suggestion of names

codegen/src/main/java/software/amazon/awssdk/codegen/model/config/customization/CustomizationConfig.java
```diff
@@ -328,6 +329,11 @@ public class CustomizationConfig {
 
     private List<CustomOperationContextParam> customOperationContextParams;
 
+    /**
+     * Special case API where an api request needs customization before the service call is made.
+     */
```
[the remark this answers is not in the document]
joviegas: Done

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/client/c2j/query/customization.config
```diff
@@ -21,5 +21,11 @@
         }
       }
     }
-  ]
+  ],
+  "customRequestTransformerMap": {
```
L-Applin, line 25: **nit**: I would drop the `map` suffix (everything is a map in json), simply `customRequestTransformers` sounds more clear to me. Other keys, like `customOperationContextParams` don't have the map suffix, it would make the customization file more consistent. But it is not a blocker for me.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/RequestModifierUtils.java
L-Applin: Oh you were thinking of reusing this class for other things. I see. Would it be better to keep this class only for CopySource purpose or not?

L-Applin (approved):

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/RequestModifierUtils.java
joviegas: I was thinking of using this class to hold the static APIs that do the requests transformations.
In this case we have two API `copyObject` and `UploadPartCopy`.
Since this utils is created to hold the CustomRequestTransformers thart will be used in Customizations I was thinking to keep its scope at this level rather than CopySource which is just a Field which gets transformed.

PR #5254, opened 2024-05-29
add completed to ResumableFileDownload ser/des

L-Applin:
Add `completedParts` field to `ResumableFileDownload` with required serialization and deserialization.

## Motivation and Context
Needed to Multipart Pause/Resume logic. The `ResumableFileDownload` can be serialized to json and deserialized back into memory to resume a download for it. 

## Modifications
- Add new field to `ResumableFileDownload` + Builder
- Add a new `TransferManagerJsonUnmarshaller` for list of Integer
- Update marshalling/unmarshalling to include `completedParts` list

## Testing
Added unit tests

### DIFF
--- a/services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/model/ResumableFileDownload.java
+++ b/services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/model/ResumableFileDownload.java
@@ -174,6 +174,7 @@ public final class ResumableFileDownload implements ResumableTransfer,
                        .add("s3ObjectLastModified", s3ObjectLastModified)
                        .add("totalSizeInBytes", totalSizeInBytes)
                        .add("downloadFileRequest", downloadFileRequest)
+                       .add("completedParts", completedParts)
                        .build();
     }
 
--- a/services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/model/ResumableFileDownload.java
+++ b/services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/model/ResumableFileDownload.java
@@ -23,6 +23,8 @@ import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.time.Instant;
+import java.util.Collections;
+import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.OptionalLong;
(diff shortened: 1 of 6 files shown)


davidh44 (approved):

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/model/ResumableFileDownload.java
```diff
@@ -149,6 +157,15 @@ public OptionalLong totalSizeInBytes() {
         return totalSizeInBytes == null ? OptionalLong.empty() : OptionalLong.of(totalSizeInBytes);
     }
 
+    /**
+     * The lists of parts that were successfully completed and saved to the file. Non-empty only for multipart downloads.
+     *
+     * @return part numbers of a multipart download that were completed saved to file.
+     */
+    public List<Integer> completedParts() {
+        return Collections.unmodifiableList(completedParts);
```
zoewangg, line 166: Can we wrap with umodifableList in the ctor instead of the getter? Also, we should probably create a new list. Example: https://github.com/aws/aws-sdk-java-v2/blob/master/core/sdk-core/src/main/java/software/amazon/awssdk/core/client/config/ClientOverrideConfiguration.java#L148

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/model/ResumableFileDownload.java
```diff
@@ -61,6 +63,7 @@ public final class ResumableFileDownload implements ResumableTransfer,
     private final Instant s3ObjectLastModified;
     private final Long totalSizeInBytes;
     private final Instant fileLastModified;
+    private final List<Integer> completedParts;
```
zoewangg, line 66: Is the reason for List that we may download different parts in parallel in the future?

L-Applin: yes, trying to prepare for when we will do parallel write to file

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/model/ResumableFileDownload.java
L-Applin: will address in full resume PR

PR #5258, opened 2024-05-30
Pause and resume for multipart download

L-Applin:
Implement transfer-manager pause and resume feature for multipart download. Multipart download to file with the java-based client will be resumed from after the last completed part.

## Motivation and Context
Previously, when resuming any download to file, a GetObject with the Range parameter was performed to s3, resulting in the remaining content of the object being downloaded as one bog chunk, bypassing multipart operations.

## Modifications
Add a new `MultipartDownloadResumeContext`, bound the the original GetRequest through a new multipart execution attribute. This context instance is updated during multipart download and can be access upon resuming.

## Testing
Unit test
Integ test
Manual validation


services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/GenericS3TransferManager.java
```diff
@@ -289,16 +295,31 @@ private CopyObjectRequest attachSdkAttribute(CopyObjectRequest copyObjectRequest
...
+                                                                             .build());
 
         return copyObjectRequest.toBuilder()
-                               .overrideConfiguration(modifiedRequestOverrideConfig)
-                               .build();
+                                .overrideConfiguration(modifiedRequestOverrideConfig)
+                                .build();
     }
 
+    private GetObjectRequest attachSdkAttributes(GetObjectRequest request,
```
davidh44, line 308: nit: singular Attribute

services-custom/s3-transfer-manager/src/it/java/software/amazon/awssdk/transfer/s3/S3TransferManagerMultipartDownloadPauseResumeIntegrationTest.java
```diff
@@ -0,0 +1,109 @@
...
+    void pauseAndResume_whenAlreadyComplete_shouldHandleGracefully() {
+        Path path = RandomTempFile.randomUncreatedFile().toPath();
+        DownloadFileRequest request = DownloadFileRequest.builder()
+                                                         .getObjectRequest(b -> b.bucket(BUCKET).key(KEY))
+                                                         .destination(path)
+                                                         .build();
+        FileDownload download = tmJava.downloadFile(request);
+        download.completionFuture().join();
+        ResumableFileDownload resume = download.pause();
+        FileDownload resumedDownload = tmJava.resumeDownloadFile(resume);
```
davidh44, line 93: Since totalParts is not currently set, completedDownload() is never called. Can we assert that `"Sending downloadFileRequest "` logging is not called a second time, and the logging for a completedDownload is called?

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/multipart/MultipartDownloadResumeContext.java
```diff
@@ -0,0 +1,132 @@
...
+
+    /**
+     * Check if the multipart download is complete or not by checking if the total amount of downloaded parts is equal to the
+     * total amount of parts.
+     *
+     * @return true if all parts were downloaded, false if not.
+     */
+    public boolean isComplete() {
+        if (totalParts == null) {
+            return false;
```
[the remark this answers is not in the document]
L-Applin: good catch, `totalPart` should be set in `MultipartDownloaderSubscribe`, I swear I had done it... must have revert it somehow. Fixing now

services-custom/s3-transfer-manager/src/it/java/software/amazon/awssdk/transfer/s3/S3TransferManagerMultipartDownloadPauseResumeIntegrationTest.java
L-Applin: Good idea, I'll add some `LogCaptor` assertions

davidh44 (approved):

PR #5185, opened 2024-05-06
Fix reload credentials not working when modified

munendrasn:
Ensure defaultSupplier is used instead of fixedSupplier in case of defaultProfileFile.

I have updated the some places where `defaultProfileFile` was used along with `fixedProfileSupplier` with `defaultSupplier`

## Motivation and Context
Fixes #4268

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/aws/aws-sdk-java-v2/blob/master/CONTRIBUTING.md) document
- [ ] Local run of `mvn install` succeeds
- [ ] My code follows the code style of this project
- [ ] My change requires a change to the Javadoc documentation
- [ ] I have updated the Javadoc documentation accordingly
- [x] I have added tests to cover my changes
- [ ] All new and existing tests passed
- [ ] I have added a changelog entry. Adding a new entry must be accomplished by running the `scripts/new-change` script and following the instructions. Commit the new file created by the script in `.changes/next-release` with your changes.
- [ ] My change is to implement 1.11 parity feature and I have updated [LaunchChangelog](https://github.com/aws/aws-sdk-java-v2/blob/master/docs/LaunchChangelog.md)

## License

- [x] I confirm that this pull request can be released under the Apache 2 license

PR #5259, opened 2024-05-31
Fix reload credentials not working when modified - external PR : #5185

anirudh9391:
Refer to external PR link : https://github.com/aws/aws-sdk-java-v2/pull/5185


munendrasn:
Fix the build test failures https://github.com/aws/aws-sdk-java-v2/pull/5185/commits/d7cd3a080184d828333d896ef10013a69a028425.
The defaultSupplier uses `ProfileFile.builder` as fallback for cases when config & credential file doesn't exist but builder expects contentLocation or content to be present. This is resolved by using dummy aggregator, other option was to use `defaultProfileFile` both would have same behaviour but defaultProfileFile would look to read and parse non-existing file (works, redundant parsing)

The failures were not surfacing in local, as either config or credential file was set, which is not the case with CI build 
FYI @anirudh9391

munendrasn:
There is one test failure in recent build, it is due to new profileFile being created every time when config/credential file is not set. Resolve it in the commit https://github.com/aws/aws-sdk-java-v2/pull/5185/commits/fd04bab07957ee8e39c7e4c76927eb60fe4e35cf FYI @anirudh9391

munendrasn:
Removed the unused import causing the checkstyle failure for auth module https://github.com/aws/aws-sdk-java-v2/pull/5185/commits/b737a7c341e255bac01fcbb01fa450eed6c79d2e FYI @anirudh9391

munendrasn:
Removed unused imports from tests too, although not causing the failures https://github.com/aws/aws-sdk-java-v2/pull/5185/commits/9088b59459609037dc4217aa160e7d058df712c6

core/auth/src/test/java/software/amazon/awssdk/auth/credentials/DefaultCredentialsProviderTest.java
```diff
@@ -97,6 +107,44 @@ void resolveCredentials_ProfileCredentialsProviderWithSupplierProfileFile_return
...
+
+        assertThat(provider.resolveCredentials()).satisfies(awsCredentials -> {
+            assertThat(awsCredentials.accessKeyId()).isEqualTo("customAccess");
+            assertThat(awsCredentials.secretAccessKey()).isEqualTo("customSecret");
+        });
+
+        generateTestCredentialsFile("modifiedAccess", "modifiedSecret");
+
+        // without sleep the assertion fails, as there is delay in config reload
+        Thread.sleep(2000);
```
munendrasn, line 125: @anirudh9391 For Java8, the test is failing pos this, had included delay to ensure latest config is picked up, but it is failing occasionally, 
```
Error:    DefaultCredentialsProviderTest.resolveCredentials_DefaultCredentialProviderWithReloadWhenModified:126->lambda$resolveCredentials_DefaultCredentialProviderWithReloadWhenModified$7:127
expected: "modifiedAccess"
 but was: "customAccess"

Should we increase the sleep value? Is there any alternative way to validate this?
```
For Java17, the test failures seems to be unrelated & from s3 module 
```
2024-06-11T16:42:29.7804890Z [ERROR] Errors:
2024-06-11T16:42:29.7805324Z [ERROR]   CrtDownloadErrorTest.verifyCrtResource:73 ? Runtime Timeout waiting for resource count to drop to zero
2024-06-11T16:42:29.7805829Z [ERROR]   CrtCredentialProviderAdapterTest.verifyCrtResource:43 ? Runtime Timeout waiting for resource count to drop to zero
2024-06-11T16:42:29.7806270Z [ERROR]   S3CrtClientWiremockTest.verifyCrtResource:92 ? Runtime Timeout waiting for resource count to drop to zero
```

anirudh9391: The Java17 error seems to be transient in nature. It seems to succeed locally. I will play around with timeout values a bit.

munendrasn: Due to the caching of ProfileFile for short duration, https://github.com/aws/aws-sdk-java-v2/blob/fb7b6ea57c5512fa6229af06b904070054ed1597/utils/src/main/java/software/amazon/awssdk/utils/cache/CachedSupplier.java#L242
the credentials are not reloaded immediately, hence, Thread.sleep is required, added 2s sleep which in most cases sufficient, is not working in CI sometimes. 
Here, sleep duration can be increased or assertion can retried with smaller timeout values

For reference, in ProfileRefresherTest https://github.com/aws/aws-sdk-java-v2/blob/fb7b6ea57c5512fa6229af06b904070054ed1597/core/profiles/src/test/java/software/amazon/awssdk/profiles/internal/ProfileFileRefresherTest.java#L134 

Custom clock is used to advance the timer to validate refresh without sleep which might not be possible here, ie to set custom clock in ProfileSupplier

anirudh9391: I tried it with a 5 second delay. Seems to trip up the java21 and java8 builds, both of which fail assertion. Please feel free to cut another revision with any modifications to add retries or anything else. I think we can keep timeout at 5 seconds.

anirudh9391: I pushed another revision with a retry. Lets see if this works

anirudh9391: Seems like the retry failed all the java builds

munendrasn: Thanks @anirudh9391 , will take a look. Going through the commit, I think retry should be added to second assertion, once credentials are modified

munendrasn: Updated the retry logic, along with additional logging to debug the test, if it fails again in the commits https://github.com/aws/aws-sdk-java-v2/pull/5185/commits/e360e9d13e1e5485110306ad27fc4f02619211c9 and https://github.com/aws/aws-sdk-java-v2/pull/5185/commits/f3304575a2b4315d01e666a50edd837abbcf472c respectively

munendrasn: @anirudh9391 could you please cherry-pick latest commit for the fix, and further debugging?

PR #5270, opened 2024-06-05
Improve SRA signer user experience, with documentation and helper methods.

millems:
1. Deprecate old signer interfaces, directing to new interfaces.
2. Add Javadoc for new v4 and v4a signers.
3. Add utility methods for creating ContentStreamProviders.


core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/signer/AwsV4FamilyHttpSigner.java
```diff
@@ -64,13 +69,21 @@ public interface AwsV4FamilyHttpSigner<T extends Identity> extends HttpSigner<T>
...
+     * support this value being disabled, especially over HTTPS where SSL provides some of its own protections against
+     * payload tampering.
      */
     SignerProperty<Boolean> PAYLOAD_SIGNING_ENABLED =
         SignerProperty.create(AwsV4FamilyHttpSigner.class, "PayloadSigningEnabled");
 
     /**
      * Whether to indicate that a payload is chunk-encoded or not. This property defaults to false. This can be set true to
-     * enable the `aws-chunk` content-encoding
+     * enable the `aws-chunk` content-encoding.
```
dagnir, line 83: minor: should this be `aws-chunked`?

dagnir (approved):

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/signer/AwsV4FamilyHttpSigner.java
millems: Fixed!

PR #5300, opened 2024-06-15
 Fix a bug that prevents a retry strategy override to get used

sugmanue:
## Motivation and Context

Fixes a bug that was found that prevents a retry strategy override from being used as it was overwritten once again to configure it for AWS. 

The original logic was there to add AWS-specific retry conditions for when the user set the retry strategy using `RetryMode` or using the mutator to tailor the default one `Consumer<RetryStrategy.Builder>`. We eagerly created the retry strategy, but since this package (`sdk-core`) does not know about AWS packages was not able to set the correct one, thus we did patch it after the fact, but this logic is not correct as it prevents user defined overrides to take place and will "over-configure" if an already AWS aware retry strategy is given.

## Modifications

We now resolve the retry strategy lazily, either when using `RetryMode` or the default one to apply the configured configurator. We need to add logic to the client builder classes and to the client to accomplish this in three different scenarios:

1. Client builder using client override configuration
2. Client builder using plugins
3. Request override configuration using plugins

For that the following method was added to the client builder and to the client to update the configuration after the plugins are run:

```java
    private void updateRetryStrategyClientConfiguration(SdkClientConfiguration.Builder configuration) {
        ClientOverrideConfiguration.Builder builder = configuration.asOverrideConfigurationBuilder();
        RetryMode retryMode = builder.retryMode();
        if (retryMode != null) {
            configuration.option(SdkClientOption.RETRY_STRATEGY, AwsRetryStrategy.forRetryMode(retryMode));
            return;
        }
        Consumer<RetryStrategy.Builder<?, ?>> configurator = builder.retryStrategyConfigurator();
        if (configurator != null) {
            RetryStrategy.Builder<?, ?> defaultBuilder = AwsRetryStrategy.defaultRetryStrategy().toBuilder();
            configurator.accept(defaultBuilder);
            configuration.option(SdkClientOption.RETRY_STRATEGY, defaultBuilder.build());
            return;
        }
        RetryStrategy retryStrategy = builder.retryStrategy();
        if (retryStrategy != null) {
            configuration.option(SdkClientOption.RETRY_STRATEGY, retryStrategy);
        }
    }
```

This implies that now the clients themselves have a direct dependency on `retries-spi` which was also added for all services updating their `pom.xml`.

## Types of changes

## Checklist

## License


joviegas:
There we lot of test files so was not able to pin point which test covers functional testing of this bug ? Could you please help with the Java test files which actually test the functionality when  retry strategy is overriden.

sugmanue:
> Could you please help with the Java test files which actually test the functionality when retry strategy is overridden.

[CanOverrideRetryStrategy.java](https://github.com/aws/aws-sdk-java-v2/pull/5300/files#diff-23d73408e8d3e173cb829ef5a8358dd96c54e120bc7de9c07ae1a332b2af9588) (the link might not work but that's the name of the file at the very bottom of the changes)

core/sdk-core/src/main/java/software/amazon/awssdk/core/client/config/SdkClientOption.java
```diff
@@ -60,9 +61,30 @@ public final class SdkClientOption<T> extends ClientOption<T> {
     /**
      * @see ClientOverrideConfiguration#retryStrategy()
      */
-    @SuppressWarnings("rawtypes")
     public static final SdkClientOption<RetryStrategy> RETRY_STRATEGY = new SdkClientOption<>(RetryStrategy.class);
 
+    /**
+     * The internal retry strategy set by the customer. This is likely only useful within configuration
```
cenedhryn, line 67: The usage of the word `internal` here is confusing to me. Can we say `The retry strategy set by the customer [where/when]` instead?

pom.xml
```diff
@@ -645,7 +645,7 @@
                             <includeModule>profiles</includeModule>
                             <includeModule>protocols</includeModule>
                             <includeModule>regions</includeModule>
-                            <includeModule>sdk-core</includeModule>
+                            <!-- <includeModule>sdk-core</includeModule> -->
```
cenedhryn, line 648: This doesn't look right.

response.json
```diff
@@ -0,0 +1 @@
+{"statusCode": 200, "body": "\"Wheel spinned and selected joviegas\""}
```
cenedhryn, line 1: Let's not include this file!

core/sdk-core/src/main/java/software/amazon/awssdk/core/client/config/ClientOverrideConfiguration.java
```diff
@@ -474,17 +499,16 @@ default Builder retryStrategy(RetryMode retryMode) {
         Builder retryStrategy(RetryStrategy retryStrategy);
 
         /**
-         * Configure the retry strategy that should be used when handling failure cases.
+         * Configure a new default retry strategy that should be used when handling failure cases using the given consumer for it.
```
cenedhryn, line 502: We should have the same wording for all three methods - I can see they're mostly similar. To modify:
- each should say that the retryMode methods are mutually exclusive
- each should say that they override policy
- this method says `new default` and the others do not - let's consistently use one or the other. 
I'm a bit nitpicky here but I think this topic requires clarity!

pom.xml
sugmanue: It's needed since `retryStrategy(RetryMode)` and `retryStrategy(Consumer<RetryStrategy.Consumer>)` are not default methods but instead abstract. I'm like 98.5% confident that this is backwards compatible so I feel like it's a false positive but still. The plan is to remove it after we release this change.

Without this the build fails with:

```
[ERROR] Failed to execute goal com.github.siom79.japicmp:japicmp-maven-plugin:0.15.6:cmp (default) on project sdk-core: There is at least one incompatibility: software.amazon.awssdk.core.client.config.ClientOverrideConfiguration$Builder.retryStrategy(software.amazon.awssdk.core.retry.RetryMode):METHOD_NOW_ABSTRACT,software.amazon.awssdk.core.client.config.ClientOverrideConfiguration$Builder.retryStrategy(java.util.function.Consumer):METHOD_NOW_ABSTRACT -> [Help 1]
```

response.json
sugmanue: 🤦 let's not.

services/athena/pom.xml
```diff
@@ -61,5 +61,10 @@
             <artifactId>http-auth-aws</artifactId>
             <version>${awsjavasdk.version}</version>
         </dependency>
+        <dependency>
```
zoewangg, line 64: Can we add `retries-spi` to `services` pom.xml(shared by all services) instead?

sugmanue: This change along reduced the complexity of this change a lot, thanks for point this out @zoewangg!

core/sdk-core/src/main/java/software/amazon/awssdk/core/client/config/ClientOverrideConfiguration.java
sugmanue: > We should have the same wording for all three methods - I can see they're mostly similar. To modify:

Can you please suggest a new wording? That would make it a lot easier than guessing what you're thinking about, e.g., 

> each should say that the retryMode methods are mutually exclusive

There's only one method using retry model, so there are not many that can be *mutually exclusive*.

> each should say that they override policy

As in retry policy? Those do not override policies, I guess we can but we don't at the moment.

> this method says new default and the others do not - let's consistently use one or the other.

The idea is that this method will create a *new* default retry strategy (as given by `AwsRetryStrategy.defaultRetryStrategy()`) and pass that to the `mutator`. If we had access to `aws-core` it would be implemented as 

```java
    default Builder retryStrategy(Consumer<RetryStrategy.Builder<?, ?>> mutator) {
        RetryStrategy.Builder<?, ?> defaultBuilder = AwsRetryStrategy.defaultRetryStrategy().toBuilder();
        configurator.accept(defaultBuilder);
        return retryStrategy(defaultBuilder.build());
    }
```

core/sdk-core/src/main/java/software/amazon/awssdk/core/client/config/SdkClientOption.java
sugmanue: I updated the docs, e.g.,

```java
    /**
     * The retry mode set by the customer using {@link ClientOverrideConfiguration.Builder#retryStrategy(RetryMode)}. This is
     * likely only useful within configuration classes, and will be converted into a {@link #RETRY_STRATEGY} for the SDK's
     * runtime.
     *
     * @see ClientOverrideConfiguration#retryMode()
     */
```

Let me know if you have a better wording that I can use.

codegen/src/main/java/software/amazon/awssdk/codegen/poet/client/ClientClassUtils.java
```diff
@@ -40,12 +41,17 @@
...
+import software.amazon.awssdk.core.client.config.SdkClientOption;
+import software.amazon.awssdk.core.retry.RetryMode;
 import software.amazon.awssdk.core.signer.Signer;
+import software.amazon.awssdk.retries.api.RetryStrategy;
 import software.amazon.awssdk.utils.HostnameValidator;
 import software.amazon.awssdk.utils.StringUtils;
 import software.amazon.awssdk.utils.Validate;
 
-final class ClientClassUtils {
+public final class ClientClassUtils {
```
sugmanue, line 54: We share the `updateRetryStrategyClientConfigurationMethod` with the client builder logic, so this class now needs to be public.

core/sdk-core/src/main/java/software/amazon/awssdk/core/client/builder/SdkDefaultClientBuilder.java
```diff
@@ -133,12 +136,12 @@ public abstract class SdkDefaultClientBuilder<B extends SdkClientBuilder<B, C>,
     protected final SdkClientConfiguration.Builder clientConfiguration = SdkClientConfiguration.builder();
 
     protected final AttributeMap.Builder clientContextParams = AttributeMap.builder();
-
+    protected ClientOverrideConfiguration overrideConfig;
```
sugmanue, line 139: The codegen clients builders need access to the client override configuration to properly configure the `RetryStrategy`.

core/aws-core/src/main/java/software/amazon/awssdk/awscore/client/builder/AwsDefaultClientBuilder.java
```diff
@@ -187,6 +189,26 @@ private SdkClientConfiguration finalizeAwsConfiguration(SdkClientConfiguration c
                             .build();
     }
 
+    /**
+     * Apply the client override configuration to the provided configuration.
+     */
+    protected SdkClientConfiguration setOverrides(SdkClientConfiguration configuration) {
```
zoewangg, line 195: Is this method intended to be overridden? If not, can we add `final`? Also, missing override annotation

core/sdk-core/src/main/java/software/amazon/awssdk/core/client/config/ClientOverrideConfiguration.java
```diff
@@ -458,33 +485,39 @@ default Builder retryPolicy(RetryMode retryMode) {
...
-         * retry policy on this builder.
+         * Configure the retry mode used to resolve the corresponding {@link RetryStrategy} that should be used when handling
+         * failure cases.
+         *
+         * @see RetryMode
          */
-        default Builder retryStrategy(RetryMode retryMode) {
-            return retryStrategy(SdkDefaultRetryStrategy.forRetryMode(retryMode));
-        }
+        Builder retryStrategy(RetryMode retryMode);
```
zoewangg, line 493: Seems like breaking change, same for other methods which we took out default implementation from

sugmanue: It's marked as breaking by japicmp but I don't know what else we can do about it. Any suggestions?

core/aws-core/src/main/java/software/amazon/awssdk/awscore/client/builder/AwsDefaultClientBuilder.java
sugmanue: No, it's not, I can add the override and make it final.

core/sdk-core/src/main/java/software/amazon/awssdk/core/client/config/ClientOverrideConfiguration.java
zoewangg: Can we keep the original implementation? Or alternatively, we can change it to throw UnsupportedOperationException

sugmanue: Oh, that's a good idea, keep it default and throw and override it in the impl, yeah, we can do that, let me update the PR.

joviegas (approved):

core/aws-core/src/main/java/software/amazon/awssdk/awscore/client/builder/AwsDefaultClientBuilder.java
```diff
@@ -187,6 +189,26 @@ private SdkClientConfiguration finalizeAwsConfiguration(SdkClientConfiguration c
...
+        SdkClientConfiguration.Builder builder = configuration.toBuilder()
+                                                              .putAll(overrideConfig);
+        overrideConfig.retryStrategy().ifPresent(retryStrategy -> builder.option(RETRY_STRATEGY, retryStrategy));
+        overrideConfig.retryMode().ifPresent(retryMode -> builder.option(RETRY_STRATEGY,
+                                                                         AwsRetryStrategy.forRetryMode(retryMode)));
+        overrideConfig.retryStrategyConfigurator().ifPresent(configurator -> {
+            RetryStrategy.Builder<?, ?> defaultBuilder = AwsRetryStrategy.defaultRetryStrategy().toBuilder();
+            configurator.accept(defaultBuilder);
+            builder.option(RETRY_STRATEGY, defaultBuilder.build());
+        });
```
cenedhryn, line 208: Should we not nullify the configured options here as well as in the parent method implementation? If not, what condition differs?

core/sdk-core/src/main/java/software/amazon/awssdk/core/client/config/SdkClientOption.java
cenedhryn: I like this wording, thanks for updating.

core/aws-core/src/main/java/software/amazon/awssdk/awscore/client/builder/AwsDefaultClientBuilder.java
sugmanue: Yes, we should, let me add a change for that.

core/sdk-core/src/main/java/software/amazon/awssdk/core/client/config/ClientOverrideConfiguration.java
cenedhryn: Thanks for the updates. 

Sorry, I meant to say that the `retryStrategy` methods are mutually exclusive settings, not retryMode. That was confusing. If we can include a reference to that, I'd appreciate it. (Nit)
`Note that retryStrategy options are mutually exclusive`?

Disregard the comment about policy, I think that wording is gone anyway.

cenedhryn (approved):

joviegas (approved):

PR #5319, opened 2024-06-21
Treat UncheckedIOException as IOExceptions

anirudh9391:
## Motivation and Context
Timeout triggered termination of the connection stream, causes SDK to not perform retries even if configured. 

## Modifications
Users of DynamoDB experienced and reported that API timeouts causing UncheckedIOException do not get retried, despite the client being configured to retry timeout exceptions. 

## Testing
Added unit test

## Types of changes

- [ x ] Bug fix (non-breaking change which fixes an issue)


joviegas:
Could you please mention some details like which service was facing this issue and how this issue , surfaced ?
In this particular who exactly was throwing UncheckedIOException  ?
Also please add changelogs

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/TimeoutExceptionHandlingStage.java
```diff
@@ -118,6 +119,10 @@ private Exception handleTimeoutCausedException(RequestExecutionContext context,
             return AbortedException.create("Thread was interrupted", e);
         }
 
+        if (e instanceof UncheckedIOException) {
+            return new IOException(e.getCause());
+        }
+
```
zoewangg, line 125: Do we need this logic? If the stream is closed by the timer, then it will return from line 107 or 118, right?

joviegas:
Something to also consider while fixing is instance like https://github.com/aws/aws-sdk-java-v2/blob/9fec27c2d2f04167ae7c8cf91a958f8f3aa4bde4/core/sdk-core/src/main/java/software/amazon/awssdk/core/BytesWrapper.java#L91 where UnCheckedIOException are exception which do not pass on retries .

anirudh9391:
> aws-sdk-java-v2/core/sdk-core/src/main/java/software/amazon/awssdk/core/BytesWrapper.java

Yes, as discussed offline yesterday, we need to cut a broader task to identify and handle UncheckedIOExceptions throughout the SDK codebase. Now that we know it most certainly has to be retried. This SIM addresses the imminent task, which is timeout caused UncheckedIOExceptions.

joviegas:
- Can we please update the title and change log to reflect that these changes modify the handling of exceptions when closing the response stream during an SDKInterrupted exception, changing it to just log the cause of failure.
- If possible, can we write a functional test or at least perform a one-off test to ensure this addresses the issue.
- Can we please also update the fix for ApiCallAttemptTimeoutTrackingStage since the reported stack trace showed this class.

zoewangg:
Closing in favor of https://github.com/aws/aws-sdk-java-v2/pull/5439

PR #5320, opened 2024-06-21
Do not add chunk data for empty ByteBuffer

davidh44:
## Motivation and Context

When doing S3 `PutObject` with flexible checksums with `S3AsyncClient`, the `ChecksumCalculatingAsyncRequestBody` adds chunk headers and trailers to each chunk. We should skip this if there is a non-final empty chunk of length 0, as it causes an error to be thrown

```
S3Exception: You did not provide the number of bytes specified by the Content-Length HTTP header
```

## Modifications

In `onNext()`, return the `ByteBuffer` as is, if it is empty and non-final chunk

## Testing

added unit tests, performed one-off integ tests

## Types of changes

## Checklist

## License


services/s3/pom.xml
```diff
@@ -224,5 +224,17 @@
             <artifactId>jimfs</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
zoewangg, line 227: We should probably avoid circular dependency in the SDK. It'll likely cause more dependency issues in the future. (Side note, I didn't know you could use exclusions trick to work around the issue)

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/ChecksumCalculatingAsyncRequestBody.java
```diff
@@ -197,9 +197,11 @@ public void onNext(ByteBuffer byteBuffer) {
                     checksumBytes = checksum.getChecksumBytes();
                     ByteBuffer allocatedBuffer = getFinalChecksumAppendedChunk(byteBuffer);
                     wrapped.onNext(allocatedBuffer);
-                } else {
+                } else if (byteBuffer.remaining() > 0) {
```
zoewangg, line 200: nit: hasRemaining()?

services/s3/pom.xml
davidh44: removed integ test, one-off testing and unit test should be sufficient

zoewangg (approved):

PR #5321, opened 2024-06-21
Codegenerate BatchManager API under AsyncClient and Initail Interfaces for BatchManager

joviegas:
## Motivation and Context

PR similar to https://github.com/aws/aws-sdk-java-v2/pull/2666. But following are considered after design reviews
1. The BatchManager Implementation will be hand written , thus we donot need codegenrated BatchFunctions and code generated Default batch Manager.
2.  We will proivide batchManager() api only to AsyncClient , since returning  completable future for sync Api doesnot make sense for Sync API  , instead sync API can directly call send/recieve with custom BatchedRequest in its request.

## Modifications

Customization 
```json
  "batchManagerSupported": true
```

Will result in code
1. Addition of scheduledExecutor similar to that of waiters()
2.  Java code as below

```java
    @Override
    public BatchManagerTestAsyncBatchManager batchManager() {
        return BatchManagerTestAsyncBatchManager.builder().client(this).scheduledExecutor(executorService).build();
    }
```

## Testing

1. Added Junits
2. Also tested with SQS , made sure other service not impacted

## License

### DIFF
--- a/codegen/src/main/java/software/amazon/awssdk/codegen/poet/client/AsyncClientClass.java
+++ b/codegen/src/main/java/software/amazon/awssdk/codegen/poet/client/AsyncClientClass.java
@@ -15,6 +15,7 @@
 
 package software.amazon.awssdk.codegen.poet.client;
 
+import static com.squareup.javapoet.TypeSpec.Builder;
 import static java.util.Collections.singletonList;
 import static java.util.stream.Collectors.toList;
 import static javax.lang.model.element.Modifier.FINAL;
@@ -104,6 +105,7 @@ public final class AsyncClientClass extends AsyncClientInterface {
     private final ClassName serviceClientConfigurationClassName;
     private final ServiceClientConfigurationUtils configurationUtils;
     private final boolean useSraAuth;
+    private boolean hasScheduledExecutor;
 
     public AsyncClientClass(GeneratorTaskParams dependencies) {
         super(dependencies.getModel());
@@ -139,7 +141,7 @@ public final class AsyncClientClass extends AsyncClientInterface {
     }
 
     @Override
-    protected void addFields(TypeSpec.Builder type) {
+    protected void addFields(Builder type) {
         type.addField(FieldSpec.builder(ClassName.get(Logger.class), "log")
                                .addModifiers(PRIVATE, STATIC, FINAL)
                                .initializer("$T.getLogger($T.class)", LoggerFactory.class,
@@ -153,6 +155,10 @@ public final class AsyncClientClass extends AsyncClientInterface {
         // Kinesis doesn't support CBOR for STS yet so need another protocol factory for JSON
         if (model.getMetadata().isCborProtocol()) {
             type.addField(AwsJsonProtocolFactory.class, "jsonProtocolFactory", PRIVATE, FINAL);
+        }
+
+        if (model.getCustomizationConfig().getBatchManagerSupported() || model.hasWaiters()) {
+            addScheduledExecutorIfNeeded(type);
         }
 
         model.getEndpointOperation().ifPresent(
@@ -180,9 +186,6 @@ public final class AsyncClientClass extends AsyncClientInterface {
 
     @Override
     protected void addWaiterMethod(TypeSpec.Builder type) {
-        type.addField(FieldSpec.builder(ClassName.get(ScheduledExecutorService.class), "executorService")
-                                       .addModifiers(PRIVATE, FINAL)
-                                       .build());
 
         MethodSpec waiter = MethodSpec.methodBuilder("waiter")
                                       .addModifiers(PUBLIC)
@@ -263,7 +266,7 @@ public final class AsyncClientClass extends AsyncClientInterface {
             builder.endControlFlow();
         }
 
-        if (model.hasWaiters()) {
+        if (model.hasWaiters() || model.getCustomizationConfig().getBatchManagerSupported()) {
             builder.addStatement("this.executorService = clientConfiguration.option($T.SCHEDULED_EXECUTOR_SERVICE)",
                                  SdkClientOption.class);
         }
@@ -547,6 +550,26 @@ public final class AsyncClientClass extends AsyncClientInterface {
                          .build();
     }
 
+    @Override
+    protected void addBatchManagerMethod(Builder type) {
+
+        String scheduledExecutor = "executorService";
+        ClassName returnType;
+
+        returnType = poetExtensions.getBatchManagerAsyncInterface();
+
+        MethodSpec batchManager = MethodSpec.methodBuilder("batchManager")
+                                            .addModifiers(PUBLIC)
+                                            .addAnnotation(Override.class)
+                                            .returns(returnType)
+                                            .addStatement("return $T.builder().client(this).scheduledExecutor($N).build()",
+                                                          returnType, scheduledExecutor)
+                                            .build();
+
+
+        type.addMethod(batchManager);
+    }
+
     private MethodSpec resolveMetricPublishersMethod() {
         String clientConfigName = "clientConfiguration";
         String requestOverrideConfigName = "requestOverrideConfiguration";
@@ -620,5 +643,14 @@ public final class AsyncClientClass extends AsyncClientInterface {
     private boolean hasStreamingV4AuthOperations() {
         return model.getOperations().values().stream()
                 .anyMatch(this::shouldUseAsyncWithBodySigner);
+    }
+
+    private void addScheduledExecutorIfNeeded(Builder classBuilder) {
+        if (!hasScheduledExecutor) {
+            classBuilder.addField(FieldSpec.builder(ClassName.get(ScheduledExecutorService.class), "executorService")
+                                           .addModifiers(PRIVATE, FINAL)
+                                           .build());
+            hasScheduledExecutor = true;
+        }
     }
 }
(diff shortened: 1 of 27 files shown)


services/sqs/src/main/java/software/amazon/awssdk/services/sqs/internal/batchmanager/DefaultSqsAsyncBatchManager.java
```diff
@@ -0,0 +1,71 @@
...
+import software.amazon.awssdk.services.sqs.SqsAsyncClient;
+import software.amazon.awssdk.services.sqs.batchmanager.SqsAsyncBatchManager;
+import software.amazon.awssdk.services.sqs.internal.batchmanager.BatchOverrideConfiguration;
+
+@SdkInternalApi
+public final class DefaultSqsAsyncBatchManager implements SqsAsyncBatchManager {
+    private final SqsAsyncClient client;
+
+    private DefaultSqsAsyncBatchManager(DefaultBuilder builder) {
+        this.client = builder.client;
```
cenedhryn, line 29: What validations would be needed here? I think this is where the scope of this PR stops, but let's keep in mind for the next PR.

services/sqs/src/main/java/software/amazon/awssdk/services/sqs/internal/batchmanager/BatchOverrideConfiguration.java
```diff
@@ -0,0 +1,192 @@
...
+        private Integer maxBatchItems;
+        private Integer maxBatchKeys;
+        private Integer maxBufferSize;
+        private Duration maxBatchOpenInMs;
+
+        private Builder() {
+        }
+
+        /**
+         * Define the the maximum number of messages that are batched together in a single request.
```
cenedhryn, line 140: Is there a maximum number for the maximum number? What is the default?

services/sqs/src/main/java/software/amazon/awssdk/services/sqs/internal/batchmanager/BatchOverrideConfiguration.java
```diff
@@ -0,0 +1,192 @@
...
+         * @return This object for method chaining.
+         */
+        public Builder maxBatchItems(Integer maxBatchItems) {
+            this.maxBatchItems = maxBatchItems;
+            return this;
+        }
+
+        /**
+         * Define the maximum number of batchKeys to keep track of. A batchKey determines which requests are batched together
+         * and is calculated by the client based on the information in a request.
```
cenedhryn, line 152: Is there an upper bound? What is the default?

services/sqs/src/main/java/software/amazon/awssdk/services/sqs/internal/batchmanager/BatchOverrideConfiguration.java
```diff
@@ -0,0 +1,192 @@
...
+         * @param maxBatchKeys the new maxBatchKeys value.
+         * @return This object for method chaining.
+         */
+        public Builder maxBatchKeys(Integer maxBatchKeys) {
+            this.maxBatchKeys = maxBatchKeys;
+            return this;
+        }
+
+        /**
+         * Define the maximum number of items to allow to be buffered for each batchKey.
```
cenedhryn, line 166: Is there an upper bound? What is the default?

services/sqs/src/main/java/software/amazon/awssdk/services/sqs/internal/batchmanager/BatchOverrideConfiguration.java
```diff
@@ -0,0 +1,192 @@
...
+         * @return This object for method chaining.
+         */
+        public Builder maxBufferSize(Integer maxBufferSize) {
+            this.maxBufferSize = maxBufferSize;
+            return this;
+        }
+
+        /**
+         * The maximum amount of time (in milliseconds) that an outgoing call waits for other requests before sending out a batch
+         * request.
```
cenedhryn, line 178: What is the default?

services/sqs/src/main/java/software/amazon/awssdk/services/sqs/internal/batchmanager/BatchOverrideConfiguration.java
```diff
@@ -0,0 +1,192 @@
...
+         *
+         * @param maxBatchOpenInMs The new maxBatchOpenInMs value.
+         * @return This object for method chaining.
+         */
+        public Builder maxBatchOpenInMs(Duration maxBatchOpenInMs) {
```
cenedhryn, line 183: The most common term to denote milliseconds is `Millis`. Can we change?

zoewangg:
Can we use the existing `additionalBuilderMethods` customization config if we are going to handwrite the sqs batch manager anyway?

https://github.com/aws/aws-sdk-java-v2/blob/master/services/s3/src/main/resources/codegen-resources/customization.config#L291

services/sqs/src/main/java/software/amazon/awssdk/services/sqs/batchmanager/SqsAsyncBatchManager.java
```diff
@@ -0,0 +1,134 @@
...
+
+    /**
+     * Buffers and batches {@link SendMessageRequest}s, sending them as a
+     * {@link software.amazon.awssdk.services.sqs.model.SendMessageBatchRequest}. Requests are grouped by queue URL and override
+     * configuration, and sent when the batch size or timeout is reached.
+     *
+     * @param request The SendMessageRequest to be buffered.
+     * @return CompletableFuture of the corresponding {@link SendMessageResponse}.
+     */
+    default CompletableFuture<SendMessageResponse> sendMessage(SendMessageRequest request) {
```
zoewangg, line 60: Probably out of scope of this PR, but we should probably add consumer builder overload (can be addressed in separate PR)

services/sqs/src/main/java/software/amazon/awssdk/services/sqs/internal/batchmanager/BatchOverrideConfiguration.java
```diff
@@ -0,0 +1,192 @@
...
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.services.sqs.internal.batchmanager;
```
zoewangg, line 16: We should move this out of internal package.

services/sqs/src/main/java/software/amazon/awssdk/services/sqs/internal/batchmanager/BatchOverrideConfiguration.java
zoewangg: Or just remove `InMs` since it's a Duration ;) We should probably discuss naming in API surface area review because I'm not sure what maxBatchOpen means...

services/sqs/src/main/java/software/amazon/awssdk/services/sqs/batchmanager/SqsAsyncBatchManager.java
joviegas: Good point , will mention it as TODO

services/sqs/src/main/java/software/amazon/awssdk/services/sqs/batchmanager/SqsAsyncBatchManager.java
```diff
@@ -0,0 +1,134 @@
...
+         * by the caller.
+         *
+         * @param client The SqsAsyncClient to use.
+         * @return This builder for method chaining.
+         */
+        Builder client(SqsAsyncClient client);
+
+        /**
+         * Sets a custom {@link ScheduledExecutorService} for periodic buffer flushes. This executor must be closed by the
+         * caller.
```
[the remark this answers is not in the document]
joviegas: Added TODO in line number 38 for class level

codegen/src/main/java/software/amazon/awssdk/codegen/poet/client/AsyncClientClass.java
```diff
@@ -155,6 +157,10 @@ protected void addFields(TypeSpec.Builder type) {
             type.addField(AwsJsonProtocolFactory.class, "jsonProtocolFactory", PRIVATE, FINAL);
         }
 
+        if (model.getCustomizationConfig().getBatchManagerSupported() || model.hasWaiters()) {
```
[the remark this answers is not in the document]
joviegas: done

services/sqs/src/main/java/software/amazon/awssdk/services/sqs/internal/batchmanager/BatchOverrideConfiguration.java
joviegas: added a class level TODO when doing actual implementaton of this class

joviegas:
> Can we use the existing `additionalBuilderMethods` customization config if we are going to handwrite the sqs batch manager anyway?
> 
> https://github.com/aws/aws-sdk-java-v2/blob/master/services/s3/src/main/resources/codegen-resources/customization.config#L291

These builder Methods are Static APIs on the Client , where as the BatchManager API are non-static methods which are invoked on the client , where client internally uses "this" to build  batch Manager.

cenedhryn (approved):

PR #5327, opened 2024-06-24
Fix SonarCloud Issues

davidh44:
## Motivation and Context

Fix SonarCloud issues


codegen/src/main/java/software/amazon/awssdk/codegen/naming/DefaultNamingStrategy.java
```diff
@@ -272,7 +272,7 @@ public String getEnumValueName(String enumValue) {
         String result = enumValue;
 
         // Special cases
-        result = result.replaceAll("textORcsv", "TEXT_OR_CSV");
+        result = result.replace("textORcsv", "TEXT_OR_CSV");
```
cenedhryn, line 275: Is this safe?

core/auth-crt/src/main/java/software/amazon/awssdk/authcrt/signer/internal/SigningUtils.java
```diff
@@ -81,7 +82,7 @@ public static Clock getSigningClock(ExecutionAttributes executionAttributes) {
 
         Clock baseClock = Clock.systemUTC();
         Optional<Integer> timeOffset = Optional.ofNullable(executionAttributes.getAttribute(
-            AwsSignerExecutionAttribute.TIME_OFFSET));
+            SdkExecutionAttribute.TIME_OFFSET));
```
cenedhryn, line 85: Is this safe?

core/sdk-core/src/test/java/utils/SdkSubscriberTest.java
```diff
@@ -146,10 +146,10 @@ public void limiting_subscriber_with_multiple_thread_publishers() throws Interru
         final int limitFactor = 5;
         LimitingSubscriber<Integer> limitingSubscriber = new LimitingSubscriber<>(mockSubscriber, limitFactor);
         limitingSubscriber.onSubscribe(new EmptySubscription(mockSubscriber));
-        final ExecutorService executorService = Executors.newFixedThreadPool(10);
+        ExecutorService executorService = Executors.newFixedThreadPool(10);
         for (int i = 0; i < 10; i++) {
-            final Integer integer = Integer.valueOf(i);
-            executorService.submit(() -> limitingSubscriber.onNext(new Integer(integer)));
+            int integer = i;
+            executorService.submit(() -> limitingSubscriber.onNext(integer));
```
cenedhryn, line 152: interesting, does the call to onNext actually autobox the int after the change? 

Otherwise if we want to submit an Integer we can remove the extra call and do 

```
Integer boxed = Integer.valueOf(i);
executorService.submit(() -> limitingSubscriber.onNext(boxed));
```

codegen/src/main/java/software/amazon/awssdk/codegen/naming/DefaultNamingStrategy.java
davidh44: yup, `replaceAll` is for regex, if chars then `replace` is less expensive

core/sdk-core/src/test/java/utils/SdkSubscriberTest.java
davidh44: Tests are passing, but updated to
`Integer boxed = i;`

cenedhryn (approved):

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/signer/io/InMemoryPublisher.java
```diff
@@ -64,8 +64,8 @@ public void request(long n) {
                 try {
                     demand.addAndGet(n);
                     fulfillDemand();
-                } catch (Throwable t) {
-                    finish(() -> s.onError(t));
+                } catch (Exception e) {
```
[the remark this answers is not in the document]
davidh44: reverted back to Throwable

zoewangg (approved):

PR #5331, opened 2024-06-25
Wrap region string with Region.of

zoewangg:
## Motivation and Context
Wrap region string with Region.of
```
SqsClient.builder()
         .region("us-west-2")
         .build();
 ```
to
```
SqsClient.builder()
         .region(Region.of("us-west-2"))
         .build();
```

### DIFF
--- a/migration-tool/src/main/java/software/amazon/awssdk/migration/internal/recipe/WrapSdkClientBuilderRegionStr.java
+++ b/migration-tool/src/main/java/software/amazon/awssdk/migration/internal/recipe/WrapSdkClientBuilderRegionStr.java
@@ -1,0 +1,121 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: migration-tool/src/main/java/software/amazon/awssdk/migration/internal/recipe/WrapSdkClientBuilderRegionStr.java [hunk 1]
```diff
@@ -0,0 +1,121 @@
...
+
+            List<JavaType> parameterTypes = methodType.getParameterTypes();
+            if (parameterTypes.size() != 1 || !TypeUtils.isString(parameterTypes.get(0))) {
+                return method;
+            }
+
+            String regionFqcn = Region.class.getCanonicalName();
+
+            JavaTemplate template = JavaTemplate
+                .builder("Region.of(#{any()})")
```
zoewangg, line 97: No unit tests are created for this class because OpenRewrite testing framework can't recognize types from the same project if JavaTemplate is used.

davidh44 (approved):

PR #5339, opened 2024-06-26
Make sure that override plays well for retry policies and strategies

sugmanue:
## Motivation and Context

Currently this bug mostly affects Kinesis but might also affect other clients when there's an override for retry policy followed by an override for a retry strategy. 

Kinesis has a custom `RetryPolicy` that was not converted to a `RetryStrategy` due to their use of non-supported features, in particular they need to get from the request the specific operation to avoid retries on `subscribeToShard` calls. This is not supported by `RetryStrategy` that does not know anything about requests. 

To keep backwards compatibility, the retry stages look for a retry policy and if found it takes precedence over retry strategies. This means that if any user overrides a retry policy and then overrides a retry strategy the retry policy will still be used.

For the particular case of Kinesis this happens by default, i.e., the service builders will create the default retry policy which cannot be overridden for anything else but another retry policy. 

This change clears a previously existing retry policy if there's a retry strategy override, thus making the override work as expected.

## Types of changes

## Checklist

## License

### DIFF
--- a/codegen/src/main/java/software/amazon/awssdk/codegen/poet/builder/BaseClientBuilderClass.java
+++ b/codegen/src/main/java/software/amazon/awssdk/codegen/poet/builder/BaseClientBuilderClass.java
@@ -403,54 +403,55 @@ public class BaseClientBuilderClass implements ClassSpec {
         builder.addStatement("return result.build()")
                .addCode("});");
 
-        builder.addCode("builder.option($1T.EXECUTION_INTERCEPTORS, interceptors)", SdkClientOption.class);
+        builder.addStatement("builder.option($1T.EXECUTION_INTERCEPTORS, interceptors)", SdkClientOption.class);
 
         if (model.getCustomizationConfig().getServiceConfig().hasDualstackProperty()) {
-            builder.addCode(".option($T.DUALSTACK_ENDPOINT_ENABLED, finalServiceConfig.dualstackEnabled())",
+            builder.addStatement("builder.option($T.DUALSTACK_ENDPOINT_ENABLED, finalServiceConfig.dualstackEnabled())",
                             AwsClientOption.class);
         }
 
         if (model.getCustomizationConfig().getServiceConfig().hasFipsProperty()) {
-            builder.addCode(".option($T.FIPS_ENDPOINT_ENABLED, finalServiceConfig.fipsModeEnabled())", AwsClientOption.class);
+            builder.addStatement("builder.option($T.FIPS_ENDPOINT_ENABLED, finalServiceConfig.fipsModeEnabled())",
+                                 AwsClientOption.class);
         }
 
         if (model.getEndpointOperation().isPresent()) {
-            builder.addCode(".option($T.ENDPOINT_DISCOVERY_ENABLED, endpointDiscoveryEnabled)\n",
+            builder.addStatement("builder.option($T.ENDPOINT_DISCOVERY_ENABLED, endpointDiscoveryEnabled)\n",
                             SdkClientOption.class);
         }
 
 
-        if (StringUtils.isNotBlank(model.getCustomizationConfig().getCustomRetryPolicy())) {
-            builder.addCode(".option($1T.RETRY_POLICY, $2T.resolveRetryPolicy(config))",
-                            SdkClientOption.class,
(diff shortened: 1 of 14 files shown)


codegen/src/main/java/software/amazon/awssdk/codegen/poet/builder/BaseClientBuilderClass.java
```diff
@@ -403,54 +403,55 @@ private MethodSpec finalizeServiceConfigurationMethod() {
...
-
         if (StringUtils.isNotBlank(model.getCustomizationConfig().getCustomRetryStrategy())) {
-            builder.addCode(".option($1T.RETRY_STRATEGY, $2T.resolveRetryStrategy(config))",
+            builder.addStatement("builder.option($1T.RETRY_STRATEGY, $2T.resolveRetryStrategy(config))",
                             SdkClientOption.class,
                             PoetUtils.classNameFromFqcn(model.getCustomizationConfig().getCustomRetryStrategy()));
         }
 
+        if (StringUtils.isNotBlank(model.getCustomizationConfig().getCustomRetryPolicy())) {
+            builder.beginControlFlow("if (builder.option($T.RETRY_STRATEGY) == null)", SdkClientOption.class);
```
sugmanue, line 431: `setOverrides` happens before this methods is called, if there's a retry strategy override we need to avoid setting the default client retry policy.

services/kinesis/src/test/java/software/amazon/awssdk/services/kinesis/KinesisClientCanOverrideDefaultRetryPolicy.java
```diff
@@ -0,0 +1,139 @@
...
+import org.junit.jupiter.api.Test;
+import software.amazon.awssdk.auth.credentials.AwsBasicCredentials;
+import software.amazon.awssdk.auth.credentials.StaticCredentialsProvider;
+import software.amazon.awssdk.core.SdkPlugin;
+import software.amazon.awssdk.core.retry.RetryMode;
+import software.amazon.awssdk.services.kinesis.model.SubscribeToShardEventStream;
+import software.amazon.awssdk.services.kinesis.model.SubscribeToShardRequest;
+import software.amazon.awssdk.services.kinesis.model.SubscribeToShardResponseHandler;
+
+public class KinesisClientCanOverrideDefaultRetryPolicy {
```
zoewangg, line 41: Maybe add a Test suffix to the class name?

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/builder/sra/test-client-builder-class.java
```diff
@@ -155,12 +155,14 @@ protected final SdkClientConfiguration finalizeServiceConfiguration(SdkClientCon
...
-                .option(AwsClientOption.FIPS_ENDPOINT_ENABLED, finalServiceConfig.fipsModeEnabled())
-                .option(SdkClientOption.RETRY_POLICY, MyServiceRetryPolicy.resolveRetryPolicy(config))
-                .option(SdkClientOption.RETRY_STRATEGY, MyServiceRetryStrategy.resolveRetryStrategy(config))
-                .option(SdkClientOption.SERVICE_CONFIGURATION, finalServiceConfig);
+        builder.option(SdkClientOption.EXECUTION_INTERCEPTORS, interceptors);
+        builder.option(AwsClientOption.DUALSTACK_ENDPOINT_ENABLED, finalServiceConfig.dualstackEnabled());
+        builder.option(AwsClientOption.FIPS_ENDPOINT_ENABLED, finalServiceConfig.fipsModeEnabled());
+        builder.option(SdkClientOption.RETRY_STRATEGY, MyServiceRetryStrategy.resolveRetryStrategy(config));
+        if (builder.option(SdkClientOption.RETRY_STRATEGY) == null) {
+            builder.option(SdkClientOption.RETRY_POLICY, MyServiceRetryPolicy.resolveRetryPolicy(config));
```
zoewangg, line 163: I understand why we are making this formatting change, but it'd be great if we can separate refactoring and actual code change to different PRs  in the future to make it easier to view the diff.

zoewangg (approved):

PR #5345, opened 2024-06-28
Allow slow recovery of tokens in the token bucket

sugmanue:
## Motivation and Context

The original implementation of the token bucket logic did not deposit back for successful invocations which means that once the buckets were exhausted it will no allow any more retries since there's no mechanism to recover from that state.

Customers using custom **`RetryPolicy`** instances or **Kinesis** are NOT affected by this bug.

## Types of changes

## Checklist

## License

### DIFF
--- a/core/retries/src/main/java/software/amazon/awssdk/retries/internal/circuitbreaker/TokenBucket.java
+++ b/core/retries/src/main/java/software/amazon/awssdk/retries/internal/circuitbreaker/TokenBucket.java
@@ -89,8 +89,7 @@ public final class TokenBucket {
                            .maxCapacity(maxCapacity);
 
         if (amountToRelease == 0) {
-            return builder.currentCapacity(capacity.get())
-                          .build();
+            amountToRelease = 1;
         }
 
         int currentCapacity;
--- a/core/retries/src/main/java/software/amazon/awssdk/retries/internal/circuitbreaker/TokenBucket.java
+++ b/core/retries/src/main/java/software/amazon/awssdk/retries/internal/circuitbreaker/TokenBucket.java
@@ -89,7 +89,8 @@ public final class TokenBucket {
                            .maxCapacity(maxCapacity);
 
         if (amountToRelease == 0) {
-            amountToRelease = 1;
+            return builder.currentCapacity(capacity.get())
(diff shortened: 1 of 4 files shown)


core/retries/src/main/java/software/amazon/awssdk/retries/internal/circuitbreaker/TokenBucket.java
```diff
@@ -89,8 +89,7 @@ public ReleaseResponse release(int amountToRelease) {
                            .maxCapacity(maxCapacity);
 
         if (amountToRelease == 0) {
-            return builder.currentCapacity(capacity.get())
-                          .build();
+            amountToRelease = 1;
```
millems, line 92: Why are we releasing zero? Should the fix be in the caller instead of here?

sugmanue: Could be the the caller, no particular reason. We release zero on a successful first attempt since we have not withdrawn any tokens.

millems: Imo, doing something different from what's asked in the token bucket is probably not the move. Calling release(0) and it transforming to release(1) is confusing.

millems: Calling release(1) on success even without removing anything is fine. It's not a zero-sum game. The token bucket enforces a max on capacity, and we already remove more than we add for a given request, so adding more than we remove is a fine action as well.

sugmanue: We don't know at success if we withdraw or not, so it needs to be something along the lines of 

```
release(Math.max(1, withdrawn))
```

Let me know if that looks better.

sugmanue: I pushed a commit with the alternative.

millems (approved):

PR #5348, opened 2024-06-30
Add throttling backoff to all retry strategies

sugmanue:
## Types of changes

## Checklist

## License

### DIFF
--- a/core/retries-spi/src/main/java/software/amazon/awssdk/retries/api/RetryStrategy.java
+++ b/core/retries-spi/src/main/java/software/amazon/awssdk/retries/api/RetryStrategy.java
@@ -212,7 +212,9 @@ public interface RetryStrategy {
         B throttlingBackoffStrategy(BackoffStrategy throttlingBackoffStrategy);
 
         /**
-         * Configure the predicate to allow the strategy categorize a Throwable as throttling exception.
+         * Configure a predicate that determines whether a retryable exception is a throttling exception. When this predicate
+         * returns true, the retry strategy will use the {@link #throttlingBackoffStrategy}. If it returns false, the
+         * {@link #backoffStrategy} will be used. This predicate will not be called for non-retryable exceptions.
          */
         B treatAsThrottling(Predicate<Throwable> treatAsThrottling);
 
--- a/core/retries-spi/src/main/java/software/amazon/awssdk/retries/api/RetryStrategy.java
+++ b/core/retries-spi/src/main/java/software/amazon/awssdk/retries/api/RetryStrategy.java
@@ -188,7 +188,7 @@ public interface RetryStrategy {
         }
 
         /**
-         * Configure the maximum number of attempts used by this executor.
+         * Configure the maximum number of attempts used by the retry strategy.
          *
          * <p>The actual number of attempts made may be less, depending on the retry strategy implementation. For example, the
          * standard and adaptive retry modes both employ short-circuiting which reduces the maximum attempts during outages.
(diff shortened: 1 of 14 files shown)


core/retries-spi/src/main/java/software/amazon/awssdk/retries/api/RetryStrategy.java
```diff
@@ -198,12 +198,24 @@ default B retryOnRootCauseInstanceOf(Class<? extends Throwable> throwable) {
...
+         * Configure the backoff strategy used for throttling exceptions by this strategy.
+         *
+         * <p>By default, this uses jittered exponential backoff.
+         */
+        B throttlingBackoffStrategy(BackoffStrategy throttlingBackoffStrategy);
+
+        /**
+         * Configure the predicate to allow the strategy categorize a Throwable as throttling exception.
+         */
+        B treatAsThrottling(Predicate<Throwable> treatAsThrottling);
```
millems, line 217: Can we clarify in the docs that this will only get called if the `Throwable` is considered retryable? Right now it's not clear when this will be used.

Maybe 
> Configure a predicate that determines whether a retryable exception is a throttling exception. When this predicate returns true, the retry strategy will use the `throttlingBackoffStrategy`. If it returns false, the `backoffStrategy` will be used. This predicate will not be called for non-retryable exceptions.

core/retries/src/main/java/software/amazon/awssdk/retries/AdaptiveRetryStrategy.java
```diff
@@ -70,8 +70,6 @@ static AdaptiveRetryStrategy.Builder builder() {
                                               .tokenBucketMaxCapacity(DefaultRetryStrategy.Standard.TOKEN_BUCKET_SIZE)
                                               .build())
             .tokenBucketExceptionCost(DefaultRetryStrategy.Standard.DEFAULT_EXCEPTION_TOKEN_COST)
-            .backoffStrategy(BackoffStrategy.exponentialDelay(DefaultRetryStrategy.Standard.BASE_DELAY,
-                                                              DefaultRetryStrategy.Standard.MAX_BACKOFF))
```
millems, line 74: Why this change? Just switching to the defaults?

core/retries-spi/src/main/java/software/amazon/awssdk/retries/api/RetryStrategy.java
millems, line 217: I'm worried about baking the concept of "throttling exceptions" into our base retry strategy API. Is it a better or worse API if we allow them to specify a `Function<Throwable, BackoffStrategy>` for having an exception-level override for backoff strategies for any reason, instead of making this API specific to throttling exceptions?

i.e. a "use this backoff strategy for this exception type, overriding the default".  We'd use it for throttling exceptions:

```
throwable -> isThrottlingException(throwable) ? throttlingBackoffStrategy : null; // null = use default  backoff strategy
```

But it would be more flexible because it could be used to override the backoff strategy for IOExceptions or DynamoDbExceptions, etc.

millems: I'm imagining just one new API instead of two:

```java
/**
 * Specify a function that overrides the backoff strategy for a particular exception type. When a throwable is considered retryable, the SDK will use the default backoff strategy unless this function returns a different one to use.
 */
B overrideBackoffStrategy(Function<Throwable, BackoffStrategy> overrideBackoffStrategyFunction);
```

millems: Discussed offline. We already have methods specific to throttling in adaptive and legacy, so this is a door we've mostly already walked through. We'll use these two methods everywhere and if we ever want to make it more flexible, we can do that then instead of trying to do it now.

millems (approved):

PR #5358, opened 2024-07-02
Consume any pending bytes on socket

dagnir:
This update checks the wrapped socket to see if there is any data available on the socket. If this is true, then there is (at least a partial) TLS record that is waiting to be read. Other than application data, this record may contain a close_notify.

If data is available, then we perform a single-byte read on the SSL socket to get the SSLSocket to consume the record.

## Types of changes

## Checklist

## License

### DIFF
--- a/http-clients/apache-client/src/main/java/software/amazon/awssdk/http/apache/internal/net/InputShutdownCheckingSslSocket.java
+++ b/http-clients/apache-client/src/main/java/software/amazon/awssdk/http/apache/internal/net/InputShutdownCheckingSslSocket.java
@@ -15,11 +15,17 @@
 
 package software.amazon.awssdk.http.apache.internal.net;
 
+import java.io.FilterInputStream;
 import java.io.FilterOutputStream;
 import java.io.IOException;
+import java.io.InputStream;
 import java.io.OutputStream;
+import java.net.Socket;
+import java.net.SocketTimeoutException;
+import java.nio.ByteBuffer;
 import javax.net.ssl.SSLSocket;
 import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.utils.Logger;
 
 /**
  * Wrapper socket that ensures the read end of the socket is still open before performing a {@code write()}. In TLS 1.3, it is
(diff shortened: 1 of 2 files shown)


http-clients/apache-client/src/main/java/software/amazon/awssdk/http/apache/internal/net/InputShutdownCheckingSslSocket.java
```diff
@@ -28,22 +34,114 @@
...
+        synchronized (readBuffer) {
+            if (!readBuffer.hasRemaining() || !dataAvailableOnLayeredSocket()) {
+                return;
+            }
+
+            InputStream is = sslSocket.getInputStream();
+            int originalSoTimeout = sslSocket.getSoTimeout();
+            try {
+                // Maximum time to block for the least amount of time possible (1ms)
+                sslSocket.setSoTimeout(1);
```
dagnir, line 70: The hope is that `bufferSslSocketRead()` on the whole doesn't get called much. This should only happen when data arrives as we're writing to the socket which shouldn't be a common occurrence at least in HTTP.

http-clients/apache-client/src/main/java/software/amazon/awssdk/http/apache/internal/net/InputShutdownCheckingSslSocket.java
```diff
@@ -65,17 +163,26 @@ public void write(byte[] b, int off, int len) throws IOException {
...
-            if (sock.isInputShutdown()) {
+            if (sslSocket.isInputShutdown()) {
                 throw new IOException("Remote end is closed.");
             }
 
             try {
-                sock.getInputStream();
+                if (dataAvailableOnLayeredSocket()) {
+                    LOG.debug(() -> String.format("%d bytes available in layered socket, calling SSLSocket's read() to try and "
+                                                 + "consume the TLS record(s) if present."));
```
millems, line 173: %d was not specified.

http-clients/apache-client/src/main/java/software/amazon/awssdk/http/apache/internal/net/InputShutdownCheckingSslSocket.java
```diff
@@ -28,22 +34,114 @@
...
+            } catch (SocketTimeoutException to) {
+                LOG.debug(() -> "Timeout doing single byte read on SSLSocket. Most likely due to incomplete record", to);
+            } finally {
+                sslSocket.setSoTimeout(originalSoTimeout);
+            }
+        }
+    }
+
+    private class BufferedInputStream extends FilterInputStream {
+        private final byte[] oneByte = new byte[1];
```
millems, line 84: `LookaheadInputStream` from `utils` does this same thing - allow peeking one byte in advance. Not sure if it's possible to reuse that, but would be nice to avoid duplication if possible.

http-clients/apache-client/src/main/java/software/amazon/awssdk/http/apache/internal/net/InputShutdownCheckingSslSocket.java
```diff
@@ -65,17 +163,26 @@ public void write(byte[] b, int off, int len) throws IOException {
...
             }
 
             try {
-                sock.getInputStream();
+                if (dataAvailableOnLayeredSocket()) {
+                    LOG.debug(() -> String.format("%d bytes available in layered socket, calling SSLSocket's read() to try and "
+                                                 + "consume the TLS record(s) if present."));
+                    bufferSslSocketRead();
+                }
+                sslSocket.getInputStream();
```
millems, line 176: Is it possible to write a unit test for this?

dagnir: Should be; essentially just flood the socket with data without ever performing a socket reads.

http-clients/apache-client/src/main/java/software/amazon/awssdk/http/apache/internal/net/InputShutdownCheckingSslSocket.java
dagnir: Oops! That's probably the build failures 😅

PR #5360, opened 2024-07-02
Update Spotbugs suppressions

davidh44:
## Motivation and Context
Upgrade to Spotbugs `4.7.3.5` flagged new bugs which we suppressed.

These rules have many false positives that we can safely ignore
`EI_EXPOSE_REP` , `EI_EXPOSE_REP2` , `MS_EXPOSE_REP` , `MC_OVERRIDABLE_METHOD_CALL_IN_CONSTRUCTOR`

Filtered out flagged classes for these rules which are determined to be false positives
`SA_FIELD_SELF_ASSIGNMENT`, `DCN_NULLPOINTER_EXCEPTION`


cenedhryn (commented):
What happens if I write a PR containing one of the bugs that we suppress in general with this PR? We'll never catch it?

davidh44:
> What happens if I write a PR containing one of the bugs that we suppress in general with this PR? We'll never catch it?

Yeah we won't catch it, will rely on code reviews. The flagged bugs are false positives and harmless so seems okay. Plus not feasible for the 2 rules that have hundreds of flagged bugs

cenedhryn (approved):

PR #5364, opened 2024-07-03
Skip transformation for S3 Transfer Manager and DynamoDB mapper

zoewangg:
## Motivation and Context
- Skip transformation for S3 Transfer Manager and DynamoDB mapper since they are not supported in the initial release.
- Only add HTTP client dependencies if `ClientConfiguration` is directly being used (http settings)

### DIFF
--- a/migration-tool/src/test/java/software/amazon/awssdk/migration/recipe/ChangeSdkTypeTest.java
+++ b/migration-tool/src/test/java/software/amazon/awssdk/migration/recipe/ChangeSdkTypeTest.java
@@ -28,7 +28,8 @@ public class ChangeSdkTypeTest implements RewriteTest {
 
     @Override
     public void defaults(RecipeSpec spec) {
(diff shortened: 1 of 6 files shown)


migration-tool/src/test/java/software/amazon/awssdk/migration/recipe/ChangeSdkTypeTest.java
```diff
@@ -215,4 +216,30 @@ void shouldChangeFieldsInInnerClass() {
             )
         );
     }
+
+    @Test
+    @EnabledOnJre({JRE.JAVA_8})
+    void hasUnsupportedFeature_shouldSkip() {
```
davidh44, line 222: nit : TODO to update/replace test

davidh44 (approved):

migration-tool/src/test/java/software/amazon/awssdk/migration/recipe/ChangeSdkTypeTest.java
zoewangg: Test will fail right away once we add support :)

PR #5365, opened 2024-07-04
Correctly consider defaultRetryMode customization

L-Applin:
Update default retry mode configuration only if not already defined by using `lazyOptionIfAbsent` instead of `lazyOption`

## Motivation and Context
The `defaultRetryMode` defined in `customization.config` file would be overridden by the `resolveDefaultRetryMode` method.

## Modifications
Call `lazyOptionIfAbsent` instead of `lazyOption` for `DEFAULT_RETRY_MODE`

## Testing
Added codegen test to validate correct default retry mode

## Types of changes


core/aws-core/src/main/java/software/amazon/awssdk/awscore/client/builder/AwsDefaultClientBuilder.java
```diff
@@ -176,7 +176,7 @@ private SdkClientConfiguration finalizeAwsConfiguration(SdkClientConfiguration c
                             .lazyOptionIfAbsent(AwsClientOption.DUALSTACK_ENDPOINT_ENABLED, this::resolveDualstackEndpointEnabled)
                             .lazyOptionIfAbsent(AwsClientOption.FIPS_ENDPOINT_ENABLED, this::resolveFipsEndpointEnabled)
                             .lazyOption(AwsClientOption.DEFAULTS_MODE, this::resolveDefaultsMode)
-                            .lazyOption(SdkClientOption.DEFAULT_RETRY_MODE, this::resolveDefaultRetryMode)
+                            .lazyOptionIfAbsent(SdkClientOption.DEFAULT_RETRY_MODE, this::resolveDefaultRetryMode)
```
joviegas, line 179: Do we remember the reason why we initially chose to do a `lazyOption` ?

L-Applin: Matt would have the background on this, I don't

millems: I don't recall.

joviegas (approved):

PR #5368, opened 2024-07-05
Add gradle support and add tests for gradle project

zoewangg:
## Motivation and Context
Add support for gradle project

## Modifications
-  Replace `org.openrewrite.maven.AddDependency` with `org.openrewrite.java.dependencies.AddDependency`, which handles dependencies update for both Maven and Gradle
- Add end-to-end tests for Gradle project
- Bump OR version


migration-tool/src/test/java/software/amazon/awssdk/migration/recipe/UpgradeSdkDependenciesTest.java
```diff
@@ -85,7 +85,6 @@ private static String getVersion() throws IOException {
     @Test
```
zoewangg, line 85: I tried to add unit tests for gradle but the testing framework requires gradle to run it :(

migration-tool/pom.xml
```diff
@@ -32,10 +32,18 @@
 
     <dependencyManagement>
         <dependencies>
+            <!-- We should not use bom-internal here because it'll cause version conflict -->
```
millems, line 35: Can we align on a single version?

migration-tool/src/test/java/software/amazon/awssdk/migration/recipe/UpgradeSdkDependenciesTest.java
millems: We could just invoke `gradle` from the unit test! (Gross, but it would work, right?)

millems (approved):

migration-tool/pom.xml
zoewangg: Yeah, all the dependent SDK modules use the same version as this module, or do you mean something else?

millems: I guess I don't understand why bom-internal would cause version conflicts, then.

zoewangg: Some of the dependencies such as jackson and Junit defined in our bom-internal conflict with those defined in `rewrite-recipe-bom`.

PR #5376, opened 2024-07-08
Add new BackoffStrategy in waiters

sugmanue:
Add the new BackoffStretegy to the Waiter polling logic

## Motivation and Context
Waiters were still using the deprecated BackoffStrategy, this PR is to remove the usage of the deprecated classes.

## Modifications
- Modified codegen waiter classes to fetch the new BackoffStrategy from the Waiter configuration.
- Updated waiter configuration to be able to retrieve the new BackoffStrategy

## Testing
- Unit test
- Codegen tests

## Types of changes


core/sdk-core/src/main/java/software/amazon/awssdk/core/waiters/WaiterOverrideConfiguration.java
```diff
@@ -163,4 +197,36 @@ public WaiterOverrideConfiguration build() {
             return new WaiterOverrideConfiguration(this);
         }
     }
+
+    @SdkInternalApi
+    static class LegacyToNonLegacyAdapter implements BackoffStrategy {
```
zoewangg, line 202: We should probably move internal API outside of a public API

core/sdk-core/src/main/java/software/amazon/awssdk/core/waiters/WaiterOverrideConfiguration.java
```diff
@@ -55,9 +57,25 @@ public Optional<Integer> maxAttempts() {
     }
 
     /**
-     * @return the optional {@link BackoffStrategy} that should be used when polling the resource
+     * @return the optional {@link software.amazon.awssdk.core.retry.backoff.BackoffStrategy} that should be used when polling
+     * the resource
+     * @deprecated Use instead {@link #backoffStrategy2()}
      */
-    public Optional<BackoffStrategy> backoffStrategy() {
+    public Optional<software.amazon.awssdk.core.retry.backoff.BackoffStrategy> backoffStrategy() {
```
zoewangg, line 64: Can we move the logic to ctor? Should we validate that only of them is specified?

sugmanue: It the new backoff strategy is given, what would this one return in that case? (it's just a cosmetic issue, this class while public, is not actually used by customers, rather is just used internally for the waiter executor) You could return  an empty value in this method for that case (not sure if that might confuse someone but for sure will reduce the complexity of it) and that should be fine. If we go that route we won't have any use for the `NonLegacyToLegacyAdapter` and can also go away.

L-Applin: We cant move it all to the ctor, as we want to keep a reference to the new BackoffStrategy as a class member

zoewangg (approved):
Let's add a changelog entry

PR #5380, opened 2024-07-09
Fixed EnhancedClient UpdateItem operation to make it work on nested attributes as well

anirudh9391:
Modified Enhanced Client to facilitate nested update operations. 

## Motivation and Context
Customer reported an issue where UpdateItem does not work for nested attributes. 

## Modifications
Modified UpdateItem query path to form the right updateQuery for nested update attributes. 
Previously nested update queries looked like 
```
SET #AMZN_MAPPED_mainAddress = :AMZN_MAPPED_mainAddress
:AMZN_MAPPED_mainAddress -> "AttributeValue(M={state=AttributeValue(S=WA)})"
```

This operates like a putItem and updates the map with a new state. 

Intead modified the update query to 
```
aws dynamodb update-item \
    --table-name TEST_TABLE3 \
    --key '{"id":{"N":"111"}}' \
    --region us-west-2 \
    --update-expression "SET #AMZN_MAPPED_mainAddress.#AMZN_MAPPED_city = :AMZN_MAPPED_mainAddress_city" \
    --expression-attribute-values '{":AMZN_MAPPED_mainAddress_city": {"S": "Salem"}}' \  
    --expression-attribute-names '{"#AMZN_MAPPED_mainAddress": "mainAddress","#AMZN_MAPPED_city":"city"}'
```
## Testing
Unit Tests pass

## Types of changes


services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/DynamoDBEnhancedRequestConfiguration.java
```diff
@@ -0,0 +1,36 @@
...
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.enhanced.dynamodb.internal;
```
cenedhryn, line 16: It can't be internal, because it's a public API. I suggest the `mapper` module

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/DynamoDBEnhancedRequestConfiguration.java
```diff
@@ -0,0 +1,36 @@
...
+
+import software.amazon.awssdk.annotations.SdkPublicApi;
+import software.amazon.awssdk.enhanced.dynamodb.mapper.AttributeMapping;
+
+/**
+ * Configuration Object to define behaviour of DynamoDB operations.
+ */
+@SdkPublicApi
+public class DynamoDBEnhancedRequestConfiguration {
```
cenedhryn, line 25: How about naming it `ItemMapConfiguration` or something similar?

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/mapper/AttributeMapping.java
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.enhanced.dynamodb.mapper;
+
+public enum AttributeMapping {
```
cenedhryn, line 18: Missing SDK API annotation

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/TableSchema.java
```diff
@@ -240,8 +241,24 @@ default T mapToItem(Map<String, AttributeValue> attributeMap, boolean preserveEm
      * @param ignoreNulls If set to true; any null values in the Java object will not be added to the output map.
      *                    If set to false; null values in the Java object will be added as {@link AttributeValue} of
      *                    type 'nul' to the output map.
+     * @param requestConfiguration denotes the level of nesting, i.e. SHALLOW or NESTED, contained in the request params based
+     * which nested update query parameters would be dealt with
```
cenedhryn, line 245: Maybe `configuration options for mapping an object to a map with attribute values`?

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/TableSchema.java
```diff
@@ -240,8 +241,24 @@ default T mapToItem(Map<String, AttributeValue> attributeMap, boolean preserveEm
...
+     * This is deprecated in favour of itemToMap(T item, boolean ignoreNulls, DynamoDBEnhancedRequestOverrideConfiguration
+     * ddbRequestOverrideConfiguration) This method takes a modelled object and converts it into a raw map of
+     * {@link AttributeValue} that the DynamoDb low-level SDK can work with.
+     *
+     * @param item        The modelled Java object to convert into a map of attributes.
+     * @param ignoreNulls If set to true; any null values in the Java object will not be added to the output map. If set to false;
+     *                    null values in the Java object will be added as {@link AttributeValue} of type 'nul' to the output map.
+     * @return A map of String to {@link AttributeValue} representing all the modelled attributes in the model object.
+     */
+    @Deprecated
```
cenedhryn, line 261: We have discussed whether we should deprecate older APIs when there are new alternatives, since some applications break when deprecated attributes are detected. In this case I think it's warranted to deprecate, because itemToMap (older version) has a bug.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/TableSchema.java
```diff
@@ -240,8 +241,24 @@ default T mapToItem(Map<String, AttributeValue> attributeMap, boolean preserveEm
      * @param ignoreNulls If set to true; any null values in the Java object will not be added to the output map.
      *                    If set to false; null values in the Java object will be added as {@link AttributeValue} of
      *                    type 'nul' to the output map.
+     * @param requestConfiguration denotes the level of nesting, i.e. SHALLOW or NESTED, contained in the request params based
+     * which nested update query parameters would be dealt with
      * @return A map of String to {@link AttributeValue} representing all the modelled attributes in the model object.
      */
+    Map<String, AttributeValue> itemToMap(T item, boolean ignoreNulls,
+                                          DynamoDBEnhancedRequestConfiguration requestConfiguration);
```
cenedhryn, line 249: Can it be just `configuration`?

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/EnhancedClientUtils.java
```diff
@@ -67,18 +71,30 @@ public static String cleanAttributeName(String key) {
...
 
     /**
      * Creates a value token to be used with an ExpressionValues map.
      */
     public static String valueRef(String value) {
-        return ":AMZN_MAPPED_" + cleanAttributeName(value);
+        String cleanAttributeName = cleanAttributeName(value);
+        cleanAttributeName = isNestedAttribute(cleanAttributeName) ?
+                             PATTERN.matcher(cleanAttributeName).replaceAll("_")
+                                                      : cleanAttributeName;
```
cenedhryn, line 96: Nit: alignment of rows

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/DynamoDBEnhancedRequestConfiguration.java
```diff
@@ -0,0 +1,36 @@
...
+    private final AttributeMapping attributeMapping;
+
+    public DynamoDBEnhancedRequestConfiguration(AttributeMapping attributeMapping) {
+        this.attributeMapping = attributeMapping;
+    }
```
cenedhryn, line 30: Because this is a public API, can we add a Builder for backwards compatibility

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/mapper/annotations/DynamoDbIgnoreNulls.java
```diff
@@ -21,14 +21,16 @@
...
- * stored in the current object should ignore the attributes with null values. Note that if this annotation is absent, NULL
- * attributes will be created.
+ * Specifies that when calling {@link TableSchema#itemToMap(Object, boolean, DynamoDBEnhancedRequestConfiguration)}, a separate
+ * DynamoDB object
+ * that is stored in the current object should ignore the attributes with null values. Note that if this annotation
+ * is absent, NULL attributes will be created.
  *
  * <p>
- * Example using {@link DynamoDbIgnoreNulls}:
+ * Example using {@link DynamoDbIgnoreNulls}:d
```
cenedhryn, line 33: Nit: extra char

cenedhryn (commented):
The PR also needs a changelog item.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/TableSchema.java
anirudh9391: Yep, I too went through the discussion points. I can probably call this out

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/TableSchema.java
anirudh9391: sure i can change this

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/mapper/annotations/DynamoDbIgnoreNulls.java
anirudh9391: oops, will change

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/TableSchema.java
zoewangg: Can we move ignoreNulls to the configuration class since it's part of the configuration?

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/DynamoDBEnhancedRequestConfiguration.java
zoewangg: +1 Let's use builder pattern instead of the ctor. Let's follow the config class convention. https://github.com/aws/aws-sdk-java-v2/blob/master/docs/design/ClientConfiguration.md

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/mapper/AttributeMapping.java
zoewangg: Let's add toString and other methods, see existing enum for reference.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/TableSchema.java
zoewangg: It needs to be a default method that throws UnsupportedOperationException to be backwards compatible.

anirudh9391: Thats a good suggestion. I can do that

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/TableSchema.java
zoewangg: I'd like to understand this a bit more. Let's discuss this as a team in post standup

joviegas:
It will be great if we mention the user experience in the description with these changes  since these changes are  made in Public classes 
Something like 
1. Args with  ignoreNulls as true
2. Args with  ignoreNulls as true and DynamoDBEnhancedRequestConfiguration
and when to use DynamoDBEnhancedRequestConfigurations values  for different value

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/TableSchema.java
```diff
@@ -237,11 +238,25 @@ default T mapToItem(Map<String, AttributeValue> attributeMap, boolean preserveEm
...
+     * This is deprecated in favour of itemToMap(T item, MappingConfiguration
+     * configuration) This method takes a modelled object and converts it into a raw map of
+     * {@link AttributeValue} that the DynamoDb low-level SDK can work with.
+     *
+     * @param item        The modelled Java object to convert into a map of attributes.
+     * @param ignoreNulls If set to true; any null values in the Java object will not be added to the output map. If set to false;
+     *                    null values in the Java object will be added as {@link AttributeValue} of type 'nul' to the output map.
+     * @return A map of String to {@link AttributeValue} representing all the modelled attributes in the model object.
+     */
+    @Deprecated
```
zoewangg, line 259: Can we add `@deprecated` to the javadoc? Example: https://github.com/aws/aws-sdk-java-v2/blob/master/core/sdk-core/src/main/java/software/amazon/awssdk/core/retry/RetryPolicy.java#L46

.changes/next-release/bugfix-AWSSDKforJavav2-2a80679.json
```diff
@@ -0,0 +1,6 @@
+{
+    "type": "bugfix",
+    "category": "AWS SDK for Java v2",
+    "contributor": "anirudh9391",
+    "description": "Modify Enhanced Client to work on successfully updating nested attributes without loss of information"
```
zoewangg, line 5: Can we add more information? Is there an associated GH issue? Suggesting something like below
> Introduce a new `TableSchema#itemMap` that fixes the issue where xxxx. See [xxx](github issue link). The existing xxx is now deprecated in favor of this new API.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/TableSchema.java
```diff
@@ -237,11 +238,25 @@ default T mapToItem(Map<String, AttributeValue> attributeMap, boolean preserveEm
...
      *
      * @param item The modelled Java object to convert into a map of attributes.
-     * @param ignoreNulls If set to true; any null values in the Java object will not be added to the output map.
-     *                    If set to false; null values in the Java object will be added as {@link AttributeValue} of
-     *                    type 'nul' to the output map.
+     * @param configuration configuration options for mapping an object to a map with attribute values
      * @return A map of String to {@link AttributeValue} representing all the modelled attributes in the model object.
      */
+    default Map<String, AttributeValue> itemToMap(T item,
+                                          MappingConfiguration configuration) {
```
zoewangg, line 245: `MappingConfiguration` seems a bit generic.  `mapToItem` doesn't take a mapping configuration. Maybe just `ItemToMapConfiguration`? 

I just noticed there's another overload down below.
```  
Map<String, AttributeValue> itemToMap(T item, Collection<String> attributes);
```

Are we gonna need to add another overload in the future? `Map<String, AttributeValue> itemToMap(T item, Collection<String> attributes, ItemToMapConfiguration configuration);`

This makes me wonder if we should just add an overload that takes a request wrapper class and returns a response wrapper class to be future proof, something like 

```
ItemToMapResponse itemToMap(ItemToMapRequest request);
```

services-custom/dynamodb-enhanced/src/test/java/software/amazon/awssdk/enhanced/dynamodb/mapper/MappingConfigurationTest.java
```diff
@@ -0,0 +1,55 @@
...
+        assertNotNull(configuration);
+    }
+
+    @Test
+    public void mappingConfigurationTest_verify_ToString() {
+        assertEquals("MappingConfiguration(attributeMapping=SHALLOW, ignoreNulls=true)", configuration.toString());
+    }
+
+    @Test
+    public void mappingConfigurationTest_verify_equals() {
```
zoewangg, line 45: nit: we can use `equalsverifier` to verify equals and hashcode. Example: https://github.com/aws/aws-sdk-java-v2/blob/master/services/s3/src/test/java/software/amazon/awssdk/services/s3/crt/S3CrtHttpConfigurationTest.java#L24

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/TableSchema.java
anirudh9391: done

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/TableSchema.java
anirudh9391: ItemToMap sounds good. I can modify that.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/mapper/AttributeMapping.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+package software.amazon.awssdk.enhanced.dynamodb.mapper;
+
+import software.amazon.awssdk.annotations.SdkPublicApi;
+import software.amazon.awssdk.utils.StringUtils;
+
+@SdkPublicApi
+public enum AttributeMapping {
+    SHALLOW,
```
joviegas, line 23: Can we please add some Java docs here  ?

joviegas:
```java
Map<String, AttributeValue> itemMap = beanTableSchema.itemToMap(bean, MappingConfiguration.builder().attributeMapping(AttributeMapping.NESTED).build());
```

I have a concern regarding the use of MappingConfiguration for the NESTED approach as it disrupts the ItemToMap representation of the item. For instance, it adds innerBean_NESTED_OBJECT_UPDATE_attribute2 to the map:

```
itemMap {innerBean_NESTED_OBJECT_UPDATE_attribute2=AttributeValue(S=Hello), id=AttributeValue(NUL=true), nested=AttributeValue(L=[AttributeValue(L=[AttributeValue(S=one), AttributeValue(S=two), AttributeValue(S=three)]), AttributeValue(L=[AttributeValue(S=a), AttributeValue(S=b), AttributeValue(S=b)])])}
```

This addition alters the original ItemToMap representation of the class, and we cannot use itemToMap to directly convert a Bean/Class/Item to AttributeValue Map and insert into DDB.

```java
DynamoDbClient dynamoDbClient = DynamoDbClient.create();

Map<String, AttributeValue> itemMap = beanTableSchema.itemToMap(bean, MappingConfiguration.builder()
    .attributeMapping(AttributeMapping.NESTED)
    .build());
dynamoDbClient.putItem(PutItemRequest.builder().item(itemMap).build());
```

Even though currently we do not have a use case where a customer would use AttributeMapping.NESTED, I believe a customer would never do that. If the user is never going to do that, why provide a public API that can cause disruption in the AttributeValue Map? WDYT

.idea/codeStyles/Project.xml
```diff
@@ -47,4 +47,4 @@
       </indentOptions>
     </codeStyleSettings>
   </code_scheme>
-</component>
\ No newline at end of file
+</component>
```
anirudh9391, line 50: Unsure as to why this shows up. This file was untouched

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/update/UpdateExpressionUtils.java
```diff
@@ -136,9 +139,18 @@ private static Function<String, String> behaviorBasedValue(UpdateBehavior update
     /**
      * Simple utility method that can create an ExpressionNames map based on a list of attribute names.
      */
-    private static Map<String, String> expressionNamesFor(String... attributeNames) {
-        return Arrays.stream(attributeNames)
-                     .collect(Collectors.toMap(EnhancedClientUtils::keyRef, Function.identity()));
-    }
+    private static Map<String, String> expressionNamesFor(String attributeNames) {
```
joviegas, line 142: Minor comment : following the pattern of older code , what if we use below ?
```java
    private static Map<String, String> expressionNamesFor(String attributeNames) {
        if (attributeNames.contains(NESTED_OBJECT_UPDATE)) {
            return Arrays.stream(PATTERN.split(attributeNames))
                         .collect(Collectors.collectingAndThen(
                             Collectors.toMap(EnhancedClientUtils::keyRef, Function.identity()),
                             Collections::unmodifiableMap));
        } else {
            return Collections.singletonMap(keyRef(attributeNames), attributeNames);
        }
    }
```

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/operations/UpdateItemOperation.java
```diff
@@ -141,6 +151,56 @@ public UpdateItemRequest generateRequest(TableSchema<T> tableSchema,
...
+                                    ignoreNulls);
+                } else {
+                    itemToMap.put(nestedAttributeKey, mapValue);
+                }
+            }
+        });
+        return itemToMap;
+    }
+    
+    private boolean attributeValueNonNullOrShouldWriteNull(boolean ignoreNulls, AttributeValue attributeValue) {
```
joviegas, line 201: i think we will need to do some additional check here to make sure the intentional null are updated

example
### value in DDB
```json
{
  "id": "ignoreNullFalse",
  "nestedRecord": {
    "id":"nestedId",
    "nestedUpdateBehaviorAttribute": "FixedOverride",
    "nestedCounter": 9999
  },
  "version": 11
}
```

I want to Update  "nestedUpdateBehaviorAttribute":  as null

in which case DDB should be updated as 
```json
{
  "id": "ignoreNullFalse",
  "nestedRecord": {
    "id":"nestedId",
    "nestedUpdateBehaviorAttribute": null,
    "nestedCounter": 9999
  },
  "version": 12
}
```

also could you please update a test case for this scenario

.idea/codeStyles/Project.xml
cenedhryn: Does `git reset` have any effect?

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/operations/UpdateItemOperation.java
```diff
@@ -145,6 +151,56 @@ public UpdateItemRequest generateRequest(TableSchema<T> tableSchema,
...
         return requestBuilder.build();
     }
+    
+    /**
+     * Method checks if a nested object parameter requires an update
+     * If so flattens out nested params separated by "_NESTED_ATTR_UPDATE_"
+     * this is consumed by @link EnhancedClientUtils to form the appropriate UpdateExpression
+     */
+    public Map<String, AttributeValue> transformItemToMapForUpdateExpression(Map<String, AttributeValue> itemToMap,
+                                                                             boolean ignoreNulls) {
```
cenedhryn, line 161: Can we simplify this method since as far as I can see, `ignoreNulls` is always true? Can we just remove it - this is an internal class.

services-custom/dynamodb-enhanced/src/test/java/software/amazon/awssdk/enhanced/dynamodb/functionaltests/UpdateBehaviorTest.java
```diff
@@ -145,6 +155,168 @@ public void updateBehaviors_transactWriteItems_secondUpdate() {
...
+        updateFlattenRecord.setCompositeRecord(updateCompositeRecord);
+        updateFlattenRecord.setId("id789");
+        
+        FlattenRecord persistedFlattenedRecord = flattenedMappedTable.updateItem(r -> r.item(updateFlattenRecord).ignoreNulls(true));
+        
+        assertThat(persistedFlattenedRecord.getCompositeRecord()).isNotNull();
+        assertThat(persistedFlattenedRecord.getCompositeRecord().getNestedRecord().getNestedCounter()).isEqualTo(100L);
+        assertThat(persistedFlattenedRecord.getCompositeRecord().getNestedRecord().getNestedRecord().getNestedCounter()).isEqualTo(50L);
+    }
+
```
joviegas, line 319: Could you please write a test case to see what happens if we just update a inner nested value

1. Put Item initially
```json
{

  "lastAutoUpdatedOn": "2024-07-26T17:58:24.647273Z",
  "nestedRecord": {
    "nestedTimeAttribute": "2020-05-03T10:00:00Z",
    "nestedRecord": null,
    "id": "id456",
    "nestedUpdateBehaviorAttribute": "testBehaviourAttribute",
    "nestedCounter": 5555,
    "nestedVersionedAttribute": null
  },
  "id": "ignoreNullFalse",
  "version": 1
}
```

Now do a update item to update the inner Nested record like
```java
        NestedRecordWithUpdateBehavior updatedNestedRecord = new NestedRecordWithUpdateBehavior();


        NestedRecordWithUpdateBehavior updatedInnerNestedRecord = new NestedRecordWithUpdateBehavior();
        updatedInnerNestedRecord.setNestedCounter(1L);
        updatedInnerNestedRecord.setId("0001");
        updatedNestedRecord.setNestedRecord(updatedInnerNestedRecord);
        RecordWithUpdateBehaviors update_record = new RecordWithUpdateBehaviors();
        update_record.setId("ignoreNullFalse");
        update_record.setVersion(1L);
        update_record.setNestedRecord(updatedNestedRecord);

        mappedTable.updateItem(r -> r.item(update_record).ignoreNulls(true));

```
Currently I saw that we get 
```
software.amazon.awssdk.services.dynamodb.model.DynamoDbException: The document path provided in the update expression is invalid for update (Service: DynamoDb, Status Code: 400, Request ID: xxx)
```

anirudh9391: I run 

```
@Test
    public void test() {
        NestedRecordWithUpdateBehavior nestedRecord1 = createNestedWithDefaults("id789", 50L);

        NestedRecordWithUpdateBehavior nestedRecord2 = createNestedWithDefaults("id456", 0L);
        nestedRecord2.setNestedRecord(nestedRecord1);

        RecordWithUpdateBehaviors record = new RecordWithUpdateBehaviors();
        record.setId("id123");
        record.setNestedRecord(nestedRecord2);

        mappedTable.putItem(record);

        NestedRecordWithUpdateBehavior updatedNestedRecord2 = new NestedRecordWithUpdateBehavior();
        long innerNestedCounter = 100L;
        updatedNestedRecord2.setNestedCounter(innerNestedCounter);

        // A Null Attribute set as part of update
        updatedNestedRecord2.setAttribute(TEST_ATTRIBUTE);

        NestedRecordWithUpdateBehavior updatedNestedRecord1 = new NestedRecordWithUpdateBehavior();
        updatedNestedRecord1.setNestedRecord(updatedNestedRecord2);
        long outerNestedCounter = 200L;
        updatedNestedRecord1.setNestedCounter(outerNestedCounter);

        RecordWithUpdateBehaviors update_record = new RecordWithUpdateBehaviors();
        update_record.setId("id123");
        update_record.setVersion(1L);
        update_record.setNestedRecord(updatedNestedRecord1);

        mappedTable.updateItem(r -> r.item(update_record).ignoreNulls(true));

        RecordWithUpdateBehaviors persistedRecord = mappedTable.getItem(r -> r.key(k -> k.partitionValue("id123")));

        verifyMultipleLevelNestingTargetedUpdateBehavior(persistedRecord, outerNestedCounter, innerNestedCounter);

        assertThat(persistedRecord.getNestedRecord().getNestedRecord().getAttribute()).isEqualTo(
            TEST_ATTRIBUTE);

    }
```

and I get a successful. In the test case specified, a new level of nesting is being created with an update. Our existing behaviour does not support it.

anirudh9391: I will add this added check as part of the existing test though. 

```
updatedNestedRecord2.setAttribute(TEST_ATTRIBUTE);
assertThat(persistedRecord.getNestedRecord().getNestedRecord().getAttribute()).isEqualTo(
            TEST_ATTRIBUTE);
```

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/update/UpdateExpressionUtils.java
anirudh9391: i can modify this

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/operations/UpdateItemOperation.java
anirudh9391: Good catch, i can do that

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/operations/UpdateItemOperation.java
anirudh9391: We do not handle updating record values to null for nested / or non-nested records as of today. 

```
@Test
    public void when_updatingNestedObjectWithSingleLevelWithNullOrEmpty_existingInformationIsPreservedAndUpdated_ignoreNulls() {

        NestedRecordWithUpdateBehavior nestedRecord = createNestedWithDefaults("id456", 5L);

        RecordWithUpdateBehaviors record = new RecordWithUpdateBehaviors();
        record.setId("id123");
        record.setCreatedOn(INSTANT_2);
        record.setNestedRecord(nestedRecord);

        mappedTable.putItem(record);

        NestedRecordWithUpdateBehavior updatedNestedRecord = new NestedRecordWithUpdateBehavior();
        long updatedNestedCounter = 10L;
        updatedNestedRecord.setNestedCounter(updatedNestedCounter);
        //updatedNestedRecord.setNestedUpdateBehaviorAttribute(null);

        RecordWithUpdateBehaviors update_record = new RecordWithUpdateBehaviors();
        update_record.setId("id123");
        update_record.setCreatedOn(null);
        update_record.setVersion(1L);

        update_record.setNestedRecord(updatedNestedRecord);

        mappedTable.updateItem(r -> r.item(update_record).ignoreNulls(true));

        RecordWithUpdateBehaviors persistedRecord = mappedTable.getItem(r -> r.key(k -> k.partitionValue("id123")));

        verifySingleLevelNestingTargetedUpdateBehavior(persistedRecord, updatedNestedCounter);
        assertThat(persistedRecord.getCreatedOn()).isNull();
```

The above test fails when we update out of our non-nested parent level attributes to null, and that is becasue of the way itemToMap models data. To change this behaviour we need to touch itemToMap. 

Diving deeper : we ignore attributes that are "null" and cannot differentiate between values set to null, and values that are null because they have not been updated in that operation. This is a bug that we had originally thought about, but it involves fundamental changes to itemToMap so we park it as an adjacent problem.  Refer to https://github.com/aws/aws-sdk-java-v2/blob/91c88084af5e0aeb779f00b68b014bdd6f9f8425/services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/mapper/StaticImmutableTableSchema.java#L520. We would need a way of differentiating between user set nulls and nulls set due to lack of updates


Im not exactly sure what you mean when we refer to an "empty" nested record. I can test it for cases when String attributes are updated to null

.idea/codeStyles/Project.xml
anirudh9391: fixed this

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/update/UpdateExpressionUtils.java
anirudh9391: modified

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/operations/UpdateItemOperation.java
anirudh9391: fixed

services-custom/dynamodb-enhanced/src/test/java/software/amazon/awssdk/enhanced/dynamodb/functionaltests/UpdateBehaviorTest.java
anirudh9391: Added test for null and empty string attribute setting in the next revision

joviegas:
```java
    DynamoDbClient myDynamoDbClient = DynamoDbClient.create();
    String MY_TABLE = "joviegas-test-update_07cfdb74-8757-4918-9a05-baeb6eae1682";


    private DynamoDbTable<RecordWithUpdateBehaviors> getTheTable() {


        DynamoDbEnhancedClient enhancedClient = DynamoDbEnhancedClient.builder()
                                                                      .dynamoDbClient(myDynamoDbClient).extensions(
                Stream.concat(ExtensionResolver.defaultExtensions().stream(),
                              Stream.of(AutoGeneratedTimestampRecordExtension.create())).collect(Collectors.toList())).build();


        DynamoDbTable<RecordWithUpdateBehaviors> mappedTable =
            enhancedClient.table(MY_TABLE, TABLE_SCHEMA);
        return mappedTable;
    }



   @Test
    public void when_updatingNestedObjectWithSingleLevel_() {


        DynamoDbTable<RecordWithUpdateBehaviors> mappedTable = getTheTable();

        String uuid = UUID.randomUUID().toString();


        NestedRecordWithUpdateBehavior nestedRecord = createNestedWithDefaults("id456", 5555L);

        RecordWithUpdateBehaviors insertRecord = new RecordWithUpdateBehaviors();
        insertRecord.setId(uuid);
        insertRecord.setNestedRecord(nestedRecord);

        mappedTable.putItem(insertRecord);

        NestedRecordWithUpdateBehavior updatedNestedRecord = new NestedRecordWithUpdateBehavior();


        NestedRecordWithUpdateBehavior updatedInnerNestedRecord = new NestedRecordWithUpdateBehavior();
        updatedInnerNestedRecord.setNestedCounter(0L);
        updatedInnerNestedRecord.setId("0000");

        updatedNestedRecord.setNestedUpdateBehaviorAttribute(null);
        updatedNestedRecord.setNestedRecord(updatedInnerNestedRecord);
        // updatedNestedRecord.setNestedCounter(9999L);

        RecordWithUpdateBehaviors update_record = new RecordWithUpdateBehaviors();
        update_record.setId(uuid);
        update_record.setVersion(1L);
        update_record.setNestedRecord(updatedNestedRecord);

        mappedTable.updateItem(r -> r.item(update_record).ignoreNulls(true));


        GetItemResponse item = myDynamoDbClient.getItem(f -> f.tableName(MY_TABLE).key(Collections.singletonMap("id",
                                                                                                                AttributeValue.builder().s(uuid).build())));

        System.out.println("item is " +item.item());

        // verifySingleLevelNestingTargetedUpdateBehavior(persistedRecord, updatedNestedCounter);

    }
```

Could you please try this on real Dynamo table the test case in this class used LocalDDB which is a mock DDB.
Might be service behaves differently for this .

If we confirm its not a SDK issue then we can raise issue on DDB team internally

joviegas:
When a map is set/updated to empty empty same should be sent to Dynamo DB
Current behavior supports this , that is when we do new `record.setNestedRecord(new NestedRecordWithUpdateBehavior());` this is transfered as empty map in wire and also set as empty Attribute Value in DDB.
But with this fix we are setting it to null , which is incorrect.
Below test case shows how earlier we used to get empty attribute value map for `getItemResponse.item().get("nestedRecord"` and now null .



```java
public class TestCaseForEmptyMapUpdate {

    private static final String TABLE_NAME = "joviegas-update-test-table";
    private static final TableSchema<RecordWithUpdateBehaviors> TABLE_SCHEMA = TableSchema.fromClass(RecordWithUpdateBehaviors.class);

    private static final DynamoDbEnhancedClient ENHANCED_CLIENT = DynamoDbEnhancedClient.builder()
                                                                                        .dynamoDbClient(DynamoDbClient.create())
                                                                                        .extensions(
                                                                                            Stream.concat(
                                                                                                ExtensionResolver.defaultExtensions().stream(),
                                                                                                Stream.of(AutoGeneratedTimestampRecordExtension.create())
                                                                                            ).collect(Collectors.toList())
                                                                                        )
                                                                                        .build();

    private static final DynamoDbTable<RecordWithUpdateBehaviors> MAPPED_TABLE = ENHANCED_CLIENT.table(TABLE_NAME, TABLE_SCHEMA);

        String randomId = UUID.randomUUID().toString();

        RecordWithUpdateBehaviors record = new RecordWithUpdateBehaviors();
        record.setId(randomId);
        record.setNestedRecord(new NestedRecordWithUpdateBehavior());

        MAPPED_TABLE.updateItem(r -> r.item(record).ignoreNulls(true));

        // Retrieve the element
        DynamoDbClient dynamoDbClient = DynamoDbClient.create();
        GetItemResponse getItemResponse = dynamoDbClient.getItem(GetItemRequest.builder()
                                                                               .key(Collections.singletonMap("id", AttributeValue.fromS(randomId)))
                                                                               .tableName(TABLE_NAME)
                                                                               .build());

        System.out.println("Retrieved nestedRecord: " + getItemResponse.item().get("nestedRecord"));
    }
}
```

anirudh9391:
> ```java
> String randomId = UUID.randomUUID().toString();
> 
>         RecordWithUpdateBehaviors record = new RecordWithUpdateBehaviors();
>         record.setId(randomId);
>         record.setNestedRecord(new NestedRecordWithUpdateBehavior());
> 
>         MAPPED_TABLE.updateItem(r -> r.item(record).ignoreNulls(true));
> 
>         // Retrieve the element
>         DynamoDbClient dynamoDbClient = DynamoDbClient.create();
>         GetItemResponse getItemResponse = dynamoDbClient.getItem(GetItemRequest.builder()
>                                                                                .key(Collections.singletonMap("id", AttributeValue.fromS(randomId)))
>                                                                                .tableName(TABLE_NAME)
>                                                                                .build());
> 
>         System.out.println("Retrieved nestedRecord: " + getItemResponse.item().get("nestedRecord"));
> ```

So going forward, any update to "Objects" in a nested item will not be made with the ignoreNulls = true configuration, as discussed in our previous design meeting. Regardless of whatever value (null or non-null).

anirudh9391:
> When a map is set or updated to an empty map, the same should be sent to DynamoDB. The current behavior supports this. Specifically, when we execute record.setNestedRecord(new NestedRecordWithUpdateBehavior());, it is transmitted as an empty map over the wire and is correctly stored as an empty AttributeValue in DynamoDB.
> 
> However, with the recent change, the map is now being set to null, which is incorrect. The test case below demonstrates the issue: previously, getItemResponse.item().get("nestedRecord") would return an empty attribute value map, but now it returns null.
> 
> ```java
> public class TestCaseForEmptyMapUpdate {
> 
>     private static final String TABLE_NAME = "joviegas-update-test-table";
>     private static final TableSchema<RecordWithUpdateBehaviors> TABLE_SCHEMA = TableSchema.fromClass(RecordWithUpdateBehaviors.class);
> 
>     private static final DynamoDbEnhancedClient ENHANCED_CLIENT = DynamoDbEnhancedClient.builder()
>                                                                                         .dynamoDbClient(DynamoDbClient.create())
>                                                                                         .extensions(
>                                                                                             Stream.concat(
>                                                                                                 ExtensionResolver.defaultExtensions().stream(),
>                                                                                                 Stream.of(AutoGeneratedTimestampRecordExtension.create())
>                                                                                             ).collect(Collectors.toList())
>                                                                                         )
>                                                                                         .build();
> 
>     private static final DynamoDbTable<RecordWithUpdateBehaviors> MAPPED_TABLE = ENHANCED_CLIENT.table(TABLE_NAME, TABLE_SCHEMA);
> 
>     public static void main(String[] args) {
>         String randomId = UUID.randomUUID().toString();
> 
>         RecordWithUpdateBehaviors record = new RecordWithUpdateBehaviors();
>         record.setId(randomId);
>         record.setNestedRecord(new NestedRecordWithUpdateBehavior());
> 
>         MAPPED_TABLE.updateItem(r -> r.item(record).ignoreNulls(true));
> 
>         // Retrieve the element
>         DynamoDbClient dynamoDbClient = DynamoDbClient.create();
>         GetItemResponse getItemResponse = dynamoDbClient.getItem(GetItemRequest.builder()
>                                                                                .key(Collections.singletonMap("id", AttributeValue.fromS(randomId)))
>                                                                                .tableName(TABLE_NAME)
>                                                                                .build());
> 
>         System.out.println("Retrieved nestedRecord: " + getItemResponse.item().get("nestedRecord"));
> // This should be an empty map , you can test the same in master branch
>     }
> }
> ```
> 
> also in cli
> 
> ```
> aws dynamodb update-item \
>     --table-name joviegas-update-test-table \
>     --key '{"id": {"S": "clitest"}}' \
>     --update-expression "SET #NR = :emptyMap" \
>     --expression-attribute-names '{"#NR": "nestedRecord"}' \
>     --expression-attribute-values '{":emptyMap": {"M": {}}}' \
>     --return-values ALL_NEW -
>     ```
> ```

Seems like this is testing out a case where the customer uses an enhancedClient for the update but a lower level client for get ? I have modified it so that we store empty map values in the table, but try testing this out using the enhanced client like this 

```
@Test
    public void when_updatingNestedObjectToEmptyWithSingleLevel_existingInformationIsPreserved_ignoreNulls() {

        NestedRecordWithUpdateBehavior nestedRecord = createNestedWithDefaults("id456", 5L);
        nestedRecord.setAttribute(TEST_ATTRIBUTE);

        RecordWithUpdateBehaviors record = new RecordWithUpdateBehaviors();
        record.setId("id123");
        record.setNestedRecord(nestedRecord);

        mappedTable.putItem(record);

        NestedRecordWithUpdateBehavior updatedNestedRecord = new NestedRecordWithUpdateBehavior();

        RecordWithUpdateBehaviors update_record = new RecordWithUpdateBehaviors();
        update_record.setId("id123");
        update_record.setVersion(1L);
        update_record.setNestedRecord(updatedNestedRecord);

        mappedTable.updateItem(r -> r.item(update_record).ignoreNulls(true));

        RecordWithUpdateBehaviors persistedRecord = mappedTable.getItem(r -> r.key(k -> k.partitionValue("id123")));

        assertThat(persistedRecord.getNestedRecord()).isNull();
        assertThat(persistedRecord.getNestedRecord().getNestedCounter()).isEqualTo(updatedNestedCounter);
        assertThat(persistedRecord.getNestedRecord().getNestedUpdateBehaviorAttribute()).isEqualTo(TEST_BEHAVIOUR_ATTRIBUTE);
        assertThat(persistedRecord.getNestedRecord().getNestedTimeAttribute()).isEqualTo(INSTANT_1);
        assertThat(persistedRecord.getNestedRecord().getAttribute()).isEqualTo(testAttribute);
    }
```

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/operations/UpdateItemOperation.java
```diff
@@ -141,6 +150,67 @@ public UpdateItemRequest generateRequest(TableSchema<T> tableSchema,
...
+                    nestedItemToMap(itemToMap, nestedAttributeKey, mapValue);
+                } else {
+                    itemToMap.put(nestedAttributeKey, mapValue);
+                }
+            }
+        });
+        return itemToMap;
+    }
+
+    private boolean isNotEmptyMap(Map<String, AttributeValue> map) {
```
joviegas, line 197:
```java
private boolean isNotEmptyMap(Map<String, AttributeValue> map) {
    return !map.isEmpty() && map.values().stream()
        .anyMatch(this::attributeValueNonNullOrShouldWriteNull);
}
```

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/update/UpdateExpressionUtils.java
```diff
@@ -99,7 +103,7 @@ private static List<RemoveAction> removeActionsFor(Map<String, AttributeValue> a
     private static RemoveAction remove(String attributeName) {
         return RemoveAction.builder()
                            .path(keyRef(attributeName))
-                           .expressionNames(Collections.singletonMap(keyRef(attributeName), attributeName))
+                           .expressionNames(expressionNamesFor(attributeName))
```
joviegas, line 106: Do we have  test case that covers `REMOVE`  action ?
What is the scenario where "REMOVE" comes in to play

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/EnhancedClientUtils.java
```diff
@@ -40,6 +43,7 @@ public final class EnhancedClientUtils {
     private static final Set<Character> SPECIAL_CHARACTERS = Stream.of(
         '*', '.', '-', '#', '+', ':', '/', '(', ')', ' ',
         '&', '<', '>', '?', '=', '!', '@', '%', '$', '|').collect(Collectors.toSet());
+    private static final Pattern PATTERN = Pattern.compile(NESTED_OBJECT_UPDATE);
```
joviegas, line 46: Can we rename this to
```
NESTED_OBJECT_PATTERN
```

joviegas:
Thank you so much for handling the changes. Unfortunately, I wasn't able to test further due to some other issues. Could you kindly confirm if the following cases work as expected?

    Updating a nested attribute value for the following types: BinaryAttributeValue, StringSetAttributeValue, NumberSetAttributeValue, BinarySetAttributeValue, ListAttributeValue, NullAttributeValue, and BooleanAttributeValue. (Covering all possible [AttributeValues-Link](https://github.com/aws/aws-sdk-java-v2/blob/71e129f3dcff9caa634316237cdb60509fa82f58/services/dynamodb/src/main/resources/codegen-resources/dynamodb/service-2.json#L1068-L1113))

There’s no need to write additional test cases, but a quick integration test would be greatly appreciated.

services-custom/dynamodb-enhanced/src/test/java/software/amazon/awssdk/enhanced/dynamodb/functionaltests/UpdateBehaviorTest.java
```diff
@@ -304,6 +302,21 @@ public void when_updatingNestedNonScalarObject_DynamoDBExceptionIsThrown() {
...
+        String randomId = "id123";
+
+        RecordWithUpdateBehaviors record = new RecordWithUpdateBehaviors();
+        record.setId(randomId);
+        record.setNestedRecord(new NestedRecordWithUpdateBehavior());
+
+        mappedTable.updateItem(r -> r.item(record).ignoreNulls(true));
+
+        RecordWithUpdateBehaviors persistedRecord = mappedTable.getItem(r -> r.key(k -> k.partitionValue("id123")));
+        assertThat(persistedRecord.getNestedRecord()).isNull();
```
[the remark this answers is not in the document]
anirudh9391: As discussed previously, the integ test written here covers that case : https://github.com/aws/aws-sdk-java-v2/pull/5380#issuecomment-2319627533. The enhanced ddb client behaves this way due to an issue while deserializing to Bean.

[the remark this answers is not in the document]
anirudh9391: I included this in the test suit to ensure we are consistent in our behaviour and it doesnt change as a result of the fix.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/update/UpdateExpressionUtils.java
anirudh9391: Yeah, remove does not come into play here. I modified it to make it neater and reuse the expressionNamesFor method. I can undo these changes to make it less confusing in the commit log.

services-custom/dynamodb-enhanced/src/test/java/software/amazon/awssdk/enhanced/dynamodb/functionaltests/UpdateBehaviorTest.java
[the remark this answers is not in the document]
anirudh9391: discussed offline. modified the test case. dropping the enhanced ddb behaviour validation

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/EnhancedClientUtils.java
anirudh9391: done

services-custom/dynamodb-enhanced/src/test/java/software/amazon/awssdk/enhanced/dynamodb/functionaltests/UpdateBehaviorTest.java
```diff
@@ -304,6 +302,21 @@ public void when_updatingNestedNonScalarObject_DynamoDBExceptionIsThrown() {
             .isInstanceOf(DynamoDbException.class);
     }
 
+    @Test
+    public void test() {
```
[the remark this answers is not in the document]
anirudh9391: done

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/operations/UpdateItemOperation.java
anirudh9391: done

anirudh9391:
> Thank you so much for handling the changes. Unfortunately, I wasn't able to test further due to some other issues. Could you kindly confirm if the following cases work as expected? Updating a nested attribute value for the following types: BinaryAttributeValue, StringSetAttributeValue, NumberSetAttributeValue, BinarySetAttributeValue, ListAttributeValue, NullAttributeValue, and BooleanAttributeValue.
> 
> (Covering all possible [AttributeValues-Link](https://github.com/aws/aws-sdk-java-v2/blob/71e129f3dcff9caa634316237cdb60509fa82f58/services/dynamodb/src/main/resources/codegen-resources/dynamodb/service-2.json#L1068-L1113))
> 
> There’s no need to write additional test cases, but a quick integration test would be greatly appreciated.

```
@Test
    public void nestedRecord() {
        RecordWithUpdateBehaviors item = new RecordWithUpdateBehaviors();
        item.setId("abc");
        item.setVersion(1L);

        MAPPED_TABLE.putItem(r -> r.item(item));*/

        NestedRecordWithUpdateBehavior nested = new NestedRecordWithUpdateBehavior();
        nested.setFlag(false);
        nested.setList(Arrays.asList("a", "b", "c"));
        nested.setNumber(8);
        nested.setNumberSet(new HashSet<>(Arrays.asList(1, 2, 3)));
        nested.setStringSet(new HashSet<>(Arrays.asList("d", "e", "f")));
        nested.setBinary(ByteBuffer.allocateDirect(1));
        nested.setByteList(Arrays.asList(ByteBuffer.allocateDirect(1), ByteBuffer.allocateDirect(1)));

        RecordWithUpdateBehaviors update = new RecordWithUpdateBehaviors();
        update.setId("abc");
        update.setVersion(16L);
        update.setNestedRecord(nested);

        MAPPED_TABLE.updateItem(r -> r.item(update).ignoreNulls(true));

        // Retrieve the element
        DynamoDbClient dynamoDbClient = DynamoDbClient.create();

        GetItemResponse getItemResponse = dynamoDbClient.getItem(GetItemRequest.builder()
                                                                               .key(Collections.singletonMap("id",
                                                                                                             AttributeValue.fromS(randomId)))
                                                                               .tableName(TABLE_NAME)
                                                                               .build());

        assertThat(getItemResponse.item().get("nestedRecord").toString()).isEqualTo("AttributeValue(M={flag=AttributeValue"
                                                                                    + "(BOOL=false), stringSet=AttributeValue"
                                                                                    + "(SS=[d, e, f]), "
                                                                                    + "nestedRecord=AttributeValue(NUL=true), "
                                                                                    + "byteList=AttributeValue"
                                                                                    + "(L=[AttributeValue(B=SdkBytes"
                                                                                    + "(bytes=0x00)), AttributeValue(B=SdkBytes"
                                                                                    + "(bytes=0x00))]), list=AttributeValue"
                                                                                    + "(L=[AttributeValue(S=a), AttributeValue"
                                                                                    + "(S=b), AttributeValue(S=c)]), "
                                                                                    + "nestedUpdateBehaviorAttribute"
                                                                                    + "=AttributeValue(NUL=true), "
                                                                                    + "nestedCounter=AttributeValue(NUL=true), "
                                                                                    + "numberSet=AttributeValue(NS=[3, 2, 1]), "
                                                                                    + "nestedVersionedAttribute=AttributeValue"
                                                                                    + "(NUL=true), number=AttributeValue(N=8), "
                                                                                    + "nestedTimeAttribute=AttributeValue"
                                                                                    + "(NUL=true), binary=AttributeValue"
                                                                                    + "(B=SdkBytes(bytes=0x00)), "
                                                                                    + "attribute=AttributeValue(NUL=true), "
                                                                                    + "id=AttributeValue(NUL=true)})");
    }
```

joviegas (approved):

phillipberndt:
This backwards incompatible change broke our existing code that relied on this behavior. I cut https://github.com/aws/aws-sdk-java-v2/issues/5584 to track that.

PR #5383, opened 2024-07-09
fix SigV4a signer incorrectly interpreting query params with '&'

L-Applin:
## Motivation and Context
Fix an issue with the default `DefaultAwsCrtV4aHttpSigner` where it split query parameters value containing ampersands into multiple query params instead of encoding it in the query string.

## Modifications
The `DefaultAwsCrtV4aHttpSigner` receive encoded string, but was decoding it (using `uri.getQuery()`) before going through the logic for appending query param in the SdkHttpRequest. This would result in ampersand in the param value to be treated as query separator. Instead, that logic now uses the encoded path value so that ampersands are uri-encoded and not seen as query param separator. This causes an additional problem since the http request builder method `.appendRawQueryParameter(...)` expect query param values to *not* be encoded. To fix this they are decoded before passing them to the http request builder.

## Testing
added unit tests

## Types of changes


dagnir (approved):

dagnir:
LGTM. Did we run integ tests?

L-Applin:
> LGTM. Did we run integ tests?

No, good point, I will start a run

PR #5393, opened 2024-07-12
Add enum getters and casing transforms

davidh44:
## Motivation and Context
Transform enums from v1 to v2
- Casing: [enum constants from pascal case to screaming snake case](https://github.com/aws/aws-sdk-java-v2/blob/master/codegen/src/main/java/software/amazon/awssdk/codegen/naming/DefaultNamingStrategy.java#L271)
- Getters: XXAsString(s) to return String instead of enum type

### DIFF
--- a/v2-migration/src/main/java/software/amazon/awssdk/v2migration/EnumCasingToV2.java
+++ b/v2-migration/src/main/java/software/amazon/awssdk/v2migration/EnumCasingToV2.java
@@ -28,6 +28,7 @@ import org.openrewrite.java.tree.JavaType;
 import org.openrewrite.java.tree.TypeUtils;
 import software.amazon.awssdk.annotations.SdkInternalApi;
 import software.amazon.awssdk.utils.StringUtils;
+import software.amazon.awssdk.v2migration.internal.utils.SdkTypeUtils;
 
 @SdkInternalApi
 public class EnumCasingToV2 extends Recipe {
@@ -85,14 +86,10 @@ public class EnumCasingToV2 extends Recipe {
             JavaType javaType = fa.getTarget().getType();
             JavaType.FullyQualified fullyQualified = TypeUtils.asFullyQualified(javaType);
             if (fullyQualified != null) {
-                return isV2ModelClass(fullyQualified.getFullyQualifiedName())
+                return SdkTypeUtils.isV2ModelClass(javaType)
                        && fullyQualified.getKind() == JavaType.FullyQualified.Kind.Enum;
             }
             return false;
-        }
-
-        public boolean isV2ModelClass(String fcqn) {
-            return fcqn.startsWith("software.amazon.awssdk.services.") && fcqn.contains(".model.");
         }
     }
 }
--- a/v2-migration/src/main/java/software/amazon/awssdk/v2migration/EnumCasingToV2.java
+++ b/v2-migration/src/main/java/software/amazon/awssdk/v2migration/EnumCasingToV2.java
@@ -1,0 +1,93 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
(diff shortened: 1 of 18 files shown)


v2-migration/src/main/resources/META-INF/rewrite/change-enum-getters.yml
```diff
@@ -0,0 +1,378 @@
...
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+---
+type: specs.openrewrite.org/v1beta/recipe
+name: software.amazon.awssdk.v2migration.EnumGettersToV2
+displayName: Change v1 enum getters to v2
+recipeList:
+  - org.openrewrite.java.ChangeMethodName:
+      methodPattern: software.amazon.awssdk.services.sqs.model.CreateQueueRequest attributes()
```
zoewangg, line 20: I think we need to perform this recipe before ChangeSdkType because v2 POJOs actually have both `xxx()` and `xxxAsString()` and this will cause issues for customers who use Java v1 and Java v2 in their workspace. Can we add a test case for it?

v2-migration/src/test/java/software/amazon/awssdk/v2migration/EnumCasingToV2Test.java
```diff
@@ -0,0 +1,60 @@
+/*
```
zoewangg, line 1: Can we add some end-to-end tests? https://github.com/aws/aws-sdk-java-v2/tree/feature/master/migration-tool/test/v2-migration-tests

v2-migration/src/main/resources/META-INF/rewrite/change-enum-getters.yml
davidh44: Good catch, updated to run before ChangeSdkType and change v1 classnames, added end-to-end tests with v2 pojos included

v2-migration/src/main/java/software/amazon/awssdk/v2migration/EnumCasingToV2.java
```diff
@@ -0,0 +1,98 @@
...
+    public String getDescription() {
+        return "Transforms V1 enum constants from pascal case to screaming snake case for v2.";
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return new Visitor();
+    }
+
+    private static String v2Casing(String enumValue) {
```
zoewangg, line 52: Can we use `DefaultNamingStrategy.getEnumValueName` here directly?

v2-migration/src/main/java/software/amazon/awssdk/v2migration/EnumCasingToV2.java
```diff
@@ -0,0 +1,98 @@
...
+            }
+
+            return id;
+        }
+
+        public boolean isV2EnumValue(J.FieldAccess fa) {
+            JavaType javaType = fa.getTarget().getType();
+            JavaType.FullyQualified fullyQualified = TypeUtils.asFullyQualified(javaType);
+            if (fullyQualified != null) {
+                return isV2ModelClass(fullyQualified.getFullyQualifiedName())
```
zoewangg, line 88: nit, we can use `SdkTypeUtils.isV2ModelClass`

v2-migration/src/main/resources/META-INF/rewrite/aws-sdk-java-v1-to-v2.yml
```diff
@@ -33,3 +34,4 @@ recipeList:
   - software.amazon.awssdk.v2migration.V1GetterToV2
   - software.amazon.awssdk.v2migration.HttpSettingsToHttpClient
   - software.amazon.awssdk.v2migration.WrapSdkClientBuilderRegionStr
+  - software.amazon.awssdk.v2migration.EnumCasingToV2
```
zoewangg, line 37: Would it make sense to move `EnumCasingToV2` before ChangeSdkType as well?

zoewangg (approved):

PR #5385, opened 2024-07-09
Add exception handling for TLS half-close in ApacheHTTPClient

joviegas:
## Motivation and Context

Implemented a feature to handle TLS half-close scenarios by throwing an exception in the ApacheHTTPClient package. In TLS 1.3, the inbound and outbound close_notify alerts are independent. When the client receives a close_notify alert, it only closes the inbound stream but continues to send data to the server. Previously, the SDK could not detect that the connection was closed on the server side, causing it to get stuck while writing to the socket and eventually timing out. This feature ensures proper detection and handling of closed connections, improving overall reliability by preventing client hangs.

## Modifications

- Merging previously reviewed https://github.com/aws/aws-sdk-java-v2/pull/5257
- Add Junits

## Testing

- Junit test cases

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

## License

- [x] I confirm that this pull request can be released under the Apache 2 license

PR #5398, opened 2024-07-15
 Add exception handling for TLS half-close in ApacheHTTPClient #5385 

joviegas:
## Motivation and Context

Re-merge of https://github.com/aws/aws-sdk-java-v2/pull/5385

The unit test case were failing for older builds of Java 1.8 B171

```
Implemented a feature to handle TLS half-close scenarios by throwing an exception in the ApacheHTTPClient package. In TLS 1.3, the inbound and outbound close_notify alerts are independent. When the client receives a close_notify alert, it only closes the inbound stream but continues to send data to the server. Previously, the SDK could not detect that the connection was closed on the server side, causing it to get stuck while writing to the socket and eventually timing out. This feature ensures proper detection and handling of closed connections, improving overall reliability by preventing client hangs.
```

## Modifications

- Modified Test case to support older Java 1.8 which do not support half close.

## Testing
-Junit Test added 

## Types of changes

## Checklist

## License


test/http-client-tests/src/main/java/software/amazon/awssdk/http/server/MockServer.java
```diff
@@ -150,10 +156,12 @@ public static class HalfCloseServerBehavior implements ServerBehaviorStrategy {
 
         @Override
         public void runServer(ServerSocket serverSocket) {
+            Socket socket = null;
             try {
                 while (true) {
-                    Socket socket = null;
+
                     try {
+                        Thread.sleep(200);
```
zoewangg, line 164: Is there any reason we need to add a sleep?

joviegas: Yes, since we are interrupting when we stopServer(), we include a sleep to ensure the interrupt signal is received. Otherwise, the accept() call continues waiting until the join timeout is over.

joviegas: I removed the sleep . Now we will return if socket exception are seen in case of halfclose and preclose.
While for success case we wait until stopserver interrupts the listener thread,

zoewangg (approved):

PR #5404, opened 2024-07-17
Use SRA auth by default for all existing and new services.

millems:
The customization is preserved as an option, but will be removed in the future.


core/auth/pom.xml
```diff
@@ -78,6 +78,11 @@
             <artifactId>http-auth-aws</artifactId>
             <version>${awsjavasdk.version}</version>
         </dependency>
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>http-auth-aws-eventstream</artifactId>
+            <version>${awsjavasdk.version}</version>
+        </dependency>
```
millems, line 85: This fixes an issue where existing non-AWS event stream services (that depend on `auth`) were not getting the event bridge signer, when they were getting it pre-SRA.

davidh44 (approved):

PR #5410, opened 2024-07-18
 Use code-generated endpoint rules for all services, and remove the customization.

millems:
Additional necessary changes to facilitate this:
1. Fixed an issue where any region parameter was assumed to be the endpoint region. The fix also removes the region as a local variable and instead relies on reading the region from the parameters. A "{name}Id" method has been added to the endpoint and auth resolver's region parameters, which allows readingthe region ID as a string.
2. Removed "auth scheme strategy" classes from code generated to just being in S3, because they're not used anywhere else, now. In the future, it would be nice to simplify this S3 logic.

This PR is split between the `customization.config` cleanups, and the codegen changes, to hopefully simplify the review.


codegen/src/main/java/software/amazon/awssdk/codegen/poet/rules/EndpointProviderSpec.java
```diff
@@ -17,191 +17,281 @@
...
 import software.amazon.awssdk.codegen.poet.ClassSpec;
 import software.amazon.awssdk.codegen.poet.PoetUtils;
+import software.amazon.awssdk.codegen.poet.rules.EndpointRulesSpecUtils;
 import software.amazon.awssdk.core.exception.SdkClientException;
-import software.amazon.awssdk.endpoints.Endpoint;
 import software.amazon.awssdk.utils.CompletableFutureUtils;
-import software.amazon.awssdk.utils.Logger;
 import software.amazon.awssdk.utils.Validate;
 
 public class EndpointProviderSpec implements ClassSpec {
```
millems, line 45: It looks like there are a lot of changes here, but this is actually just `EndpointProviderSpec2`, which has already been code reviewed and is listed as a deletion below. You can skip reviewing this class if you want, since it wasn't changed.

services/s3/src/main/java/software/amazon/awssdk/services/s3/endpoints/internal/EndpointAttributeProvider.java
```diff
@@ -0,0 +1,33 @@
...
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.services.s3.endpoints.internal;
```
millems, line 16: These s3.endpoints.internal files look new, but they are just copies of the .java.resource files that have already been code generated. I just copied them into S3 because this is the only service that uses them.

You can skip reviewing these files if you want, because they were already reviewed. I didn't change them.

codegen/src/main/java/software/amazon/awssdk/codegen/poet/rules/EndpointProviderSpec.java
anirudh9391: Looks like it wasnt merged in ? Would be great if you could provide a PR link

millems: https://github.com/aws/aws-sdk-java-v2/blame/9a3c99e5b70f608d20c18248aa9bdb5c8d98b684/codegen/src/main/java/software/amazon/awssdk/codegen/poet/rules2/EndpointProviderSpec2.java is the file already in the repo.

Original PR for it was https://github.com/aws/aws-sdk-java-v2/commit/db355d31b851a617b5d2c39b52b9fdcd7f2b7d4b

anirudh9391: got it thanks

codegen/src/main/java/software/amazon/awssdk/codegen/poet/rules/SymbolTable.java
```diff
@@ -13,22 +13,23 @@
...
     private final Map<String, RuleType> params;
     private final Map<String, RuleType> locals;
-    private final String regionParamName;
+    private final Map<String, BuiltInParameter> builtInParamTypes;
 
     SymbolTable(Builder builder) {
         this.params = Collections.unmodifiableMap(new LinkedHashMap<>(builder.params));
         this.locals = Collections.unmodifiableMap(new LinkedHashMap<>(builder.locals));
-        this.regionParamName = builder.regionParamName;
+        this.builtInParamTypes = Collections.unmodifiableMap(new LinkedHashMap<>(builder.builtInParamTypes));
```
anirudh9391, line 32: Do we need the overhead of maintaining a list with the LinkedHashMap ? I do not see us needing element ordering here in this class. Could we just do HashMap instead ?

services/s3/src/main/java/software/amazon/awssdk/services/s3/endpoints/internal/Value.java
```diff
@@ -364,15 +381,15 @@ public static Endpoint fromNode(JsonNode node) {
             JsonNode propertiesNode = objNode.get(PROPERTIES);
             if (propertiesNode != null) {
                 propertiesNode.asObject()
-                    .forEach((k, v) -> {
-                        b.property(k, Value.fromNode(v));
-                    });
+                              .forEach((k, v) -> {
```
anirudh9391, line 384: Do we want indentation changes like these to go through ? Feels like we have a lot to indent if we start looking into our entire codebase. Would it be better to leave them as they are ? It also shows up in the "git blame" section

services/sts/src/main/java/software/amazon/awssdk/services/sts/internal/StsAuthUtils.java
```diff
@@ -31,9 +31,11 @@ public static String accountIdFromArn(AssumedRoleUser assumedRoleUser) {
         if (assumedRoleUser == null) {
             return null;
         }
-        return Arn.parse(assumedRoleUser.arn())
-                  .map(Arn::accountId)
-                  .orElse(null);
+        RuleArn arn = RuleArn.parse(assumedRoleUser.arn());
+        if (arn == null) {
```
anirudh9391, line 35: nit : would be great to extract this into a method arnAccountId(RuleArn arn);

```
RuleArn arn = RuleArn.parse(assumedRoleUser.arn()); //or
RuleArn arn = RuleArn.parse(federatedUser.arn());

return arnAccountId(RuleArn arn);
```

codegen/src/main/java/software/amazon/awssdk/codegen/poet/rules/SymbolTable.java
millems: Seems reasonable. Let me see if there's any tests that assume order.

services/s3/src/main/java/software/amazon/awssdk/services/s3/endpoints/internal/Value.java
millems: It just did that since it was a copy+paste of code, which reformats things automatically in Intellij. I didn't do it intentionally. I can try to revert these types of changes. Hopefully I don't miss any.

services/sts/src/main/java/software/amazon/awssdk/services/sts/internal/StsAuthUtils.java
millems: Since they're in different classes (so the code is harder to share) and there's only 2 places it's done like this (so [the rule of three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)) isn't satisfied), I'd usually leave it like this. I don't feel strongly about it, though, so LMK if you'd prefer it to be shared. I'd probably do a convenience method on RuleArn, like RuleArn.accountId(arn).

services/s3/src/main/java/software/amazon/awssdk/services/s3/endpoints/internal/Value.java
millems: I actually don't need all of these files! I've deleted a lot of them.

anirudh9391 (approved):

PR #5418, opened 2024-07-23
Internal classes and RequestBatchManager Impelementation

joviegas:
# RequestManager Implementation details

## Overview

The `RequestManager` is responsible for managing and executing batched requests in an efficient manner. It utilizes various batch managers to handle different types of requests such as sending messages, deleting messages, changing message visibility, and handling generic request batches. The design ensures that requests are processed in an optimized and scalable way by leveraging concurrent processing and batching techniques.

### 1. `SendMessageBatchManager`
- **Purpose**: Manages the batching and sending of messages to SQS.
- **Key Responsibilities**:
  - Collects messages to be sent.
  - Batches messages into the appropriate size.
  - Sends the batched messages to SQS.
  - Handles responses and errors for the batched requests.

### 2. `RequestBatchManager`
- **Purpose**: Manages generic request batching for various types of SQS operations.
- **Key Responsibilities**:
  - Aggregates individual requests.
  - Creates batches of requests based on configuration.
  - Executes batched requests.
  - Provides mechanisms for retrying failed requests.

### 3. `RequestBatchBuffer`
- **Purpose**: Buffers requests before they are batched and sent.
- **Key Responsibilities**:
  - Temporarily stores requests.
  - Provides methods to retrieve and clear buffered requests.
  - Ensures thread-safe operations for request storage and retrieval.

### 4. `IdentifiableMessage`
- **Purpose**: Represents a message with a unique identifier.
- **Key Responsibilities**:
  - Stores message content and unique identifier.
  - Provides methods to access message details.
  
### 5. `DeleteMessageBatchManager`
- **Purpose**: Manages the batching and deletion of messages from SQS.
- **Key Responsibilities**:
  - Collects messages to be deleted.
  - Batches delete requests.
  - Sends batched delete requests to SQS.
  - Handles responses and errors for the delete requests.

### 6. `DefaultSqsAsyncBatchManager`
- **Purpose**: Default implementation for managing asynchronous SQS batch operations.
- **Key Responsibilities**:
  - Provides a default mechanism for handling various SQS batch operations asynchronously.
  - Ensures efficient processing of batch requests.
  
### 7. `ChangeMessageVisibilityBatchManager`
- **Purpose**: Manages the batching and visibility timeout changes for SQS messages.
- **Key Responsibilities**:
  - Collects visibility change requests.
  - Batches visibility change requests.
  - Sends batched requests to SQS.
  - Handles responses and errors for the visibility change requests.

### 8. `BatchingMap`
- **Purpose**: Utility class for managing batching contexts and operations.
- **Key Responsibilities**:
  - Stores and manages batch contexts.
  - Provides methods for adding and retrieving batches.
  - Ensures thread-safe access to batch contexts.
  
### 9. `BatchingExecutionContext`
- **Purpose**: Execution context for batching operations.
- **Key Responsibilities**:
  - Manages the lifecycle of batch executions.
  - Provides context-specific configurations and state management.
  
### 10. `RequestBatchConfiguration`
- **Purpose**: Configuration class for request batching.
- **Key Responsibilities**:
  - Stores configuration settings for batching operations.
  - Provides methods to access and modify configuration settings.

## Key Design Considerations

- **Concurrency**: The design ensures thread-safe operations using appropriate synchronization mechanisms.
- **Scalability**: Batching of requests allows for efficient handling of a large number of requests.
- **Error Handling**: Each batch manager includes mechanisms for handling and retrying failed requests.
- **Flexibility**: The use of different batch managers allows for easy extension and customization for various types of SQS operations.

## Conclusion

The `RequestManager` design leverages batching and asynchronous processing to efficiently manage and execute SQS requests. The modular design with specific batch managers for different request types ensures flexibility and scalability, making it suitable for high-throughput applications.

## License


services/sqs/src/main/java/software/amazon/awssdk/services/sqs/internal/batchmanager/core/RequestBatchManager.java
```diff
@@ -0,0 +1,228 @@
...
+     * Takes a list of identified requests in addition to a destination and batches the requests into a batch request.
+     * It then sends the batch request and returns a CompletableFuture of the response.
+     */
+    private final BatchAndSend<RequestT, BatchResponseT> batchFunction;
+
+    /**
+     * Unpacks the batch response, then transforms individual entries to the appropriate response type. Each entry's batch ID
+     * is mapped to the individual response entry.
+     */
+    private final BatchResponseMapper<BatchResponseT, ResponseT> responseMapper;
```
dagnir, line 60: Since we agreed we don't need to support batch managers generically, is this complexity necessary? e.g. do we need to abstract away the concept of batching and sending (BatchSend) or can we just implement all of that internally in this class?

dagnir (commented):
Overall, it seems like we're still carrying over some of complexity from the original feature that intended to support batching across multiple services. Since we know that this is all going to be SQS specific can we simplify this design further, knowing that we just need to support SQS?

joviegas:
> Overall, it seems like we're still carrying over some of complexity from the original feature that intended to support batching across multiple services. Since we know that this is all going to be SQS specific can we simplify this design further, knowing that we just need to support SQS?


I removed the Generic Interface for BatchManager and will have sepaarte Batch Manager for Response and Request.
We still need the Generic for RequestBatch manger apis since we have 3 APIs in it , this complexity is still needed unless we decide we will break it out and do simple implementation as V1

services/sqs/src/main/java/software/amazon/awssdk/services/sqs/internal/batchmanager/core/RequestBatchManager.java
joviegas: Yeap the old generic design doesnot hold for ReceiveMessage but it holds good for 3 other APIs like ChangeVisibility , SendMessage and DeleteMessage the generics are used here to handle these 3 APIs

services/sqs/src/main/java/software/amazon/awssdk/services/sqs/internal/batchmanager/core/RequestBatchManager.java
```diff
@@ -0,0 +1,228 @@
...
+                                   Map<String, BatchingExecutionContext<RequestT, ResponseT>> flushableRequests) {
+        requestsAndResponsesMaps.cancelScheduledFlush(batchKey);
+        flushBuffer(batchKey, flushableRequests);
+        requestsAndResponsesMaps.putScheduledFlush(batchKey, scheduleBufferFlush(batchKey, maxBatchOpenInMs.toMillis(),
+                                                                                 scheduledExecutor));
+    }
+
+    // Flushes the buffer for the given batchKey and fills in the response map with the returned responses.
+    // Returns exception in completableFuture if batchingFunction.apply throws an exception.
+    private void flushBuffer(String batchKey, Map<String, BatchingExecutionContext<RequestT, ResponseT>> flushableRequests) {
```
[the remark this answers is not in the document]
joviegas: We need to keep the track of individual requests that were send and update those requests with its corresponding response. While sending we create unique ids for messages and while receiving we map it back.
Note that flushBuffer is done asynchronously and we are not returning any values but updating the CompletableFutures of the responses from the map, I can explain you this offline .

services/sqs/src/main/java/software/amazon/awssdk/services/sqs/internal/batchmanager/core/BatchBuffer.java
```diff
@@ -0,0 +1,80 @@
...
+
+
+/**
+ * Interface representing a buffer that handles batching of requests and responses.
+ *
+ * @param <RequestT>  the type of the request
+ * @param <ResponseT> the type of the response
+ */
+@SdkInternalApi
+public interface BatchBuffer<RequestT, ResponseT> {
```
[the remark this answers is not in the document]
joviegas: Agree, I removed this interface . I created it initially thinking this could be shared with ResponseManager but response batch manager buffer behaviour is different so  we dont need the interface.

services/sqs/src/main/java/software/amazon/awssdk/services/sqs/internal/batchmanager/core/BatchKeyMapper.java
```diff
@@ -0,0 +1,29 @@
...
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.services.sqs.internal.batchmanager.core;
+
+import software.amazon.awssdk.annotations.SdkInternalApi;
+
+/**
+ * Takes a request and extracts a batchKey as determined by the caller.
+ * TODO: For right now, the batchKey is a String but this may change as needed in the future.
```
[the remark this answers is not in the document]
joviegas: Yeap I have create `JAVA-6450 ` will close this task before Integ and test case addition task

services/sqs/src/main/java/software/amazon/awssdk/services/sqs/internal/batchmanager/SqsBatchFunctions.java
```diff
@@ -0,0 +1,316 @@
...
+import software.amazon.awssdk.services.sqs.model.SendMessageBatchRequestEntry;
+import software.amazon.awssdk.services.sqs.model.SendMessageBatchResponse;
+import software.amazon.awssdk.services.sqs.model.SendMessageBatchResultEntry;
+import software.amazon.awssdk.services.sqs.model.SendMessageRequest;
+import software.amazon.awssdk.services.sqs.model.SendMessageResponse;
+import software.amazon.awssdk.services.sqs.model.SqsException;
+import software.amazon.awssdk.utils.Either;
+
+@SdkInternalApi
+public final class SqsBatchFunctions {
```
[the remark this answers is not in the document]
joviegas: This had private constructor  , anyways made it abstract now.

Do you want to make it abstract and make DefaultSqsAsyncBatchManager extend SqsBatchFunctions ?

services/sqs/src/main/java/software/amazon/awssdk/services/sqs/internal/batchmanager/RequestBatchManager.java
```diff
@@ -0,0 +1,198 @@
...
+import software.amazon.awssdk.utils.Validate;
+
+@SdkInternalApi
+public abstract class RequestBatchManager<RequestT, ResponseT, BatchResponseT> {
+    private final int maxBatchItems;
+    private final Duration maxBatchOpenInMs;
+    private final BatchingMap<RequestT, ResponseT> requestsAndResponsesMaps;
+    private final BatchAndSend<RequestT, BatchResponseT> batchFunction;
+    private final BatchResponseMapper<BatchResponseT, ResponseT> responseMapper;
+    private final BatchKeyMapper<RequestT> batchKeyMapper;
```
dagnir, line 38: These should just be `protected abstract` methods:

```java
    protected abstract CompletableFuture<BatchResponseT> batchAndSend(List<IdentifiableMessage<RequestT>> identifiedRequests,
                                                                      String batchKey);

    protected abstract String getBatchKey(RequestT request);

    protected abstract List<Either<IdentifiableMessage<ResponseT>, IdentifiableMessage<Throwable>>> mapBatchResponse(BatchResponseT batchResponse);

```

services/sqs/src/main/java/software/amazon/awssdk/services/sqs/internal/batchmanager/RequestBatchManager.java
```diff
@@ -0,0 +1,198 @@
...
+        BatchConfiguration batchConfiguration = new BatchConfiguration(builder.overrideConfiguration);
+        this.requestsAndResponsesMaps = new BatchingMap<>(batchConfiguration.maxBatchKeys(),
+                                                          batchConfiguration.maxBufferSize(),
+                                                          RequestBatchBuffer::new);
+        this.maxBatchItems = batchConfiguration.maxBatchItems();
+        this.maxBatchOpenInMs = batchConfiguration.maxBatchOpenInMs();
+        this.batchFunction = Validate.notNull(builder.batchFunction, "Null batchFunction");
+        this.responseMapper = Validate.notNull(builder.responseMapper, "Null responseMapper");
+        this.batchKeyMapper = Validate.notNull(builder.batchKeyMapper, "Null batchKeyMapper");
+        this.scheduledExecutor = Validate.notNull(builder.scheduledExecutor, "Null scheduledExecutor");
```
dagnir, line 51: We shouldn't need these right? See comment above re: making these protected abstract methods

services/sqs/src/main/resources/codegen-resources/customization.config
```diff
@@ -12,6 +12,7 @@
         }
     },
 
-    "enableGenerateCompiledEndpointRules": true
+    "enableGenerateCompiledEndpointRules": true,
+    "batchManagerSupported": true
```
dagnir, line 16: Should this be more generic now that batch managers could differ between services, e.g. where the actual FQCN, construct method, and argument are all specified? Maybe similar to the way the `utilities()` method is generated? 

https://github.com/aws/aws-sdk-java-v2/blob/4f967b987d651a8438fff99604127f19df676218/services/s3/src/main/resources/codegen-resources/customization.config#L285

services/sqs/src/main/java/software/amazon/awssdk/services/sqs/internal/batchmanager/BatchingMap.java
```diff
@@ -0,0 +1,89 @@
...
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ScheduledFuture;
+import java.util.function.BiConsumer;
+import java.util.function.BiFunction;
+import java.util.function.Supplier;
+import software.amazon.awssdk.annotations.SdkInternalApi;
+
+/**
+ * Outer map maps a batchKey (ex. queueUrl, overrideConfig etc.) to a nested BatchingGroupMap map.
```
dagnir, line 28: What is `BatchingGroup`? Is this supposed to be `RequestBatchBuffer`?

services/sqs/src/main/java/software/amazon/awssdk/services/sqs/internal/batchmanager/RequestBatchManager.java
```diff
@@ -0,0 +1,198 @@
...
+        if (!flushableRequests.isEmpty()) {
+            manualFlushBuffer(batchKey, flushableRequests);
+        }
+    }
+
+    private void manualFlushBuffer(String batchKey,
+                                   Map<String, BatchingExecutionContext<RequestT, ResponseT>> flushableRequests) {
+        requestsAndResponsesMaps.cancelScheduledFlush(batchKey);
+        flushBuffer(batchKey, flushableRequests);
+        requestsAndResponsesMaps.putScheduledFlush(batchKey, scheduleBufferFlush(batchKey, maxBatchOpenInMs.toMillis(),
```
dagnir, line 81: What happens if the flush can't be cancelled because it's already running?

services/sqs/src/main/java/software/amazon/awssdk/services/sqs/internal/batchmanager/RequestBatchBuffer.java
```diff
@@ -0,0 +1,137 @@
...
+        }
+        return Integer.toString(nextBatchEntry++);
+    }
+
+    public void putScheduledFlush(ScheduledFuture<?> scheduledFlush) {
+        this.scheduledFlush = scheduledFlush;
+    }
+
+    public void cancelScheduledFlush() {
+        scheduledFlush.cancel(false);
```
dagnir, line 124: This class still seems a little weird to me since it tells you what entries are flushable, and knows about removes about entries that are about to be flushed, but doesn't actually perform the flushing.

Probably not easy to refactor this without prolonging this review though. Can we add a tracking item to possibly refactor this in the future?

joviegas: Created `JAVA-7832` for this. Current IM was based on reusing as much as possible from the Project branch , we will need more time to refactor this , created a task for this , WDYT if we refactor this class after the release , else it will delay the release of the feature

services/sqs/src/main/java/software/amazon/awssdk/services/sqs/internal/batchmanager/RequestBatchManager.java
joviegas: done

services/sqs/src/main/java/software/amazon/awssdk/services/sqs/internal/batchmanager/RequestBatchManager.java
joviegas: yeap , removed these

services/sqs/src/main/java/software/amazon/awssdk/services/sqs/internal/batchmanager/BatchingMap.java
joviegas: Typo error

services/sqs/src/main/java/software/amazon/awssdk/services/sqs/internal/batchmanager/RequestBatchManager.java
joviegas: Currently it would just wait until response is recieved and relied on user to cancel it or complete it , but now I have fixed it to explicitly cancel them whenever BatchManager is closed

services/sqs/src/main/java/software/amazon/awssdk/services/sqs/internal/batchmanager/ChangeMessageVisibilityBatchManager.java
```diff
@@ -0,0 +1,156 @@
...
+
+    @Override
+    protected CompletableFuture<ChangeMessageVisibilityBatchResponse> batchAndSend(
+        List<IdentifiableMessage<ChangeMessageVisibilityRequest>> identifiedRequests, String batchKey) {
+        return changeMessageVisibilityBatchAsyncFunction(sqsAsyncClient).batchAndSend(identifiedRequests, batchKey);
+    }
+
+    @Override
+    protected String getBatchKey(ChangeMessageVisibilityRequest request) {
+        return changeMessageVisibilityBatchKeyMapper().getBatchKey(request);
```
dagnir, line 148: Can we just implement the logic directly in this method instead of using the lambda returned by `changeMessageVisibilityBatchKeyMapper()`? I think we can just get rid of these interfaces (`BatchAndSend`, `BatchKeyMapper`, `BatchResponseMapper`)

```java
    @Override
    protected String getBatchKey(ChangeMessageVisibilityRequest request) {
        return request.overrideConfiguration()
                      .map(overrideConfig -> request.queueUrl() + overrideConfig.hashCode())
                      .orElse(request.queueUrl());
    }
```

similar comment for `mapBatchResponse()`

joviegas: +1 , removed these classes , Thanks

dagnir (approved):

services/sqs/src/main/java/software/amazon/awssdk/services/sqs/internal/batchmanager/RequestBatchBuffer.java
joviegas: @dagnir To avoid the confusing refactored the method name to make it as clear we are extracting the buffered entries and actually flushing which is done in RequestBatchManager https://github.com/aws/aws-sdk-java-v2/pull/6485

PR #5429, opened 2024-07-25
S3 PutObject String overload

davidh44:
## Motivation and Context

Transforming `putObject` overload for String payload:
`putObject(String bucketName, String key, String content)`

## Testing

Added unit and end-to-end tests

### DIFF
--- a/v2-migration/src/main/java/software/amazon/awssdk/v2migration/S3StreamingRequestToV2.java
+++ b/v2-migration/src/main/java/software/amazon/awssdk/v2migration/S3StreamingRequestToV2.java
@@ -94,6 +94,7 @@ public class S3StreamingRequestToV2 extends Recipe {
             Expression getObjectExpr = bucketAndKeyToPutObject(bucketExpr, keyExpr);
             newArgs.add(getObjectExpr);
 
+            // This is to maintain the formatting/spacing of original code, getPrefix() retrieves the leading whitespace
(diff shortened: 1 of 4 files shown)


v2-migration/src/main/java/software/amazon/awssdk/v2migration/S3StreamingRequestToV2.java
```diff
@@ -66,12 +69,78 @@ public TreeVisitor<?, ExecutionContext> getVisitor() {
...
+
+            Expression bucketExpr = originalArgs.get(0);
+            Expression keyExpr = originalArgs.get(1);
+            Expression stringExpr = originalArgs.get(2);
+
+            List<Expression> newArgs = new ArrayList<>();
+            Expression getObjectExpr = bucketAndKeyToPutObject(bucketExpr, keyExpr);
+            newArgs.add(getObjectExpr);
+
+            Space stringArgPrefix = stringExpr.getPrefix();
```
anirudh9391, line 97: nit : could we add a short comment on what we are doing here and why ? I get that we are separating the prefix and the string with a space, is that correct ?

davidh44: added comment, we are preserving the original formatting

anirudh9391 (approved):

Issue #4935, opened 2024-02-15
StackOverflowError on Alpine when using DefaultS3CrtAsyncClient


ffried:
### Describe the bug

Using the `DefaultS3CrtAsyncClient` as produced by `S3AsyncClient.crtBuilder().build()` fails with a `StackOverflowError` on Alpine when running a `ListObjectsRequest`.

Using `S3AsyncClient.builder().build()` (no CRT) succeeds.

### Expected Behavior

The CRT client works just as fine as the normal client (if not better 😉).

### Current Behavior

A `StackOverflowError` is raised.

### Reproduction Steps

Reproduction files:

<details>
<summary><code>Main.java</code></summary>

```java
package de.sersoft;

import software.amazon.awssdk.regions.Region;
import software.amazon.awssdk.services.s3.S3AsyncClient;

public class Main {
    public static void main(String[] args) {
        var bucketName = args[0];
        try (var s3 = S3AsyncClient.crtBuilder().region(Region.US_EAST_1).build()) {
            System.out.println("Listing objects in bucket " + bucketName);
            var objects = s3.listObjects(builder -> builder.bucket(bucketName)).join().contents();
            System.out.println("Objects in bucket " + bucketName + ":");
            for (var myValue : objects) {
                System.out.println("-" + myValue.key());
            }
        }
    }
}
```
</details>

<details>
<summary><code>pom.xml</code></summary>

```pom.xml
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0"
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
    <modelVersion>4.0.0</modelVersion>

    <groupId>de.sersoft</groupId>
    <artifactId>AWSAlpine</artifactId>
    <version>1.0.0-SNAPSHOT</version>
    <packaging>jar</packaging>

    <properties>
        <maven.compiler.source>21</maven.compiler.source>
        <maven.compiler.target>21</maven.compiler.target>
        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
    </properties>

    <dependencyManagement>
        <dependencies>
            <dependency>
                <groupId>software.amazon.awssdk</groupId>
                <artifactId>bom</artifactId>
                <version>2.22.2</version>
                <type>pom</type>
[...]

PR #5433, opened 2024-07-26
Bump CRT version to 0.30.3

zoewangg:
## Motivation and Context
Bump CRT version to 0.30.3 which contains fix for StackOverflowError on Alpine for CRT client.
Fix https://github.com/aws/aws-sdk-java-v2/issues/4935


embano1:
Thx for the work! Would be great to get this in since we're pending on this one as well to move to the official Kafka docker images (using Alpine)

PR #5471, opened 2024-08-06
Added QueueAttributesManager 

joviegas:
## Motivation and Context

- We need new option for ReceiveBatchManager thus added those in BatchOverrideConfiguration. 
- We need Queue Attributes like ReceiveWaitTimeOut and Visibility Timeout while doing `receiveMessage()` thus added QueueAttributesManager class
## Modifications

# QueueAttributesManager

## Overview

`QueueAttributesManager` is a utility class responsible for managing the visibility timeout and receive message wait time for an Amazon SQS queue. It leverages asynchronous operations to fetch and cache these configurations, ensuring efficient use of resources by minimizing repeated API calls.

## Key Responsibilities

- **Fetch Visibility Timeout**: Retrieves the visibility timeout for the SQS queue and caches it to avoid repeated API calls.
- **Fetch Receive Message Wait Time**: Retrieves the receive message wait time for the SQS queue and caches it similarly.
- **Calculate Wait Time**: Calculates the effective wait time for receiving messages, ensuring it meets a specified minimum duration.

## Implementation Details

- **Asynchronous Fetching**: Utilizes `CompletableFuture` to fetch configurations asynchronously, ensuring non-blocking operations.
- **Caching**: Uses `AtomicReference<CompletableFuture<Duration>>` to cache the results of the fetch operations, preventing redundant API calls.
- **Validation**: Ensures that the fetched attributes are valid and not null, throwing appropriate exceptions if validation fails.
- **Concurrency Handling**: Manages concurrent access to the cached values, ensuring thread-safety.

### `getReceiveMessageTimeout(ReceiveMessageRequest rq)`
Fetches or retrieves the cached receive message wait time and calculates the effective wait time for the given request.

### `getVisibilityTimeout()`
Fetches or retrieves the cached visibility timeout for the SQS queue.

### `fetchQueueWaitTime(CompletableFuture<Duration> newWaitTimeFuture)`
Helper method to asynchronously fetch the receive message wait time from SQS and complete the provided future.

### `fetchVisibilityTimeout(CompletableFuture<Duration> newTimeoutFuture)`
Helper method to asynchronously fetch the visibility timeout from SQS and complete the provided future.

### `calculateWaitTime(ReceiveMessageRequest rq, Duration defaultWaitTime)`
Calculates the effective wait time for receiving messages, ensuring it meets the minimum specified wait time.

## Context

The `QueueAttributesManager` is used when performing a `ReceiveMessage` operation to get the `ReceiveMessageWaitTimeSeconds` and `VisibilityTimeout`. For more details on these attributes, refer to the [AWS SQS API Reference](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_SetQueueAttributes.html).

## Benefits

- **Efficiency**: Reduces redundant API calls by caching configuration values.
- **Non-blocking**: Uses asynchronous operations to fetch values without blocking the main thread.
- **Thread-safe**: Ensures thread-safe access to cached values.

## Testing
- Junits added

## License


services/sqs/src/main/java/software/amazon/awssdk/services/sqs/internal/batchmanager/QueueAttributesManager.java
```diff
@@ -0,0 +1,153 @@
...
+import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.services.sqs.SqsAsyncClient;
+import software.amazon.awssdk.services.sqs.model.GetQueueAttributesRequest;
+import software.amazon.awssdk.services.sqs.model.QueueAttributeName;
+import software.amazon.awssdk.services.sqs.model.ReceiveMessageRequest;
+import software.amazon.awssdk.utils.Validate;
+
+
+@SdkInternalApi
+public class QueueAttributesManager {
```
dagnir, line 31: Should be final. Please also include a description

services/sqs/src/main/java/software/amazon/awssdk/services/sqs/internal/batchmanager/QueueAttributesManager.java
```diff
@@ -0,0 +1,153 @@
...
+        this.minReceiveWaitTime = minReceiveWaitTime;
+    }
+
+    /**
+     * Retrieves the received message timeout based on the provided request and queue attributes.
+     *
+     * @param rq The receive message request
+     * @return CompletableFuture with the calculated receive message timeout in milliseconds
+     */
+    public CompletableFuture<Duration> getReceiveMessageTimeout(ReceiveMessageRequest rq) {
```
dagnir, line 51: Should we short circuit here if `rq` has a custom wait time set? If so we don't need to know the default value right?

services/sqs/src/main/java/software/amazon/awssdk/services/sqs/internal/batchmanager/QueueAttributesManager.java
```diff
@@ -0,0 +1,153 @@
...
+                                              .get(QueueAttributeName.RECEIVE_MESSAGE_WAIT_TIME_SECONDS),
+                                          QueueAttributeName.RECEIVE_MESSAGE_WAIT_TIME_SECONDS +
+                                          " attribute is null in sqs.");
+
+                     return Duration.ofSeconds(Integer.parseInt(messageWaitTime));
+                 })
+                 .thenAccept(newWaitTimeFuture::complete)
+                 .exceptionally(ex -> {
+                     newWaitTimeFuture.completeExceptionally(ex);
+                     return null;
```
dagnir, line 90: Can we just combine this into a `whenComplete`?

services/sqs/src/main/java/software/amazon/awssdk/services/sqs/internal/batchmanager/QueueAttributesManager.java
```diff
@@ -0,0 +1,153 @@
...
+import software.amazon.awssdk.services.sqs.model.ReceiveMessageRequest;
+import software.amazon.awssdk.utils.Validate;
+
+
+@SdkInternalApi
+public class QueueAttributesManager {
+
+    private final SqsAsyncClient sqsClient;
+    private final String queueUrl;
+    private final AtomicReference<CompletableFuture<Duration>> defaultWaitTimeSecondsFuture = new AtomicReference<>();
```
zoewangg, line 35: Is there any reason we are storing the future? Can't we store the value directly?

services/sqs/src/main/java/software/amazon/awssdk/services/sqs/internal/batchmanager/QueueAttributesManager.java
zoewangg: yes, seems like we can just use `volatile Duration` for waitTimeSeconds and do something like the following


```
    public CompletableFuture<Duration> getReceiveMessageTimeout(ReceiveMessageRequest request) {
        if (request.waitTimeSeconds() != null) {
            return CompletableFuture.completedFuture(Duration.ofSeconds(request.waitTimeSeconds()));
        }

        if (defaultWaitTimeSecondsFuture == null) {
            synchronized (this) {
                if (defaultWaitTimeSecondsFuture == null) {
                    GetQueueAttributesRequest req = GetQueueAttributesRequest.builder()
                                                                             .queueUrl(queueUrl)
                                                                             .attributeNames(
                                                                                 QueueAttributeName.RECEIVE_MESSAGE_WAIT_TIME_SECONDS)
                                                                             .build();

                    return sqsClient.getQueueAttributes(req)
                                    .thenApply(response -> {
                                        String messageWaitTime =
                                            Validate.notNull(response
                                                                 .attributes()
                                                                 .get(QueueAttributeName.RECEIVE_MESSAGE_WAIT_TIME_SECONDS),
                                                             QueueAttributeName.RECEIVE_MESSAGE_WAIT_TIME_SECONDS +
                                                             " attribute is null in sqs.");

                                        Duration duration = Duration.ofSeconds(Integer.parseInt(messageWaitTime));
                                        defaultWaitTimeSecondsFuture = duration;
                                        return duration;
                                    });
                }
            }
        }
        
        return CompletableFuture.completedFuture(defaultWaitTimeSecondsFuture);
    }
```

services/sqs/src/main/java/software/amazon/awssdk/services/sqs/internal/batchmanager/QueueAttributesManager.java
```diff
@@ -0,0 +1,153 @@
...
+                                          QueueAttributeName.RECEIVE_MESSAGE_WAIT_TIME_SECONDS +
+                                          " attribute is null in sqs.");
+
+                     return Duration.ofSeconds(Integer.parseInt(messageWaitTime));
+                 })
+                 .thenAccept(newWaitTimeFuture::complete)
+                 .exceptionally(ex -> {
+                     newWaitTimeFuture.completeExceptionally(ex);
+                     return null;
+                 });
```
zoewangg, line 91: Should we set up future cancellation? If newWaitTimeFuture is cancelled, we should cancel `getQueueAttributes` future

services/sqs/src/main/java/software/amazon/awssdk/services/sqs/internal/batchmanager/QueueAttributesManager.java
```diff
@@ -0,0 +1,153 @@
...
+
+    /**
+     * Calculates the wait time for receiving a message, ensuring it meets the minimum wait time.
+     *
+     * @param rq              The receive message request
+     * @param defaultWaitTime The default wait time from the queue attributes
+     * @return The calculated wait time in milliseconds
+     */
+    private Duration calculateWaitTime(ReceiveMessageRequest rq, Duration defaultWaitTime) {
+        int waitTimeSeconds = (rq.waitTimeSeconds() != null) ? rq.waitTimeSeconds() : (int) defaultWaitTime.getSeconds();
```
zoewangg, line 102: nit: we can use `NumericUtils.saturatedCast` to cast long to int

services/sqs/src/main/java/software/amazon/awssdk/services/sqs/internal/batchmanager/QueueAttributesManager.java
zoewangg: NVM, got it now.

services/sqs/src/main/java/software/amazon/awssdk/services/sqs/internal/batchmanager/QueueAttributesManager.java
joviegas: No,  we need that because we take the maximum of the value 
https://github.com/aws/aws-sdk-java/blob/9acdd7a744fef85a56ca298a955890dce6111688/aws-java-sdk-sqs/src/main/java/com/amazonaws/services/sqs/buffered/ReceiveQueueBuffer.java#L182-L197

In order to determin max we have to pull the Value as configured on actual queue

joviegas: good catch

PR #5481, opened 2024-08-07
Allow calling `close` after `abort` on a `ResponseInputStream`.

millems:
Before this change, calling `close` after `abort` would raise an IOException. It's well-understood that streams should be `close`d after they're done being used. This was not possible before this change, if `abort` was called.

As a side effect of this change, `abort` now calls `close` on the underlying input stream. This is the behavior of the 1.x SDK, so it is low-risk.


core/sdk-core/src/main/java/software/amazon/awssdk/core/ResponseInputStream.java
```diff
@@ -57,10 +58,15 @@ public ResponseT response() {
...
+    /**
+     * Close the underlying connection, dropping all remaining data in the stream, and not leaving the
+     * connection open to be used for future requests.
+     */
     @Override
     public void abort() {
         if (abortable != null) {
             abortable.abort();
         }
+        IoUtils.closeQuietly(in, null);
```
L-Applin, line 70: This will log the potential error at `DEBUG` level, would we want a higher visibility level for that or no?

L-Applin (approved):
Like you said the risk of breaking change is low. We still log the error

core/sdk-core/src/main/java/software/amazon/awssdk/core/ResponseInputStream.java
millems: We use DEBUG for 1.x as well. In both 1.x and 2.x, it seems to always lead to a "ConnectionClosedException", which... yeah, that's what we wanted. I think logging it at a higher level than DEBUG is useful if there's some action that people could perform, but I don't know what we'd want people to do differently.

jigsaw18june: I looked at the IoUtils that is being used here is https://github.com/aws/aws-sdk-java-v2/blob/master/utils/src/main/java/software/amazon/awssdk/utils/IoUtils.java#L71. This one catches all exceptions instead of catching only IO exceptions. Is that intentional ? I think this might result in an unexpected behavior where we just catch all exceptions and silently log it instead of catching IO exceptions only and if consumers are expecting that behavior.

millems: I think it's a reasonable behavior to not fail on non-IO errors as well. This was the behavior in 1.x, and it didn't seem to cause any problem.

anil-kj: yes, that seemsreasonable . Although I did check SDK1 and its actually only catching IOException on `closeQuietly` in v1. If its not a problem then its ok to keep this behavior as is

PR #5494, opened 2024-08-09
Spotbugs rule for blocking calls

L-Applin:
Implementation of a new spotbugs/findbugs rule to detect blocking calls:
- `CompletableFuture.join()`
- `CompletableFuture.get()`
- `CompletableFutureUtils.joinLikeSync()`
- `Thread.sleep()`
- `Object.wait()`
- `HttpResourcesUtils.readResource()`
- `HttpCredentialsLoader.loadloadCredentials()`

 The rule will flag any method calls made from anywhere and fails the maven build. To supress legitimate calls made to these methods outside of the async code path, the `spotbugs-suppressions.xml` must be updated.

## Motivation and Context
As part of our effort to remove blocking call from the async code path, we need a way to detect existing blocking calls and prevent new blocking calls from making their way into the async code path.

## Modifications
New spotbugs/findbugs rule and suppression for it.

## Testing
Manually build the sdk with the new rule enabled, adding suppression for any the new rule findings


build-tools/src/main/java/software/amazon/awssdk/buildtools/findbugs/DisallowMethodCall.java
```diff
@@ -44,6 +45,18 @@ public class DisallowMethodCall extends OpcodeStackDetector {
...
         PROHIBITED_METHODS.add(new SimpleEntry<>("software/amazon/awssdk/http/SdkHttpFullRequest", "rawQueryParameters"));
         PROHIBITED_METHODS.add(new SimpleEntry<>("software/amazon/awssdk/http/SdkHttpFullRequest$Builder", "rawQueryParameters"));
+
+        // Blocking calls in async code path
+        PROHIBITED_ASYNC_BLOCKING_METHODS.add(new SimpleEntry<>("java/lang/Thread", "sleep"));
+        PROHIBITED_ASYNC_BLOCKING_METHODS.add(new SimpleEntry<>("java/util/concurrent/CompletableFuture", "join"));
+        PROHIBITED_ASYNC_BLOCKING_METHODS.add(new SimpleEntry<>("java/util/concurrent/CompletableFuture", "get"));
+        PROHIBITED_ASYNC_BLOCKING_METHODS.add(new SimpleEntry<>(
+            "software/amazon/awssdk/utils/CompletableFutureUtils", "joinLikeSync"));
+        PROHIBITED_ASYNC_BLOCKING_METHODS.add(new SimpleEntry<>(
```
anirudh9391, line 55: do we also want to include wait() - https://docs.oracle.com/javase/8/docs/api/java/lang/Object.html#:~:text=wait,-public%20final%20void&text=()%20throws%20InterruptedException-,Causes%20the%20current%20thread%20to%20wait%20until%20another%20thread%20invokes,must%20own%20this%20object's%20monitor.

L-Applin: added Object.wait(), good call

anirudh9391 (approved):

Issue #5382, opened 2024-07-09
DynamoDB enhanced client - Add "Select" in ScanEnhancedRequest (short issue description)


shetsa-amzn:
### Describe the feature

DynamoDB enhanced client - Add "Select" in ScanEnhancedRequest 

Similar to this issue: https://github.com/aws/aws-sdk-java-v2/issues/2083

### Use Case

ScanEnhancedRequest.builder().select(Select select)
The select should has those enums

    ALL_ATTRIBUTES
    ALL_PROJECTED_ATTRIBUTES
    SPECIFIC_ATTRIBUTES
    COUNT

### Acknowledgements

- [X] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### AWS Java SDK version used

2.26.16

### JDK version used

1.8

### Operating System and version

macOs

PR #5519, opened 2024-08-20
Adding support for Select in ScanEnhancedRequest

shetsa-amzn:
Adding support for Select in ScanEnhancedRequest

## Motivation and Context
https://github.com/aws/aws-sdk-java-v2/issues/5382

## Modifications
Adding support for Select Attribute in ScanEnhancedRequest

## Types of changes

## Checklist

- [] My change requires a change to the Javadoc documentation

## License


services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/operations/ScanOperation.java
```diff
@@ -82,6 +82,7 @@ public ScanRequest generateRequest(TableSchema<T> tableSchema,
             .exclusiveStartKey(this.request.exclusiveStartKey())
             .consistentRead(this.request.consistentRead())
             .returnConsumedCapacity(this.request.returnConsumedCapacity())
+            .select(this.request.select())
```
millems, line 85: In case the "select" enum value isn't known to the SDK version, it would be nice to use 'String' internally, like the code-generated models do. That way, customers that can't upgrade their SDK version can still specify new "select" options we might add in the future.

I see we're not doing that for other enums, and I don't think "select" changes much, so "in a separate PR" is totally fine.

shetsa-amzn: I think we should do as part of separate PR for all the Operations where select is used. AFAIK its for Query and Scan.

millems (approved):

PR #5534, opened 2024-08-26
Adding tests for overriding credentials or credentials providers 

cenedhryn:
The motivation for these tests are to show how the SDK supports use cases where customers want to override credentials per request through execution interceptors.

### DIFF
--- a/test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/IdentityResolutionOverrideTest.java
+++ b/test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/IdentityResolutionOverrideTest.java
@@ -1,0 +1,214 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.services;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.assertj.core.api.Assertions.fail;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+import java.util.function.Consumer;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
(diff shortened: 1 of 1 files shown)


test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/IdentityResolutionOverrideTest.java
```diff
@@ -0,0 +1,214 @@
...
+            .hasMessageContaining("stop");
+
+        assertSelectedAuthSchemeBeforeTransmissionContains(OVERRIDE_CREDENTIALS);
+    }
+
+    // Changing the credentials provider in modifyRequest does not work in SRA identity resolution
+    // Identity is resolved in beforeExecution (and happens before user applied interceptors) and cannot
+    // be affected by execution interceptors.
+    @Test
+    void when_credentialsProviderIsOverridden_inExecutionInterceptor_modifyRequest_itIsNotUsed() {
```
joviegas, line 78: In some test cases I see "itIsNotUsed" and "thayAreNotUsed"
You mean to say the overridden credentials in the Interceptors and if they are not used then what credentials get used 
like 
when_user_Provides_ProfileCredendial_And_InterceptorOverrideCredential_thenProfileCredential gets used. something like this would be helpful to clarify

joviegas:
How will the users discover this test cases , are we linking this file in Developer guide or any other places ?

test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/IdentityResolutionOverrideTest.java
```diff
@@ -0,0 +1,214 @@
...
+
+        @Override
+        public SdkRequest modifyRequest(Context.ModifyRequest context, ExecutionAttributes executionAttributes) {
+            SdkRequest request = context.request();
+            AwsRequest awsRequest = (AwsRequest) request;
+
+            AwsRequestOverrideConfiguration.Builder overrideConfigBuilder = getOrCreateOverrideConfig(request);
+            modifier.accept(overrideConfigBuilder);
+
+            //overrideConfigBuilder.putExecutionAttribute(AwsSignerExecutionAttribute.AWS_CREDENTIALS, overrideCredentials);
```
joviegas, line 184: Can we remove this comment ?

test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/IdentityResolutionOverrideTest.java
cenedhryn: I can update the language if this heading doesn't make sense

test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/IdentityResolutionOverrideTest.java
cenedhryn: Yes

cenedhryn:
It's not necessarily the customers who need to discover these tests, but that we didn't have test coverage for these scenarios before but there are legitimate use cases where these types of overrides are necessary.

joviegas (approved):

PR #5544, opened 2024-08-29
Refactor to avoid impacting JSON with RPCv2 logic

sugmanue:
## Motivation and Context

This change refactors the previously introduced (see #[5543](https://github.com/aws/aws-sdk-java-v2/pull/5539)) logic to handle RPCv2 unmarshalling to move it into the RPCv2 package and keeping the JSON unmarshalling logic as it was.

In order to accomplish this we changed the way the unmarshaller gets its registry to allow it to be pass down using the protocol factory builder. 

This will reduce the potential impact that this new protocol might have on the already existing ones.

## Types of changes

## Checklist

## License


core/protocols/aws-json-protocol/src/main/java/software/amazon/awssdk/protocols/json/BaseAwsJsonProtocolFactory.java
```diff
@@ -181,13 +183,6 @@ protected JsonContentTypeResolver getContentTypeResolver() {
         return AWS_JSON;
     }
 
-    /**
-     * @return Default JsonNode value factory.
-     */
-    protected JsonValueNodeFactory getJsonValueNodeFactory() {
```
sugmanue, line 187: Introduced in the previous PR, it can be removed now that we carry this in the builder and we don't need to expose it here.

core/protocols/aws-json-protocol/src/main/java/software/amazon/awssdk/protocols/json/BaseAwsJsonProtocolFactory.java
```diff
@@ -322,10 +319,40 @@ public final SubclassT hasAwsQueryCompatible(boolean hasAwsQueryCompatible) {
...
+         *
+         * @param timestampFormatRegistryFactory the factory to create an unmarshall registry
+         * @return This builder for method chaining.
+         */
+        protected final SubclassT timestampFormatRegistryFactory(
+            TimestampFormatRegistryFactory timestampFormatRegistryFactory
+        ) {
+            this.timestampFormatRegistryFactory = timestampFormatRegistryFactory;
+            return getSubclass();
+        }
```
millems, line 335: Is there a way to make this not specific to timestamps? It feels better to have protocol-specific marshalling for any type, not just Instants.

core/protocols/aws-json-protocol/src/main/java/software/amazon/awssdk/protocols/json/internal/unmarshall/TimestampFormatRegistryFactory.java
```diff
@@ -0,0 +1,34 @@
...
+
+package software.amazon.awssdk.protocols.json.internal.unmarshall;
+
+import java.util.Map;
+import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.core.protocol.MarshallLocation;
+import software.amazon.awssdk.core.traits.TimestampFormatTrait;
+
+/**
+ * Interface used to create an {@link DefaultJsonUnmarshallerRegistry} that will unmarshall using the given default formats for
```
millems, line 24: s/DefaultJsonUnmarshallerRegistry/JsonUnmarshallerRegistry/?

core/protocols/aws-json-protocol/src/main/java/software/amazon/awssdk/protocols/json/BaseAwsJsonProtocolFactory.java
sugmanue: Yeah, that would be ideal, not sure how though, suggestions?. I'm trying to generalize the functionality statically defined before in the [`createUnmarshallerRegistry`](https://github.com/aws/aws-sdk-java-v2/blob/3bc3427813e75a86530714075dcc67d37e1e9acd/core/protocols/aws-json-protocol/src/main/java/software/amazon/awssdk/protocols/json/internal/unmarshall/JsonProtocolUnmarshaller.java#L72)

sugmanue: Oh, the formats themselves are part of the protocol, so yeah, let me make that happen.

sugmanue: Pushed a commit to change this, now we take a single object with all the dependencies which include the registry.

millems (approved):

PR #5559, opened 2024-09-04
Fixed a bug in AttributeMap, which causes a ConcurrentModificationException…

millems:
… when lazy properties depend on lazy properties that depend on any other property.

This is achieved by changing a `compute` into a `get` followed by a `put`.

This code path was not hit before the endpoint URL changes, which are forthcoming. Those changes also add a path that tests this change.

### DIFF
--- a/utils/src/main/java/software/amazon/awssdk/utils/AttributeMap.java
+++ b/utils/src/main/java/software/amazon/awssdk/utils/AttributeMap.java
@@ -319,6 +319,7 @@ public final class AttributeMap implements ToCopyableBuilder<AttributeMap.Builde
             if (oldValue != null) {
                 dependencyGraph.valueUpdated(oldValue, value);
             }
+
         }
 
         /**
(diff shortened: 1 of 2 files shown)


utils/src/main/java/software/amazon/awssdk/utils/AttributeMap.java
```diff
@@ -319,6 +319,7 @@ private void internalPut(Key<?> key, Value<?> value) {
             if (oldValue != null) {
                 dependencyGraph.valueUpdated(oldValue, value);
             }
+
```
anirudh9391, line 322: nit : unintentional ?

utils/src/main/java/software/amazon/awssdk/utils/AttributeMap.java
```diff
@@ -327,17 +328,17 @@ private void internalPut(Key<?> key, Value<?> value) {
...
             checkCopyOnUpdate();
-            return attributes.compute(key, (k, v) -> {
-                if (v == null || resolveValue(v) == null) {
-                    Value<?> newValue = value.get();
-                    Validate.notNull(newValue, "Supplied value must not be null.");
-                    if (v != null) {
-                        dependencyGraph.valueUpdated(v, newValue);
-                    }
-                    return newValue;
+            Value<?> currentValue = attributes.get(key);
```
anirudh9391, line 331: Looks good. Is there any way in which we tested this ? If a unit test is difficult, A one-time runnable test that validates these changes would be helpful here. Lmk if its non-trivial to test / validate

utils/src/main/java/software/amazon/awssdk/utils/AttributeMap.java
millems: Fixed!

utils/src/main/java/software/amazon/awssdk/utils/AttributeMap.java
millems: This new behavior is exercised in https://github.com/aws/aws-sdk-java-v2/pull/5562. Let me see if I can create a standalone test for it, for the purposes of this PR.

anirudh9391 (approved):

PR #5560, opened 2024-09-04
Update ProfileFile to support reading sub-properties.

millems:
Sub-properties are those nested under another property. E.g.
```
[default]
s3 =
  endpoint_url = foo
```

Before this change, only `s3` would be available, and would contain the raw string underneath the property. After this change, `s3` remains the same, but `foo` can be retrieved through the property `s3.endpoint_url`.

This change also adds the property for `endpoint_url` and a convenience setter for profile content strings.


zoewangg (approved):

zoewangg (commented):
Can we add a changelog entry?

Issue #3894, opened 2023-04-07
VersionedRecordExtension does not support 0 as a starting value


akiesler:
### Describe the bug

The VersionedRecordExtension says it will increment a version and verify that the version in DynamoDB is equal to the previous value when updating. However, the documentation is unclear that the starting value must be `null` and cannot be `0` which would be a reasonable starting version. Using a value of `0` results in the conditional check failing when attempting to put the item.

### Expected Behavior

I would expect the extension to treat `0` the same as `null` acting as a clear starting point for the versioning. This would result in an update expression where the version attribute is checked not to exist before updating.

### Current Behavior

The DynamoDB client throws a ConditionalUpdateFailedException because the the version field does not exist in the record.

### Reproduction Steps

A simple test case modified from the SDK:
```
   @Test
    public void beforeWrite_initialVersionDueToExplicitZero_expressionAndTransformedItemIsCorrect() {
        FakeItem fakeItem = createUniqueFakeItem();

        Map<String, AttributeValue> inputMap =
            new HashMap<>(FakeItem.getTableSchema().itemToMap(fakeItem, true));
        inputMap.put("version", AttributeValue.builder().n("0").build());

        Map<String, AttributeValue> fakeItemWithInitialVersion =
            new HashMap<>(FakeItem.getTableSchema().itemToMap(fakeItem, true));
        fakeItemWithInitialVersion.put("version", AttributeValue.builder().n("1").build());

        WriteModification result =
            versionedRecordExtension.beforeWrite(DefaultDynamoDbExtensionContext
                                                     .builder()
                                                     .items(inputMap)
                                                     .tableMetadata(FakeItem.getTableMetadata())
                                                     .operationContext(PRIMARY_CONTEXT).build());

        assertThat(result.transformedItem(), is(fakeItemWithInitialVersion));
        assertThat(result.additionalConditionalExpression(),
                   is(Expression.builder()
                                .expression("attribute_not_exists(#AMZN_MAPPED_version)")
                                .expressionNames(singletonMap("#AMZN_MAPPED_version", "version"))
                                .build()));
    }
```

### Possible Solution

Update the check in `beforeWrite` for the VersionedRecordExtension to include a check for `0`.
If there is concern that this may break backwards compatibility it could be added behind a feature flag in the extension itself.

### AWS Java SDK version used

2.20.43

### JDK version used

openjdk version "1.8.0_362"

### Operating System and version

macOS 12.6.4

[...]

PR #5565, opened 2024-09-05
DynamoDB Enhanced Client Versioned Record start at 0

akiesler:
This allows for clients using the DynamoDB Versioned Record Extension to start their versions at `0` instead of `null` thus preventing clients from being required to use the `Integer` type and instead allowing them to use the `int` primitive.
This change also makes the extension more flexible by allowing clients to set the starting value and incrementing value if they have a use-case to do so. 

## Motivation and Context

Most developers would expect a version to start at 0 and increment from their instead of having to have a special case where the value must be initialized to null (see https://github.com/aws/aws-sdk-java-v2/issues/3894).

## Modifications

Updates the internal logic of `VersionedRecordExtension` to accept either a null value or a numeric value equal to the starting value.

## Testing

Added a unit tests for explicitly setting the version to 0 during initial creation.

## Types of changes

## Checklist

## License


akiesler:
@dagnir could you please review this feature?

debora-ito:
Superseded by https://github.com/aws/aws-sdk-java-v2/pull/6019.

PR #5571, opened 2024-09-06
Fix issue where the Scheduled Timeout was incorrectly completing the futures with empty messages

joviegas:
## Motivation and Context

The key modification was ensuring that the same future passed to `receiveQueueBuffer.receiveMessage` is the one that gets completed, regardless of whether a timeout occurs.

Previously, a new future (timeoutFuture) was scheduled to complete after the timeout period, but this future was independent of the one added to the futures queue. Now, instead of creating a new timeout future, we directly complete the original future (receiveMessageFuture) if it hasn't already been completed before the timeout occurs.

### Root Cause

The issue occurred because a new CompletableFuture (timeout future) was being created and completed during timeouts, rather than completing the original future added to the ReceiveBuffer. This left the original future uncompleted, causing the buffer to return empty responses.

### Fix

The fix ensures that the same `receiveMessageFuture`passed to the buffer is completed, even when a timeout occurs. The timeout logic now checks if the original future is complete and directly completes it, avoiding the creation of a separate timeout future. This ensures that the buffer behaves correctly and returns the expected messages.

## Modifications

The key modification was ensuring that the same future passed to `receiveQueueBuffer.receiveMessage` is the one that gets completed, regardless of whether a timeout occurs.

Previously, a new future (timeoutFuture) was scheduled to complete after the timeout period, but this future was independent of the one added to the futures queue. Now, instead of creating a new timeout future, we directly complete the original future (receiveMessageFuture) if it hasn't already been completed before the timeout occurs.

- Also we need to honor the user passed waitReceiveTime as the timeoOut time , same as done in V1 , thus removed bypassing .

## Testing

- Integ test
- Junits added

## License

### DIFF
--- a/services/sqs/src/main/java/software/amazon/awssdk/services/sqs/internal/batchmanager/ReceiveBatchManager.java
+++ b/services/sqs/src/main/java/software/amazon/awssdk/services/sqs/internal/batchmanager/ReceiveBatchManager.java
@@ -60,11 +60,9 @@ public class ReceiveBatchManager implements SdkAutoCloseable {
         return queueAttributesManager.getReceiveMessageTimeout(rq, config.messageMinWaitDuration()).thenCompose(waitTimeMs -> {
             CompletableFuture<ReceiveMessageResponse> receiveMessageFuture = new CompletableFuture<>();
             receiveQueueBuffer.receiveMessage(receiveMessageFuture, numMessages);
-            executor.schedule(() -> {
-                if (!receiveMessageFuture.isDone()) {
-                    receiveMessageFuture.complete(ReceiveMessageResponse.builder().build());
-                }
-            }, waitTimeMs.toMillis(), TimeUnit.MILLISECONDS);
+            executor.schedule(() -> receiveMessageFuture.complete(ReceiveMessageResponse.builder().build()),
+                              waitTimeMs.toMillis(),
+                              TimeUnit.MILLISECONDS);
             return receiveMessageFuture;
 
         });
--- a/services/sqs/src/main/java/software/amazon/awssdk/services/sqs/internal/batchmanager/ReceiveBatchManager.java
+++ b/services/sqs/src/main/java/software/amazon/awssdk/services/sqs/internal/batchmanager/ReceiveBatchManager.java
@@ -20,7 +20,6 @@ import static software.amazon.awssdk.services.sqs.internal.batchmanager.Response
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
-import java.util.function.Function;
 import software.amazon.awssdk.annotations.SdkInternalApi;
 import software.amazon.awssdk.services.sqs.SqsAsyncClient;
 import software.amazon.awssdk.services.sqs.model.ReceiveMessageRequest;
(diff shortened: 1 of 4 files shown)


services/sqs/src/main/java/software/amazon/awssdk/services/sqs/internal/batchmanager/ReceiveBatchManager.java
```diff
@@ -61,10 +60,12 @@ public CompletableFuture<ReceiveMessageResponse> processRequest(ReceiveMessageRe
         return queueAttributesManager.getReceiveMessageTimeout(rq, config.messageMinWaitDuration()).thenCompose(waitTimeMs -> {
             CompletableFuture<ReceiveMessageResponse> receiveMessageFuture = new CompletableFuture<>();
             receiveQueueBuffer.receiveMessage(receiveMessageFuture, numMessages);
-            CompletableFuture<ReceiveMessageResponse> timeoutFuture = new CompletableFuture<>();
-            executor.schedule(() -> timeoutFuture.complete(ReceiveMessageResponse.builder().build()), waitTimeMs.toMillis(),
-                              TimeUnit.MILLISECONDS);
-            return receiveMessageFuture.applyToEither(timeoutFuture, Function.identity());
+            executor.schedule(() -> {
+                if (!receiveMessageFuture.isDone()) {
```
zoewangg, line 64: I don't think we need to check isDone here, complete will handle it.

services/sqs/src/main/java/software/amazon/awssdk/services/sqs/internal/batchmanager/ReceiveMessageBatchManager.java
```diff
@@ -91,9 +91,6 @@ private String checkBatchingEligibility(ReceiveMessageRequest rq) {
         if (rq.overrideConfiguration().isPresent()) {
             return "Request has override configurations.";
         }
-        if (rq.waitTimeSeconds() != null && rq.waitTimeSeconds() != 0) {
```
zoewangg, line 94: Why are we removing this?

services/sqs/src/main/java/software/amazon/awssdk/services/sqs/internal/batchmanager/ReceiveBatchManager.java
joviegas: If the user passes this value then user can control the Polling time out , this is behaviour same as https://github.com/aws/aws-sdk-java/blob/61d73631fac8535ad70666bbce9e70a1d2cea2ca/aws-java-sdk-sqs/src/main/java/com/amazonaws/services/sqs/buffered/ReceiveQueueBuffer.java#L192-L197

I removed this in earlier PR but when I was doing V1 comparison I realized that V1 uses this value to determine the Client side Wait time for that  particular request.

services/sqs/src/main/java/software/amazon/awssdk/services/sqs/internal/batchmanager/ReceiveMessageBatchManager.java
joviegas: If the user passes this value then user can control the Polling time out , this is behaviour same as https://github.com/aws/aws-sdk-java/blob/61d73631fac8535ad70666bbce9e70a1d2cea2ca/aws-java-sdk-sqs/src/main/java/com/amazonaws/services/sqs/buffered/ReceiveQueueBuffer.java#L192-L197

I removed this in earlier PR but when I was doing V1 comparison I realized that V1 uses this value to determine the Client side Wait time for that  particular request.

joviegas: We already supported in V2 https://github.com/aws/aws-sdk-java-v2/blob/dd9379d67d08d932eeda472e6616e4e06b893bcd/services/sqs/src/main/java/software/amazon/awssdk/services/sqs/internal/batchmanager/QueueAttributesManager.java#L76-L82
But after handling the Surface API review comments where I added a check to bypass, I went ahead and added check to this since it was an additional Attribute related to long polling but when used with Automatic Batch manager we use it client side polling time so thus we need this to go through the batch manager

zoewangg (approved):

L-Applin (approved):

PR #5593, opened 2024-09-12
Fixed EnhancedClient UpdateItem operation to make it work on nested attributes as well

anirudh9391:
Modified Enhanced Client to facilitate nested update operations. 

## Motivation and Context
Customer reported an issue where UpdateItem does not work for nested attributes. 

## Modifications
Modified UpdateItem query path to form the right updateQuery for nested update attributes. 
Previously nested update queries looked like 
```
SET #AMZN_MAPPED_mainAddress = :AMZN_MAPPED_mainAddress
:AMZN_MAPPED_mainAddress -> "AttributeValue(M={state=AttributeValue(S=WA)})"
```

This operates like a putItem and updates the map with a new state. 

Intead modified the update query to 
```
aws dynamodb update-item \
    --table-name TEST_TABLE3 \
    --key '{"id":{"N":"111"}}' \
    --region us-west-2 \
    --update-expression "SET #AMZN_MAPPED_mainAddress.#AMZN_MAPPED_city = :AMZN_MAPPED_mainAddress_city" \
    --expression-attribute-values '{":AMZN_MAPPED_mainAddress_city": {"S": "Salem"}}' \  
    --expression-attribute-names '{"#AMZN_MAPPED_mainAddress": "mainAddress","#AMZN_MAPPED_city":"city"}'
```

### Public APIs

1) ignoreNullsMode(IgnoreNullsMode ignoreNullsMode) - Sets the ignoreNullsMode, defining the configuration in which the update is to be carried out. 

Usage : 
```
dynamoDBTable.updateItem(r -> r.item(update_record).ignoreNullsMode(IgnoreNullsMode.SCALAR_ONLY));
```
2) public IgnoreNullsMode ignoreNullsMode() - Retrieves the mode of usage

Usage : 
```
updateItemEnhancedRequest.ignoreNullsMode() == IgnoreNullsMode.SCALAR_ONLY ? : clause1 : clause2
```

## Testing
Unit Tests pass

## Types of changes


services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/operations/UpdateItemOperation.java
```diff
@@ -90,7 +94,21 @@ public UpdateItemRequest generateRequest(TableSchema<T> tableSchema,
...
+
+        if (ignoreNullsMode == IgnoreNullsMode.SCALAR_ONLY) {
+            ignoreNulls = true;
+        }
+        Map<String, AttributeValue> itemMapImmutable = tableSchema.itemToMap(item, Boolean.TRUE.equals(ignoreNulls));
+        
+        // If ignoreNulls is set to true, check for nested params to be updated
+        // If needed, Transform itemMap for it to be able to handle them.
+
+        Map<String, AttributeValue> itemMap = ignoreNullsMode == IgnoreNullsMode.SCALAR_ONLY ?
```
anirudh9391, line 109: This line is different from the approved PR

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/model/IgnoreNullsMode.java
```diff
@@ -0,0 +1,32 @@
+/*
```
anirudh9391, line 1: This interface is new

services-custom/dynamodb-enhanced/src/test/java/software/amazon/awssdk/enhanced/dynamodb/functionaltests/UpdateBehaviorTest.java
```diff
@@ -145,6 +162,355 @@ public void updateBehaviors_transactWriteItems_secondUpdate() {
         assertThat(persistedRecord.getCreatedAutoUpdateOn()).isEqualTo(firstUpdatedRecord.getCreatedAutoUpdateOn());
     }
 
+    @Test
```
anirudh9391, line 165: I have modified and added more tests

joviegas:
Since we are adding Public APIS for this could you please mention the Public APIS in PR description and user experience in PR description

services-custom/dynamodb-enhanced/src/test/java/software/amazon/awssdk/enhanced/dynamodb/functionaltests/UpdateBehaviorTest.java
```diff
@@ -145,6 +162,355 @@ public void updateBehaviors_transactWriteItems_secondUpdate() {
...
+        update_record.setNestedRecord(updatedNestedRecord);
+
+        mappedTable.updateItem(r -> r.item(update_record).ignoreNullsMode(IgnoreNullsMode.DEFAULT));
+
+        RecordWithUpdateBehaviors persistedRecord = mappedTable.getItem(r -> r.key(k -> k.partitionValue("id123")));
+
+        verifySingleLevelNestingDefaultUpdateBehavior(persistedRecord.getNestedRecord(), updatedNestedCounter);
+    }
+
+    @Test
```
joviegas, line 219: Can we have parameterized test so that,s it easy to find  expected behaviours

```java
public static List<Arguments> getDifferentCombinationOfignoreNullsModeAndExpectedResult() {
        return Arrays.asList(
            Arguments.of(simpleNestedAttrib, enum1, expectedMap1),
  Arguments.of(simpleNestedAttrib, enum2, expectedMap1),
  Arguments.of(nestedAttribWithEmpty, enum2, expectedMap1),
  
  etc etc


@ParameterizedTest
@MethodSource("getDifferentCombinationOfignoreNullsModeAndExpectedResult")
void testUpdateWithDifferentModes(NestedAttibute attrib,  IgnoreNullsMode ignoreNullsMode, Map<String, AttributeVale> expectedUpdate)
{
//existing common code
        mappedTable.updateItem(r -> r.item(attrib).ignoreNullsMode(ignoreNullsMode));

Map<String, AttributeVale> valueInDDB = nativeClient.get(item);

assertThat(expectedUpdate).isEqualTo(valueInDDB)

}
```

anirudh9391: This is not going to make it neater. For each case, i.e. simple nested attribute, empty attribute or a multi nested attribute, the validation functions are different. If at all we group some methods together, it will not make it neater.

anirudh9391: To make it more readable, I have grouped some of the validation methods together in the next revision into : 

- verifyMultipleLevelNestingTargetedUpdateBehavior
- verifySingleLevelNestingTargetedUpdateBehavior

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/model/IgnoreNullsMode.java
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * This enum offers users different modes of performing DDB item updates
+ * <p>
+ * In the SCALAR_ONLY mode, updates to nested scalar attributes are supported
+ * <p>
+ * In the MAPS_ONLY mode, updates to nested map structures are supported
+ * <p>
+ * The DEFAULT mode operates by setting ignoreNulls to false, and requires the user to
+ * fetch existing DDB item, make modifications to it and then update the item
+ */
```
dagnir, line 29: This probably needs more detail, especially regarding tradeoffs

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/model/IgnoreNullsMode.java
```diff
@@ -0,0 +1,35 @@
...
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.enhanced.dynamodb.model;
+
+import software.amazon.awssdk.annotations.SdkPublicApi;
+
+/**
+ * This enum offers users different modes of performing DDB item updates
```
dagnir, line 21: > different modes of performing DDB item updates

This statement seems too generic

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/model/IgnoreNullsMode.java
```diff
@@ -0,0 +1,35 @@
...
+ * In the MAPS_ONLY mode, updates to nested map structures are supported
+ * <p>
+ * The DEFAULT mode operates by setting ignoreNulls to false, and requires the user to
+ * fetch existing DDB item, make modifications to it and then update the item
+ */
+@SdkPublicApi
+public enum IgnoreNullsMode {
+    SCALAR_ONLY,
+    MAPS_ONLY,
+    DEFAULT
```
dagnir, line 34: What is the difference between MAPS_ONLY and DEFAULT? In the code, seems like we behave the same unless its `SCALAR_ONLY`

anirudh9391: I have modified this so that we do things differently. Basically, 

SCALAR_ONLY - when user wants to update nested attributes while only providing the changeset, i.e. today's ignoreNulls = true mode. 

MAPS_ONLY - This is a weird case that works the same for when ignoreNulls = true / false, i.e. when user provides only the new data, or all the data. The reason it works is because this mode ONLY supports updating with new maps, so the delta of changes constitutes the entire map. 

DEFAULT - when the user provides the whole item with the modified data. equivalent to ignoreNulls = false. 

We only need query transformations when the user sets SCALAR_ONLY mode and not otherwise. I had the modes defined jsut to make it clear. I will discuss a bit to see if we can remove it

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/model/IgnoreNullsMode.java
```diff
@@ -0,0 +1,37 @@
...
+
+package software.amazon.awssdk.enhanced.dynamodb.model;
+
+import software.amazon.awssdk.annotations.SdkPublicApi;
+
+/**
+ * <p>
+ * In the SCALAR_ONLY mode, updates to nested scalar attributes are supported in the ignoreNulls = true mode, i.e. when the user
+ * wants to update nested scalar attributes by providing only the delta of changes to be updated. This mode does not support
+ * updates to maps and is expected to throw a 400x DynamoDB exception if done so.
```
dagnir, line 24: nit: 400x looks like "400 times" to me. Should this be `4xx`?

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/operations/UpdateItemOperation.java
dagnir: So just to confirm, MAPS_ONLY mode is what we originally supported with `ignoreNulls == true`, and `SCALAR_ONLY` is new support we added?

anirudh9391: Yes that is correct. with MAPS_ONLY, essentially a new map is created as part of a every item update regardless of whether its scalar or non-scalar. This does not work for scalar updates, and hence the SCALAR_ONLY

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/model/IgnoreNullsMode.java
anirudh9391: Yeah i will change it

dagnir (approved):

PR #5596, opened 2024-09-13
Update user agent construction and deprecate old SdkUserAgent class

cenedhryn:
## Motivation and Context
New cross-SDK specifications changes the format of the Java user agent header. This change refactors Java SDK to conform to those specifications. 

### Creates a SystemUserAgent and deprecates SdkUserAgent
The existing `software.amazon.awssdk.core.util.SdkUserAgent` class is used for two types of requests:
1. Standard SDK client requests to AWS services
2. Other types of calls, usually to local HTTP endpoints

The latter can only use system level information in the user agent, while the former also can include client- and request-level information. 

The new `SystemUserAgent` stores a general system user agent string to be used in both use cases. The old user agent class is deprecated, because the name is confusing. `SystemUserAgent` has public methods to retrieve the individual system properties it supports, in addition to the system user agent string (which is a concatenated value). 

### Minor changes
- Removes system property `user.region` and adds `user.country`. The JVM doesn't populate user.region anymore, and the lang/region value (like `en_US`) has not been added to the user agent string for a while.
- Added a separate properties class `SdkClientUserAgentProperties.java` as input to building the main part of the SDK user agent string. It is currently a grab bag because at another point in the implementation it was stored in execution attributes. It can be a POJO as well, but the upcoming business metrics refactor also gives ample opportunities to refactor the code again. 
- Removes adding default values of empty string for `USER_AGENT_PREFIX` and `USER_AGENT_SUFFIX`. It's unclear why the implementation was that way. 

### How to handle old SdkUserAgent ### 
We can 
1. Deprecate (current choice)
2. Delete it
3. Log a warning if people use it.

## Types of changes

## Checklist

## License


core/sdk-core/src/main/java/software/amazon/awssdk/core/client/builder/SdkDefaultClientBuilder.java
```diff
@@ -283,8 +289,6 @@ private SdkClientConfiguration mergeGlobalDefaults(SdkClientConfiguration config
                                                   .lazyOption(PROFILE_FILE, conf -> conf.get(PROFILE_FILE_SUPPLIER).get())
                                                   .option(PROFILE_NAME,
                                                           ProfileFileSystemSetting.AWS_PROFILE.getStringValueOrThrow())
-                                                  .option(USER_AGENT_PREFIX, SdkUserAgent.create().userAgent())
-                                                  .option(USER_AGENT_SUFFIX, "")
```
cenedhryn, line 287: Can't see the motivation for this setup. We should try to figure out if this is a safe change.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/useragent/UserAgentLangValues.java
```diff
@@ -0,0 +1,92 @@
...
+        String scalaVersion = "";
+        try {
+            Class<?> scalaProperties = Class.forName("scala.util.Properties");
+            scalaVersion = "scala";
+            String version = (String) scalaProperties.getMethod("versionNumberString").invoke(null);
+            scalaVersion = concat(scalaVersion, version, "/");
+        } catch (ClassNotFoundException e) {
+            //Ignore
+        } catch (Exception e) {
+            if (log.isTraceEnabled()) {
```
dagnir, line 58: nit: don't think this is necessary

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/useragent/UserAgentLangValues.java
```diff
@@ -0,0 +1,92 @@
...
+        try {
+            Class<?> kotlinUnit = Class.forName("kotlin.Unit");
+            kotlinVersion = "kotlin";
+            kotlinJar = new JarInputStream(kotlinUnit.getProtectionDomain().getCodeSource().getLocation().openStream());
+            String version = kotlinJar.getManifest().getMainAttributes().getValue("Implementation-Version");
+            kotlinVersion = concat(kotlinVersion, version, "/");
+        } catch (ClassNotFoundException e) {
+            //Ignore
+        } catch (Exception e) {
+            if (log.isTraceEnabled()) {
```
dagnir, line 84: Not necessary

core/sdk-core/src/main/java/software/amazon/awssdk/core/client/builder/SdkDefaultClientBuilder.java
dagnir: I'd feel more comfortable just keeping these values here. They're public so it's entirely possible some customers are reading or modifying these values.

cenedhryn: It's frustrating to not be able to remove cruft, but I'm willing to put it back for the sake of less risk affecting edge cases. 

I do want to scrap the definition of `USER_AGENT_PREFIX, SdkUserAgent.create().userAgent()` and set it to the empty string like the suffix. It doesn't make sense to put a huge chunk of text in this param that we know we won't use in any case.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/useragent/UserAgentLangValues.java
cenedhryn: Kept previous code, but can remove.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/useragent/UserAgentLangValues.java
cenedhryn: Same

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/useragent/UserAgentConstant.java
```diff
@@ -0,0 +1,121 @@
...
+        if (!StringUtils.isEmpty(value)) {
+            appendFieldAndSpace(builder, name, value);
+        }
+    }
+
+    /**
+     * Replace any spaces, parentheses in the input with underscores.
+     */
+    public static String sanitizeInput(String input) {
+        return input == null ? UNKNOWN : input.replaceAll(UA_DENYLIST_REGEX, "_");
```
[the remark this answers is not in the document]
cenedhryn: The current usage doesn't require it, since it's mainly used in the singleton class, but I'm adding it still.

core/sdk-core/src/main/java/software/amazon/awssdk/core/client/builder/SdkDefaultClientBuilder.java
dagnir: Okay, I think leaving it as empty string should be fine.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/useragent/UserAgentConstant.java
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    /**
+     * Concatenates two values with the specified separator, if the second value is not null/empty, otherwise
+     * returns the first value.
+     */
+    public static String concat(String prefix, String suffix, String separator) {
+        return suffix != null && !suffix.isEmpty() ? prefix + separator + suffix : prefix;
```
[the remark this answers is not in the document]
cenedhryn: It's fine; it's on a few different call paths but mainly it's legacy code. I don't think there is any case where there would only be a whitespace value - don't think that can happen with system setting values.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/useragent/UserAgentLangValues.java
cenedhryn: To clarify, this code is unchanged from before. I want to leave it as is to not risk any issues with errors.

dagnir (approved):

PR #5602, opened 2024-09-16
Migration tool exclude custom sdks

davidh44:
## Motivation and Context

Skip transforming v1 custom SDK code and imports

## Modifications

Added list of v1 packages, will skip transform if type not in list
Removed packages to skip, was redundant since they are filtered out by the v1 model/class Patterns already


v2-migration/src/test/java/software/amazon/awssdk/v2migration/ChangeSdkTypeTest.java
```diff
@@ -216,30 +215,4 @@ void shouldChangeFieldsInInnerClass() {
             )
         );
     }
-
-    @Test
-    @EnabledOnJre({JRE.JAVA_8})
-    void hasUnsupportedFeature_shouldSkip() {
```
zoewangg, line 222: Can we still keep the test? It should still pass, right?

davidh44: Yup, adding back test, with Lambda invoke class

PR #5607, opened 2024-09-18
Reuse connections for 5xx errors

zoewangg:
## Motivation and Context
Reuse connections for 5xx errors for all supported HTTP clients. This change reverts https://github.com/aws/aws-sdk-java-v2/pull/1768 and https://github.com/aws/aws-sdk-java-v2/pull/2960


jtjeferreira:
Hi @zoewangg,

I am working on an AWS SDK Http client based on pekko [here](https://github.com/apache/pekko-connectors/pull/829/files/f3f04f8027aa362e10b89894ebc7bf83e9ea2bbd..56dc24550d56ff8b6992b150a329ad7479ca9d81), and the only failing test from `SdkAsyncHttpClientH1TestSuite` is `connectionReceiveServerErrorStatusShouldNotReuseConnection` which I am ignoring for now. 

Luckily I stumbled upon this PR, and I am just wondering why this functionality is being reverted?

zoewangg:
Hi @jtjeferreira, we are still evaluating the change. The main reason is that closing connections for 5xx errors could cause more TCP connection establishment in the unlikely event of outage, which could add request latency and add more load on the server side. In general, we try to keep custom logic in the SDK minimal and let the server side decide if a connection should be kept open or not.

millems (approved):

PR #5610, opened 2024-09-18
Add customization in S3 to keep expires as timestamp type

zoewangg:
## Motivation and Context
Add customization in S3 to keep expires as timestamp type

## Modifications
- Update `ShapeSubstitutionsProcessor` to handle `emitAsType`
- Add customization in S3 to keep expires as timestamp type

## Testing
- Added unit tests
- Changed `expires` to string type in s3 and verified with japicmp that no breaking changes are introduced.

## Types of changes

## Checklist

## License

### DIFF
--- a/codegen/src/main/java/software/amazon/awssdk/codegen/customization/processors/ShapeSubstitutionsProcessor.java
+++ b/codegen/src/main/java/software/amazon/awssdk/codegen/customization/processors/ShapeSubstitutionsProcessor.java
@@ -31,6 +31,8 @@ import software.amazon.awssdk.codegen.model.service.Member;
 import software.amazon.awssdk.codegen.model.service.Operation;
 import software.amazon.awssdk.codegen.model.service.ServiceModel;
 import software.amazon.awssdk.codegen.model.service.Shape;
+import software.amazon.awssdk.utils.CollectionUtils;
+import software.amazon.awssdk.utils.Validate;
 
 /**
  * This processor internally keeps track of all the structure members whose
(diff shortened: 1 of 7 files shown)


codegen/src/main/java/software/amazon/awssdk/codegen/customization/processors/ShapeSubstitutionsProcessor.java
```diff
@@ -209,8 +219,24 @@ private void preprocessSubstituteShapeReferencesInShape(
...
+            member.setShape(emitAsShape);
+            return substitute;
+        }
+
+        String emitAsType = substitute.getEmitAsType();
+
+        if (emitAsType != null) {
+            Shape newShapeForType = new Shape();
+            newShapeForType.setType(emitAsType);
+            String shapeName = "SdkCustomization_" + emitAsType;
```
L-Applin, line 237: Just for my understanding, what is this `SdkCustomization_` name prefix for?

L-Applin (approved):

codegen/src/main/java/software/amazon/awssdk/codegen/customization/processors/ShapeSubstitutionsProcessor.java
zoewangg: This is for the case where we need to add a new type to the service model. In this case, we are adding `"SdkCustomization_ expires": {"type": "timestamp"}` to S3 since it doesn't have a generic timestamp type.

PR #5637, opened 2024-10-01
Deprecate internal checksum APIs

zoewangg:
## Motivation and Context
We have duplicate checksum classes, which is really confusing when we add new checksum algorithms.. Example: old [Md5Checksum](https://github.com/aws/aws-sdk-java-v2/blob/master/core/sdk-core/src/main/java/software/amazon/awssdk/core/checksums/Md5Checksum.java), new [Md5Checksum](https://github.com/aws/aws-sdk-java-v2/blob/master/core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/signer/checksums/Md5Checksum.java), same for other checksum algorithms.

## Modifications
- Deprecated internal checksum APIs. Since they are not in internal sub packages unfortunately, removing them could break users who didn't notice `SdkInternalApi` annotations, even though we don't maintain backwards compatibility guarantees for internal APIs.
- Changed annotation fo checksum classes that are used in generated sdk clients from `SdkPublicApi` to `SdkProtectedApi`

## Types of changes

## Checklist

## License

### DIFF
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/checksums/Algorithm.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/checksums/Algorithm.java
@@ -16,6 +16,7 @@
 package software.amazon.awssdk.core.checksums;
 
 import java.util.Map;
+import software.amazon.awssdk.annotations.SdkProtectedApi;
 import software.amazon.awssdk.annotations.SdkPublicApi;
 import software.amazon.awssdk.utils.StringUtils;
 import software.amazon.awssdk.utils.internal.EnumUtils;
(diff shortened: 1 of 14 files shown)


core/sdk-core/src/main/java/software/amazon/awssdk/core/checksums/Algorithm.java
```diff
@@ -16,14 +16,15 @@
...
+import software.amazon.awssdk.annotations.SdkProtectedApi;
 import software.amazon.awssdk.annotations.SdkPublicApi;
 import software.amazon.awssdk.utils.StringUtils;
 import software.amazon.awssdk.utils.internal.EnumUtils;
 
 /**
  * Enum that indicates all the checksums supported by Flexible checksums in a Service Request/Response Header.
  */
-@SdkPublicApi
+@SdkProtectedApi
```
millems, line 27: Moving things from public to protected seems unsafe, since people may have already read it and taken a dependency on it. In the future, we might assume it's protected and treat it differently than we might if we knew it used to be public.

zoewangg: Fair point. I can add a note instead

zoewangg:
Closing the PR, the change was merged to feature branch for now

PR #5648, opened 2024-10-08
Add retry conditions to the strategy if none is configured

sugmanue:
## Motivation and Context

This change adds the default retry conditions in the client builder if none are configured to behave similarly to retry policies that are configured behind the scenes without the users having to do that themselves. 

This will prevent customers using directly the retry strategies builders from `DefaultRetryStrategy` to end up with a no-op strategy.

## Types of changes

## Checklist

## License

### DIFF
--- a/core/aws-core/src/main/java/software/amazon/awssdk/awscore/client/builder/AwsDefaultClientBuilder.java
+++ b/core/aws-core/src/main/java/software/amazon/awssdk/awscore/client/builder/AwsDefaultClientBuilder.java
@@ -51,6 +51,7 @@ import software.amazon.awssdk.core.client.config.SdkClientConfiguration;
 import software.amazon.awssdk.core.client.config.SdkClientOption;
 import software.amazon.awssdk.core.interceptor.ExecutionInterceptor;
 import software.amazon.awssdk.core.internal.SdkInternalTestAdvancedClientOption;
+import software.amazon.awssdk.core.internal.retry.SdkDefaultRetryStrategy;
 import software.amazon.awssdk.core.retry.RetryMode;
 import software.amazon.awssdk.core.retry.RetryPolicy;
 import software.amazon.awssdk.http.SdkHttpClient;
@@ -64,6 +65,7 @@ import software.amazon.awssdk.regions.ServiceMetadata;
 import software.amazon.awssdk.regions.ServiceMetadataAdvancedOption;
 import software.amazon.awssdk.regions.providers.DefaultAwsRegionProviderChain;
 import software.amazon.awssdk.retries.api.RetryStrategy;
+import software.amazon.awssdk.retries.internal.BaseRetryStrategy;
 import software.amazon.awssdk.utils.AttributeMap;
 import software.amazon.awssdk.utils.AttributeMap.LazyValueSource;
 import software.amazon.awssdk.utils.CollectionUtils;
@@ -422,6 +424,20 @@ public abstract class AwsDefaultClientBuilder<BuilderT extends AwsClientBuilder<
     private void configureRetryStrategy(SdkClientConfiguration.Builder config) {
         RetryStrategy strategy = config.option(SdkClientOption.RETRY_STRATEGY);
         if (strategy != null) {
+            // Fix the retry strategy if no retry predicates were configured. Users using
+            // DefaultRetryStrategy do not have any conditions configured as that package
+            // does not know anything about the SDK or AWS retry conditions. This mimics the
+            // retry policies behavior that are configured behind the scenes and avoids the
+            // risk of customers inadvertently using a no-op retry strategy.
+            if (strategy.maxAttempts() > 1
+                && (strategy instanceof BaseRetryStrategy)
+                && !((BaseRetryStrategy) strategy).hasRetryPredicates()
+            ) {
+                RetryStrategy.Builder<?, ?> builder = strategy.toBuilder();
+                SdkDefaultRetryStrategy.configureStrategy(builder);
+                AwsRetryStrategy.configureStrategy(builder);
+                config.option(SdkClientOption.RETRY_STRATEGY, builder.build());
+            }
             return;
         }
         config.lazyOption(SdkClientOption.RETRY_STRATEGY, this::resolveAwsRetryStrategy);
--- a/core/retries/src/main/java/software/amazon/awssdk/retries/internal/BaseRetryStrategy.java
+++ b/core/retries/src/main/java/software/amazon/awssdk/retries/internal/BaseRetryStrategy.java
@@ -187,6 +187,15 @@ public abstract class BaseRetryStrategy implements RetryStrategy {
         return 0;
     }
 
+    /**
+     * Returns true if there are retry predicates configured for this retry strategy.
+     *
+     * @return true if there are retry predicates configured for this retry strategy.
+     */
+    public boolean hasRetryPredicates() {
+        return !retryPredicates.isEmpty();
+    }
+
     private DefaultRetryToken refreshToken(RefreshRetryTokenRequest request, AcquireResponse acquireResponse) {
         DefaultRetryToken token = asDefaultRetryToken(request.token());
         return token.toBuilder()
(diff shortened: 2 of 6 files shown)


core/aws-core/src/main/java/software/amazon/awssdk/awscore/client/builder/AwsDefaultClientBuilder.java
```diff
@@ -422,6 +424,20 @@ private void configureRetryPolicy(SdkClientConfiguration.Builder config) {
     private void configureRetryStrategy(SdkClientConfiguration.Builder config) {
         RetryStrategy strategy = config.option(SdkClientOption.RETRY_STRATEGY);
         if (strategy != null) {
+            // Fix the retry strategy if no retry predicates were configured. Users using
+            // DefaultRetryStrategy do not have any conditions configured as that package
+            // does not know anything about the SDK or AWS retry conditions. This mimics the
+            // retry policies behavior that are configured behind the scenes and avoids the
+            // risk of customers inadvertently using a no-op retry strategy.
+            if (strategy.maxAttempts() > 1
```
zoewangg, line 432: Why do we need `maxAttempts() > 1` again?

core/retries/src/main/java/software/amazon/awssdk/retries/internal/BaseRetryStrategy.java
```diff
@@ -187,6 +187,15 @@ protected int exceptionCost(RefreshRetryTokenRequest request) {
         return 0;
     }
 
+    /**
+     * Returns true if there are retry predicates configured for this retry strategy.
+     *
+     * @return true if there are retry predicates configured for this retry strategy.
+     */
+    public boolean hasRetryPredicates() {
```
zoewangg, line 195: nit: `final`?

core/aws-core/src/main/java/software/amazon/awssdk/awscore/client/builder/AwsDefaultClientBuilder.java
sugmanue: Not really needed, but avoids changing the strategy for when the user meant to avoid retries.

core/retries/src/main/java/software/amazon/awssdk/retries/internal/BaseRetryStrategy.java
sugmanue, line 195:
```suggestion
    public final boolean hasRetryPredicates() {
```

core/aws-core/src/main/java/software/amazon/awssdk/awscore/client/builder/AwsDefaultClientBuilder.java
```diff
@@ -422,6 +424,20 @@ private void configureRetryPolicy(SdkClientConfiguration.Builder config) {
...
+                && (strategy instanceof BaseRetryStrategy)
+                && !((BaseRetryStrategy) strategy).hasRetryPredicates()
+            ) {
+                RetryStrategy.Builder<?, ?> builder = strategy.toBuilder();
+                SdkDefaultRetryStrategy.configureStrategy(builder);
+                AwsRetryStrategy.configureStrategy(builder);
+                config.option(SdkClientOption.RETRY_STRATEGY, builder.build());
+            }
```
millems, line 440: The condition in this "if" feels fragile. If the thing we're trying to fix is "the customer thinks they are creating the default strategy and adding to it", this wouldn't address it. i.e. them adding predicates means they don't get any of the built-in ones.

Should we have an equivalent of "can the client add its own defaults?" to `RetryStrategy`? We could default it to false in the interface and true in our strategies. We'd use that method in the if statement instead of these conditions. If a customer wants to exclude the client-provided conditions, they can set the flag to false.

test/protocol-tests/src/test/java/software/amazon/awssdk/protocol/tests/retry/RetryStrategyBackfillsEmtpyRetryConditions.java
```diff
@@ -0,0 +1,162 @@
...
+import software.amazon.awssdk.auth.credentials.AwsBasicCredentials;
+import software.amazon.awssdk.auth.credentials.StaticCredentialsProvider;
+import software.amazon.awssdk.regions.Region;
+import software.amazon.awssdk.retries.DefaultRetryStrategy;
+import software.amazon.awssdk.services.protocoljsonrpc.ProtocolJsonRpcClient;
+import software.amazon.awssdk.services.protocoljsonrpc.model.AllTypesRequest;
+import software.amazon.awssdk.services.protocoljsonrpc.model.AllTypesResponse;
+import software.amazon.awssdk.services.protocoljsonrpc.model.ProtocolJsonRpcException;
+
+public class RetryStrategyBackfillsEmtpyRetryConditions {
```
millems, line 40: nit: empty

core/aws-core/src/main/java/software/amazon/awssdk/awscore/client/builder/AwsDefaultClientBuilder.java
```diff
@@ -422,6 +424,20 @@ private void configureRetryPolicy(SdkClientConfiguration.Builder config) {
     private void configureRetryStrategy(SdkClientConfiguration.Builder config) {
```
millems, line 424: What about DynamoDB or other service-specific retry strategy configurations? How do we include those as well as defaults, if the customer hasn't overridden the defaults?

core/aws-core/src/main/java/software/amazon/awssdk/awscore/client/builder/AwsDefaultClientBuilder.java
sugmanue: Not a big issue but we might end up with duplicated conditions. Say, they are using the correct way of building an AWS retry strategy but they just want to increase the number of retries, they would do

```
.overrideConfiguration(o -> o.retryStrategy(AwsRetryStrategy.standardRetryStrategy()
 .toBuilder()
 .maxAttemtps(10)
 .build()))
```

So we will add the default conditions twice. 

Also, the flag cannot be toggled unless is part of the state of the strategy and set in the builder, or they would have to override the method. Is that better than, say, just add the defaults here unconditionally?

core/aws-core/src/main/java/software/amazon/awssdk/awscore/client/builder/AwsDefaultClientBuilder.java
sugmanue: For DynamoDB, or any other service with custom retry strategy we codegen the setting as in 

```
        builder.option(SdkClientOption.RETRY_STRATEGY, DynamoDbRetryPolicy.resolveRetryStrategy(config));
```

So they will have a non-empty set of retry conditions already set. And the specified method should take care of setting the correct defaults already. E.g., for DynamoDB the method already returns

```

        return AwsRetryStrategy.forRetryMode(retryMode)
            .toBuilder()
            .maxAttempts(MAX_ATTEMPTS)
            .backoffStrategy(exponentialDelay(BASE_DELAY, SdkDefaultRetrySetting.MAX_BACKOFF))
            .build();
```

test/protocol-tests/src/test/java/software/amazon/awssdk/protocol/tests/retry/RetryStrategyBackfillsEmtpyRetryConditions.java
sugmanue: done

core/aws-core/src/main/java/software/amazon/awssdk/awscore/client/builder/AwsDefaultClientBuilder.java
```diff
@@ -422,6 +424,20 @@ private void configureRetryPolicy(SdkClientConfiguration.Builder config) {
...
         RetryStrategy strategy = config.option(SdkClientOption.RETRY_STRATEGY);
         if (strategy != null) {
+            // Fix the retry strategy if no retry predicates were configured. Users using
```
zoewangg, line 427: Can we add a TODO here as we discussed?

zoewangg (approved):

PR #5678, opened 2024-10-24
Allow redirect for 400 - IllegalLocationConstraintException in s3 cross region client

L-Applin:
The cross region client was not picking up on some redirect cases, especially when dealing with opt-in regions. This change  is to ensure the cross-region client redirect s3 calls to the correct region when dealing with opt-in regions.

## Motivation and Context
If a request to a standard region is addressing a bucket in an opt-in region (or vice-versa), S3 returns 400: IllegalLocationConstraintException. The sdk was previously only checking for 301 and 307 status.

## Modifications
Add a check for 400 IllegalLocationConstraintException in the logic for redirect check.

## Testing
Added new cases for 400 to existing unit tests.

## Types of changes


.changes/next-release/bugfix-AmazonS3-527812c.json
```diff
@@ -0,0 +1,6 @@
+{
+    "type": "bugfix",
+    "category": "Amazon S3",
+    "contributor": "",
+    "description": "Update the cross-region client to correctly handle redirect cases for opt-in regions."
```
zoewangg, line 5: nit, customers may not know what `cross-region` client is, suggesting something like the following:
```
Update the S3 client to correctly handle redirect cases for opt-in regions when crossRegionAccessEnabled is used.
```

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crossregion/utils/CrossRegionUtils.java
```diff
@@ -59,7 +61,16 @@ public static boolean isS3RedirectException(Throwable exception) {
...
     private static boolean isRedirectError(S3Exception exceptionToBeChecked) {
-        if (REDIRECT_STATUS_CODES.stream().anyMatch(status -> status.equals(exceptionToBeChecked.statusCode()))) {
+        boolean matchRedirectStatusCode =
+            REDIRECT_STATUS_CODES.stream()
+                                 .anyMatch(status -> status.equals(exceptionToBeChecked.statusCode()));
+        if (matchRedirectStatusCode) {
+            return true;
+        }
+        boolean is400IllegalLocationConstraintException =
+            exceptionToBeChecked.statusCode() == ILLEGAL_LOCATION_CONSTRAINT_EXCEPTION_STATUS_CODE
```
zoewangg, line 71: Can we just use 400 here? 400 is easier to read than ILLEGAL_LOCATION_CONSTRAINT_EXCEPTION_STATUS_CODE IMO...

services/s3/src/test/java/software/amazon/awssdk/services/s3/internal/crossregion/S3CrossRegionAsyncClientRedirectTest.java
```diff
@@ -49,8 +49,10 @@ public void setup() {
     @Override
     protected void stubRedirectSuccessSuccess(Integer redirect) {
         when(mockDelegateAsyncClient.listObjects(any(ListObjectsRequest.class)))
-            .thenReturn(CompletableFutureUtils.failedFuture(new CompletionException(redirectException(redirect, CROSS_REGION.id(),
-                                                                                                      null, null))))
+            .thenReturn(CompletableFutureUtils.failedFuture(new CompletionException(redirectException(redirect,
```
zoewangg, line 52: Can we add a test case where it's 400 with a different error code to verify no redirect? Or do we have an existing test somewhere covering that case?

L-Applin: good point, adding it

zoewangg (approved):

PR #5685, opened 2024-10-30
Update value to path for operation context params

davidh44:
## Motivation and Context

Per the spec, should be `path` instead of `value`

## Modifications

Renamed to `path` in `OperationContextParam` and updated `EndpointResolverInterceptorSpec`

## Testing

updated codegen test files

### DIFF
--- a/codegen/src/main/java/software/amazon/awssdk/codegen/model/service/OperationContextParam.java
+++ b/codegen/src/main/java/software/amazon/awssdk/codegen/model/service/OperationContextParam.java
@@ -18,13 +18,13 @@ package software.amazon.awssdk.codegen.model.service;
 import com.fasterxml.jackson.core.TreeNode;
 
 public class OperationContextParam {
-    private TreeNode value;
+    private TreeNode path;
 
-    public TreeNode getValue() {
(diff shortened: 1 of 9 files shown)


L-Applin (approved):

codegen/src/main/java/software/amazon/awssdk/codegen/model/service/OperationContextParam.java
```diff
@@ -18,13 +18,13 @@
...
 
-    public TreeNode getValue() {
-        return value;
+    public TreeNode getPath() {
+        return path;
     }
 
-    public void setValue(TreeNode value) {
-        this.value = value;
+    public void setPath(TreeNode path) {
```
sugmanue, line 27: Path is always a string, which we can map directly without having to use a `TreeNode` instance here. As in 

```
package software.amazon.awssdk.codegen.model.service;

public class OperationContextParam {
    private String path;

    public String getPath() {
        return path;
    }

    public void setPath(String path) {
        this.path = path;
    }
}
```
This will also simplify the code that uses those values.

```

PR #5692, opened 2024-10-31
Load the CRC checksum from CRT using Class path loaders inst…

joviegas:
Load the CRC checksum from CRT using Class path loaders instead of direct reference of crt related classes

## Motivation and Context

This change enables the CRC64NVME checksum class to be loaded dynamically using class loaders instead of directly referencing CRT-specific classes (e.g., software.amazon.awssdk.crt.checksums.CRC64NVME). This approach avoids hard dependencies on CRT-related classes and allows for optional functionality, preventing potential runtime errors if CRT dependencies are removed externally.

By loading classes dynamically, we ensure compatibility with configurations where the aws-crt library is unavailable, maintaining the optional nature of the CRC64NVME feature. This also utilizes ClassLoadHelpers, which handles conditional class loading based on class path availability.

## Modifications

- Replaced direct CRC64NVME imports with a dynamic load using ClassLoadHelpers to check class path availability.
- Updated Crc64NvmeChecksum to use a dynamically loaded CRC64NVME instance.
- Removed redundant optional dependencies from the project configuration, as the new approach ensures dynamic loading only when aws-crt is present.
- Refactored Crc64NvmeChecksum constructor to load CRC64NVME via CrcChecksumProvider.createCrtCrc64Nvme, leveraging conditional class loading.
- 
## Testing

Added specific test cases for CRC64NVME functionality in the auth-crt module, where the CRT dependency is available.
Ran tests in environments with and without CRT dependencies to verify the class loading behavior and ensure fallback compatibility.
Validated that the code operates seamlessly without aws-crt dependency, ensuring graceful handling of missing classes.

## Types of changes

## License


core/checksums/pom.xml
```diff
@@ -51,13 +51,6 @@
             <version>${awsjavasdk.version}</version>
         </dependency>
 
-        <dependency>
-            <groupId>software.amazon.awssdk.crt</groupId>
-            <artifactId>aws-crt</artifactId>
-            <version>${awscrt.version}</version>
-            <optional>true</optional>
```
dagnir, line 58: Does this need to be removed? It's an optional dependency already so it wouldn't be loaded at runtime

core/checksums/src/main/java/software/amazon/awssdk/checksums/SdkChecksum.java
```diff
@@ -39,7 +39,7 @@ public interface SdkChecksum extends Checksum {
     static SdkChecksum forAlgorithm(ChecksumAlgorithm algorithm) {
         switch (algorithm.algorithmId()) {
             case "CRC32C":
-                return Crc32cProvider.create();
+                return CrcChecksumProvider.create();
```
dagnir, line 42: I think this is confusing. Nothing about `CrcChecksumProvider.create()` says that we're returning a CRC32C checksum implementation. Can we make the name more descriptive?

core/checksums/src/main/java/software/amazon/awssdk/checksums/internal/CrcChecksumProvider.java
```diff
@@ -18,23 +18,24 @@
 import java.util.zip.Checksum;
 import software.amazon.awssdk.annotations.SdkInternalApi;
 import software.amazon.awssdk.checksums.SdkChecksum;
-import software.amazon.awssdk.crt.checksums.CRC32C;
 
 /**
  * Utility class to provide different implementations of CRC32C checksum. This class supports the use of: 1. Java-based CRC32C
  * (Java 9+ when available) 2. CRT-based CRC32C (when available) 3. SDK-based CRC32C (as fallback)
  */
 @SdkInternalApi
```
dagnir, line 26: Javadoc needs to be updated

core/checksums/src/main/java/software/amazon/awssdk/checksums/internal/CrcCloneOnMarkChecksum.java
```diff
@@ -32,22 +32,31 @@
...
+            try {
+                Method cloneMethod = checksum.getClass().getMethod("clone");
+                return (Checksum) cloneMethod.invoke(checksum);
+            } catch (Exception e) {
+                throw new IllegalStateException("Failed to clone the checksum", e);
+            }
+        }
+        throw new UnsupportedOperationException("Checksum does not support cloning");
+    }
+}
```
dagnir, line 62: Does this need to be public?

core/checksums/src/main/java/software/amazon/awssdk/checksums/internal/CrcChecksumProvider.java
```diff
@@ -65,14 +66,38 @@ public static SdkChecksum create() {
     static SdkChecksum createCrtCrc32C() {
         return CONSTRUCTOR_CACHE.getConstructor(CRT_CRC32C_CLASS_PATH).map(constructor -> {
             try {
-                return new CrcCloneOnMarkChecksum((Checksum) constructor.newInstance(), checksumToClone ->
-                    (Checksum) ((CRC32C) checksumToClone).clone());
-            } catch (ClassCastException | ReflectiveOperationException e) {
-                throw new IllegalStateException("Failed to instantiate " + JAVA_CRC32C_CLASS_PATH, e);
+                Checksum checksumInstance = (Checksum) constructor.newInstance();
+                return new CrcCloneOnMarkChecksum(checksumInstance);
+            } catch (ReflectiveOperationException e) {
```
dagnir, line 71: Curious, why is `ClassCastException` removed?

core/checksums/pom.xml
joviegas: Yes ,Users needs to explicitly add the crt depencies if they want to load the CRT classes. 
We donot need it is as compile time as well as not as optional since we are loading classes at runtime,

I needed this dependency for testing so kept it as test dependency

core/checksums/src/main/java/software/amazon/awssdk/checksums/SdkChecksum.java
joviegas: made it to crc32cImplementation

core/checksums/src/main/java/software/amazon/awssdk/checksums/internal/CrcChecksumProvider.java
joviegas: done

core/checksums/src/main/java/software/amazon/awssdk/checksums/internal/CrcChecksumProvider.java
joviegas: Earlier we were doing (Checksum) , we no longer need this since CrcCloneOnMarkChecksum internally does check for Cloneable inetrface

core/checksums/src/main/java/software/amazon/awssdk/checksums/internal/CrcCloneOnMarkChecksum.java
joviegas: made it private and non static  , this was just an additional function to do a instance of check

dagnir (approved):

core/checksums/src/main/java/software/amazon/awssdk/checksums/internal/CrcCloneOnMarkChecksum.java
```diff
@@ -32,22 +32,31 @@
...
         return new byte[] { valueBytes[4], valueBytes[5], valueBytes[6], valueBytes[7] };
     }
-}
\ No newline at end of file
+
+    // Function to clone any Checksum that implements Cloneable
+    private Checksum cloneIfCloneable(Checksum checksum) {
+        if (checksum instanceof Cloneable) {
+            try {
+                Method cloneMethod = checksum.getClass().getMethod("clone");
```
scrocquesel, line 54: @joviegas  Is there a test that cover this reflection call in native compilation ? I guess all checksum classes that would be passed to this class must be registered for reflection. Would it be possible to keep the `Function<>`  for both compatibility with native compilation and performance ?

PR #5704, opened 2024-11-11
Fix ApacheHttpClient's handling of request bodies on DELETE, GET, HEAD & OPTIONS requests

Xtansia:
## Motivation and Context
Though providing request bodies on these HTTP methods is uncommon, it is [not disallowed by the spec](https://www.rfc-editor.org/rfc/rfc9110.html#section-9.3.1-6), and is used in certain Elasticsearch/OpenSearch APIs. Authenticating with Amazon OpenSearch Service can be done via SigV4 and we re-use the SDK's signing logic to provide that in https://github.com/opensearch-project/opensearch-java.
We've received multiple bug reports of users having trouble making certain requests using the `ApacheHttpClient` (https://github.com/opensearch-project/opensearch-java/issues/712, https://github.com/opensearch-project/opensearch-java/issues/521), where the client itself doesn't complain or error but silently drops the request body resulting in a mismatched signature on the server.
The Netty & CRT `Sdk(Async)HttpClient` implementations correctly send the request body for all methods, URL Connection does as well with the exception of `GET` which is hardcoded to swap to `POST` (and doesn't support `PATCH` at all).
As such I think it is not harmful to bring the Apache implementation in line with the other client implementations.

## Modifications
Generalize the Apache HTTP request factory implementation to attach the request body entity on all HTTP methods.
This matches the original implementations of the separate method classes:
- https://github.com/apache/httpcomponents-client/blob/4.5.x/httpclient/src/main/java/org/apache/http/client/methods/HttpDelete.java
- https://github.com/apache/httpcomponents-client/blob/4.5.x/httpclient/src/main/java/org/apache/http/client/methods/HttpGet.java
- https://github.com/apache/httpcomponents-client/blob/4.5.x/httpclient/src/main/java/org/apache/http/client/methods/HttpHead.java
- https://github.com/apache/httpcomponents-client/blob/4.5.x/httpclient/src/main/java/org/apache/http/client/methods/HttpOptions.java
- https://github.com/apache/httpcomponents-client/blob/4.5.x/httpclient/src/main/java/org/apache/http/client/methods/HttpPatch.java
- https://github.com/apache/httpcomponents-client/blob/4.5.x/httpclient/src/main/java/org/apache/http/client/methods/HttpPost.java
- https://github.com/apache/httpcomponents-client/blob/4.5.x/httpclient/src/main/java/org/apache/http/client/methods/HttpPut.java

## Testing
Added unit tests

## Types of changes

## Checklist

## License

### DIFF
--- a/http-clients/apache-client/src/main/java/software/amazon/awssdk/http/apache/internal/impl/ApacheHttpRequestFactory.java
+++ b/http-clients/apache-client/src/main/java/software/amazon/awssdk/http/apache/internal/impl/ApacheHttpRequestFactory.java
@@ -24,14 +24,7 @@ import java.util.Optional;
 import org.apache.http.HttpEntity;
 import org.apache.http.HttpHeaders;
 import org.apache.http.client.config.RequestConfig;
-import org.apache.http.client.methods.HttpDelete;
 import org.apache.http.client.methods.HttpEntityEnclosingRequestBase;
-import org.apache.http.client.methods.HttpGet;
-import org.apache.http.client.methods.HttpHead;
-import org.apache.http.client.methods.HttpOptions;
-import org.apache.http.client.methods.HttpPatch;
-import org.apache.http.client.methods.HttpPost;
-import org.apache.http.client.methods.HttpPut;
 import org.apache.http.client.methods.HttpRequestBase;
 import software.amazon.awssdk.annotations.SdkInternalApi;
 import software.amazon.awssdk.http.HttpExecuteRequest;
@@ -116,23 +109,18 @@ public class ApacheHttpRequestFactory {
 
 
     private HttpRequestBase createApacheRequest(HttpExecuteRequest request, URI uri) {
-        switch (request.httpRequest().method()) {
+        SdkHttpMethod method = request.httpRequest().method();
+        switch (method) {
             case HEAD:
-                return new HttpHead(uri);
             case GET:
-                return new HttpGet(uri);
             case DELETE:
-                return new HttpDelete(uri);
             case OPTIONS:
-                return new HttpOptions(uri);
             case PATCH:
-                return wrapEntity(request, new HttpPatch(uri));
             case POST:
-                return wrapEntity(request, new HttpPost(uri));
             case PUT:
-                return wrapEntity(request, new HttpPut(uri));
+                return wrapEntity(request, new HttpRequestImpl(method, uri));
             default:
-                throw new RuntimeException("Unknown HTTP method name: " + request.httpRequest().method());
+                throw new RuntimeException("Unknown HTTP method name: " + method);
         }
     }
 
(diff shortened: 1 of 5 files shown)


victorekpo:
Awesome work! 👏 I just ran into this issue with the scroll API because the DELETE requests have a body. Thanks for looking into this.

http-clients/apache-client/src/main/java/software/amazon/awssdk/http/apache/internal/impl/ApacheHttpRequestFactory.java
```diff
@@ -116,23 +109,18 @@ private void addRequestConfig(final HttpRequestBase base,
...
+        SdkHttpMethod method = request.httpRequest().method();
+        switch (method) {
             case HEAD:
-                return new HttpHead(uri);
             case GET:
-                return new HttpGet(uri);
             case DELETE:
-                return new HttpDelete(uri);
             case OPTIONS:
-                return new HttpOptions(uri);
```
davidh44, line 127: For `Options`, can we continue to use `HttpOptions` since it includes the `getAllowedMethods` method

```
return wrapEntity(request, new HttpOptions(uri));
```

davidh44:
Thanks for the PR, looks good overall!

Builds are failing due to protocol tests which previously expected absent content length header, can we update the tests accordingly?

```
[ERROR] Failures:
--
[ERROR]   RestJsonContentTypeProtocolTest.runProtocolTest:48 Header 'Content-Length' was expected to be absent
```

- https://github.com/aws/aws-sdk-java-v2/blob/master/test/protocol-tests/src/test/java/software/amazon/awssdk/protocol/tests/RestJsonContentTypeProtocolTest.java
- https://github.com/aws/aws-sdk-java-v2/blob/master/test/protocol-tests-core/src/main/resources/software/amazon/awssdk/protocol/suites/cases/rest-json-contenttype.json

http-clients/apache-client/src/main/java/software/amazon/awssdk/http/apache/internal/impl/ApacheHttpRequestFactory.java
Xtansia: `getAllowedMethods` is not used by Apache HttpClient itself or by aws-sdk, and the Apache request objects aren't ever exposed to end users of the SDK. So I don't think there's any significant benefit to retaining it, but happy to rework things. Would however mean no entity on `OPTIONS` or having to reimplement all entity handling just on a subtype of `HttpOptions`

dagnir (approved):

http-clients/apache-client/src/main/java/software/amazon/awssdk/http/apache/internal/impl/ApacheHttpRequestFactory.java
davidh44: Makes sense, we can keep as is

davidh44 (approved):

davidh44:
https://github.com/all-contributors please add @Xtansia for code

davidh44:
@all-contributors please add @Xtansia for code

PR #5708, opened 2024-11-11
Pin rewrite-maven-plugin version for migration test

davidh44:
## Motivation and Context

Regression in rewrite-maven-plugin 5.44.0

Filed bug report
https://github.com/openrewrite/rewrite-maven-plugin/issues/894

## Modifications

Pin to previous working version 5.43.0

## Testing

build passes locally

### DIFF
--- a/test/v2-migration-tests/src/test/java/software/amazon/awssdk/v2migrationtests/MavenProjectTest.java
+++ b/test/v2-migration-tests/src/test/java/software/amazon/awssdk/v2migrationtests/MavenProjectTest.java
@@ -29,10 +29,8 @@ import java.util.ArrayList;
 import java.util.List;
 import org.apache.commons.io.FileUtils;
 import org.junit.jupiter.api.BeforeAll;
-import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.condition.EnabledIf;
-import software.amazon.awssdk.testutils.SdkVersionUtils;
 import software.amazon.awssdk.utils.Logger;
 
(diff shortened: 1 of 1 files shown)


test/v2-migration-tests/src/test/java/software/amazon/awssdk/v2migrationtests/MavenProjectTest.java
```diff
@@ -69,16 +67,17 @@ private static void deleteTempDirectories() throws IOException {
...
         verifyTransformation();
         verifyCompilation();
     }
 
     private static void verifyTransformation() throws IOException {
         List<String> rewriteArgs = new ArrayList<>();
-        addAll(rewriteArgs, "mvn", "org.openrewrite.maven:rewrite-maven-plugin:run",
+        // pin version since updates have broken tests
+        String rewriteMavenPluginVersion = "5.43.0";
+        addAll(rewriteArgs, "mvn", "org.openrewrite.maven:rewrite-maven-plugin:" + rewriteMavenPluginVersion + ":run",
```
dagnir, line 80: Do we need the separate variable?

davidh44: just so easier to see and update version in future, to pin to more recent version once it works. wdyt?

dagnir: Seems like you could just as easily update the version in the if were all a single string. I don't feel strongly either way though.

dagnir (approved):

PR #5720, opened 2024-11-21
Codegen direct index access and map member model

davidh44:
## Motivation and Context

1) Add support for parsing direct index access in endpoint rules set, e.g., `[123]`

2) Add support for map MemberModel in client endpoint tests

## Testing

Added codegen tests
Builds pass locally for new feature

### DIFF
--- a/codegen/src/main/java/software/amazon/awssdk/codegen/poet/rules/EndpointRulesClientTestSpec.java
+++ b/codegen/src/main/java/software/amazon/awssdk/codegen/poet/rules/EndpointRulesClientTestSpec.java
@@ -20,6 +20,7 @@ import static software.amazon.awssdk.codegen.poet.rules.TestGeneratorUtils.getHo
 import com.fasterxml.jackson.core.TreeNode;
 import com.fasterxml.jackson.jr.stree.JrsArray;
 import com.fasterxml.jackson.jr.stree.JrsObject;
+import com.fasterxml.jackson.jr.stree.JrsString;
 import com.fasterxml.jackson.jr.stree.JrsValue;
 import com.squareup.javapoet.AnnotationSpec;
 import com.squareup.javapoet.ClassName;
@@ -68,6 +69,7 @@ import software.amazon.awssdk.core.rules.testing.SyncTestCase;
 import software.amazon.awssdk.core.rules.testing.util.EmptyPublisher;
 import software.amazon.awssdk.core.sync.RequestBody;
 import software.amazon.awssdk.regions.Region;
+import software.amazon.awssdk.utils.ImmutableMap;
 import software.amazon.awssdk.utils.Validate;
 
 public class EndpointRulesClientTestSpec implements ClassSpec {
@@ -781,8 +783,26 @@ public class EndpointRulesClientTestSpec implements ClassSpec {
         }
 
         if (memberModel.isMap()) {
-            // Not necessary at the moment
-            throw new RuntimeException("Don't know how to create map member.");
+            Iterator<Map.Entry<String, JrsValue>> fieldsIter = ((JrsObject) valueNode).fields();
+            MemberModel keyModel = memberModel.getMapModel().getKeyModel();
+            MemberModel valueModel = memberModel.getMapModel().getValueModel();
+
+            b.add("$T.of(", ImmutableMap.class);
+            boolean firstKey = true;
+            while (fieldsIter.hasNext()) {
+                Map.Entry<String, JrsValue> entry = fieldsIter.next();
+                JrsString keyNode = new JrsString(entry.getKey());
+
+                if (!firstKey) {
+                    b.add(", ");
+                }
+                firstKey = false;
+                b.add(createMemberValue(keyModel, keyNode));
+                b.add(", ");
+                b.add(createMemberValue(valueModel, entry.getValue()));
+            }
+            b.add(")");
+            return b.build();
         }
 
         return createModelClass(model.getShapes().get(memberModel.getC2jShape()), valueNode);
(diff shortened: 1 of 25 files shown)


codegen/src/main/java/software/amazon/awssdk/codegen/poet/rules2/AssignTypesVisitor.java
```diff
@@ -90,12 +88,6 @@ public RuleExpression visitFunctionCallExpression(FunctionCallExpression e) {
             return expr;
         }
 
-        List<RuleType> args = expr.arguments().stream().map(RuleExpression::type).collect(Collectors.toList());
-        if (!function.matches(args)) {
-            addError("Arguments for function `%s` doesn't match, expected: `%s`, got: `%s`",
-                     name,
-                     function.arguments().values(), args);
-        }
```
davidh44, line 98: We need to remove this error check here and above since they will fail when there is two assigns (LetExpression) in a row where second assign depends on the first assigned variable.

Because of the order of execution of the visitors, the first variable will not be stored in SymbolTable.locals before the second assign executes initially. The second assign will execute again successfully after the first variable is stored when super.visitXX is invoked

codegen/src/main/java/software/amazon/awssdk/codegen/poet/rules/EndpointRulesClientTestSpec.java
```diff
@@ -781,8 +783,26 @@ private CodeBlock createMemberValue(MemberModel memberModel, TreeNode valueNode)
         }
 
         if (memberModel.isMap()) {
-            // Not necessary at the moment
-            throw new RuntimeException("Don't know how to create map member.");
+            Iterator<Map.Entry<String, JrsValue>> fieldsIter = ((JrsObject) valueNode).fields();
+            MemberModel keyModel = memberModel.getMapModel().getKeyModel();
+            MemberModel valueModel = memberModel.getMapModel().getValueModel();
+
+            b.add("$T.of(", ImmutableMap.class);
```
millems, line 790: should we use `builder()`, since `of` only supports up to 4 KV pairs?

millems (approved):

codegen/src/main/java/software/amazon/awssdk/codegen/poet/rules/EndpointRulesClientTestSpec.java
davidh44: I tried to use that initially, but `ImmutableMap.builder().build()` returns `ImmutableMap<K, V>`, so its not compatible. We'd have to update `build()` to return `<K, V> ImmutableMap<K, V>` like `of()` does , which would entail an unchecked cast. WDYT?

```
        public <K, V> ImmutableMap<K, V> build() {
            HashMap<K, V> builtMap = (HashMap<K, V>) new HashMap<>(entries);
            return new ImmutableMap<>(builtMap);
        }
```

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/rules2/endpoint-provider-know-prop-override-class.java
```diff
@@ -179,6 +183,31 @@ private static RuleResult endpointRule11(QueryEndpointParams params, LocalState
...
+            && params.arnList() != null) {
+            String firstArn = null;
+            RuleArn parsedArn = null;
+            if ((firstArn = RulesFunctions.listAccess(params.arnList(), 0)) != null) {
+                locals = locals.toBuilder().firstArn(firstArn).build();
+            } else {
+                return RuleResult.carryOn();
+            }
+            if ((parsedArn = RulesFunctions.awsParseArn(locals.firstArn())) != null) {
+                locals = locals.toBuilder().parsedArn(parsedArn).build();
```
sugmanue, line 197: I was thinking more along the lines of 

```
            if ((firstArn = RulesFunctions.listAccess(params.arnList(), 0)) != null
                && (parsedArn = RulesFunctions.awsParseArn(firstArn)) != null) {
                locals = locals.toBuilder().parsedArn(parsedArn).firstArn(firstArn).build();
```

What you have works I think, just feeling uneasy about being able to reach the end of the second if with locals modified. That doesn't matter now, but before I would've never matter as the side effects would only happen in the if body scope.

sugmanue: NOTE, No action item or change is expected, just leaving this as a side note.

Nesting could also work leaving the previous semantics, e.g.,

```java
        if ((firstArn = RulesFunctions.listAccess(params.arnList(), 0)) != null) {
                locals = locals.toBuilder().firstArn(firstArn).build();
                if ((parsedArn = RulesFunctions.awsParseArn(locals.firstArn())) != null) {
                    locals = locals.toBuilder().parsedArn(parsedArn).build();
                    return RuleResult.endpoint(Endpoint
                                                   .builder()
                                                   .url(URI.create("https://" + params.endpointId() + ".query."
                                                                   + locals.partitionResult().dualStackDnsSuffix()))
                                                   .putAttribute(
                                                       AwsEndpointAttribute.AUTH_SCHEMES,
                                                       Arrays.asList(SigV4aAuthScheme.builder().signingName("query")
                                                                                     .signingRegionSet(Arrays.asList("*")).build())).build());
                }
        }
        return RuleResult.carryOn();
```

sugmanue (approved):

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/rules2/endpoint-provider-know-prop-override-class.java
davidh44: Yeah I also came up with that at first when handwriting it, just not sure how to modify the logic to do so. Need to plumb deeper through the call stack which is already pretty complicated

codegen/src/main/java/software/amazon/awssdk/codegen/poet/rules/EndpointRulesClientTestSpec.java
davidh44: Will merge as is, can update in follow up PR if needed, or can also add new `of()` overloads to support more pairs if necessary in future

PR #5724, opened 2024-11-22
Add default retry predicates to user provided ones

L-Applin:
Add default retry predicates to retry strategies on top of user provided predicates. 

## Motivation and Context
Currently, the SDK will only add default retry predicates when user do not specified any, effectively overriding defaults whenever a single other predicate is defined by user.

## Modifications
- Added an opt out configuration: The `boolean useClientDefaults()` method has been added to the RetryStrategy and builder interface to check whether this strategy should add the default predicates or not.
- Keep track of which defaults have been added to prevent adding them multiple time at client build time.
- Moved `DefaultAwsClientBuilderTest` to junit 5

## Testing
Added unit tests

## Types of changes


core/aws-core/src/main/java/software/amazon/awssdk/awscore/retry/AwsRetryStrategy.java
```diff
@@ -162,4 +167,16 @@ private static RetryStrategy legacyAdaptiveRetryStrategy() {
                                  .build();
     }
 
+    public static void applyDefault(RetryStrategy.Builder<?, ?> builder) {
```
millems, line 170: nit: `applyDefaults`? (here and SDK)

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/retry/SdkDefaultRetryStrategy.java
```diff
@@ -165,6 +167,10 @@ public static AdaptiveRetryStrategy.Builder adaptiveRetryStrategyBuilder() {
         if (maxAttempts != null) {
             builder.maxAttempts(maxAttempts);
         }
+        if (builder instanceof BaseRetryStrategy.Builder) {
+            BaseRetryStrategy.Builder b = (BaseRetryStrategy.Builder) builder;
+            b.markDefaultAdded(DEFAULTS_NAME);
+        }
```
millems, line 173: BaseRetryStrategy is an internal API, so we shouldn't use it from here or anywhere else outside of the module. Should we push up the BaseRetryStrategy methods we're using here, or migrate the base to be a protected API? I feel like the former, but I'm easy to convince.

L-Applin: Right now `BaseRetryStrategy.Builder` has no parent, so, yeah, we could create a `DefaultRetryStrategyBuilder` interface that is protected API for the methods used here, and cast to that type instead of `BaseRetryStrategy.Builder`, I think this would be clean and would work with the current hierarchy of the Builder inheritance which is tricky to work with already.

L-Applin: wait the `BaseRetryStrategy.Builder` is already marked `@SdkProtectedApi`

L-Applin: Decided to name it `DefaultAwareRetryStrategy` since we already have a [DefaultRetryStrategy](https://github.com/aws/aws-sdk-java-v2/blob/master/core/retries/src/main/java/software/amazon/awssdk/retries/DefaultRetryStrategy.java) class

core/retries/src/main/java/software/amazon/awssdk/retries/internal/DefaultAwareRetryStrategy.java
```diff
@@ -0,0 +1,46 @@
...
+        return false;
+    }
+
+    interface Builder {
+
+        /**
+         * Identify the Builder has having the specified set of predicate added to it.
+         * @param defaultPredicatesName the name of the set of predicate
+         */
+        default void markDefaultAdded(String defaultPredicatesName) {
```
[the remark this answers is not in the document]
L-Applin: right I see what you mean. The caller would just attempt to add the default every time,  which would just be a noop within the retry strategy if those defaults were already added. It's not a big change, I can see the benefit of doing it this way.

core/retries/src/main/java/software/amazon/awssdk/retries/internal/DefaultAwareRetryStrategy.java
```diff
@@ -22,25 +22,21 @@
...
         /**
-         * Identify the Builder has having the specified set of predicate added to it.
-         * @param defaultPredicatesName the name of the set of predicate
+         * Identify the Builder as having the specified defaults to it.
+         * @param defaultPredicatesName the name the defaults to mark as added
          */
-        default void markDefaultAdded(String defaultPredicatesName) {
-
-        }
+        void markDefaultAdded(String defaultPredicatesName);
```
millems, line 40: Is this method needed anymore?

core/retries/src/main/java/software/amazon/awssdk/retries/internal/RetryStrategyDefaults.java
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * The set of retry predicates that are by default added to a retry strategy.
+ */
+@SdkProtectedApi
+public interface RetryStrategyDefaults {
+
+    /**
+     * @return The unique name that identifies this set of predicates
+     */
+    String name();
```
millems, line 30: If we implement `equals()` and `hashCode()`, we don't even need `name()`.

core/retries/src/main/java/software/amazon/awssdk/retries/internal/DefaultAwareRetryStrategy.java
L-Applin: we do if we want to avoid having the core retry classes reference `BaseRetryStrategy` like you proposed eariler, which I think is the better approach. For example, `AwsRetryStrategy` needs to be able to mark defaults as being added when instantiated directly, to not duplicate them when the default client builder tries to add them later at client build time.

core/retries/src/main/java/software/amazon/awssdk/retries/internal/RetryStrategyDefaults.java
millems: That or just use identity equality and singletons, but I suppose that names are less opaque.

core/retries/src/main/java/software/amazon/awssdk/retries/internal/DefaultAwareRetryStrategy.java
millems: Got it, I imagined `AwsRetryStrategy` using `addDefaults` on instantiation to ensure they're marked as already included. I can see why that's weird now. For some reason I also imagined `addDefaults` being on the builder, not on the retry strategy itself.

millems (approved):

core/retries/src/main/java/software/amazon/awssdk/retries/internal/RetryStrategyDefaults.java
L-Applin: I think I slightly prefer singleton and identity, lets got with that. We wouldn't need to implement hashCode and  equals then.

L-Applin: changed my mind, lets just keep `name()`,  it's simpler

core/retries/src/main/java/software/amazon/awssdk/retries/internal/BaseRetryStrategy.java
```diff
@@ -66,6 +71,8 @@ public abstract class BaseRetryStrategy implements RetryStrategy {
         this.treatAsThrottling = Validate.paramNotNull(builder.treatAsThrottling, "treatAsThrottling");
         this.exceptionCost = Validate.paramNotNull(builder.exceptionCost, "exceptionCost");
         this.tokenBucketStore = Validate.paramNotNull(builder.tokenBucketStore, "tokenBucketStore");
+        this.defaultsAdded = Validate.paramNotNull(builder.defaultsAdded, "defaultsAdded");
```
sugmanue, line 74: You need to copy the set here instead of using the same as the builder, or, better yet, copy and pass it trough `Collections.unmodifiableSet`. Otherwise you end up sharing state that you shouldn't. For instance:

```java
RetryStategy.Builder builder = standardRetryStrategy().toBuilder();
addSomeFooDefaults(builder);
RetryStategy strategyA = builder.build();
addSomeBarDefaults(builder);
RetryStategy strategyB = builder.build();
```
After this, both, `strategyA` and `strategyB` think they have both, `Foo` and `Bar` defaults whereas only `strategyA` does.

PR #5735, opened 2024-11-30
feat(instance metadata service conection) : Added env variable,  system and profile file  property for AWS EC2 Metatdat Service timeout

joviegas:
…stem property for AWS EC2 Metatdat Service timeout

## Motivation and Context

This change introduces the AWS_METADATA_SERVICE_TIMEOUT setting to allow configuration of connection and read timeouts (in seconds) for the instance metadata service. It solves the problem of handling slow or unresponsive metadata service connections in certain environments by allowing users to customize the timeout. This setting is supported for InstanceProfileCredentialsProvider, and the IMDS SDK client, ensuring broader applicability across different metadata-related use cases.

Note : EC2MetadataUtils is a internal class and users are recommended to use imds client instead of EC2MetadataUtils

## Modifications

Added AWS_METADATA_SERVICE_TIMEOUT to configure connection and read timeouts for the ec2  metadata service timeouts.
Supported for  InstanceProfileCredentialsProvider  and the IMDS SDK client.

## Testing

- Junits added

## Types of changes

## License


core/imds/pom.xml
```diff
@@ -105,16 +105,6 @@
             <artifactId>test-utils</artifactId>
             <version>${awsjavasdk.version}</version>
             <scope>test</scope>
-            <exclusions>
```
joviegas, line 108: removed this because `import software.amazon.awssdk.testutils.EnvironmentVariableHelper; ` had dependency on Junit4  `import org.junit.rules.ExternalResource;
`

core/regions/src/main/java/software/amazon/awssdk/regions/internal/util/ConnectionUtils.java
```diff
@@ -47,4 +63,27 @@ public HttpURLConnection connectToEndpoint(URI endpoint, Map<String, String> hea
         return connection;
     }
 
+    private int resolveMetadataServiceTimeoutMillis() {
+
+        String timeoutValue = SystemSettingUtils.resolveSetting(SdkSystemSetting.AWS_METADATA_SERVICE_TIMEOUT)
```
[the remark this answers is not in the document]
joviegas: I agree with this observation and attempted to move the timeout setting out of the util class. However, ConnectionUtils is shared by InstanceProfileCredentialsProvider, EC2MetadataUtils, and indirectly by AutoDefaultsModeDiscovery and InstanceProfileRegionProvider.

Additionally, the util package already includes IMDS-specific classes, such as Ec2MetadataConfigProvider. Adding this logic to ConnectionUtils helps avoid modifying SdkProtected classes, where overloading methods to pass timeouts could result in non-overloaded methods not using the configured timeouts. Given the current design, with ConnectionUtils shared across these components, it seems like the most practical place to add this logic during instantiation.

Do you think we should proceed with adding overloaded methods to SdkProtected APIs, or is there a better alternative we could consider?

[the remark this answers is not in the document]
L-Applin: I see what you mean, my concern is that if `ConnectionUtils` is used for accessing non-imds endpoints, those http calls will use the timeouts configured for imds, it might be confusing for users. Should that be a concern? Do you think there is another better way to avoid that?

L-Applin:
it seems `HttpCredentialsUtilsTest` is failing

joviegas:
> it seems `HttpCredentialsUtilsTest` is failing with NPE

Thanks.
Done , updated code

core/regions/src/main/java/software/amazon/awssdk/regions/internal/util/ConnectionUtils.java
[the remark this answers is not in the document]
zoewangg: It seems `ConnectionUtils` is also used by `ContainerCredentialsProvider`, so we may want to avoid applying the same env var to `ContainerCredentialsProvider`

core/sdk-core/src/main/java/software/amazon/awssdk/core/SdkSystemSetting.java
```diff
@@ -90,6 +90,16 @@ public enum SdkSystemSetting implements SystemSetting {
...
 
+    /**
+     * The number of seconds (either as an integer or double) before a connection to the instance
+     * metadata service should time out. This value is applied to both the socket connect and read timeouts.
+     *
+     * The timeout can be configured using the system property "aws.ec2MetadataServiceTimeout". If not set,
+     * a default timeout is used. This setting is crucial for ensuring timely responses from the instance
+     * metadata service in environments with varying network conditions.
+     */
+    AWS_METADATA_SERVICE_TIMEOUT("aws.ec2MetadataServiceTimeout", "1"),
```
zoewangg, line 101: Should we also add profile config? https://docs.aws.amazon.com/sdkref/latest/guide/feature-ec2-instance-metadata.html

joviegas: done

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/StaticResourcesEndpointProvider.java
```diff
@@ -27,16 +30,25 @@
 public final class StaticResourcesEndpointProvider implements ResourcesEndpointProvider {
     private final URI endpoint;
     private final Map<String, String> headers;
+    private final Optional<Duration> connectionTimeout;
```
zoewangg, line 33: Optional should not be used as fields. Too bad this test is ignored for now...

https://github.com/aws/aws-sdk-java-v2/blob/master/test/architecture-tests/src/test/java/software/amazon/awssdk/archtests/CodingConventionTest.java#L87-L91

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/InstanceProfileCredentialsProvider.java
```diff
@@ -110,6 +112,7 @@ private InstanceProfileCredentialsProvider(BuilderImpl builder) {
                                      .profileName(profileName)
                                      .build();
         this.ec2MetadataDisableV1Resolver = Ec2MetadataDisableV1Resolver.create(profileFile, profileName);
+        this.ec2MetadataServiceTimeoutResolver = Ec2MetadataServiceTimeoutResolver.create(profileFile, profileName);
```
zoewangg, line 115: Instead of introducing yet another resolver, should we make it more flexible to support new settings in the future? Something like, `Ec2MetadataResolver`? (I know this ship has sailed for `Ec2metadataDisableV1Resolver` :( ).

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/Ec2MetadataServiceTimeoutResolver.java
```diff
@@ -0,0 +1,82 @@
...
+import java.util.Optional;
+import java.util.function.Supplier;
+import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.core.SdkSystemSetting;
+import software.amazon.awssdk.profiles.ProfileFile;
+import software.amazon.awssdk.profiles.ProfileProperty;
+import software.amazon.awssdk.utils.Lazy;
+import software.amazon.awssdk.utils.OptionalUtils;
+
+@SdkInternalApi
```
zoewangg, line 28: I'd expect InternalApiBoundaryTest to fail for this one... Let's see

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/StaticResourcesEndpointProvider.java
joviegas: My bad ,  missed it , thanks for the reference 👍

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/InstanceProfileCredentialsProvider.java
joviegas: We already have `Ec2MetadataConfigProvider` I have model all the config related to this class now

core/regions/src/main/java/software/amazon/awssdk/regions/internal/util/ConnectionUtils.java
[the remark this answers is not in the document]
joviegas: Done

core/imds/src/main/java/software/amazon/awssdk/imds/internal/BaseEc2MetadataClient.java
```diff
@@ -29,21 +30,21 @@
...
     protected static final Duration DEFAULT_TOKEN_TTL = Duration.of(21_600, ChronoUnit.SECONDS);
-    protected static final AttributeMap IMDS_HTTP_DEFAULTS =
-        AttributeMap.builder()
-                    .put(SdkHttpConfigurationOption.CONNECTION_TIMEOUT, Duration.ofSeconds(1))
-                    .put(SdkHttpConfigurationOption.READ_TIMEOUT, Duration.ofSeconds(1))
-                    .build();
 
     private static final Logger log = Logger.loggerFor(BaseEc2MetadataClient.class);
 
+    protected final Lazy<Duration> metadataServiceDuration = new Lazy<>(this::resolveMetadataServiceTimeoutDuration);
```
joviegas, line 46: @zoewangg 
Will update Ec2MetadataEndpointProvider to have this config since Ec2MetadataEndpointProvider has the respolveProfilFile method and this serviceTimeOut is related to Endpoint
In future is more settings then we can name Ec2MetadataEndpointProvider as Ec2MetadataConfigProvider.
Working on this changes , let me know any thoughts

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/Ec2MetadataServiceTimeoutResolver.java
joviegas: IMDS doesnot depend on auth and region module thus we need to add new classes

core/imds/src/main/java/software/amazon/awssdk/imds/Ec2MetadataClientBuilder.java
```diff
@@ -63,7 +63,7 @@ public interface Ec2MetadataClientBuilder<B, T> extends SdkBuilder<Ec2MetadataCl
      * Define the endpoint of IMDS.
      * <p>
      * If not specified, the IMDS client will attempt to automatically resolve the endpoint value
-     * based on the logic of {@link Ec2MetadataEndpointProvider}.
+     * based on the logic of {@link Ec2MetadataConfigProvider}.
```
zoewangg, line 66: Can we remove the mention of this internal API, everything after `resolve the endpoint value`?

core/imds/src/main/java/software/amazon/awssdk/imds/Ec2MetadataClientBuilder.java
```diff
@@ -85,7 +85,7 @@ public interface Ec2MetadataClientBuilder<B, T> extends SdkBuilder<Ec2MetadataCl
      * but not both. If both are specified in the Builder, a {@link IllegalArgumentException} will be thrown.
      * <p>
      * If not specified, the IMDS client will attempt to automatically resolve the endpoint mode value
-     * based on the logic of {@link Ec2MetadataEndpointProvider}.
+     * based on the logic of {@link Ec2MetadataConfigProvider}.
```
zoewangg, line 88: Same here

core/imds/src/main/java/software/amazon/awssdk/imds/Ec2MetadataClientBuilder.java
joviegas: The reason for removal is we do not want to expose internal classes anywhere right ?

PR #5743, opened 2024-12-06
Fix ApacheHttpClient's handling of request bodies on DELETE, GET, HEAD & OPTIONS requests

Xtansia:
This is a rework of https://github.com/aws/aws-sdk-java-v2/pull/5704 that was reverted.

Have modified logic to only use the entity enclosing request type on `DELETE`, `GET`, `HEAD` & `OPTIONS` when a content stream provider is present, ensuring the previous behavior of no Content-Length header being sent is retained.

Summary copied from previous PR:

## Motivation and Context
Though providing request bodies on these HTTP methods is uncommon, it is [not disallowed by the spec](https://www.rfc-editor.org/rfc/rfc9110.html#section-9.3.1-6), and is used in certain Elasticsearch/OpenSearch APIs. Authenticating with Amazon OpenSearch Service can be done via SigV4 and we re-use the SDK's signing logic to provide that in https://github.com/opensearch-project/opensearch-java.
We've received multiple bug reports of users having trouble making certain requests using the `ApacheHttpClient` (https://github.com/opensearch-project/opensearch-java/issues/712, https://github.com/opensearch-project/opensearch-java/issues/521), where the client itself doesn't complain or error but silently drops the request body resulting in a mismatched signature on the server.
The Netty & CRT `Sdk(Async)HttpClient` implementations correctly send the request body for all methods, URL Connection does as well with the exception of `GET` which is hardcoded to swap to `POST` (and doesn't support `PATCH` at all).
As such I think it is not harmful to bring the Apache implementation in line with the other client implementations.

## Modifications
Generalize the Apache HTTP request factory implementation to attach the request body entity on all HTTP methods.
This matches the original implementations of the separate method classes:
- https://github.com/apache/httpcomponents-client/blob/4.5.x/httpclient/src/main/java/org/apache/http/client/methods/HttpDelete.java
- https://github.com/apache/httpcomponents-client/blob/4.5.x/httpclient/src/main/java/org/apache/http/client/methods/HttpGet.java
- https://github.com/apache/httpcomponents-client/blob/4.5.x/httpclient/src/main/java/org/apache/http/client/methods/HttpHead.java
- https://github.com/apache/httpcomponents-client/blob/4.5.x/httpclient/src/main/java/org/apache/http/client/methods/HttpOptions.java
- https://github.com/apache/httpcomponents-client/blob/4.5.x/httpclient/src/main/java/org/apache/http/client/methods/HttpPatch.java
- https://github.com/apache/httpcomponents-client/blob/4.5.x/httpclient/src/main/java/org/apache/http/client/methods/HttpPost.java
- https://github.com/apache/httpcomponents-client/blob/4.5.x/httpclient/src/main/java/org/apache/http/client/methods/HttpPut.java

## Testing
Added unit tests

## Types of changes

## Checklist

## License


uorhsohk:
We also have the same problem with clearing the scroll context. Are there any updates? can this be merged?

PR #5752, opened 2024-12-13
EmfMetricPublisher class added

Fred1155:
## Motivation and Context

The existing CloudWatchMetricPublisher  has limitations when used with AWS Lambda functions. When current implementation be applied to lambda functions, it can results in metrics being lost when the lambda function terminates before the scheduled batch upload.   EMF allows metrics to be published through CloudWatch Logs using a structured JSON format, which CloudWatch automatically processes to extract and publish metrics.  The proposed EMFMetricPublisher will convert SDK metric collections into EMF-formatted log entries, leveraging Lambda's as well as other execution environment that has built-in integration with cloudwatch logs such as ECS. 
## Modifications

 Add a new metric publisher that transforms an metricCollection to a list of emf format Strings and publish it to CouldWatch using cloudwatch agents
## Testing

1. Unit test covering different cases of metricCollection in convertMetricCollectionToEMF method
2. End to End test (In process)
## Types of changes

## Checklist

## License


metric-publishers/emf-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/emf/EmfMetricPublisher.java
```diff
@@ -0,0 +1,551 @@
...
+ *   <li>CloudWatch Metrics directives that specify how to process the metrics</li>
+ * </ul>
+ *
+ *
+ */
+
+@ThreadSafe
+@Immutable
+@SdkPublicApi
+public final class EmfMetricPublisher implements MetricPublisher {
```
zoewangg, line 89: Let's break up this class a bit more.

metric-publishers/emf-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/emf/EmfMetricPublisher.java
```diff
@@ -0,0 +1,551 @@
...
+import software.amazon.awssdk.metrics.MetricLevel;
+import software.amazon.awssdk.metrics.MetricPublisher;
+import software.amazon.awssdk.metrics.MetricRecord;
+import software.amazon.awssdk.metrics.SdkMetric;
+import software.amazon.awssdk.thirdparty.jackson.core.JsonFactory;
+import software.amazon.awssdk.thirdparty.jackson.core.JsonGenerator;
+import software.amazon.awssdk.utils.Logger;
+
+/**
+ * A metric publisher implementation that converts metrics into CloudWatch Embedded Metric Format (EMF).
```
zoewangg, line 49: Can we try markdown for javadoc? 

https://docs.oracle.com/en/java/javase/23/javadoc/using-markdown-documentation-comments.html#GUID-0E4D6135-AAE9-4641-8427-CEE7F65E4F6E

metric-publishers/emf-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/emf/EmfMetricPublisher.java
```diff
@@ -0,0 +1,551 @@
...
+ * log entries.</p>
+ *
+ * <p>The EMF publisher converts metric collections into JSON-formatted log entries that conform to the
+ * CloudWatch EMF specification. These logs are written to the "/aws/emf/metrics" log group by default.
+ * CloudWatch automatically processes these logs to generate corresponding metrics that can be used for
+ * monitoring and alerting.</p>
+ *
+ * <p>Example usage:</p>
+ * <pre>
+ * EmfMetricPublisher publisher = PublisherBuilder.dimensions(CoreMetric.SERVICE_ID)
```
zoewangg, line 65: We should use `@snippet` for code sample for javadoc

https://github.com/aws/aws-sdk-java-v2/blob/master/docs/design/APIReference.md#style-guideline

metric-publishers/emf-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/emf/EmfMetricPublisher.java
```diff
@@ -0,0 +1,551 @@
...
+        }
+
+
+        /**
+         * Configure the unitTest field to the publisher. It tells the publisher if this publisher is used in unit test.
+         * If so, it will use a fixed timestamp when generating emf string instead of the current time.
+         *
+         * <p>If this is not specified, {@code false} will be used.
+         */
+        public Builder unitTest(boolean unitTest) {
```
zoewangg, line 535: We should use dependency injection for testing if we need to mock a dependency. eg: https://github.com/aws/aws-sdk-java-v2/blob/master/services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/GenericS3TransferManager.java#L121

metric-publishers/emf-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/emf/EmfMetricPublisher.java
```diff
@@ -0,0 +1,551 @@
...
+            return this;
+        }
+
+        /**
+         * Build a {@link EmfMetricPublisher} using the configuration currently configured on this publisher.
+         */
+        public EmfMetricPublisher build() {
+            // if (this.logGroupName == null || this.logGroupName.trim().isEmpty()) {
+            //     throw new IllegalStateException("logGroupName is required");
+            // }
```
zoewangg, line 546: Commented-out lines

metric-publishers/emf-metric-publisher/pom.xml
```diff
@@ -0,0 +1,53 @@
...
+            <version>${awsjavasdk.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>annotations</artifactId>
+            <version>${awsjavasdk.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>third-party-jackson-core</artifactId>
```
zoewangg, line 48: Can we use `json-utils` instead of using jackson directly?

metric-publishers/emf-metric-publisher/src/test/java/software/amazon/awssdk/metrics/publishers/emf/EmfMetricPublisherTest.java
```diff
@@ -0,0 +1,196 @@
...
+    void testConvertMetricCollectionToEMF_EmptyCollection(){
+        EmfMetricPublisher publisher = publisherBuilder.logGroupName("my-log-group-name")
+                                                       .build();
+
+        List<String> emfLogs = publisher.convertMetricCollectionToEMF(MetricCollector.create("test").collect());
+        for (String emfLog : emfLogs) {
+            assertThat(emfLog).isEqualTo("{\"_aws\":{\"Timestamp\":12345678,\"LogGroupName\":\"my-log-group-name\","
+                                         + "\"CloudWatchMetrics\":[{\"Namespace\":\"AwsSdk/JavaSdk2\",\"Dimensions\":[[]],"
+                                         + "\"Metrics\":[]}]}}");
+            System.out.println(emfLog);
```
zoewangg, line 49: Do we need log it? Seems unnecessary.

metric-publishers/emf-metric-publisher/src/test/java/software/amazon/awssdk/metrics/publishers/emf/EmfMetricPublisherTest.java
```diff
@@ -0,0 +1,196 @@
...
+
+    private EmfMetricPublisher.Builder publisherBuilder;
+
+    @BeforeEach
+    void setUp() {
+        publisherBuilder = EmfMetricPublisher.builder().unitTest(true);
+    }
+
+    @Test
+    void testConvertMetricCollectionToEMF_EmptyCollection(){
```
zoewangg, line 40: nit:`test` prefix is redundant, see [Test naming convention](https://github.com/aws/aws-sdk-java-v2/blob/master/docs/design/NamingConventions.md#tests-naming)

metric-publishers/emf-metric-publisher/src/test/java/software/amazon/awssdk/metrics/publishers/emf/EmfMetricPublisherTest.java
```diff
@@ -0,0 +1,196 @@
...
+
+    @Test
+    void testConvertMetricCollectionToEMF_SingleMetric(){
+        EmfMetricPublisher publisher = publisherBuilder.build();
+
+        MetricCollector metricCollector = MetricCollector.create("test");
+        metricCollector.reportMetric(HttpMetric.AVAILABLE_CONCURRENCY, 5);
+        List<String> emfLogs = publisher.convertMetricCollectionToEMF(metricCollector.collect());
+
+        for (String emfLog : emfLogs) {
```
zoewangg, line 61: Are there multiple logs?

metric-publishers/emf-metric-publisher/src/test/java/software/amazon/awssdk/metrics/publishers/emf/EmfMetricPublisherTest.java
```diff
@@ -0,0 +1,196 @@
...
+
+    @Test
+    void testConvertMetricCollectionToEMF_nonExistDimensions(){
+        EmfMetricPublisher publisher = publisherBuilder.dimensions(CoreMetric.SERVICE_ID)
+                                                       .build();
+
+        MetricCollector metricCollector = MetricCollector.create("test");
+        metricCollector.reportMetric(HttpMetric.AVAILABLE_CONCURRENCY, 5);
+        Object emf = publisher.convertMetricCollectionToEMF(metricCollector.collect());
+        System.out.println(emf);
```
zoewangg, line 117: Missing assertion. What are we testing?

metric-publishers/emf-metric-publisher/src/test/java/software/amazon/awssdk/metrics/publishers/emf/EmfMetricPublisherTest.java
```diff
@@ -0,0 +1,196 @@
...
+
+    @Test
+    void testConvertMetricCollectionToEMF_extraDimensions(){
+        EmfMetricPublisher publisher = publisherBuilder.build();
+
+        MetricCollector metricCollector = MetricCollector.create("test");
+        metricCollector.reportMetric(CoreMetric.SERVICE_ID, "ServiceId1234");
+        metricCollector.reportMetric(HttpMetric.AVAILABLE_CONCURRENCY, 5);
+        Object emf = publisher.convertMetricCollectionToEMF(metricCollector.collect());
+        System.out.println(emf);
```
zoewangg, line 128: Same here

metric-publishers/emf-metric-publisher/src/test/java/software/amazon/awssdk/metrics/publishers/emf/EmfMetricPublisherTest.java
zoewangg, line 40: Consider using parameterized tests if we are just testing with different input

zoewangg (commented):
Left some high level comments, still reviewing the implementation

metric-publishers/emf-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/emf/EmfMetricPublisher.java
Fred1155: Do you think I should separate the whole converting method into another class?

metric-publishers/emf-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/emf/MetricEmfConverter.java
```diff
@@ -0,0 +1,368 @@
...
+     *   "AvailableConcurrency": 5
+     * }
+     * ```
+     *
+     * @param metricCollection Collection of SDK metrics to be converted
+     * @return List of EMF-formatted metrics ready for CloudWatch ingestion
+     */
+    public List<String> convertMetricCollectionToEmf(MetricCollection metricCollection) {
+        // Map to store aggregated metrics
+        Map<String, List<Object>> aggregatedMetrics = new HashMap<>();
```
zoewangg, line 150: We should avoid using Object if possible because it's not type safe. Can we follow the same approach in cw metric publisher?

metric-publishers/emf-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/emf/MetricEmfConverter.java
```diff
@@ -0,0 +1,368 @@
...
+            logger.warn(() -> "failed to add metricCollection to the queue");
+        }
+
+        while (!queue.isEmpty()) {
+            MetricCollection current = queue.poll();
+
+            // Process all metrics in current collection
+            current.stream().forEach(r -> {
+                String metricName = r.metric().name();
+                Object metricValue = r.value();
```
zoewangg, line 164: We should use `metricRecord.metric().valueClass()` to determine the type

metric-publishers/emf-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/emf/MetricEmfConverter.java
```diff
@@ -0,0 +1,368 @@
...
+        Queue<MetricCollection> queue = new LinkedList<>();
+        if (!queue.offer(metricCollection)) {
+            logger.warn(() -> "failed to add metricCollection to the queue");
+        }
+
+        while (!queue.isEmpty()) {
+            MetricCollection current = queue.poll();
+
+            // Process all metrics in current collection
+            current.stream().forEach(r -> {
```
zoewangg, line 162: Let's use a more descriptive variable name instead of `r`

metric-publishers/emf-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/emf/MetricEmfConverter.java
```diff
@@ -0,0 +1,368 @@
...
+        if (metricValue == null) {
+            return 0.0;
+        }
+
+        // Changes boolean value from true/false to 1/0
+        if (metricValue instanceof Boolean) {
+            return metricValue.equals(true) ? 1.0 : 0.0;
+        }
+
+        // Changes duration value to a number of ms
```
zoewangg, line 92: Same here, the code should be self explanatory ideally, and we should only add comments in places that truly needs context.

Could you check other comments in this PR as well?

metric-publishers/emf-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/emf/MetricEmfConverter.java
```diff
@@ -0,0 +1,368 @@
...
+     * The method handles various input types and normalizes them according to EMF requirements:
+     * - `null` : 0.0
+     * - `Boolean` : 1.0 (true) or 0.0 (false)
+     * - `Duration` : milliseconds value
+     * - `Double`: normalized to 0.0 if below threshold
+     *
+     * @param metricValue The metric value to process
+     * @return The processed value in EMF-compatible format
+     */
+    private Object processValue(Object metricValue) {
```
zoewangg, line 82: We should avoid using raw Object because it's not type safe.

metric-publishers/emf-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/emf/MetricEmfConverter.java
```diff
@@ -0,0 +1,368 @@
...
+ * - Duration values are converted to milliseconds
+ *
+ * @since 2.0
+ * @see EmfMetricConfiguration
+ */
+@SdkInternalApi
+public class MetricEmfConverter {
+    private static final Logger logger = Logger.loggerFor("software.amazon.awssdk.metrics.publishers.emf");
+    private static final double ZERO_THRESHOLD = 0.0001;
+    private final List<String> realDimensionStrings = new ArrayList<>();
```
zoewangg, line 61: What's `realDimensionStrings`?

metric-publishers/emf-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/emf/MetricEmfConverter.java
```diff
@@ -0,0 +1,368 @@
...
+            writeAwsObject(jsonWriter, metricNames);
+            // Write metric values
+            writeMetricValues(jsonWriter, metrics);
+
+            jsonWriter.writeEndObject(); // End root object
+
+            return new String(jsonWriter.getBytes(), StandardCharsets.UTF_8);
+
+        } catch (SdkClientException e) {
+            logger.error(() -> "Failed to create EMF format string");
```
zoewangg, line 239: https://github.com/aws/aws-sdk-java-v2/blob/master/test/architecture-tests/src/test/java/software/amazon/awssdk/archtests/CodingConventionWithSuppressionTest.java#L80 this test should fail :)  

logger.error is not necessary here because we are throwing an exception anyway

metric-publishers/emf-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/emf/MetricEmfConverter.java
```diff
@@ -0,0 +1,368 @@
...
+    }
+
+    private void writeAwsObject(JsonWriter jsonWriter, Set<String> metricNames) {
+        // Start with _aws section
+        jsonWriter.writeFieldName("_aws");
+        jsonWriter.writeStartObject();
+
+        // Write Timestamp
+        jsonWriter.writeFieldName("Timestamp");
+        //jsonWriter.writeValue(Instant.now().toEpochMilli());
```
zoewangg, line 251: Commented out lines

metric-publishers/emf-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/emf/MetricEmfConverter.java
```diff
@@ -0,0 +1,368 @@
...
+        //Unit Test
+        jsonWriter.writeValue(12345678);
```
zoewangg, line 253: Unintentional change?

metric-publishers/emf-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/emf/MetricEmfConverter.java
```diff
@@ -0,0 +1,368 @@
...
+
+
+        jsonWriter.writeFieldName("LogGroupName");
+        jsonWriter.writeValue(config.getLogGroupName());
```
zoewangg, line 257: Suggesting writing a help method, eg: https://github.com/aws/aws-sdk-java-v2/blob/master/services-custom/iam-policy-builder/src/main/java/software/amazon/awssdk/policybuilder/iam/internal/DefaultIamPolicyWriter.java#L138

metric-publishers/emf-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/emf/MetricEmfConverter.java
```diff
@@ -0,0 +1,368 @@
...
+
+            return new String(jsonWriter.getBytes(), StandardCharsets.UTF_8);
+
+        } catch (SdkClientException e) {
+            logger.error(() -> "Failed to create EMF format string");
+            throw e;
+        }
+    }
+
+    private void writeAwsObject(JsonWriter jsonWriter, Set<String> metricNames) {
```
zoewangg, line 244: Suggesting breaking up this method to something like this:
```java
private void writeAwsObject(JsonWriter jsonWriter, Set<String> metricNames) {
       jsonWriter.writeFieldName("_aws");
       jsonWriter.writeStartObject();
       writeCloudWatchMetricsArray(jsonWriter, metricNames);
       jsonWriter.writeEndObject();
}

private void writeCloudWatchMetricsArray(JsonWriter jsonWriter, Set<String> metricNames) {
       jsonWriter.writeFieldName("CloudWatchMetrics");
       jsonWriter.writeStartArray();
       for (...) {
          writeCloudWatchMetricsObjects(jsonWriter, metricsName);
       }
       jsonWriter.writeEndArray();
}

...
```

metric-publishers/emf-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/emf/MetricEmfConverter.java
```diff
@@ -0,0 +1,368 @@
...
+        jsonWriter.writeValue(12345678);
+
+
+        jsonWriter.writeFieldName("LogGroupName");
+        jsonWriter.writeValue(config.getLogGroupName());
+
+        // Write CloudWatchMetrics array
+        jsonWriter.writeFieldName("CloudWatchMetrics");
+        jsonWriter.writeStartArray();
+        jsonWriter.writeStartObject();
```
zoewangg, line 262: Is this correct? It seems this CloudWatchMetrics only has one object

metric-publishers/emf-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/emf/MetricEmfConverter.java
```diff
@@ -0,0 +1,368 @@
...
+
+    private void writeMetricValues(JsonWriter jsonWriter, Map<String, List<Object>> metrics) {
+        for (Map.Entry<String, List<Object>> entry : metrics.entrySet()) {
+            String metricName = entry.getKey();
+            List<Object> values = entry.getValue();
+
+            // For dimension metrics, write the last value
+            if (isDimension(metricName)) {
+                jsonWriter.writeFieldName(metricName);
+                jsonWriter.writeValue((String) values.get(values.size() - 1));
```
zoewangg, line 312: Why are we writing the last value?

metric-publishers/emf-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/emf/MetricEmfConverter.java
```diff
@@ -0,0 +1,368 @@
...
+    }
+
+
+    private boolean isDimension(String metricName) {
+        return metricName != null && config.getDimensionStrings().contains(metricName);
+    }
+
+    private boolean hasUnit(String metricName) {
+        // Implement logic to determine if metric should have a unit
+        return metricName.contains("Duration");
```
zoewangg, line 342: Hmm, should we use type to determine instead? We have other numeric metrics that doesn't contain `Duration` such as `PendingConcurrencyAcquires`

metric-publishers/emf-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/emf/MetricEmfConverter.java
```diff
@@ -0,0 +1,368 @@
...
+    }
+
+    private String getMetricUnit(String metricName) {
+        if (metricName.endsWith("Duration")) {
```
zoewangg, line 346: Same here

metric-publishers/emf-metric-publisher/pom.xml
```diff
@@ -0,0 +1,75 @@
...
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>software.amazon.awssdk</groupId>
+        <artifactId>metric-publishers</artifactId>
+        <version>2.29.35-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>emf-metric-publisher</artifactId>
```
zoewangg, line 27: We don't have to change it now, but we should discuss it in API surface area review. Should we rename it to `emf-metric-logger`, since technically it's not publishing metrics?

metric-publishers/emf-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/emf/EmfMetricPublisher.java
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+
+    @Override
+    public void publish(MetricCollection metricCollection) {
+        if (metricCollection == null) {
+            return;
+        }
+
+        List<String> emfStrings = metricConverter.convertMetricCollectionToEmf(metricCollection);
```
zoewangg, line 91: We should probably add a try-catch here because we don't want to fail the request if metrics publishing fails.

metric-publishers/emf-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/emf/MetricEmfConverter.java
```diff
@@ -0,0 +1,368 @@
...
+ * - Boolean values are converted to numeric values (1.0 for true, 0.0 for false)
+ * - Null values are converted to 0.0
+ * - Very small numeric values (below 0.0001) are normalized to 0.0
+ * - Duration values are converted to milliseconds
+ *
+ * @since 2.0
+ * @see EmfMetricConfiguration
+ */
+@SdkInternalApi
+public class MetricEmfConverter {
```
zoewangg, line 58: Missing unit tests for this class

metric-publishers/emf-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/emf/EmfMetricPublisher.java
```diff
@@ -0,0 +1,202 @@
...
+    @Override
+    public void close() {
+    }
+
+    public static final class Builder {
+        String namespace;
+        String logGroupName;
+        Collection<SdkMetric<String>> dimensions;
+        Collection<MetricCategory> metricCategories;
+        MetricLevel metricLevel;
```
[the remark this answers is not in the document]
Fred1155: The EmfMetricConfig class need the access to read that, do i need to write get methods for them in the builder?

metric-publishers/emf-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/emf/MetricEmfConverter.java
Fred1155: DimensionString is the dimension string user passes in to the config, but I think user's may pass in dimensions that do not really exist in the metricCollection,  realDimensionString is the strings that will be output in the emf String, which both exists in user's config and metricCollection

metric-publishers/emf-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/emf/MetricEmfConverter.java
Fred1155: Yep I'm changing to using dependency injection now

metric-publishers/emf-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/emf/MetricEmfConverter.java
Fred1155: So should I test the converting method in this class and the publish method in the main class?

zoewangg: Yes, ideally, each class should have its own test focusing on its own functionality.

metric-publishers/emf-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/emf/MetricEmfConverter.java
Fred1155: I think so. In the example https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch_Embedded_Metric_Format_Specification.html, CloudWatchMetrics should have one object, because we only have one namespace and one set of dimensions

metric-publishers/emf-metric-publisher/pom.xml
```diff
@@ -0,0 +1,75 @@
...
+                    </archive>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
+
+    <dependencies>
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>sdk-core</artifactId>
```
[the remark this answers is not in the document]
Fred1155: The only core part code used is core.exception.SdkClientException

metric-publishers/emf-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/emf/internal/MetricEmfConverter.java
```diff
@@ -0,0 +1,339 @@
...
+
+
+    private void writeMetricValues(JsonWriter jsonWriter, Map<String, List<MetricRecord<?>>> metrics) {
+        for (Map.Entry<String, List<MetricRecord<?>>> entry : metrics.entrySet()) {
+            String metricName = entry.getKey();
+            List<MetricRecord<?>> records = entry.getValue();
+
+            if (isDimension(metricName)) {
+                jsonWriter.writeFieldName(metricName);
+                jsonWriter.writeValue((String) records.get(0).value());
```
zoewangg, line 291: I understand dimension metric could only have one element, but for the sake of defensive coding, let's add validation here.

metric-publishers/emf-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/emf/internal/MetricEmfConverter.java
```diff
@@ -0,0 +1,339 @@
...
+            } else {
+                if (records.get(0).value() instanceof String) {
+                    continue;
+                }
```
zoewangg, line 295: Should we use `SdkMetric` as the key instead of String (metricName) of aggregated metrics? This way, we don't need to do this awkward type check and we can just use `entry.getKey().valueClass() instanceof String `

metric-publishers/emf-metric-publisher/src/test/java/software/amazon/awssdk/metrics/publishers/emf/EmfMetricPublisherTest.java
```diff
@@ -0,0 +1,67 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import software.amazon.awssdk.http.HttpMetric;
+import software.amazon.awssdk.metrics.MetricCollector;
+import software.amazon.awssdk.testutils.LogCaptor;
+
+
+public class EmfMetricPublisherTest extends LogCaptor.LogCaptorTestBase{
```
zoewangg, line 27: Can we add a test case where metricCollector throws an exception?

metric-publishers/emf-metric-publisher/src/test/java/software/amazon/awssdk/metrics/publishers/emf/internal/MetricEmfConverterTest.java
```diff
@@ -0,0 +1,178 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import software.amazon.awssdk.http.HttpMetric;
+import software.amazon.awssdk.metrics.MetricCategory;
+import software.amazon.awssdk.metrics.MetricCollector;
+import software.amazon.awssdk.metrics.MetricLevel;
+
+public class MetricEmfConverterTest {
```
zoewangg, line 33: Can we add a test case where metric records size is larger than 100?

metric-publishers/emf-metric-publisher/src/test/java/software/amazon/awssdk/metrics/publishers/emf/internal/MetricEmfConverterTest.java
```diff
@@ -0,0 +1,178 @@
...
+    @Test
+    void ConvertMetricCollectionToEMF_Dimensions(){
+
+        MetricCollector metricCollector = MetricCollector.create("test");
+        metricCollector.reportMetric(HttpMetric.HTTP_CLIENT_NAME, "apache-http-client");
+        metricCollector.reportMetric(HttpMetric.AVAILABLE_CONCURRENCY, 5);
+        metricCollector.reportMetric(HttpMetric.HTTP_STATUS_CODE, 404);
+        List<String> emfLogs = metricEmfConverterDefault.convertMetricCollectionToEmf(metricCollector.collect());
+
+        for (String emfLog : emfLogs) {
```
zoewangg, line 93: If we only expect one log statement, let's just assert that instead of for loop. Same for other test cases.

```
assertThat(emfLogs).containsOnly(...)
```

metric-publishers/emf-metric-publisher/src/test/java/software/amazon/awssdk/metrics/publishers/emf/internal/MetricEmfConverterTest.java
```diff
@@ -0,0 +1,178 @@
...
+import software.amazon.awssdk.http.HttpMetric;
+import software.amazon.awssdk.metrics.MetricCategory;
+import software.amazon.awssdk.metrics.MetricCollector;
+import software.amazon.awssdk.metrics.MetricLevel;
+
+public class MetricEmfConverterTest {
+
+    private MetricEmfConverter metricEmfConverterDefault;
+    private MetricEmfConverter metricEmfConverterCustom;
+    private final EmfMetricConfiguration testConfigDefault = new EmfMetricConfiguration("AwsSdk/Test/JavaSdk2",
```
zoewangg, line 37: Is this the default config though?

metric-publishers/emf-metric-publisher/src/test/java/software/amazon/awssdk/metrics/publishers/emf/internal/MetricEmfConverterTest.java
```diff
@@ -0,0 +1,178 @@
...
+        for (String emfLog : emfLogs) {
+            assertThat(emfLog).isEqualTo("{\"_aws\":{\"Timestamp\":12345.678,\"LogGroupName\":\"my_log_group_name\",\"CloudWatchMetrics\":[{\"Namespace\":"
+                                         + "\"AwsSdk/Test/JavaSdk2\",\"Dimensions\":[[]],"
+                                         + "\"Metrics\":[{\"Name\":\"AvailableConcurrency\"}]}]},\"AvailableConcurrency\":5}");
+
+        }
+    }
+
+    @Test
+    void ConvertMetricCollectionToEMF_metricLevel(){
```
zoewangg, line 117: How is this test different from `ConvertMetricCollectionToEMF_metricCategory`?

metric-publishers/emf-metric-publisher/src/test/java/software/amazon/awssdk/metrics/publishers/emf/internal/MetricEmfConverterTest.java
```diff
@@ -0,0 +1,168 @@
...
+import java.util.stream.Stream;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import software.amazon.awssdk.core.metrics.CoreMetric;
+import software.amazon.awssdk.http.HttpMetric;
+import software.amazon.awssdk.metrics.MetricCategory;
+import software.amazon.awssdk.metrics.MetricCollector;
+import software.amazon.awssdk.metrics.MetricLevel;
+
+public class MetricEmfConverterTest {
```
zoewangg, line 34: Can we add a test case where the number of metrics exceed 100?

Can we also add a test case where we drop metric values that are not numeric types?

metric-publishers/emf-metric-publisher/src/main/resources/log4j2.properties
```diff
@@ -0,0 +1,38 @@
+#
```
zoewangg, line 1: We should not add a default log4j config for customers.

zoewangg (commented):
It's probably easier if you see the code, so I pushed suggested changes for `MetricEmfConverter` here :) Feel free to change as needed. https://github.com/aws/aws-sdk-java-v2/compare/bole/EMFMetricPublisher...zoewang/reviewPR5752

metric-publishers/emf-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/emf/EmfMetricPublisher.java
zoewangg: Discussed offline, this is not a blocker

metric-publishers/emf-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/emf/EmfMetricPublisher.java
```diff
@@ -0,0 +1,213 @@
...
+@Immutable
+@SdkPublicApi
+public final class EmfMetricPublisher implements MetricPublisher {
+
+    private static final Logger logger = Logger.loggerFor("software.amazon.awssdk.metrics.publishers.emf");
+    private final MetricEmfConverter metricConverter;
+
+
+    private EmfMetricPublisher(Builder builder) {
+        EmfMetricConfiguration config = new EmfMetricConfiguration(
```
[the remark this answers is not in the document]
zoewangg: Discussed offline, we are keeping EmfMetricConfiguration and using builder pattern for it.

metric-publishers/emf-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/emf/EmfMetricPublisher.java
```diff
@@ -0,0 +1,213 @@
...
+        public Builder metricLevel(MetricLevel metricLevel) {
+            this.metricLevel = metricLevel;
+            return this;
+        }
+
+        /**
+         * Configure the LogGroupName key that will be put into the emf log to this publisher. This is a required key for
+         * using the CloudWatch agent to send embedded metric format logs that tells the agent which log group to use.
+         *
+         * <p>If this is not specified, {@code /aws/emf/metrics} will be used.
```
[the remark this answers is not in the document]
zoewangg: Discussed offline, this is required for cw agent and we decided to require users to provide it instead of reading the lambda env var because 1) this emf publisher is not specific to Lambda 2) defaulting to log to the default lambda logger may pollute customers' logs since it'll introduce more logs. 3) it's a two way door

metric-publishers/emf-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/emf/internal/MetricEmfConverter.java
```diff
@@ -0,0 +1,339 @@
...
+        } else if (Boolean.class.isAssignableFrom(valueClass)) {
+            jsonWriter.writeValue(value.equals(true) ? 1.0 : 0.0);
+        } else if (Duration.class.isAssignableFrom(valueClass)) {
+            Duration duration = (Duration) value;
+            double millisValue = duration.toMillis();
+            jsonWriter.writeValue(millisValue);
+        } else if (Double.class.isAssignableFrom(valueClass)) {
+            double doubleValue = (Double) value;
+            if (Math.abs(doubleValue) < ZERO_THRESHOLD) {
+                jsonWriter.writeValue(0.0);
```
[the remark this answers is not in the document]
zoewangg: Discussed offline, this is the existing behavior in CW metric publisher. https://github.com/aws/aws-sdk-java-v2/blob/3f81c5c99c6e0f23960bd58cfca65bf366b74ea5/metric-publishers/cloudwatch-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/cloudwatch/internal/transform/MetricValueNormalizer.java Decided to copy the class to utils module so that it can be reused in this module.

metric-publishers/emf-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/emf/internal/EmfMetricConfiguration.java
```diff
@@ -0,0 +1,134 @@
...
+    private final MetricLevel metricLevel;
+    private final List<String> dimensionStrings;
+
+    private EmfMetricConfiguration(Builder builder) {
+        this.namespace = builder.namespace == null ? DEFAULT_NAMESPACE : builder.namespace;
+        this.logGroupName = builder.logGroupName == null ? DEFAULT_LOG_GROUP_NAME : builder.logGroupName;
+        this.dimensions = builder.dimensions == null ? DEFAULT_DIMENSIONS : new HashSet<>(builder.dimensions);
+        this.metricCategories = builder.metricCategories == null ? DEFAULT_CATEGORIES : new HashSet<>(builder.metricCategories);
+        this.metricLevel = builder.metricLevel == null ? DEFAULT_METRIC_LEVEL : builder.metricLevel;
+        this.dimensionStrings = resolveDimensionStrings(builder.dimensions);
```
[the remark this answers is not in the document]
zoewangg: This comment can be ignored now since we can remove dimensionStrings

metric-publishers/emf-metric-publisher/pom.xml
```diff
@@ -0,0 +1,81 @@
...
+        </dependency>
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>json-utils</artifactId>
+            <version>${awsjavasdk.version}</version>
+            <scope>compile</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.logging.log4j</groupId>
+            <artifactId>log4j-slf4j-impl</artifactId>
```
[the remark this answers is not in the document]
Fred1155: 👍

metric-publishers/emf-metric-publisher/pom.xml
```diff
@@ -0,0 +1,81 @@
...
+            <version>2.20.0</version>
+        </dependency>
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>sdk-core</artifactId>
```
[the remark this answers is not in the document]
Fred1155: The default Dimensions: CoreMetric.SERVICE_ID, CoreMetric.OPERATION_NAME😞

metric-publishers/emf-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/emf/EmfMetricPublisher.java
```diff
@@ -0,0 +1,222 @@
...
+
+    private EmfMetricPublisher(Builder builder) {
+        EmfMetricConfiguration config = new EmfMetricConfiguration.Builder()
+            .namespace(builder.namespace)
+            .logGroupName(builder.logGroupName)
+            .dimensions(builder.dimensions)
+            .metricLevel(builder.metricLevel)
+            .metricCategories(builder.metricCategories)
+            .build()
+        ;
```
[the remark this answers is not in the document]
Fred1155: 👍

metric-publishers/emf-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/emf/EmfMetricPublisher.java
```diff
@@ -0,0 +1,222 @@
...
+         * publisher.
+         *
+         * <p>If this is not specified,  {@link CoreMetric#SERVICE_ID} and {@link CoreMetric#OPERATION_NAME} will be used.
+         */
+        public Builder dimensions(Collection<SdkMetric<String>> dimensions) {
+            this.dimensions = new ArrayList<>(dimensions);
+            return this;
+        }
+
+        @SafeVarargs
```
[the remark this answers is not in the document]
Fred1155: 👍

metric-publishers/emf-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/emf/internal/MetricEmfConverter.java
```diff
@@ -0,0 +1,376 @@
...
+            jsonWriter.writeFieldName("Unit");
+            jsonWriter.writeValue(unit);
+        }
+
+        jsonWriter.writeEndObject();
+    }
+
+    private void writeMetricValues(JsonWriter jsonWriter, Map<String, List<MetricRecord<?>>> metrics) {
+        metrics.forEach((metricName, records) -> {
+            if (records.isEmpty()) {
```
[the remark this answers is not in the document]
Fred1155: I'm not sure what should be put on the emf string if it is empty. since "A metric target MUST be a numeric value or an array of numeric values" Should I put an empty array?

metric-publishers/emf-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/emf/EmfMetricPublisher.java
```diff
@@ -0,0 +1,222 @@
...
+ *
+ * <p>
+ * The EMF publisher converts metric collections into JSON-formatted log entries that conform to the
+ * CloudWatch EMF specification. These logs are written to the "/aws/emf/metrics" log group by default.
+ * CloudWatch automatically processes these logs to generate corresponding metrics that can be used for
+ * monitoring and alerting.
+ * </p>
+ *
+ * <pre>
+ * // Create a CloudWatchMetricPublisher using a custom namespace.
```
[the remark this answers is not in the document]
Fred1155: @snippet seems only support language level 18 and above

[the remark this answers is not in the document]
Fred1155: `@snippet` seems only be supported by language level 18 and above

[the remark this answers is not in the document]
zoewangg: Yes, and we use Java 23 to build our javadocs :) https://github.com/aws/aws-sdk-java-v2/blob/master/buildspecs/release-javadoc.yml#L8

metric-publishers/emf-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/emf/internal/MetricEmfConverter.java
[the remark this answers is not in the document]
zoewangg: Yes, I'd say empty array.

metric-publishers/emf-metric-publisher/src/test/java/software/amazon/awssdk/metrics/publishers/emf/EmfMetricPublisherTest.java
```diff
@@ -38,14 +39,15 @@ void setUp() {
...
     void Publish_NoLogGroupName() {
-        try {
+        assertThatThrownBy(() -> {
             EmfMetricPublisher publisher = publisherBuilder.build();
             publisher.publish(null);
-        } catch (Exception e) {
-            assertThat(e).hasMessage("logGroupName must be configured for publishing emf format log");
-        }
+        })
+            .isInstanceOf(Exception.class)
```
zoewangg, line 46: Exception.class is a pretty generic type and we should probably assert the exact type in tests, suggesting `NullPointerException`

metric-publishers/emf-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/emf/internal/MetricEmfConverter.java
```diff
@@ -47,8 +50,11 @@
  */
 @SdkInternalApi
 public class MetricEmfConverter {
+    private static final int MAX_RECORD_SIZE = 100;
+    private static final int MAX_METRIC_NUM = 100;
```
zoewangg, line 54: We could use more explanations on those numbers. This is one of the few places where comments are beneficial :)

metric-publishers/emf-metric-publisher/src/test/java/software/amazon/awssdk/metrics/publishers/emf/internal/MetricEmfConverterTest.java
```diff
@@ -165,4 +166,16 @@ void ConvertMetricCollectionToEMF_OverSizedRecords(){
 
     }
 
+    @Test
+    void ConvertMetricCollectionToEMF_LargeCollection(){
```
zoewangg, line 170: Let's be more descriptive on the naming following `methodToTest_when_should`, suggesting `ConvertMetricCollectionToEMF_metricCollectionExceedsMaxSize_shouldSplit`

metric-publishers/emf-metric-publisher/src/test/java/software/amazon/awssdk/metrics/publishers/emf/internal/MetricEmfConverterTest.java
zoewangg: Suggesting renaming it to `convertMetricCollectionToEmf_traceEnabled_shouldIncludeTrace` to make it more clear what we are testing

metric-publishers/emf-metric-publisher/src/test/java/software/amazon/awssdk/metrics/publishers/emf/internal/MetricEmfConverterTest.java
zoewangg: Any updates on the test case where we drop metric values that are not numeric types?

metric-publishers/emf-metric-publisher/src/test/java/software/amazon/awssdk/metrics/publishers/emf/internal/MetricEmfConverterTest.java
```diff
@@ -178,4 +189,46 @@ void ConvertMetricCollectionToEMF_LargeCollection(){
...
+            JsonSchemaFactory factory = JsonSchemaFactory.getInstance(SpecVersion.VersionFlag.V4);
+            JsonSchema schema = factory.getSchema(jsonSchema);
+
+
+                ObjectMapper mapper = new ObjectMapper();
+                JsonNode jsonNode = mapper.readTree(emfLogs.get(0));
+                Set<ValidationMessage> errors = schema.validate(jsonNode);
+
+                assertThat(errors).isEmpty();
+        } catch (Exception e){
```
zoewangg, line 231: This will silently fail. I actually meant not catching the exception and just letting it add `throws XXX` in the method signature

metric-publishers/emf-metric-publisher/src/test/java/software/amazon/awssdk/metrics/publishers/emf/internal/MetricEmfConverterTest.java
```diff
@@ -178,4 +189,46 @@ void ConvertMetricCollectionToEMF_LargeCollection(){
...
+
+            List<String> emfLogs = metricEmfConverterDefault.convertMetricCollectionToEmf(metricCollector.collect());
+
+            JsonSchemaFactory factory = JsonSchemaFactory.getInstance(SpecVersion.VersionFlag.V4);
+            JsonSchema schema = factory.getSchema(jsonSchema);
+
+
+                ObjectMapper mapper = new ObjectMapper();
+                JsonNode jsonNode = mapper.readTree(emfLogs.get(0));
+                Set<ValidationMessage> errors = schema.validate(jsonNode);
```
zoewangg, line 228: formatting is a bit weird

metric-publishers/emf-metric-publisher/src/test/java/software/amazon/awssdk/metrics/publishers/emf/internal/MetricEmfConverterTest.java
```diff
@@ -178,4 +189,46 @@ void ConvertMetricCollectionToEMF_LargeCollection(){
...
+    void ConvertMetricCollectionToEMF_shouldConformToSchema() {
+        try {
+            String jsonSchema = IOUtils.toString(
+                Objects.requireNonNull(getClass().getResourceAsStream("/emfSchema.json")),
+                StandardCharsets.UTF_8
+            );
+
+            MetricCollector metricCollector = MetricCollector.create("test");
+            metricCollector.reportMetric(HttpMetric.AVAILABLE_CONCURRENCY, 5);
+            metricCollector.reportMetric(CoreMetric.SERVICE_ID, "serviceId");
```
zoewangg, line 218: Can we add more metrics to simulate real-case scenarios? especially the ones with multiple values

metric-publishers/emf-metric-publisher/pom.xml
```diff
@@ -71,6 +67,18 @@
             <version>${awsjavasdk.version}</version>
             <scope>compile</scope>
         </dependency>
+        <dependency>
+            <groupId>com.networknt</groupId>
+            <artifactId>json-schema-validator</artifactId>
+            <version>1.4.0</version>
```
zoewangg, line 73: Looks like 1.5.4 is the latest version https://github.com/networknt/json-schema-validator/blob/master/pom.xml#L26C11-L26C17 

We should define all versions in the top pom.xml to make it easier to manage, see example: https://github.com/aws/aws-sdk-java-v2/blob/master/pom.xml#L146

metric-publishers/emf-metric-publisher/pom.xml
```diff
@@ -71,6 +67,18 @@
...
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>commons-io</groupId>
+            <artifactId>commons-io</artifactId>
+            <version>2.11.0</version>
```
zoewangg, line 79: We probably don't need this dependency since we have our own `IoUtils`

zoewangg (approved):
Good job!

Issue #5247, opened 2024-05-27
InstanceProfileCredentialsProvider unable to refresh/recover from throttling or network problems at time of credential cache refresh


steveloughran:
### Describe the bug

The full details and analysis are in [HADOOP-19181. IAMCredentialsProvider throttle failures](https://issues.apache.org/jira/browse/HADOOP-19181)

* the credentials retrieved by IAMCredentialsProvider are set to expire() only 1s before the expiry time returned by the Instance metadata service
* this is not enough time for the CachedSupplier to recover from any failure as on a failure it sets a jitter on a retry to `ComparableUtils.minimum(Duration.ofMillis(exponentialBackoffMillis), Duration.ofSeconds(10))`. That is: up to 9 seconds after the credentials expire.

It's notable that `ContainerCredentialsProvider` has more resilience in
* credentials are set to expire 15 minutes before the credentials become invalid
* it also retries 5 times with no sleep on a request failure.

### Expected Behavior

`IAMCredentialsProvider` to always issue valid credentials. This should include asynchronous refreshing of credentials when enabled, far enough in advance of their expiry that recovery attempts can be repeated.

### Current Behavior

invalid credentials were returned when requesting credentials to sign a request.

### Reproduction Steps

see the hadoop bug report. The deployment was a single EC2 node running many services, long enough for the EC2 credentials to expire and need refreshing.

### Possible Solution

I really don't see what we can do in the hadoop codebase to recover from this. We could consider extending our own `IAMInstanceCredentialsProvider` to wrap the retry failures with our own sleep/retry, but as it'd take > 10 seconds, API calls awaiting signing (e.g. S3 Express CreateSession) will time out.

I'd propose copying ContainerCredentialsProvider, if not with the retries then at least the expiry time many minutes ahead of actual credential expiry

### AWS Java SDK version used

2.24.6

### JDK version used

an openjdk java 8 build

### Operating System and version

linux

debora-ito:
Hi @steveloughran 

we added a task to analyze and improve the credential experience. Thank you for the detailed bug report - [HADOOP-19181](https://issues.apache.org/jira/browse/HADOOP-19181).

steveloughran:
Is there an approximate timeline for this? I don't want to have to reimplement our own refresh thread code -not because it is hard, but because maintenance and testing with fault injection are the pain points.

steveloughran:
[...]

PR #5758, opened 2024-12-18
Add stale time config to InstanceProfileCredentialsProvider

L-Applin:
Add stale time config to `InstanceProfileCredentialsProvider` to allow for refreshing credentials earlier due to stale value. This will help prevent returning invalid credentials when an error is encountered during asynchronous refresh.

## Motivation and Context
Fixes for #5247 , this will allow user to provide a higher value for prefetch to prevent returning expired credentails without changing default value. Increasing this value may lead to a higher rate of request to IMDS, so we ought to avoid changing the base default value for it.

## Modifications
- Added the `staleTime` configuration option to `InstanceProfileCredentialsProvider`
- Added retry strategy for `InstanceProfileCredentialsProvider`

## Testing
Added unit tests
## Types of changes


core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/StaticResourcesEndpointProvider.java
```diff
@@ -58,6 +59,11 @@ public Map<String, String> headers() {
         return Collections.unmodifiableMap(headers);
     }
 
+    @Override
+    public ResourcesEndpointRetryPolicy retryPolicy() {
+        return new ContainerCredentialsRetryPolicy();
+    }
```
zoewangg, line 65: I don't think we can add retry at this point for IMDS credentials provider because it may increase latency and break customers.

L-Applin: right, that was a ask from Hadoop. Instead we can add it as an optional config as well

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/InstanceProfileCredentialsProvider.java
```diff
@@ -174,7 +178,7 @@ private Instant staleTime(Instant expiration) {
             return null;
         }
 
-        return expiration.minusSeconds(1);
+        return expiration.minus(staleTime);
```
joviegas, line 181: I was trying to understand comment in https://github.com/aws/aws-sdk-java-v2/pull/3314#discussion_r926112575

Looks like the 1 sec is intentional , if a user configures an excessively large stale time (e.g., expiration.minusMinutes(10)), credentials might be marked as stale too early, causing unnecessary refreshes and increased load on the credential provider. WDYT ?

steveloughran: the default must be > the retry sleep delay of 10s. Otherwise when a call is throttled it is going to fail.

steveloughran (changes requested):
you need to change that default to be > 10s; without that any 503 response is not going be recovered from.

Those of us who have encountered the problem will know to explicitly change it, but people who haven't yet hit it will only find out when things go wrong.

There is also the subtlety that all processes running in the same VM will be using the same timeout, as it is based on when the shared credentials expire. The more processes running, the more risk of that 503 -regardless of what the value is. This means that any constant value can create a Thundering Herd.

The "extra load" concern is marginal, given refresh time is every hour. spreading out that load is likely to be better as any herd-triggered retry is load in itself.

I'd suggest some larger number and maybe a jitter of a few seconds to spread out load from many processes.

L-Applin:
> you need to change that default to be > 10s; without that any 503 response is not going be recovered from.
> 
> Those of us who have encountered the problem will know to explicitly change it, but people who haven't yet hit it will only find out when things go wrong.
> 
> There is also the subtlety that all processes running in the same VM will be using the same timeout, as it is based on when the shared credentials expire. The more processes running, the more risk of that 503 -regardless of what the value is. This means that any constant value can create a Thundering Herd.
> 
> The "extra load" concern is marginal, given refresh time is every hour. spreading out that load is likely to be better as any herd-triggered retry is load in itself.
> 
> I'd suggest some larger number and maybe a jitter of a few seconds to spread out load from many processes.

There is already jitter, up to 1 minute before expiration time. Also, I am weary of changing default values, it may cause unexpected behaviour change in user applications and we try to avoid those. Is the ability to change to value manually by yourself not a viable solution for your use case?

L-Applin:
> I'm happy with the change -but do think you should consider documenting this for others

Good point, I added a note about the issue in the javadoc for `staleTime`

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/InstanceProfileCredentialsProvider.java
L-Applin: yes 1 sec was intentional but as @steveloughran mentioned, it causes issues when retrying during prefetch, it may lead to situation, specifically during high load, where it is not long enough to ensure the SDK returns valid credentials

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/StaticResourcesEndpointProvider.java
```diff
@@ -58,10 +62,16 @@ public Map<String, String> headers() {
...
+    @Override
+    public ResourcesEndpointRetryPolicy retryPolicy() {
+        return this.retryPolicy;
+    }
+
     public static class Builder {
         private URI endpoint;
         private Map<String, String> additionalHeaders = new HashMap<>();
         private Duration customTimeout;
+        private ResourcesEndpointRetryPolicy retryPolicy = ResourcesEndpointRetryPolicy.NO_RETRY;
```
zoewangg, line 74: Can we set the default in the ctor instead? Ideally, the builder shouldn't have any logic

zoewangg (approved):

PR #5762, opened 2024-12-26
Sigv4aSigningRegionSetProvider and updating Execution Attributes

joviegas:
## Motivation and Context

- RegionSet for Sigv4a configuration must also be supported as an environment variable called AWS_SIGV4A_SIGNING_REGION_SET and as a configuration file property called sigv4a_signing_region_set. In both of these locations, the value is a non-empty, comma-delmited list of non-empty strings. They follow the standard precedence rules of in-code configuration, then environment configuration, then file-based configuration.

## Types of changes

## Checklist

## License


core/aws-core/src/main/java/software/amazon/awssdk/awscore/endpoint/Sigv4aSigningRegionSetProvider.java
```diff
@@ -0,0 +1,111 @@
...
+
+        ProfileFile file = this.profileFile.get();
+        Optional<Profile> profile = file.profile(profileName());
+        return profile
+            .flatMap(p -> p.property(ProfileProperty.SIGV4A_SIGNING_REGION_SET))
+            .flatMap(this::parseRegionSet)
+            .orElse(null);
+    }
+
+    private Optional<Set<String>> parseRegionSet(String value) {
```
zoewangg, line 73: We don't normally use Optional with collection or map. Should we just return empty set if the value is blank?

test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/Sigv4aSigningRegionSetTest.java
```diff
@@ -0,0 +1,215 @@
+package software.amazon.awssdk.services;
```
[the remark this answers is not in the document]
joviegas: Moved to auth folder now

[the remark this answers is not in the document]
joviegas: added

core/aws-core/src/main/java/software/amazon/awssdk/awscore/endpoint/Sigv4aSigningRegionSetProvider.java
joviegas: Agree , Updated to return Null directly if we donot find valid non empty values

core/aws-core/src/main/java/software/amazon/awssdk/awscore/endpoint/Sigv4aSigningRegionSetProvider.java
```diff
@@ -0,0 +1,109 @@
...
+import software.amazon.awssdk.core.SdkSystemSetting;
+import software.amazon.awssdk.profiles.Profile;
+import software.amazon.awssdk.profiles.ProfileFile;
+import software.amazon.awssdk.profiles.ProfileFileSystemSetting;
+import software.amazon.awssdk.profiles.ProfileProperty;
+import software.amazon.awssdk.utils.Lazy;
+import software.amazon.awssdk.utils.StringUtils;
+import software.amazon.awssdk.utils.Validate;
+
+@SdkProtectedApi
```
zoewangg, line 35: Should this be internal API? Or is this class used outside of this module? (Sorry, didn't notice this previously)

core/aws-core/src/main/java/software/amazon/awssdk/awscore/endpoint/Sigv4aSigningRegionSetProvider.java
zoewangg: The reason we don't use Optional with collection is that we can use empty to avoid null handling. Should we return empty set in that case?

core/aws-core/src/main/java/software/amazon/awssdk/awscore/endpoint/Sigv4aSigningRegionSetProvider.java
joviegas: I referred to existing
```
@SdkProtectedApi
public class DualstackEnabledProvider {
    private final Supplier<ProfileFile> profileFile;
    private final String profileName;
```
    
What do you recommend ?

core/aws-core/src/main/java/software/amazon/awssdk/awscore/endpoint/Sigv4aSigningRegionSetProvider.java
joviegas: Empty Set is like supporting "  " regionset
If we allow empty Set it like the set had a entry of empty Value.
The SEP says its a non empty value if it exist.
WDYT ?

zoewangg: I meant using empty set to indicate no value is available so that we don't need to handle null. We can check whether it's empty in the caller code.

joviegas: Updated to return Empty Set also updated test case

zoewangg (approved):

PR #5767, opened 2024-12-31
Remove IoT 1-Click services - Devices and Projects

davidh44:
## Motivation and Context

IoT 1 Click services are being shut down


joviegas:
We need to handle the reference in 
https://github.com/aws/aws-sdk-java-v2/blob/61d16e083a27a146665aec19f14bfd31e54e9f83/v2-migration/src/main/resources/META-INF/rewrite/upgrade-sdk-dependencies.yml#L110-L116

Either we need to delete the reference of deprecated version
or Pin the last version where this was supported

davidh44:
> We need to handle the reference in
> 
> https://github.com/aws/aws-sdk-java-v2/blob/61d16e083a27a146665aec19f14bfd31e54e9f83/v2-migration/src/main/resources/META-INF/rewrite/upgrade-sdk-dependencies.yml#L110-L116
> 
> Either we need to delete the reference of deprecated package or Pin the last version where this was supported
> 
> I think removing these services is fine since the service themselves are deprecated . WDYT ?

Good catch, yeah I think removing is best, pushed update

joviegas (approved):

PR #5768, opened 2025-01-01
Codegen changes for AuthSchemeInterceptorSpecs to update the RegionSet from Execution Attributes

joviegas:
## Motivation and Context

- Modify interceptor to update the RegionSet from Execution Attributes
- Fall back region is always configured Region if no RegionSet configured on Client (Requirement doc)
(Note Will raise a separate PR later for EndpointBased Auth case that is legacy service like SES , S3 etc)
## Modifications

Generated Code looks like below
```java
    private MultiauthAuthSchemeParams authSchemeParams(SdkRequest request, ExecutionAttributes executionAttributes) {
        String operation = executionAttributes.getAttribute(SdkExecutionAttribute.OPERATION_NAME);
        MultiauthAuthSchemeParams.Builder builder = MultiauthAuthSchemeParams.builder().operation(operation);
        Region region = executionAttributes.getAttribute(AwsExecutionAttribute.AWS_REGION);
        builder.region(region);
        // Comment on PR : This code is generate only if Service has Sigv4a Auth
        RegionSet regionSet = executionAttributes.getOptionalAttribute(AwsExecutionAttribute.AWS_SIGV4A_SIGNING_REGION_SET)
                .filter(regions -> !regions.isEmpty()).map(regions -> RegionSet.create(String.join(", ", regions)))
                .orElseGet(() -> {
                    Region fallbackRegion = executionAttributes.getAttribute(AwsExecutionAttribute.AWS_REGION);
                    return fallbackRegion != null ? RegionSet.create(fallbackRegion.toString()) : null;
                });
        ;
        builder.regionSet(regionSet);
        return builder.build();
```

## Testing

- Junits added for codegen
- Junits for interceptors to check AuthScheme Params are updated

## License

### DIFF
--- a/codegen/src/main/java/software/amazon/awssdk/codegen/poet/auth/scheme/AuthSchemeInterceptorSpec.java
+++ b/codegen/src/main/java/software/amazon/awssdk/codegen/poet/auth/scheme/AuthSchemeInterceptorSpec.java
@@ -158,7 +158,7 @@ public final class AuthSchemeInterceptorSpec implements ClassSpec {
                                      AwsExecutionAttribute.class);
                 builder.addStatement("builder.region(region)");
             }
-            addRegionSet(builder);
+            generateSigv4aRegionSet(builder);
             builder.addStatement("return builder.build()");
             return builder.build();
         }
@@ -449,7 +449,7 @@ public final class AuthSchemeInterceptorSpec implements ClassSpec {
         return result;
     }
 
-    private void addRegionSet(MethodSpec.Builder builder) {
+    private void generateSigv4aRegionSet(MethodSpec.Builder builder) {
         if (authSchemeSpecUtils.usesSigV4a()) {
             builder.addStatement(
                 "$T regionSet = executionAttributes.getOptionalAttribute($T.AWS_SIGV4A_SIGNING_REGION_SET)\n" +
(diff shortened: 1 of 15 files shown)


codegen/src/main/java/software/amazon/awssdk/codegen/poet/auth/scheme/AuthSchemeInterceptorSpec.java
```diff
@@ -148,20 +149,17 @@ private MethodSpec generateAuthSchemeParams() {
         if (!authSchemeSpecUtils.useEndpointBasedAuthProvider()) {
             builder.addStatement("$T operation = executionAttributes.getAttribute($T.OPERATION_NAME)", String.class,
                                  SdkExecutionAttribute.class);
+            builder.addStatement("$T.Builder builder = $T.builder().operation(operation)",
```
joviegas, line 152: This part is needed because earlier Builder was built in a single line
Since we have conditional bases builder for sigv4a so splited this generation code
```
QueryAuthSchemeParams.builder().operation(operation).region(region).build();
```

test/codegen-generated-classes-test/src/main/resources/codegen-resources/multiauth/customization.config
```diff
@@ -0,0 +1,4 @@
+{
+    "skipEndpointTestGeneration": true,
+     "useMultiAuth": true
```
zoewangg, line 3: Discussed offline, we should consider removing useMultiAuth

zoewangg (approved):

test/codegen-generated-classes-test/src/main/resources/codegen-resources/multiauth/customization.config
joviegas: Added `JAVA-8023` and also added in planning

PR #5772, opened 2025-01-04
Add sigv4aRegionSet to Client Builder for Multi-Auth Services Supporting Sigv4a

joviegas:
## Motivation and Context

- Clients with Sigv4a Multi auth should provide API on client to set the region set
- The Sigv4aRegion Set build takes highest Precedence

Note : Even though we have [RegionScope](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/regions/RegionScope.html)  this has been not used because SRA introduced RegionSet so using the latest representation class for 
Region set
### Interface

- Code gen changes that will generate API on Client Builder  as below 
```java
    /**
     * Sets the {@link RegionSet} to be used for operations using Sigv4a signing requests. This is optional; if not
     * provided, the following precedence is used:
     * <ol>
     * <li>{@link software.amazon.awssdk.core.SdkSystemSetting#AWS_SIGV4A_SIGNING_REGION_SET}.</li>
     * <li>as <code>sigv4a_signing_region_set</code> in the configuration file.</li>
     * <li>The region configured for the client.</li>
     * </ol>
     */
    default B sigv4aRegionSet(RegionSet sigv4aRegionSet) {
    }
```
### Default Builder

```java
abstract class DefaultMultiauthBaseClientBuilder<B extends MultiauthBaseClientBuilder<B, C>, C> extends
        AwsDefaultClientBuilder<B, C> {
// ...code 
    public B sigv4aRegionSet(RegionSet sigv4aRegionSet) {
        clientConfiguration.option(AwsClientOption.AWS_SIGV4A_SIGNING_REGION_SET,
                sigv4aRegionSet == null ? Collections.emptySet() : sigv4aRegionSet.asSet());
        return thisBuilder();
    }
```
- Also added change to set ClientOption when builder is built

## Testing

- Junit for codegen
- Test case to verify precedence of sigv4aRegionSet on Builder

## License


codegen/src/main/java/software/amazon/awssdk/codegen/poet/builder/BaseClientBuilderClass.java
```diff
@@ -721,6 +726,18 @@ private MethodSpec authSchemeProviderMethod() {
                          .build();
     }
 
+    private MethodSpec sigv4aRegionSetMethod() {
+        return MethodSpec.methodBuilder("sigv4aRegionSet")
```
zoewangg, line 730: Should this just be `regionSet`? I guess we can finalize it in API surface area review

zoewangg (approved):

codegen/src/main/java/software/amazon/awssdk/codegen/poet/builder/BaseClientBuilderClass.java
joviegas: Sure , I have this discussion during Surface API review

PR #5776, opened 2025-01-07
Adding Codegen support for unsignedPayload model trait for multi-auth auth schemes

joviegas:
## Motivation and Context

 If the unsignedPayload trait is present and has a value of true, then the HTTP payload must not be signed. For sigv4 and sigv4a, the static string UNSIGNED-PAYLOAD must be used in place of the payload hash.
```json
"FooOperation":{
    "name": "FooOperation",
    "http":{
        "method": "POST",
        "requestUri": "/FooOperation"
    },
    "input": {"shape": "FooOperationRequest"},
    "auth": ["aws.auth#sigv4a", "aws.auth#sigv4", "smithy.api#httpBearerAuth"],
    "unsignedPayload": true
}

```

## Modifications

- Added unsignedPayload in operation Model
- Grouping of Operation done based on AuthTrait = AutType + unsignedPayload   () earlier it was done based on only AutType)

## Testing

- Added Junit Test

## License


codegen/src/main/java/software/amazon/awssdk/codegen/poet/auth/scheme/ModelAuthSchemeKnowledgeIndex.java
```diff
@@ -72,44 +72,58 @@ public Map<List<String>, List<AuthSchemeCodegenMetadata>> operationsToMetadata()
...
     }
 
     /**
      * Returns a map from list of operations to the list of auth-types modeled for those operations. The values are taken directly
      * from the model {@link OperationModel#getAuth()} method.
      */
-    private Map<List<String>, List<AuthType>> operationsToAuthType() {
-        Map<List<AuthType>, List<String>> authSchemesToOperations =
+    private Map<List<String>, List<AuthTrait>> operationsToAuthOptions() {
+        // Create a map from lists of AuthTrait to lists of operation names
```
L-Applin, line 86: It took me a little while to understand exactly what that method does. Maybe this comment or the javadoc could be a little more explicit, something like "First, group together in a map operations (map value) that share the same AuthTraits (map key)."

joviegas: Done

L-Applin (approved):

PR #5789, opened 2025-01-10
Adding retry count to exceptions

RanVaknin:
## Motivation and Context
Improve debugging visibility by making attempt count immediately visible in exception messages. Currently, attempt information is only available through suppressed exceptions, requiring additional code to access. This change makes attempts visible directly in the exception message, matching behavior in other AWS SDKs.

## Modifications
- Added attempt count tracking and message formatting in `SdkException`
- Modified `AwsServiceException` to include attempts in AWS error details
- Modified `RetryableStageHelper2` to track and propagate attempt counts during retries

## Testing
- Added new functional test `ExceptionAttemptMessageBehaviorTest` testing with different retry scenarios
- Modified existing tests in `AwsServiceExceptionTest` and `SdkExceptionSerializationTest` to verify attempt count behavior
- Added test cases in `SdkExceptionMessageTest` for new attempt count functionality

### Before change:
```java
// non-retryable scenario
try {
    // request to a non-existent bucket
    s3Client.listObjects(g -> g.bucket("some-non-existent-bucket"));
} catch (AwsServiceException e) {
    System.out.println(e);
}
// outputs:
// NoSuchBucketException: The specified bucket does not exist (Service: S3, Status Code: 404, Request ID: abc123, Extended Request ID: xyzdef123456)


// retryable scenario
try {
    // request that might cause a retryable error (e.g., throttling)
    dynamoDbClient.getItem(...);
} catch (AwsServiceException e) {
    System.out.println(e);
}
// outputs:
// ThrottlingException: Rate of requests exceeds the allowed throughput (Service: DynamoDB, Status Code: 429, Request ID: abc123, Extended Request ID: xyzdef123456)
// Note: attempts only visible through e.getSuppressed()
```
### After change:

```java
// non-retryable scenario
try {
    // request to a non-existent bucket
    s3Client.listObjects(g -> g.bucket("some-non-existent-bucket"));
} catch (AwsServiceException e) {
    System.out.println(e);
}
// outputs:
// NoSuchBucketException: The specified bucket does not exist (Service: S3, Status Code: 404, Request ID: abc123, Extended Request ID: xyzdef123456) (Attempts: 1)


// retryable scenario
try {
    // request that might cause a retryable error (e.g., throttling)
    dynamoDbClient.getItem(...);
} catch (AwsServiceException e) {
    System.out.println(e);
}
// outputs:
// ThrottlingException: Rate of requests exceeds the allowed throughput (Service: DynamoDB, Status Code: 429, Request ID: abc123, Extended Request ID: xyzdef123456) (Attempts: 4)
// Note: attempts now visible directly in the exception message
```

## Potential Impact:

Test or error handling code that relies on exact exception message matching might break as messages now include attempt count. We recommend using partial message matching for more resilient tests and error handling.

```java
// may break
assertEquals("Resource not found", e.getMessage());
if (e.getMessage().equals("Resource not found")) { ... }

// more resilient approach
assertTrue(e.getMessage().contains("Resource not found"));
```

```java
try {
    client.someOperation();
} catch (AwsServiceException e) {
	// may break
    if (e.getMessage().equals("Resource not found")) {
        handleNotFound();
    }
	// more resilient
    if (e.getMessage().contains("Resource not found")) {
        handleNotFound();
    }
}
```


core/aws-core/src/main/java/software/amazon/awssdk/awscore/exception/AwsServiceException.java
```diff
@@ -68,17 +69,22 @@ public String getMessage() {
...
             }
-            String message = super.getMessage();
             if (message == null) {
                 message = awsErrorDetails().errorMessage();
             }
             if (message == null) {
                 return details.toString();
             }
-            return message + " " + details;
+            StringBuilder formattedMessage = new StringBuilder(message);
```
joviegas, line 78: Can we add a comment stating that appending the attempt information after the details is done to preserve the order, ensuring that AWS error details are printed after the AWS server exception message?

test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/retry/ExceptionAttemptMessageBehaviorTest.java
```diff
@@ -0,0 +1,166 @@
...
+
+        ClientT client = clientBuilder()
+            .overrideConfiguration(o -> o.retryStrategy(
+                b -> b.retryOnException(AwsServiceException.class)))
+            .build();
+
+        AwsServiceException exception = assertThrows(AwsServiceException.class,
+                                                     () -> callAllTypes(client));
+
+        assertThat(exception.getMessage()).contains("(Attempts: 1");
```
joviegas, line 103: Can we `"(Attempts: 1)"` since this test case will also pass if we out put as "(Attempts: 10" or "(Attempts: 1__SPACE_" etc

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/utils/RetryableStageHelper2.java
```diff
@@ -162,8 +162,10 @@ public SdkException retryPolicyDisallowedRetryException() {
                                   .message("Request attempt " + (i + 1) + " failure: " + exceptionMessageHistory.get(i))
                                   .writableStackTrace(false)
                                   .build();
+
             lastException.addSuppressed(pastException);
         }
+        lastException.setAttempts(retriesAttemptedSoFar() + 1);
```
zoewangg, line 168: We should probably use `toBuilder().numAttempts(x).build()`

core/sdk-core/src/test/java/software/amazon/awssdk/core/exception/SdkExceptionMessageTest.java
```diff
@@ -47,4 +47,27 @@ public void noMessage_causeWithoutMessage_implies_noMessage() {
...
+
+    @Test
+    public void defaultAttemptCount_shouldReturnOne() {
+        assertThat(SdkException.builder().build().getAttempts()).isEqualTo(0);
+    }
+
+    @Test
+    public void getAttempts_WithExplicitAttemptCount_ReturnsOneBased() {
+        assertThat(SdkException.builder().message("foo").attemptCount(2).build().getAttempts()).isEqualTo(2);
+    }
```
zoewangg, line 59: Consider using ParameterizedTest if we are verifying the same thing with different values.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/utils/RetryableStageHelper2.java
RanVaknin: I wanted to do that initially, but toBuilder will cause the re-consturction of the exception message and will result in a duplication of the attempt count (Attempt: N) (Attempt N).

Discussed this with @millems and @dagnir  and since we are already mutating the exception that mutating is reasonable.

core/sdk-core/src/main/java/software/amazon/awssdk/core/exception/SdkException.java
```diff
@@ -28,9 +28,23 @@
...
+
+    public int getAttempts() {
+        return attempts;
+    }
+
+    public void setAttempts(int attempts) {
+        this.attempts = attempts;
+    }
+    
+    public String getRawMessage() {
```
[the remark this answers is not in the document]
RanVaknin: Yes, this method is used in `AwsServiceException` as a template method maintain the order of the exception message in the desired format of `<message> (<metadata>) (<attempts>)`

Because its defined in `sdk-core` but used in `aws-core` it has to be made public. 

It was introduced to avoid this:
```
Expected :"errorMessage (Service: serviceName, Status Code: 500, Request ID: requestId) (Attempts: 6)"
Actual   :"errorMessage (Attempts: 6) (Service: serviceName, Status Code: 500, Request ID: requestId)"
```

test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/retry/ExceptionAttemptMessageBehaviorTest.java
RanVaknin: good catch! Ill fix that.

RanVaknin:
New PR: https://github.com/aws/aws-sdk-java-v2/pull/5834

PR #5794, opened 2025-01-14
ALPN H2 Support for Netty Client

davidh44:
## Motivation and Context

Adds ALPN H2 support to Netty client. ALPN is required when using an ALB in order to negotiate H2 connection.
This can be enabled when using Netty client with h2 protocol.

ALPN support was [backported to Java 8](https://bugs.openjdk.org/browse/JDK-8242894) , only works with `1.8.0_251` and newer

## Modifications

Introduces new `ProtocolNegotiation` enum class: 
`ASSUME_PROTOCOL` - Default existing setting, client uses prior knowledge
`ALPN`- Uses ALPN, without fallback. Only works with h2, throws error if set with `Protocol.HTTP1_1`

## Testing

Added unit and integ tests

## Types of changes


.changes/next-release/feature-AWSSDKforJavav2-abb9c7e.json
```diff
@@ -0,0 +1,6 @@
+{
+    "type": "feature",
+    "category": "AWS SDK for Java v2",
```
zoewangg, line 3: nit: `Netty NIO HTTP Client`

codegen/src/main/java/software/amazon/awssdk/codegen/model/config/customization/CustomizationConfig.java
```diff
@@ -258,6 +258,11 @@ public class CustomizationConfig {
      */
     private boolean generateEndpointClientTests;
 
+    /**
+     * Whether to skip using ALPN for H2 service.
+     */
+    private boolean skipAlpn;
```
zoewangg, line 264: Maybe `usePriorKnowledgeForH2`?

http-clients/aws-crt-client/src/main/java/software/amazon/awssdk/http/crt/AwsCrtHttpClientBase.java
```diff
@@ -73,6 +75,11 @@ abstract class AwsCrtHttpClientBase implements SdkAutoCloseable {
                                                + "NettyNioAsyncHttpClient instead.");
         }
 
+        if (config.get(PROTOCOL_NEGOTIATION) == ProtocolNegotiation.ALPN) {
```
zoewangg, line 78: Do we need to add validation here? CRT will only support ALPN for h2 (once we add h2 support)

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/NettyNioAsyncHttpClient.java
```diff
@@ -371,6 +407,20 @@ public interface Builder extends SdkAsyncHttpClient.Builder<NettyNioAsyncHttpCli
          */
         Builder protocol(Protocol protocol);
 
+        /**
+         * If set to {@code ProtocolNegotiation.ALPN}, the request will be made using ALPN, without a fallback protocol.
+         * If ALPN is not supported by the server an exception will be thrown.
+         * <p>
+         * Default value is true for services with H2 protocol setting, with the exception of the following services:
```
zoewangg, line 414: `Default value is true` -> `Default value is {@link ProtocolNegotiation.ALPN}`

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ChannelPipelineInitializer.java
```diff
@@ -107,30 +114,26 @@ public void channelCreated(Channel ch) {
...
+        configurePostProtocolHandlers(pipeline, protocol);
+    }
 
-        // Only add it for h1 channel because it does not apply to
-        // h2 connection channel. It will be attached
-        // to stream channels when they are created.
-        if (protocol == Protocol.HTTP1_1) {
-            pipeline.addLast(UnusedChannelExceptionHandler.getInstance());
+    private void configureProtocolHandlers(Channel ch, ChannelPipeline pipeline, Protocol protocol) {
+        if (protocolNegotiation == ProtocolNegotiation.ASSUME_PROTOCOL) {
```
zoewangg, line 122: Suggesting using switch statement

```
switch (protocolNegotiation):
   ASSUME_PROTOCOL:
         configureAssumeProtocol(xx);
         break;
  ALPN:
         configureALPN(xxx);
         break;
   default:
        throw xxx
 
```

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ChannelPipelineInitializer.java
```diff
@@ -107,30 +114,26 @@ public void channelCreated(Channel ch) {
...
+            if (protocol == Protocol.HTTP1_1) {
+                configureHttp11(ch, pipeline);
+            } else if (protocol == Protocol.HTTP2) {
+                configureHttp2(ch, pipeline);
+            }
+        } else if (protocolNegotiation == ProtocolNegotiation.ALPN) {
+            if (protocol == Protocol.HTTP1_1) {
+                // TODO - verify functionality once we implement ALPN with H1
+                // we will never reach here since we throw error in NettyNioAsyncHttpClient for ALPN + H1
+                configureAlpnH1(pipeline);
```
zoewangg, line 132: Maybe just throw exception here?

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ChannelPipelineInitializer.java
```diff
@@ -167,11 +170,52 @@ private void configureHttp11(Channel ch, ChannelPipeline pipeline) {
         ch.attr(PROTOCOL_FUTURE).get().complete(Protocol.HTTP1_1);
     }
 
+    private void configureAlpnH2(ChannelPipeline pipeline) {
+        pipeline.addLast(new ApplicationProtocolNegotiationHandler("") {
+            @Override
+            protected void configurePipeline(ChannelHandlerContext ctx, String protocol) {
+                if (ApplicationProtocolNames.HTTP_2.equals(protocol)) {
```
zoewangg, line 177: Should we handle the case where the returned protocol is not h2? 
Can you point to me the logic where we handle the case where alpn is not supported on the server side?

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/SslContextProvider.java
```diff
@@ -37,30 +40,46 @@
...
+                                                         .sslProvider(sslProvider)
+                                                         .ciphers(getCiphers(), SupportedCipherSuiteFilter.INSTANCE)
+                                                         .trustManager(trustManagerFactory)
+                                                         .keyManager(keyManagerFactory);
+
+            if (protocolNegotiation == ProtocolNegotiation.ALPN) {
+                builder.applicationProtocolConfig(
+                    new ApplicationProtocolConfig(ApplicationProtocolConfig.Protocol.ALPN,
+                                                  ApplicationProtocolConfig.SelectorFailureBehavior.NO_ADVERTISE,
+                                                  ApplicationProtocolConfig.SelectedListenerFailureBehavior.ACCEPT,
```
zoewangg, line 69: Question: how is `SelectorFailureBehavior` different from `SelectedListenerFailureBehavior`? 
Should we use `FATAL_ALERT`?

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/utils/NettyUtils.java
```diff
@@ -388,4 +388,35 @@ public static void runAndLogError(NettyClientLogger log, String errorMsg, Functi
             log.error(null, () -> errorMsg, e);
         }
     }
+
+    //  ALPN supported backported in u251
+    //  https://bugs.openjdk.org/browse/JDK-8242894
+    public static boolean isAlpnSupported() {
```
zoewangg, line 394: What if customers are using openssl, then Java version is no longer a limitation, right?

Instead of checking java version (that seems error-prone), should we check if ALPN related method is there?

http-client-spi/src/main/java/software/amazon/awssdk/http/ProtocolNegotiation.java
```diff
@@ -0,0 +1,35 @@
...
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.http;
+
+import software.amazon.awssdk.annotations.SdkPublicApi;
+
+/**
+ * The protocol negotiation used by the HTTP client to establish connections
```
dagnir, line 21: Maybe "The protocol negotiation selection scheme..."?

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/NettyNioAsyncHttpClient.java
```diff
@@ -162,6 +168,36 @@ private SslProvider resolveSslProvider(DefaultBuilder builder) {
...
+        }
+
+        ProtocolNegotiation protocolNegotiation = serviceDefaultsMap.get(SdkHttpConfigurationOption.PROTOCOL_NEGOTIATION);
+        if (protocolNegotiation == ProtocolNegotiation.ALPN) {
+            if (!isAlpnSupported()) {
+                // fallback to prior knowledge if not supported and SDK defaults to ALPN
+                protocolNegotiation = ProtocolNegotiation.ASSUME_PROTOCOL;
+                log.warn(null, () -> "ALPN is not supported in the current Java version, falling back to prior knowledge for "
+                                     + "protocol negotiation");
+            }
```
dagnir, line 195: I think we should throw if the ALPN is set on the HTTP client builder and the JDK doesn't support it.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/NettyNioAsyncHttpClient.java
```diff
@@ -162,6 +168,36 @@ private SslProvider resolveSslProvider(DefaultBuilder builder) {
         return SslContext.defaultClientProvider();
     }
 
+    private ProtocolNegotiation resolveProtocolNegotiation(ProtocolNegotiation userSetValue, AttributeMap serviceDefaultsMap,
+                                                           Protocol protocol) {
+        if (userSetValue == ProtocolNegotiation.ALPN) {
+            // TODO - remove once we implement support for ALPN with HTTP1
+            if (protocol == Protocol.HTTP1_1) {
+                throw new UnsupportedOperationException("ALPN with HTTP1 is not yet supported, use prior knowledge instead with "
```
dagnir, line 176: "HTTP/1.1"

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/NettyNioAsyncHttpClient.java
davidh44: This scenario is future new H2 service, async client, with old Java 8 version.So we want to fail by default and force user to upgrade Java version or explicitly set ASSUME_PROTOCOL on the Netty client?

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ChannelPipelineInitializer.java
davidh44: Updated to throw UnsupportedOperationException

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/SslContextProvider.java
davidh44: Seems like `SelectorFailureBehavior` comes first, is the peer in the connection. `SelectedListenerFailureBehavior` is for a configured listener. Updating to `FATAL_ALERT` for both

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/utils/NettyUtils.java
davidh44: Updated to check for `ALPNExtension`, copied from Netty

https://netty.io/4.0/xref/io/netty/handler/ssl/JettyAlpnSslEngine.html#44

http-clients/aws-crt-client/src/main/java/software/amazon/awssdk/http/crt/AwsCrtHttpClientBase.java
davidh44: Removed validation for CRT

codegen/src/main/java/software/amazon/awssdk/codegen/model/config/customization/CustomizationConfig.java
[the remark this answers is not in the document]
davidh44: updated to `usePriorKnowledgeForH2` as suggested by Zoe

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/NettyNioAsyncHttpClient.java
dagnir: Oh I see, I thought this was an explicit setting on the builder, set by the customer. I think this is fine then. If the service doesn't support prior knowledge at least we log a warning.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ChannelPipelineInitializer.java
```diff
@@ -167,11 +183,43 @@ private void configureHttp11(Channel ch, ChannelPipeline pipeline) {
...
 
+    private void configureAlpnH2(ChannelPipeline pipeline) {
+        pipeline.addLast(new ApplicationProtocolNegotiationHandler("") {
+            @Override
+            protected void configurePipeline(ChannelHandlerContext ctx, String protocol) {
+                if (protocol.equals(ApplicationProtocolNames.HTTP_2)) {
+                    configureHttp2(ctx.channel(), ctx.pipeline());
+                } else {
+                    throw new UnsupportedOperationException("The server does not support ALPN with H2");
+                }
```
dagnir, line 194: We should
 - call `ctx.fireExceptionCaught()` here to propagate the exception down the pipeline
 - call `ctx.close()` to close the channel

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/utils/NettyUtils.java
davidh44: Nvm, this doesn't work, always throws error. Its not clear what classes/methods we need to check, the [old deprecated Netty class](https://github.com/netty/netty/blob/56ef371676ed48262a46ce827d5462cd6f218a6d/handler/src/main/java/io/netty/handler/ssl/JdkAlpnApplicationProtocolNegotiator.java#L29) checked from a handful of sources, the [new class](https://github.com/netty/netty/blob/56ef371676ed48262a46ce827d5462cd6f218a6d/handler/src/main/java/io/netty/handler/ssl/ApplicationProtocolConfig.java) doesn't have checks. Reverting to checking Java version for now, will update

zoewangg: This is the check implemented in in Netty 5.x fwiw https://github.com/netty/netty/blob/netty-5.0.0.Alpha5/handler/src/main/java/io/netty5/handler/ssl/JdkAlpnSslUtils.java#L108

Any updates on non JDK Ssl provider? Seems we can skip the check if SslProvider is not JDK
> What if customers are using openssl, then Java version is no longer a limitation, right?

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ChannelPipelineInitializer.java
```diff
@@ -107,30 +114,39 @@ public void channelCreated(Channel ch) {
...
-            configureHttp11(ch, pipeline);
-        }
+        configureProtocolHandlers(ch, pipeline, protocol);
+        configurePostProtocolHandlers(pipeline, protocol);
+    }
 
-        if (configuration.reapIdleConnections()) {
-            pipeline.addLast(new IdleConnectionReaperHandler(configuration.idleTimeoutMillis()));
+    private void configureProtocolHandlers(Channel ch, ChannelPipeline pipeline, Protocol protocol) {
+        switch (protocolNegotiation) {
```
zoewangg, line 122: Can we add functional tests for those different combinations using a local mock server?

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/utils/NettyUtils.java
dagnir: +1, I was going to suggest a similar check to This is the check implemented in in Netty 5.x fwiw https://github.com/netty/netty/blob/netty-5.0.0.Alpha5/handler/src/main/java/io/netty5/handler/ssl/JdkAlpnSslUtils.java#L108

davidh44: Updated check to return true ifSslProvider is OpenSsl

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ChannelPipelineInitializer.java
```diff
@@ -167,11 +183,43 @@ private void configureHttp11(Channel ch, ChannelPipeline pipeline) {
         ch.attr(PROTOCOL_FUTURE).get().complete(Protocol.HTTP1_1);
     }
 
+    private void configureAlpnH2(ChannelPipeline pipeline) {
+        pipeline.addLast(new ApplicationProtocolNegotiationHandler("") {
+            @Override
+            protected void configurePipeline(ChannelHandlerContext ctx, String protocol) {
```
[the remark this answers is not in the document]
davidh44: no, only a single protocol will be returned

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ChannelPipelineInitializer.java
davidh44: Added NettyClientAlpnTest using MockH2Server from benchmarks

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/SslContextProvider.java
davidh44: Turns out OpenSsl Provider does not support `FATAL_ALERT`
https://github.com/netty/netty/blob/4.1/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslContext.java#L767

Reverting back to original values if OpenSsl provider used, added tests

dagnir (approved):

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/utils/NettyUtils.java
```diff
@@ -388,4 +395,45 @@ public static void runAndLogError(NettyClientLogger log, String errorMsg, Functi
...
+            MethodHandles.Lookup lookup = MethodHandles.lookup();
+
+            MethodHandle getApplicationProtocol = AccessController.doPrivileged(
+                (PrivilegedExceptionAction<MethodHandle>) () ->
+                lookup.findVirtual(SSLEngine.class, "getApplicationProtocol", MethodType.methodType(String.class)));
+
+            getApplicationProtocol.invoke(engine);
+            return true;
+        } catch (NoSuchMethodException e) {
+            log.info(() -> "ALPN not supported: SSLEngine.getApplicationProtocol() method not found.");
```
zoewangg, line 419: Is info the right level? Should it be debug?

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/utils/NettyUtils.java
```diff
@@ -388,4 +395,45 @@ public static void runAndLogError(NettyClientLogger log, String errorMsg, Functi
...
+            return false;
+        } catch (Throwable t) {
+            log.info(() -> "ALPN support check failed: " + t.getMessage());
```
zoewangg, line 422: Is info the right level? Should it be debug?

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/utils/NettyUtils.java
```diff
@@ -388,4 +395,45 @@ public static void runAndLogError(NettyClientLogger log, String errorMsg, Functi
             log.error(null, () -> errorMsg, e);
         }
     }
+
+    //  ALPN supported backported in u251
+    //  https://bugs.openjdk.org/browse/JDK-8242894
+    public static boolean isAlpnSupported(SslProvider sslProvider) {
```
zoewangg, line 401: Can we just do the check once (per JVM) and lazily?

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/utils/NettyUtils.java
```diff
@@ -388,4 +395,45 @@ public static void runAndLogError(NettyClientLogger log, String errorMsg, Functi
...
+        if (sslProvider != SslProvider.JDK) {
+            return true;
+        }
+
+        try {
+            SSLContext context = SSLContext.getInstance("TLS");
```
zoewangg, line 407: Just to double check, did we test this behavior (manually) with older java version?

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/NettyNioAsyncHttpClient.java
```diff
@@ -371,6 +408,20 @@ public interface Builder extends SdkAsyncHttpClient.Builder<NettyNioAsyncHttpCli
...
+        /**
+         * If set to {@code ProtocolNegotiation.ALPN}, the request will be made using ALPN, without a fallback protocol.
+         * If ALPN is not supported by the server an exception will be thrown.
+         * <p>
+         * Default value is {@code ProtocolNegotiation.ALPN} for services with H2 protocol setting, with the exception of the
+         * following services: Kinesis, TranscribeStreaming, Lex Runtime v2, Q Business
+         * <p>
+         * Note: For Java 8, ALPN is only supported in versions 1.8.0_251 and newer.
+         * If on unsupported Java version:
+         *  1) Default SDK setting of true -> SDK will fallback to prior knowledge and not use ALPN
```
zoewangg, line 420: nit: `Default SDK setting of true` and `User explicitly sets value of true` probably need to be updated

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ChannelPipelineInitializer.java
```diff
@@ -107,30 +114,39 @@ public void channelCreated(Channel ch) {
...
+    private void configureProtocolHandlers(Channel ch, ChannelPipeline pipeline, Protocol protocol) {
+        switch (protocolNegotiation) {
+            case ASSUME_PROTOCOL:
+                configureAssumeProtocol(ch, pipeline, protocol);
+                break;
+            case ALPN:
+                configureAlpn(pipeline, protocol);
+                break;
+            default:
+                throw new UnsupportedOperationException("");
```
zoewangg, line 130: Can we add the enum to the error message?

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ChannelPipelineInitializer.java
```diff
@@ -167,11 +183,45 @@ private void configureHttp11(Channel ch, ChannelPipeline pipeline) {
...
+    private void configureAlpnH2(ChannelPipeline pipeline) {
+        pipeline.addLast(new ApplicationProtocolNegotiationHandler("") {
+            @Override
+            protected void configurePipeline(ChannelHandlerContext ctx, String protocol) {
+                if (protocol.equals(ApplicationProtocolNames.HTTP_2)) {
+                    configureHttp2(ctx.channel(), ctx.pipeline());
+                } else {
+                    ctx.channel().attr(PROTOCOL_FUTURE).get()
+                       .completeExceptionally(new UnsupportedOperationException("The server does not support ALPN with H2"));
+                    ctx.close();
```
zoewangg, line 195: Should we invoke `ctx.fireExceptionCaught(new UnsupportedOperationException("The server does not support ALPN with H2")`?

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/NettyNioAsyncHttpClient.java
```diff
@@ -86,6 +89,9 @@ public final class NettyNioAsyncHttpClient implements SdkAsyncHttpClient {
     private NettyNioAsyncHttpClient(DefaultBuilder builder, AttributeMap serviceDefaultsMap) {
         this.configuration = new NettyConfiguration(serviceDefaultsMap);
         Protocol protocol = serviceDefaultsMap.get(SdkHttpConfigurationOption.PROTOCOL);
+        SslProvider sslProvider = resolveSslProvider(builder);
+        ProtocolNegotiation protocolNegotiation = resolveProtocolNegotiation(builder.protocolNegotiation, serviceDefaultsMap,
```
zoewangg, line 93: Should we add a validation to throw exception if `http` is used with ALPN enabled?

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/utils/NettyUtils.java
davidh44: Tested on Jdoodle with JDK `1.8.0_66`, fails as intended and succeeds with Java 9 and above. Updated to catch  `PrivilegedActionException` instead of `NoSuchMethodException`

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/NettyNioAsyncHttpClient.java
davidh44: added check during `execute()` and added unit test

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/NettyNioAsyncHttpClient.java
```diff
@@ -117,19 +118,29 @@ private NettyNioAsyncHttpClient(DefaultBuilder builder, AttributeMap serviceDefa
...
+        failIfAlpnUsedWithHttp(request);
         RequestContext ctx = createRequestContext(request);
         ctx.metricCollector().reportMetric(HTTP_CLIENT_NAME, clientName()); // TODO: Can't this be done in core?
         return new NettyRequestExecutor(ctx).execute();
     }
 
+    private void failIfAlpnUsedWithHttp(AsyncExecuteRequest request) {
+        if (protocolNegotiation == ProtocolNegotiation.ALPN
+            && "http".equals(request.request().protocol())) {
+            throw new UnsupportedOperationException("ALPN can only be used with HTTPS, not HTTP.");
```
zoewangg, line 140: Should we complete the future exceptionally instead? What happens if we throw exception here? Does SDK core handle it correctly?

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/utils/NettyUtils.java
```diff
@@ -403,6 +406,20 @@ public static boolean isAlpnSupported(SslProvider sslProvider) {
...
 
+        Boolean supported = alpnSupported;
+        if (supported != null) {
+            return supported;
+        }
+
+        synchronized (NettyUtils.class) {
+            if (alpnSupported == null) {
+                alpnSupported = checkAlpnSupport();
+            }
```
zoewangg, line 417: We need to double check the value here. Can we use Lazy directly, which has built-in lazy initialization support?

https://github.com/aws/aws-sdk-java-v2/blob/master/utils/src/main/java/software/amazon/awssdk/utils/Lazy.java

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ChannelPipelineInitializer.java
zoewangg: I see, I guess this is fine then.

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/NettyNioAsyncHttpClient.java
davidh44: Yes, SDK core handles throwing exception correctly and fails right away. Returned a future completed exceptionally will cause the request to hang

PR #5812, opened 2025-01-20
Fix intermittent ConcurrentModificationException thrown from codegen 

zoewangg:
## Motivation and Context
Fix intermittent `ConcurrentModificationException` thrown from codegen

```
Failed to execute goal software.amazon.awssdk:codegen-maven-plugin:2.30.3-SNAPSHOT:generate (default) on project codegen-generated-classes-test: Execution default of goal software.amazon.awssdk:codegen-maven-plugin:2.30.3-SNAPSHOT:generate failed: Failed to generate code. Exception message : java.lang.RuntimeException: java.util.ConcurrentModificationException -> [Help 1]
```

The root cause is that we are sorting an ArrayList while iterating.

## Modifications
Updated the code to not sort the list while iterating

### DIFF
--- a/codegen/src/main/java/software/amazon/awssdk/codegen/CodeGenerator.java
+++ b/codegen/src/main/java/software/amazon/awssdk/codegen/CodeGenerator.java
@@ -26,9 +26,10 @@ import software.amazon.awssdk.codegen.emitters.tasks.AwsGeneratorTasks;
 import software.amazon.awssdk.codegen.internal.Jackson;
 import software.amazon.awssdk.codegen.internal.Utils;
 import software.amazon.awssdk.codegen.model.intermediate.IntermediateModel;
+import software.amazon.awssdk.utils.Logger;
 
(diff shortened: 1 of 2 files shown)


L-Applin (approved):
This works, but just pointing out that we probably could as well return a new list here in the **[HttpChecksum class](https://github.com/aws/aws-sdk-java-v2/blob/master/codegen/src/main/java/software/amazon/awssdk/codegen/checksum/HttpChecksum.java#L59-L61)**

zoewangg:
>This works, but just pointing out that we probably could as well return a new list here in the [HttpChecksum class](https://github.com/aws/aws-sdk-java-v2/blob/master/codegen/src/main/java/software/amazon/awssdk/codegen/checksum/HttpChecksum.java#L59-L61)

The root cause seems to be the use of collections.sort on an ArrayList, so unless we return a concurrently modifiable list, it may have the same issue.

PR #5813, opened 2025-01-21
Add logic to handle number acceptors

RanVaknin:
## Motivation and Context
Currently the Java SDK waiter matcher acceptor only support Integers. According to the cross SDK spec waiter matcher acceptor must support all number types. 

## Modifications

- Changed  jmspathruntime numbers internally to use BigDecimal for all number types in order to support arbitrary precision.
- Updated Jackson Jr version to 2.17.3 to support Big Decimal.

## Testing

- Added WaitersPathMatcherTest for numeric matchers
- Added coverage to existing numeric handling tests


dagnir (commented):
Can we add tests in `codegen` for generated waiters that use numbers?

codegen/src/main/java/software/amazon/awssdk/codegen/poet/waiters/BaseWaiterClassSpec.java
```diff
@@ -586,18 +602,31 @@ private CodeBlock pathAllAcceptorBody(Acceptor acceptor) {
...
+        } else if (isString) {
+            builder.add("$S", expected);
+        } else {
+            builder.add("$L", expected);
+        }
+
+        builder.add("));")
+               .add("}");
+
+        return builder.build();
```
dagnir, line 629: Can we make the same change as in `pathAcceptorBody` to make it more readable?

test/codegen-generated-classes-test/pom.xml
```diff
@@ -279,6 +279,12 @@
             <artifactId>mockito-junit-jupiter</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>backport-util-concurrent</groupId>
+            <artifactId>backport-util-concurrent</artifactId>
+            <version>3.1</version>
+            <scope>test</scope>
+        </dependency>
```
dagnir, line 287: What's this dependency for?

dagnir:
Can we merge in master and run integ tests again?

dagnir (approved):

PR #5814, opened 2025-01-21
Adds business metrics for credential providers

cenedhryn:
## Motivation and Context
Tracking how credentials were resolved

## Modifications
- Adds a number of new business metrics feature IDs representing credential sources
- Changes existing providerNames to the new metrics values
- Adds more combinations and sources of credential providers
- Introduces a new public API param `source` on some builders for credential providers. It's the only non-trivial way to link more than one metrics value together

Deviations from specs:
- Not using SSO legacy `u` because it doesn't make sense - It's already tracked as legacy via the`CREDENTIALS_PROFILE_SSO_LEGACY` `t` value. Once in the SSO provider legacy doesn't really matter. The regular `CREDENTIALS_PROFILE_SSO` value of `s` can be used for both cases. You can't really set a legacy client without coming from the profile
- Not tracking the exact parent provider for source profile and/or named provider / credential_source use cases. The method (source profile or named provider) is tracked in both the source provider call and the final credential provider call, but the final credential provider call will not emit exactly what the source provider was. For instance, if named provider with IMDS is used, we should track `p,0` for the imds call and `p,0,i` for the credential metrics: named provider followed by imds followed by assume role. Instead, we'll have `p,0` and `p,i` for the imds call and the assume roll call respectively. 
- Not tracking named provider where the source provider is env vars provider. It's an unusual case and doesn't warrant rewriting the provider. 

## Types of changes

## Checklist

## License


test/auth-tests/src/it/java/software/amazon/awssdk/auth/source/UserAgentProviderTest.java
```diff
@@ -74,8 +74,8 @@ void userAgentString_containsCredentialProviderNames_IfPresent(IdentityProvider<
 
     private static Stream<Arguments> credentialProviders() {
         return Stream.of(
-            Arguments.of(StaticCredentialsProvider.create(SESSION_IDENTITY), "stat"),
-            Arguments.of(StaticCredentialsProvider.create(BASIC_IDENTITY), "stat")
+            Arguments.of(StaticCredentialsProvider.create(SESSION_IDENTITY), "m/D"),
+            Arguments.of(StaticCredentialsProvider.create(BASIC_IDENTITY), "m/D")
```
dagnir, line 78: Probably missed something, but why are we checking for `m/D` here? (`D` being `RETRY_MODE_LEGACY`).

core/auth/src/test/java/software/amazon/awssdk/auth/credentials/SystemSettingsCredentialsProviderTest.java
dagnir, line 1: Why did we remove this test?

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/ChildProfileCredentialsProviderFactory.java
```diff
@@ -38,7 +38,9 @@ public interface ChildProfileCredentialsProviderFactory {
      * provider. This credentials provider should be closed when it is no longer used.
      * @param profile The profile that should be used to load the configuration necessary to create the child credentials
      * provider.
+     * @param source A string list of {@link software.amazon.awssdk.core.useragent.BusinessMetricFeatureId} denoting
+     * previous credentials providers that are chained with this one.
      * @return The credentials provider with permissions derived from the source credentials provider and profile.
      */
-    AwsCredentialsProvider create(AwsCredentialsProvider sourceCredentialsProvider, Profile profile);
-}
+    AwsCredentialsProvider create(AwsCredentialsProvider sourceCredentialsProvider, Profile profile, String source);
```
dagnir, line 45: The javadoc implies this this might better as a `List<String>`, any reason it isn't?

dagnir:
> Not tracking the exact parent provider for source profile and/or named provider / credential_source use cases. The method (source profile or named provider) is tracked in both the source provider call and the final credential provider call, but the final credential provider call will not emit exactly what the source provider was. For instance, if named provider with IMDS is used, we should track p,0 for the imds call and p,0,i for the credential metrics: named provider followed by imds followed by assume role. Instead, we'll have p,0 and p,i for the imds call and the assume roll call respectively.

Can you elaborate further on why getting `p,0,1` is not possible or difficult? I don't think I quite understood from this explanation.

cenedhryn:
> > Not tracking the exact parent provider for source profile and/or named provider / credential_source use cases. The method (source profile or named provider) is tracked in both the source provider call and the final credential provider call, but the final credential provider call will not emit exactly what the source provider was. For instance, if named provider with IMDS is used, we should track p,0 for the imds call and p,0,i for the credential metrics: named provider followed by imds followed by assume role. Instead, we'll have p,0 and p,i for the imds call and the assume roll call respectively.
> 
> Can you elaborate further on why getting `p,0,1` is not possible or difficult? I don't think I quite understood from this explanation.

When I look at the code again, I think it will work. In this example, ideally the user agent should contain p,0,i for the main call. First, the profile credentials provider is called and in ProfileCredentialsUtils, the use case for named provider is selected based on the settings in `roleAndCredentialSourceBasedProfileCredentialsProvider`. That's where the `p` is recorded. Then, the code selects the base credentials provider. This is where I went astray before, but retrieving the source from the base credentials provider, `0` should be doable, and then concatenate with the `p`, and submitting `p,0` to STS. 

Overall the question of whether composite sources should be concatenated String or List<String> - I'm still leaning towards concatenated String. 

There is a fair amount of refactoring that needs to be done because toString cannot and should not be used to retrieve the business metric. I have to roll back some weird code.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/ChildProfileCredentialsProviderFactory.java
cenedhryn: See the general comment. I think the question of List or not List will become clearer if I go through the APIs for source again.

core/auth/src/test/java/software/amazon/awssdk/auth/credentials/SystemSettingsCredentialsProviderTest.java
cenedhryn: Because there is also a `SystemSettingCredentialsProvidersTest` that is more extensive 🙄

test/auth-tests/src/it/java/software/amazon/awssdk/auth/source/UserAgentProviderTest.java
cenedhryn: I think there is a pretty large bug in how the user agent is constructed, putting credential provider business metrics in auth source but not in `m`.

bhoradc:
Closing, since we added the business metrics tracking for credential providers in https://github.com/aws/aws-sdk-java-v2/pull/6457

PR #5823, opened 2025-01-24
allow single "=" as query

Fred1155:
## Motivation and Context

There is a issue on SdkHttpFullRequest throws exception when constructing with query string is a single "=".  The "=" query is valid according to RFC spec. We should treat it as a valid query string. This is because when spliting query strings, "=" will be splited to an empty list, resulting in array out of bound.
Fixed the bug "ArrayIndexOutOfBoundsException: Index 0 out of bounds for length 0" when "=" as the query string
## Modifications

1. filter out the length 0 list when splitQueryString
## Types of changes

## Checklist

## License


joviegas:
- Can we please mention some details about this bug fix , how is SDK user impacted by this or in what circumstances will this issue be seen ?
- How did we discover this bug ?
- Do we need change log ?

utils/src/test/java/software/amazon/awssdk/utils/SdkHttpUtilsTest.java
```diff
@@ -238,6 +238,13 @@ public void uriParams() throws URISyntaxException {
                                        entry("decoded&Part", Arrays.asList("equals=val")));
     }
 
+    @Test
+    public void uriParamsWithSingleEqualQuery() throws URISyntaxException {
+        URI uri = URI.create("http://example.com?=");
+        Map<String, List<String>> uriParams = SdkHttpUtils.uriParams(uri);
+        assertThat(uriParams).isEmpty();
```
joviegas, line 245: Consider
```
    @Test
    public void uriParamsWithEmptyKeyAndValue() throws URISyntaxException {     //Interested
        URI uri = URI.create("http://example.com?=myValue");
        Map<String, List<String>> uriParams = SdkHttpUtils.uriParams(uri);
        System.out.println("uriParams " +uriParams);

    }
   ```
   
   For `http://example.com?=myValue`  the uriParams looks like 
   `uriParams {=[myValue]}`
   
   If you see here the Key in the Map is empty .

Now based on this for `"http://example.com?="` we should have empty key with empty list as below
`uriParams {=[]}`
   
   Can you please check on that ?

Comment (Fred1155) [reply utils/src/test/java/software/amazon/awssdk/utils/SdkHttpUtilsTest.java:245]: Discussed offline,  {=null} conforms with current implementation

File: utils/src/test/java/software/amazon/awssdk/utils/SdkHttpUtilsTest.java [hunk 2]
```diff
@@ -238,6 +238,13 @@ public void uriParams() throws URISyntaxException {
                                        entry("decoded&Part", Arrays.asList("equals=val")));
     }
 
+    @Test
```
joviegas, line 241:
```suggestion
    @Test
    void uriParamsWithSingleEqualQuery()  {
        URI uri = URI.create("http://example.com?=");
        Map<String, List<String>> uriParams = SdkHttpUtils.uriParams(uri);
        assertThat(uriParams).containsKey("");
        assertThat(uriParams.get(""))
            .isNotNull()
            .hasSize(1)
            .containsNull();
    }

    @Test
    void uriParamsWithSingleEqualWithValueQuery()  {
        URI uri = URI.create("http://example.com?=nokeyvalue");
        Map<String, List<String>> uriParams = SdkHttpUtils.uriParams(uri);
        assertThat(uriParams).containsKey("");
        assertThat(uriParams.get(""))
            .isNotNull()
            .hasSize(1)
            .contains("nokeyvalue");
    }
```

joviegas: Also for completeness can we add test in SdkHttpRequestResponseTest since this tests the Public API

```java
    @Test
    public void testSdkHttpRequestWithEmptyEqualQueryParameterName() {
        final String expected = "http://example.com?=";
        URI myUri = URI.create(expected);
        SdkHttpRequest actual = SdkHttpRequest.builder().method(SdkHttpMethod.POST).uri(myUri).build();
        assertThat(actual.getUri()).hasToString(expected);
    }

```

.changes/next-release/bugfix-AWSSDKforJavav2-466bf4c.json
```diff
@@ -0,0 +1,6 @@
+{
+    "type": "bugfix",
+    "category": "AWS SDK for Java v2",
+    "contributor": "",
+    "description": "Fixed the bug \"ArrayIndexOutOfBoundsException\" when single \"=\" as the query string"
```
joviegas, line 5: Can we be specific about the  fix is made in SdkHttpFullRequest 
```
Fixed an issue in SdkHttpFullRequest where constructing with a query string consisting of a single \"=\" would throw an ArrayIndexOutOfBoundsException..
```

or something that is more specific ?

utils/src/test/java/software/amazon/awssdk/utils/SdkHttpUtilsTest.java
Fred1155: This test will fail because our current implementation will eliminate "=" when there are no query values. I added a test that the expected result string is "http://example.com?"

joviegas (approved):

utils/src/main/java/software/amazon/awssdk/utils/http/SdkHttpUtils.java
```diff
@@ -405,6 +405,7 @@ public static Map<String, List<String>> uriParams(URI uri) {
         return splitQueryString(uri.getRawQuery())
             .stream()
             .map(s -> s.split("="))
+            .map(s -> s.length == 0 ? new String[] {""} : s)
             .map(s -> s.length == 1 ? new String[] { s[0], null } : s)
```
zoewangg, line 409:
```suggestion
            .map(s -> {
                if (s.length == 0) {
                    return new String[] {""};
                }
                return s.length == 1 ? new String[] {s[0], null} : s;
            })
```

Fred1155: Do you mean `return new String[] {"", null};` in line 410? cause otherwise it will fail on original line 410

zoewangg: ah I see,  yup

PR #5836, opened 2025-01-29
Fixed an issue that could cause checksum mismatch errors in S3 uploads.

millems:
This error is sometimes encountered when a customer uses:
1. The AsyncS3Client
2. A ChecksumAlgorithm of SHA1 or SHA256 (instead of the default CRC32)
3. Parallel uploads

The root cause was the SDK using thread locals to cache the SHA1 or SHA256 message digest implementations. This meant that if a single event loop thread was processing multiple requests, those two requests would use the same digest implementation to calculate the checksum.

This PR updates the SHA1 and SHA256 (and MD5, though it's not used by S3) to use a LIFO cache.


core/checksums/src/main/java/software/amazon/awssdk/checksums/internal/DigestAlgorithm.java
```diff
@@ -27,41 +31,92 @@ public enum DigestAlgorithm {
...
+         */
+        @Override
+        public void close() {
+            if (digest == closedDigest) {
+                return;
+            }
+
+            // Avoid over-caching after large traffic bursts. The maximum chosen here is arbitrary. It's also not strictly
+            // enforced, since these statements aren't synchronized.
+            if (digestCache.size() <= MAX_CACHED_DIGESTS) {
```
zoewangg, line 108: size() may be expensive, should we track the size using atomic integer instead?

core/checksums/src/main/java/software/amazon/awssdk/checksums/internal/DigestAlgorithm.java
```diff
@@ -27,41 +31,92 @@ public enum DigestAlgorithm {
...
+         * Retrieve the message digest instance.
+         */
+        public MessageDigest messageDigest() {
+            return digest.get();
+        }
 
-        DigestThreadLocal(String algorithmName) {
-            this.algorithmName = algorithmName;
+        /**
+         * Retrieve the message digest bytes. This will close the message digest when invoked. This is because the underlying
```
zoewangg, line 86: Question: where do we close messageDigest in this method?

core/checksums/src/main/java/software/amazon/awssdk/checksums/internal/DigestAlgorithm.java
jeking3: How would you coordinate the atomic size and the concurrent deque?

core/checksums/src/main/java/software/amazon/awssdk/checksums/internal/DigestAlgorithm.java
```diff
@@ -27,41 +31,92 @@ public enum DigestAlgorithm {
...
         @Override
-        protected MessageDigest initialValue() {
+        public CloseableMessageDigest clone() {
             try {
-                return MessageDigest.getInstance(algorithmName);
-            } catch (NoSuchAlgorithmException e) {
-                throw new RuntimeException("Unable to fetch message digest instance for Algorithm "
-                                           + algorithmName + ": " + e.getMessage(), e);
+                return new CloseableMessageDigest((MessageDigest) digest.get().clone());
+            } catch (CloneNotSupportedException e) { // should never occur
```
jeking3, line 118: Any time a comment says "should never occur", it seems to happen.  Why can't this method declare that it throws CloneNotSupportedException instead?

core/checksums/src/main/java/software/amazon/awssdk/checksums/internal/DigestAlgorithm.java
```diff
@@ -27,41 +31,92 @@ public enum DigestAlgorithm {
...
+         * Release this message digest back to the cache. Once released, you must not use the digest anymore.
+         */
+        @Override
+        public void close() {
+            if (digest == closedDigest) {
+                return;
+            }
+
+            // Avoid over-caching after large traffic bursts. The maximum chosen here is arbitrary. It's also not strictly
+            // enforced, since these statements aren't synchronized.
```
dagnir, line 107: Would it be possible to have a "prefilled" cache of lazy loaded digests, and then when we can't use a cached digest, we have a different instance that just closes and doesn't need to interact with the cache?

core/checksums/src/main/java/software/amazon/awssdk/checksums/internal/DigestAlgorithm.java
millems: Good catch, that needs a test added.

core/checksums/src/main/java/software/amazon/awssdk/checksums/internal/DigestAlgorithm.java
millems: That's an option. It might be trickier to decide on the size of the cache, and then it means that not releasing a message digest to the cache will have long-term performance implications if it's one of those special "cached" digests. The advantage to this implementation is that the odd error that could fail to release the digest back to the cache doesn't really hurt.

core/checksums/src/main/java/software/amazon/awssdk/checksums/internal/DigestAlgorithm.java
millems: `CloneNotSupportedException` is a checked exception. Because the SDKs don't throw checked exceptions, the callers would need to wrap it in an unchecked exception themselves.

core/checksums/src/main/java/software/amazon/awssdk/checksums/internal/DigestAlgorithm.java
millems: I'll switch to a linked blocking deque. Uses locking, but it's still likely faster than creating new checksums and has a constant-time size() method. We can do benchmarking later to verify this is the fastest method.

Is that reasonable?

core/checksums/src/main/java/software/amazon/awssdk/checksums/internal/DigestAlgorithm.java
millems: I've improved the exception message if this scenario does happen.

dagnir (approved):

core/checksums/src/main/java/software/amazon/awssdk/checksums/internal/DigestAlgorithm.java
```diff
@@ -103,20 +104,18 @@ public void close() {
                 return;
             }
 
-            // Avoid over-caching after large traffic bursts. The maximum chosen here is arbitrary. It's also not strictly
-            // enforced, since these statements aren't synchronized.
-            if (digestCache.size() <= MAX_CACHED_DIGESTS) {
-                digestCache.addFirst(digest.get());
-            }
+            // Drop this digest is the cache is full.
+            digestCache.offerFirst(digest.get());
```
dagnir, line 108: Nice

zoewangg (approved):

PR #5841, opened 2025-01-29
Buffer if custom content provider stream

dagnir:
This updates the
`RequestBody.fromContentProvider(ContentStreamProvider,long,String)` override such that the underlying implementation will buffer the contents of the stream in memory during the first pass through the stream.

This is a followup to #5837.

## Types of changes

## Checklist

## License

### DIFF
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/sync/BufferingContentStreamProvider.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/sync/BufferingContentStreamProvider.java
@@ -34,19 +34,21 @@ import software.amazon.awssdk.utils.IoUtils;
 @NotThreadSafe
 public final class BufferingContentStreamProvider implements ContentStreamProvider {
     private final ContentStreamProvider delegate;
-    private InputStream bufferedStream;
+    private final Long expectedLength;
+    private BufferStream bufferedStream;
 
     private byte[] bufferedStreamData;
     private int count;
 
-    public BufferingContentStreamProvider(ContentStreamProvider delegate) {
+    public BufferingContentStreamProvider(ContentStreamProvider delegate, Long expectedLength) {
         this.delegate = delegate;
+        this.expectedLength = expectedLength;
     }
 
     @Override
     public InputStream newStream() {
         if (bufferedStreamData != null) {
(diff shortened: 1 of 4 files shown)


core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/sync/BufferingContentStreamProvider.java
```diff
@@ -59,36 +61,54 @@ public InputStream newStream() {
...
 
         @Override
-        public synchronized int read(byte[] b, int off, int len) throws IOException {
-            int read = super.read(b, off, len);
-            if (read < 0) {
+        public void close() throws IOException {
+            if (!hasExpectedLength() || expectedLengthReached()) {
                 saveBuffer();
+                super.close();
             }
```
millems, line 95: Can we add a comment on why we're ignoring close sometimes here? (I assume, because people may be reusing RequestBody, so we don't want to break subsequent requests).

core/sdk-core/src/main/java/software/amazon/awssdk/core/sync/RequestBody.java
```diff
@@ -209,6 +209,14 @@ public static RequestBody empty() {
 
     /**
      * Creates a {@link RequestBody} from the given {@link ContentStreamProvider}.
+     * <p>
+     * Important: Be aware that is implementation requires buffering the contents for {@code ContentStreamProvider}, which can
```
millems, line 213: nit: that is -> this

core/sdk-core/src/main/java/software/amazon/awssdk/core/sync/RequestBody.java
```diff
@@ -217,17 +225,14 @@ public static RequestBody empty() {
      * @return The created {@code RequestBody}.
      */
     public static RequestBody fromContentProvider(ContentStreamProvider provider, long contentLength, String mimeType) {
```
millems, line 227: There are places we delegate to these methods (e.g. fromInputStream). Do we need those features to also buffer in memory, or should we move them away from delegating to these methods?

dagnir: Good callout. Done.

millems (approved):

PR #5854, opened 2025-02-03
Skip ALPN if request is HTTP endpoint

davidh44:
## Motivation and Context

Mock tests with h2 clients making requests to HTTP endpoints fail with [recent ALPN addition](https://github.com/aws/aws-sdk-java-v2/pull/5794)

## Modifications

Skip adding ALPN handler if request endpoint is HTTP and the default client setting is ALPN. This maintains previous behavior of prior knowledge protocol negotiation, and avoids breaking changes for existing generated clients with h2 protocol setting

Continue to throw error if request endpoint is HTTP and user configures ALPN on client

## Testing

Added unit tests

## Types of changes

### DIFF
--- a/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/NettyNioAsyncHttpClient.java
+++ b/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/NettyNioAsyncHttpClient.java
@@ -136,10 +136,9 @@ public final class NettyNioAsyncHttpClient implements SdkAsyncHttpClient {
     }
 
     private void failIfAlpnUsedWithHttp(AsyncExecuteRequest request) {
-        if ("http".equals(request.request().protocol())) {
-            if (isAlpnUserConfigured) {
-                throw new UnsupportedOperationException("ALPN can only be used with HTTPS, not HTTP.");
-            }
+        if (isAlpnUserConfigured && "http".equals(request.request().protocol())) {
+            throw new UnsupportedOperationException("ALPN can only be used with HTTPS, not HTTP. "
+                                                    + "Use ProtocolNegotiation.ASSUME_PROTOCOL instead.");
         }
     }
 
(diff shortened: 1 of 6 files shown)


http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/NettyNioAsyncHttpClient.java
```diff
@@ -86,6 +86,7 @@ public final class NettyNioAsyncHttpClient implements SdkAsyncHttpClient {
     private final SdkChannelPoolMap<URI, ? extends SdkChannelPool> pools;
     private final NettyConfiguration configuration;
     private final ProtocolNegotiation protocolNegotiation;
+    private boolean userSetAlpn;
```
joviegas, line 89: some suggestion 
```
private boolean isAlpnUserConfigured;
// or
private boolean hasUserDefinedAlpn;

```

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ChannelPipelineInitializer.java
```diff
@@ -99,7 +99,9 @@ public void channelCreated(Channel ch) {
         ch.attr(CHANNEL_DIAGNOSTICS).set(new ChannelDiagnostics(ch));
         ch.attr(PROTOCOL_FUTURE).set(new CompletableFuture<>());
         ChannelPipeline pipeline = ch.pipeline();
-        if (sslCtx != null) {
+
+        boolean sslCtxPresent = sslCtx != null;
```
joviegas, line 103: whats the reason for this change ?

davidh44: We are saving the boolean as variable since we pass to `configureProtocolHandlers()` to fallback to prior knowledge if sslCtx is null

joviegas:
Can we please  add change logs .

http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ChannelPipelineInitializer.java
davidh44: This is existing behavior, to use prior knowledge. Covered by test here, to make sure we don't throw `UnsupportedOperationException`
https://github.com/aws/aws-sdk-java-v2/blob/ed58871b0bc410944537c433a2b02346fea111a0/test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/h2/AlpnHttpTest.java#L49

joviegas (approved):

PR #5862, opened 2025-02-04
Decouple policy logic from resource url for getSignedUrlWithCustomPolicy

### LINKED ISSUE #5577: CloudFrontUtilities : cannot specify wildcard (*) resource url for a custom policy
### Describe the bug

When using the new CloudFrontUtilities class to sign resource urls, it is not possible to specify a **wildcard resource URL policy** ('*') as the `CustomSignerRequest.resourceUrl` is used for both the policy and the URL to be signed.

The `SigningUtils.buildCustomPolicyForSignedUrl(...)` method which is called from the `CloudFrontUtilities.getSignedUrlWithCustomPolicy(CustomSignerRequest request)` method does appear to cater for this by defaulting to the wildcard if the resourceUrl is `NULL` BUT it is not possible to specify a null CustomSignerRequest.resourceUrl as this is also used for the URL to be signed.

### Expected Behavior

Should be able to specify a custom policy resourceUrl as the 'wildcard' (or any other policy required) whilst also specify the resourceUrl to be signed.
- In the older AWS SDK this was possible as the custom policy was provided as a separate parameter string.

### Current Behavior

Currently cannot specify a 'wildcard' resourceUrl for a custom policy separate to the URL to be signed.

Example: 
When use a signed URL with primefaces '<p:graphicImage ... cache=false>' on a client page primefaces adds a '&pfdrid_c' paramter to the URL. Without the ability to specify a wildcard policy for the resource url then this enforces the primefaces parameter to be present in the URL when signing.

### Reproduction Steps

For example: 
As indicate above, the code does try to set the default wildcard for the custom policy if a NULL resourceUrl is specified.
However if try to do this then will get a NPE as this value is also used as the URL to be signed.

i.e. if you try this in a test then you will get a NPE exception. 

```
CustomSignerRequest.builder()
	.resourceUrl( null ) // wildcard policy
	. etc
	.build();
CloudFrontUtilities.create()
	.getSignedUrlWithCustomPolicy(customSignerRequest)
	.url()
```

So the code as it stands forces the URL to be signed to always be specified & does not enable a custom resource URL policy to be specified which is different to the URL to be signed.

### Possible Solution

- Could add a new/separate `CustomSignedResource` parameter to enable the policy resource url to be specified in addition to the URL to be signed.

OR

- Could make the `CustomSignedResource.resourceUrl`  be specific to the custom policy & then add a new/separate parameter to the `CloudFrontUtilities.create()` API call

_No response_

### AWS Java SDK version used

2.27.19

### JDK version used

openjdk version "17.0.8" 2023-07-18 LTS

macOS Sonoma

RanVaknin:
## Motivation and Context
Fixes #5577 where users cannot specify a wildcard (*) resource URL policy separate from the actual URL to be signed
Current implementation couples the resource URL for both signing and policy specification
This change allows for more flexible URL signing scenarios.

## Modifications
Added new `resourceUrlPattern` field to `CustomSignerRequest`
Updated `CloudFrontUtilities.getSignedUrlWithCustomPolicy` to use resourceUrlPattern when specified

## Testing
Added a unit test to verify `resourceUrlPattern` applies the correct logic to the policy.
Added integration tests to verify real-world scenarios: (Passing on intg test acct ✅ )
- Testing with query parameters
- Testing with wildcard paths
- Testing with full wildcard policy

### DIFF
--- a/.changes/next-release/bugfix-AmazonCloudfront-1fa6f75.json
+++ b/.changes/next-release/bugfix-AmazonCloudfront-1fa6f75.json
@@ -1,0 +1,6 @@
+{
+    "type": "bugfix",
+    "category": "Amazon Cloudfront",
+    "contributor": "",
+    "description": "Decouple policy logic from resource url for getSignedUrlWithCustomPolicy"
+}
--- a/services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
+++ b/services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
@@ -253,9 +253,9 @@ public final class CloudFrontUtilities {
     public SignedUrl getSignedUrlWithCustomPolicy(CustomSignerRequest request) {
         String resourceUrl = request.resourceUrl();
         try {
-            String resourceUrlPattern = request.resourceUrlPattern() != null && !StringUtils.isEmpty(request.resourceUrlPattern())
-                                    ? request.resourceUrlPattern()
-                                    : request.resourceUrl();
+            String resourceUrlPattern = StringUtils.isEmpty(request.resourceUrlPattern())
+                                        ? request.resourceUrl()
+                                        : request.resourceUrlPattern();
 
             String policy = SigningUtils.buildCustomPolicyForSignedUrl(resourceUrlPattern,
                                                                        request.activeDate(),
--- a/services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
+++ b/services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
@@ -217,8 +217,13 @@ public final class CloudFrontUtilities {
      *
      * @param request
      *            A {@link CustomSignerRequest} configured with the following values:
-     *            resourceUrl, privateKey, keyPairId, expirationDate, activeDate (optional), ipRange (optional), resourceUrlPattern
-     *            (optional)
+     *            resourceUrl,
+     *            privateKey,
+     *            keyPairId,
+     *            expirationDate,
+     *            activeDate (optional),
+     *            ipRange (optional),
+     *            resourceUrlPattern (optional)
      * @return A signed URL that will permit access to distribution and S3
      *         objects as specified in the policy document.
      *
--- a/services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
+++ b/services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
@@ -210,13 +210,17 @@ public final class CloudFrontUtilities {
      * based on an access time window and an ip range. The custom policy itself
      * is included as part of the signed URL (For a signed URL with canned
      * policy, there is no policy included in the signed URL).
+     * <p>
+     * <b>Note: </b> If you require a wildcard or a different resource path in the policy than the actual {@code resourceUrl},
+     * use the optional {@code policyResource} field in {@link CustomSignerRequest}.
      * For more information, see <a href=
      * "https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-creating-signed-url-custom-policy.html"
      * >Creating a signed URL using a custom policy</a>.
      *
      * @param request
      *            A {@link CustomSignerRequest} configured with the following values:
-     *            resourceUrl, privateKey, keyPairId, expirationDate, activeDate (optional), ipRange (optional)
+     *            resourceUrl, privateKey, keyPairId, expirationDate, activeDate (optional), ipRange (optional), policyResource
+     *            (optional)
      * @return A signed URL that will permit access to distribution and S3
      *         objects as specified in the policy document.
      *
@@ -233,6 +237,7 @@ public final class CloudFrontUtilities {
      *     Path keyFile = myKeyFile;
      *     Instant activeDate = Instant.now().plus(Duration.ofDays(2));
      *     String ipRange = "192.168.0.1/24";
+     *     String policyResource = resourceUrl + "*";
      *
      *     CustomSignerRequest customRequest = CustomSignerRequest.builder()
      *                                                            .resourceUrl(resourceUrl)
@@ -241,19 +246,29 @@ public final class CloudFrontUtilities {
      *                                                            .expirationDate(expirationDate)
      *                                                            .activeDate(activeDate)
      *                                                            .ipRange(ipRange)
+     *                                                             .policyResource(policyResource)
      *                                                            .build();
      *     SignedUrl signedUrl = utilities.getSignedUrlWithCustomPolicy(customRequest);
      *     String url = signedUrl.url();
      * }
      */
     public SignedUrl getSignedUrlWithCustomPolicy(CustomSignerRequest request) {
+        String resourceUrl = request.resourceUrl();
+
+        if (resourceUrl == null) {
+            throw SdkClientException.create("resourceUrl must not be null. "
+                                            + "To provide a custom policy resource please use the policyResource parameter");
+        }
+
         try {
-            String resourceUrl = request.resourceUrl();
-            String policy = SigningUtils.buildCustomPolicyForSignedUrl(request.resourceUrl(),
+            String policyResource = request.policyResource() != null
+                                    ? request.policyResource()
+                                    : request.resourceUrl();
+
+            String policy = SigningUtils.buildCustomPolicyForSignedUrl(policyResource,
                                                                        request.activeDate(),
                                                                        request.expirationDate(),
-                                                                       request.ipRange(),
-                                                                       request.policyResourceUrl());
+                                                                       request.ipRange());
 
             byte[] signatureBytes = SigningUtils.signWithSha1Rsa(policy.getBytes(UTF_8), request.privateKey());
             String urlSafePolicy = SigningUtils.makeStringUrlSafe(policy);
--- a/services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
+++ b/services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
@@ -254,12 +254,6 @@ public final class CloudFrontUtilities {
      */
     public SignedUrl getSignedUrlWithCustomPolicy(CustomSignerRequest request) {
         String resourceUrl = request.resourceUrl();
-
-        if (resourceUrl == null) {
-            throw SdkClientException.create("resourceUrl must not be null. "
-                                            + "To provide a custom policy resource please use the policyResource parameter");
-        }
-
         try {
             String policyResource = request.policyResource() != null
                                     ? request.policyResource()
--- a/services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
+++ b/services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
@@ -249,8 +249,12 @@ public final class CloudFrontUtilities {
     public SignedUrl getSignedUrlWithCustomPolicy(CustomSignerRequest request) {
         try {
             String resourceUrl = request.resourceUrl();
-            String policy = SigningUtils.buildCustomPolicyForSignedUrl(request.resourceUrl(), request.activeDate(),
-                                                                       request.expirationDate(), request.ipRange());
+            String policy = SigningUtils.buildCustomPolicyForSignedUrl(request.resourceUrl(),
+                                                                       request.activeDate(),
+                                                                       request.expirationDate(),
+                                                                       request.ipRange(),
+                                                                       request.policyResourceUrl());
+
             byte[] signatureBytes = SigningUtils.signWithSha1Rsa(policy.getBytes(UTF_8), request.privateKey());
             String urlSafePolicy = SigningUtils.makeStringUrlSafe(policy);
             String urlSafeSignature = SigningUtils.makeBytesUrlSafe(signatureBytes);
--- a/services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
+++ b/services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
@@ -33,6 +33,7 @@ import software.amazon.awssdk.services.cloudfront.internal.utils.SigningUtils;
 import software.amazon.awssdk.services.cloudfront.model.CannedSignerRequest;
 import software.amazon.awssdk.services.cloudfront.model.CustomSignerRequest;
 import software.amazon.awssdk.services.cloudfront.url.SignedUrl;
+import software.amazon.awssdk.utils.StringUtils;
 
 /**
  *
@@ -210,16 +211,13 @@ public final class CloudFrontUtilities {
      * based on an access time window and an ip range. The custom policy itself
      * is included as part of the signed URL (For a signed URL with canned
      * policy, there is no policy included in the signed URL).
-     * <p>
-     * <b>Note: </b> If you require a wildcard or a different resource path in the policy than the actual {@code resourceUrl},
-     * use the optional {@code policyResource} field in {@link CustomSignerRequest}.
      * For more information, see <a href=
      * "https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-creating-signed-url-custom-policy.html"
      * >Creating a signed URL using a custom policy</a>.
      *
      * @param request
      *            A {@link CustomSignerRequest} configured with the following values:
-     *            resourceUrl, privateKey, keyPairId, expirationDate, activeDate (optional), ipRange (optional), policyResource
+     *            resourceUrl, privateKey, keyPairId, expirationDate, activeDate (optional), ipRange (optional), resourceUrlPattern
      *            (optional)
      * @return A signed URL that will permit access to distribution and S3
      *         objects as specified in the policy document.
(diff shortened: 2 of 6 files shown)


services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/model/CustomSignerRequest.java
```diff
@@ -50,6 +50,7 @@ private CustomSignerRequest(DefaultBuilder builder) {
         this.expirationDate = builder.expirationDate;
```
davidh44, line 50: I know it was already there, can we remove the null validation for `expirationDate` from `SigningUtils.buildCustomPolicyForSignedUrl()` and do a `Validate.notNull()` here? This way it will cover `cookiesForCustomPolicy()` as well.

services/cloudfront/src/test/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilitiesIntegrationTest.java
```diff
@@ -267,6 +271,114 @@ void getCookiesForCustomPolicy_withFutureActiveDate_shouldReturn403Response() th
...
+        SdkHttpClient client = ApacheHttpClient.create();
+        HttpExecuteResponse response = client.prepareRequest(HttpExecuteRequest.builder()
+                                                                               .request(SdkHttpRequest.builder()
+                                                                                                      .encodedPath(modifiedUri.getRawPath() + "?" + modifiedUri.getRawQuery())
+                                                                                                      .host(modifiedUri.getHost())
+                                                                                                      .method(SdkHttpMethod.GET)
+                                                                                                      .protocol("https")
+                                                                                                      .build())
+                                                                               .build()).call();
+        assertThat(response.httpResponse().statusCode()).isEqualTo(200);
```
davidh44, line 344: Looks like both tests are failing with 403

RanVaknin: now passing

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/model/CustomSignerRequest.java
```diff
@@ -35,21 +36,22 @@
...
     private final PrivateKey privateKey;
     private final String keyPairId;
     private final Instant expirationDate;
     private final Instant activeDate;
     private final String ipRange;
+    private final String policyResource;
 
     private CustomSignerRequest(DefaultBuilder builder) {
-        this.resourceUrl = builder.resourceUrl;
+        this.resourceUrl = Validate.notNull(builder.resourceUrl, "resourceUrl must not be null");
```
joviegas, line 48: Can we add a junit
```java
    @Test
    void testNPE(){
        assertThatNullPointerException().isThrownBy(() -> CustomSignerRequest.builder().build())
                                        .withMessageContaining("resourceUrl must not be null");
    }
```

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
```diff
@@ -233,6 +237,7 @@ public SignedUrl getSignedUrlWithCustomPolicy(Consumer<CustomSignerRequest.Build
      *     Path keyFile = myKeyFile;
      *     Instant activeDate = Instant.now().plus(Duration.ofDays(2));
      *     String ipRange = "192.168.0.1/24";
+     *     String policyResource = resourceUrl + "*";
```
joviegas, line 240: The '+' in the Java context could be misinterpreted as string concatenation. 
Can we just say 
String policyResource = resourceUrl(by default) or something like this which is not ambiguous

services/cloudfront/src/test/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilitiesTest.java
```diff
@@ -331,4 +334,44 @@ void getCookiesForCustomPolicy_withActiveDateAndIpRangeOmitted_producesValidCook
         assertThat(cookiesForCustomPolicy.keyPairIdHeaderValue()).isEqualTo("CloudFront-Key-Pair-Id=keyPairId");
     }
 
+    @Test
```
joviegas, line 337: Can we add a parameterized test to test different possibilities of resourceUrlPattern ? 
Something like 
```java
@ParameterizedTest
@MethodSource("provideUrlPatternsAndExpectedResources")
void getSignedURLWithCustomPolicy_policyResourceUrlShouldHandleVariousPatterns(
        String resourceUrlPattern, String expectedResource) {
    String baseUrl = "https://d1234.cloudfront.net/images/photo.jpg";
    Instant expiration = Instant.now().plusSeconds(3600);
    CustomSignerRequest request = CustomSignerRequest.builder()
            .resourceUrl(baseUrl)
            .privateKey(keyPair.getPrivate())
            .keyPairId("keyPairId")
            .resourceUrlPattern(resourceUrlPattern)
            .expirationDate(expiration)
            .build();

    SignedUrl signedUrl = cloudFrontUtilities.getSignedUrlWithCustomPolicy(request);

    // Extract and decode the policy
    String encodedPolicy = signedUrl.url().split("Policy=")[1].split("&")[0];
    String standardBase64 = encodedPolicy
            .replace('-', '+')
            .replace('_', '=')
            .replace('~', '/');
    String decodedPolicy = new String(Base64.getDecoder().decode(standardBase64), UTF_8);
    
    // Build expected policy
    StringBuilder expectedPolicy = new StringBuilder();
    StringJoiner conditions = new StringJoiner(",", "{", "}");
    conditions.add("\"DateLessThan\":{\"AWS:EpochTime\":" + expiration.getEpochSecond() + "}");
    
    expectedPolicy.append("{\"Statement\": [{")
            .append("\"Resource\":\"").append(expectedResource).append("\",")
            .append("\"Condition\":").append(conditions)
            .append("}]}");

    assertThat(decodedPolicy.trim()).isEqualTo(expectedPolicy.toString().trim());
    
    // Verify URL is properly encoded
    assertThat(signedUrl.url()).doesNotContain(" ");
    assertThat(signedUrl.url()).matches("^[\\x20-\\x7E]+$"); // Printable ASCII characters only
}

private static Stream<Arguments> provideUrlPatternsAndExpectedResources() {
    return Stream.of(
        // Basic patterns
        Arguments.of("*", "*"),
        Arguments.of("https://d1234.cloudfront.net/*", "https://d1234.cloudfront.net/*"),
        Arguments.of("https://d1234.cloudfront.net/images/*", "https://d1234.cloudfront.net/images/*"),
        
        // Specific file patterns
        Arguments.of("234.cloudfront.net/images/photo.jpg", 
                    "https://d1234.cloudfront.net/images/photo.jpg"),
        Arguments.of("https://d1234.cloudfront.net/*/photo.jpg", 
                    "234.cloudfront.net/*/photo.jpg"),
        
        // Multiple wildcards
        Arguments.of("https://*.cloudfront.net/*/photo.jpg", 
                    "https://*.cloudfront.net/*/photo.jpg"),
        
        // Patterns with special characters that need encoding
        Arguments.of("https://d1234.cloudfront.net/images/photo with spaces.jpg", 
                    "https://d1234.cloudfront.net/images/photo%20with%20spaces.jpg"),
        Arguments.of("https://d1234.cloudfront.net/images/photo+with+plus.jpg", 
                    "234.cloudfront.net/images/photo%2Bwith%2Bplus.jpg"),
        Arguments.of("https://d1234.cloudfront.net/images/photo?param=value", 
                    "234.cloudfront.net/images/photo%3Fparam%3Dvalue"),
        
        // Unicode characters
        Arguments.of("https://d1234.cloudfront.net/images/photo-ümlaut.jpg", 
                    "https://d1234.cloudfront.net/images/photo-%C3%BCmlaut.jpg"),
        
        // Path traversal patterns
        Arguments.of("https://d1234.cloudfront.net/images/../photo.jpg", 
                    "https://d1234.cloudfront.net/images/../photo.jpg"),
        
        // Patterns with query parameters
        Arguments.of("https://d1234.cloudfront.net/images/photo.jpg?size=large&format=png", 
                    "https://d1234.cloudfront.net/images/photo.jpg%3Fsize%3Dlarge%26format%3Dpng")
   
   ```

Comment (RanVaknin) [reply services/cloudfront/src/test/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilitiesTest.java:337]: The parameterized test suite is testing for URL encoding transformations that don't exist in the code. Looking at `getSignedUrlWithCustomPolicy`, the resource pattern is inserted directly into the policy document as a JSON string without modification, and URL-safe encoding (`makeStringUrlSafe`) happens later to the entire Base64-encoded policy. The test cases expect the Resource field in the policy to contain URL-encoded values (like %20 for spaces), but this is incorrect since the policy is a JSON document, not a URL, and these transformations never occur in the codepath. 

For this reason, in this test suite's case, both the Input and Output arguments will be the same every time.

Comment (joviegas) [reply services/cloudfront/src/test/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilitiesTest.java]: Thanks Ran 
I saw that some how the current implementation on resourceUrl does not handle special cases like below

In CloudFrontUtilitiesIntegrationTest I updated S3_OBJECT_KEY_ON_SUB_PATH to have special charatcters and also updated  getSignedUrlWithCustomPolicy_wildCardPath test to cehck for specific url

```java 
    private static final String S3_OBJECT_KEY_ON_SUB_PATH = "foo/specific space+plus-üspecial";


    @Test
    void getSignedUrlWithCustomPolicy_wildCardPath() throws Exception {

        System.out.println("bucket® "+bucket);
        String resourceUri = "https://" + domainName;
        Instant expirationDate = LocalDate.of(2050, 1, 1)
                                          .atStartOfDay()
                                          .toInstant(ZoneOffset.of("Z"));

        Instant activeDate = LocalDate.of(2022, 1, 1)
                                      .atStartOfDay()
                                      .toInstant(ZoneOffset.of("Z"));

        CustomSignerRequest request = CustomSignerRequest.builder()
                                                         .resourceUrl(resourceUri + "/foo/specific space+plus-üspecial")
                                                         .privateKey(keyFilePath)
                                                         .keyPairId(keyPairId)
                                                         .policyResource(resourceUri + "/foo/specific space+plus-üspecial")
                                                         .activeDate(activeDate)
                                                         .expirationDate(expirationDate)
                                                         .build();

        SignedUrl signedUrl = cloudFrontUtilities.getSignedUrlWithCustomPolicy(request);


        URI modifiedUri = URI.create(signedUrl.url().replace("/specific space+plus-üspecial","/other-file"));
        SdkHttpClient client = ApacheHttpClient.create();
        HttpExecuteResponse response = client.prepareRequest(HttpExecuteRequest.builder()
                                                                               .request(SdkHttpRequest.builder()
                                                                                                      .encodedPath(modifiedUri.getRawPath() + "?" + modifiedUri.getRawQuery())
                                                                                                      .host(modifiedUri.getHost())
                                                                                                      .method(SdkHttpMethod.GET)
                                                                                                      .protocol("https")
                                                                                                      .build())
                                                                               .build()).call();
        assertThat(response.httpResponse().statusCode()).isEqualTo(200);
    }
```
But I get
```
java.lang.IllegalArgumentException: Illegal character in path at index 50: https://d3j5oja9i0ormb.cloudfront.net/foo/specific space+plus-üspecial


```

We need not fix this in this PR but raise an internal task and discuss this.

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/model/CustomSignerRequest.java
```diff
@@ -179,6 +187,16 @@ public interface Builder extends CopyableBuilder<CustomSignerRequest.Builder, Cu
          * IPv6 format is not supported.
          */
         Builder ipRange(String ipRange);
+
+        /**
+         * Configure the resource URL pattern to be used in the policy
+         * <p>
+         * For custom policies, this specifies the URL pattern that determines which files
+         * can be accessed with this signed URL. This can include wildcard characters (*) to
```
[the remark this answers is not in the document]
joviegas: Please ack here once we have created request for above

joviegas (approved):

services/cloudfront/src/test/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilitiesTest.java
RanVaknin: Sounds good. I can create a new task on next steps. 

For reference, both the JS SDK v3 and Go SDK v2 rely on the standard library's URL parsing, and the provided URLs get properly encoded:
Go SDK impl https://github.com/aws/aws-sdk-go-v2/blob/main/feature/cloudfront/sign/sign_url.go#L194

```go
func main() {
	rawURL := "https://example.cloudfront.net/foo/specific space+plus-üspecial"
	uri, err := url.Parse(rawURL)
	if err != nil {
		fmt.Println("Error:", err)
	} else {
		fmt.Println("Parsed successfully", uri)
	}
}
// Parsed successfully https://example.cloudfront.net/foo/specific%20space+plus-%C3%BCspecial
``` 

JS SDK impl https://github.com/aws/aws-sdk-js-v3/blob/main/packages/cloudfront-signer/src/sign.ts#L126

```js
try {
const uri = new URL("https://d3j5oja9i0ormb.cloudfront.net/foo/specific space+plus-üspecial")
    console.log("Parsed ", uri.href)
} catch (e) {
    console.log(e);
}
// "Parsed https://d3j5oja9i0ormb.cloudfront.net/foo/specific%20space+plus-%C3%BCspecial
```

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/model/CustomSignerRequest.java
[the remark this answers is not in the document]
RanVaknin: Created a task to add documentation to the migration guide + updating the blogpost written by @davidh44.

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
```diff
@@ -241,16 +249,24 @@ public SignedUrl getSignedUrlWithCustomPolicy(Consumer<CustomSignerRequest.Build
...
      *     String url = signedUrl.url();
      * }
      */
     public SignedUrl getSignedUrlWithCustomPolicy(CustomSignerRequest request) {
+        String resourceUrl = request.resourceUrl();
         try {
-            String resourceUrl = request.resourceUrl();
-            String policy = SigningUtils.buildCustomPolicyForSignedUrl(request.resourceUrl(), request.activeDate(),
-                                                                       request.expirationDate(), request.ipRange());
+            String resourceUrlPattern = StringUtils.isEmpty(request.resourceUrlPattern())
```
zoewangg, line 261: Is there any reason we are treating empty as null?

.changes/next-release/bugfix-AmazonCloudfront-1fa6f75.json
```diff
@@ -0,0 +1,6 @@
+{
+    "type": "bugfix",
+    "category": "Amazon Cloudfront",
+    "contributor": "",
+    "description": "Decouple policy logic from resource url for getSignedUrlWithCustomPolicy"
```
zoewangg, line 5: This seems to be implementation detail.

Suggesting: 
```
Allow users to specify resource URL pattern in `CloudFrontUtilities#getSignedUrlWithCustomPolicy`. See [#5577](https://github.com/aws/aws-sdk-java-v2/issues/5577)
```

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
RanVaknin: Because `resourceUrlPattern` maps to the JSON Policy document's `Resource` field, if a user provides `null` or `""` we default to the `resourceUrl` itself - meaning, there is no pattern, just use the base url as the `Resource` field in the policy. 

The server side implication of send a policy with a `Resource` field set to "" will result in a 403. I can't foresee a situation in which users are actively prohibiting all access to their distribution just by passing `""`.

For example: 

```java
        CustomSignerRequest request = CustomSignerRequest.builder()
                                                         .resourceUrl(resourceUrl)
                                                         .privateKey(keyFilePath)
                                                         .keyPairId(keyPairId)
                                                         .resourceUrlPattern("")
                                                         .activeDate(activeDate)
                                                         .expirationDate(expirationDate)
                                                         .build();
```
Will result in a 403.

.changes/next-release/bugfix-AmazonCloudfront-1fa6f75.json
RanVaknin: Agreed. Do you want me to raise a PR to amend the changelog?

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
zoewangg: Right, my question is more around: is this something we want to handle in the SDK? Should we just let it fail on the server side instead?

My concern is that some customers may not realize `""` is not supported or have bugs in their code to pass empty string to resourceUrlPattern unintentionally, which will be hidden by this SDK logic.

.changes/next-release/bugfix-AmazonCloudfront-1fa6f75.json
zoewangg: Yeah, that'd be great. It may be easier after release is finished.

services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
RanVaknin: Ack. I don't feel strongly about it. I can just use a standard null check.

PR #5871, opened 2025-02-08
Validate sent body len, remove buffering

dagnir:
## Motivation and Context

This commit removes the buffering of
`RequestBody.fromContentProvider(ContentStreamProvider,long,String)`, which was used to ensure that the provider always provided the expected number of bytes given by `length`.

Instead of doing the buffering, this commit updates `SdkLengthAwareInputstream` to validate the number of bytes read is correct, and throws an error if EOF is reached prematurely. We use this stream to ensure that we sent the correct amount of bytes to the server when sending the HTTP request.

Note that this a weaker check because we're only testing for the *number* of bytes sent rather than their contents but has the benefit of sidestepping buffering.

Note: we do a similar wrapping here https://github.com/aws/aws-sdk-java-v2/blob/9d5af05cd90d1a9181b52d7b0d95f111a5d29f76/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/handler/BaseClientHandler.java#L139, I don't think we necessarily want to remove it there since it still serves truncating purposes.

## Types of changes

## Checklist

## License

### DIFF
--- a/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/MakeHttpRequestStage.java
+++ b/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/MakeHttpRequestStage.java
@@ -16,6 +16,7 @@
 package software.amazon.awssdk.core.internal.http.pipeline.stages;
 
 import java.time.Duration;
+import java.util.Optional;
 import software.amazon.awssdk.annotations.SdkInternalApi;
 import software.amazon.awssdk.core.client.config.SdkClientOption;
 import software.amazon.awssdk.core.interceptor.SdkInternalExecutionAttribute;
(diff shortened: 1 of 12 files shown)


L-Applin:
why the 3 changelog files?

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/MakeHttpRequestStage.java
```diff
@@ -94,4 +103,39 @@ private static long updateMetricCollectionAttributes(RequestExecutionContext con
...
+    }
+
+    private static Optional<Long> contentLength(SdkHttpFullRequest request) {
+        Optional<String> contentLengthHeader = request.firstMatchingHeader(CONTENT_LENGTH);
+
+        if (contentLengthHeader.isPresent()) {
+            try {
+                return Optional.of(Long.parseLong(contentLengthHeader.get()));
+            } catch (NumberFormatException e) {
+                LOG.warn(() -> "Unable to parse 'Content-Length' header. Treating it as non existent.");
```
L-Applin, line 136: are we sure we want WARN level here?

L-Applin (approved):

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/io/SdkLengthAwareInputStream.java
```diff
@@ -48,8 +49,13 @@ public int read() throws IOException {
         }
 
         int read = super.read();
+
         if (read != -1) {
             remaining--;
+        } else if (remaining != 0) { // EOF, ensure we've read the number of expected bytes
+            throw new IllegalStateException("The request content has fewer bytes than the "
+                                            + "specified "
+                                            + "content-length: " + length + " bytes.");
```
millems, line 58: Was there a use-case that existed before, which we don't support anymore? E.g. someone specifies a content-length of 50 for a stream of length 100 so that they only send the first 50 bytes?

zoewangg: This should still be supported. We only throw exception if content-length > number of bytes in the stream. If the stream contains more data, we truncate the stream (stop reading after content-length is reached.

zoewangg:
Closing the PR since it's merged upstream

Issue #5878, opened 2025-02-11
S3 DeleteObjects errors on multi-region access point due to checksum headers


winter-s:
### Describe the bug

SDK 2.30.17 fails with an InvalidRequestException when trying to delete objects on an S3 multi-region access point.
The same code works with SDK 2.29.52

### Regression Issue

- [x] Select this option if this issue appears to be a regression.

### Expected Behavior

The S3 DeleteObjects API should work as it did with SDK 2.29

### Current Behavior

> software.amazon.awssdk.services.s3.model.InvalidRequestException: x-amz-sdk-checksum-algorithm specified, but no corresponding x-amz-checksum-* or x-amz-trailer headers were found. (Service: S3, Status Code: 400, Request ID: 7FBWJHRPYNQXGZ4G, Extended Request ID: 7q8nLrvVN+p/hIB3gCe/KlIFComVGNdSVGkPPV+8mnpxuLWyF8vyITHt3qzAyVMwi29DLVInycQ=)

### Reproduction Steps

```
      var s3Client = S3Client.builder()
          .region("eu-west-1);
          .build();
      var apArn = "arn:aws:s3::1234567890:accesspoint/abcdefg.mrap";

      var paginatedResponse = s3Client.listObjectsV2Paginator(ListObjectsV2Request.builder()
          .bucket(apArn)
          .prefix(path)
          .maxKeys(1000)
          .build());
      for (var aPage : paginatedResponse) {
        var objectIds = aPage.contents().stream()
            .map(s3Object -> ObjectIdentifier.builder()
            .key(s3Object.key())
            .build())
            .toList();
        if (objectIds.isEmpty()) {
          break;
        }
        s3Client.deleteObjects(DeleteObjectsRequest.builder()
            .bucket(apArn)
            .delete(Delete.builder()
                .objects(objectIds)
                .build())
            .build());
      }
```

### Possible Solution

Downgrade to SDK 2.29

### AWS Java SDK version used

2.30.17

### JDK version used

21.0.5-11

### Operating System and version

AWS ECS Fargate arm64 running a bellsoft/liberica-openjre-alpine-musl:21.0.5-11 container

steveloughran:
Related to this?

https://github.com/aws/aws-sdk-java-v2/issues/5801

zoewangg:
We released a fix in `2.30.19`. Please try with the latest version and let us know if you continue to see the issue.

PR #5880, opened 2025-02-12
Updated the SDK to calculate checksums for checksum required operatio…

zoewangg:
…ns that require sigV4a signer

## Motivation and Context
#5878 
Fixed the issue where the SDK skipped checksum calculation for operations that require checksum and use SigV4a scheme

## Modifications

The root cause is that  [DefaultAwsCrtV4aHttpSigner](https://github.com/aws/aws-sdk-java-v2/blob/master/core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/crt/internal/signer/DefaultAwsCrtV4aHttpSigner.java#L63) did not add checksum header, similar to what [AwsV4HttpSigner](https://github.com/aws/aws-sdk-java-v2/blob/master/core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/signer/DefaultAwsV4HttpSigner.java#L61) does. It had been relying on the SDK core to calculate the legacy content-md5 checksum for those operations,  and as part of the checksums by default feature, we removed the legacy md5 calculation and thus no checksum was added. 

We may consider moving checksum header logic back to the core so that we decouple checksum from signing logic in the future

## Testing
Added tests

## Types of changes

## Checklist

## License

### DIFF
--- a/core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/crt/internal/signer/DefaultAwsCrtV4aHttpSigner.java
+++ b/core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/crt/internal/signer/DefaultAwsCrtV4aHttpSigner.java
@@ -25,6 +25,7 @@ import static software.amazon.awssdk.crt.auth.signing.AwsSigningConfig.AwsSignin
 import static software.amazon.awssdk.http.auth.aws.crt.internal.util.CrtHttpRequestConverter.toRequest;
 import static software.amazon.awssdk.http.auth.aws.crt.internal.util.CrtUtils.sanitizeRequest;
 import static software.amazon.awssdk.http.auth.aws.crt.internal.util.CrtUtils.toCredentials;
+import static software.amazon.awssdk.http.auth.aws.internal.signer.util.ChecksumUtil.checksummer;
 import static software.amazon.awssdk.http.auth.aws.internal.signer.util.ChecksumUtil.hasChecksumHeader;
 import static software.amazon.awssdk.http.auth.aws.internal.signer.util.ChecksumUtil.isPayloadSigning;
 import static software.amazon.awssdk.http.auth.aws.internal.signer.util.ChecksumUtil.useChunkEncoding;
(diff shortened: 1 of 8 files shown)


core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/crt/internal/signer/DefaultAwsCrtV4aHttpSigner.java
```diff
@@ -68,10 +70,11 @@ public final class DefaultAwsCrtV4aHttpSigner implements AwsV4aHttpSigner {
 
     @Override
     public SignedRequest sign(SignRequest<? extends AwsCredentialsIdentity> request) {
+        Checksummer checksummer = checksummer(request, null);
```
zoewangg, line 73: This is the same as what we have in https://github.com/aws/aws-sdk-java-v2/blob/master/core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/signer/DefaultAwsV4HttpSigner.java#L62

core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/crt/internal/signer/DefaultAwsCrtV4aHttpSigner.java
```diff
@@ -205,20 +208,22 @@ private static void configurePayloadSigning(AwsSigningConfig signingConfig, bool
...
 
     private static SignedRequest doSign(SignRequest<? extends AwsCredentialsIdentity> request,
+                                        Checksummer checksummer,
                                         AwsSigningConfig signingConfig,
                                         V4aPayloadSigner payloadSigner) {
+        SdkHttpRequest.Builder requestBuilder = request.request().toBuilder();
         ContentStreamProvider contentStreamProvider = request.payload().orElse(null);
+
+        checksummer.checksum(contentStreamProvider, requestBuilder);
+        payloadSigner.beforeSigning(requestBuilder, contentStreamProvider, signingConfig.getSignedBodyValue());
```
L-Applin, line 218: just double checking, could moving `beforeSigning` call ahead for anonymous request cause any issue?

zoewangg: It seems anonymous requests are not supported in sigv4a, so the ordering doesn't really matter, but I moved  beforeSigning and checksum methods call down to preserve the existing behavior

zoewangg:
Closing since we merged the change upstream

PR #5885, opened 2025-02-17
Custom lookup for reflective schemas

dagnir:
## Motivation and Context
This commit updates `BeanTableSchema` and `ImmutableTableSchema` to allow
customers to provide their own instance of `MethodHandle.Lookup`.

This is a necessary feature for situations where either of these schemas
are used in an application where the item classes and the SDK classes
are loaded by different classloaders. Since these schemas work by
reflectively scanning the item class to find the attributes' getters and
setters (or builder for immutable classes), the enhanced client makes
use of a LambdaMetaFactory to construct lambdas that directly call these
methods to avoid reflection when mapping to and from a
`AttributeValue`s. At least as of Java 21, the way that lambdas are
constructed (all lambdas, not just the ones created by the enhanced
client), is they are made internal classes of the "caller" class, i.e.
the class that defined it. The calling class is identified by the class
that created the `Lookup` object. This means that when using the default
`Lookup` created `LambdaToMethodBridgeBuilder`, all of these lambdas are
inner classes of of `LambdaToMethodBridgeBuilder`. This is an issue when
the item class is from a different classloader (or module) because when
the lambda attempts to reference the item class (e.g. to call a getter),
it's `ClassLoader` will throw a `ClassNotFoundException` because because
it's not a class that it knows about.

By allowing the customer to provide a `Lookup` object, they can
guarantee that the the `Lookup` has visibility and access to the item
classes.

## Testing
 - Added new tests
 
## Types of changes

## Checklist

## License


services-custom/dynamodb-enhanced/src/test/resources/software/amazon/awssdk/enhanced/dynamodb/mapper/test-pojos.jar
dagnir, line 1: This JAR just contains the test POJOs.

I opted to just have a one-off JAR since it's simpler than setting up a new module and using that as a test dependency for the enhanced client, since it involves updating the import config and release yml files.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/mapper/BeanTableSchemaParams.java
```diff
@@ -0,0 +1,109 @@
...
+import java.lang.invoke.MethodHandles;
+import software.amazon.awssdk.annotations.Immutable;
+import software.amazon.awssdk.annotations.SdkPublicApi;
+import software.amazon.awssdk.utils.builder.CopyableBuilder;
+import software.amazon.awssdk.utils.builder.ToCopyableBuilder;
+
+/**
+ * Parameter class used for creating {@link TableSchema}'s that leverage reflection.
+ * @param <T>
+ */
```
millems, line 27: Unused `@param`.

Can we be a bit more specific? E.g. `Configuration for {@link TableSchema#fromBean} or {@link BeanTableSchema#create}`?

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/mapper/BeanTableSchemaParams.java
```diff
@@ -0,0 +1,109 @@
...
+    public Builder<T> toBuilder() {
+        return new BuilderImpl<>(this);
+    }
+
+    public static <T> Builder<T> builder(Class<T> beanClass) {
+        return new BuilderImpl<>(beanClass);
+    }
+
+    public interface Builder<T> extends CopyableBuilder<Builder<T>, BeanTableSchemaParams<T>> {
+        Builder<T> beanClass(Class<T> beanClass);
```
millems, line 58: If we allow specifying this, we should document it.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/mapper/BeanTableSchema.java
```diff
@@ -125,31 +127,61 @@ private BeanTableSchema(StaticTableSchema<T> staticTableSchema) {
...
+     * Generally, this method should be preferred over {@link #create(Class)} because it allows you to use a custom
+     * {@link MethodHandles.Lookup} instance, which is necessary when your application runs in an environment where your
+     * application code and dependencies like the AWS SDK for Java are loaded by different classloaders.
+     *
+     * @param params The parameters object.
+     * @param <T> The bean class type.
+     * @return An initialized {@link BeanTableSchema}
+     */
+    @SuppressWarnings("unchecked")
+    public static <T> BeanTableSchema<T> create(BeanTableSchemaParams<T> params) {
```
millems, line 163: What about `TableSchema#fromBean`?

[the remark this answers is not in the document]
dagnir: There's no precedent in the Enhanced Client for naming of parameters objects, but in core, we do have several public API's that follow the 'Params' naming, e.g. in `auth`: https://sdk.amazonaws.com/java/api/2.27.12/software/amazon/awssdk/auth/signer/params/package-summary.html

The endpoint providers, also expose follow this pattern https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/s3/endpoints/S3EndpointParams.html

dagnir: Completely forgot about this. Done!

dagnir:
Integ failure not related, it's Kinesis

millems (approved):

PR #5892, opened 2025-02-19
Fixing cloudfront signer logic, and changelog

RanVaknin:
Fixing changelog and minor logic for https://github.com/aws/aws-sdk-java-v2/pull/5862

### DIFF
--- a/services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
+++ b/services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
@@ -258,7 +258,7 @@ public final class CloudFrontUtilities {
     public SignedUrl getSignedUrlWithCustomPolicy(CustomSignerRequest request) {
         String resourceUrl = request.resourceUrl();
         try {
-            String resourceUrlPattern = StringUtils.isEmpty(request.resourceUrlPattern())
+            String resourceUrlPattern = request.resourceUrlPattern() == null
                                         ? request.resourceUrl()
                                         : request.resourceUrlPattern();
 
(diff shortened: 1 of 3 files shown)


services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/CloudFrontUtilities.java
```diff
@@ -258,7 +258,7 @@ public SignedUrl getSignedUrlWithCustomPolicy(Consumer<CustomSignerRequest.Build
     public SignedUrl getSignedUrlWithCustomPolicy(CustomSignerRequest request) {
         String resourceUrl = request.resourceUrl();
         try {
-            String resourceUrlPattern = StringUtils.isEmpty(request.resourceUrlPattern())
+            String resourceUrlPattern = request.resourceUrlPattern() == null
```
zoewangg, line 261: Don't we need to update tests accordingly?

RanVaknin: No. There is no test case that checks for `""` specifically. 
Both before and after the change this codepath never runs into an issue with null because it uses `resourceUrl` as a backfill which is [guaranteed](https://github.com/aws/aws-sdk-java-v2/blob/9fa37f9d4a057291d830654f94609ec291130058/services/cloudfront/src/main/java/software/amazon/awssdk/services/cloudfront/model/CustomSignerRequest.java#L48) to be not null or empty.

zoewangg: Gotcha, should we have tests though ;)

zoewangg (approved):

PR #5895, opened 2025-02-20
Do not sign Transfer-Encoding Header  if present on the request

joviegas:
## Motivation and Context

Currently, AWS SDKs sign the Transfer-Encoding chunked header during SigV4 authentication, causing signature validation failures in specific scenarios.

The issue occurs because Transfer-Encoding is a hop-by-hop header that intermediate proxies can modify. 

This change proposes to stop signing the Transfer-Encoding header in SigV4(a).

Note : for Sigv4a signers using CRT it is done in  crt side with https://github.com/awslabs/aws-c-auth/pull/261 

## Modifications

- Added Transfer-Encoding in list of `LIST_OF_HEADERS_TO_IGNORE_IN_LOWER_CASE` which already skips some of predefined headers

## Testing

- Added Junits
- Added codegen-tst and checked Authorization doesnot have ` Transfer-Encoding`

## Types of changes

## License

### DIFF
--- a/core/auth/src/main/java/software/amazon/awssdk/auth/signer/internal/AbstractAws4Signer.java
+++ b/core/auth/src/main/java/software/amazon/awssdk/auth/signer/internal/AbstractAws4Signer.java
@@ -68,7 +68,7 @@ public abstract class AbstractAws4Signer<T extends Aws4SignerParams, U extends A
     private static final FifoCache<SignerKey> SIGNER_CACHE =
         new FifoCache<>(SIGNER_CACHE_MAX_SIZE);
     private static final List<String> LIST_OF_HEADERS_TO_IGNORE_IN_LOWER_CASE =
-        Arrays.asList("connection", "x-amzn-trace-id", "user-agent", "expect");
+        Arrays.asList("connection", "x-amzn-trace-id", "user-agent", "expect", "transfer-encoding");
 
     protected SdkHttpFullRequest.Builder doSign(SdkHttpFullRequest request,
                                                 Aws4SignerRequestParams requestParams,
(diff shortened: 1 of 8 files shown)


core/auth/src/main/java/software/amazon/awssdk/auth/signer/internal/AbstractAws4Signer.java
```diff
@@ -68,7 +68,7 @@ public abstract class AbstractAws4Signer<T extends Aws4SignerParams, U extends A
     private static final FifoCache<SignerKey> SIGNER_CACHE =
         new FifoCache<>(SIGNER_CACHE_MAX_SIZE);
     private static final List<String> LIST_OF_HEADERS_TO_IGNORE_IN_LOWER_CASE =
-        Arrays.asList("connection", "x-amzn-trace-id", "user-agent", "expect");
+        Arrays.asList("connection", "x-amzn-trace-id", "user-agent", "expect", "transfer-encoding");
```
davidh44, line 71: We have the same list in HeaderTransformsHelper used by AwsS3V4ChunkSigner, do we need to add TE there also?

joviegas: Nope , it is not required here because SDK add only one Header(singletonMap) via this function , that is checksum header and its always "x-amz-checksum-XYZ" 
https://github.com/aws/aws-sdk-java-v2/blob/6c74e87c0ef16472b33fb2dc33de6a18806dbe3a/core/auth/src/main/java/software/amazon/awssdk/auth/signer/internal/chunkedencoding/AwsS3V4ChunkSigner.java#L104-L106

joviegas: Done.

davidh44 (approved):

PR #5834, opened 2025-01-28
Adding attempt count to error message

RanVaknin:
## Motivation and Context
Improve debugging visibility by making attempt count immediately visible in exception messages. Currently, attempt information is only available through suppressed exceptions, requiring additional code to access. This change makes attempts visible directly in the exception message, matching behavior in other AWS SDKs.

## Modifications
- Added attempt count tracking and message formatting in `SdkException`
- Added SdkDiagnostics class to handle the formatting of the Diagnostics clause in error messages.
- Modified `AwsServiceException` to include attempts in AWS error details
- Modified `RetryableStageHelper` to track and propagate attempt counts during retries

## Testing
- Added new functional test `ExceptionAttemptMessageBehaviorTest` testing with different retry scenarios
- Modified existing tests in `AwsServiceExceptionTest` and `SdkExceptionSerializationTest` to verify attempt count behavior
- Added test cases in `SdkExceptionMessageTest` for new attempt count functionality

### Before change:
```java
// non-retryable scenario
try {
    // request to a non-existent bucket
    s3Client.listObjects(g -> g.bucket("some-non-existent-bucket"));
} catch (AwsServiceException e) {
    System.out.println(e);
}
// outputs:
// NoSuchBucketException: The specified bucket does not exist (Service: S3, Status Code: 404, Request ID: abc123, Extended Request ID: xyzdef123456)


// retryable scenario
try {
    // request that might cause a retryable error (e.g., throttling)
    dynamoDbClient.getItem(...);
} catch (AwsServiceException e) {
    System.out.println(e);
}
// outputs:
// ThrottlingException: Rate of requests exceeds the allowed throughput (Service: DynamoDB, Status Code: 429, Request ID: abc123, Extended Request ID: xyzdef123456)
// Note: attempts only visible through e.getSuppressed()
```
### After change:

```java
// non-retryable scenario
try {
    // request to a non-existent bucket
    s3Client.listObjects(g -> g.bucket("some-non-existent-bucket"));
} catch (AwsServiceException e) {
    System.out.println(e);
}
// outputs:
// NoSuchBucketException: The specified bucket does not exist (Service: S3, Status Code: 404, Request ID: abc123, Extended Request ID: xyzdef123456) (SDK Diagnostics: numAttempts = 1)


// retryable scenario
try {
    // request that might cause a retryable error (e.g., throttling)
    dynamoDbClient.getItem(...);
} catch (AwsServiceException e) {
    System.out.println(e);
}
// outputs:
[...]

PR #5907, opened 2025-02-26
Fixing nullability issue of message and adding test coverage

RanVaknin:
## Motivation and Context
#5834 refactors `AwsServiceException` to add the SDK diagnostics conditionally and refactors the `getMessage` impl to use String joiner. This change makes AwsServiceException's getMessage to always be not null which can break customers who are relying on nullability. 

This change preserves the behavior pre #5834. 

Added test coverage.


zoewangg (approved):
Can we add a changelog entry?

dagnir:
@all-contributors please add @RanVaknin for code

Issue #5846, opened 2025-01-31
High CPU Usage in S3 Java SDK Due to Excessive URL Parsing and Exception Stack Trace Generation During getObject Requests


jasonstack:
### Describe the bug

We are using `software.amazon.awssdk:S3:2.29.21`: `DefaultS3AuthSchemeProvider#resolveAuthScheme` seems to parse the `String endpoint` on every request and catching exceptions when some endpoint rule is mismatched in `DefaultS3EndpointProvider#endpointRule64`.

This is how we configure the client.
```
        S3CrtAsyncClient.builder()
                .credentialsProvider(StaticCredentialsProvider.create(AwsBasicCredentials.create(accessKey, secretKey))
                .retryConfiguration(b -> b.numRetries(3))
                .checksumValidationEnabled(true)
                .forcePathStyle(true)
                .endpointOverride(URI.create(endpoint));
```

Attached CPU profile:

<img width="1718" alt="Image" src="https://github.com/user-attachments/assets/aa6b5dca-caea-4e78-9c29-a3ee889416ba" />

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

Avoid parsing URL and filing stacktrace on endpoint rule mismatch per request

### Current Behavior

Parsing URL on every request and filling stacktrace when endpoint rule mismatches

### Reproduction Steps

Run `getObjects` request

### AWS Java SDK version used

2.29.21

### JDK version used

21

### Operating System and version

Ubuntu

sugmanue:
This is not a bug, but the expected behavior. Granted, we could improve its performance a bit, but the behavior is required for S3 to work correctly.

jasonstack:
thanks for improving it!

sugmanue:
Thank you for the through report!

PR #5911, opened 2025-02-27
Improve the IP addresses detection

sugmanue:
## Motivation and Context

The previous code relied on the exception thrown by `Integer.parseInt` to detect when the given segment was not a integer number. This uses exception as control flow and it can be very inefficient as creating an exception is costly.

Instead of using `Integer.parseInt` we just handle the few cases that we care about depending on the length of the segment. If the length is one then check that the char is a digit, if the length is two then check that both chars are digits, if the length is 3, check that all the chars are digits and that the converted number is less than 255.

See also #5846 reporting "High CPU Usage in S3 Java SDK Due to Excessive URL Parsing and Exception Stack Trace Generation During getObject Requests"

## Types of changes

## Checklist

## License


codegen/src/main/resources/software/amazon/awssdk/codegen/rules2/RuleUrl.java.resource
```diff
@@ -48,29 +48,14 @@ public final class RuleUrl {
...
-            })) {
-                isIpAddr = true;
-            }
-        }
+        boolean isIpAddr = (host.startsWith("[") && host.endsWith("]")) || isIpAddr(host);
         String normalizedPath;
         if (StringUtils.isBlank(path)) {
             normalizedPath = "/";
+        } else if (path.startsWith("/") && path.endsWith("/")) {
+            normalizedPath = path;
```
millems, line 58: Is it worth it to reuse the result of startsWith and endsWith? Honestly, the compiler might already be handling that.

```java
        if (StringUtils.isBlank(path)) {
            normalizedPath = "/";
        } else {
            boolean startsWithSlash = path.startsWith("/");
            boolean endsWithSlash = path.endsWith("/");

            if (startsWithSlash && endsWithSlash) {
                normalizedPath = path;
            } else {
                StringBuilder builder = new StringBuilder();
                if (!startsWithSlash) {
                    builder.append("/");
                }
                builder.append(path);
                if (!endsWithSlash) {
                    builder.append("/");
                }
                normalizedPath = builder.toString();
            }
        }
```

Regardless, it feels like it's time to put the "normalize this path" code in a new function, which will flatten it a lot:

```java
if (StringUtils.isBlank(path)) {
    return "/";
}

boolean startsWithSlash = path.startsWith("/");
boolean endsWithSlash = path.endsWith("/");

if (startsWithSlash && endsWithSlash) {
    return path;
}

StringBuilder builder = new StringBuilder();
if (!startsWithSlash) {
    builder.append("/");
}
builder.append(path);
if (!endsWithSlash) {
    builder.append("/");
}
return builder.toString();
```

codegen/src/main/resources/software/amazon/awssdk/codegen/rules2/RuleUrl.java.resource
```diff
@@ -85,4 +70,58 @@ public final class RuleUrl {
...
+            int length = index - from;
+            if (length == 1) {
+                char ch0 = host.charAt(from);
+                if (ch0 < '0' || ch0 > '9') {
+                    return false;
+                }
+            } else if (length == 2) {
+                char ch0 = host.charAt(from);
+                char ch1 = host.charAt(from + 1);
+                if ((ch0 < '0' || ch0 > '9') || (ch1 < '0' || ch1 > '9')) {
```
millems, line 100: We should probably add a comment that we don't validate leading zeroes. E.g. 000.000.000.000 is not a valid IP address, but this would think it is an IP address.

That's probably fine?

codegen/src/main/resources/software/amazon/awssdk/codegen/rules2/RuleUrl.java.resource
```diff
@@ -85,4 +70,58 @@ public final class RuleUrl {
...
+                char ch2 = host.charAt(from + 2);
+                if ((ch0 < '0' || ch0 > '9')
+                        || (ch1 < '0' || ch1 > '9')
+                        || (ch2 < '0' || ch2 > '9')
+                ) {
+                    return false;
+                }
+                int value = (host.charAt(from) - '0') * 100
+                        + (host.charAt(from + 1) - '0') * 10
+                        + (host.charAt(from + 2) - '0');
```
millems, line 115: nit: reuse ch0, ch1, ch2 instead of doing another charAt.

codegen/src/main/resources/software/amazon/awssdk/codegen/rules2/RuleUrl.java.resource
```diff
@@ -85,4 +70,58 @@ public final class RuleUrl {
 
         return new RuleUrl(parsed.getProtocol(), parsed.getAuthority(), path, normalizedPath, isIpAddr);
     }
+
+    private static boolean isIpAddr(String host) {
```
millems, line 74: Can we make this package-protected and add tests for it?

codegen/src/main/resources/software/amazon/awssdk/codegen/rules2/RuleUrl.java.resource
sugmanue: 🤦 Not sure how I missed that.

codegen/src/main/resources/software/amazon/awssdk/codegen/rules2/RuleUrl.java.resource
sugmanue: Ended up just validating that the first char is > '0'. That should do. It won't be backwards compatible but more correct, so let's do it.

test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/compiledendpointrules/endpoints/internal/RuleUrlTest.java
```diff
@@ -0,0 +1,142 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import java.util.Arrays;
+import java.util.Collection;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
+
+class RuleUrlTest {
+
+    public static Collection<TestIsIpCase> testIsIpCases() {
```
millems, line 27: Extra test cases (which will pass, so just add to protect against regressions):
* Three-segment IP addresses, e.g. 1.2.3
* Double-zero prefixed three-digit IP addresses, e.g. 001.0.0.0

millems (approved):

test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/compiledendpointrules/endpoints/internal/RuleUrlTest.java
sugmanue: Added in #5929.

PR #5916, opened 2025-02-28
bugfix: Fixed memory leak issue in SqsBatchManager for RequestBatchManager sendRequest Calls in SqsBatchManager

joviegas:
## Motivation and Context

- Sqs Batch Manager user reported Memory leak while using Batch Send Messages

## Modifications

- Cleared  the  entries  in the Collections used in RequestBatchManager after Completable futures are completdd

## Testing

- Functionalty not impacted so existing test cases cover this.
- Locally ran profiler without fix and observed there was a memory leak in some time , then used the fix and ran again and made sure there is no leak due to Sqs Batch Manager.

## Types of changes

## License

### DIFF
--- a/services/sqs/src/main/java/software/amazon/awssdk/services/sqs/internal/batchmanager/RequestBatchManager.java
+++ b/services/sqs/src/main/java/software/amazon/awssdk/services/sqs/internal/batchmanager/RequestBatchManager.java
@@ -17,7 +17,6 @@ package software.amazon.awssdk.services.sqs.internal.batchmanager;
 
 import java.time.Duration;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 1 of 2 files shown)


services/sqs/src/main/java/software/amazon/awssdk/services/sqs/internal/batchmanager/RequestBatchManager.java
```diff
@@ -67,6 +67,7 @@ protected RequestBatchManager(RequestBatchConfiguration overrideConfiguration,
     public CompletableFuture<ResponseT> batchRequest(RequestT request) {
         CompletableFuture<ResponseT> response = new CompletableFuture<>();
         pendingResponses.add(response);
```
millems, line 69: I notice when we're creating this set, we're using:
`Collections.newSetFromMap(new ConcurrentHashMap<>());`

For ConcurrentHashMap, it's more standard to use:
`new ConcurrentHashMap<>().newKeySet()`

millems (approved):

services/sqs/src/main/java/software/amazon/awssdk/services/sqs/internal/batchmanager/RequestBatchManager.java
joviegas: Done

dagnir:
Removed from queue since we're merging upstream

joviegas:
Merged it upstream

PR #5920, opened 2025-03-01
Migration tool : TransferManager constructor and methods

davidh44:
## Motivation and Context

Adds transforms for some TransferManager constructors and methods

```
TransferManager()
TransferManagerBuilder.standard().build()
TransferManagerBuilder.defaultTransferManager()

copy(CopyObjectRequest copyObjectRequest)
copy(String sourceBucketName, String sourceKey, String destinationBucketName, String destinationKey)
download(GetObjectRequest getObjectRequest, File file)
download(GetObjectRequest getObjectRequest, File file, long timeoutMillis)
download(String bucket, String key, File file)
download(String bucket, String key, File file, long timeoutMillis)
upload(String bucketName, String key, File file)
```

## Modifications

- new recipe `TransferManagerMethodsToV2`
- Updated `V1BuilderVariationsToV2Builder` and `SdkTypeUtils` to transform v1 `TransferManager` empty constructor and `TransferManagerBuilder` methods
- new separate composite recipe `AwsSdkJavaV1ToV2Experimental` which adds transfer manager recipes to parent recipe `AwsSdkJavaV1ToV2`

## Testing

Added unit and end-to-end tests


test/v2-migration-tests/src/test/resources/software/amazon/awssdk/v2migrationtests/maven/after/src/main/java/foo/bar/TransferManagerS3.java
```diff
@@ -0,0 +1,67 @@
...
+            .build();
+        S3TransferManager tmBuilderDefault = S3TransferManager.create();
+        S3TransferManager tmBuilderWithS3 = S3TransferManager.builder().build();
+    }
+
+    void download(S3TransferManager tm, String bucket, String key) {
+        FileDownload download = tm.downloadFile(DownloadFileRequest.builder().getObjectRequest(GetObjectRequest.builder().bucket(bucket).key(key).build()).destination(file).build());
+
+        long timeout = 89;
+        FileDownload download2 = tm.downloadFile(DownloadFileRequest.builder().getObjectRequest(GetObjectRequest.builder().bucket(bucket).key(key).overrideConfiguration(AwsRequestOverrideConfiguration.builder().apiCallTimeout(Duration.ofMillis(timeout)).build()).build()).destination(file).build());
```
Fred1155, line 46: nit: since I'm new to migration tool, do we normally include indentation for builder methods?

Fred1155 (approved):

test/v2-migration-tests/src/test/resources/software/amazon/awssdk/v2migrationtests/maven/after/src/main/java/foo/bar/TransferManagerS3.java
davidh44: This is just generated files, we can ignore indentation :)

v2-migration/src/main/resources/META-INF/rewrite/aws-sdk-java-v1-to-v2-with-tm.yml
```diff
@@ -0,0 +1,45 @@
...
+---
+type: specs.openrewrite.org/v1beta/recipe
+name: software.amazon.awssdk.v2migration.AwsSdkJavaV1ToV2WithTransferManager
+displayName: Migrate from the AWS SDK for Java v1 to the AWS SDK for Java v2 including S3 Transfer Manager.
+description: This recipe will apply changes required for migrating from the AWS SDK for Java v1 to the AWS SDK for Java v2, including S3 Transfer Manager.
+tags:
+  - aws
+  - sdk
+recipeList:
+  - software.amazon.awssdk.v2migration.AddTransferManagerDependency
```
zoewangg, line 24: Can we just use `software.amazon.awssdk.v2migration.AwsSdkJavaV1ToV2`?

davidh44: Might be easier to have a separate composite recipe, to reference in internal testing. Plus, we may need to add recipes in the middle of the order

zoewangg: I see, my concern is that since we updated our tests to test with this new recipe, we won't be able to catch regressions for the main recipe and this diverge would make it harder to maintain in the long term. I wonder if there's a way to add a configuration to the main recipe to toggle the behavior of enabling TM recipe

davidh44: > my concern is that since we updated our tests to test with this new recipe, we won't be able to catch regressions for the main recipe and this diverge would make it harder to maintain in the long term

The TM recipes only add functionality, it shouldn't affect any existing core transforms. The goal is to have one single parent recipe eventually once the TM recipes are complete. 

The current TM transforms make use of V1BuilderVariationsToV2Builder, NewClassToBuilder, and V1SetterToV2, may be more efficient to re-use the logic instead of copying them to new classes. TM code will include S3 POJOs so will need the S3 recipes as well

davidh44: Updated end to end tests to have separate testing for AwsSdkJavaV1ToV2 and AwsSdkJavaV1ToV2WithTransferManager

zoewangg: I think creating separate recipes would make it easier to maintain and extend IMO. For example, if we want to support migrating individual service client, we'd need to have separate recipes for those anyway.  We can probably create common utilities to reuse code to reduce duplication.  

With that said, this is not a blocker, but let's add a backlog item to separate them

v2-migration/src/main/resources/META-INF/rewrite/aws-sdk-java-v1-to-v2-with-tm.yml
```diff
@@ -0,0 +1,45 @@
...
+#  http://aws.amazon.com/apache2.0
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+#
+---
+type: specs.openrewrite.org/v1beta/recipe
+name: software.amazon.awssdk.v2migration.AwsSdkJavaV1ToV2WithTransferManager
```
zoewangg, line 17: Instead of calling `AwsSdkJavaV1ToV2WithTransferManager`, can we call it something like `AwsSdkJavaV1ToV2Beta` or `AwsSdkJavaV1ToV2Experimental`

v2-migration/src/main/resources/META-INF/rewrite/aws-sdk-java-v1-to-v2-with-tm.yml
davidh44: Makes sense, creating tracking ticket for refactor

zoewangg (approved):

PR #5925, opened 2025-03-04
Fix Flaky Test in AsyncApiCallTimeoutTest.streamingOperation_finishedWithinTime_shouldNotTimeout test cases by adding retry

joviegas:
## Motivation and Context

In the last 3 weeks AsyncApiCallTimeoutTest has been failing for streamingOperation_finishedWithinTime_shouldNotTimeout  test 
| Date | Execution Time |
|------|---------------|
| 2025-03-03 | 0.419 seconds |
| 2025-02-27 | 0.359 seconds |
| 2025-02-27 | 3.157 seconds |
| 2025-02-26 | 0.449 seconds |
| 2025-02-26 | 2.749 seconds |
| 2025-02-21 | 1.503 seconds |
| 2025-02-20 | 2.604 seconds |

Based on Above there is no point in increasing the timeout , instead we need to Retry the test same as stability test.

## Modifications

- Added retry only for this test since for last 3 weeks we have data for this test case failure and it always passes on retry.

## Testing
- Locally tested by giving TIMEOUT to lower values in IDE and it was sucessfully retyring and passing in 2nd and 3rd attempts
```
    protected static final Duration TIMEOUT = Duration.ofMillis(20);
```

## License


L-Applin:
Everything looks good to me, just need to fix checkstyle 👍

L-Applin (approved):

Issue #5779, opened 2025-01-08
Service endpoint_url from config file ignored


fkjaekel:
### Describe the bug

In order to use the least environment variables possible with localstack, I created the following profile:

```
[services localstack-specific]
s3 =
  endpoint_url = http://s3.localhost.localstack.cloud:4566

[profile localstack]
region = us-east-1
output = json
endpoint_url=http://localhost:4566
services = localstack-specific
```

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

Be able to use the endpoint_url from config file instead of the AWS_ENDPOINT_URL_S3 environment variable

### Current Behavior

Since the endpoint_url from config is not being used, if I don't set AWS_ENDPOINT_URL_S3 I get this at localstack:

File "/opt/code/localstack/.venv/lib/python3.11/site-packages/localstack/aws/protocol/parser.py", line 779, in _parse_xml_string_to_dom
    raise ProtocolParserError(
localstack.aws.protocol.parser.ProtocolParserError: Unable to parse request (not well-formed (invalid token): line 1, column 0), invalid XML received:
b'{"sourceId":"1","eventDateTime":"2024-11-14T12:13:18.6870-03:00","code":"1    ","sourceApplicationId":1,"display":"TESTE","active":"true "}'
 (Service: S3, Status Code: 500, Request ID: 4d0a9fbd-6893-4f2d-9f55-138ddb31fbd6, Extended Request ID: s9lzHYrFp76ZVxRcpX9+5cjAnEH2ROuNkd2BHfIa6UkFVdtjf5mKR3/eTPFvsiP/XV/VLi31234=)
	Suppressed: software.amazon.awssdk.core.exception.SdkClientException: Request attempt 2 failure: exception while calling s3 with unknown operation: Traceback (most recent call last):
  File "/opt/code/localstack/.venv/lib/python3.11/site-packages/localstack/aws/protocol/parser.py", line 776, in _parse_xml_string_to_dom
    parser.feed(xml_string)
xml.etree.ElementTree.ParseError: not well-formed (invalid token): line 1, column 0

### Reproduction Steps

set AWS_PROFILE=localstack and use S3AsyncClient

### Additional Information/Context

localstack version: 4

### AWS Java SDK version used

2.29.47

### JDK version used

[...]

PR #5562, opened 2024-09-04
Add support for service-specific endpoint overrides without code changes.

millems:
This adds support for this feature: https://docs.aws.amazon.com/sdkref/latest/guide/feature-ss-endpoints.html

The commit has been broken up into 6 commits, to hopefully simplify the code review:

**[Commit 1](https://github.com/aws/aws-sdk-java-v2/pull/5562/commits/d298fcaf9f0026d5574fe382ae3f5d46b6f715b2) - Made protected API changes in support of service-specific endpoint overrides.**

This introduces new `ClientEndpointProvider` abstraction, which is used by clients to resolve the "default" endpoint for the client. The main implementation is the new `AwsClientEndpointProvider` class, which checks the client-level endpoint override, environment variables, system properties and profile properties, in priority order. More information about what it checks is documented [here](https://docs.aws.amazon.com/sdkref/latest/guide/feature-ss-endpoints.html).

Other Protected API changes:
* DefaultServiceEndpointBuilder - Deprecated in favor of the new ClientEndpointProvider and AwsClientEndpointProvider
* Deprecated SdkClientOption.ENDPOINT and SdkClientOption.ENDPOINT_OVERRIDDEN in favor of a new SdkClientOption.CLIENT_ENDPOINT_PROVIDER. The new property allows updating both values in a single field. It was a necessary improvement so that these properties can stay in sync, reliably.
* Deprecated SdkExecutionAttribute.CLIENT_ENDPOINT and SdkClientOption.ENDPOINT_OVERRIDDEN in favor of a new SdkInternalExecutionAttribute.CLIENT_ENDPOINT_PROVIDER. The new property exists for much the same reasons as the new SdkClientOption, but also provides an opportunity to hide these should-be-internal-to-the-SDK attributes.

**[Commit 2](https://github.com/aws/aws-sdk-java-v2/pull/5562/commits/7860e1f75d61bb4b9e226ce551def6d5cf14e32f) - Wires up the new AwsClientEndpointProvider for service clients**

In this change, service client builders set the `AwsClientEndpointProvider` when they are created. This allows setting the service-specific environment variables, system settings and profile properties.

We also update the default AWS client builder to set the `ClientEndpointProvider`, so that old client versions get the new CLIENT_ENDPOINT_PROVIDER option set. The default SDK client builder is also updated to use the new provider abstraction for endpoint overrides.

This also includes EndpointSharedConfigTest, which tests to make sure that the new ways of overriding the endpoint work for clients.

[...]

PR #5944, opened 2025-03-10
Adding service endpoint section in config file

RanVaknin:
## Motivation and Context
Support for service-specific endpoint overrides in the Java SDK was introduced in a #5562 , allowing services to resolve endpoints from environment variables and system properties. 
However, service-specific overrides from `[services]` section of the config file was not supported. Called out in #5779.

This PR adds the functionality to provide an endpoint override through  ~/.aws/config file's `[services]` section.

Example:

```
[profile foo]
region = us-east-1
endpoint_url = https://my-general-endpoint.com
services = dev

[profile bar]
region = us-west-2

[services dev]
s3 =
  endpoint_url = https://my-s3-specific-endpoint.com
```

## Modifications

- Added `servicesProperty()` method in `AwsClientEndpointProvider` to extract and resolve service-specific endpoints from the [services] section.
- Updated `clientEndpointFromEnvironment()` to use the new `servicesProperty` in the correct load order. 
- Added `ProfileSection` enum to include a SERVICES section.

## Testing

- Updated parameterized tests in `EndpointSharedConfigTest` to include [services] as a valid source for endpoint resolution.
- Added `ProfileFileServicesTest` class to verify parsing and retrieval of service-specific endpoint settings.


core/aws-core/src/main/java/software/amazon/awssdk/awscore/endpoint/AwsClientEndpointProvider.java
```diff
@@ -129,6 +130,7 @@ private Optional<ClientEndpoint> clientEndpointFromEnvironment(Builder builder)
                                           () -> systemProperty(GLOBAL_ENDPOINT_OVERRIDE_SYSTEM_PROPERTY),
                                           () -> environmentVariable(builder.serviceEndpointOverrideEnvironmentVariable),
                                           () -> environmentVariable(GLOBAL_ENDPOINT_OVERRIDE_ENVIRONMENT_VARIABLE),
+                                          () -> servicesProperty(builder),
                                           () -> profileProperty(builder,
                                                                 builder.serviceProfileProperty + "."
                                                                 + ProfileProperty.ENDPOINT_URL),
```
dagnir, line 136: Let's add a comment here for context about why we support

```
s3 =
    endpoint_url = ...
```

in both `profile` and `services`

dagnir:
Can you kick off integ tests?

RanVaknin:
> Can you kick off integ tests?

Intg tests are passing for 5944.

core/aws-core/src/main/java/software/amazon/awssdk/awscore/endpoint/AwsClientEndpointProvider.java
```diff
@@ -156,6 +166,21 @@ private Optional<URI> profileProperty(Builder builder, String profileProperty) {
...
+    private Optional<URI> servicesProperty(Builder builder) {
+        Optional<ProfileFile> profileFile = Optional.ofNullable(builder.profileFile.get());
+        Optional<String> servicesSectionName = profileFile
+            .flatMap(pf -> pf.profile(builder.profileName))
+            .flatMap(p -> p.property("services"));
+
+        Optional<URI> serviceEndpoint = servicesSectionName
+            .flatMap(name -> profileFile.flatMap(pf -> pf.getSection("services", name)))
+            .flatMap(p -> Optional.ofNullable(p.properties().get(builder.serviceProfileProperty
+                                                                 + "." + ProfileProperty.ENDPOINT_URL)))
```
millems, line 178: You could have used a normal `map` here and not wrap it in `Optional`. 

Is there a reason you're not using `p.property(builder.serviceProfileProperty + "." + ProfileProperty.ENDPOINT_URL)`?

core/aws-core/src/main/java/software/amazon/awssdk/awscore/endpoint/AwsClientEndpointProvider.java
```diff
@@ -156,6 +166,21 @@ private Optional<URI> profileProperty(Builder builder, String profileProperty) {
...
+            .flatMap(uri -> createUri("services section property", Optional.of(uri)));
```
millems, line 179: Wrapping uri in an optional is a weird way of doing it, even though it works. I'd have personally just saved the result of the last method as and then returned createUrl(...). Not blocking for you to make a change.

test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/EndpointSharedConfigTest.java
```diff
@@ -76,6 +76,15 @@ public void resolvesCorrectEndpoint() {
                                   .append("  endpoint_url = ").append(testCase.serviceProfileSetting).append("\n");
             }
 
+            if (testCase.serviceSectionProfileSetting != null) {
+                if (!profileFileContent.toString().contains("services = dev")) {
```
millems, line 80: Why is this if necessary?

RanVaknin: It's not. I was running into an issue with duplication, but it was an unrelated issue. Just an oversight that is safe to remove.

core/aws-core/src/main/java/software/amazon/awssdk/awscore/endpoint/AwsClientEndpointProvider.java
RanVaknin: No particular reason, just unfamiliar with this API. Fixed now.

dagnir (approved):

PR #5949, opened 2025-03-11
Flaky integ test where the SubscribeToShardIntegrationTest

joviegas:
## Motivation and Context

- Test cases fail in @setup() stage
```
< ERROR! java.util.concurrent.CompletionException: software.amazon.awssdk.services.kinesis.model.ResourceInUseException: 
Stream subscribe-to-shard-integ-test-123456789 under account 123456789 
not ACTIVE or UPDATING, instead in state CREATING 
(Service: Kinesis, Status Code: 400, Request ID: ca299ffa-90c6-9006-aa63-1a1ca6eea3b3, Extended Request ID: ABC
```

- The above log indicates that stream is still in CREATING
## Modifications
- Added additional verification by adding a list call to ensure that stream is fully operational

## License


services/kinesis/src/it/java/software/amazon/awssdk/services/kinesis/SubscribeToShardIntegrationTest.java
```diff
@@ -290,8 +291,22 @@ private static void waitForStreamToBeActive() {
...
               .orFailAfter(Duration.ofMinutes(5));
-    }
 
+        // Additional verification to ensure stream is fully operational
+        Waiter.run(() -> {
+                  try {
+                      asyncClient.listShards(r -> r.streamName(streamName)).join();
+                      return true;
+                  } catch (Exception e) {
+                      if (e.getCause() instanceof ResourceInUseException) {
```
L-Applin, line 301: We already have a Waiter waiting for the `ACTIVE` state, but we are failing because the stream is in the `CREATING` stage? That seems weird, are we sure this is helping?

joviegas: Yeap , I have attached the complete log in the internal issue.
The above log is taken from the codebuild logs.

joviegas: The additional check will ensure that the actual state is properly reflected in the system and will provide additional time with retry logic rather than immediately failing(same as we retrying the test case when it fails)
I recommend proceeding with this safe approach since the current failure of this test case is unnecessarily blocking the release and creating significant churn. This solution maintains reliability while addressing our immediate need to unblock the release process

L-Applin (approved):

PR #5948, opened 2025-03-11
Fix flakyness in LoggingTransferListenerTest where LogCaptor captures logs from other parrallely running test cases

joviegas:
## Motivation and Context

LogCaptor events are not filtered before comparing that causes incorrect assertion on logs and irrelevant logs are added in assertions

```
[ERROR] Tests run: 2, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.039 s <<< FAILURE! -- in software.amazon.awssdk.transfer.s3.progress.LoggingTransferListenerTest
[ERROR] software.amazon.awssdk.transfer.s3.progress.LoggingTransferListenerTest.test_customTicksListener_successfulTransfer -- Time elapsed: 0.027 s <<< FAILURE!
org.opentest4j.AssertionFailedError:

expected: "|==== | 80.0%"
but was: "Retrying Request: DefaultSdkHttpFullRequest(httpMethod=PUT, protocol=http, host=localhost, port=42741, encodedPath=/bucket/key, headers=[amz-sdk-invocation-id, Content-encoding, Content-Length, Content-Type, Expect, User-Agent, x-amz-content-sha256, x-amz-decoded-content-length, x-amz-sdk-checksum-algorithm, x-amz-trailer], queryParameters=[partNumber, uploadId])"
at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
at software.amazon.awssdk.transfer.s3.progress.LoggingTransferListenerTest.assertLogged(LoggingTransferListenerTest.java:123)
```
## Modifications

- LogCaptor Events are filtered , to consider only `LoggingTransferListener`
## License

- [x] I confirm that this pull request can be released under the Apache 2 license

PR #5972, opened 2025-03-18
Filter out retry events for progress bar assertion on intg tests

RanVaknin:
Our S3 transfer manager integ test has a progress bar assertion that is intermittently failing due to an unaccounted retry message.

```
[ERROR] software.amazon.awssdk.transfer.s3.progress.LoggingTransferListenerTest.test_customTicksListener_successfulTransfer -- Time elapsed: 0.027 s <<< FAILURE!
org.opentest4j.AssertionFailedError:

expected: "|==== | 80.0%"
but was: "Retrying Request: DefaultSdkHttpFullRequest(httpMethod=PUT, protocol=http, host=localhost, port=42741, encodedPath=/bucket/key, headers=[amz-sdk-invocation-id, Content-encoding, Content-Length, Content-Type, Expect, User-Agent, x-amz-content-sha256, x-amz-decoded-content-length, x-amz-sdk-checksum-algorithm, x-amz-trailer], queryParameters=[partNumber, uploadId])"
```

We can filter out this "Retrying Request" log from the stream so that only the progression bar logs appear.

### DIFF
--- a/services-custom/s3-transfer-manager/src/test/java/software/amazon/awssdk/transfer/s3/progress/LoggingTransferListenerTest.java
+++ b/services-custom/s3-transfer-manager/src/test/java/software/amazon/awssdk/transfer/s3/progress/LoggingTransferListenerTest.java
@@ -62,6 +62,7 @@ class LoggingTransferListenerTest {
             String loggerName = LoggingTransferListener.class.getName();
             List<LogEvent> filteredEvents = events.stream()
                                                   .filter(e -> e.getLoggerName().equals(loggerName))
+                                                  .filter(e -> !e.getMessage().getFormattedMessage().contains("Retrying Request"))
                                                   .collect(Collectors.toList());
             assertLogged(filteredEvents, Level.INFO, "Transfer initiated...");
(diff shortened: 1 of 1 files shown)


RanVaknin:
Integ tests are passing

L-Applin (commented):
Do we know why the request was retrying in the first place?

L-Applin (approved):

RanVaknin:
The test in question tests the `TransferListenerContext` directly, and not uses the underlying transfer manager/ s3 client. Therefore I don’t think the log for the “Retrying Request” we are seeing is actually from this test.
The logger might get a shared scope somehow ([maybe here?](https://github.com/aws/aws-sdk-java-v2/blob/master/test/test-utils/src/main/java/software/amazon/awssdk/testutils/LogCaptor.java#L125)) and the error messages are getting polluted from other s3 tests that are running concurrently.

joviegas:
I think we donot require this fix after https://github.com/aws/aws-sdk-java-v2/pull/5948/

PR #5986, opened 2025-03-24
Migration tool - Date to Instant

davidh44:
## Motivation and Context

v1 takes `Date` args, while v2 takes `Instant` args

## Modifications

Added new recipe `DateToInstant` to append `toInstant()` to `Date` args

## Testing

Added end to end tests

### DIFF
--- a/v2-migration/src/main/java/software/amazon/awssdk/v2migration/DateToInstant.java
+++ b/v2-migration/src/main/java/software/amazon/awssdk/v2migration/DateToInstant.java
@@ -1,0 +1,88 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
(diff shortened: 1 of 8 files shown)


v2-migration/src/main/java/software/amazon/awssdk/v2migration/DateToInstant.java
```diff
@@ -0,0 +1,88 @@
...
+                return false;
+            }
+
+            JavaType javaType = parameterTypes.get(0);
+            if (javaType == null) {
+                return false;
+            }
+
+            boolean isDateParam = javaType.isAssignableFrom(DATE_PATTERN);
+            return SdkTypeUtils.isV2ModelBuilder(declaringType) && isDateParam;
```
Fred1155, line 85: Do we have special cases where Date objects should NOT be transformed?

davidh44: No Date in v2, will all be Instant afaik. At least for POJO setters

Fred1155 (approved):

PR #5990, opened 2025-03-25
Migration Tool AmazonS3URI transform

davidh44:
## Motivation and Context

Transform v1 `AmazonS3URI` to v2 `S3Uri`

https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/migration-s3-uri-parser.html

## Modifications

new recipe `S3UriToV2`

## Testing

added end to end testing

### DIFF
--- a/test/v2-migration-tests/src/test/resources/software/amazon/awssdk/v2migrationtests/maven/after/src/main/java/foo/bar/S3.java
+++ b/test/v2-migration-tests/src/test/resources/software/amazon/awssdk/v2migrationtests/maven/after/src/main/java/foo/bar/S3.java
@@ -15,9 +15,13 @@
 
 package foo.bar;
 
+import java.net.URI;
 import java.util.ArrayList;
 import java.util.List;
+import software.amazon.awssdk.regions.Region;
 import software.amazon.awssdk.services.s3.S3Client;
+import software.amazon.awssdk.services.s3.S3Uri;
+import software.amazon.awssdk.services.s3.S3Utilities;
 import software.amazon.awssdk.services.s3.model.AbortMultipartUploadRequest;
 import software.amazon.awssdk.services.s3.model.AccelerateConfiguration;
 import software.amazon.awssdk.services.s3.model.AnalyticsConfiguration;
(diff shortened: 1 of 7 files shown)


test/v2-migration-tests/src/test/resources/software/amazon/awssdk/v2migrationtests/maven/after/src/main/java/foo/bar/S3.java
```diff
@@ -384,4 +388,18 @@ private void setBucketNameTest(S3Client s3, String bucket) {
...
+    private void s3Uri(URI uri, String uriAsString) {
+        S3Uri s3Uri = S3Utilities.builder().build().parseUri(uri);
+
+        String versionId = s3Uri.firstMatchingRawQueryParameter("versionId").orElse(null);
+        String bucket = s3Uri.bucket().orElse(null);
+        String key = s3Uri.key().orElse(null);
+        String region = s3Uri.region().map(Region::id).orElse(null);
+        boolean isPathStyle = s3Uri.isPathStyle();
+
+        S3Uri s3UriFromString = /*AWS SDK for Java v2 migration: v2 S3Uri does not URL-encode a String URI. If you relied on this functionality in v1 you must update your code to manually encode the String.*/S3Utilities.builder().build().parseUri(URI.create(uriAsString));
```
L-Applin, line 401: the Comment is kind of in a weird spot, is that what we do elsewhere for similar warning in migration tool?

davidh44: Yeah Dongie had same concern. We can't maintain the spacing if put on separate line though. Also for this specific transform can't do it on the line before since we transform the left and right hand sides separately. We could possibly do it on the next line. On the other hand, having it here makes it really obvious to the user

We do the same for other comments fwiw
https://github.com/aws/aws-sdk-java-v2/blob/master/v2-migration/src/test/java/software/amazon/awssdk/v2migration/AddCommentToMethodTest.java#L51

L-Applin: Yeah, it's great that it is really obvious, just a bit weird. I don't mind it.

PR #5993, opened 2025-03-26
Providing an idle publisher to resolve race condition

RanVaknin:
## Motivation and Context
This test had two separate race conditions: 

1. the Transformer's `exceptionOccurred` function and the Subscriber's `failed()` method both race to complete the completeablefuture exceptionally resulting in:

```
java.lang.AssertionError: 
Expecting a cause with type:
  "java.lang.RuntimeException"
and message:
  "oops"
but type was:
  "java.nio.channels.ClosedChannelException"
and message was:
  null.
```

This can be solved by making the assertion more relaxed. This test is meant to test the delete file behavior, and not the exception handling. I believe this exception assertion that was added in the `stubException` helper function was to enforce the contract (an exception is propagated when the file transformer is terminated). This race condition caused that the specific exception could not be guaranteed. 

2. The FileSubscriber's `preformWrite` finished writing before the Transformer's `exceptionOccurred` handler was able to complete the future exceptionally:

```
java.lang.AssertionError: 
Expecting
  <CompletableFuture[Completed: "foobar"]>
to have failed within 1L SECONDS.
Be aware that the state of the future in this message might not reflect the one at the time when the assertion was performed as it is evaluated later on
```

The fix replaces the original `SdkPublisher` that emits data with `Flowable.just(content, content)` with an idle publisher that emits data but never completes. 
The original publisher would emit and immediately complete, creating a race between normal completion and exception handling. The idle publisher keeps the stream open, while the exception handling path has time to execute before completion.

## Testing
1. Created a `@RepeatedTest` wrapper with a 1000 executions to check for failure rates. The test was failing 0.4-1.2% of tests. After the fix I see a 0% failure rate. 

2. Added logging to each handler of the FileSubscriber and the Transformer to observe the race conditions. For example:

```
transformer calling onSubscribe
Transformer's exceptionOccurred start block
Transformer's exceptionOccurred: closing the channel
Transformer's exceptionOccurred: cleanup
transformer completed method is called
File Subscriber failed, future.completeExceptionally(exc) is called: java.nio.channels.ClosedChannelException
Transofmer exceptionOccured completes the future exceptionally

java.lang.AssertionError: 
Expecting a cause with type:
  "java.lang.RuntimeException"
and message:
  "oops"
but type was:
  "java.nio.channels.ClosedChannelException"
and message was:
  null.
```

### DIFF
--- a/core/sdk-core/src/test/java/software/amazon/awssdk/core/internal/async/FileAsyncResponseTransformerTest.java
+++ b/core/sdk-core/src/test/java/software/amazon/awssdk/core/internal/async/FileAsyncResponseTransformerTest.java
@@ -47,6 +47,7 @@ import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.MethodSource;
+import org.reactivestreams.Subscriber;
 import org.reactivestreams.Subscription;
 import software.amazon.awssdk.core.FileTransformerConfiguration;
 import software.amazon.awssdk.core.FileTransformerConfiguration.FileWriteOption;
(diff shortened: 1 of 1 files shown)


core/sdk-core/src/test/java/software/amazon/awssdk/core/internal/async/FileAsyncResponseTransformerTest.java
```diff
@@ -331,12 +332,25 @@ private static void stubException(String newContent, FileAsyncResponseTransforme
 
         RuntimeException runtimeException = new RuntimeException("oops");
         ByteBuffer content = ByteBuffer.wrap(newContent.getBytes(StandardCharsets.UTF_8));
-        transformer.onStream(SdkPublisher.adapt(Flowable.just(content, content)));
+        SdkPublisher<ByteBuffer> idlePublisher = new SdkPublisher<ByteBuffer>() {
+            @Override
+            public void subscribe(Subscriber<? super ByteBuffer> subscriber) {
+                subscriber.onSubscribe(new Subscription() {
```
L-Applin, line 338: I'd rather use a `NoopSubscription` like we discussed. Here the content is still sent to the transformer by calling `subscriber.onNext()`, which might make the issue reappear.

core/sdk-core/src/test/java/software/amazon/awssdk/core/internal/async/FileAsyncResponseTransformerTest.java
```diff
@@ -331,12 +332,25 @@ private static void stubException(String newContent, FileAsyncResponseTransforme
...
+                });
+            }
+        };
+        transformer.onStream(idlePublisher);
         transformer.exceptionOccurred(runtimeException);
 
-        assertThat(future).failsWithin(1, TimeUnit.SECONDS)
-            .withThrowableOfType(ExecutionException.class)
-            .withCause(runtimeException);
+        assertThatThrownBy(future::join).isInstanceOf(Exception.class);
```
L-Applin, line 353: With a new publisher that wont cause problems with unexpected file write errors, I think we can keep the  previous assertions as-is.

L-Applin (approved):

PR #5998, opened 2025-03-31
Add all streaming operation to s3 checksums tests

L-Applin:
Add more test cases to s3 checksums tests:
- Add all method from `RequestBody` and `ResponseTransformer` to the test suite. 
- Add multipart download/upload.
- Add Transfer-manager based tests.

Prepare for Codebuild execution for the tests suite


test/s3-tests/src/it/java/software/amazon/awssdk/services/s3/regression/UploadStreamingRegressionTesting.java
```diff
@@ -0,0 +1,753 @@
...
+            case INPUTSTREAM_RESETABLE: {
+                byte[] content = contentSize.byteContent();
+                AsyncRequestBody asyncRequestBody = AsyncRequestBody.fromInputStream(new ByteArrayInputStream(content),
+                                                                                     (long) content.length,
+                                                                                     ASYNC_REQUEST_BODY_EXECUTOR);
+                return new TestAsyncBody(asyncRequestBody, content.length, crc32(content), bodyType);
+            }
+            case INPUTSTREAM_NOT_RESETABLE: {
+                byte[] content = contentSize.byteContent();
+                AsyncRequestBody asyncRequestBody = AsyncRequestBody.fromInputStream(new NonResettableByteStream(content),
```
zoewangg, line 362: Can we add a test for INPUTSTREAM_NO_LENGTH?

test/s3-tests/src/it/java/software/amazon/awssdk/services/s3/regression/UploadStreamingRegressionTesting.java
```diff
@@ -0,0 +1,753 @@
...
+                                         bodyType);
+            }
+            case BUFFERS_UNSAFE: {
+                byte[] content1 = contentSize.byteContent();
+                byte[] content2 = contentSize.byteContent();
+                AsyncRequestBody asyncRequestBody = AsyncRequestBody.fromByteBuffersUnsafe(ByteBuffer.wrap(content1),
+                                                                                           ByteBuffer.wrap(content2));
+                byte[] crcArray = new byte[content2.length + content2.length];
+                System.arraycopy(content1, 0, crcArray, 0, content1.length);
+                System.arraycopy(content2, 0, crcArray, content1.length, content2.length);
```
zoewangg, line 445: If we do this once in set up, would it accelerate the test?

pom.xml
```diff
@@ -876,6 +876,52 @@
             </build>
         </profile>
 
+        <profile>
```
zoewangg, line 879: Can we add a README? Something like what we have for stability tests will do https://github.com/aws/aws-sdk-java-v2/tree/master/test/stability-tests

test/s3-tests/src/it/java/software/amazon/awssdk/services/s3/regression/UploadStreamingRegressionTesting.java
```diff
@@ -0,0 +1,753 @@
...
+            && (config.payloadSigning()
+                // MRAP requires body signing
+                || config.getBaseConfig().getBucketType() == BucketType.MRAP),
+            "Async payload signing doesn't work with Java based clients");
+
+        // For testing purposes, ContentProvider is Publisher<ByteBuffer> for async clients
+        // Async java based clients don't currently support unknown content-length bodies
+        Assumptions.assumeFalse(
+            (config.getBaseConfig().getFlavor() == S3ClientFlavor.ASYNC_JAVA_BASED ||
+             config.getBaseConfig().getFlavor() == S3ClientFlavor.TM_JAVA)
```
zoewangg, line 126: This should be supported in multipart client, right?

zoewangg: Do we have a test case covering multipart client + unknown content-length?

test/s3-tests/src/it/java/software/amazon/awssdk/services/s3/regression/DownloadStreamingRegressionTesting.java
```diff
@@ -0,0 +1,616 @@
...
+        ByteArrayOutputStream os = new ByteArrayOutputStream();
+        byte[] rand = new byte[8 * 1024 * 1024];
+        for (int i = 0; i < 10; i++) {
+            new Random().nextBytes(rand);
+            os.write(rand);
+        }
+        byte[] fullContent = os.toByteArray();
+        String crc32 = crc32(fullContent);
+        for (BucketType bucketType : BucketType.values()) {
+            doMultipartUpload(bucketType, name, fullContent, crc32);
```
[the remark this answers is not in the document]
zoewangg: We could technically use v1 TM if we want to completely avoid v2 ;) Just trying to see if we can speed up the test.

test/s3-tests/src/it/java/software/amazon/awssdk/services/s3/regression/S3ChecksumsTestUtils.java
```diff
@@ -0,0 +1,375 @@
...
+    private static final SdkChecksum CRC32 = SdkChecksum.forAlgorithm(DefaultChecksumAlgorithm.CRC32);
+
+    private S3ChecksumsTestUtils() {
+    }
+
+    public static String createBucket(S3Client s3, String name, Logger log) {
+        log.debug(() -> "Creating bucket: " + name);
+        createBucket(s3, name, 3, log);
+        s3.putBucketAccelerateConfiguration(r -> r.bucket(name)
+                                                  .accelerateConfiguration(c -> c.status(BucketAccelerateStatus.ENABLED)));
```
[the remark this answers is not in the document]
zoewangg: @dagnir any ideas? 

I'm leaning towards keeping the default config (don't think we need to add bucket accelerate status to the test config since there's no customization code for it).

test/s3-tests/src/it/java/software/amazon/awssdk/services/s3/regression/DownloadStreamingRegressionTesting.java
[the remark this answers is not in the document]
L-Applin: it won't make a big difference on test duration, it is only uploaded once per whole test run (`@BeforeAll` )

test/s3-tests/src/it/java/software/amazon/awssdk/services/s3/regression/UploadStreamingRegressionTesting.java
L-Applin: Yes, we have `INPUTSTREAM_NO_LENGTH` which uses `AsyncRequestBody
                    .fromInputStream(AsyncRequestBodyFromInputStreamConfiguration)` without length and `CONTENT_PROVIDER_NO_LENGTH` which uses `AsyncRequestBody.fromPublisher(Publisher)` which are used with both multipart and single part

test/s3-tests/README.md
```diff
@@ -0,0 +1,31 @@
...
+```
+mvn clean install -P s3-regression-tests -pl :stability-tests
+```
+
+## Adding New Tests
+
+- The tests are built using [JUnit 5](https://junit.org/junit5/). Make sure you are using the correct APIs and mixing of
+  Junit 4 and Junit 5 APIs on the same test can have unexpected results.
+
+- All tests should have the suffix of `RegressionTesting`, eg: ``
```
zoewangg: Missing example?

test/s3-tests/src/test/resources/log4j2.xml
```diff
@@ -24,6 +24,7 @@
         <Root level="WARN">
             <AppenderRef ref="ConsoleAppender"/>
         </Root>
+        <Logger name="software.amazon.awssdk.services.s3.regression" level="DEBUG" />
```
zoewangg: nit: maybe just changing root level to INFO and making those logs in this class INFO so that we don't end up printing tons of logs?

zoewangg (approved):

zoewangg (approved):

PR #6012, opened 2025-04-07
Lowering ack callback to avoid race condition

RanVaknin:
## Motivation and Context
This PR addresses an intermittent test failure in `Http2PingHandlerTest.delayedPingFlushDoesntTerminateConnectionPrematurely()`. The test was occasionally failing with a PingFailedException due to timing issues between ping scheduling, and delayed writes. The test verifies that a delayed ping flush doesn't prematurely terminate the connection, but the race condition in the test setup was causing inconsistent results.

## Testing
Even though this test failed intermittently in our CI builds, I ran the test locally with `@RepeatedTest(1000)` before making changes and observed a 100% success rate locally.
While this wont deterministically solve the issue, the change aims to make these failures even more rare by reducing the response time window.

### DIFF
--- a/http-clients/netty-nio-client/src/test/java/software/amazon/awssdk/http/nio/netty/internal/http2/Http2PingHandlerTest.java
+++ b/http-clients/netty-nio-client/src/test/java/software/amazon/awssdk/http/nio/netty/internal/http2/Http2PingHandlerTest.java
@@ -238,7 +238,7 @@ public class Http2PingHandlerTest {
         EmbeddedChannel channel = createHttp2Channel(fastChecker, catcher, pingResponder, delayingWriter);
 
         pingResponder.setCallback(() -> channel.writeInbound(new DefaultHttp2PingFrame(0, true)),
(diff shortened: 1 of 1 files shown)


davidh44 (approved):
Just to check, we don't see any failures with `schedulingDelayDoesNotCausePingTimeout()` and don't need to make a similar change there?

https://github.com/aws/aws-sdk-java-v2/blob/master/http-clients/netty-nio-client/src/test/java/software/amazon/awssdk/http/nio/netty/internal/http2/Http2PingHandlerTest.java#L125schedulingDelayDoesNotCausePingTimeout()

RanVaknin:
I re-scanned our logs and in the last 10 we saw 9 total failures for `Http2PingHandlerTest`, all failures only occurred in `delayedPingFlushDoesntTerminateConnectionPrematurely. I decided to add a retryable annotation because Iam less confident now in lowering the callback.

davidh44 (approved):

PR #6019, opened 2025-04-09
DDB Enhanced: Allow custom versioning

RanVaknin:
## Motivation and Context
This PR builds on the work started in issue #3894 where the `VersionedRecordExtension` didn't support starting version numbers at 0, requiring clients to use Integer objects (with null initial values) rather than long primitives. As mentioned by @akiesler in the original issue, developers might expect versions to start at 0 and increment from there, rather than having a special case where the value must be initialized to null.

This implementation allows the extension to be more flexible by allowing:
- Starting versions at 0 (or any non-negative long)
- Configuring custom increment values (only positive numbers)
- Supporting these configs both at the extension level and via annotations

## Modifications
- Refactored the `VersionedRecordExtension` to support explicit `startAt` and `incrementBy` values via builder methods (making it opt-in).
- Expanded the `DynamoDbVersionAttribute` annotation to support `startAt` and `incrementBy` parameters
- Added validation to prevent negative startAt values and non-positive incrementBy values

## Testing
  - Custom `startAt` and `incrementBy` values through both builder and annotations
  - Validation of illegal values (negative `startAt`, zero/negative `incrementBy`)
  - Precedence rules between annotation and builder configurations
  - Version incrementation for both initial and existing records
  - Edge cases with different configuration combinations


services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/Expression.java
```diff
@@ -311,8 +311,17 @@ public int hashCode() {
...
+        return "Expression{" +
+               "expression='" + expression + '\'' +
+               ", expressionValues=" + expressionValues +
+               ", expressionNames=" + expressionNames +
+               '}';
+    }
+
     /**
-     * A builder for {@link Expression}
+     * A builder for {@link Expression}v
```
davidh44, line 324:
```suggestion
     * A builder for {@link Expression}
```

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/extensions/VersionedRecordExtension.java
```diff
@@ -75,19 +88,50 @@ private AttributeTags() {
...
+                throw new IllegalArgumentException("StartAt cannot be negative.");
+            }
+
+            if (incrementBy != null && incrementBy < 1) {
+                throw new IllegalArgumentException("IncrementBy must be greater than 0.");
             }
 
-            return metadata -> metadata.addCustomMetadataObject(CUSTOM_METADATA_KEY, attributeName)
+            return metadata -> metadata
+                                       .addCustomMetadataObject(CUSTOM_METADATA_KEY, attributeName)
```
davidh44, line 132:
```suggestion
            return metadata -> metadata.addCustomMetadataObject(CUSTOM_METADATA_KEY, attributeName)
```

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/extensions/VersionedRecordExtension.java
```diff
@@ -142,13 +159,131 @@ public WriteModification beforeWrite(DynamoDbExtensionContext.BeforeWrite contex
                                 .build();
     }
 
+    private Pair<AttributeValue, Expression> getRecordUpdates(String versionAttributeKey,
+                                                              DynamoDbExtensionContext.BeforeWrite context) {
+        Map<String, AttributeValue> itemToTransform = context.items();
+
+        // Default to NUL if not present to reduce additional checks further along
```
davidh44, line 166:
```suggestion
        // Default to NULL if not present to reduce additional checks further along
```

services-custom/dynamodb-enhanced/pom.xml
```diff
@@ -76,6 +76,20 @@
...
                 </configuration>
             </plugin>
+            <plugin>
+                <groupId>com.github.siom79.japicmp</groupId>
+                <artifactId>japicmp-maven-plugin</artifactId>
+                <version>${japicmp-maven-plugin.version}</version>
+                <configuration>
+                    <parameter>
+                        <excludes>
+                            <!-- japicmp flagging these incorrectly as a breaking change -->
```
zoewangg, line 86: Hmm, this is odd, can you paste the error logs? If we have to add exclusion, we should add it to the top level pom https://github.com/aws/aws-sdk-java-v2/blob/master/pom.xml#L681

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/extensions/VersionedRecordExtension.java
```diff
@@ -75,19 +88,50 @@ private AttributeTags() {
...
         public Consumer<StaticTableMetadata.Builder> modifyMetadata(String attributeName,
                                                                     AttributeValueType attributeValueType) {
             if (attributeValueType != AttributeValueType.N) {
                 throw new IllegalArgumentException(String.format(
                     "Attribute '%s' of type %s is not a suitable type to be used as a version attribute. Only type 'N' " +
-                        "is supported.", attributeName, attributeValueType.name()));
+                    "is supported.", attributeName, attributeValueType.name()));
+            }
+
+            if (startAt != null && startAt < 0) {
```
zoewangg, line 123: nit: we can use `Validate.isNotNegativeOrNull`

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/extensions/VersionedRecordExtension.java
```diff
@@ -142,13 +159,131 @@ public WriteModification beforeWrite(DynamoDbExtensionContext.BeforeWrite contex
...
+                                                                                  Integer.class);
+        if (versionIncrementByFromAnnotation.isPresent()) {
+            return AttributeValue.fromN(Integer.toString(version + versionIncrementByFromAnnotation.get()));
+        }
+        return AttributeValue.fromN(Integer.toString(version + this.incrementBy));
+    }
+
     @NotThreadSafe
     public static final class Builder {
+        private int startAt = 0;
```
zoewangg, line 249: Can we define the default value in the outer class instead of builder? IMO builder class should not have any logic

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/extensions/VersionedRecordExtension.java
```diff
@@ -142,13 +159,131 @@ public WriteModification beforeWrite(DynamoDbExtensionContext.BeforeWrite contex
...
         }
 
+        /**
+         * Sets the startAt used to compare if a record is the initial version of a record.
+         * Default value - {@code 0}.
+         *
+         * @param startAt
+         * @return the builder instance
+         */
+        public Builder startAt(int startAt) {
```
zoewangg, line 262: We should use boxed type

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/extensions/VersionedRecordExtension.java
```diff
@@ -101,39 +145,12 @@ public WriteModification beforeWrite(DynamoDbExtensionContext.BeforeWrite contex
...
-                                  .expression(String.format("attribute_not_exists(%s)", attributeKeyRef))
-                                  .expressionNames(Collections.singletonMap(attributeKeyRef, versionAttributeKey.get()))
-                                  .build();
-        } else {
-            // Existing record, increment version
-            if (existingVersionValue.get().n() == null) {
-                // In this case a non-null version attribute is present, but it's not an N
-                throw new IllegalArgumentException("Version attribute appears to be the wrong type. N is required.");
-            }
+        Pair<AttributeValue, Expression> updates = getRecordUpdates(versionAttributeKey.get(), context);
```
zoewangg, line 148: Refactoring (extracting code to a separate method) is great, but it could make it a bit hard to review the new change. Can we remove the refactoring from this commit and add it to a separate commit, if possible so that we can review the new change?

zoewangg (changes requested):

akiesler:
Thank you for picking up this change! It will greatly simplify our version handling. Please let me know if I can do anything to help.

services-custom/dynamodb-enhanced/pom.xml
RanVaknin: the error only was shown on CI, but I was able to raise with running `mvn japicmp:cmp -pl :dynamodb-enhanced`
```
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  4.557 s
[INFO] Finished at: 2025-04-21T11:29:14-07:00
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal com.github.siom79.japicmp:japicmp-maven-plugin:0.15.6:cmp (default-cli) on project dynamodb-enhanced: There is at least one incompatibility: software.amazon.awssdk.enhanced.dynamodb.extensions.annotations.DynamoDbVersionAttribute.incrementBy():METHOD_ABSTRACT_ADDED_TO_CLASS,software.amazon.awssdk.enhanced.dynamodb.extensions.annotations.DynamoDbVersionAttribute.startAt():METHOD_ABSTRACT_ADDED_TO_CLASS -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoFailureException
```

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/extensions/VersionedRecordExtension.java
RanVaknin: The refactoring here was done by the person who  opened the [original PR](https://github.com/aws/aws-sdk-java-v2/pull/5565/files). I just ported their changes over.
I’m going to end up reverting those changes because IMO it makes it harder to read.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/extensions/VersionedRecordExtension.java
```diff
@@ -123,9 +178,12 @@ public WriteModification beforeWrite(DynamoDbExtensionContext.BeforeWrite contex
                 throw new IllegalArgumentException("Version attribute appears to be the wrong type. N is required.");
             }
 
-            int existingVersion = Integer.parseInt(existingVersionValue.get().n());
+            long existingVersion = Long.parseLong(existingVersionValue.get().n());
             String existingVersionValueKey = VERSIONED_RECORD_EXPRESSION_VALUE_KEY_MAPPER.apply(versionAttributeKey.get());
-            newVersionValue = AttributeValue.builder().n(Integer.toString(existingVersion + 1)).build();
+
+            long increment = versionIncrementByFromAnnotation.orElse(this.incrementBy);
+            newVersionValue = AttributeValue.builder().n(Long.toString(existingVersion + increment)).build();
```
dagnir, line 185: It doesn't look like the original version handled this, but how do we deal with integer overflow? Originally `incrementBy` was always `1` so it's unlikely to happen, but with `incrementBy` being configurable now, it's technically more likely now if the user sets a relatively large value.

Can we check how v1's mapper handles this (if at all)?

dagnir (changes requested):

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/extensions/VersionedRecordExtension.java
```diff
@@ -144,11 +202,38 @@ public WriteModification beforeWrite(DynamoDbExtensionContext.BeforeWrite contex
 
     @NotThreadSafe
     public static final class Builder {
+        private Long startAt = 0L;
```
zoewangg, line 205: Is there any reason we are setting the default values in Builder? Builder class should not have any logic ideally and it also seems we are already setting the default values in the top level classes

services-custom/dynamodb-enhanced/src/test/java/software/amazon/awssdk/enhanced/dynamodb/extensions/VersionedRecordExtensionTest.java
```diff
@@ -168,15 +178,344 @@ public void beforeWrite_returnsNoOpModification_ifVersionAttributeNotDefined() {
...
             DefaultDynamoDbExtensionContext.builder()
-                                           .items(fakeItemWIthBadVersion)
+                                           .items(fakeItemWithBadVersion)
                                            .operationContext(PRIMARY_CONTEXT)
                                            .tableMetadata(FakeItem.getTableMetadata())
                                            .build());
     }
+
+    @Test
+    public void beforeWrite_versionEqualsStartAt_treatedAsInitialVersion() {
```
zoewangg, line 194: Can we add tests for different table schemas such as static table schema, immutable table scheme and bean table schema? https://github.com/aws/aws-sdk-java-v2/pull/6067/files#diff-9826429ca566bd1325f9ffa5d4fe1141cd26df01ec67262bab6f4acb12c1aa28

zoewangg (changes requested):

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/extensions/VersionedRecordExtension.java
RanVaknin: Doesn't seem like v1 had any sort of integer overflow protection? 

https://github.com/aws/aws-sdk-java/blob/c34aa7f14d7598f033091b7b9be96ee82d2c3860/aws-java-sdk-dynamodb/src/main/java/com/amazonaws/services/dynamodbv2/datamodeling/DynamoDBMapper.java#L2454-L2470

dagnir: Thanks for looking! I'd still like us to add protection for 2.x

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/Expression.java
```diff
@@ -311,6 +312,20 @@ public int hashCode() {
         return result;
     }
 
+    @Override
+    public String toString() {
+        // return "Expression{" +
+        //        "expression='" + expression + '\'' +
+        //        ", expressionValues=" + expressionValues +
+        //        ", expressionNames=" + expressionNames +
+        //        '}';
```
dagnir, line 321: Delete

dagnir (changes requested):

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/extensions/VersionedRecordExtension.java
```diff
@@ -142,13 +194,55 @@ public WriteModification beforeWrite(DynamoDbExtensionContext.BeforeWrite contex
                                 .build();
     }
 
+    private boolean isInitialVersion(Optional<AttributeValue> existingVersionValue, Optional<Long> versionStartAtFromAnnotation) {
```
zoewangg, line 197: Use of Optional as method parameter is a bit code smell. https://github.com/aws/aws-sdk-java-v2/blob/master/docs/design/UseOfOptional.md

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/extensions/VersionedRecordExtension.java
zoewangg: Oh I see, I completely missed this. So just to double check, in order to make it backwards compatible, the new behavior should be like the following:
- if startAt is not configured: we only add version if the version attribute does not exist or is null
- if startAt is configured and the version attribute is equal to startAt: we will treat it as initial version

Is that correct?

services-custom/dynamodb-enhanced/src/test/java/software/amazon/awssdk/enhanced/dynamodb/extensions/VersionedRecordExtensionTest.java
```diff
@@ -18,19 +18,31 @@
...
 import software.amazon.awssdk.enhanced.dynamodb.functionaltests.models.FakeItemWithSort;
+import software.amazon.awssdk.enhanced.dynamodb.functionaltests.models.FakeVersionedImmutableItem;
+import software.amazon.awssdk.enhanced.dynamodb.functionaltests.models.FakeVersionedStaticImmutableItem;
 import software.amazon.awssdk.enhanced.dynamodb.internal.extensions.DefaultDynamoDbExtensionContext;
 import software.amazon.awssdk.enhanced.dynamodb.internal.operations.DefaultOperationContext;
+import software.amazon.awssdk.enhanced.dynamodb.mapper.annotations.DynamoDbBean;
+import software.amazon.awssdk.enhanced.dynamodb.mapper.annotations.DynamoDbPartitionKey;
 import software.amazon.awssdk.services.dynamodb.model.AttributeValue;
 
 public class VersionedRecordExtensionTest {
```
zoewangg, line 48: Can we add functional tests as well that tests end to end functionality with dynamodb local? https://github.com/aws/aws-sdk-java-v2/blob/master/services-custom/dynamodb-enhanced/src/test/java/software/amazon/awssdk/enhanced/dynamodb/functionaltests/VersionedRecordTest.java

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/extensions/VersionedRecordExtension.java
RanVaknin: @zoewangg yes you are correct.

services-custom/dynamodb-enhanced/src/test/java/software/amazon/awssdk/enhanced/dynamodb/extensions/VersionedRecordExtensionTest.java
RanVaknin: done

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/extensions/VersionedRecordExtension.java
zoewangg: Any updates on overflow protection? +1 to what Dongie mentioned, it's probably better to fail explicitly with a clear error message for that case

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/extensions/VersionedRecordExtension.java
```diff
@@ -75,19 +87,47 @@ private AttributeTags() {
...
                 throw new IllegalArgumentException(String.format(
                     "Attribute '%s' of type %s is not a suitable type to be used as a version attribute. Only type 'N' " +
-                        "is supported.", attributeName, attributeValueType.name()));
+                    "is supported.", attributeName, attributeValueType.name()));
+            }
+
+            Validate.isNotNegativeOrNull(startAt, "startAt");
+
+            if (incrementBy != null && incrementBy < 1) {
+                throw new IllegalArgumentException("IncrementBy must be greater than 0.");
```
zoewangg: Do we need additional validations in this class? If so, can we move it to ctor

RanVaknin: This validation is for the annotation. The constructor validation only applies to setting those startAt and incrementBy values from the builder.

I actually have a test case covering this. If I were to remove these validations Id see this test case failing:
`invalidAnnotationValues_shouldThrowException()` with:

```

org.opentest4j.AssertionFailedError: Expected java.lang.IllegalArgumentException to be thrown, but nothing was thrown.

	at org.junit.jupiter.api.AssertionFailureBuilder.build(AssertionFailureBuilder.java:152)
	at org.junit.jupiter.api.AssertThrows.assertThrows(AssertThrows.java:73)
	at org.junit.jupiter.api.AssertThrows.assertThrows(AssertThrows.java:35)
	at org.junit.jupiter.api.Assertions.assertThrows(Assertions.java:3111)
	at software.amazon.awssdk.enhanced.dynamodb.extensions.VersionedRecordExtensionTest.invalidAnnotationValues_shouldThrowException(VersionedRecordExtensionTest.java:478)
	at java.base/java.lang.reflect.Method.invoke(Method.java:569)


```

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/extensions/VersionedRecordExtension.java
RanVaknin: Added overflow protection + test case.

RanVaknin:
Fixed edge case in the `isInitialVersion` function. Previously it was not evaluating the precedence that annotation values should take over builder values. 

This caused inconsistent behavior when both annotation and builder values were both provided with different values, potentially treating records as initial versions incorrectly when their version matched the builder's startAt value but not the annotation's value.

This is now fixed. Added a test case that combines input from both the builder and the annotation.

services-custom/dynamodb-enhanced/src/test/java/software/amazon/awssdk/enhanced/dynamodb/extensions/VersionedRecordExtensionTest.java
```diff
@@ -168,15 +180,465 @@ public void beforeWrite_returnsNoOpModification_ifVersionAttributeNotDefined() {
     @Test(expected = IllegalArgumentException.class)
     public void beforeWrite_throwsIllegalArgumentException_ifVersioPnAttributeIsWrongType() {
         FakeItem fakeItem = createUniqueFakeItem();
-        Map<String, AttributeValue> fakeItemWIthBadVersion =
+        Map<String, AttributeValue> fakeItemWithBadVersion =
```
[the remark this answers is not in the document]
RanVaknin: ... ok?

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/extensions/VersionedRecordExtension.java
```diff
@@ -75,19 +87,47 @@ private AttributeTags() {
...
         public Consumer<StaticTableMetadata.Builder> modifyMetadata(String attributeName,
                                                                     AttributeValueType attributeValueType) {
             if (attributeValueType != AttributeValueType.N) {
                 throw new IllegalArgumentException(String.format(
                     "Attribute '%s' of type %s is not a suitable type to be used as a version attribute. Only type 'N' " +
-                        "is supported.", attributeName, attributeValueType.name()));
+                    "is supported.", attributeName, attributeValueType.name()));
+            }
+
+            Validate.isNotNegativeOrNull(startAt, "startAt");
```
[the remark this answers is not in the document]
RanVaknin: isNotNegativeOrNull exist in Utils Validate

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/extensions/VersionedRecordExtension.java
```diff
@@ -109,9 +149,24 @@ public WriteModification beforeWrite(DynamoDbExtensionContext.BeforeWrite contex
...
+
+        Optional<Long> versionIncrementByFromAnnotation = context.tableMetadata()
+                                                                 .customMetadataObject(VersionAttribute.INCREMENT_BY_METADATA_KEY,
+                                                                                       Long.class);
+
+        if (!existingVersionValue.isPresent() || isNullAttributeValue(existingVersionValue.get()) ||
+            (existingVersionValue.get().n() != null &&
+             ((versionStartAtFromAnnotation.isPresent() &&
+               Long.parseLong(existingVersionValue.get().n()) == versionStartAtFromAnnotation.get()) ||
+              Long.parseLong(existingVersionValue.get().n()) == this.startAt))) {
```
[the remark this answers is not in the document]
RanVaknin: Fixed edge case in the isInitialVersion function. Previously it was not evaluating the precedence that annotation values should take over builder values.

This caused inconsistent behavior when both annotation and builder values were both provided with different values, potentially treating records as initial versions incorrectly when their version matched the builder's startAt value but not the annotation's value.

This is now fixed. Added a test case that combines input from both the builder and the annotation.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/extensions/VersionedRecordExtension.java
```diff
@@ -106,31 +146,53 @@ public WriteModification beforeWrite(DynamoDbExtensionContext.BeforeWrite contex
...
             }
 
-            int existingVersion = Integer.parseInt(existingVersionValue.get().n());
+            long existingVersion = Long.parseLong(existingVersionValue.n());
             String existingVersionValueKey = VERSIONED_RECORD_EXPRESSION_VALUE_KEY_MAPPER.apply(versionAttributeKey.get());
-            newVersionValue = AttributeValue.builder().n(Integer.toString(existingVersion + 1)).build();
+
+            long increment = versionIncrementByFromAnnotation;
+
+            /*
```
zoewangg, line 179: nit: the convention for inline comment is `//` rather than `/*`

zoewangg (approved):

dagnir (approved):

PR #6031, opened 2025-04-14
Add duplicateAndRenameSharedEvents customization to support event shapes shared with multiple evenstreams

alextwoods:
Add processor to detect shared events + customization to duplicate.

## Motivation and Context
When an event shape is shared between multiple eventstreams, it causes SDK generation/compilation failures.  The top level shape POJO implements the event stream interface for each stream and the return type of the `sdkEventType` method conflicts.  

## Modifications
This PR adds two things:
1. A new customization config (`duplicateAndRenameSharedEvents`)
2. A new processor (`EventStreamSharedEventProcessor`) which applies the customization and raises helpful error messages if shared events are detected.

## Testing
Added new unit tests to cover processor/config.  Manual testing of internal service model that raised this issue.

N/A

## Types of changes

- [ x ] New feature (non-breaking change which adds functionality)

## Checklist

- [ x] I have added tests to cover my changes

## License


codegen/src/test/resources/software/amazon/awssdk/codegen/emitters/customizations/processors/uselegacyeventgenerationscheme/service-2.json
```diff
@@ -452,14 +452,23 @@
...
-          "shape": "InputEvent"
+          "shape": "InputEventOne"
         },
         "InputEventTwo": {
           "shape": "InputEventTwo"
         }
       },
       "eventstream": true
     },
+    "InputEventOne": {
```
alextwoods, line 463: There are a number of existing test models which have shared events like this - They don't cause issues in the current test because they aren't used to generate/build a full SDK.

We have a few options, we can:
1.  Modify the test model to duplicate the shape (as is done here).
2. Add the new customization flag to customization.config for each model to do the rename.
3.  Disable the new processor for these cases

codegen/src/main/java/software/amazon/awssdk/codegen/customization/processors/EventStreamSharedEventProcessor.java
```diff
@@ -0,0 +1,128 @@
...
+                }
+
+                String newShapeName = eventEntry.getValue();
+                if (serviceModel.getShapes().containsKey(newShapeName)) {
+                    throw new IllegalStateException(
+                        String.format("Error: [%s] is already in the model when processing "
+                                      + "customization config duplicateAndRenameSharedEvents.%s",
+                                      newShapeName, eventStreamName));
+                }
+                serviceModel.getShapes().put(newShapeName, eventMemberShape);
```
Fred1155: I may be misunderstanding the intent here, but In this case, is the newShapeName and shapeToDuplicate refers to the same Shape object? Is this the intended behavior, or should we create a deep copy of the Shape object?

alextwoods: Yea, good call out. I had originally done this to reduce memory but thinking about it more, I think we should deep copy here to prevent possible surprise if anything else is modifying the shapes.  I've updated.

Fred1155 (approved):

codegen/src/main/java/software/amazon/awssdk/codegen/customization/processors/EventStreamSharedEventProcessor.java
```diff
@@ -0,0 +1,143 @@
...
+        }
+        if (!shape.isEventstream()) {
+            throw new IllegalStateException(
+                String.format("Error: [%s] must be an EventStream when processing "
+                              + "customization config duplicateAndRenameSharedEvents.%s", name, name));
+        }
+    }
+
+    @Override
+    public void postprocess(IntermediateModel intermediateModel) {
```
RanVaknin, line 134: Can we add a test case to cover the post processing codepath?

RanVaknin (changes requested):

codegen/src/main/java/software/amazon/awssdk/codegen/customization/processors/EventStreamSharedEventProcessor.java
alextwoods: There is a test in `EventStreamSharedEventProcessorTest` - see: `modelWithSharedEvents_raises` - the test doesn't call postprocess directly, but instead relies on the IntermediateModel/C2jModel builders to call all of the processors.

RanVaknin (approved):

PR #6042, opened 2025-04-16
Add support for etag validation in resumableFileDownload

RanVaknin:
## Motivation and Context
Added support for ETag validation in resumable file downloads. The Transfer Manager now validates both the `lastModified` timestamp and the ETag when resuming a download. This aims to make sure that if the object has changed on S3 (indicated by a different ETag), the download will restart from the beginning rather than attempting to resume, which could lead to corrupted files.

## Modifications

- Added `s3ObjectEtag` field to `ResumableFileDownload` to store the ETag 
- Updated the serialization/deserialization logic in `ResumableFileDownloadSerializer` to include the ETag values.
- Updated `ResumableRequestConverter` to compare the stored ETag with the current S3 object's ETag.
- Added logging when an ETag mismatch is detected

## Testing

- Tests for different ETag values causing downloads to restart
- Tests for matching ETag values allowing downloads to 
- Tests for null/missing ETag values
-  Tests for case sensitivity in ETag comparisons
- Test for logging the hint about download being restarted

### DIFF
--- a/.changes/next-release/feature-AmazonS3TransferManager-21d3a50.json
+++ b/.changes/next-release/feature-AmazonS3TransferManager-21d3a50.json
@@ -1,0 +1,6 @@
+{
+    "type": "feature",
+    "category": "Amazon S3 Transfer Manager",
+    "contributor": "",
+    "description": "Add support for etag validation in resumableFileDownload"
+}
--- a/.changes/next-release/feature-AmazonS3TransferManager-21d3a50.json
+++ b/.changes/next-release/feature-AmazonS3TransferManager-21d3a50.json
@@ -2,5 +2,5 @@
     "type": "feature",
     "category": "Amazon S3 Transfer Manager",
     "contributor": "",
-    "description": "Add support for etag validation in resumableFileDownload"
+    "description": "Add support for etag validation in resumableFileDownload: restart paused downloads when etag does not match"
 }
--- a/services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/serialization/ResumableFileDownloadSerializer.java
+++ b/services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/serialization/ResumableFileDownloadSerializer.java
@@ -62,6 +62,11 @@ public final class ResumableFileDownloadSerializer {
                                        jsonGenerator,
                                        "s3ObjectLastModified");
         }
+        if (download.s3ObjectEtag().isPresent()) {
+            TransferManagerJsonMarshaller.STRING.marshall(download.s3ObjectEtag().get(),
+                                                           jsonGenerator,
+                                                           "s3ObjectEtag");
+        }
         marshallDownloadFileRequest(download.downloadFileRequest(), jsonGenerator);
         TransferManagerJsonMarshaller.LIST.marshall(download.completedParts(), jsonGenerator, "completedParts");
         jsonGenerator.writeEndObject();
@@ -127,6 +132,8 @@ public final class ResumableFileDownloadSerializer {
             (TransferManagerJsonUnmarshaller<Long>) getUnmarshaller(MarshallingType.LONG);
         TransferManagerJsonUnmarshaller<Instant> instantUnmarshaller =
             (TransferManagerJsonUnmarshaller<Instant>) getUnmarshaller(MarshallingType.INSTANT);
+        TransferManagerJsonUnmarshaller<String> stringUnmarshaller =
+            (TransferManagerJsonUnmarshaller<String>) getUnmarshaller(MarshallingType.STRING);
 
         ResumableFileDownload.Builder builder = ResumableFileDownload.builder();
         builder.bytesTransferred(longUnmarshaller.unmarshall(downloadNodes.get("bytesTransferred")));
@@ -137,6 +144,10 @@ public final class ResumableFileDownloadSerializer {
 
         if (downloadNodes.get("s3ObjectLastModified") != null) {
             builder.s3ObjectLastModified(instantUnmarshaller.unmarshall(downloadNodes.get("s3ObjectLastModified")));
+        }
+
+        if (downloadNodes.get("s3ObjectEtag") != null) {
+            builder.s3ObjectEtag(stringUnmarshaller.unmarshall(downloadNodes.get("s3ObjectEtag")));
         }
         builder.downloadFileRequest(parseDownloadFileRequest(downloadNodes.get("downloadFileRequest")));
         if (downloadNodes.get("completedParts") != null) {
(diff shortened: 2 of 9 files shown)


.changes/next-release/feature-AmazonS3TransferManager-21d3a50.json
```diff
@@ -0,0 +1,6 @@
+{
+    "type": "feature",
+    "category": "Amazon S3 Transfer Manager",
+    "contributor": "",
+    "description": "Add support for etag validation in resumableFileDownload"
```
alextwoods, line 5: Not sure what the general standards are for these - but maybe add description of the behavior? (something like "restart paused downloads when etag does not match")

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/utils/ResumableRequestConverter.java
```diff
@@ -60,15 +60,21 @@ private ResumableRequestConverter() {
         GetObjectRequest getObjectRequest = originalDownloadRequest.getObjectRequest();
         DownloadFileRequest newDownloadFileRequest;
         Instant lastModified = resumableFileDownload.s3ObjectLastModified().orElse(null);
-        boolean s3ObjectModified = !headObjectResponse.lastModified().equals(lastModified);
+        String resumableFileDownloadEtag = resumableFileDownload.s3ObjectEtag().orElse(null);
+
+        String s3ObjectEtag = headObjectResponse.eTag();
+        boolean etagModified = resumableFileDownloadEtag != null &&
```
alextwoods, line 66: naming nit (feel free to ignore) - the other variables here use s3 prefix (eg: s3EtagModified) and I think that helps indicate that the modification is on the s3 object side.

services-custom/s3-transfer-manager/src/test/java/software/amazon/awssdk/transfer/s3/util/ResumableRequestConverterTest.java
```diff
@@ -116,6 +124,65 @@ void toDownloadFileAndTransformer_fileLastModifiedTimeChanged_shouldStartFromBeg
         verifyActualGetObjectRequest(getObjectRequest, actual.left().getObjectRequest(), null);
     }
 
+    static Stream<Arguments> EtagTestCases() {
+        return Stream.of(
+            Arguments.of("acbd18db4cc2f85cedef654fccc4a4d8", null), // etag of "foo"
```
alextwoods, line 129: Since these are used below, maybe create a constant

RanVaknin:
Working on adding an integ test.

alextwoods (commented):
Nice - looks good overall.

I think could use an integration test case added in S3TransferManagerDownloadPauseResumeIntegrationTest

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/utils/ResumableRequestConverter.java
RanVaknin: the s3 prefix is for the headobject part, and then the `Modified` suffix is for the resumable part. It follows the existing convention that is underneath it:

```diff
        String resumableFileDownloadEtag = resumableFileDownload.s3ObjectEtag().orElse(null);

+        String s3ObjectEtag = headObjectResponse.eTag();
+        boolean etagModified = resumableFileDownloadEtag != null &&
+                               !resumableFileDownloadEtag.equals(s3ObjectEtag);


        boolean s3ObjectModified = !headObjectResponse.lastModified().equals(lastModified);
        boolean fileModified = !fileNotModified(resumableFileDownload.bytesTransferred(),
                                                resumableFileDownload.fileLastModified(),
                                                resumableFileDownload.downloadFileRequest().destination());
```

RanVaknin: I believe the `s3` prefix here is representing the headObjectResponse parameter, and the non `s3` prefixed variable represents the resumableDownload parameter?

just like this part underneath:

```diff
+	String s3ObjectEtag = headObjectResponse.eTag();
+       boolean etagModified = resumableFileDownloadEtag != null &&
+			!resumableFileDownloadEtag.equals(s3ObjectEtag);
        
       boolean s3ObjectModified = !headObjectResponse.lastModified().equals(lastModified);
       boolean fileModified = !fileNotModified(resumableFileDownload.bytesTransferred(),
                                                resumableFileDownload.fileLastModified(),
                                                resumableFileDownload.downloadFileRequest().destination());
```

Happy to change it either way.

.changes/next-release/feature-AmazonS3TransferManager-21d3a50.json
zoewangg: +1, we don't have a written guidance, but the general rule is that describe the bahvior instead of implementation to make it clear to customers

zoewangg (commented):
Don't we need to update `DefaultFileDownload#doPause` to capture etag?

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/serialization/ResumableFileDownloadSerializer.java
```diff
@@ -62,6 +62,11 @@ public static byte[] toJson(ResumableFileDownload download) {
                                        jsonGenerator,
                                        "s3ObjectLastModified");
         }
+        if (download.s3ObjectEtag().isPresent()) {
```
zoewangg, line 65: Can we add unit tests in ResumableFileDownloadSerializerTest?

zoewangg (changes requested):

.changes/next-release/feature-AmazonS3TransferManager-21d3a50.json
RanVaknin: done

services-custom/s3-transfer-manager/src/main/java/software/amazon/awssdk/transfer/s3/internal/serialization/ResumableFileDownloadSerializer.java
RanVaknin: Good call. I added coverage for that new Etag field in the existing test. Do I need to add a standalone test just for that, or is this sufficient?

services-custom/s3-transfer-manager/src/test/java/software/amazon/awssdk/transfer/s3/util/ResumableRequestConverterTest.java
RanVaknin: done

RanVaknin:
> Don't we need to update `DefaultFileDownload#doPause` to capture etag?

Yes good call. Actually the integ test I wrote actually caught that.

Fixed now.

zoewangg (approved):
Let's kick off integ tests to make sure they are passing if we haven't already

alextwoods (approved):

PR #6043, opened 2025-04-17
updated transact delete item flow to check the version for optimistic…

roamariei:
… locking

## Types of changes

## Checklist

## License


joviegas:
This is a breaking change right ?

What if currently user was using this API to delete any API m and such cases we will start getting exceptions ?

anasatirbasa:
> This is a breaking change right ?
> 
> What if currently user was using this API to delete any API m and such cases we will start getting exceptions ?


Hello, @joviegas

This is not a breaking change, as described in the [Solution Proposal](https://github.com/gorelov1/ddb-endava/blob/main/DesignDocuments/Java/%5BIssue2358%5DOptimisticLockingDeleteOperations.md). A validation step was added to ensure that the version in the delete request matches the version stored in the database.

If the versions do not match, the operation now throws a TransactionCanceledException, which is the expected behavior for conditional operations in DefaultDynamoDbClient in case a condition (in one of the condition expressions) is not met. This scenario is covered by the integration test _deleteItemWithTransactWrite_shouldFailIfVersionMismatch_ in _CrudWithResponseIntegrationTest.java_.

Before the changes from the current PR, the item would have been deleted even if the version did not match, and no exception was thrown. The new behavior enforces optimistic locking to improve data integrity.

joviegas:
> > This is a breaking change right ?
> > What if currently user was using this API to delete any API m and such cases we will start getting exceptions ?
> 
> Hello, @joviegas
> 
> This is not a breaking change, as described in the [Solution Proposal](https://github.com/gorelov1/ddb-endava/blob/main/DesignDocuments/Java/%5BIssue2358%5DOptimisticLockingDeleteOperations.md). A validation step was added to ensure that the version in the delete request matches the version stored in the database.
> 
> If the versions do not match, the operation now throws a TransactionCanceledException, which is the expected behavior for conditional operations in DefaultDynamoDbClient in case a condition (in one of the condition expressions) is not met. This scenario is covered by the integration test _deleteItemWithTransactWrite_shouldFailIfVersionMismatch_ in _CrudWithResponseIntegrationTest.java_.
> 
> Before the changes from the current PR, the item would have been deleted even if the version did not match, and no exception was thrown. The new behavior enforces optimistic locking to improve data integrity.

Right now, customers can delete items without worrying about versions, and it just works. But after our update, they're going to start getting TransactionCanceledException errors out of nowhere.

Think about it this way: let's say someone has code that tries to delete item X with version 1. They don't really care if the actual version is 2 or 3 - they just want the item gone. Today, that works fine. Tomorrow, after they upgrade, boom - exception thrown and their code breaks.

That's a classic breaking change. Their apps that work perfectly today will suddenly fail. They'll have to go in and modify their code to handle these new exceptions. And there's no way for them to get the old "just delete it" behavior back without changing their code.

Don't get me wrong - I totally get that this change makes things better from a data integrity standpoint with the optimistic locking.  Some customers probably built their apps specifically counting on that "delete regardless of version" behavior.
We need to think of approach which is not breaking the existing customers. What do you think?

anasatirbasa:
Hello @joviegas,
 
From this standpoint indeed it would be a breaking change. To address the potential breaking change and ensure full backward compatibility while introducing optimistic locking support, we propose the addition of a new overload for the addDeleteItem API:

- Current method signature (no Optimistic Locking support):
`public <T> Builder addDeleteItem(MappedTableResource<T> mappedTableResource, T keyItem)`

- Proposed overload (with Optimistic Locking support): 
`public <T> Builder addDeleteItem(MappedTableResource<T> mappedTableResource, T keyItem, Boolean useOptimisticLocking)`
 

The current method will internally delegate to the new method using useOptimisticLocking = false, thereby maintaining the existing "delete regardless of version" behavior without introducing any breaking change.
Developers who opt in to optimistic locking can explicitly call the new overload with useOptimisticLocking = true, which will enforce version checks during deletion and improve data integrity.

 We will refactor the tests to add coverage on both current and new behavior. This approach allows us to introduce the feature in a backward-compatible manner, giving users the choice to adopt the new behavior at their discretion.

We look forward to your feedback on this approach.

joviegas:
> Hello @joviegas,   From this standpoint indeed it would be a breaking change. To address the potential breaking change and ensure full backward compatibility while introducing optimistic locking support, we propose the addition of a new overload for the addDeleteItem API:
> 
>     * Current method signature (no Optimistic Locking support):
>       `public <T> Builder addDeleteItem(MappedTableResource<T> mappedTableResource, T keyItem)`
> 
>     * Proposed overload (with Optimistic Locking support):
>       `public <T> Builder addDeleteItem(MappedTableResource<T> mappedTableResource, T keyItem, Boolean useOptimisticLocking)`
> 
> 
> The current method will internally delegate to the new method using useOptimisticLocking = false, thereby maintaining the existing "delete regardless of version" behavior without introducing any breaking change. Developers who opt in to optimistic locking can explicitly call the new overload with useOptimisticLocking = true, which will enforce version checks during deletion and improve data integrity.
> 
> We will refactor the tests to add coverage on both current and new behavior. This approach allows us to introduce the feature in a backward-compatible manner, giving users the choice to adopt the new behavior at their discretion.
> 
> We look forward to your feedback on this approach.

Thanks anasatirbasa@ for the new version , will take a look at it in between this week

joviegas:
> Hello @joviegas,   From this standpoint indeed it would be a breaking change. To address the potential breaking change and ensure full backward compatibility while introducing optimistic locking support, we propose the addition of a new overload for the addDeleteItem API:
> 
>     * Current method signature (no Optimistic Locking support):
>       `public <T> Builder addDeleteItem(MappedTableResource<T> mappedTableResource, T keyItem)`
> 
>     * Proposed overload (with Optimistic Locking support):
>       `public <T> Builder addDeleteItem(MappedTableResource<T> mappedTableResource, T keyItem, Boolean useOptimisticLocking)`
> 
> 
> The current method will internally delegate to the new method using useOptimisticLocking = false, thereby maintaining the existing "delete regardless of version" behavior without introducing any breaking change. Developers who opt in to optimistic locking can explicitly call the new overload with useOptimisticLocking = true, which will enforce version checks during deletion and improve data integrity.
> 
> We will refactor the tests to add coverage on both current and new behavior. This approach allows us to introduce the feature in a backward-compatible manner, giving users the choice to adopt the new behavior at their discretion.
> 
> We look forward to your feedback on this approach.

Thanks anasatirbasa@ for the new version , will take a look at it in between this week

joviegas:
Thanks @anasatirbasa for the PR!

I've reviewed the changes and tested some scenarios. I have a concern about the current implementation scope: optimistic locking shouldn't be limited to just `TransactWriteItemsEnhancedRequest`. Since this feature is tied to the "VersionIdExtension," its behavior should remain consistent across all APIs, same as  `putItem`, as demonstrated below:

```java
@Test
public void deleteExistingRecordVersionMatches() {
    mappedTable.putItem(r -> r.item(new Record().setId("id").setAttribute("one")));

    mappedTable.putItem(new Record().setId("id").setAttribute("one").setVersion(1));
    mappedTable.putItem(new Record().setId("id").setAttribute("one").setVersion(2));

    // This fails as expected
    // mappedTable.putItem(new Record().setId("id").setAttribute("one").setVersion(1));
    
    // This should also fail for consistency when user is expecting Optimistic operation [In our case they will provide settings]
    mappedTable.deleteItem(new Record().setId("id").setAttribute("one").setVersion(1));
}
```

**Alternative Approach Suggestion:**

Instead of overriding specific APIs, I recommend adding a feature flag to the VersionIdExtension (similar to the approach in [PR #6019](https://github.com/aws/aws-sdk-java-v2/pull/6019)). We could introduce an "optimisticDelete" flag that:
- Defaults to `false` to maintain backward compatibility
- Can be enabled by users who want optimistic locking for delete operations across all APIs
- Ensures consistent behavior throughout the extension

**Next Steps:**

Could you please conduct a surface API review considering:
1. The recommended feature flag approach
2. Alternative implementation strategies
3. Consistency requirements across all VersionIdExtension operations

This would help ensure we're taking the most comprehensive and user-friendly approach.

anasatirbasa:
Hello, @joviegas 

We understand your concern and we've analyzed your proposal (defining a feature flag for optimistic locking on delete operations at extension level - VersionRecordExtension.java)
 
We have identified the following SDK APIs used for delete operations:

| **Class**                                   | **Method Signature**                                                                                                                                                          |
|---------------------------------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| `DefaultDynamoDbTable<T>`                   | - `public void deleteItem(DeleteItemEnhancedRequest request)`<br>- `public T deleteItem(Consumer<DeleteItemEnhancedRequest.Builder> requestConsumer)`<br>- `public T deleteItem(Key key)`<br>- `public T deleteItem(T keyItem)` |
| `TransactWriteItemsEnhancedRequest.Builder` | - `public <T> Builder addDeleteItem(MappedTableResource<T> mappedTableResource, TransactDeleteItemEnhancedRequest request)`<br> - `@Deprecated public <T> Builder addDeleteItem(MappedTableResource<T> mappedTableResource, DeleteItemEnhancedRequest request)`<br>- `public <T> Builder addDeleteItem(MappedTableResource<T> mappedTableResource, Key key)`<br>- `public <T> Builder addDeleteItem(MappedTableResource<T> mappedTableResource, T keyItem)` |
| `DefaultDynamoDbAsyncTable<T>`              | - `public CompletableFuture<T> deleteItem(DeleteItemEnhancedRequest request)`<br>- `public CompletableFuture<T> deleteItem(Consumer<DeleteItemEnhancedRequest.Builder> requestConsumer)`<br>- `public CompletableFuture<T> deleteItem(Key key)`<br>- `public CompletableFuture<T> deleteItem(T keyItem)` |
| `WriteBatch.Builder<T>`                     | - `public Builder<T> addDeleteItem(DeleteItemEnhancedRequest request)`<br>- `public Builder<T> addDeleteItem(Consumer<DeleteItemEnhancedRequest.Builder> requestConsumer)`<br>- `public Builder<T> addDeleteItem(Key key)`<br>- `public Builder<T> addDeleteItem(T keyItem)` |


### Proposed Behavior with Optimistic Locking Flag

- **Flag disabled**  
  – Delete operations proceed without version checks; the item is removed unconditionally.

- **Flag enabled & version provided**  
  – The supplied version is compared to the stored version; deletion occurs only if they match.

- **Flag enabled & no version provided**  
  – Version is not checked; the item is removed unconditionally.

---

### Implementation Strategies

#### 1. DynamoDbEnhancedClient level configuration

```java
DynamoDbEnhancedClient client = DynamoDbEnhancedClient.builder()
    .dynamoDbClient(getDynamoDbClient())
    .extensions(VersionedRecordExtension.builder()
        .optimisticLocking(true)
        .build())
    .build();
```

*We will have to pass the optimistic locking flag when we attach the custom VersionedRecordExtension when building DynamoDbEnhancedClient.*

#### 2. Annotation-based configuration

```java
@DynamoDbVersionAttribute(optimisticLocking = true)
public Integer getVersion() {
    return version;
}
```

*Allows per-entity control of optimistic locking via annotations.*

---

The first approach offers a single, centralized configuration but lacks per-schema flexibility. The annotation-based approach provides control at the model level, allowing different entities to opt in or out independently.

Could you please advise which solution fits better from your perspective?
Thank you!

joviegas:
> Hello, @joviegas
> 
> We understand your concern and we've analyzed your proposal (defining a feature flag for optimistic locking on delete operations at extension level - VersionRecordExtension.java)   We have identified the following SDK APIs used for delete operations:
> **Class** 	**Method Signature**
> `DefaultDynamoDbTable<T>` 	- `public void deleteItem(DeleteItemEnhancedRequest request)`
> - `public T deleteItem(Consumer<DeleteItemEnhancedRequest.Builder> requestConsumer)`
> - `public T deleteItem(Key key)`
> - `public T deleteItem(T keyItem)`
> `TransactWriteItemsEnhancedRequest.Builder` 	- `public <T> Builder addDeleteItem(MappedTableResource<T> mappedTableResource, TransactDeleteItemEnhancedRequest request)`
> - `@Deprecated public <T> Builder addDeleteItem(MappedTableResource<T> mappedTableResource, DeleteItemEnhancedRequest request)`
> - `public <T> Builder addDeleteItem(MappedTableResource<T> mappedTableResource, Key key)`
> - `public <T> Builder addDeleteItem(MappedTableResource<T> mappedTableResource, T keyItem)`
> `DefaultDynamoDbAsyncTable<T>` 	- `public CompletableFuture<T> deleteItem(DeleteItemEnhancedRequest request)`
> - `public CompletableFuture<T> deleteItem(Consumer<DeleteItemEnhancedRequest.Builder> requestConsumer)`
> - `public CompletableFuture<T> deleteItem(Key key)`
> - `public CompletableFuture<T> deleteItem(T keyItem)`
> `WriteBatch.Builder<T>` 	- `public Builder<T> addDeleteItem(DeleteItemEnhancedRequest request)`
> - `public Builder<T> addDeleteItem(Consumer<DeleteItemEnhancedRequest.Builder> requestConsumer)`
> - `public Builder<T> addDeleteItem(Key key)`
> - `public Builder<T> addDeleteItem(T keyItem)`
> ### Proposed Behavior with Optimistic Locking Flag
> 
>     * **Flag disabled**
>       – Delete operations proceed without version checks; the item is removed unconditionally.
> 
>     * **Flag enabled & version provided**
>       – The supplied version is compared to the stored version; deletion occurs only if they match.
> 
>     * **Flag enabled & no version provided**
>       – Version is not checked; the item is removed unconditionally.
> 
> 
> ### Implementation Strategies
> #### 1. DynamoDbEnhancedClient level configuration
> 
> ```java
> DynamoDbEnhancedClient client = DynamoDbEnhancedClient.builder()
>     .dynamoDbClient(getDynamoDbClient())
>     .extensions(VersionedRecordExtension.builder()
>         .optimisticLocking(true)
>         .build())
>     .build();
> ```
> 
> _We will have to pass the optimistic locking flag when we attach the custom VersionedRecordExtension when building DynamoDbEnhancedClient._
> #### 2. Annotation-based configuration
> 
> ```java
> @DynamoDbVersionAttribute(optimisticLocking = true)
> public Integer getVersion() {
>     return version;
> }
> ```
> 
> _Allows per-entity control of optimistic locking via annotations._
> 
> The first approach offers a single, centralized configuration but lacks per-schema flexibility. The annotation-based approach provides control at the model level, allowing different entities to opt in or out independently.
> 
> Could you please advise which solution fits better from your perspective? Thank you!

Thank you so much for the detailed documentation and thorough analysis of the APIs. I really appreciate the comprehensive breakdown of the implementation approach.

I apologize for the delay - I had posted in our internal Slack asking for a quick API design review and was waiting for feedback. However, I see you've already provided excellent documentation directly in this PR, which is very helpful.

Important clarification needed: Since we're using the generic name optimisticLocking, could you please confirm that when this flag is enabled, it will apply to all write operations (create, update, delete) consistently? Per the DynamoDB [documentation](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DynamoDBMapper.OptimisticLocking.html):

    "Optimistic locking is a strategy to ensure that the client-side item that you are updating (or deleting) is the same as the item in Amazon DynamoDB."

This would mean (please help me confirm):

    Create operations: Would fail if an item with the same key already exists
    Update operations: Would require version matching (existing behavior)
    Delete operations: Would require version matching (your proposed new behavior)

Regarding your implementation approach:

I see you're proposing both client-level configuration via the builder AND annotation support. A few questions that would help me better understand the design:

    Precedence: When both are specified, which takes precedence - the client-level setting or the annotation?
    Batch operations: How would batch/transactional operations handle mixed scenarios where some items are not in sync with version number? Is it possible to have a batch with some items having optmisticlocking enabled and some disabled
- Why are we deprecating addDeleteItem with this approach ?
```
- @Deprecated public <T> Builder addDeleteItem(MappedTableResource<T> mappedTableResource, 
```
If you could help clarify these points, I'll complete a quick internal review with the team and provide feedback.

anasatirbasa:
Hello @joviegas,


Thanks for your feedback and the request for consistency across all operations.

Just to clarify, we have provided two alternatives for how the optimistic locking flag could be introduced, either at Version extension level or at annotation level, but not both together. For the moment let’s consider only the flag at extension level option.

Regarding the consistency across all operations, below is a clarification of our design options and the trade‑offs we see:

1) **Option A: Flag at Version Extension level across all operations**
- Introduce a single global flag (e.g., `optimisticLocking`) on the Version extension.
- All operations (create, update, delete) will behave based on this flag consistently.
- Major drawback: this changes the default behavior. Today, create/update implicitly enforce optimistic version checks, while delete does not. Under this model, users must explicitly enable the Version extension with `optimisticLocking = true` when building their enhanced client to retain existing create/update behavior. If they omit it, those checks go away.
- This breaks backward compatibility unless we make the flag default to true, but in that case it will break the compatibility for delete operations, as currently the delete doesn’t consider optimisticLocking checks.


2) **Option B: Flag at Version Extension level only for Delete Operations**
- Introduce a flag (e.g. `optimisticDelete`) scoped only to delete behavior.
- Create/update continue working exactly as today (optimistic by default), so existing clients remain totally unaffected. Delete becomes version‑checked only if extension is explicitly defined with `optimisticDelete = true`.
- Drawback: create/update/delete won't be symmetric, but that's already the case today.


Also: the `@Deprecated addDeleteItem(...)` method was deprecated previously; we are not deprecating anything new as part of this change.


**Question for you**  
Given these constraints, which path would you prefer?

- **Option A**: Introduce a globally scoped optimistic‑locking flag to enforce consistent versioning behavior across create, update, and delete, but requiring explicit opt‑in to maintain existing behavior.  
- **Option B**: Introduce a delete‑specific flag (e.g. `optimisticDelete`)—preserving all current behavior unchanged for create/update, at the cost of not having full consistency.

Happy to refine either path further given your input.

Thanks again and looking forward to your feedback!

joviegas:
> Introduce a single global flag (e.g., `optimisticLocking`) on the Version extension.

@anasatirbasa 
Thanks for the response
`Option B: Flag at Version Extension level only for Delete Operations` makes more sense.
Something like https://github.com/aws/aws-sdk-java-v2/pull/6327 , I just did some quick Proof of concept PR , here we can even avoid the new api for beforeWrite, what do you think of this approach.

anasatirbasa:
Hello, @joviegas!

I really appreciate the effort you put into this POC - the proposed solution is very good and closely aligned with an approach I had previously implemented locally (but had not committed at that time, prior to your review of the PR).

I have used my local changes as a foundation and incorporated your logic for **defining the flag at both the extension and annotation levels**, as well as the precedence rules you proposed.

In addition, I extended the **VersionedRecordExtension to also support the table.deleteItem() operation** (DeleteItemEnhancedRequest). Previously, the implementation supported only enhancedClient.transactWriteItems() (TransactDeleteItemEnhancedRequest) - see the code in DeleteItemOperation.generateRequest().

I have also added **unit and integration tests** to cover all relevant scenarios (see CrudWithResponseIntegrationTest, AsyncCrudWithResponseIntegrationTest, VersionedRecordExtensionTest, and DeleteItemOperationTest).

Please have a look at the committed changes and share your thoughts so we can move this PR toward completion.
Thank you very much!

joviegas:
> Hello, @joviegas!
> 
> I really appreciate the effort you put into this POC - the proposed solution is very good and closely aligned with an approach I had previously implemented locally (but had not committed at that time, prior to your review of the PR).
> 
> I have used my local changes as a foundation and incorporated your logic for **defining the flag at both the extension and annotation levels**, as well as the precedence rules you proposed.
> 
> In addition, I extended the **VersionedRecordExtension to also support the table.deleteItem() operation** (DeleteItemEnhancedRequest). Previously, the implementation supported only enhancedClient.transactWriteItems() (TransactDeleteItemEnhancedRequest) - see the code in DeleteItemOperation.generateRequest().
> 
> I have also added **unit and integration tests** to cover all relevant scenarios (see CrudWithResponseIntegrationTest, AsyncCrudWithResponseIntegrationTest, VersionedRecordExtensionTest, and DeleteItemOperationTest).
> 
> Please have a look at the committed changes and share your thoughts so we can move this PR toward completion. Thank you very much!

Hi @anasatirbasa,

First, I want to thank you for your patience and perseverance through multiple iterations of this implementation. Your dedication to getting this feature right is truly appreciated.

We had an internal surface API review for your changes. While avoiding generics by using [Map<String, AttributeValue>](https://github.com/roamariei/aws-sdk-java-v2/blob/078f16d90b42a9e2455fe7d360a556bca7972d54/services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/model/DeleteItemEnhancedRequest.java#L47) was a creative approach, the team raised concerns about adding this to the public API `DeleteItemEnhancedRequest`. The main issues identified were:

1. **API Confusion**: `DeleteItemEnhancedRequest` already has a `Key` field that identifies the item to be deleted. Adding an `items` map creates ambiguity about which field to use and when.

2. **User Experience**: If users need to use the items field outside the VersionedRecordExtension context, they would need to manually call `itemToMap()`, which isn't an intuitive experience.

Based on our discussion, here are the decision logs from the review:

| Decision | Rationale |
|----------|-----------|
| **Revisit current implementation approach** | The Map<String, AttributeValue> solution compromises type safety and creates a confusing user experience |
| **Implement client-level optimistic locking by default** | Automatic version checking without configuration flags provides better default behavior and consistency with UPDATE operations |
| **Create new type-safe DeleteItemEnhancedRequest API similar to PutItemEnhancedRequest which takes in generic T item** | Accept generic Item<T> instead of Key to maintain type safety and enable optimistic locking by default. Since it will be a new API, there will be no backward compatibility concerns |

I understand this means another revision, but we believe this approach will provide a cleaner, more intuitive API for our users. Would you be open to exploring this new direction? I'm happy to collaborate on the design if that would be helpful.

Thank you again for all your hard work on this feature!
Is it possible to have an internal review with the surface API team like we did for other features? I can also do it, but I'm currently working on other project which might cause delays.

anasatirbasa:
Hi @joviegas,

Thank you for the detailed review and for providing clear guidance on the improvements. I have implemented the suggested changes and pushed them to the branch.

The branch has been updated to incorporate these adjustments. When convenient, I would greatly appreciate it if you could review the latest changes and advise if any additional refinements are required.

Thank you once again for the time and effort you have invested in reviewing this contribution!

joviegas:
```java

import java.time.Duration;
import org.junit.jupiter.api.AfterAll;
import org.junit.jupiter.api.AfterEach;
import org.junit.jupiter.api.BeforeAll;
import software.amazon.awssdk.enhanced.dynamodb.extensions.annotations.DynamoDbVersionAttribute;
import software.amazon.awssdk.enhanced.dynamodb.mapper.annotations.DynamoDbBean;
import software.amazon.awssdk.enhanced.dynamodb.mapper.annotations.DynamoDbPartitionKey;
import software.amazon.awssdk.services.dynamodb.DynamoDbClient;
import software.amazon.awssdk.services.dynamodb.model.ResourceNotFoundException;
import software.amazon.awssdk.services.dynamodb.waiters.DynamoDbWaiter;

public class OptimisticLockingIntegrationTest {

    private static final String TABLE_NAME = "test-optimistic-locking";
    private static DynamoDbClient dynamoDbClient;
    private static DynamoDbEnhancedClient enhancedClient;
    private static DynamoDbTable<TestItem> table;

    @Test
    void testDeleteVersionConflict() {
        // Arrange
        TestItem item = new TestItem("item-1", "data");
        table.putItem(item);

        TestItem localItem = table.getItem(r -> r.key(k -> k.partitionValue("item-1")));

        TestItem currentItem = new TestItem("item-1", "data");
        currentItem.setData("updated-data");
        currentItem.setVersion(localItem.getVersion());
        table.putItem(currentItem);

        // Delete local version
        table.deleteItem(localItem);

        TestItem afterDelete = table.getItem(r -> r.key(k -> k.partitionValue("item-1")));
        assertNull(afterDelete);
    }

    @BeforeAll
    static void setUpClass() {
        // Use real AWS DynamoDB
        dynamoDbClient = DynamoDbClient.builder().build();

        enhancedClient = DynamoDbEnhancedClient.builder()
                                               .dynamoDbClient(dynamoDbClient)
                                               .build();

        table = enhancedClient.table(TABLE_NAME, TableSchema.fromBean(TestItem.class));

        // Delete table if it exists
        try {
            table.deleteTable();
            System.out.println("Existing table found, deleting...");

            // Wait for table to be deleted
            try (DynamoDbWaiter waiter = DynamoDbWaiter.builder()
                                                       .client(dynamoDbClient)
                                                       .overrideConfiguration(o -> o.waitTimeout(Duration.ofMinutes(5)))
                                                       .build()) {
                waiter.waitUntilTableNotExists(r -> r.tableName(TABLE_NAME));
            }
            System.out.println("Table deleted successfully");
        } catch (ResourceNotFoundException e) {
            System.out.println("Table doesn't exist, proceeding with creation");
        }

        // Create table
        System.out.println("Creating table...");
        table.createTable();

        // Wait for table to be active
        try (DynamoDbWaiter waiter = DynamoDbWaiter.builder()
                                                   .client(dynamoDbClient)
                                                   .overrideConfiguration(o -> o.waitTimeout(Duration.ofMinutes(5)))
                                                   .build()) {
            waiter.waitUntilTableExists(r -> r.tableName(TABLE_NAME));
        }
        System.out.println("Table created and active");
    }

    @AfterAll
    static void tearDownClass() {
        try {
            // Delete table
            System.out.println("Deleting table...");
            table.deleteTable();

            // Wait for table to be deleted
            try (DynamoDbWaiter waiter = DynamoDbWaiter.builder()
                                                       .client(dynamoDbClient)
                                                       .overrideConfiguration(o -> o.waitTimeout(Duration.ofMinutes(5)))
                                                       .build()) {
                waiter.waitUntilTableNotExists(r -> r.tableName(TABLE_NAME));
            }
            System.out.println("Table deleted successfully");
        } catch (Exception e) {
            System.err.println("Error during table cleanup: " + e.getMessage());
        }

        if (dynamoDbClient != null) {
            dynamoDbClient.close();
        }
    }

    @AfterEach
    void cleanupItems() {
        // Clean up test data after each test
        try {
            // Delete test items - adjust based on your test item IDs
            for (String itemId : new String[] {"item-1", "item-2", "item-3"}) {
                try {
                    table.deleteItem(r -> r.key(k -> k.partitionValue(itemId)));
                } catch (Exception e) {
                    // Ignore if item doesn't exist
                }
            }
        } catch (Exception e) {
            // Ignore cleanup errors
        }
    }




    @DynamoDbBean
    public static class TestItem {

        private String id;
        private String data;
        private Long version;

        public TestItem() {
        }

        public TestItem(String id, String data) {
            this.id = id;
            this.data = data;
        }

        @DynamoDbPartitionKey
        public String getId() {
            return id;
        }

        public void setId(String id) {
            this.id = id;
        }

        public String getData() {
            return data;
        }

        public void setData(String data) {
            this.data = data;
        }

        @DynamoDbVersionAttribute
        public Long getVersion() {
            return version;
        }

        public void setVersion(Long version) {
            this.version = version;
        }
    }
}
```

joviegas:
> Hi @joviegas,
> 
> Thank you for the detailed review and for providing clear guidance on the improvements. I have implemented the suggested changes and pushed them to the branch.
> 
> The branch has been updated to incorporate these adjustments. When convenient, I would greatly appreciate it if you could review the latest changes and advise if any additional refinements are required.
> 
> Thank you once again for the time and effort you have invested in reviewing this contribution!

Thanks @anasatirbasa I had a quick review and internal discussion and following were the concerns:

The current approach still causes backward compatibility issue as shown in [testDeleteVersionConflict](https://github.com/aws/aws-sdk-java-v2/pull/6043#issuecomment-3234803388). In order to avoid this there can be two possible solutions:

a. Have a new api say deleteItemWithOptimisticLocking(item) or deleteItem(Item item, boolean optimisticLocking) both this approach appear to solve the problem but it can cause confusion since these APIs do not do anything when VersionIdExtension is not used

b. Other option is same as before, where we have a boolean flag on versionId and we perform optimistic locking as done in your current PR only when this feature is set to true. But here we had concern that these new classes DeleteCompleteItemEnhancedRequest has its meaning or serves it purpose only when there is VersionIdExtension

Even if we go with 1.a approach we see a lot of new public APIs are introduced just to support Optimistic locking for delete with VersionIdExtension, and we will need to maintain the new and old DeleteEnhanced classes, thus were wondering if there is any approach where we provide OptimisticLockingHelper
```java
public static DeleteItemEnhancedRequest withOptimisticLocking(
    DeleteItemEnhancedRequest request,
    AttributeValue oldVersionValue,
    String versionAttributeName) {

    Expression conditionExpression = Expression.builder()
                                               .expression(versionAttributeName + " = " + VERSION_VALUE_PLACEHOLDER)
                                               .putExpressionValue(VERSION_VALUE_PLACEHOLDER, oldVersionValue)
                                               .build();

    return request.toBuilder()
                  .conditionExpression(conditionExpression)
  ```
  and use this as wrapper on DeleteItemEnhancedRequest ? something like this , Is there any approach could you think of ? or else we need to handle the backward incompatibility with current approach and then I will have another round of surface API review, will do it after you raise the next version.

PR #6051, opened 2025-04-22
Update java doc on ContentLength in requestBody for s3 uploads

Fred1155:
## Motivation and Context

Customers may get partial object puts when they SDK to upload an input stream, but they incorrectly set the content length. This PR improved the documentation surrounding this usage to warn user from inadvertently setting an incorrect Content-Length.
## Modifications

Added documentation in
```
AsyncRequestBody.fromInputStream
AsyncRequestBody.forBlockingInputStream
AsyncRequestBody.forBlockingOutputStream
RequestBody.fromInputStream
RequestBody.fromContentProvider
```
## Types of changes

## Checklist

## License


core/sdk-core/src/main/java/software/amazon/awssdk/core/sync/RequestBody.java
```diff
@@ -129,6 +129,8 @@ public static RequestBody fromFile(File file) {
      * 128 KiB. If you need more control, use {@link #fromContentProvider(ContentStreamProvider, long, String)} or
      * {@link #fromContentProvider(ContentStreamProvider, String)}.
      * <p>
+     * If a content length smaller than the actual size of the object is set, the client will truncate the stream to the
+     * specified content length and only send exactly the number of bytes equal to the content length.
      *
      * @param inputStream   Input stream to send to the service. The stream will not be closed by the SDK.
      * @param contentLength Content length of data in input stream.
```
joviegas, line 136: It will be of direct help if we mention this right next to @param contentLength , WDYT ?
```
 * @param inputStream   Input stream to send to the service. The stream will not be closed by the SDK.
 * @param contentLength Content length of data in input stream. 
 *                      If a content length smaller than the actual size of the 
 *                      object is set, the client will truncate the stream to the specified content length and only send 
 *                      exactly the number of bytes equal to the content length.
 * @return RequestBody instance.
 */


```

core/sdk-core/src/main/java/software/amazon/awssdk/core/async/AsyncRequestBody.java
```diff
@@ -424,6 +429,8 @@ static BlockingInputStreamAsyncRequestBody forBlockingInputStream(Long contentLe
      *
      * <p>By default, it will time out if streaming hasn't started within 10 seconds, and you can configure the timeout
      * via {@link BlockingOutputStreamAsyncRequestBody#builder()}
+     * <p>If a content length smaller than the actual size of the object is set, the client will truncate the stream to the
+     * specified content length and only send exactly the number of bytes equal to the content length.
```
joviegas, line 433: Can we add as below
```java
 * @param contentLength Content length of data to be written to the output stream. If a content length smaller than the 
 *                      actual size of the object is set, the client will truncate the stream to the specified content 
 *                      length and only send exactly the number of bytes equal to the content length.
 * @return A {@link BlockingOutputStreamAsyncRequestBody} instance.
```

joviegas (approved):

PR #6052, opened 2025-04-22
Duplicate all eventstream event shapes + add new legacy event modes

alextwoods:
Fixes codegeneration issues for eventstreams with shared event shapes: Duplicate and rename all eventstream event shapes with a unique name + add new disableUniqueEventStreamShapePreprocessing customization config.

## Motivation and Context
When an event shape is shared between multiple eventstreams, it causes SDK generation/compilation failures. The top level shape POJO implements the event stream interface for each stream and the return type of the sdkEventType method conflicts.

## Modifications
This PR does two things:
1.  Adds a new `disableUniqueEventStreamShapePreprocessing` customization config.
2.  If both `useLegacyEventGenerationScheme` and `disableUniqueEventStreamShapePreprocessing` is unset a for an event, the event shape will be duplicated and given a unique name: `<ShapeName><EventStreamName>`.

Note: There are a large number of changes in the codegen test to add the new customization mode to existing test eventstreams.

### What naming scheme should we use to produce unique names?
This PR currently uses `<EventStreamName><ShapeName>`, for example an event named `Payload` in eventstream `StreamA` would get the name: `StreamAPayload`.    We could reverse this (`PayloadStreamA`).

The downside of this naming scheme is that it can produce long names.  For example the existing [HeartbeatEvent](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/lexruntimev2/model/HeartbeatEvent.html) from LexRuntimeV2's [StartConversationResponseEventStream](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/lexruntimev2/model/StartConversationResponseEventStream.html) would be (without the legacy customization) renamed to: `HeartbeatEventStartConversationResponseEventStream`.

### Why do we always need to rename the event shape? Couldn't we do that only when there is a shared event?
We must ensure that shapes names are deterministic and do not depend on the shapes in the model because if a shared event shape is added in the future, we have no way of telling which was the existing shape.

## Alternatives
For POCs of alternative approaches see:
* Support through customization: https://github.com/aws/aws-sdk-java-v2/pull/6031
* Remove eventstream interface from generic POJOs: https://github.com/alextwoods/aws-sdk-java-v2/pull/2/files
* add SDKEventType interface: https://github.com/alextwoods/aws-sdk-java-v2/pull/3

## User Experience
There are no changes for users for any existing Eventstreams/events.  New eventstreams or new events added to existing evenstreams will have unique, longer names, but the fluid visitor and builder methods will continue to use the member names.  For example, assuming we have `Payload` event (with member name "PayloadEvent") in eventstream `StreamA`, the handler and builder methods will still be `onPayloadEvent` and `payloadEventBuilder`.

```java
// Handling events
StreamAResponseHandler.builder()
    .subscriber(
        StreamAResponseHandler.Visitor.builder()
            .onPayloadEvent(event -> { // NOTE: onPayload is still used because the member name is "PayloadEvent"
                System.out.println("Payload event: " + event);
            })
            .build()
    );

// sending events
subscriber.onNext(
    StreamA.payloadEventBuilder() // NOTE: We still use `payloadEventBuilder` because the member name is "Payload"
        .payload("text/plain;charset=utf-8")
        .build()
);
```

Note: For users using switch/case or if/else on `sdkEventType` or checking `instance of`, when casting the events they will need to use the longer shape names.  Using the example from above:

```java
subsciber(event -> {
    if (event instanceof PayloadStreamA) { // NOTE: we need to check the rename/unique shape name here
        handlePayloadEvent((PayloadStreamA)event);
    }
});

subscriber(event -> {
    switch(event.sdkEventType()) {
        case PAYLOAD_EVENT: // the sdkEventType will still use the member name
            handlePayloadEvent((PayloadStreamA)event); // Note: Need to cast to the renamed/uniqe shape name
            break;
    }
})
```

## Testing
New Unit tests - see shared event stream cases (model was modified to add new streams that share an event)

## Types of changes

## Checklist

## License

- [x] I confirm that this pull request can be released under the Apache 2 licens


codegen/src/main/java/software/amazon/awssdk/codegen/customization/processors/EventStreamUniqueEventShapesProcessor.java
```diff
@@ -0,0 +1,92 @@
...
+            .getOrDefault(eventStreamName, Collections.emptyMap());
+
+        eventStreamShape.getMembers().forEach((memberName, member) -> {
+            String eventShapeName = member.getShape();
+            Shape memberTargetShape = serviceModel.getShape(eventShapeName);
+            LegacyEventGenerationMode legacyEventGenerationMode = eventLegacyModes
+                .getOrDefault(memberName, LegacyEventGenerationMode.DISABLED);
+
+            if (memberTargetShape.isEvent() && legacyEventGenerationMode == LegacyEventGenerationMode.DISABLED) {
+                String newShapeName = eventShapeName + eventStreamName;
```
dagnir, line 70: Maybe put this in `NamingStrategy`

dagnir:
> The downside of this naming scheme is that it can produce long names. For example the existing [HeartbeatEvent](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/lexruntimev2/model/HeartbeatEvent.html) from LexRuntimeV2's [StartConversationResponseEventStream](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/lexruntimev2/model/StartConversationResponseEventStream.html) would be (without the legacy customization) renamed to: HeartbeatEventStartConversationResponseEventStream.

The current non-legacy way is there's a package per eventstream, e.g. `software.amazon.awssdk.services.lexruntimev2.model.startconversationresponseeventstream` that contains the event stream specific event classes. Could we do something similar in this case?

alextwoods:
> The current non-legacy way is there's a package per eventstream, e.g. software.amazon.awssdk.services.lexruntimev2.model.startconversationresponseeventstream that contains the event stream specific event classes. Could we do something similar in this case?

I did try to put together a POC for this but It would require a much more significant refactoring to get working.  There are a few issues/sharp edges:
1. In some cases we need both the model.<Event> and model.<EventStream>.<Event> POJOs  (an event in a legacy ES + in a new, non-legacy ES).
2. There are package private model classes that are required by the event pojos.  We would need to make these public to allow access/usage in the eventstream model package).
3. Fairly extensive updates to code generation for marshallers, async client (usage of marshallers), Builders, Handlers to ensure the correct types are used (each place needs to switch behavior on legacy modes).

codegen/src/main/java/software/amazon/awssdk/codegen/customization/processors/EventStreamUniqueEventShapesProcessor.java
alextwoods: Good suggestion - done!

codegen/src/main/java/software/amazon/awssdk/codegen/model/config/customization/CustomizationConfig.java
dagnir, line 1: Can we do a dryrun internally to see if there any usages of `useLegacyEventGenerationScheme` that break?

codegen/src/main/java/software/amazon/awssdk/codegen/naming/DefaultNamingStrategy.java
```diff
@@ -411,6 +412,12 @@ public String getUnionEnumTypeName(MemberModel memberModel) {
         return screamCase(memberModel.getName());
     }
 
+    @Override
+    public String getUniqueEventStreamEventShapeName(Member eventMember, String eventStreamName) {
+        return eventMember.getShape() + eventStreamName;
```
dagnir, line 417: Just checking: is it necessary to have `eventmember.getShape()` be capitalized here?

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/model/sharedstream/streamdeathsinputeventstream.java
```diff
@@ -0,0 +1,79 @@
...
+ * Base interface for all event types in StreamDeathsInputEventStream.
+ */
+@Generated("software.amazon.awssdk:codegen")
+@SdkPublicApi
+public interface StreamDeathsInputEventStream {
+    /**
+     * Create a builder for the {@code Person} event type for this stream.
+     */
+    static PersonStreamDeathsInputEventStream.Builder personBuilder() {
+        return DefaultPerson.builder();
```
dagnir, line 21: Can we have tests to verify the code for `DefaultPerson`?

services/bedrockagentruntime/src/main/resources/codegen-resources/customization.config
```diff
@@ -1,4 +1,76 @@
 {
     "enableGenerateCompiledEndpointRules": true,
-    "enableFastUnmarshaller": true
-}
+    "enableFastUnmarshaller": true,
+    "useLegacyEventGenerationScheme": {
+        "FlowResponseStream": {
```
dagnir, line 5: Just to check my understanding, what happens if/when Bedrock adds a new event to this eventstream? Will it default to `DISABLED` or since the mapping doesn't exist it will fail?

dagnir (changes requested):

codegen/src/main/java/software/amazon/awssdk/codegen/naming/DefaultNamingStrategy.java
alextwoods: What do you mean? The parameter name `eventMember`? Or do you mean we should apply some naming transform of the shape name to ensure it has correct capitalization?

RE: the parameter name, theres no reason it couldn't be `eventmember`.

RE: applying capitalization to the existing shape name - Thats a good question - I was assuming that the existing shape name has already been validated (and any divergence from standard is intended and should be preserved).

services/bedrockagentruntime/src/main/resources/codegen-resources/customization.config
alextwoods: New event members added to existing eventstreams will just have their shape's renamed to ensure uniqueness, there aren't any issues mixing the legacy behavior with the new behavior in the same eventstream.

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/model/sharedstream/streamdeathsinputeventstream.java
alextwoods: The `SharedStreamAwsModelSpecTest` is testing only the interfaces, but `EventModelSpecTest` (which uses a different model) tests tests the `Default<Shape>` generation (but not the interfaces) for event stream event shapes - between the two I think we get decent coverage of all of the files we're generating.  

It might be a little tricky, but the SharedStreamAwsModelSpecTest could probably be updated to test the implementations (ie `Default<shape>.java`) as well but we would need to add files for all of the shapes I think.  What do you think?

dagnir: Yeah let's do that please. I don't think we need to do it for every renamed shape, just for a sample.

dagnir (approved):

dagnir:
@zoewangg Can you take a look as well please?

If we haven't done so already, can we do a before and after check with the services with existing eventstream to make sure there aren't any changes?

PR #6055, opened 2025-04-24
Add LegacyMd5Plugin for MD5 checksum calculations in S3 operations requiring checksums

joviegas:
## Motivation and Context
https://github.com/aws/aws-sdk-java-v2/discussions/5802#discussioncomment-12281124

- The recent S3 changes introduced challenges for customers using S3A (Apache Spark, Iceberg) with various third-party storage providers. 
- While some providers like MinIO have fixes available, many customers use different storage solutions ( or cannot easily upgrade their third-party stores. 
- This plugin provides a backwards-compatible solution allowing customers to maintain MD5 checksum calculations where required while adopting S3 Motorcade features.

- Note: This Plugin adds MD5 checksum for operations which require checksums along with the default checksum provided by SDK for the operation which require checksum .

```java
 // For synchronous S3 client
 S3Client s3Client = S3Client.builder()
                            .addPlugin(LegacyMd5Plugin.create())
                            .build();

 // For asynchronous S3 client
 S3AsyncClient asyncClient = S3AsyncClient.builder()
                                         .addPlugin(LegacyMd5Plugin.create())
                                         .build();
```
If you want to add MD5 checksums to the operations that require checksums and want to skip adding of SDK Default checksums for operations that support checksums but not required, then you can enable ClientBuilder options requestChecksumCalculation and responseChecksumValidation as WHEN_REQUIRED, this will add SDK default checksums only to operation that required checksums
```java
// Use LegacyMd5Plugin with requestChecksumCalculation and responseChecksumValidation set to WHEN_REQUIRED
S3AsyncClient asyncClient = S3AsyncClient.builder()
                                         .addPlugin(LegacyMd5Plugin.create())
                                         .requestChecksumCalculation(RequestChecksumCalculation.WHEN_REQUIRED)
                                         .responseChecksumValidation(ResponseChecksumValidation.WHEN_REQUIRED)
                                         .build();
```

## Modifications

- A new plugin which updates the configs
- New APIs on ClientConfigBuilders
- Interceptor that is same as [HttpChecksumStage-legacyChecksum](https://github.com/aws/aws-sdk-java-v2/blob/03343864b9e9135f659d29f73de83794734ac9d5/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/HttpChecksumStage.java#L174-L181)

## Testing

- Junits
- One-off integs

- A Test class was created with following APIs
```java
    public void runAllTests() {

            createBucket();
            putObject();
            listObjects();
            getObject();
            deleteObjects();
            deleteBucket();
            System.out.println("All tests completed successfully!");
    }
```

Tested with `minio version RELEASE.2025-01-18T00-31-37Z (commit-id=4b6eadbd80313711c01039bfa9a05167291a8c50)
` where the issue existed and all test passed , below config were tested

** With LegacyMd5Plugin **
```
        S3Client s3Client = S3Client.builder()
            .addPlugin(LegacyMd5Plugin.create())
            .endpointOverride(URI.create("http://127.0.0.1:9000"))
            .credentialsProvider(StaticCredentialsProvider.create(
                         AwsBasicCredentials.create("minioadmin", "minioadmin")))
                          .forcePathStyle(true) // Required for MinIO
                           .build();
```

** With LegacyMd5Plugin  with  ChecksumCalculation and  ChecksumValidation  as when required**
```
        S3Client s3Client = S3Client.builder()
            .addPlugin(LegacyMd5Plugin.create())
            .requestChecksumCalculation(RequestChecksumCalculation.WHEN_REQUIRED)
            .responseChecksumValidation(ResponseChecksumValidation.WHEN_REQUIRED)
            .endpointOverride(URI.create("http://127.0.0.1:9000"))
            .credentialsProvider(StaticCredentialsProvider.create(
                         AwsBasicCredentials.create("minioadmin", "minioadmin")))
                          .forcePathStyle(true) // Required for MinIO
                           .build();
```

** Also tested the above test cases and configutation with a real bucket in S3 and test passed **

## Types of changes

## License


services/s3/src/main/java/software/amazon/awssdk/services/s3/LegacyMd5Plugin.java
```diff
@@ -0,0 +1,67 @@
...
+    private LegacyMd5Plugin() {
+    }
+
+    public static SdkPlugin create() {
+        return new LegacyMd5Plugin();
+    }
+
+    @Override
+    public void configureClient(SdkServiceClientConfiguration.Builder config) {
+        S3ServiceClientConfiguration.Builder s3Config = (S3ServiceClientConfiguration.Builder) config;
```
L-Applin, line 59: Should we check it's s3 before casting that the config?

```suggestion
if (!(config instanceof S3ServiceClientConfiguration.Builder)) {
    return;
}
S3ServiceClientConfiguration.Builder s3Config = (S3ServiceClientConfiguration.Builder) config;
```

L-Applin (changes requested):
`architecture-tests` fails because `HttpChecksumUtils` (intenal API in sdk-core) is used in another package (s3). We should either mark this utils class protected api or update the archunit store.

services/s3/src/main/java/software/amazon/awssdk/services/s3/LegacyMd5Plugin.java
```diff
@@ -0,0 +1,67 @@
...
+        s3Config.responseChecksumValidation(ResponseChecksumValidation.WHEN_REQUIRED);
```
zoewangg, line 60: Should we not set it if responseChecksumValidation is already configured? Same for RequestChecksumCalculation (I missed it when I wrote the sample code😅)

services/s3/src/main/java/software/amazon/awssdk/services/s3/LegacyMd5Plugin.java
```diff
@@ -0,0 +1,67 @@
...
+ *       required by the S3 service for specific operations</li>
+ *   <li>Set response checksum validation to WHEN_REQUIRED mode, which validates checksums only when
+ *       the S3 service provides them in responses</li>
+ *   <li>Add an interceptor that maintains backward compatibility with older SDK versions
+ *       that automatically calculated MD5 checksums for certain operations</li>
+ * </ul>
+ *
+ * <p>Use this plugin only when you need to maintain compatibility with applications that depend on the
+ * legacy MD5 checksum behavior, particularly for operations that previously calculated MD5 checksums
+ * automatically.
```
zoewangg, line 41: Can we add sample code of how customers can configure it on the client with `@snippet`?

zoewangg (changes requested):

services/s3/src/main/java/software/amazon/awssdk/services/s3/LegacyMd5Plugin.java
joviegas: Can you you please elaborate , the default values of RequestChecksumCalculation and ResponseChecksumValidation is `WHEN_SUPPORTED` thus this values will always have some values.

Can you please me with a Scenario where this additional check might be helpful ?

joviegas: On second thought, I believe we should not be doing:
```java
s3Config.responseChecksumValidation(ResponseChecksumValidation.WHEN_REQUIRED);
s3Config.requestChecksumCalculation(RequestChecksumCalculation.WHEN_REQUIRED);
        
```    
This is to maintain the single responsibility of the LegacyMd5Plugin, which is solely added to add the MD5 header.

The responseChecksumValidation and requestChecksumCalculation should instead be set on the client or in the environment variables. What do you think?

services/s3/src/main/java/software/amazon/awssdk/services/s3/LegacyMd5Plugin.java
joviegas: Done

joviegas:
> `architecture-tests` fails because `HttpChecksumUtils` (intenal API in sdk-core) is used in another package (s3). We should either mark this utils class protected api or update the archunit store.

Done , removed the HttpChecksumUtils apiand made a inline.

services/s3/src/main/java/software/amazon/awssdk/services/s3/LegacyMd5Plugin.java
```diff
@@ -0,0 +1,79 @@
...
+import software.amazon.awssdk.core.SdkServiceClientConfiguration;
+import software.amazon.awssdk.core.checksums.RequestChecksumCalculation;
+import software.amazon.awssdk.core.checksums.ResponseChecksumValidation;
+import software.amazon.awssdk.services.s3.internal.handlers.LegacyMd5ExecutionInterceptor;
+/**
+ * Plugin that enables legacy MD5 checksum behavior for S3 operations.
+ *
+ * <p>This plugin configures the S3 client to:
+ * <ul>
+ *   <li>Set request checksum calculation to WHEN_REQUIRED mode, which calculates default checksums only when
```
zoewangg, line 29: Looks like we need to update this documentation since we removed configuration

services/s3/src/main/java/software/amazon/awssdk/services/s3/LegacyMd5Plugin.java
```diff
@@ -0,0 +1,79 @@
...
+ * <p>
+ * {@snippet :
+ * // For synchronous S3 client
+ *  S3Client s3Client = S3Client.builder()
+ *                             .addPlugin(LegacyMd5Plugin.create())
+ *                             .build();
+ *
+ *  // For asynchronous S3 client
+ *  S3AsyncClient asyncClient = S3AsyncClient.builder()
+ *                                          .addPlugin(create())
```
zoewangg, line 50: `LegacyMd5Plugin.create()`. Can we configure the request checksum calculation and response checksum validation to when_required in the code snippet?

services/s3/src/main/java/software/amazon/awssdk/services/s3/LegacyMd5Plugin.java
joviegas: done

services/s3/src/main/java/software/amazon/awssdk/services/s3/LegacyMd5Plugin.java
joviegas: Done

zoewangg (approved):

services/s3/src/test/java/software/amazon/awssdk/services/s3/plugin/legacymd5/LegacyMd5PluginOperationTest.java
L-Applin, line 1: Can we add a test cases with the plugin when user manually set md5 on the request body?

joviegas: Added a test case

L-Applin (approved):

steveloughran:
thank you for this!

have you tried testing it against a third party store? google cloud is easy to set up & documented in the s3a thirdparty and qualification docs. Even if you only run it weekly, it'd restore my confidence in the SDK changes.

thanks

joviegas:
> thank you for this!
> 
> have you tried testing it against a third party store? google cloud is easy to set up & documented in the s3a thirdparty and qualification docs. Even if you only run it weekly, it'd restore my confidence in the SDK changes.
> 
> thanks

HI Steve ,
- I tested it with  minio version RELEASE.2025-01-18T00-31-37Z where this issue was easily repeatable.
_ let me do a when off test with  google cloud , thanks for the reference
- `run it weekly` will keep this  in consideration and discuss about it .

uwolfer:
Has somebody already successfully tested this change against GCP? I have enabled the new plugin, but `putObject` still fails with a `SignatureDoesNotMatch` error while it works with the same code with 2.29.52.

joviegas:
Hi  @uwolfer  , 
I have only tested with minio  and it worked fine.
I spent some time using gcp but it was giving me lot of authentication issues.
@bhoradc  did you get a chance to check it out GCP.

Can you please share some simple docs/link/steps which can help me to do a quick test with GCP.

uwolfer:
@joviegas I would appreciate if you could take a look into my question above.

joviegas:
@uwolfer I checked with GCP today.
> Is it really the idea to disable optional checksum checks?

Nope, the idea of LegacyMd5Plugin is to add "Content-MD5" to the S3 APIs which were marked for httpChecksum as required.

Some of the third-party tools like MinIO were always expecting "Content-MD5" to be present and would error if it was absent. With the 2.30.0 release, we removed "Content-MD5" since the SDK was always sending checksums by default for all the APIs which are marked as Required and supported. However, to support the third-party tools that were always expecting "Content-MD5", we had to add this plugin.

One example is [DeleteObjects](https://github.com/aws/aws-sdk-java-v2/blob/eac6dafd8d9d817b32fd6f2fe6c5583a5a2bf65c/services/s3/src/main/resources/codegen-resources/service-2.json#L322-L335) - this API fails in older versions of MinIO since this API is marked as checksum required. However, I saw that this API failed in GCP even prior to 2.30.0 with some other errors.


I was not able to find an API in S3 which was marked as required checksum and worked fine with GCP. Do you have any reference of a API which is marked as Checksum required and worked prior to 2.30.0 , if yes please could you please provide me a code snippet, I can test it out 

Also, if there are any issues with using GCP despite the LegacyMd5Plugin, RequestChecksumCalculation, and ResponseChecksumValidation options, which worked fine prior to 2.30.0 and are not working after 2.30.0, I would recommend raising a new issue, since comments on merged PRs can go unnoticed sometimes.

steveloughran:
> I saw that this API failed in GCP even prior to 2.30.0 with some other errors.

google gcs doesn't do bulk deletes, multipart uploads and ignores if-match/if-none-match headers on PUT

PR #6059, opened 2025-04-24
Added support for s3-event-notifications in migration tool

Fred1155:
## Motivation and Context

This PR added support for s3-event-notifications in migration tool. s3-event-notifications is a separate module in v2, while in v1 it is in s3.event folder. All nested class in S3EventNotification class in v1 is a separated class in v2.
## Modifications

Added transform for all package names.
Added transform for all different method names.
Added transform for different return types (getEventNameAsEnum() and getEventTime())
Added recipe to add wrapper for methods that return DateTime in v1 and Instant in v2
## Testing

Added new integration test class for s3-event-notifications module
## Types of changes

## Checklist

## License


test/v2-migration-tests/src/test/resources/software/amazon/awssdk/v2migrationtests/maven-en/after/src/main/java/foo/bar/S3EventNotificationTest.java
```diff
@@ -0,0 +1,70 @@
...
+            S3Bucket bucket = s3.getBucket();
+
+            S3Object object = s3.getObject();
+
+            String eventName = record.getEventName();
+
+            String eventNameEnum = record.getEventName();
+
+            DateTime eventTime = /*AWS SDK for Java v2 migration: getEventTime returns Instant instead of DateTime in v2. AWS SDK v2 does not include org.joda.time as a dependency. If you want to keep using DateTime, you'll need to manually add "org.joda.time:joda-time" dependency to your project after migration.*/new DateTime(record.getEventTime().toEpochMilli());
+
```
dagnir, line 49: Is it possible to put these comments above the statement?

davidh44: Not easily, we are visiting the method invocation here, and would have to check if there is a variable assignment and add it there instead if so. Even then, putting a newline after the comment causes the code to have no indentation, which is why I didn't do it for other comments.

I'd imagine users would remove all comments regardless so I think its okay

v2-migration/src/main/resources/META-INF/rewrite/aws-sdk-java-v1-to-v2.yml
```diff
@@ -21,16 +21,20 @@ tags:
   - aws
   - sdk
 recipeList:
+  - software.amazon.awssdk.v2migration.AddS3EventNotificationDependency
```
davidh44, line 24: We should update tm recipe as well (which includes all of this recipe)

https://github.com/aws/aws-sdk-java-v2/blob/master/v2-migration/src/main/resources/META-INF/rewrite/aws-sdk-java-v1-to-v2-with-tm.yml

test/v2-migration-tests/src/test/java/software/amazon/awssdk/v2migrationtests/MavenS3EventNotificationTest.java
```diff
@@ -0,0 +1,54 @@
...
+import static software.amazon.awssdk.v2migrationtests.TestUtils.replaceVersion;
+
+import java.io.File;
+import java.io.IOException;
+import org.apache.commons.io.FileUtils;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.condition.EnabledIf;
+
+public class MavenS3EventNotificationTest extends MavenTestBase {
```
davidh44, line 27: Is this separate module because of DateTime needing joda-time? If so, can we test most transforms in `maven` and the DateTime ones in `maven-nocompile`?

test/v2-migration-tests/src/test/resources/software/amazon/awssdk/v2migrationtests/maven-en/after/src/main/java/foo/bar/S3EventNotificationTest.java
dagnir: I really think we should investigate if we can do without too much effort. As is, the long, single line comments are difficult to read; if we're adding comments it's probably because it's re: something the customer should know about so we should make it easier for them to get that information.

Shouldn't we be able to use the indent information from the select (`select.getAfter().getIndent()`) to preserve the formatting?


This is arguably much easier to read for example
```java
        DateTime eventTime = /* AWS SDK for Java v2 migration: getEventTime returns Instant instead of DateTime in v2. 
                                AWS SDK v2 does not include org.joda.time as a dependency. If you want to keep using
                                DateTime, you'll need to manually add "org.joda.time:joda-time" dependency to your 
                                project after migration.
                             */
                             new DateTime(record.getEventTime().toEpochMilli());
```

Fred1155: Currently all added comments are like this. It might require some change the comments recipe, as well as all the integration test cases. We can address this problem in a separate PR.

dagnir: That's fine with me. Can we please create a backlog item?

test/v2-migration-tests/src/test/java/software/amazon/awssdk/v2migrationtests/MavenS3EventNotificationTest.java
Fred1155: Yes. Will do that.

dagnir (approved):

davidh44 (approved):

PR #6087, opened 2025-05-05
URI cache for DynamoDB account id based endpoint

L-Applin:
## Motivation and Context

## Modifications
- Added a `SdkURI` class which caches calls to the various URI constructors methods.
- Added customization configuration to codegen to enable usage of `SdkURI` class in generated endpoint resolution code. Enabled this config only for DynamoDB. 

## Testing
- Unit test 
- Canary testing in progress

## Types of changes


codegen/src/main/java/software/amazon/awssdk/codegen/poet/rules2/CodeGeneratorVisitor.java
```diff
@@ -274,7 +278,11 @@ private void codegenTreeBody(RuleSetExpression expr) {
     @Override
     public Void visitEndpointExpression(EndpointExpression e) {
         builder.add("return $T.endpoint(", typeMirror.rulesResult().type());
-        builder.add("$T.builder().url($T.create(", Endpoint.class, URI.class);
+        if (endpointCaching) {
+            builder.add("$T.builder().url($T.getInstance().create(", Endpoint.class, SdkUri.class);
```
zoewangg, line 282: Can we add a codegen tests? 

Actually, should we just do it for all service clients at this point? I don't really see us removing the logic ever...

.changes/next-release/feature-AWSSDKforJavav2-b405876.json
```diff
@@ -0,0 +1,6 @@
+{
+    "type": "feature",
+    "category": "AWS SDK for Java v2",
+    "contributor": "",
+    "description": "URI cache for DynamoDB account id based endpoint"
```
zoewangg, line 5: Can we describe the behavior rather than implementation?

zoewangg (changes requested):

utils/src/main/java/software/amazon/awssdk/utils/uri/SdkUri.java
```diff
@@ -0,0 +1,287 @@
...
+
+    public URI create(String s) {
+        if (!isAccountIdUri(s)) {
+            log.trace(() -> "skipping cache for uri " + s);
+            return URI.create(s);
+        }
+        StringConstructorArgs key = new StringConstructorArgs(s);
+        boolean containsK = cache.contains(key);
+        URI uri = cache.get(key);
+        logCacheUsage(containsK, uri);
```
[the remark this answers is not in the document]
L-Applin: Yep we can, I'll just need to move around the logs a bit (calling `lruCache.containsKey()` after `get()` would always be true)

codegen/src/main/java/software/amazon/awssdk/codegen/poet/rules2/CodeGeneratorVisitor.java
L-Applin: I would avoid doing it for all service clients to minimize blast radius. But good point on codegen test, I'll ad them.

utils/src/main/java/software/amazon/awssdk/utils/uri/SdkUri.java
```diff
@@ -0,0 +1,287 @@
...
+@SdkProtectedApi
+public final class SdkUri {
+    private static final String HTTPS_PREFIX = "https://";
+    private static final String HTTP_PREFIX = "http://";
+
+    private static final Logger log = Logger.loggerFor(SdkUri.class);
+
+    private static final SdkUri INSTANCE = new SdkUri();
+
+    private final LruCache<UriConstructorArgs, URI> cache = LruCache.builder(UriConstructorArgs::newInstance)
```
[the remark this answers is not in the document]
L-Applin: Spotbugs complain about double checking with the typical lazy initialization for singleton: https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#dc-possible-double-check-of-field-dc-doublecheck

It seems to consider it an anti pattern. The linked article mentions that static initializing it as a field is fine, left it like that. We can discuss more if you want.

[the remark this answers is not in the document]
L-Applin: Using `Lazy<>` for lazy initialization

build-tools/src/main/resources/software/amazon/awssdk/spotbugs-suppressions.xml
```diff
@@ -359,4 +359,8 @@
         <Class name="software.amazon.awssdk.v2migration.EnumCasingToV2$Visitor"/>
         <Bug pattern="NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE"/>
     </Match>
+    <Match>
+        <Class name="software.amazon.awssdk.utils.uri.SdkUri" />
+        <Bug pattern="BC_UNCONFIRMED_CAST_OF_RETURN_VALUE" />
```
zoewangg, line 364: Question: where do we cast?

.changes/next-release/feature-AWSSDKforJavav2-b405876.json
```diff
@@ -0,0 +1,6 @@
+{
+    "type": "feature",
+    "category": "AWS SDK for Java v2",
```
zoewangg, line 3: nit: Amazon DyanmoDB

build-tools/src/main/resources/software/amazon/awssdk/spotbugs-suppressions.xml
L-Applin: here: https://github.com/aws/aws-sdk-java-v2/pull/6087/files#diff-ba4e76ea7c37f0a322912c476915fa2de501d6ff5f3af283be7f54babf846f00R81-R88

utils/src/main/java/software/amazon/awssdk/utils/uri/SdkUri.java
```diff
@@ -0,0 +1,294 @@
...
+    }
+
+    public URI create(String s) {
+        if (!isAccountIdUri(s)) {
+            log.trace(() -> "skipping cache for uri " + s);
+            return URI.create(s);
+        }
+        StringConstructorArgs key = new StringConstructorArgs(s);
+        boolean containsK = cache.containsKey(key);
+        URI uri = cache.get(key);
```
dagnir, line 65: This isn't 100% reliable since the key could have been inserted between 64 and 65. Do we actually need to log this? If there is, should be do it within `LruCache` instead?

utils/src/main/java/software/amazon/awssdk/utils/uri/SdkUri.java
```diff
@@ -0,0 +1,294 @@
...
+            }
+            throw e;
+        }
+    }
+
+    public URI newUri(String scheme,
+                      String authority,
+                      String path, String query, String fragment) throws URISyntaxException {
+        if (!isAccountIdUri(authority)) {
+            log.trace(() -> "skipping cache for authority" + authority);
```
dagnir, line 115: nit: space after authority

utils/src/main/java/software/amazon/awssdk/utils/uri/SdkUri.java
```diff
@@ -0,0 +1,294 @@
...
+     * Best-effort check for uri string being account-id based.
+     *
+     * The troublesome uris are of the form 'https://123456789012.ddb.us-east-1.amazonaws.com' The heuristic chosen to detect such
+     * candidate URI is to check the first char after the scheme, and then the char 10 places further down the string. If both
+     * are digits, there is a potential for that string to represent a number that would exceed the value of Integer.MAX_VALUE,
+     * which would cause the performance degradation observed with such URIs.
+     */
+    private boolean isAccountIdUri(String s) {
+        int firstCharAfterScheme = 0;
+        int maxIntSizeBase10 = 10;
```
dagnir, line 142: Make static constant. Also maybe `maxIntDigitsBase10`

utils/src/main/java/software/amazon/awssdk/utils/uri/SdkUri.java
```diff
@@ -0,0 +1,294 @@
...
+            }
+            throw e;
+        }
+    }
+
+    public URI newUri(String scheme,
+                      String userInfo, String host, int port,
+                      String path, String query, String fragment) throws URISyntaxException {
+        if (!isAccountIdUri(host)) {
+            log.trace(() -> "skipping cache for host" + host);
```
dagnir, line 94: nit: space after host

dagnir (changes requested):

zoewangg (approved):

utils/src/main/java/software/amazon/awssdk/utils/uri/SdkUri.java
L-Applin: we don't actually _need_ to log it, but thought it could help debug if push come to shove.

dagnir: Okay in that case this should be fine, as long as we're not relying on it to be completely accuraet.

dagnir (approved):

PR #6050, opened 2025-04-22
Remove fast unmarshaller dialup code

sugmanue:
## Motivation and Context

Clean up the code to enable the fast unmarshalling code path. Now that the last wave went in we are removing it.

I will create a follow up to clean up the `customization.config` files to avoid that noise in this change.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)

## Checklist

- [ ] I have read the [CONTRIBUTING](https://github.com/aws/aws-sdk-java-v2/blob/master/CONTRIBUTING.md) document
- [ ] Local run of `mvn install` succeeds
- [ ] My code follows the code style of this project
- [ ] My change requires a change to the Javadoc documentation
- [ ] I have updated the Javadoc documentation accordingly
- [ ] I have added tests to cover my changes
- [ ] All new and existing tests passed
- [ ] I have added a changelog entry. Adding a new entry must be accomplished by running the `scripts/new-change` script and following the instructions. Commit the new file created by the script in `.changes/next-release` with your changes.
- [ ] My change is to implement 1.11 parity feature and I have updated [LaunchChangelog](https://github.com/aws/aws-sdk-java-v2/blob/master/docs/LaunchChangelog.md)

## License

- [ ] I confirm that this pull request can be released under the Apache 2 license

PR #6089, opened 2025-05-05
Enable the fast unmarshalling codepath by default

sugmanue:
## Motivation and Context

This is a partial clean up of the fast unmarshalling dialup. This change only enables adding the client config to enable the fast unmarshalling codepath by default.

## Types of changes

## Checklist

## License


joviegas (approved):

sugmanue:
> Since this is True by default , should we also consider removing unused code from JsonProtocolUnmarshaller ?

We tried that on #6050 but that broke some customers that are using newer core libraries with old clients. We don't support that use case but we need to find out a better way of preventing it so meanwhile we will keep this for a while.

PR #6095, opened 2025-05-08
Fix S3 event parsing when eventName is present but eventTime is not

reifiedbeans:
## Motivation and Context
Currently, `S3EventNotification::fromJson` throws an exception when `eventTime` is `null`. This is problematic, since the code intends to allow for this field to be `null`, but [it uses the wrong variable when performing a guard](https://github.com/aws/aws-sdk-java-v2/blob/cb7f5c5d692ecf657e9545a48b02e78d7a132b2c/services-custom/s3-event-notifications/src/main/java/software/amazon/awssdk/eventnotifications/s3/internal/DefaultS3EventNotificationReader.java#L104).

## Modifications
This patch fixes the `null` guard by using the correct variable, `eventTime`, before attempting to parse it as an `Instant`.

## Testing
Built the project locally (couldn't build some modules locally, so I just built the one modified)

```shell
./mvnw --projects :s3-event-notifications --also-make install
```

N/A

## Types of changes

## Checklist

## License

### DIFF
--- a/.changes/next-release/bugfix-AWSS3EventNotifications-3774889.json
+++ b/.changes/next-release/bugfix-AWSS3EventNotifications-3774889.json
@@ -1,0 +1,6 @@
+{
+    "type": "bugfix",
+    "category": "AWS S3 Event Notifications",
+    "contributor": "reifiedbeans",
+    "description": "Fixes parsing of S3 event notifications to allow eventTime to be null when eventName is not"
+}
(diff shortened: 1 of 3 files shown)


debora-ito (changes requested):
Hi @reifiedbeans thank you for the fix.

Can you add a test case? We should have a test to catch this in the first place :)

debora-ito (approved):

L-Applin (approved):

debora-ito:
@all-contributors please add @reifiedbeans for code.

PR #6099, opened 2025-05-13
Add CI for api surface area review verification

zoewangg:
## Motivation and Context
Add CI for api surface area review verification. It detects protected/public api changes by checking the code and exclude files in "internal", "codegen" or "test" folders. If there is protected/public api change, it requires `no-api-surface-area-change` or `api-surface-area-change` label to be added


.github/workflows/api-surface-area-review-verification.yml
```diff
@@ -0,0 +1,36 @@
...
+  api-surface-area-review-verification:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Verifies updates to protected/public APIs have been reviewed and approved by the team, if any
+        id: api-surface-area-review-verification
+        if: ${{ !contains(github.event.pull_request.labels.*.name, 'no-api-surface-area-change') }}
+        run: |
+          git fetch origin ${{ github.base_ref }} --depth 1
+          FILES=$( git diff remotes/origin/${{ github.base_ref }} --name-only | grep "\.java$" | grep -v -E "(^|/)(internal|test|codegen|v2-migration)/"  || true)
```
joviegas, line 19: Nice CI action +1
I had below suggestions
1. Would it be possible to narrow down the checks to only files containing @SdkPublic and SdkProtectedInterface annotations?
2. Could we enhance the output to include the specific file names that failed the check?
3. Would it be possible to add a tag to the PR using `github.rest.issues.addLabels` ? This way we can mark it as 'api-surface-area-review-required'

zoewangg: 1. It's just more complexity and maintenance to inspect the exact diff for each file and may increase the check time. IMO we really shouldn't expand on `internal|test|codegen|v2-migration` since this repo is already huge, so I'm not sure if the effort it worth it.
2. Yes, it is in the output.https://github.com/aws/aws-sdk-java-v2/pull/6099/files#diff-03e23d2b8990386383cd6d9efa82ad2dd53af6d72b4f33a6be1d0932442b71c1R21
3. I actually thought about it and tried with the GH labeler action, I didn't go with that because it's a bit trickier to flag false positives since we want it to always automatically add `api-surface-area-review-required` (it's not straightforward to tell it to not re-add the label for new revision if the label was removed previously) and having both `api-surface-area-review-required` and `no-api-surface-area-change` is just confusing

joviegas (approved):

PR #6101, opened 2025-05-13
Copy only req extra files if using compiled rules

dagnir:
## Motivation and Context
When generating compiled endpoint rules, the majority of the classes used for the older intepreted rules don't need to be copied. To save on JAR space, only copy the files that are required.

This is part of slimming down our JARs.

## Types of changes

## Checklist

## License


codegen/src/main/java/software/amazon/awssdk/codegen/emitters/tasks/RulesEngineRuntimeLiteGeneratorTask.java
```diff
@@ -0,0 +1,49 @@
...
+
+/** A version of {@link software.amazon.awssdk.codegen.emitters.tasks.RulesEngineRuntimeGeneratorTask} that copies a minimal
+ * set of the interpreter related classes. This set represents the only classes that need to be copied when compiled rules are
+ * enabled.
+ *
+ * @see CustomizationConfig#isEnableGenerateCompiledEndpointRules()
+ */
+public final class RulesEngineRuntimeLiteGeneratorTask extends RulesEngineRuntimeGeneratorTask {
+    private static final List<String> FILES_TO_COPY = Stream.of("Outputs.java.resource",
+                                                                "RegionOverride.java.resource",
```
alextwoods, line 33: I think we may need a few more files than this - I think at least the Arn class is still needed.

This may be out of scope for this change, and. maybe this doesn't matter if we eventually refactor this to completely remove the runtime generation paths - but is there any way to automatically get this list and have it be based on the endpoint features required - ie, Arn I think may only be needed when the service's rules use it.

As an aside - is there a reason these are copied to generated service libraries instead of provided in a core library?

alextwoods (changes requested):

codegen/src/main/java/software/amazon/awssdk/codegen/emitters/tasks/RulesEngineRuntimeLiteGeneratorTask.java
```diff
@@ -0,0 +1,50 @@
...
+
+/** A version of {@link software.amazon.awssdk.codegen.emitters.tasks.RulesEngineRuntimeGeneratorTask} that copies a minimal
+ * set of the interpreter related classes. This set represents the only classes that need to be copied when compiled rules are
+ * enabled.
+ *
+ * @see CustomizationConfig#isEnableGenerateCompiledEndpointRules()
+ */
+public final class RulesEngineRuntimeLiteGeneratorTask extends RulesEngineRuntimeGeneratorTask {
+    private static final List<String> FILES_TO_COPY = Stream.of("/Outputs.java.resource",
+                                                                "/RegionOverride.java.resource",
```
alextwoods, line 33: Requiring the  leading `/` is slightly confusing - are we including it because a file like `Arn.java.resource` could also match something like `<otherprefix>Arn.java.resource`?

If thats true - could we either add a comment about it or in the filter below check the fileName only (ie, use something like [Path#getFileName](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Path.html#getFileName--))

codegen/src/main/java/software/amazon/awssdk/codegen/emitters/tasks/RulesEngineRuntimeLiteGeneratorTask.java
dagnir: > I think we may need a few more files than this - I think at least the Arn class is still needed.

Yep, working on it. Didn't expect some services to use different subsets!

> This may be out of scope for this change, and. maybe this doesn't matter if we eventually refactor this to completely remove the runtime generation paths - but is there any way to automatically get this list and have it be based on the endpoint features required - ie, Arn I think may only be needed when the service's rules use it.

Yeah the eventual goal is to remove everything from the "legacy" rules runtime so I don't want o invest too much time into this.

> As an aside - is there a reason these are copied to generated service libraries instead of provided in a core library?

Yep - putting them in core would mean making them protected API's, and much harder to remove as a result. We knew that we would remove the interpreter code at some point which is why they're generated as internal classes instead.

codegen/src/main/java/software/amazon/awssdk/codegen/emitters/tasks/RulesEngineRuntimeLiteGeneratorTask.java
dagnir: Yep, exactly right, to disambiguate between files that share the same suffix. I thought about using `Path` but the current codepaths use String so this was simpler. I'll add a comment!

alextwoods (approved):

Issue #2542, opened 2021-06-18
[DynamoDbEnhancedClient] Support to flatten a Map into top level attributes of the object.


ghost:
In my usecase, my dynamodb table has a fixed set of fields which are defined as fields in the class. There are a arbitrary fields which are represented by a Map<String, String>. I need the ability to use `@DynamoDbFlatten` on the Map similar to using it on a class.

## Describe the Feature
Support `@DyanmoDbFlatten` annotation for Map as well.

## Is your Feature Request related to a problem?
Without this feature I won't be able to use DynamoDbEnhancedClient

## Describe alternatives you've considered
The only alternatives to this I can think of is to not use the DynamoDbEnhanced client at all.

## Additional Context
The table we use has a fixed fields which can be represented using class members.
However, there are arbitrary set of top level fields which need to be returned in a Map.

- [ ] I may be able to implement this feature request

## Your Environment

* AWS Java SDK version used:
* JDK version used:
* Operating System and version:

debora-ito:
@rakshith-vzm thank you for reaching out, sorry for the delayed response. Marking as a feature request.

DianaElenaGusa:
Hi! I am interested in working on this issue and, after conducting an initial analysis, I have identified some key areas that require clarification regarding the expected behavior of _@DynamoDbFlatten_ for Map types in V2. To ensure the implementation aligns with the intended design, I would like to clarify the following points.

**Findings from V1**
- In V1, Map types are stored as a single JSON object within a column. The column name corresponds to the map attribute's name, and the value contains the serialized JSON.
- Unlike nested objects, there is no mechanism in V1 to dynamically create columns based on map keys.

**Expected Behavior in V2**
My understanding is that V2 aims to introduce a new annotation that enables flattening maps by dynamically adding attributes/columns to the table schema. Specifically:
- Each key in the map would become a column name, with its corresponding value stored as the column value.
- However, this raises several design considerations that require clarification.

**Key Questions**

**1. Dynamic Column Creation**
- Should the new annotation support dynamically adding attributes/columns based on the keys in the map?

**2. Handling Multiple Maps in a Model**
- Can multiple maps exist in a single model?
- If so, how should deserialization work?
- A major challenge is that dynamically generated attributes may not have a clear association with a specific map if multiple maps exist in the model.

**3. Supported Map Types**
- Should this feature support only Map<String, String>? Or should it extend to Map<String, Object>?

**Potential Deserialization Approaches**
[...]

PR #6102, opened 2025-05-14
DDBEnhanced - Support to flatten a Map into top level attributes of the object

anasatirbasa:
Added the facility of using an annotation that will flatten a map of attributes in a class,
similar to what @DynamoDbFlatten does for attributes from another class.

## Description

Added a new annotation named DynamoDbFlattenMap that can be applied at method level for only a single map of attributes within a class (Map<String, String>). If used on multiple maps of attributes, the serialization will throw an exception.

## Motivation and Context
[#2542](https://github.com/aws/aws-sdk-java-v2/issues/2542)

## Modifications
Followed the same approach as for DybamoDbFlatten annotation,  that is used to flatten all the attributes of a separate DynamoDb bean that is stored in the current bean

## Testing
The changes have already been tested by running the existing tests and also added new unit/integration tests for the new flow.

## Types of changes

## Checklist

## License


.changes/next-release/feature-AmazonDynamoDBEnhancedClient-22723fc.json
```diff
@@ -0,0 +1,6 @@
+{
+    "type": "feature",
+    "category": "Amazon DynamoDB Enhanced Client",
+    "contributor": "",
+    "description": "Added the support to flatten a Map into top level attributes of the object"
```
alextwoods: Nit: mention the annotation name here.  Maybe:
```
Add support for @DynamoDBFlattenMap to flatten a Map to top level attributes of an object
```

services-custom/dynamodb-enhanced/README.md
```diff
@@ -685,3 +685,53 @@ private static final StaticTableSchema<Customer> CUSTOMER_TABLE_SCHEMA =
 ```
 Just as for annotations, you can flatten as many different eligible classes as you like using the
 builder pattern. 
+
+#### Using composition
```
alextwoods: There is already a "Using composition" header above this (so this already falls under that section).  

I think ths should be its own `###` level header instead - maybe "### Flattening map attributes"

services-custom/dynamodb-enhanced/README.md
```diff
@@ -685,3 +685,53 @@ private static final StaticTableSchema<Customer> CUSTOMER_TABLE_SCHEMA =
...
+
+Using composition, the @DynamoDBFlattenMap annotation support to flatten a Map:
```
alextwoods: this reads a bit weirdly, consider rephrasing.

services-custom/dynamodb-enhanced/README.md
```diff
@@ -685,3 +685,53 @@ private static final StaticTableSchema<Customer> CUSTOMER_TABLE_SCHEMA =
...
+    public void setCity(String city) { this.city = city;}
+    public String getAddress() { return this.address; }
+    public void setAddress(String address) { this.name = address;}
+
+    @DynamoDbFlattenMap
+    public Map<String, String> getDetailsMap() { return this.detailsMap; }
+    public void setDetailsMap(Map<String, String> record) { this.detailsMap = detailsMap;}
+}
+```
+You can flatten only one map present on a record, otherwise it will be thrown an exception
```
alextwoods: May want to add that this includes maps on any flattened classes as well.

Phrasing nit: "otherwise it will be thrown an exception" -> "otherwise an exception will be thrown".

Additionally - are there other constraints we should mention here?  What key and value types are supported?  It is only <String, String>?

services-custom/dynamodb-enhanced/README.md
```diff
@@ -685,3 +685,53 @@ private static final StaticTableSchema<Customer> CUSTOMER_TABLE_SCHEMA =
...
+    public String getName() { return this.name; }
+    public void setName(String name) { this.name = name;}
+    public String getCity() { return this.city; }
+    public void setCity(String city) { this.city = city;}
+    public String getAddress() { return this.address; }
+    public void setAddress(String address) { this.name = address;}
+
+    @DynamoDbFlattenMap
+    public Map<String, String> getDetailsMap() { return this.detailsMap; }
+    public void setDetailsMap(Map<String, String> record) { this.detailsMap = detailsMap;}
```
alextwoods: parameter name `record` here is incorrect, it should be `detailsMap`.
```suggestion
    public void setDetailsMap(Map<String, String> detailsMap) { this.detailsMap = detailsMap;}
```

.changes/next-release/feature-AmazonDynamoDBEnhancedClient-22723fc.json
alextwoods:
```suggestion
    "description": "Add support for @DynamoDBFlattenMap to flatten a Map to top level attributes of an object"
```

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/mapper/BeanTableSchema.java
```diff
@@ -222,6 +224,14 @@ private static <T> StaticTableSchema<T> createStaticTableSchema(Class<T> beanCla
             throw new IllegalArgumentException(e);
         }
 
+        List<PropertyDescriptor> mappableProperties = Arrays.stream(beanInfo.getPropertyDescriptors())
+                                                            .filter(p -> isMappableProperty(beanClass, p))
+                                                            .collect(Collectors.toList());
+
+        if (dynamoDbFlattenMapAnnotationHasInvalidUse(mappableProperties)) {
```
alextwoods: What about moving this to a `validateDynamoDbFlattenMapAnnotations` which would throw the specific exceptions (thinking about what we would do if we wanted to add additional validations in the future - with the current method there isn't an easy way to pass information about how the usage is invalid)

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/mapper/BeanTableSchema.java
```diff
@@ -286,6 +302,15 @@ private static Optional<Method> findFluentSetter(Class<?> beanClass, String prop
                      .findFirst();
     }
 
+    private static boolean dynamoDbFlattenMapAnnotationHasInvalidUse(List<PropertyDescriptor> mappableProperties) {
```
alextwoods: Right now we only support Map<String,String> right?  How is that validated?

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/mapper/BeanTableSchema.java
```diff
@@ -286,6 +302,15 @@ private static Optional<Method> findFluentSetter(Class<?> beanClass, String prop
...
+        return mappableProperties.stream()
```
alextwoods: What about cases where `@ DynamoDbFlatten` is used and the other class also has a @ DynamoDBFlattenMap annotation?  Would we catch those here?

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/mapper/StaticImmutableTableSchema.java
```diff
@@ -144,7 +145,57 @@ private AttributeValue attributeValue(T item, String attributeName) {
             return isNullAttributeValue(attributeValue) ? null : attributeValue;
         }
     }
-    
+
+    private static final class FlattenedMapperForMaps<T, B> {
```
alextwoods, line 149: Naming nit - this is slightly unwieldy - how about `FlattenedMapMapper`

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/mapper/StaticImmutableTableSchema.java
```diff
@@ -144,7 +145,57 @@ private AttributeValue attributeValue(T item, String attributeName) {
...
+            }
+            this.mapItemSetter.accept(thisBuilder, flattenedMap);
+
+            return thisBuilder;
+        }
+
+        private Map<String, AttributeValue> itemToMap(T item, boolean ignoreNulls, List<String> attributeNames) {
+            Map<String, String> mapItem = this.mapItemGetter.apply(item);
+
+            Map<String, AttributeValue> result  = new HashMap<>();
```
alextwoods: white space nit - make sure you've run checkstyle.
```suggestion
            Map<String, AttributeValue> result = new HashMap<>();
```

services-custom/dynamodb-enhanced/README.md
```diff
@@ -685,3 +685,53 @@ private static final StaticTableSchema<Customer> CUSTOMER_TABLE_SCHEMA =
...
+
+private static final StaticTableSchema<Customer> CUSTOMER_TABLE_SCHEMA =
+  StaticTableSchema.builder(Customer.class)
+    .newItemSupplier(Customer::new)
+    .addAttribute(String.class, a -> a.name("name")
+                                      .getter(Customer::getName)
+                                      .setter(Customer::setName))
+    // Because we are flattening a Map object, we supply a getter and setter so the
+    // mapper knows how to access it
+    .flattenMap(Map::detailsMap, Map::detailsMap)
```
alextwoods:
```suggestion
    .flattenMap(Map::getDetailsMap, Map::setDetailsMap)
```

services-custom/dynamodb-enhanced/src/test/java/software/amazon/awssdk/enhanced/dynamodb/functionaltests/FlattenMapTest.java
```diff
@@ -0,0 +1,170 @@
...
+
+        persistedRecord = mappedTable.getItem(record);
+        assertThat(persistedRecord.getId()).isEqualTo("444");
+        assertThat(persistedRecord.getRootAttribute1()).isNull();
+        assertThat(persistedRecord.getRootAttribute2()).isNull();
+        assertThat(persistedRecord.getAttributesMap()).isNull();
+    }
+
+    @Test
+    public void updateItemWithFlattenMap_withDuplicateAttributeName_throwsIllegalArgumentException() {
```
alextwoods, line 157: Can we also add a similar test for behavior with the FlattenMapAndFlattenRecordBean with a duplicate attribute in the flattened record?

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/mapper/BeanTableSchema.java
alextwoods: Sorry more validation questions - what about cases where multiple traits are applied, eg:

```
@DynamoDbUpdateBehavior(UpdateBehavior.WRITE_IF_NOT_EXISTS)
@DynamoDbFlattenMap
public Map<String, String> getDetailsMap() { return this.detailsMap; }
```

services-custom/dynamodb-enhanced/src/it/java/software/amazon/awssdk/enhanced/dynamodb/model/Record.java
```diff
@@ -27,6 +28,8 @@ public class Record {
 
     private String stringAttribute;
 
+    private Map<String, String> attributesMap;
```
alextwoods, line 31: This is a generic question - not specifically related to this line only.

We are only supporting maps of <String, String> with this change right?  

How could we extend this to support different value types long term - at minimum I'd think the standard types we support, potentially even allowing custom attribute converters to be applied.  

I think releasing this with only String values is probably reasonable, but I want to make sure that we could add more support in the future without it being a breaking change.

services-custom/dynamodb-enhanced/README.md
anasatirbasa: I've updated the README file.

.changes/next-release/feature-AmazonDynamoDBEnhancedClient-22723fc.json
anasatirbasa: The description was changed.

services-custom/dynamodb-enhanced/README.md
anasatirbasa: I've changed the parameter name to "detailsMap".

services-custom/dynamodb-enhanced/README.md
anasatirbasa: Readme file updated

services-custom/dynamodb-enhanced/README.md
anasatirbasa: Yes, it's only Map<String, String> - both key and value must be String type. Other constraints include: only one flattened map per class hierarchy, map keys can't conflict with existing attribute names, and validation occurs at schema creation time throwing exceptions for violations.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/mapper/StaticImmutableTableSchema.java
anasatirbasa: Fixed

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/mapper/BeanTableSchema.java
anasatirbasa: Updated the README documentation to clarify that additional annotations such as @DynamoDbUpdateBehavior are not supported on flattened maps, ensuring consistency with the existing object flattening implementation where such annotations are not applied to flattened properties.

services-custom/dynamodb-enhanced/src/test/java/software/amazon/awssdk/enhanced/dynamodb/functionaltests/FlattenMapTest.java
anasatirbasa: Added a new test covering this scenario: updateItemWithFlattenMap_duplicateAttributeInFlattenedRecord_throwsIllegalArgumentException.

services-custom/dynamodb-enhanced/README.md
anasatirbasa: Updated README file.

anasatirbasa:
Hello @alextwoods,

Thank you very much for your feedback, we have addressed the comments from the pull request. 

After further analysis, we changed from the original proposal of separate annotations to a unified @DynamoDbFlatten approach that handles both objects and maps. We chose this design for better API consistency. You can find the updated Solution Proposal [here](https://github.com/gorelov1/ddb-endava/blob/main/DesignDocuments/Java/%5BIssue2542%5DDynamoDbFlattenMap.md).

Could you please review this approach and provide your feedback?
Thank you very much!

alextwoods:
It looks like checkstyle is failing - make sure you run checkstyle locally.

services-custom/dynamodb-enhanced/src/it/java/software/amazon/awssdk/enhanced/dynamodb/ScanQueryWithFlattenMapIntegrationTest.java
```diff
@@ -0,0 +1,117 @@
...
+                                                  .limit(3)
+                                                  .build())
+                       .iterator();
+
+        Page<Record> page1 = results.next();
+        assertThat(results.hasNext(), is(true));
+        Page<Record> page2 = results.next();
+        assertThat(results.hasNext(), is(false));
+
+        Map<String, String> resultedAttributesMap = new HashMap<>();
```
alextwoods: naming nit - maybe `exepctedAttributesMap` instead

services-custom/dynamodb-enhanced/src/it/java/software/amazon/awssdk/enhanced/dynamodb/DynamoDbEnhancedIntegrationTestBase.java
```diff
@@ -87,6 +118,22 @@ protected static DynamoDbAsyncClient createAsyncDynamoDbClient() {
                      .setGsiSort(i))
                  .collect(Collectors.toList());
 
+    protected static final List<Record> RECORDS_WITH_FLATTEN_MAP =
```
alextwoods: These seem specific to just one test class - I'd lean towards moving these into that test class (ditto on the schema definition above).

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/mapper/annotations/DynamoDbFlatten.java
```diff
@@ -22,9 +22,26 @@
...
+ * <ul>
+ *   <li>If the property is a {@code Map<String, String>}, map flattening is used</li>
+ *   <li>If the property is a DynamoDb bean class, object flattening is used</li>
+ * </ul>
+ * <p>
+ * <b>Constraints:</b>
+ * <ul>
+ *   <li>Only one map flattening is allowed per class hierarchy</li>
+ *   <li>For map flattening, the property must be of type {@code Map<String, String>}</li>
+ *   <li>Attribute names from flattened content must not conflict with existing attributes</li>
```
alextwoods: Nit - since this applies to both object and map flattening, make it the first constraint in this list.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/mapper/StaticImmutableTableSchema.java
```diff
@@ -246,6 +303,7 @@ public static final class Builder<T, B> {
         private final List<ResolvedImmutableAttribute<T, B>> additionalAttributes = new ArrayList<>();
         private final List<FlattenedMapper<T, B, ?>> flattenedMappers = new ArrayList<>();
 
+        private FlattenedMapperForMaps<T, B> flattenedMap;
```
alextwoods: To disambiguate these - maybe use `flattenedObjectMappers` and `flattenedMapMapper`

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/mapper/annotations/DynamoDbFlatten.java
alextwoods: Additionally - generally the other constraints are all validated at schema creation time right?  However, with a flattened map, duplicate attribute names will only be detected when on an operation (like update) is performed right?  If so - we should make this difference clear here.

services-custom/dynamodb-enhanced/src/test/java/software/amazon/awssdk/enhanced/dynamodb/functionaltests/FlattenMapTest.java
```diff
@@ -0,0 +1,214 @@
...
+        mappedTable.createTable(r -> r.provisionedThroughput(getDefaultProvisionedThroughput()));
+    }
+
+    @After
+    public void deleteTable() {
+        getDynamoDbClient().deleteTable(r -> r.tableName(getConcreteTableName("table-name")));
+    }
+
+    @Test
+    public void updateItemWithFlattenMap_correctlyFlattensMapAttributes() {
```
alextwoods, line 72: Can we add a similar test case for a flattened map on a nested/flattened class?

services-custom/dynamodb-enhanced/src/test/java/software/amazon/awssdk/enhanced/dynamodb/functionaltests/FlattenMapTest.java
```diff
@@ -0,0 +1,214 @@
...
+            assertThatThrownBy(() -> composedTable.updateItem(flattenMapAndFlattenRecord))
+                .isInstanceOf(IllegalArgumentException.class)
+                .hasMessageContaining("Attempt to add an attribute to a mapper that already has one with the same name. ");
+        } finally {
+            getDynamoDbClient().deleteTable(r -> r.tableName(composedTableName));
+        }
+    }
+
+    @Test
+    public void updateItemWithFlattenMap_withMultipleAnnotatedMaps_throwsIllegalArgumentException() {
```
alextwoods, line 200: Can we add a similar test case for multiple flattened maps on nested/flattened classes?

services-custom/dynamodb-enhanced/src/it/java/software/amazon/awssdk/enhanced/dynamodb/model/Record.java
alextwoods: Un-resolving this comment - I understand we're only support Map<String,String> with this change - and I agree thats the right support to launch with.  However, I want to make sure this isn't a one-way decision that would prevent us from supporting maps of other types.

services-custom/dynamodb-enhanced/src/test/java/software/amazon/awssdk/enhanced/dynamodb/functionaltests/FlattenMapTest.java
anasatirbasa: Yes. I've added a new test for this scenario in FlattenMapBeanTableSchemaTest: **serializeBean_withNestedMultipleFlattenMaps_throwsIllegalArgumentException**.

services-custom/dynamodb-enhanced/src/test/java/software/amazon/awssdk/enhanced/dynamodb/functionaltests/FlattenMapTest.java
anasatirbasa: Yes. I've added a new test for this scenario in FlattenMapTest: **updateItemWithNestedFlattenMap_correctlyFlattensMapAttributes**.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/mapper/StaticImmutableTableSchema.java
anasatirbasa: Updated the names for these two.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/mapper/annotations/DynamoDbFlatten.java
anasatirbasa: Yes, flattened attribute names conflicts will be evaluated at runtime — for maps and at schema creation — for objects. I've updated the Javadoc in order to reflect this.

services-custom/dynamodb-enhanced/src/it/java/software/amazon/awssdk/enhanced/dynamodb/DynamoDbEnhancedIntegrationTestBase.java
anasatirbasa: You're right, I've moved these to the specific test class — ScanQueryWithFlattenMapIntegrationTest.

services-custom/dynamodb-enhanced/src/it/java/software/amazon/awssdk/enhanced/dynamodb/ScanQueryWithFlattenMapIntegrationTest.java
anasatirbasa: I've changed the variable name to “expectedAttributesMap”.

services-custom/dynamodb-enhanced/src/it/java/software/amazon/awssdk/enhanced/dynamodb/model/Record.java
anasatirbasa: Yes, we're starting with Map<String, String> but we can extend to other types without breaking changes.

The current  FlattenedMapperForMaps class can be generalized to **FlattenedMapperForMapsT, B, V>** where V is the value type, and we'd add overloaded builder methods like:

- .flatten(mapName, getter, setter, Integer.class) for primitives
- .flatten(mapName, getter, setter, customConverter) for complex objects

The existing .flatten(mapName, getter, setter) method would remain unchanged for backward compatibility.

anasatirbasa:
Hello @alextwoods,

Thank you very much for the review! I have addressed the comments raised and resolved the checkstyle issues.
Could you please take another look when you have a chance? Thank you!

PR #6118, opened 2025-05-16
Add script to automatically set up a new module

zoewangg:
## Motivation and Context
Add script to automatically set up a new module based on the checklist.


scripts/setup-new-module
```diff
@@ -0,0 +1,352 @@
+#!/bin/bash
```
alextwoods, line 1: It looks like almost all of the other scripts are in python (except weirdly the minor-version-bump, but its fairly simple).  Is there a reason to do this in bash instead?  Personally I tend to find longer bash scripts to be somewhat more difficult to read.

scripts/setup-new-module
```diff
@@ -0,0 +1,352 @@
...
+
+    # Determine the module path based on whether it's a test module
+    local module_path="$module_name"
+    if [ "$IS_TEST_MODULE" = true ]; then
+        module_path="test/$module_name"
+    fi
+
+    # Check if module already exists
+    if grep -q "<module>$module_path</module>" "$root_pom"; then
+        echo "Module already exists in root pom.xml. Skipping."
```
alextwoods, line 185: When would this happen and is this something that should be an error?

scripts/setup-new-module
```diff
@@ -0,0 +1,352 @@
...
+    
+    # Update .brazil.json
+    update_brazil_json "$MODULE_NAME" true
+    
+    # Update root pom.xml
+    update_root_pom "$MODULE_NAME"
+fi
+
+echo ""
+echo "Module setup complete! Please review the changes and complete any manual steps mentioned above."
```
alextwoods, line 352: I'm not sure how much output this produces, but it might be good to collect the required manual steps and provide them all together here so they are less likely to be missed. (feel free to ignore if its never much output).

scripts/setup-new-module
zoewangg: It could happen when the user has already set up some files manually because they may not know this script exist when they created the new module

scripts/setup-new-module
zoewangg: No specific reason, agreed it's more readable in python. Converted to python

scripts/setup-new-module
zoewangg: There's actually no manual steps anymore :) Will update it

scripts/setup-new-module
```diff
@@ -0,0 +1,406 @@
...
+
+
+def create_pom_xml(module_dir, module_name, sdk_version, is_test_module):
+    """Create a basic pom.xml file for the module."""
+    pom_path = os.path.join(module_dir, 'pom.xml')
+    
+    print(f"Using SDK version: {sdk_version}")
+    
+    with open(pom_path, 'w') as f:
+        f.write(f'''<?xml version="1.0" encoding="UTF-8"?>
```
alextwoods, line 62: Nit: put the start of text on new line for readability and consistency with other text below

alextwoods (approved):
Nice! Python version is much easier to read (at least for me) :)

PR #6120, opened 2025-05-20
Use partitions.json to generate Region class

S-Saranya1:
Updated Region class generation to use Partitions.json instead of the Endpoints.json and removed the hardcoded global regions.

## Motivation and Context
Currently we are using deprecated endpoints.json to generate Region class and using hardcoded global regions.
Switching to use partisions.json.resource to generate Region class which contains all regions.

## Modifications
Updated Region class generation to use partisions.json.resource file
Removed the hardcoded global regions
Added tests 

## Types of changes

## Checklist

## License

### DIFF
--- a/codegen-lite-maven-plugin/src/main/java/software/amazon/awssdk/codegen/lite/maven/plugin/RegionGenerationMojo.java
+++ b/codegen-lite-maven-plugin/src/main/java/software/amazon/awssdk/codegen/lite/maven/plugin/RegionGenerationMojo.java
@@ -59,19 +59,24 @@ public class RegionGenerationMojo extends AbstractMojo {
         "${basedir}/src/main/resources/software/amazon/awssdk/regions/internal/region/endpoints.json")
     private File endpoints;
 
+    @Parameter(property = "partitionsjson", defaultValue =
+        "${basedir}/src/main/resources/software/amazon/awssdk/regions/internal/region/partitions.json.resource")
+    private File partitionsjson;
+
     @Override
     public void execute() throws MojoExecutionException {
         Path baseSourcesDirectory = Paths.get(outputDirectory).resolve("generated-sources").resolve("sdk");
         Path testsDirectory = Paths.get(outputDirectory).resolve("generated-test-sources").resolve("sdk-tests");
 
         Partitions partitions = RegionMetadataLoader.build(endpoints);
+        Partitions regionpartitions = RegionMetadataLoader.build(partitionsjson);
 
         generatePartitionMetadataClass(baseSourcesDirectory, partitions);
-        generateRegionClass(baseSourcesDirectory, partitions);
+        generateRegionClass(baseSourcesDirectory, regionpartitions);
         generateServiceMetadata(baseSourcesDirectory, partitions);
-        generateRegions(baseSourcesDirectory, partitions);
+        generateRegions(baseSourcesDirectory, regionpartitions);
         generatePartitionProvider(baseSourcesDirectory, partitions);
-        generateRegionProvider(baseSourcesDirectory, partitions);
+        generateRegionProvider(baseSourcesDirectory, regionpartitions);
         generateServiceProvider(baseSourcesDirectory, partitions);
         generateEndpointTags(baseSourcesDirectory, partitions);
 
--- a/codegen-lite/src/main/java/software/amazon/awssdk/codegen/lite/regions/model/Partition.java
+++ b/codegen-lite/src/main/java/software/amazon/awssdk/codegen/lite/regions/model/Partition.java
@@ -62,6 +62,16 @@ public final class Partition {
      */
     private Endpoint defaults;
 
+    /**
+     * The partition id.
+     */
+    private String id;
+
+    /**
+     * Configuration outputs for the partition.
+     */
+    private PartitionOutputs outputs;
+
     public Partition() {
     }
 
@@ -69,10 +79,44 @@ public final class Partition {
                      @JsonProperty(value = "regions") Map<String, PartitionRegion>
                              regions,
                      @JsonProperty(value = "services") Map<String,
-                             Service> services) {
+                             Service> services,
+                     @JsonProperty(value = "id") String id,
+                     @JsonProperty(value = "outputs") PartitionOutputs outputs,
+                     @JsonProperty(value = "regionRegex") String regionRegex) {
         this.partition = Validate.paramNotNull(partition, "Partition");
         this.regions = regions;
         this.services = services;
+        this.id = id;
+        this.outputs = outputs;
+        this.regionRegex = regionRegex;
+    }
+
+    /**
+     * Returns the partition ID.
+     */
+    public String getId() {
+        return id;
+    }
+
+    /**
+     * Sets the partition id.
+     */
+    public void setId(String id) {
+        this.id = id;
+    }
+
+    /**
+     * Returns the configuration outputs for the partition.
+     */
+    public PartitionOutputs getOutputs() {
+        return outputs;
+    }
+
+    /**
+     * Sets the configuration outputs for the partition.
+     */
+    public void setOutputs(PartitionOutputs outputs) {
+        this.outputs = outputs;
     }
 
     /**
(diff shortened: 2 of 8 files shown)


codegen-lite-maven-plugin/src/main/java/software/amazon/awssdk/codegen/lite/maven/plugin/RegionGenerationMojo.java
```diff
@@ -59,19 +59,24 @@ public class RegionGenerationMojo extends AbstractMojo {
         "${basedir}/src/main/resources/software/amazon/awssdk/regions/internal/region/endpoints.json")
     private File endpoints;
 
+    @Parameter(property = "partitionsjson", defaultValue =
```
alextwoods, line 62: Is there a reason the `j` is lower case? (ie, why not `partitionsJson`?

codegen-lite/src/main/java/software/amazon/awssdk/codegen/lite/regions/model/Partition.java
```diff
@@ -62,17 +62,61 @@ public final class Partition {
...
+
     public Partition() {
     }
 
     public Partition(@JsonProperty(value = "partition") String partition,
                      @JsonProperty(value = "regions") Map<String, PartitionRegion>
                              regions,
                      @JsonProperty(value = "services") Map<String,
-                             Service> services) {
+                             Service> services,
```
alextwoods, line 82: I feel like its a little weird to squish both the legacy endponts.json partitions and the new partitions metadata (partitions.json) model together - How much of a refactor would be required to add a new PartitionsMetadata model that models only the new partitions.json?

I think that would make the code in RegionGenerationMojo more clear as well (since its not clear to me which of the generation functions require which version of partitions).

codegen-lite-maven-plugin/src/main/java/software/amazon/awssdk/codegen/lite/maven/plugin/RegionGenerationMojo.java
```diff
@@ -59,19 +59,24 @@ public class RegionGenerationMojo extends AbstractMojo {
...
+        "${basedir}/src/main/resources/software/amazon/awssdk/regions/internal/region/partitions.json.resource")
+    private File partitionsjson;
+
     @Override
     public void execute() throws MojoExecutionException {
         Path baseSourcesDirectory = Paths.get(outputDirectory).resolve("generated-sources").resolve("sdk");
         Path testsDirectory = Paths.get(outputDirectory).resolve("generated-test-sources").resolve("sdk-tests");
 
         Partitions partitions = RegionMetadataLoader.build(endpoints);
+        Partitions regionpartitions = RegionMetadataLoader.build(partitionsjson);
```
alextwoods, line 72: Ditto here - is there a reason not to name this `regionPartitions`?

Alternative naming thought - rename `partitions` above to `legacyPartitions` and the new one to just `partitions`?

codegen-lite-maven-plugin/src/main/java/software/amazon/awssdk/codegen/lite/maven/plugin/RegionGenerationMojo.java
S-Saranya1: Sorry, my bad. I'll update it.

codegen-lite-maven-plugin/src/main/java/software/amazon/awssdk/codegen/lite/maven/plugin/RegionGenerationMojo.java
S-Saranya1: Sure, that's a good idea. I'll rename the partitions above to legacyPartitions and the new one to just partitions.

codegen-lite/src/main/java/software/amazon/awssdk/codegen/lite/regions/model/Partition.java
alextwoods: Yeah that makes sense - I'm not suggesting that we completly replace the existing Partition model since its still required.  I'm just suggesting that we create a new `PartitionsMetadata` model and use it in the places where we can.  That way we don't have `Partition` objects that have missing/incomplete data (eg, in the current code, `regionpartitions` will have empty service maps, since partitions.json doesn't have that information).

S-Saranya1: Yeah, got it. I'll make that change.

codegen-lite-maven-plugin/src/main/java/software/amazon/awssdk/codegen/lite/maven/plugin/RegionGenerationMojo.java
```diff
@@ -59,19 +59,24 @@ public class RegionGenerationMojo extends AbstractMojo {
         "${basedir}/src/main/resources/software/amazon/awssdk/regions/internal/region/endpoints.json")
     private File endpoints;
 
+    @Parameter(property = "partitionsjson", defaultValue =
+        "${basedir}/src/main/resources/software/amazon/awssdk/regions/internal/region/partitions.json.resource")
```
zoewangg, line 63: Can we use the same file in codegen?  Otherwise, we'd need to update our release system to keep it updated and if it's not possible, I'd prefer to keep the file in codegen-lite instead of core since we don't want anyone to use it or load it at runtime.

zoewangg (changes requested):

codegen-lite-maven-plugin/src/main/java/software/amazon/awssdk/codegen/lite/maven/plugin/RegionGenerationMojo.java
S-Saranya1: Updated it.

codegen-lite/src/main/java/software/amazon/awssdk/codegen/lite/regions/model/Partition.java
S-Saranya1: Made the changes as we discussed.

codegen-lite-maven-plugin/src/main/java/software/amazon/awssdk/codegen/lite/maven/plugin/RegionGenerationMojo.java
S-Saranya1: Yes, updated to use the file in codegen.

codegen-lite-maven-plugin/src/main/java/software/amazon/awssdk/codegen/lite/maven/plugin/RegionGenerationMojo.java
S-Saranya1: I named it as regionPartitions and didn't rename the partitions to legacyPartitions since there are other functions like generateServiceProvider and generateEndpointTags which are mainly using enpoints.json, so didn't want to add the legacy tag, kept it as partitions only and named the new one to regionPartitions.

alextwoods (approved):

codegen-lite/src/main/java/software/amazon/awssdk/codegen/lite/regions/model/PartitionMetadata.java
```diff
@@ -0,0 +1,177 @@
...
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.codegen.lite.regions.model;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import software.amazon.awssdk.annotations.SdkInternalApi;
+import java.util.Map;
```
alextwoods: Checkstyle nit - I think this is import is out of order and should be above sdkInternalApi.

zoewangg (approved):

alextwoods (approved):

PR #6130, opened 2025-05-23
Add Environment Token support

alextwoods:
## Motivation and Context
Adds support for configuring bearer auth using a token sourced from the environment.

## Modifications
* Adds customization config to enable environment bearer token support.
* When customization flag is set, the BaseClientBuilder will check if the ENV variable is set, and if so, will configure both a prefered auth scheme (using the mechanism from #6083) and add a static token identity provider.
* Includes additional token identity improvements (improve client configuration).

## Testing
* New and existing unit tests
* Manual testing against service endpoints.

## Types of changes

## Checklist

## License


codegen/src/main/java/software/amazon/awssdk/codegen/poet/builder/BaseClientBuilderClass.java
```diff
@@ -302,11 +307,52 @@ private MethodSpec mergeServiceDefaultsMethod() {
...
 
-        builder.addCode(");");
+        if (model.getCustomizationConfig().isEnableEnvironmentBearerToken()) {
+            configureEnvironmentBearerToken(builder);
+        }
+        builder.endControlFlow(")");
         return builder.build();
     }
 
+    private void configureEnvironmentBearerToken(MethodSpec.Builder builder) {
```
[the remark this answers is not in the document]
alextwoods: Added, see: `test-env-bearer-token-client-builder-class.java` (in particular [here](https://github.com/aws/aws-sdk-java-v2/blob/e9afd67a78144c1cfbe837b2920284a52e0a8237/codegen/src/test/resources/software/amazon/awssdk/codegen/poet/builder/sra/test-env-bearer-token-client-builder-class.java#L74-L88))

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/client/test-environment-token-metrics-interceptor-class.java
```diff
@@ -0,0 +1,38 @@
...
+            .getAttribute(SdkInternalExecutionAttribute.SELECTED_AUTH_SCHEME);
+        if (selectedAuthScheme != null && selectedAuthScheme.authSchemeOption().schemeId().equals(BearerAuthScheme.SCHEME_ID)
+            && selectedAuthScheme.identity().isDone()) {
+            if (selectedAuthScheme.identity().getNow(null) instanceof TokenIdentity) {
+                TokenIdentity configuredToken = (TokenIdentity) selectedAuthScheme.identity().getNow(null);
+                if (configuredToken.token().equals(tokenFromEnv)) {
+                    executionAttributes.getAttribute(SdkInternalExecutionAttribute.BUSINESS_METRICS).addMetric(
+                        BusinessMetricFeatureId.BEARER_SERVICE_ENV_VARS.value());
+                }
+            }
```
zoewangg, line 35: Is there a way to not use execution interceptor to capture metrics? We've had issues (executions out of order issue) in the past with doing things in execution interceptor and are moving away from that model for internal implementations

test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/environmenttokenprovider/EnvironmentTokenProviderTest.java
```diff
@@ -0,0 +1,187 @@
...
+    public static final String SYSTEM_TEST_TOKEN = "system-test-token";
+
+    private MockSyncHttpClient mockHttpClient;
+    private String systemPropertyBeforeTest;
+
+    private final EnvironmentVariableHelper environmentVariableHelper = new EnvironmentVariableHelper();
+
+    @BeforeEach
+    void setUp() {
+        mockHttpClient = new MockSyncHttpClient();
```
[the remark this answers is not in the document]
alextwoods: Good call - will do.

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/client/test-environment-token-metrics-interceptor-class.java
alextwoods: Summarizing offline discussion:

The ordering problem is a good callout. The new `<service>EnvironmentTokenMetricsInterceptor` must run after the service specific, code generated auth scheme interceptor (`<Service>AuthSchemeInterceptor`).  The alternatives I see are:

[**non-Interceptor**] - Adding logic in a stage in the request pipeline - specifically to [ApplyUserAgentStage](https://github.com/aws/aws-sdk-java-v2/blob/master/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/pipeline/stages/ApplyUserAgentStage.java).

We need to evaluate, per request, whether the configured env token was used (since a request configuration can override it).  So, we would need a client configuration thats added to the execution attributes the value of the ENV sourced token.  The ApplyUserAgentStage would then check for the presence of this execution context value and if set, would then check the resolved auth is bearer and the resolved identity matches, and if so, set the metric.
Thats logic thats specific to this customization that would then be in core and that check would be done for every service since the pipeline stage is in core and not code generated.  I'm not sure how we feel about that tradeoff.  I'd generally lean towards trying to keep some of this service specific customization out of the core pipeline.  

[**Addressing with changes to generated interceptors**]
We could address this in a few ways:
1. Move the logic in `<service>EnvironmentTokenMetricsInterceptor` from `beforeExecution` to `beforeMarshalling`.  This ensures its in a different step and eliminates ordering concerns.
2. Modifying code generation of the `<Service>AuthSchemeInterceptor` to include this logic when the customization is enabled.  This eliminates any issue with ordering, but isn't an ideal separation of concerns.

I've updated this PR with #1 from above and moved the logic from `beforeExecution` to `beforeMarshalling` for now.  I can prototype out the changes required to update the ApplyUserAgentStage though if you think the tradeoffs of having that customization logic in core make sense.

alextwoods: Update - I've now updated the PR with alternative 2 mentioned above - I've moved the metrics logic into the generated auth scheme interceptor which will ensure we will never have ordering issues.

davidh44 (approved):

zoewangg (approved):

PR #6132, opened 2025-05-26
Fix HTTP authentication retry failures by improving RepeatableInputStreamRequestEntity repeatability

joviegas:
### Problem
- HTTP authentication retry scenarios failing with: `Cannot retry non-repeatable request`
- Occurs when initial request receives 401 Unauthorized and retry attempt fails
- Authentication challenges not handled properly
- Test cases for HTTP 301/302 responses failing due to Apache 5.x auto-redirect behavior

### Root Cause
- RepeatableInputStreamRequestEntity implementation not properly integrated with Apache 5.x
- Inconsistent repeatability detection logic
- Apache 5.x follows redirects by default, while Apache 4.x returns redirect responses to client
  - Tests expect to receive 301/302 responses for validation
  - Apache 5.x auto-follows redirects, preventing test validation

### Modifications
1. Improved Stream Repeatability
   - Changed inheritance to implement HttpEntity directly
   - Enhanced repeatability detection logic
   - Proper mark/reset support

2. Enhanced Content Handling
   - Better ContentType parsing
   - Improved chunked transfer encoding support
   - Proper resource cleanup

3. Disabled Default Redirect Handling
   - Added `builder.disableRedirectHandling()` in client creation
   - Maintains backward compatibility with Apache 4.x behavior
   - Allows tests to properly validate redirect responses:
     ```java
     @Test
     public void supportsResponseCode301() throws Exception {
         testForResponseCode(HttpURLConnection.HTTP_MOVED_PERM);
     }

     @Test
     public void supportsResponseCode302() throws Exception {
         testForResponseCode(HttpURLConnection.HTTP_MOVED_TEMP);
     }
     ```

## Testing
- Added a test  in Apache 4.x to test RepeatableInputStreamRequestEntity as it is and then ported the same test to Apache 5.x to make sure its functionality is same.
- Redirect response handling (301/302)

## Impact
- Reliable authentication retry handling
- Consistent redirect behavior with Apache 4.x
- Improved resource management

## License

### DIFF
--- a/http-clients/apache-client/src/test/java/software/amazon/awssdk/http/apache/internal/RepeatableInputStreamRequestEntityTest.java
+++ b/http-clients/apache-client/src/test/java/software/amazon/awssdk/http/apache/internal/RepeatableInputStreamRequestEntityTest.java
@@ -26,12 +26,7 @@ import static org.junit.jupiter.api.Assertions.assertSame;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.mockito.Mockito.any;
-import static org.mockito.Mockito.doThrow;
 import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.never;
-import static org.mockito.Mockito.spy;
-import static org.mockito.Mockito.times;
-import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
 import java.io.ByteArrayInputStream;
@@ -181,6 +176,7 @@ class RepeatableInputStreamRequestEntityTest {
             public int read() {
                 return -1;
             }
+
             @Override
             public boolean markSupported() {
                 return false;
@@ -208,7 +204,7 @@ class RepeatableInputStreamRequestEntityTest {
         AtomicInteger resetCallCount = new AtomicInteger(0);
         ByteArrayInputStream trackingStream = new ByteArrayInputStream(content.getBytes()) {
             @Override
-            public synchronized void reset()  {
+            public synchronized void reset() {
                 resetCallCount.incrementAndGet();
                 super.reset();
             }
@@ -239,7 +235,7 @@ class RepeatableInputStreamRequestEntityTest {
         AtomicInteger resetCallCount = new AtomicInteger(0);
         ByteArrayInputStream trackingStream = new ByteArrayInputStream(content.getBytes()) {
             @Override
-            public synchronized void reset()  {
+            public synchronized void reset() {
                 resetCallCount.incrementAndGet();
                 super.reset();
             }
@@ -452,8 +448,8 @@ class RepeatableInputStreamRequestEntityTest {
     }
 
     @Test
-    @DisplayName("Entity should handle mixed headers correctly")
-    void constructor_WithMixedHeaders_HandlesAllCorrectly() {
+    @DisplayName("Entity should handle multiple headers correctly")
+    void constructor_WithMultiHeaders_HandlesAllCorrectly() {
         // Given
         SdkHttpRequest httpRequest = httpRequestBuilder
             .putHeader("Content-Length", "2048")
@@ -548,8 +544,8 @@ class RepeatableInputStreamRequestEntityTest {
     }
 
     @Test
-    @DisplayName("Entity should handle partial stream reads")
-    void writeTo_PartialReads_CompletesSuccessfully() throws IOException {
+    @DisplayName("Entity should handle non repeatable data arriving in chunks")
+    void writeTo_withChunkedReads_CompletesSuccessfully() throws IOException {
         // Given - Stream that returns data in small chunks
         String content = "This is a test content that will be read in chunks";
         InputStream chunkingStream = new InputStream() {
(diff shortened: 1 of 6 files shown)


http-clients/apache5-client/src/main/java/software/amazon/awssdk/http/apache5/Apache5HttpClient.java
```diff
@@ -148,6 +148,8 @@ private ConnectionManagerAwareHttpClient createClient(Apache5HttpClient.DefaultB
         ApacheConnectionManagerFactory cmFactory = new ApacheConnectionManagerFactory();
 
         HttpClientBuilder builder = HttpClients.custom();
+        //This is done to keep backward compatibility with Apache 4.x
```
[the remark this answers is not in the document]
joviegas: Good catch. No reason , just wanted to keep it at the top to make it noticeable , removed it now.

http-clients/apache-client/src/test/java/software/amazon/awssdk/http/apache/internal/RepeatableInputStreamRequestEntityTest.java
```diff
@@ -0,0 +1,799 @@
...
+
+        entity = new RepeatableInputStreamRequestEntity(request);
+
+        IOException thrown = assertThrows(IOException.class,
+                                          () -> entity.writeTo(new ByteArrayOutputStream()));
+        assertSame(originalException, thrown);
+    }
+
+    @Test
+    @DisplayName("writeTo should throw original exception on subsequent failures")
```
RanVaknin, line 296: why do we want to preserve the original exception?

http-clients/apache5-client/src/test/java/software/amazon/awssdk/http/apache5/Apache5HttpClientWireMockTest.java
```diff
@@ -115,44 +115,43 @@ public void routePlannerIsInvoked() throws Exception {
...
-    //
-    //                                                     ))
-    //                                             .buildWithDefaults(AttributeMap.builder()
-    //                                                                            .put(TRUST_ALL_CERTIFICATES, Boolean.TRUE)
-    //                                                                            .build());
-    //     testForResponseCodeUsingHttps(client, HttpURLConnection.HTTP_OK);
-    //
-    //     mockProxyServer.verify(2, RequestPatternBuilder.allRequests());
-    // }
+    @Test
```
RanVaknin, line 118: Should we have more types of functional tests? end to end tests? 
Again, I know these tests didnt exist for Apache 4, but we don't know what we don't know.

http-clients/apache-client/src/test/java/software/amazon/awssdk/http/apache/internal/RepeatableInputStreamRequestEntityTest.java
```diff
@@ -0,0 +1,799 @@
...
+        entity.writeTo(output3);
+
+        // All outputs should contain the same content
+        assertEquals(content, output1.toString());
+        assertEquals(content, output2.toString());
+        assertEquals(content, output3.toString());
+    }
+
+    @Test
+    @DisplayName("Entity should handle mixed headers correctly")
```
RanVaknin, line 455: nit: should we name this "Entity should handle **_multiple_** headers correctly" ? since the headers are unique?

http-clients/apache5-client/src/test/java/software/amazon/awssdk/http/apache5/Apache5HttpClientWireMockTest.java
joviegas: Our current migration focus is to ensure the Apache 5.x SDK client maintains the same behavior as Apache 4.x. For new modification like RepeatableInputStreamRequestEntity, we have added detailed unit tests.
End to end in test will be done with another task which is in plan.
 Is there any specific functionality you'd like us to focus on at this stage?

http-clients/apache-client/src/test/java/software/amazon/awssdk/http/apache/internal/RepeatableInputStreamRequestEntityTest.java
```diff
@@ -0,0 +1,799 @@
...
+        ByteArrayOutputStream output = new ByteArrayOutputStream();
+        entity.writeTo(output);
+        assertEquals(0, output.size());
+    }
+
+    @Test
+    @DisplayName("Entity should handle large content streams")
+    void writeTo_LargeContent_HandlesCorrectly() throws IOException {
+        // Given - 10MB of data
+        int size = 10 * 1024 * 1024;
```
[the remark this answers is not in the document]
joviegas: Large compared to InputStream's [MAX_SKIP_BUFFER_SIZE](https://github.com/openjdk/jdk/blob/b7f0f480cefb7295bbd5f8830458b69baf2eaff6/src/java.base/share/classes/java/io/InputStream.java#L54-L58) and other internal buffers

http-clients/apache-client/src/test/java/software/amazon/awssdk/http/apache/internal/RepeatableInputStreamRequestEntityTest.java
joviegas: For better user experience , since this is mark and reset in cases of retry the stream will be restarted and if it fails the user need to know the initial error that caused the retry.

PR #6150, opened 2025-06-02
Fix NPE in ProfileFileSupplier.defaultSupplier

### LINKED ISSUE #5635: ProfileFileSupplier.defaultSupplier() leads to NullPointer when config files are missing
### Describe the bug

We recently introduced automatic credentials reload in the [EventBridge Kafka connector](https://github.com/awslabs/eventbridge-kafka-connector) by leveraging _ProfileFileSupplier.defaultSupplier_  as recommended in the [official docs](https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/credentials-profiles.html#profile-reloading):

```java
var provider = DefaultCredentialsProvider.builder()
         //.profileFile(ProfileFile.defaultProfileFile()) <- this is the default by the SDK 
          .profileFile(ProfileFileSupplier.defaultSupplier()) //<- This supports automatic reload
          .build();

var creds = provider.resolveCredentials();
```

However, when no credential files are present (_~/.aws/credentials_ and _~/.aws/config_ are missing) the DefaultCredentialProviderChain.resolveCredentials() process [will result in NullPointer Exceptions](https://github.com/aws/aws-sdk-java-v2/blob/832c1b93a2c2712e65d03f687d11fde8a8a6faa2/core/profiles/src/main/java/software/amazon/awssdk/profiles/ProfileFile.java#L314) in both ProfileCredentialsProvider and InstanceProfileCredentialProvider: `java.lang.NullPointerException: Cannot invoke "java.nio.file.Path.getFileSystem()" because "path" is null`. (See Additional Information below).

This will happen first in the `ProfileCredentialsProvider` and will be ignored to move to the next provider as outlined [here](https://github.com/aws/aws-sdk-java-v2/blob/3c6c397238870c423bd5e9cb30c1bbce1a3615a7/core/auth/src/main/java/software/amazon/awssdk/auth/credentials/AwsCredentialsProviderChain.java#L116). But the process will run into it again for the _InstanceProfileCredentialProvider_, for example in [Ec2MetaDataConfigProvider](https://github.com/aws/aws-sdk-java-v2/blob/3c6c397238870c423bd5e9cb30c1bbce1a3615a7/core/regions/src/main/java/software/amazon/awssdk/regions/internal/util/Ec2MetadataConfigProvider.java#L129). 

This behavior essentially influences (and breaks) the DefaultCredentialsProvider chain. It behaves differently than the current SDK default value `ProfileFile.defaultProfileFile()` which does not produce any exceptions and lets _InstanceProfileCredentialProvider_ resolve even though files are missing.

**Related issues:**

https://github.com/awslabs/eventbridge-kafka-connector/issues/396

### Expected Behavior

_ProfileFileSupplier.defaultSupplier()_ behaves the same as  _ProfileFile.defaultProfileFile()_ when credential files are missing.

### Current Behavior

_ProfileFileSupplier.defaultSupplier()_ leads to Nullpointer during resolution process and breaks _DefaultCredentialsProvider_ (or at least behaves different than the default)

### Reproduction Steps

1. First, delete local files such as _~/.aws/credentials_ and _~/.aws/config_
2. Run the below code with the default value of ProfileFile.defaultProfileFile()
3. Run the below code with the non-default value: profileFile(ProfileFileSupplier.defaultSupplier()

```java
public class Main {
        var provider = DefaultCredentialsProvider.builder()
                    //    .profileFile(ProfileFile.defaultProfileFile()) <- This is the default and is used implicit
                        .profileFile(ProfileFileSupplier.defaultSupplier()) // <- This produces NullPointer when no credentials files are present (Check Errors in the ProviderChain) 
                        .build();

        var creds = provider.resolveCredentials();
        System.out.println(creds);
    }
}
```

4. Check for Nullpointer in the resolution process and see different path for InstanceProfileCredentialsProvider 

### Possible Solution

When no credential files are present (_~/.aws/credentials_ and _~/.aws/config._ are missing) 
 [ProfileFileSupplier.defaultSupplier() effectively returns ](https://github.com/aws/aws-sdk-java-v2/blob/832c1b93a2c2712e65d03f687d11fde8a8a6faa2/core/profiles/src/main/java/software/amazon/awssdk/profiles/ProfileFileSupplier.java#L58)`ProfileFile.builder().build()`since _credentialsSupplierOptional_ and _configurationSupplierOptional_ are empty. Later during resolveCredentials of the DefaultCredentialsProviderChain the _contentLocation_ within ProfileFile Builder will be null and therefore [lead to the Nullpointer](https://github.com/aws/aws-sdk-java-v2/blob/832c1b93a2c2712e65d03f687d11fde8a8a6faa2/core/profiles/src/main/java/software/amazon/awssdk/profiles/ProfileFile.java#L314):

```java
@Override
public ProfileFile build() {
    InputStream stream = content != null ? content :
                         FunctionalUtils.invokeSafely(() -> Files.newInputStream(contentLocation));

```

It would make sense to provide a sane default here such as ProfileFile.defaultProfileFile().

This will happen first in the `ProfileCredentialsProvider` and will be ignored to move to the next provider as outlined [here](https://github.com/aws/aws-sdk-java-v2/blob/3c6c397238870c423bd5e9cb30c1bbce1a3615a7/core/auth/src/main/java/software/amazon/awssdk/auth/credentials/AwsCredentialsProviderChain.java#L116). But InstanceProfileCredentialProvider will run into it again, for example in [Ec2MetaDataConfigProvider](https://github.com/aws/aws-sdk-java-v2/blob/3c6c397238870c423bd5e9cb30c1bbce1a3615a7/core/regions/src/main/java/software/amazon/awssdk/regions/internal/util/Ec2MetadataConfigProvider.java#L129).

### AWS Java SDK version used

2.28.8

### JDK version used

17

MacOS

alextwoods:
Fix NPE in `ProfileFileSupplier.defaultSupplier` when both credentials and config files do not exist.

Fixes #5635 

## Motivation and Context
When neither credentials or config files exist, calling `ProfileFileSupplier.defaultSupplier()` will return a supplier that raises an NPE when the value is accessed through `get` because the `ProfileFile.Builder` requires that both content and type are set. Note that `ProfileFile.defaultProfileFile()` will effectively return an empty ProfileFile in this case.

## Modifications
* Adds a package private `ProfileFile.empty()` method to create empty profile files.
* When When neither credentials or config files exist, `defaultSupplier` will return a static supplier with an empty profileFile - matching the behavior of `ProfileFile.defaultProfileFile()`.
* Improve the error message in `ProfileFile.Builder.build()` when content is missing.

## Testing
Add new test case, fails without the change and now passes.

## Types of changes

## Checklist

## License

### DIFF
--- a/core/profiles/src/main/java/software/amazon/awssdk/profiles/ProfileFile.java
+++ b/core/profiles/src/main/java/software/amazon/awssdk/profiles/ProfileFile.java
@@ -317,12 +317,10 @@ public final class ProfileFile {
 
         @Override
         public ProfileFile build() {
-            InputStream stream = null;
-            if (content != null) {
-                stream = content;
-            } else if (contentLocation != null) {
-                stream = FunctionalUtils.invokeSafely(() -> Files.newInputStream(contentLocation));
-            }
+            Validate.isTrue(content != null || contentLocation != null,
+                            "content or contentLocation must be set.");
+            InputStream stream = content != null ? content :
+                                  FunctionalUtils.invokeSafely(() -> Files.newInputStream(contentLocation));
 
             Validate.paramNotNull(type, "type");
             Validate.paramNotNull(stream, "content");
--- a/core/profiles/src/main/java/software/amazon/awssdk/profiles/ProfileFile.java
+++ b/core/profiles/src/main/java/software/amazon/awssdk/profiles/ProfileFile.java
@@ -86,6 +86,13 @@ public final class ProfileFile {
     }
 
     /**
+     * Create an empty profile file.
+     */
+    static ProfileFile empty() {
+        return new ProfileFile(Collections.emptyMap());
+    }
+
+    /**
      * Get the default profile file, using the credentials file from "~/.aws/credentials", the config file from "~/.aws/config"
      * and the "default" profile. This default behavior can be customized using the
      * {@link ProfileFileSystemSetting#AWS_SHARED_CREDENTIALS_FILE}, {@link ProfileFileSystemSetting#AWS_CONFIG_FILE} and
@@ -310,8 +317,12 @@ public final class ProfileFile {
 
         @Override
         public ProfileFile build() {
-            InputStream stream = content != null ? content :
-                                 FunctionalUtils.invokeSafely(() -> Files.newInputStream(contentLocation));
+            InputStream stream = null;
+            if (content != null) {
+                stream = content;
+            } else if (contentLocation != null) {
+                stream = FunctionalUtils.invokeSafely(() -> Files.newInputStream(contentLocation));
+            }
 
             Validate.paramNotNull(type, "type");
             Validate.paramNotNull(stream, "content");
--- a/.changes/next-release/bugfix-AWSSDKforJavav2-30203bb.json
+++ b/.changes/next-release/bugfix-AWSSDKforJavav2-30203bb.json
@@ -1,0 +1,6 @@
+{
+    "type": "bugfix",
+    "category": "AWS SDK for Java v2",
+    "contributor": "",
+    "description": "Fix NPE in `ProfileFileSupplier.defaultSupplier` when both credentials and config files do not exist."
+}
--- a/core/profiles/src/main/java/software/amazon/awssdk/profiles/ProfileFileSupplier.java
+++ b/core/profiles/src/main/java/software/amazon/awssdk/profiles/ProfileFileSupplier.java
@@ -55,13 +55,15 @@ public interface ProfileFileSupplier extends Supplier<ProfileFile> {
             = ProfileFileLocation.configurationFileLocation()
                                  .map(path -> reloadWhenModified(path, ProfileFile.Type.CONFIGURATION));
 
-        ProfileFileSupplier supplier = () -> ProfileFile.builder().build();
+        ProfileFileSupplier supplier;
         if (credentialsSupplierOptional.isPresent() && configurationSupplierOptional.isPresent()) {
             supplier = aggregate(credentialsSupplierOptional.get(), configurationSupplierOptional.get());
         } else if (credentialsSupplierOptional.isPresent()) {
             supplier = credentialsSupplierOptional.get();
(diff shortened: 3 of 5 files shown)


alextwoods:
Note on `no-api-surface-area-change` label - both ProfileFile and ProfileFileSupplier have changes, but neither changes the public api (the new `empty` method in ProfileFile is package private).

core/profiles/src/main/java/software/amazon/awssdk/profiles/ProfileFile.java
```diff
@@ -85,6 +85,13 @@ public static Aggregator aggregator() {
         return new Aggregator();
     }
 
+    /**
+     * Create an empty profile file.
+     */
+    static ProfileFile empty() {
```
joviegas, line 91: What is the behavior when we have empty file (In this issue case the file doesnot exist) ? Even without this changes does it return empty map for a Empty profile file ?

alextwoods: Without this change (adding the `empty` method), we don't really have a way to construct an empty profile file other than using an empty aggregate - we could do that instead of using this method, but it requires checking the presence of the config/credentials files twice instead of just once.

> What is the behavior when we have empty file (In this issue case the file doesnot exist)?

See #5635 for more details on the behavior.  Right now, calling [ProfileFile.defaultProfileFile()](https://github.com/aws/aws-sdk-java-v2/blob/master/core/profiles/src/main/java/software/amazon/awssdk/profiles/ProfileFile.java#L97) will return an empty profile file (or more specifically, it returns an Aggregator with no files, which will effectively create a ProfileFile with an empty map).   However, calling [ProfileFileSupplier.defaultSupplier()](https://github.com/aws/aws-sdk-java-v2/blob/master/core/profiles/src/main/java/software/amazon/awssdk/profiles/ProfileFileSupplier.java#L49) will raise an NPE since it tries to use the builder incorrectly.  This PR changes that to explicitly use an empty ProfileFile in that case.

joviegas (approved):

PR #6151, opened 2025-06-02
S3 transfer manager client constructor and method recipe

Fred1155:
## Motivation and Context

Added recipe to cover S3 transfer manager client constructor and simple client method
## Modifications

Added recipe for 
TransferManager(AWSCredentials credentials)
TransferManager(AWSCredentialsProvider credentialsProvider)
downloadDirectory(String bucketName, String keyPrefix, File destinationDirectory)
resumeDownload(PersistableDownload persistableDownload)
resumeUpload(PersistableUpload persistableUpload)
shutdownNow()
## Testing

Added integration test cases
## Types of changes

## Checklist

## License


v2-migration/src/main/java/software/amazon/awssdk/v2migration/TransferManagerMethodsToV2.java
```diff
@@ -95,10 +111,100 @@ public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, Execu
...
             return super.visitMethodInvocation(method, executionContext);
         }
 
+        @Override
+        public J visitNewClass(J.NewClass newClass, ExecutionContext executionContext) {
+            JavaType type = newClass.getType();
+            if (!(type instanceof JavaType.FullyQualified)) {
+                return newClass;
+            }
+            if (type.isAssignableFrom(S3_TM_CREDENTIAL) &&
```
L-Applin, line 124: I'm probably misunderstanding open rewrite, but isn't that check checking if the call to v1 constructor is of v2 type? Isn't there a type mismatch between v1 and v2 here?

v2-migration/src/main/java/software/amazon/awssdk/v2migration/TransferManagerMethodsToV2.java
```diff
@@ -95,10 +111,100 @@ public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, Execu
...
+                newClass.getArguments().size() == 1 &&
+                newClass.getArguments().get(0).getType() != null) {
+                if (newClass.getArguments().get(0).getType().isAssignableFrom(V2_AWSCREDENTAIL)) {
```
L-Applin: similar here, I thought `newClass` was a reference to the v1 TM new instace call, but we are checking for v2 credentials in its argument?

v2-migration/src/main/java/software/amazon/awssdk/v2migration/TransferManagerMethodsToV2.java
```diff
@@ -95,10 +111,100 @@ public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, Execu
...
+                    return JavaTemplate
+                        .builder("S3TransferManager.builder()" +
+                                 ".s3Client(S3AsyncClient.builder()" +
+                                 ".credentialsProvider(StaticCredentialsProvider.create(#{any()}))" +
+                                 ".build())" +
+                                 ".build()")
+                        .build()
+                        .apply(getCursor(), newClass.getCoordinates().replace(), newClass.getArguments().get(0));
+                }
+                if (newClass.getArguments().get(0).getType().isAssignableFrom(V2_CREDENTIAL_PROVIDER)) {
```
L-Applin: same here

L-Applin:
Are the migration unit test including the new code of the before / after test classes?

v2-migration/src/main/java/software/amazon/awssdk/v2migration/TransferManagerMethodsToV2.java
Fred1155: Yes your understanding is correct and we are checking v2 type. That is because the v1-to-v2 type transformation recipe is applied before this recipe, which means when this recipe is applied, all types have already been transformed to v2 type.

Fred1155:
> Are the migration unit test including the new code of the before / after test classes?

No I didn't add unit test for that. Adding the integration test cases should be sufficient I think.

v2-migration/src/main/java/software/amazon/awssdk/v2migration/TransferManagerMethodsToV2.java
```diff
@@ -47,6 +50,19 @@ public class TransferManagerMethodsToV2 extends Recipe {
...
 
+    private static final MethodMatcher DOWNLOAD_DIR = v2TmMethodMatcher("downloadDirectory(String, String, java.io.File)");
+
+    private static final MethodMatcher RESUME_DOWNLOAD = v2TmMethodMatcher("resumeDownload(..)");
+    private static final MethodMatcher RESUME_UPLOAD = v2TmMethodMatcher("resumeUpload(..)");
+    private static final MethodMatcher SHUT_DOWN_NOW = v2TmMethodMatcher("shutdownNow()");
+
+
+
+    private static final Pattern S3_TM_CREDENTIAL = Pattern.compile(V2_TM_CLIENT);
```
davidh44:
```suggestion
    private static final MethodMatcher SHUT_DOWN_NOW = v2TmMethodMatcher("shutdownNow()");

    private static final Pattern S3_TM_CREDENTIAL = Pattern.compile(V2_TM_CLIENT);
```

nit: remove extra whitespace lines

v2-migration/src/main/java/software/amazon/awssdk/v2migration/TransferManagerMethodsToV2.java
davidh44: Yup all v2 types here is expected

v2-migration/src/main/java/software/amazon/awssdk/v2migration/TransferManagerMethodsToV2.java
davidh44: Can we extract `newClass.getArguments().get(0)` into a variable here before the second `if()`?

v2-migration/src/main/java/software/amazon/awssdk/v2migration/TransferManagerMethodsToV2.java
```diff
@@ -95,10 +111,100 @@ public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, Execu
...
+        }
+
+        private J.MethodInvocation transformResumeUpload(J.MethodInvocation method) {
+            String v2Method = "#{any()}.resumeUploadFile(#{any()})";
+
+            method = JavaTemplate.builder(v2Method).build()
+                                 .apply(getCursor(), method.getCoordinates().replace(), method.getSelect(),
+                                        method.getArguments().get(0));
+            return method;
+        }
```
davidh44: For these name change only transforms, should we just use ChangeMethodName yaml recipe like https://github.com/aws/aws-sdk-java-v2/blob/master/v2-migration/src/main/resources/META-INF/rewrite/change-s3-methods.yml ?

davidh44:
> > Are the migration unit test including the new code of the before / after test classes?
> 
> No I didn't add unit test for that. Adding the integration test cases should be sufficient I think.

Yeah we've been doing just integ tests, easier to write and better coverage than individual unit tests

davidh44 (approved):

Issue #4295, opened 2023-08-11
DynamoDbAutoGeneratedTimestampAttribute format in DynamoDb is incompatible with Time To Live


acouvreur:
### Describe the bug

DynamoDbAutoGeneratedTimestampAttribute format in DynamoDb is incompatible with Time To Live.

When using attributes such as `createdAt` or `upatedAt` like the following:

```java
@DynamoDbAutoGeneratedTimestampAttribute
@DynamoDbUpdateBehavior(UpdateBehavior.WRITE_IF_NOT_EXISTS)
public Instant getCreatedAt() {
    return createdAt;
}

@DynamoDbAutoGeneratedTimestampAttribute
public Instant getUpdatedAt() {
    return updatedAt;
}
```

The value in DynamoDb is stored with the following format: `2023-06-06T18:16:40.632760903Z`.

---

This format is not compatible with a Time To Live attribute as documented here: https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/time-to-live-ttl-before-you-start.html

> The TTL attribute’s value must be a timestamp in [Unix epoch time format](https://en.wikipedia.org/wiki/Unix_time) in seconds. If you use any other format, the TTL processes ignore the item. For example, if you set the value of the attribute to 1645119622, that is Thursday, February 17, 2022 17:40:22 (GMT), the item will be expired after that time. For help formatting your epoch timestamps, you can use third-party tools such as [Epoch Converter](https://epochconverter.com/) to get a visual web form.

### Expected Behavior

I would expect that properties that specifically use the `DynamoDbAutoGeneratedTimestampAttribute` annotation would be compatible with TTL format.

### Current Behavior

And as per the general usage and documented in the project (see https://github.com/search?q=repo%3Aaws%2Faws-sdk-java-v2%20DynamoDbAutoGeneratedTimestampAttribute&type=code) we are using an `Instant`.

If this is not compatible with the TTL format on purpose, how can I store these timestamp as TTL compatible attributes?

### Reproduction Steps

```java
@DynamoDbAutoGeneratedTimestampAttribute
@DynamoDbUpdateBehavior(UpdateBehavior.WRITE_IF_NOT_EXISTS)
public Instant getCreatedAt() {
    return createdAt;
}

@DynamoDbAutoGeneratedTimestampAttribute
public Instant getUpdatedAt() {
    return updatedAt;
}
```

And save something to DynamoDb, check the format is not unix timestamp.

### Additional Information/Context

[...]

PR #6152, opened 2025-06-03
Support timeToLive attributes in DynamoDb Enhanced Client

roamariei:
## Description
Support TimeToLive attributes in DynamoDb Enhanced Client

## Motivation and Context
https://github.com/aws/aws-sdk-java-v2/issues/4295

## Modifications
Added DynamoDbTimeToLiveAttribute annotation, TimeToLive extension to run before every write, added new operations DescribeTimeToLive and UpdateTimeToLive in enhanced client.

## Testing
Added unit tests for TimeToLiveExtension, DescribeTimeToLiveOperation, UpdateTimeToLiveOperation and integration tests for update timeToLive with enhanced client in TimeToLiveRecordTest.

## Test Coverage Checklist

| Scenario | Done | Comments if Not Done |
|---------|:----:|---------------------|
| **1. Different TableSchema Creation Methods** | | Not impacted |
| a. TableSchema.fromBean(Customer.class) | [ ] | |
| b. TableSchema.fromImmutableClass(Customer.class) for immutable classes | [ ] | |
| c. TableSchema.documentSchemaBuilder().build() | [ ] | |
| d. StaticTableSchema.builder(Customer.class) | [ ] | |
| **2. Nesting of Different TableSchema Types** | | Not impacted |
| a. @DynamoDbBean with nested @DynamoDbBean as NonNull | [ ] | |
| b. @DynamoDbBean with nested @DynamoDbImmutable as NonNull | [ ] | |
| c. @DynamoDbImmutable with nested @DynamoDbBean as NonNull | [ ] | |
| d. @DynamoDbBean with nested @DynamoDbBean as Null | [ ] | |
| e. @DynamoDbBean with nested @DynamoDbImmutable as Null| [ ] | |
| f. @DynamoDbImmutable with nested @DynamoDbBean as Null | [ ] | |
| **3. CRUD Operations** | | |
| a. scan() | [ ] | |
| b. query() | [ ] | |
| c. updateItem() | [ ] | |
| d. putItem() | [x] | |
| e. getItem() | [x] | |
| f. deleteItem() | [ ] | |
| g. batchGetItem() | [ ] | |
| h. batchWriteItem() | [ ] | |
| i. transactGetItems() | [ ] | |
| j. transactWriteItems()  | [ ] | |
| **4. Data Types and Null Handling** | |  Not impacted |
| a. top-level null attributes | [ ] | |
| b. collections with null elements | [ ] | |
| c. maps with null values | [ ] | |
| d. conversion between null Java values and AttributeValue | [ ] | |
| e. full serialization/deserialization cycle with null values | [ ] | |
| **5. AsyncTable and SyncTable** | | |
| a. DynamoDbAsyncTable Testing | [ ] | |
| b. DynamoDbTable Testing | [x] | |
| **6. New/Modification in Extensions** | | |
| a. Tables with Scenario in ScenarioSl No.1 (All table schemas are Must) | [x] | |
| b. Test with Default Values in Annotations | [x] | |
| c. Combination of Annotation and Builder passes extension | [x] | |
| **7. New/Modification in Converters** | | Not impacted |
| a. Tables with Scenario in ScenarioSl No.1 (All table schemas are Must) | [ ] | |
| b. Test with Default Values in Annotations | [ ] | |
| c. Test All Scenarios from 1 to 5 | [ ] | |

## Types of changes

## Checklist

## License


andreas-grafenberger:
Cannot edit the PR description, so I'll attach here the test coverage evidence:
image

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/extensions/TimeToLiveExtension.java
```diff
@@ -0,0 +1,186 @@
...
+        public static StaticAttributeTag timeToLiveAttribute(String baseField, long duration, ChronoUnit unit) {
+            return new TimeToLiveAttribute(baseField, duration, unit);
+        }
+    }
+
+    private static final class TimeToLiveAttribute implements StaticAttributeTag {
+
+        public String baseField;
+        public long duration;
+        public ChronoUnit unit;
```
amzn-erdemkemer: Should these be public  or private ?

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/extensions/TimeToLiveExtension.java
```diff
@@ -0,0 +1,186 @@
...
+import software.amazon.awssdk.enhanced.dynamodb.mapper.StaticTableMetadata;
+import software.amazon.awssdk.services.dynamodb.model.AttributeValue;
+import software.amazon.awssdk.utils.StringUtils;
+import software.amazon.awssdk.utils.Validate;
+
+@SdkPublicApi
+@ThreadSafe
+public final class TimeToLiveExtension implements DynamoDbEnhancedClientExtension {
+
+    public static final String CUSTOM_METADATA_KEY = "TimeToLiveExtension:TimeToLiveAttribute";
```
amzn-erdemkemer, line 47: 1- Consider renaming this to something like, TTL_ATTRIBUTE_METADATA_KEY, current name is too vague
2- This Key is under @SdkPublicApi class, because the class is @SdkPublicApi, everything public on it becomes part of the public API contract, but it is only used from UpdateTimeToLiveOperation which is an SdkInternalAPI for any backward compatibility issues going forward. Consider either making the field package-private (so UpdateTimeToLiveOperation in the same package can still access it), or moving the constant to an @SdkInternalApi shared utility class...

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/extensions/TimeToLiveExtension.java
```diff
@@ -0,0 +1,186 @@
...
+        public Consumer<StaticTableMetadata.Builder> modifyMetadata(String attributeName,
+                                                                    AttributeValueType attributeValueType) {
+            Map<String, Object> customMetadataMap = new HashMap<>();
+            customMetadataMap.put("attributeName", attributeName);
+            customMetadataMap.put("baseField", baseField);
+            customMetadataMap.put("duration", duration);
+            customMetadataMap.put("unit", unit);
+
+            return metadata -> metadata.addCustomMetadataObject(CUSTOM_METADATA_KEY,
+                                                                Collections.unmodifiableMap(customMetadataMap));
```
amzn-erdemkemer, line 198: BeanTableSchema validates that only one TTL attribute exists, but StaticTableSchema has no such guard. Since modifyMetadata stores a Map (not a Collection) via addCustomMetadataObject, a second TTL attribute silently replaces the first with no error:

```java
StaticTableSchema.builder(MyItem.class)
    .addAttribute(Long.class, a -> a.name("ttl1")
        .getter(MyItem::getTTL1).setter(MyItem::setTTL1)
        .tags(timeToLiveAttribute("createdAt", 30, ...)))
    .addAttribute(Long.class, a -> a.name("ttl2")
        .getter(MyItem::getTTL2).setter(MyItem::setTTL2)
        .tags(timeToLiveAttribute("updatedAt", 7, ...)))
    .build();
// ttl1 config is silently lost, only ttl2 is tracked
```

DynamoDB only supports one TTL attribute per table, so this is always a user error, but the SDK should fail loudly rather than silently dropping one.

Recommendation: You may consider moving the validation into TimeToLiveAttribute.modifyMetadata() so it works for all schema types

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/client/ExtensionResolver.java
```diff
@@ -33,9 +34,10 @@ public final class ExtensionResolver {
         VersionedRecordExtension.builder().build();
     private static final DynamoDbEnhancedClientExtension DEFAULT_ATOMIC_COUNTER_EXTENSION =
         AtomicCounterExtension.builder().build();
-
+    private static final DynamoDbEnhancedClientExtension DEFAULT_TIME_TO_LIVE_EXTENSION =
+        TimeToLiveExtension.builder().build();
     private static final List<DynamoDbEnhancedClientExtension> DEFAULT_EXTENSIONS =
-        Arrays.asList(DEFAULT_VERSIONED_RECORD_EXTENSION, DEFAULT_ATOMIC_COUNTER_EXTENSION);
+        Arrays.asList(DEFAULT_VERSIONED_RECORD_EXTENSION, DEFAULT_ATOMIC_COUNTER_EXTENSION, DEFAULT_TIME_TO_LIVE_EXTENSION);
```
amzn-erdemkemer, line 40: This means every existing DynamoDbEnhancedClient using default extensions will now run TimeToLiveExtension.beforeWrite() on every write operation, even for tables
with no TTL attributes. The extension does an early return when no metadata is found, but we still change the behavior:

1. Performance: An additional extension in the chain for every write across the entire SDK, for all users, even those who never use TTL. The customMetadataObject() lookup + Optional.orElse(null) runs on every putItem/updateItem/batchWriteItem/transactWriteItems.

2. Double registration: Users who extend the defaults will register TTL twice:

```java
   // This is how TimeToLiveRecordTest in this PR does it — TTL extension runs twice per write
   DynamoDbEnhancedClient.builder()
       .extensions(Stream.concat(
           ExtensionResolver.defaultExtensions().stream(),
           Stream.of(timeToLiveExtension))
           .collect(Collectors.toList()))
       .build();
```

Should we think about making this extension opt-in instead ?

also, if we want to keep this, The Javadoc comment on defaultExtensions() says "Currently this is just the VersionedRecordExtension", we may need to update it

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/DynamoDbTable.java
```diff
@@ -926,4 +928,12 @@ default void deleteTable() {
     default DescribeTableEnhancedResponse describeTable() {
         throw new UnsupportedOperationException();
     }
+
```
amzn-erdemkemer, line 931: consider adding javadoc, also applies to UpdateTimeToLiveEnhancedResponse and to Async table

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/extensions/TimeToLiveExtension.java
```diff
@@ -0,0 +1,186 @@
...
+        Map<String, ?> customTTLMetadata = context.tableMetadata()
+                                                  .customMetadataObject(CUSTOM_METADATA_KEY, Map.class).orElse(null);
+
+        if (customTTLMetadata != null) {
+            String ttlAttributeName = (String) customTTLMetadata.get("attributeName");
+            String baseFieldName = (String) customTTLMetadata.get("baseField");
+            Long duration = (Long) customTTLMetadata.get("duration");
+            TemporalUnit unit = (TemporalUnit) customTTLMetadata.get("unit");
+
+            Map<String, AttributeValue> itemToTransform = new HashMap<>(context.items());
```
amzn-erdemkemer: I think, we create a new Hashmap and copy items, this even there is no TTL, consider doing the check first and then init

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/extensions/TimeToLiveExtension.java
```diff
@@ -0,0 +1,186 @@
...
+        }
+        if (baseValue instanceof LocalDate) {
+            return ((LocalDate) baseValue).atStartOfDay(ZoneOffset.UTC).plus(duration, unit).toEpochSecond();
+        }
+        if (baseValue instanceof LocalDateTime) {
+            return ((LocalDateTime) baseValue).plus(duration, unit).toEpochSecond(ZoneOffset.UTC);
+        }
+        if (baseValue instanceof LocalTime) {
+            return LocalDate.now().atTime((LocalTime) baseValue).plus(duration, unit).toEpochSecond(ZoneOffset.UTC);
+        }
```
amzn-erdemkemer, line 116: Do we want to support local time ?

```java
if (baseValue instanceof LocalTime) {
    return LocalDate.now().atTime((LocalTime) baseValue).plus(duration, unit).toEpochSecond(ZoneOffset.UTC);
}
```

two potential problems:

1. Midnight jump: LocalDate.now() defines what "today" means, and that changes at midnight. Items with the same LocalTime base value and duration get TTLs that differ by a day depending on which side of midnight the write lands:

- Write at 11:59 PM March 10 -> LocalDate.now() = March 10-> TTL = March 11 00:05 (expires in ~6 minutes)
- Write at 12:01 AM March 11 -> LocalDate.now() = March 11-> TTL = March 12 00:05 (expires in ~24 hours)

2. JVM timezone dependency: LocalDate.now() uses ZoneId.systemDefault(), but the epoch conversion uses ZoneOffset.UTC. If the JVM runs in America/New_York, LocalDate.now() could return a different date than UTC, shifting the TTL by hours depending on where the server runs.

If you say, LocalDate.now(ZoneOffset.UTC), it would fix problem 2, but problem 1 is still there to LocalTime having no date component, there's always a midnight boundary where the TTL jumps by 24 hours.

Recommendation: You may consider removiong LocalTime support and let it fall through to the existing IllegalArgumentException if that's acceptable. The other five types (Instant, LocalDate, LocalDateTime,
ZonedDateTime, Long) are all deterministic and cover every real TTL use case. Users who need time-of-day-based TTL can compute the epoch seconds themselves with explicit control over the date and timezone.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/extensions/TimeToLiveExtension.java
```diff
@@ -0,0 +1,186 @@
...
+    @Override
+    public WriteModification beforeWrite(DynamoDbExtensionContext.BeforeWrite context) {
+        Map<String, ?> customTTLMetadata = context.tableMetadata()
+                                                  .customMetadataObject(CUSTOM_METADATA_KEY, Map.class).orElse(null);
+
+        if (customTTLMetadata != null) {
+            String ttlAttributeName = (String) customTTLMetadata.get("attributeName");
+            String baseFieldName = (String) customTTLMetadata.get("baseField");
+            Long duration = (Long) customTTLMetadata.get("duration");
+            TemporalUnit unit = (TemporalUnit) customTTLMetadata.get("unit");
```
amzn-erdemkemer: If any key is missing or has a wrong type (like someone manually constructs metadata via StaticTableMetadata.builder().addCustomMetadataObject(...)), these lines will throw NullPointerException with no context about which key failed or why. Can consider to use Validate.notNull() with descriptive messages..

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/operations/UpdateTimeToLiveOperation.java
```diff
@@ -0,0 +1,92 @@
...
+import software.amazon.awssdk.services.dynamodb.model.UpdateTimeToLiveRequest;
+import software.amazon.awssdk.services.dynamodb.model.UpdateTimeToLiveResponse;
+
+@SdkInternalApi
+public class UpdateTimeToLiveOperation<T> implements TableOperation<T, UpdateTimeToLiveRequest, UpdateTimeToLiveResponse,
+    UpdateTimeToLiveEnhancedResponse> {
+
+    private final boolean enabled;
+
+    public UpdateTimeToLiveOperation(boolean enabled) {
```
amzn-erdemkemer: Is this supposed to be private ?

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/model/UpdateTimeToLiveEnhancedResponse.java
```diff
@@ -0,0 +1,94 @@
...
+    private UpdateTimeToLiveEnhancedResponse(Builder builder) {
+        this.response = Validate.paramNotNull(builder.response, "response");
+    }
+
+    /**
+     * The properties of the timeToLive specification of the table.
+     *
+     * @return The properties of the timeToLive specification.
+     */
+    public TimeToLiveSpecification table() {
```
amzn-erdemkemer: table() function name is misleading, may consider renaming it to something like timeToLiveSpecification() ?

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/model/UpdateTimeToLiveEnhancedResponse.java
```diff
@@ -0,0 +1,94 @@
...
+import software.amazon.awssdk.annotations.ThreadSafe;
+import software.amazon.awssdk.enhanced.dynamodb.DynamoDbAsyncTable;
+import software.amazon.awssdk.enhanced.dynamodb.DynamoDbTable;
+import software.amazon.awssdk.services.dynamodb.model.TimeToLiveSpecification;
+import software.amazon.awssdk.services.dynamodb.model.UpdateTimeToLiveResponse;
+import software.amazon.awssdk.utils.ToString;
+import software.amazon.awssdk.utils.Validate;
+
+/**
+ * Defines the elements returned by DynamoDB from a {@code DescribeTimeToLive} operation, such as
```
amzn-erdemkemer: I guess not `DescribeTimeToLive` , probably mistake ?

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/extensions/TimeToLiveExtension.java
valeriodelbello-amazon: I strongly suggest we only use UTC and only support `Instant` with a `Clock` parameter like in `AutoGeneratedTimestampRecordExtension`.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/extensions/TimeToLiveExtension.java
valeriodelbello-amazon: I think these should be `private` and `final` to easily ensure the `@ThreadSafe` is respected.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/extensions/TimeToLiveExtension.java
valeriodelbello-amazon: I agree and a wrong type would throw a different exception and require a dedicated check.

services-custom/dynamodb-enhanced/src/test/java/software/amazon/awssdk/enhanced/dynamodb/functionaltests/models/RecordWithDefaultTTL.java
```diff
@@ -0,0 +1,61 @@
...
+package software.amazon.awssdk.enhanced.dynamodb.functionaltests.models;
+
+import java.time.Instant;
+import software.amazon.awssdk.enhanced.dynamodb.extensions.annotations.DynamoDbAutoGeneratedTimestampAttribute;
+import software.amazon.awssdk.enhanced.dynamodb.extensions.annotations.DynamoDbTimeToLiveAttribute;
+import software.amazon.awssdk.enhanced.dynamodb.mapper.annotations.DynamoDbBean;
+import software.amazon.awssdk.enhanced.dynamodb.mapper.annotations.DynamoDbPartitionKey;
+
+@DynamoDbBean
+public class RecordWithDefaultTTL {
```
valeriodelbello-amazon, line 25: Is this and `RecordWithSimpleTTL` dead code?

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/mapper/BeanTableSchema.java
```diff
@@ -283,10 +286,21 @@ private static <T> StaticTableSchema<T> createStaticTableSchema(Class<T> beanCla
...
+                                                                              DynamoDbTimeToLiveAttribute.class);
+            if (ttlAnnotation != null) {
+                ttlAttributesCount.getAndIncrement();
+            }
         });
 
         builder.attributes(attributes);
 
+        if (ttlAttributesCount.intValue() > 1) {
+            throw new IllegalArgumentException(
```
valeriodelbello-amazon, line 300: I think we should cover this in the unit tests.

valeriodelbello-amazon:
Should we add TTL validation to `ImmutableTableSchema` too?

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/extensions/TimeToLiveExtension.java
```diff
@@ -0,0 +1,186 @@
...
+
+    private static final class TimeToLiveAttribute implements StaticAttributeTag {
+
+        public String baseField;
+        public long duration;
+        public ChronoUnit unit;
+
+        private TimeToLiveAttribute(String baseField, long duration, ChronoUnit unit) {
+            this.baseField = baseField;
+            this.duration = duration;
```
valeriodelbello-amazon, line 168: Should we check this is not negative somewhere?

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/extensions/TimeToLiveExtension.java
andreas-grafenberger: 1) Agree, but this is the pattern used for the other extensions.
2) I would go with the last approach, and create an @SdkInternalApi shared utility class - and place there all *_ATTRIBUTE_METADATA_KEY for extensions. Would make sense? wouldn't make too much sense to create the class only for TimeToLive extension attributes... what do you think?

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/extensions/TimeToLiveExtension.java
andreas-grafenberger: Added validation for duration in modifyMetadata and beforeWrite, I will have a look if there is a better place for this validation.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/extensions/TimeToLiveExtension.java
andreas-grafenberger: Agree, the validation is applied only on BeanTableSchema and I am preparing parameterized tests to cover the remaining schemas with validation updated - tests still in progress, have pretty many scenarios to cover with different record types.

Also, I feel like `addCustomMetadataObject` is not consistent in terms of validation: 
```java
       private void mergeCustomMetaDataObject(String key, Object object) {
            if (object instanceof Collection) {
                this.addCustomMetadataObject(key, (Collection<Object>) object);
            } else if (object instanceof Map) {
                this.addCustomMetadataObject(key, (Map<Object, Object>) object);
            } else {
                this.addCustomMetadataObject(key, object);
            }
        }
```
because the else branch throws an exception when finds duplicates:
```java
        public Builder addCustomMetadataObject(String key, Object object) {
            if (customMetadata.containsKey(key)) {
                throw new IllegalArgumentException("Attempt to set a custom metadata object that has already been set. "
                                                   + "Custom metadata object key: " + key);
            }

            customMetadata.put(key, object);
            return this;
        }
```

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/DynamoDbTable.java
andreas-grafenberger: javadoc added

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/extensions/TimeToLiveExtension.java
andreas-grafenberger: Agree, updated from public to private final.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/extensions/TimeToLiveExtension.java
andreas-grafenberger: Added validation.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/extensions/TimeToLiveExtension.java
andreas-grafenberger: Should be okay now, moved the new HashMap right before putting ttlAttributeName.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/model/UpdateTimeToLiveEnhancedResponse.java
andreas-grafenberger: Yes, updated.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/model/UpdateTimeToLiveEnhancedResponse.java
andreas-grafenberger: Updated, should be fine now.

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/operations/UpdateTimeToLiveOperation.java
```diff
@@ -0,0 +1,92 @@
...
+    @Override
+    public UpdateTimeToLiveRequest generateRequest(TableSchema<T> tableSchema,
+                                                   OperationContext operationContext,
+                                                   DynamoDbEnhancedClientExtension extension) {
+        Map<String, ?> customTTLMetadata = tableSchema.tableMetadata()
+                                                      .customMetadataObject(CUSTOM_METADATA_KEY, Map.class).orElse(null);
+        if (customTTLMetadata == null) {
+            throw new IllegalArgumentException("Custom TTL metadata object is null");
+        }
+        String ttlAttributeName = (String) customTTLMetadata.get("attributeName");
```
ysunio, line 62: Why are we not using validateMetadataValue helper method ?

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/internal/operations/UpdateTimeToLiveOperation.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    @Override
+    public UpdateTimeToLiveRequest generateRequest(TableSchema<T> tableSchema,
+                                                   OperationContext operationContext,
+                                                   DynamoDbEnhancedClientExtension extension) {
+        Map<String, ?> customTTLMetadata = tableSchema.tableMetadata()
+                                                      .customMetadataObject(CUSTOM_METADATA_KEY, Map.class).orElse(null);
+        if (customTTLMetadata == null) {
+            throw new IllegalArgumentException("Custom TTL metadata object is null");
```
ysunio, line 60: I think this Message is Confusing a bit, it exposing Internal Logic, instead of helping caller to help fix the issue.

It should be more readable to caller having something like `No TTL attribute is configured in the table schema for table, Use @DynamoDbTimeToLiveAttribute on a Long getter....`

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/extensions/annotations/DynamoDbTimeToLiveAttribute.java
```diff
@@ -0,0 +1,68 @@
...
+ * To use this, the annotated method should return a {@link Long} value, which will be populated by the SDK at write time.
+ * The {@code baseField} can be a temporal type such as {@link java.time.Instant}, {@link java.time.LocalDate},
+ * {@link java.time.LocalDateTime}, etc., or a {@link Long} representing epoch seconds directly, serving as the reference point
+ * for TTL calculation.
+ */
+@Target(ElementType.METHOD)
+@Retention(RetentionPolicy.RUNTIME)
+@BeanTableSchemaAttributeTag(TimeToLiveAttributeTags.class)
+@SdkPublicApi
+public @interface DynamoDbTimeToLiveAttribute {
```
ysunio, line 43: Classes / Interfaces should have @ since tags

services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/extensions/TimeToLiveExtension.java
```diff
@@ -0,0 +1,201 @@
...
+        Map<?, ?> customTTLMetadata = context.tableMetadata()
+                                             .customMetadataObject(CUSTOM_METADATA_KEY, Map.class).orElse(null);
+
+        if (customTTLMetadata != null) {
+            String ttlAttributeName = validateMetadataValue(customTTLMetadata, "attributeName", String.class);
+            String baseFieldName = validateMetadataValue(customTTLMetadata, "baseField", String.class);
+            long duration = validateMetadataValue(customTTLMetadata, "duration", Long.class);
+            TemporalUnit unit = validateMetadataValue(customTTLMetadata, "unit", TemporalUnit.class);
+
+            Validate.isTrue(duration >= 0, "Custom TTL metadata key 'duration' must not be negative.");
```
ysunio, line 74: I think we should add baseFieldName check as well, here.
What happens if User does @DynamoDbTimeToLiveAttribute(duration=30, unit=DAYS) — forgot baseField.

It would be silent no-op when duration > 0 and baseField="".

PR #6154, opened 2025-06-03
Fix Apache5 HTTP client retry failures with non-resettable streams

joviegas:
## Motivation and Context

When using S3Client with Apache5 HTTP client, PutObject/GetObject requests fail immediately with the following error:
```
java.lang.RuntimeException: Could not create new stream:
at software.amazon.awssdk.http.auth.aws.internal.signer.io.ResettableContentStreamProvider.newStream(ResettableContentStreamProvider.java:38)
...
Caused by: java.io.IOException: mark/reset not supported
```

### Root Cause:
- RepeatableInputStreamRequestEntity constructor calls getContent() which triggers ContentStreamProvider.newStream()
-  ResettableContentStreamProvider.newStream() attempts to reset the stream before returning it
- The underlying stream doesn't support mark/reset operations, causing immediate failure
- This happens during request preparation, before any retry attempts

## Modifications

Added EntityCreationResult helper class

- Captures both the InputStreamEntity and content stream during creation
- Avoids calling getContent() which would trigger ContentStreamProvider.newStream() and cause the reset failure
- Ensures we use the original stream instance without triggering resets

Stored content stream reference as a field
- Prevents multiple ContentStreamProvider.newStream() calls
- Avoids triggering stream reset operations during entity lifecycle

Fixed resource management
- Removed manual stream closing to prevent double-closing
- InputStreamEntity properly handles stream lifecycle

## Testing

- Added Junits
- Tested S3 Integ and Junits , all passed with Apache5.x upgrade

## License


http-clients/apache-client/src/test/java/software/amazon/awssdk/http/apache/internal/RepeatableInputStreamRequestEntityTest.java
```diff
@@ -790,5 +743,161 @@ void multipleOperations_StatePreservation_WorksCorrectly() throws IOException {
...
+    void getContent_ReusesExistingStream() throws IOException {
+        InputStream originalStream = new ByteArrayInputStream("content".getBytes());
+        entity = createEntity(originalStream);
+        InputStream content1 = entity.getContent();
+        InputStream content2 = entity.getContent();
+        assertSame(content1, content2);
+    }
+
+    @Test
+    @DisplayName("Empty stream should be repeatable")
```
Fred1155, line 880: Hi just trying to understand, in this circumstance, is it equivalent to create an Entity with InputStream with zero length?

http-clients/apache-client/src/test/java/software/amazon/awssdk/http/apache/internal/RepeatableInputStreamRequestEntityTest.java
Fred1155, line 1: Should we add tests for concurrent access scenarios to ensure concurrent access to writeTo method is safe, or that has already been covered in other places?

RanVaknin: I think the reason why `writeTo_ConcurrentWrites_HandlesCorrectly` was removed is because concurrent reads from an inputStream is not thread safe?

joviegas: Yes, concurrent reads from an InputStream are not thread-safe, and if we need to implement this functionality, we would need to add proper synchronization mechanisms. Since the Apache 4.x client doesn't have support for this, I removed this test case.
Note that I added this test case in my previous PR where I included several test cases to RepeatableInputStreamRequestEntityTest to verify all possible scenarios in both 4.x and 5.x versions to ensure consistent functionality. However, at that time, I didn't test for flakiness by running it in loop mode. When I later ran it as a @Repeated test for 1000 iterations, it started failing.
So initially I thought it was fine doing concurrent access for Apache 4.x since it was passing (it fails intermittently )

http-clients/apache-client/src/test/java/software/amazon/awssdk/http/apache/internal/RepeatableInputStreamRequestEntityTest.java
joviegas: Yeap , that is done here 
https://github.com/aws/aws-sdk-java-v2/blob/0bdf8786cd7ff5d5b12872d8c78c70f4169fe3ce/http-clients/apache5-client/src/main/java/software/amazon/awssdk/http/apache5/internal/RepeatableInputStreamRequestEntity.java#L151-L154

Fred1155 (approved):

PR #6157, opened 2025-06-04
Fix expiration in past warning from ProfileFileRefresher.

alextwoods:
Fixes a warning logged from the `CachedSupplier` used by the `ProfileFileRefresher` that expiration (stale time) is in the past.

## Motivation and Context
`ProfileFileSupplier.defaultSupplier()` and `ProfileFileSupplier.reloadWhenModified()` always [logs a warning message](https://github.com/aws/aws-sdk-java-v2/blob/6f4d952240dbc34b4a2478d6938f64a7cfb44adc/utils/src/main/java/software/amazon/awssdk/utils/cache/CachedSupplier.java#L243) indicating that the expiration/stale time is in the past.  This happens every time the value is accessed from the cache, since we compare `clock.instant()` calculated at 2 separate instances:
* first in `ProfileFileRefresher` - https://github.com/aws/aws-sdk-java-v2/blob/master/core/profiles/src/main/java/software/amazon/awssdk/profiles/internal/ProfileFileRefresher.java#L86
* then in `CachedSupplier` - https://github.com/aws/aws-sdk-java-v2/blob/master/utils/src/main/java/software/amazon/awssdk/utils/cache/CachedSupplier.java#L236

## Modifications
This change adds a stale time to `now` in the `ProfileFileRefresher` preventing the warning from being logged - and preventing the CachedSupplier from [adding 1 second to the stale time](https://github.com/aws/aws-sdk-java-v2/blob/6f4d952240dbc34b4a2478d6938f64a7cfb44adc/utils/src/main/java/software/amazon/awssdk/utils/cache/CachedSupplier.java#L242) when it detects this case.

**Does this change any refresh behavior? Does this change when changes in the profile files are reloaded?**
No - there should not be behavior changes.
In the current code, the `ProfileFileRefresher` sets a stale time of `now`, which always triggers the `ProfileFileRefresher` to detect a past expiration and [adds 1 second to the stale time](https://github.com/aws/aws-sdk-java-v2/blob/6f4d952240dbc34b4a2478d6938f64a7cfb44adc/utils/src/main/java/software/amazon/awssdk/utils/cache/CachedSupplier.java#L242).  This happens even when the values of `clock.instant()` are identical because the values are checked using [isBefore](https://github.com/aws/aws-sdk-java-v2/blob/6f4d952240dbc34b4a2478d6938f64a7cfb44adc/utils/src/main/java/software/amazon/awssdk/utils/cache/CachedSupplier.java#L236) which checks strictly less than and not equal.  

This change adds the same 1 second delay to the stale time, but does so explicitly in the `ProfileFileRefresher`. 

## Testing
Modified existing test cases.  Note that the test cases were incorrectly assuming that the CachedSupplier would use the jitter path - however, jitter is only applied in prefetch cases and when the `staleValueBehavior` is set to `ALLOW` - but in the `ProfileFileRefresher`, prefetch is never done (because prefetch time defaults to Instant.MAX) and staleValueBehavior is always `STRICT`.

## Types of changes

## Checklist

## License


.changes/next-release/bugfix-AWSSDKforJavav2-8565660.json
```diff
@@ -0,0 +1,6 @@
+{
+    "type": "bugfix",
+    "category": "AWS SDK for Java v2",
+    "contributor": "",
+    "description": "Fix expiration in past warning from ProfileFileRefresher."
```
RanVaknin: nit: Since ProfileFileRefresher is an internal implementation detail, I wouldn't mention in user-facing release notes. Maybe something like  "Fix expiration in past warning during credential profile loading"

RanVaknin (approved):

core/profiles/src/main/java/software/amazon/awssdk/profiles/internal/ProfileFileRefresher.java
```diff
@@ -39,6 +39,7 @@ public final class ProfileFileRefresher {
     private static final ProfileFileRefreshRecord EMPTY_REFRESH_RECORD = ProfileFileRefreshRecord.builder()
                                                                                                  .refreshTime(Instant.MIN)
                                                                                                  .build();
+    private static final long STALE_TIME_MS = 1000;
```
joviegas, line 42: Can we add `LogCaptor ` to make sure that warning  is no longer logged  twice after this fix 
something like https://github.com/aws/aws-sdk-java-v2/blob/f272552bff83a46dbe3cc8fa11718d4157529a42/core/http-auth-aws/src/test/java/software/amazon/awssdk/http/auth/aws/internal/signer/DefaultRequestSignerTest.java#L72-L74

alextwoods: Good call - added an assertion that no warnings are logged.

joviegas (approved):

PR #6161, opened 2025-06-05
Add support for service model protocols field

davidh44:
## Motivation and Context

Add support for `protocols` field in service model which takes list of protocols

## Modifications

- Update codegen `ServiceMetadata` to add `protocols` field
- Add `ProtocolUtils` class to resolve protocol from `protocols` and `protocol` service model fields
  - `protocols` supersedes `protocol`
  - Highest priority protocol supported by the SDK that is present in `protocols` will be selected
  - Priority ordered list in SDK: `[ "smithy-rpc-v2-cbor", "json", "rest-json", "rest-xml", "query", "ec2"]`
- Update all usages of `ServiceMetadata.getProtocol()` to use `ProtocolUtils.resolveProtocol()`

## Testing

- Added unit test for `ProtocolUtils`
- Updated RpcV2 Cbor codegenerated test service model to include multiple values in `protocols` and removed `protocol`

## Types of changes

## Checklist

## License


codegen/src/main/java/software/amazon/awssdk/codegen/model/service/ServiceMetadata.java
```diff
@@ -111,6 +113,14 @@ public void setProtocol(String protocol) {
         this.protocol = protocol;
     }
 
+    public List<String> getProtocols() {
```
zoewangg, line 125: Can we deprecate `getProtocol()` method and add documentation to link to ProtocolUtils.resolveProtocol?

Should we add checkstyle rule to prevent the use of getProtocol so that we don't use it unintentionally in the future?

zoewangg (changes requested):

codegen/src/main/java/software/amazon/awssdk/codegen/model/service/ServiceMetadata.java
davidh44: Updated to deprecate getProtocol() , and added cbor check for Kinesis customization

There are other getProtocol() methods of same name in other classes so can't use simple checkstyle rule, would need to write custom class. I think the deprecation should be fine for now, wdyt?

codegen/src/main/java/software/amazon/awssdk/codegen/utils/ProtocolUtils.java
```diff
@@ -0,0 +1,59 @@
...
+     * {@code protocols} supersedes {@code protocol}. The highest priority protocol supported by the SDK that is present in the
+     * service model {@code protocols} list will be selected. If none of the values in {@code protocols} is supported by the
+     * SDK, an error will be thrown. If {@code protocols} is empty or null, the value from {@code protocol} will be returned.
+     */
+    public static String resolveProtocol(ServiceMetadata serviceMetadata) {
+
+        List<String> protocols = serviceMetadata.getProtocols();
+        String protocol = serviceMetadata.getProtocol();
+
+        // Kinesis uses customization.config customServiceMetadata to set cbor
```
zoewangg: Instead of hardcoding cbor, should we honor customServiceMetadata instead, i.e., checking the presence of custom service metadata

davidh44: The field is not currently exposed, we check for it in the MetadataModifiersProcessor and override protocol there.

https://github.com/aws/aws-sdk-java-v2/blob/master/codegen/src/main/java/software/amazon/awssdk/codegen/customization/processors/MetadataModifiersProcessor.java#L47

codegen/src/main/java/software/amazon/awssdk/codegen/utils/ProtocolUtils.java
```diff
@@ -43,11 +43,15 @@ public static String resolveProtocol(ServiceMetadata serviceMetadata) {
...
         String protocol = serviceMetadata.getProtocol();
 
-        // Kinesis uses customization.config customServiceMetadata to set cbor
-        if ("cbor".equals(protocol) || protocols == null || protocols.isEmpty()) {
+        if (protocols == null || protocols.isEmpty()) {
             return protocol;
         }
 
+        // Kinesis uses customization.config customServiceMetadata to set cbor
+        if ("cbor".equals(protocols.get(0))) {
```
zoewangg, line 51: Can we add a test case for cbor?

zoewangg (approved):

Issue #6108, opened 2025-05-14
Fast Arn.fromString implementation


rgithubli:
### Describe the feature

When parsing arn from a string, while throwing an exception is a way to convey the string is invalid, it's expensive, especially if it's on a hot code path and the client doesn't use the stacktrace at all. [code](https://github.com/aws/aws-sdk-java-v2/blob/master/core/arns/src/main/java/software/amazon/awssdk/arns/Arn.java#L150-L183). Based on internal Amazon profiling results, it can cause > 5% CPU cycles for some of the services to fill the stack trace when parsing arn, while spending >1% CPU cycles on exceptions can be considered needing optimizations.

To improve the performance, we can provide an API that doesn't throw exception, e.g. the API returns an Optional or throws exception but overrides its fillInStackTrace to avoid the expensive stacktrace filling. 

### Use Case

`Arn.fromString` parsing takes long CPU cycle due to unnecessary stacktrace filling.

### Proposed Solution

Keep existing [Arn.fromString](https://github.com/aws/aws-sdk-java-v2/blob/master/core/arns/src/main/java/software/amazon/awssdk/arns/Arn.java#L150) API but also provide another API that either returns Optional or throws exception that does not fill stacktrace.

### Acknowledgements

- [x] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### AWS Java SDK version used

Both v1 and v2

### JDK version used

17, but it doesn't matter

### Operating System and version

Linux

debora-ito:
@rgithubli We'll look into this, thank you for bringing it up.

debora-ito:
Changed implemented via https://github.com/aws/aws-sdk-java-v2/pull/6163 and released in version `2.31.66`.

PR #6163, opened 2025-06-06
Better arn parsing

RanVaknin:
## Motivation and Context
Addresses #6108 .

Currently the SDK uses exceptions as control flow for arn parsing. This is an anti pattern in Java and can cause performance overhead. 

## Modifications
This PR adds another arn parsing method `tryFromString(arn)` that would return an Optional of Arn when the arn is valid, and an empty when it is invalid - this avoids the hot path of throwing exceptions as control flow.

## Testing
Unit tests:
- Valid ARNs
- Invalid ARNs
- Exception throwing ARNs (valid ARNs missing a partition or service)

### DIFF
--- a/core/arns/src/main/java/software/amazon/awssdk/arns/Arn.java
+++ b/core/arns/src/main/java/software/amazon/awssdk/arns/Arn.java
@@ -142,7 +142,6 @@ public final class Arn implements ToCopyableBuilder<Arn.Builder, Arn> {
      * Attempts to parse the given string into an {@link Arn}. If the input string is not a valid ARN,
      * this method returns {@link Optional#empty()} instead of throwing an exception.
      * <p>
-     * Returns an empty Optional if the input string is not a valid ARN.
      * When successful, the resource is accessible entirely as a string through
      * {@link #resourceAsString()}. Where correctly formatted, a parsed resource
      * containing resource type, resource and qualifier is available through
--- a/core/arns/src/main/java/software/amazon/awssdk/arns/Arn.java
+++ b/core/arns/src/main/java/software/amazon/awssdk/arns/Arn.java
@@ -149,57 +149,10 @@ public final class Arn implements ToCopyableBuilder<Arn.Builder, Arn> {
      *
      * @param arn A string containing an ARN to parse.
      * @return An {@link Optional} containing the parsed {@link Arn} if valid, or empty if invalid.
+     * @throws IllegalArgumentException if the ARN contains empty partition or service fields
      */
     public static Optional<Arn> tryFromString(String arn) {
-        if (arn == null) {
-            return Optional.empty();
-        }
-
-        int arnColonIndex = arn.indexOf(':');
-        if (arnColonIndex < 0 || !"arn".equals(arn.substring(0, arnColonIndex))) {
-            return Optional.empty();
-        }
-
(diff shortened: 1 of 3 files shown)


davidh44 (approved):
Looks good, just need a a changelog

core/arns/src/main/java/software/amazon/awssdk/arns/Arn.java
```diff
@@ -138,6 +138,70 @@ public static Builder builder() {
...
+     * <p>
+     * When successful, the resource is accessible entirely as a string through
+     * {@link #resourceAsString()}. Where correctly formatted, a parsed resource
+     * containing resource type, resource and qualifier is available through
+     * {@link #resource()}.
+     *
+     * @param arn A string containing an ARN to parse.
+     * @return An {@link Optional} containing the parsed {@link Arn} if valid, or empty if invalid.
+     */
+    public static Optional<Arn> tryFromString(String arn) {
```
zoewangg, line 154: It contains a lot of duplicate code. Can we extract common code?

One idea is to create a helper method that takes a boolean for throwing behavior `private static Optional<Arn> parseArn(String arn, boolean throwOnError) {...}`

core/arns/src/test/java/software/amazon/awssdk/arns/ArnTest.java
```diff
@@ -311,4 +317,78 @@ public void invalidArnWithoutAccountId_ThrowsIllegalArgumentException() {
...
+        assertThat(optionalArn).isEmpty();
+    }
+
+    @ParameterizedTest(name = "{0}")
+    @MethodSource("exceptionThrowingArnTestCases")
+    public void tryFromString_InvalidArns_ShouldThrowExceptions(String testName, String arnString) {
+        assertThrows(IllegalArgumentException.class, () -> {
+            Arn.tryFromString(arnString);
+        });
+    }
```
zoewangg, line 392: Hmm, I thought it would no longer throw exception, did I miss anything?

zoewangg (changes requested):
Don't we need to update account ID code path to use this new method?

core/arns/src/test/java/software/amazon/awssdk/arns/ArnTest.java
RanVaknin: I agree, but I matched the behavior to the behavior that was backported to v1.
I think the original implementation was when `service` or `partition` are not provided, we don't explicitly throw an exception within this function. Instead we try to build the arn using the Arn.Builder and use `Validate.paramNotBlank` in the constructor https://github.com/aws/aws-sdk-java-v2/blob/master/core/arns/src/main/java/software/amazon/awssdk/arns/Arn.java#L84-L85

One could argue this is valid validation rather than the same type of "exceptions used as control flow" functionality that currently exists in `fromString`. I don't feel strongly about it either way. Happy to change it.

RanVaknin:
> Don't we need to update account ID code path to use this new method?

Separate internal Arn class used in endpoints logic already returns Optional
https://github.com/aws/aws-sdk-java-v2/blob/master/codegen/src/main/resources/software/amazon/awssdk/codegen/rules/Arn.java.resource#L26

zoewangg (approved):

PR #6165, opened 2025-06-09
Fix connectionPoolingWorks test for Apache5.x by fixing setConnectionTimeToLive for Apache5.x

joviegas:
### Issue

After migrating from Apache HttpClient 4.x to 5.x,  `connectionPoolingWorks()` connection pooling tests are failing with the assertion error:
```
java.lang.AssertionError: 
Expecting actual:
  5
to be less than or equal to:
  2
```
Expected Behavior: HTTP connections should be reused from the connection pool for subsequent requests, creating only 1-2 total connections for 5 requests.

Actual Behavior: Each request creates a new connection (5 total connections), indicating that connections are not being reused despite being properly returned to the pool.

### Observed Symptoms:
- Pool correctly shows Available: 1, Leased: 0 between requests
- Connections are marked as "can be kept alive for 60000 MILLISECONDS"
- However, connections are immediately closed with "close connection GRACEFUL" when leased from pool
-  New connections are created for each subsequent request

### Root Cause Analysis

The Core Issue: Semantic Breaking Change

Through systematic debugging, we isolated the issue to the setConnectionTimeToLive() configuration in ApacheConnectionManagerFactory.create():

```
// Apache 4.x PoolEntry constructor logic
if (timeToLive > 0) {
    final long deadline = this.created + timeUnit.toMillis(timeToLive);
    this.validityDeadline = deadline > 0 ? deadline : Long.MAX_VALUE;
} else {
    // timeToLive <= 0 treated as INFINITE (never expires)
    this.validityDeadline = Long.MAX_VALUE;  
}
```

Apache 5.x Behavior:
```
    timeToLive = 0: Immediate expiration (connections expire instantly)
    timeToLive > 0: Normal expiration behavior
```

## Modifications

- We set the `setConnectionTimeToLive` on Apache 5.x only if a non Zero value is set , since Apache 5.x defaults it to -1 which is Infinite value

## Testing

- Junit added back with this PR , earlier it was commented.

## License


http-clients/apache5-client/src/main/java/software/amazon/awssdk/http/apache5/Apache5HttpClient.java
```diff
@@ -694,16 +694,17 @@ public PoolingHttpClientConnectionManager create(Apache5HttpClient.DefaultBuilde
...
                                                          .setSchemePortResolver(DefaultSchemePortResolver.INSTANCE)
-                                                         .setDnsResolver(configuration.dnsResolver)
-                                                         .setConnectionTimeToLive(
-                                                             TimeValue.of(standardOptions.get(
-                                                                 SdkHttpConfigurationOption.CONNECTION_TIME_TO_LIVE).toMillis(),
-                                                                          TimeUnit.MILLISECONDS))
-                                                         .build();
+                                                         .setDnsResolver(configuration.dnsResolver);
+            Duration connectionTtl = standardOptions.get(SdkHttpConfigurationOption.CONNECTION_TIME_TO_LIVE);
+            if (!connectionTtl.isZero()) {
```
joviegas, line 703: Note that SdkHttpConfigurationOption sets default value as 0 in https://github.com/aws/aws-sdk-java-v2/blob/86bb4402dfebb782122b36a7f9e4a744bbfb81ba/http-client-spi/src/main/java/software/amazon/awssdk/http/SdkHttpConfigurationOption.java#L145

http-clients/apache5-client/src/test/java/software/amazon/awssdk/http/apache5/Apache5HttpClientWireMockTest.java
```diff
@@ -153,12 +153,6 @@ public void credentialPlannerIsInvoked() throws Exception {
         mockProxyServer.verify(2, RequestPatternBuilder.allRequests());
     }
 
-
-    @Override
-    public void connectionPoolingWorks() throws Exception {
-        // TODO : future PR will handle this.
-    }
-
```
davidh44, line 161: > Junit added back with this PR , earlier it was commented.

Is that this test? Its being removed not added back

joviegas: Since Apache5HttpClientWireMockTest extends SdkHttpClientTestSuite

Its been added back , since we removed the empty test from here it will pick up the actual test from 
https://github.com/aws/aws-sdk-java-v2/blob/0af784da5b9134204cad8bd691af5f4530ef86ef/test/http-client-tests/src/main/java/software/amazon/awssdk/http/SdkHttpClientTestSuite.java#L121-L146

davidh44 (approved):

PR #6172, opened 2025-06-11
Add validation for .brazil.json

dagnir:
## Motivation and Context
This commit adds the `validate-brazil-config` script, which validates the .brazil.json file for correctness. It validates that all modules of the SDK are mapped to a package internally, or explicitly skipped. Additional, it ensures that all external dependencies are also mapped to internal packages.

The buildspecs/validate-brazil-config.yml is so that this validation can run as PR build check.

## Testing
 - Setup codebuild job that uses buildspecs/validate-brazil-config.yml
 - Test against commit with unmapped dependency and ensure build fails.

## Types of changes

## Checklist

## License


.github/workflows/codebuild-ci.yml
```diff
@@ -139,6 +139,20 @@ jobs:
...
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          role-to-assume: ${{ secrets.CI_AWS_ROLE_ARN }}
+          aws-region: us-west-2
+          role-duration-seconds: 7200
+      - name: Validate Brazil config
+        uses: aws-actions/aws-codebuild-run-build@v1
+        with:
+          project-name: aws-java-sdk-v2-validate-brazil-config
```
L-Applin, line 155: `aws-java-sdk` or `aws-sdk-java` ? this file uses both 😭

L-Applin (approved):

.github/workflows/codebuild-ci.yml
dagnir: I just copied one of the existing jobs, I don't think we've been very consistent unfortunately!

.github/workflows/codebuild-ci.yml
```diff
@@ -139,6 +139,20 @@ jobs:
         uses: aws-actions/aws-codebuild-run-build@v1
         with:
           project-name: aws-sdk-java-v2-endpoints-test
+  brazil-json-validation:
+    if: github.repository == 'aws/aws-sdk-java-v2'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v1
```
RanVaknin, line 147: nit: is there a reason we are still using v1? Configure AWS credentials is currently on v4

scripts/validate-brazil-config
```diff
@@ -0,0 +1,109 @@
...
+#   ...
+
+brazil_import_config_path=".brazil.json"
+
+with open(brazil_import_config_path) as f:
+  brazil_import_config = json.loads(f.read())
+
+modules_path=sys.argv[1]
+
+service_modules = set()
```
RanVaknin: You are creating `service_modules` and adding to this set in line 64, but not doing anything with it after. There is only core module validation, is this intentional? 

Is this related to the 
`"services": { "skipImport": true }` part in the `.brazil.json` file somehow?

.github/workflows/codebuild-ci.yml
```diff
@@ -139,6 +139,20 @@ jobs:
         uses: aws-actions/aws-codebuild-run-build@v1
         with:
           project-name: aws-sdk-java-v2-endpoints-test
+  brazil-json-validation:
```
zoewangg, line 142: Do we actually need code build job? Can we just use GH CI directly?

.github/workflows/codebuild-ci.yml
dagnir: No particular reason, just what we're using everywhere else

scripts/validate-brazil-config
dagnir: Good point, I'll remove it

.github/workflows/codebuild-ci.yml
dagnir: Only advantage is we still need to build the project, so being on codebuild lets us keep the build time low by using larger instances.

RanVaknin (approved):

PR #6176, opened 2025-06-12
EC2 IMDS Changes to Support Account ID

S-Saranya1:
Add support for extracting account ID from Instance Metadata Credentials Provider

## Motivation and Context
Add support in IMDS credentials provider to retrieve accountID.

## Modifications
- Added support for new `/latest/meta-data/iam/security-credentials-extended/` IMDS endpoint
- Added fallback logic to legacy endpoint '/latest/meta-data/iam/security-credentials/{ROLE}'
- Added support for AccountId to extract it from IMDS response.

## Testing

Added unit tests using WireMock to verify:

- Successful retrieval of credentials with account ID from extended endpoint
- Proper fallback to legacy endpoint when extended endpoint returns 404

Performed integration testing on EC2:

  - Verified successful account ID resolution on allowlisted instances
  - Confirmed fallback to legacy endpoint on non-allowlisted instances

Tested Code:

```
public class ImdsAccountIdTest {
        try {
            InstanceProfileCredentialsProvider provider = InstanceProfileCredentialsProvider.create();
            AwsCredentials credentials = provider.resolveCredentials();
            System.out.println("Successfully resolved credentials!");
            System.out.println("Access Key ID: " + credentials.accessKeyId().substring(0, 5) + "...");
            System.out.println("Account ID: " + credentials.accountId().orElse("Not available"));

        } catch (Exception e) {
            System.err.println("Error: " + e.getMessage());
            e.printStackTrace();
            System.exit(1);
        }
    }
}

```

## Types of changes

## Checklist

## License


core/auth/src/main/java/software/amazon/awssdk/auth/credentials/InstanceProfileCredentialsProvider.java
```diff
@@ -70,9 +72,19 @@ public final class InstanceProfileCredentialsProvider
...
+
+    private enum ApiVersion {
+        UNKNOWN,
+        LEGACY,
+        EXTENDED
+    }
+
     private static final String EC2_METADATA_TOKEN_TTL_HEADER = "x-aws-ec2-metadata-token-ttl-seconds";
     private static final String DEFAULT_TOKEN_TTL = "21600";
+    private Lazy<ApiVersion> apiVersion = new Lazy<>(() -> ApiVersion.UNKNOWN);
```
alextwoods: Is there a reason we are using a Lazy for this?  We don't really need to cache the value of it

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/InstanceProfileCredentialsProvider.java
```diff
@@ -70,9 +72,19 @@ public final class InstanceProfileCredentialsProvider
...
+    private Lazy<String> resolvedProfile = new Lazy<>(() -> null);
```
L-Applin: Why Lazy for those?

core/auth/src/test/java/software/amazon/awssdk/auth/credentials/InstanceProfileCredentialsProviderAccountIDTest.java
```diff
@@ -0,0 +1,181 @@
...
+import software.amazon.awssdk.utils.DateUtils;
+
+import java.time.Duration;
+import java.time.Instant;
+
+/**
+ * Tests verifying IMDS credential resolution with account ID support.
+ */
+@WireMockTest
+public class InstanceProfileCredentialsProviderAccountIDTest {
```
alextwoods: Does this fully cover all of the tests from the SEP? (ie, the json file).

Also a nit on naming - we might want to focus on testing the extended api rather than just accountID.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/InstanceProfileCredentialsProvider.java
zoewangg: Hmm, what I originally had in mind for this is to initialize `Lazy(() -> findApiVersion())`

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/InstanceProfileCredentialsProvider.java
S-Saranya1: Updated them.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/InstanceProfileCredentialsProvider.java
S-Saranya1: Yeah, agree. Updated it.

core/auth/src/test/java/software/amazon/awssdk/auth/credentials/InstanceProfileCredentialsProviderAccountIDTest.java
S-Saranya1: Added additional tests, now it covers all the tests from the SEP json, it doesn't cover the tests with the "ec2InstanceProfileName" configuration.

S-Saranya1: Updated the test file name.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/InstanceProfileCredentialsProvider.java
S-Saranya1: The implementation of findApiVersion() method will have extra request to check if the ApiVersion is legacy or extended right, now with the current implementation we just check the ApiVersion in the getSecurityCredentials() method, change ApiVersion to LEGACY or EXTENDED and follow that ApiVersion throughout. We are not making the extra request again to check the ApiVersion once it's resolved in getSecurityCredentials().

.changes/next-release/feature-AWSEC2-9b178a4.json
```diff
@@ -0,0 +1,6 @@
+{
+    "type": "feature",
+    "category": "AWS EC2",
```
zoewangg: nit: should be `AWS SDK for Java v2`

core/auth/src/it/java/software/amazon/awssdk/auth/credentials/InstanceProfileCredentialsProviderIntegrationTest.java
```diff
@@ -35,7 +35,7 @@ public class InstanceProfileCredentialsProviderIntegrationTest {
     /** Starts up the mock EC2 Instance Metadata Service. */
     @Before
     public void setUp() throws Exception {
-        mockServer = new EC2MetadataServiceMock("/latest/meta-data/iam/security-credentials/");
+        mockServer = new EC2MetadataServiceMock("/latest/meta-data/iam/security-credentials-extended/");
```
zoewangg: Can we keep the existing tests and add new one for the extended uri?

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/InstanceProfileCredentialsProvider.java
```diff
@@ -70,9 +71,19 @@ public final class InstanceProfileCredentialsProvider
...
+
+    private enum ApiVersion {
+        UNKNOWN,
+        LEGACY,
+        EXTENDED
+    }
+
     private static final String EC2_METADATA_TOKEN_TTL_HEADER = "x-aws-ec2-metadata-token-ttl-seconds";
     private static final String DEFAULT_TOKEN_TTL = "21600";
+    private ApiVersion apiVersion = ApiVersion.UNKNOWN;
```
zoewangg, line 89: volatile?

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/InstanceProfileCredentialsProvider.java
```diff
@@ -157,13 +168,24 @@ private RefreshResult<AwsCredentials> refreshCredentials() {
...
                                 .staleTime(staleTime(expiration))
                                 .prefetchTime(prefetchTime(expiration))
                                 .build();
+        } catch (Ec2MetadataClientException e) {
+            if (apiVersion == ApiVersion.UNKNOWN) {
+                apiVersion = ApiVersion.LEGACY;
+                resolvedProfile = null;
+                return refreshCredentials();
+            }
+            throw SdkClientException.create("Failed to load credentials from IMDS.", e);
```
zoewangg: Hmm, this seems brittle, should we check status code to determine whether we should fallback or throw exception?

core/regions/src/main/java/software/amazon/awssdk/regions/util/HttpResourcesUtils.java
```diff
@@ -118,9 +119,9 @@ public String readResource(ResourcesEndpointProvider endpointProvider, String me
                     return IoUtils.toUtf8String(inputStream);
                 } else if (statusCode == HttpURLConnection.HTTP_NOT_FOUND) {
                     // This is to preserve existing behavior of EC2 Instance metadata service.
-                    throw SdkClientException.builder()
-                                            .message("The requested metadata is not found at " + connection.getURL())
-                                            .build();
+                    throw Ec2MetadataClientException.builder()
+                                                    .message("The requested metadata is not found at " + connection.getURL())
```
zoewangg, line 124: I thought we want to include status code here

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/InstanceProfileCredentialsProvider.java
```diff
@@ -285,21 +313,40 @@ private boolean isInsecureFallbackDisabled() {
...
+
+            if (securityCredentials.length == 0) {
+                throw SdkClientException.builder().message("Unable to load credentials path").build();
+            }
 
-        if (securityCredentials.length == 0) {
-            throw SdkClientException.builder().message("Unable to load credentials path").build();
+            if (apiVersion == ApiVersion.UNKNOWN) {
+                apiVersion = ApiVersion.EXTENDED;
+            }
```
zoewangg, line 354: Is there any reason we are setting api version here again?

zoewangg (changes requested):

alextwoods (approved):

.changes/next-release/feature-AWSEC2-9b178a4.json
S-Saranya1: Updated it.

core/auth/src/it/java/software/amazon/awssdk/auth/credentials/InstanceProfileCredentialsProviderIntegrationTest.java
S-Saranya1: Updated it.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/InstanceProfileCredentialsProvider.java
S-Saranya1: Included the status code.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/InstanceProfileCredentialsProvider.java
S-Saranya1: Made it Volatile.

core/regions/src/main/java/software/amazon/awssdk/regions/util/HttpResourcesUtils.java
S-Saranya1: Included the status code.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/InstanceProfileCredentialsProvider.java
S-Saranya1: Discussed offline: This is method where we are setting the initial ApiVersion and will be using it throughout.

Removed the same logic that we had in refreshCredentials() method, since once we set ApiVersion in getSecurityCredentials(), we won't be having UNKNOWN state again in refreshCredentials().

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/InstanceProfileCredentialsProvider.java
```diff
@@ -285,21 +309,40 @@ private boolean isInsecureFallbackDisabled() {
...
+            resolvedProfile = securityCredentials[0];
+            return securityCredentials;
 
-        if (securityCredentials.length == 0) {
-            throw SdkClientException.builder().message("Unable to load credentials path").build();
+        } catch (Ec2MetadataClientException e) {
+            if (apiVersion == ApiVersion.UNKNOWN) {
```
zoewangg: Do we need to check status code here?
Can we add a test case?

core/auth/src/test/java/software/amazon/awssdk/auth/credentials/InstanceProfileCredentialsProviderExtendedApiTest.java
```diff
@@ -0,0 +1,311 @@
...
+
+        InstanceProfileCredentialsProvider provider = InstanceProfileCredentialsProvider.builder().build();
+
+        assertThatThrownBy(() -> provider.resolveCredentials())
+            .isInstanceOf(SdkClientException.class)
+            .hasMessageContaining("Failed to load credentials from IMDS.");
+    }
+
+    @Test
+    void resolveCredentials_withUnstableProfile_noAccountId_refreshesCredentials() {
```
zoewangg: What's unstable profile?

core/auth/src/test/java/software/amazon/awssdk/auth/credentials/InstanceProfileCredentialsProviderExtendedApiTest.java
```diff
@@ -0,0 +1,311 @@
...
+import software.amazon.awssdk.utils.DateUtils;
+
+import java.time.Duration;
+import java.time.Instant;
+
+/**
+ * Tests verifying IMDS credential resolution with account ID support.
+ */
+@WireMockTest
+public class InstanceProfileCredentialsProviderExtendedApiTest {
```
zoewangg, line 49: Do we have a test case to validate that we don't fall back if the status code is not 404?

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/HttpCredentialsLoader.java
```diff
@@ -64,15 +64,17 @@ public LoadedCredentials loadCredentials(ResourcesEndpointProvider endpoint) {
...
             return new LoadedCredentials(accessKey.text(),
-                                         secretKey.text(),
-                                         token != null ? token.text() : null,
-                                         expiration != null ? expiration.text() : null,
-                                         providerName);
+                                       secretKey.text(),
+                                       token != null ? token.text() : null,
+                                       expiration != null ? expiration.text() : null,
+                                       accountId != null ? accountId.text() : null,
+                                       providerName);
```
zoewangg: nit: indention is a bit off here

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/InstanceProfileCredentialsProvider.java
```diff
@@ -70,9 +71,19 @@ public final class InstanceProfileCredentialsProvider
...
+    private enum ApiVersion {
+        UNKNOWN,
+        LEGACY,
+        EXTENDED
+    }
+
     private static final String EC2_METADATA_TOKEN_TTL_HEADER = "x-aws-ec2-metadata-token-ttl-seconds";
     private static final String DEFAULT_TOKEN_TTL = "21600";
+    private volatile ApiVersion apiVersion = ApiVersion.UNKNOWN;
+    private String resolvedProfile = null;
```
zoewangg, line 90: This also needs to be atomicreference or volatile

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/InstanceProfileCredentialsProvider.java
```diff
@@ -164,6 +175,13 @@ private RefreshResult<AwsCredentials> refreshCredentials() {
                                 .staleTime(staleTime(expiration))
                                 .prefetchTime(prefetchTime(expiration))
                                 .build();
+        } catch (Ec2MetadataClientException e) {
+            if (e.statusCode() == 404 && apiVersion == ApiVersion.EXTENDED) {
```
zoewangg: Can we add some debug-level logs for troubleshooting?

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/InstanceProfileCredentialsProvider.java
zoewangg: Same here, let's add some logs

S-Saranya1: Included the status code, also added the test case for this.

core/auth/src/test/java/software/amazon/awssdk/auth/credentials/InstanceProfileCredentialsProviderExtendedApiTest.java
S-Saranya1: Sorry it's missed, Thanks for checking, added the test case now.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/HttpCredentialsLoader.java
S-Saranya1: Modified it.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/InstanceProfileCredentialsProvider.java
S-Saranya1: Updated apiVersion and resolvedProfile to atomicreference.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/InstanceProfileCredentialsProvider.java
S-Saranya1: Added the debug logs.

core/auth/src/test/java/software/amazon/awssdk/auth/credentials/InstanceProfileCredentialsProviderExtendedApiTest.java
S-Saranya1: This is the test case from SEP. This "unstable profile" test case checks an EC2 instance profile that changes or becomes unavailable between credential refresh attempts, requiring the SDK to re-discover available profiles. But this makes more sense when we added Ec2InstanceProfileName, removing this test case for now.

.changes/next-release/feature-AWSEC2-9b178a4.json
```diff
@@ -0,0 +1,6 @@
+{
+    "type": "feature",
+    "category": "AWS SDK for Java v2",
+    "contributor": "",
+    "description": "EC2 IMDS Changes to Support Account ID"
```
dagnir: This should be more descriptive for customers; as-is I think it would be hard for them to understand what actual changes have been made. Maybe something like "Include the account ID associated with the credentials retrieved from IMDS when available."

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/InstanceProfileCredentialsProvider.java
```diff
@@ -285,21 +310,39 @@ private boolean isInsecureFallbackDisabled() {
...
+
+            apiVersion.compareAndSet(ApiVersion.UNKNOWN, ApiVersion.EXTENDED);
+            resolvedProfile.set(securityCredentials[0]);
+            return securityCredentials;
 
-        if (securityCredentials.length == 0) {
-            throw SdkClientException.builder().message("Unable to load credentials path").build();
+        } catch (Ec2MetadataClientException e) {
+            if (e.statusCode() == 404 && apiVersion.compareAndSet(ApiVersion.UNKNOWN, ApiVersion.LEGACY)) {
+                log.debug(() -> "Unable to load credential path from extended API. Falling back to legacy API.");
```
dagnir: Instead of "Unable to load credential path from extended API.", it might be more clear to say "Instance does not support IMDS extended API."

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/InstanceProfileCredentialsProvider.java
```diff
@@ -164,6 +176,13 @@ private RefreshResult<AwsCredentials> refreshCredentials() {
                                 .staleTime(staleTime(expiration))
                                 .prefetchTime(prefetchTime(expiration))
                                 .build();
+        } catch (Ec2MetadataClientException e) {
+            if (e.statusCode() == 404 && apiVersion.compareAndSet(ApiVersion.EXTENDED, ApiVersion.LEGACY)) {
```
dagnir: We're switching from `EXTENDED` to `LEGACY` here. Is that intended? I don't see this behavior specified in the spec document.

dagnir (changes requested):

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/InstanceProfileCredentialsProvider.java
```diff
@@ -285,21 +310,39 @@ private boolean isInsecureFallbackDisabled() {
...
+            }
+
+            apiVersion.compareAndSet(ApiVersion.UNKNOWN, ApiVersion.EXTENDED);
+            resolvedProfile.set(securityCredentials[0]);
+            return securityCredentials;
 
-        if (securityCredentials.length == 0) {
-            throw SdkClientException.builder().message("Unable to load credentials path").build();
+        } catch (Ec2MetadataClientException e) {
+            if (e.statusCode() == 404 && apiVersion.compareAndSet(ApiVersion.UNKNOWN, ApiVersion.LEGACY)) {
```
dagnir: So it just occurred to me that there might be an issue with just the atomic swap here.

If there are two threads that attempt to get credentials at the same time, T1 and T2, and apiVersion == UNKNOWN, and the instance doesn't support EXTENDED, both T1 and T2 will initially get a 404. BUT since `apiVersion.compareAndSet(ApiVersion.UNKNOWN, ApiVersion.LEGACY)` would only succeed for _either_ T1 or T2, one of them will get `Failed to load credentials from IMDS.` exception. Does that sound right?

dagnir: @zoewangg WDYT?

dagnir: Something like this might work 

```java
ApiVersion currentVersion = apiVersion.get();

if (currentVersion == UNKNOWN) {
    try {
        // get credentials from extended endpoint
        apiVersion.compareAndSet(UNKNOWN, EXTENDED);
    } catch (Ec2MetadataClientException e) {
        if (apiVersion.compareAndSet(UNKNOWN, LEGACY) || apiVersion.get() == LEGACY) {
          // get credentials from LEGACY
        } else {
          throw e;
        }
    }
} else if (currentVersion == EXTENDED) {
    // get credentials from extended or throw
} else {
    // get credentials from legacy or throw
}
```

alextwoods: I'm still a little skeptical of using Atomics and compareAndSet at all here - `refreshCredentials` is a private method called only in a synchronized block (by the cached supplier).  I understand that its being defensive - but I think it also makes the code read as if its intended to be thread safe, but I think we generally wouldn't want to allow multiple concurrent imds http calls and we'd probably want to ensure the logic we're using for getting/using the token would work with concurrent calls....

dagnir: > called only in a synchronized block

Ah yeah Zoe also reminded me about this. I think the AtomicReferences led me down the path to thinking this could be called concurrently

.changes/next-release/feature-AWSEC2-9b178a4.json
S-Saranya1: Updated it.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/InstanceProfileCredentialsProvider.java
S-Saranya1: Updated .

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/InstanceProfileCredentialsProvider.java
S-Saranya1: This is not needed, Thanks for checking. Modified it.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/InstanceProfileCredentialsProvider.java
```diff
@@ -164,6 +177,13 @@ private RefreshResult<AwsCredentials> refreshCredentials() {
                                 .staleTime(staleTime(expiration))
                                 .prefetchTime(prefetchTime(expiration))
                                 .build();
+        } catch (Ec2MetadataClientException e) {
+            if (e.statusCode() == 404) {
```
zoewangg, line 187: Don't we need to check apiVersion here to avoid infinite loop, i.e., if refreshCredentials() fallback call returns 404, we want to throw exception immediately? Can we add tests to catch this?

S-Saranya1: Discussed offline- Decided to keep the counter to retry, we are setting the retry count to 1 to align with the existing behavior(no profile cache).

zoewangg (approved):

alextwoods (approved):
LGTM!

PR #6182, opened 2025-06-13
Add validation for missing request URI

dagnir:
Throw InvalidModelException if we can't find the request URI for an operation when a member has URI mapping.

## Types of changes

## Checklist

## License


zoewangg (approved):

codegen/src/test/java/software/amazon/awssdk/codegen/CodeGeneratorTest.java
```diff
@@ -201,17 +216,28 @@ private static Path validationReportPath(Path root) {
...
+        String json = resourceAsString("incorrect-endpoint-tests.json");
         return Jackson.load(EndpointTestSuiteModel.class, json);
     }
 
+    private ServiceModel getMissingRequestUriServiceModel() throws IOException {
+        String json = resourceAsString("no-request-uri-operation-service.json");
+        return Jackson.load(ServiceModel.class, json);
+    }
+
+    private String resourceAsString(String name) throws IOException {
```
RanVaknin, line 228: why did we need this refactor? is it just so we could use try-with-resources?

codegen-maven-plugin/src/main/java/software/amazon/awssdk/codegen/maven/plugin/GenerationMojo.java
```diff
@@ -216,6 +234,17 @@ private <T> Optional<T> loadOptionalModel(Class<T> clzz, Path location) {
         return ModelLoaderUtils.loadOptionalModel(clzz, location.toFile());
     }
 
+    private void emitValidationReport(ModelValidationReport report) {
+        Path modelsDir = sourcesDirectory.resolve("models");
+        try {
+            Writer writer = Files.newBufferedWriter(modelsDir.resolve("validation-report.json"),
```
RanVaknin, line 240: why do we need to write the validation report to a file? Are you not doing something similar with the optional exception in `AddShapes`?

dagnir: That exception is caught here in the mojo.

codegen/src/test/java/software/amazon/awssdk/codegen/CodeGeneratorTest.java
dagnir: No, it's so we don't repeat the same code here and in `getBrokenEndpointTestSuiteModel`

PR #6186, opened 2025-06-16
Fix a bug in ConstructorCache when classes are GC'ed but not removed from cache

alextwoods:
Fix a bug in ConstructorCache when classes are GC'ed but not removed from cache

## Motivation and Context
The ConstructorCache uses WeakReferences to hold Class objects, which can be garbage-collected, leading to Optional.empty() results even though the class may still be load-able.  The ConstructorCache is using a [WeakHashMap](https://docs.oracle.com/javase/8/docs/api/java/util/WeakHashMap.html) which handles GC of the keys only, but not when values (the class Objects in this case) are cleaned up.  

The current design of the constructor cache was introduced to store the classloader and class with a weak reference, so that we don't prevent them from being unloaded.  It was intended to follow We do something similar in DynamoDB enhanced's [BeanTableSchema](https://github.com/aws/aws-sdk-java-v2/blob/master/services-custom/dynamodb-enhanced/src/main/java/software/amazon/awssdk/enhanced/dynamodb/mapper/BeanTableSchema.java) but, unlike the BeanTableSchema, this cache *also* uses weakreferences to the values being cached.

## Modifications
Adds an additional check in the `getClass` method to check if the referent in the WeakReference has been garbage collected and if so, remove it from the cache.

**Q: Why not use a `ReferenceQueue` to handle removals as in the implementation of WeakHashMap?**
Short answer: it increases complexity more than is justified to handle an edge case error.  Long answer: The schema of the cache (specially `Map<String, Map<ClassLoader, Optional<WeakReference<Class<?>>>>>`) makes this more difficult to track - we would need to create a new entry class that extends WeakReference and stores the className and classLoader and would need to ensure the reference to classLoader is also a weak reference in this entry.  We would then need to add synchronization around check the reference queue and logic to remove cache entries from the nested cache correctly. 

## Testing
Testing this edge case is extremely difficult - I was able to manually reproduce the issue with a hacked custom class loader + multiple threads, but its not consistently reproducible, so I have not added an automated unit test to cover this case.

## Types of changes

## Checklist

## License

### DIFF
--- a/core/checksums/src/main/java/software/amazon/awssdk/checksums/internal/ConstructorCache.java
+++ b/core/checksums/src/main/java/software/amazon/awssdk/checksums/internal/ConstructorCache.java
@@ -63,12 +63,17 @@ public final class ConstructorCache {
                 return Optional.empty();
             }
         });
+
         // if the WeakReference to the class has been garbage collected, remove it from the cache and try again
-        if (classRef.isPresent() && classRef.get().get() == null) {
+        if (classRef.isPresent()) {
+            Class<?> clazz = classRef.get().get();
+            if (clazz != null) {
+                return Optional.of(clazz);
+            }
             classesByClassLoader.remove(classLoader);
             return getClass(className);
         }
-        return classRef.map(WeakReference::get);
+        return Optional.empty();
     }
 
     /**
--- a/core/checksums/src/main/java/software/amazon/awssdk/checksums/internal/ConstructorCache.java
+++ b/core/checksums/src/main/java/software/amazon/awssdk/checksums/internal/ConstructorCache.java
@@ -63,6 +63,11 @@ public final class ConstructorCache {
                 return Optional.empty();
             }
         });
+        // if the WeakReference to the class has been garbage collected, remove it from the cache and try again
+        if (classRef.isPresent() && classRef.get().get() == null) {
+            classesByClassLoader.remove(classLoader);
+            return getClass(className);
+        }
         return classRef.map(WeakReference::get);
     }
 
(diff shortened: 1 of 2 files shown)


core/checksums/src/main/java/software/amazon/awssdk/checksums/internal/ConstructorCache.java
```diff
@@ -63,6 +63,11 @@ private Optional<Class<?>> getClass(String className) {
                 return Optional.empty();
             }
         });
+        // if the WeakReference to the class has been garbage collected, remove it from the cache and try again
+        if (classRef.isPresent() && classRef.get().get() == null) {
+            classesByClassLoader.remove(classLoader);
+            return getClass(className);
+        }
         return classRef.map(WeakReference::get);
```
dagnir: Is it still possible for the weak ref to get GC'd between 67 and 71? Would it be better to store `classRef.get().get()` to a temporary var to avoid that happening?

alextwoods: Ooff - yeah, good catch - I've updated.

On a related note - I had been focused on checking that this was thread safe - and because of the usage of synchronized maps, I believe it is - in particular the `classesByClassLoader.remove(classLoader)` and `classesByClassLoader.computeIfAbsent` generally ensure the cache is consistent.  There is an issue where a class could be removed and re-loaded multiple times which would result in extra work, but not in an inconsistent/incorrect cache state.

dagnir: Cool, yeah agree don't think there's an issue where the cache could get corrupted in that case

dagnir (approved):

PR #6190, opened 2025-06-17
Make apache5.x as default client if in class path

joviegas:
## Motivation and Context
Make apache5.x as default client if in class path
### This PR is just to test the CI builds


joviegas:
testing Apache5.x as default

PR #6191, opened 2025-06-17
Disable Client based retries and define httpcore5 httpclient5 in .brazil.json

joviegas:
## Motivation and Context

1. Codegen-generated-tst were failing , the test were failing for retries where SDK retries were tested. 
2. The build check was failing for .brazil.json

## Modifications

- Disable retries on http.

## Testing
- Junit added

## License

### DIFF
--- a/test/http-client-tests/src/main/java/software/amazon/awssdk/http/SdkAsyncHttpClientH1TestSuite.java
+++ b/test/http-client-tests/src/main/java/software/amazon/awssdk/http/SdkAsyncHttpClientH1TestSuite.java
@@ -172,6 +172,26 @@ public abstract class SdkAsyncHttpClientH1TestSuite {
 
     }
 
+    @Test
+    public void doesNotRetryOn429StatusCode() throws InterruptedException {
+        server.return429OnFirstRequest = true;
+        server.closeConnection = false;
+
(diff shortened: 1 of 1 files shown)


test/http-client-tests/src/main/java/software/amazon/awssdk/http/SdkHttpClientTestSuite.java
```diff
@@ -218,6 +221,48 @@ public void testCustomTlsTrustManagerAndTrustAllFails() throws Exception {
         assertThatThrownBy(() -> createSdkHttpClient(httpClientOptions)).isInstanceOf(IllegalArgumentException.class);
     }
 
+    @Test
```
joviegas, line 224: Adding test cases in SdkHttpClientTestSuite makes sure that all the other SDK clients gets testes since this class is extended by other HTTP clients too.

zoewangg: Can we add it in async test suite as well?

joviegas: done

PR #6196, opened 2025-06-19
Merge feature/master/codegen-validation to master

dagnir:
## Motivation and Context
This PR merges the `feature/master/codegen-validation` feature branch to `master`.

This feature branch contains the following updates:

 - Add support for writing a `validation-report.json` during the build if any validation errors are found during code generation.
 - Add support for defining validators that run during codegen.
 - Add support for the following validations
   - Validate that models shared between two services are completely identical
   - Validate that shape members referenced by endpoint tests exist
   - Validate that request URI exists when an input member is mapped to the URI

## Types of changes

## Checklist

## License


codegen/src/main/java/software/amazon/awssdk/codegen/emitters/tasks/BaseGeneratorTasks.java
```diff
@@ -71,6 +71,8 @@ protected void compute() {
                 ForkJoinTask.invokeAll(createTasks());
                 log.info("  Completed " + taskName + ".");
             }
+        } catch (RuntimeException e) {
+            throw e;
```
millems, line 75: Nit: Not needed

millems (approved):
This was already reviewed, so I only reviewed for obvious issues.

codegen/src/main/java/software/amazon/awssdk/codegen/emitters/tasks/BaseGeneratorTasks.java
dagnir: I added this so that exceptions that are already `RuntimeException`s that bubble up here don't get wrapped further

codegen-maven-plugin/src/main/java/software/amazon/awssdk/codegen/maven/plugin/GenerationMojo.java
```diff
@@ -169,6 +234,17 @@ private <T> Optional<T> loadOptionalModel(Class<T> clzz, Path location) {
         return ModelLoaderUtils.loadOptionalModel(clzz, location.toFile());
     }
 
+    private void emitValidationReport(ModelValidationReport report) {
+        Path modelsDir = sourcesDirectory.resolve("models");
+        try {
```
RanVaknin: Should we use try-with-resources?

codegen/src/main/java/software/amazon/awssdk/codegen/AddShapes.java
```diff
@@ -346,11 +352,20 @@ private boolean isGreedy(Shape parentShape, Map<String, Shape> allC2jShapes, Par
...
+        Optional<Operation> operation = builder.getService().getOperations().values().stream()
+                                               .filter(o -> o.getInput() != null)
+                                               .filter(o -> allC2jShapes.get(o.getInput().getShape()).equals(parentShape))
+                                               .findFirst();
+
+        return operation.map(o -> o.getHttp().getRequestUri())
+                        .orElseThrow(() -> {
+                            String detailMsg = "Could not find request URI for input shape";
+                            ValidationEntry entry =
+                                new ValidationEntry().withErrorId(ValidationErrorId.REQUEST_URI_NOT_FOUND)
```
RanVaknin, line 364: nit: we should use static factory method names
`.errorId()` rather than `.withErrorId()` etc

RanVaknin (changes requested):
Just trying to get a good understanding about this feature as whole. 

If I understand correctly, we need this validation for services that share a model, (like dynamodb and streams) through the `customization.config`using the `shareModelWith` field, to make sure that the models are identical.

However, I don't see that there is a separate model for streams. 

For example: 
```
.
└── aws-sdk-java-v2/
    └── services/
        ├── dynamodb/
        │   └── services-2.json
        └── streams(doesnt exist)/
```
How are those validated against each other if they share the model? where is the extra model that is being validated against?

---

That aside, if my read is correct on the validation itself, it checks if both models have the same members (no extra members or different named members) and then if configured throws a validation error and potentially emits a report, with the option to provide your own validator. 

Does that sound right?

codegen/src/main/java/software/amazon/awssdk/codegen/AddShapes.java
dagnir: Model classes in `codegen` tend to use `with` so sticking with the convention here

dagnir:
> Just trying to get a good understanding about this feature as whole.
> 
> If I understand correctly, we need this validation for services that share a model, (like dynamodb and streams) through the `customization.config`using the `shareModelWith` field, to make sure that the models are identical.
> 
> However, I don't see that there is a separate model for streams.
> 
> For example:
> 
> ```
> .
> └── aws-sdk-java-v2/
>     └── services/
>         ├── dynamodb/
>         │   └── services-2.json
>         └── streams(doesnt exist)/
> ```
> 
> How are those validated against each other if they share the model? where is the extra model that is being validated against?

There is not a separate module for `streams`, both models are in the `dynamodb` module: https://github.com/aws/aws-sdk-java-v2/tree/master/services/dynamodb/src/main/resources/codegen-resources

> 
> That aside, if my read is correct on the validation itself, it checks if both models have the same members (no extra members or different named members) and then if configured throws a validation error and potentially emits a report, with the option to provide your own validator.
> 
> Does that sound right?

Yes, that right.

codegen-maven-plugin/src/main/java/software/amazon/awssdk/codegen/maven/plugin/GenerationMojo.java
dagnir: Good catch, done.

RanVaknin (approved):

PR #6197, opened 2025-06-19
Add service version info

RanVaknin:
This PR adds a code generated constant value to very service's package under `ServiceVersionInfo.VERSION`.


codegen/src/main/java/software/amazon/awssdk/codegen/poet/client/specs/VersionCompatibilityTestSpec.java
RanVaknin, line 1: I think generating a fixture test for a code generated test might be overkill, but for better readability here is what this generated test would look like for SQS:

```java
/*
 * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 * 
 * Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance with
 * the License. A copy of the License is located at
 * 
 * http://aws.amazon.com/apache2.0
 * 
 * or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
 * CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions
 * and limitations under the License.
 */

package software.amazon.awssdk.services.sqs;

import org.assertj.core.api.Assertions;
import software.amazon.awssdk.annotations.Generated;
import software.amazon.awssdk.core.util.VersionInfo;
import software.amazon.awssdk.services.sqs.internal.ServiceVersionInfo;

public class VersionCompatibilityTest {
    @Test
    public void checkCompatibility() {
        String coreVersion = VersionInfo.SDK_VERSION;
        String serviceVersion = ServiceVersionInfo.VERSION;
        Assertions
                .assertThat(isVersionCompatible(coreVersion, serviceVersion))
                .withFailMessage("Core version %s must be equal to or newer than service version %s", coreVersion, serviceVersion)
                .isTrue();
    }

    private boolean isVersionCompatible(String coreVersion, String serviceVersion) {
        String normalizedCore = coreVersion.replace("-SNAPSHOT", "");
        String normalizedService = serviceVersion.replace("-SNAPSHOT", "");
        String[] coreParts = normalizedCore.split("\\.");
        String[] serviceParts = normalizedService.split("\\.");

        int coreMajor = Integer.parseInt(coreParts[0]);
        int serviceMajor = Integer.parseInt(serviceParts[0]);
        if (coreMajor != serviceMajor) {
            return coreMajor >= serviceMajor;
        }

        int coreMinor = Integer.parseInt(coreParts[1]);
        int serviceMinor = Integer.parseInt(serviceParts[1]);
        if (coreMinor != serviceMinor) {
            return coreMinor >= serviceMinor;
        }

        int corePatch = Integer.parseInt(coreParts[2]);
        int servicePatch = Integer.parseInt(serviceParts[2]);
        return corePatch >= servicePatch;
    }
}
```

codegen/src/main/java/software/amazon/awssdk/codegen/poet/client/specs/VersionCompatibilityTestSpec.java
```diff
@@ -0,0 +1,99 @@
...
+                         .addModifiers(Modifier.PRIVATE)
+                         .returns(boolean.class)
+                         .addParameter(String.class, "coreVersion")
+                         .addParameter(String.class, "serviceVersion")
+                         .addStatement("String normalizedCore = coreVersion.replace(\"-SNAPSHOT\", \"\")")
+                         .addStatement("String normalizedService = serviceVersion.replace(\"-SNAPSHOT\", \"\")")
+                         .addStatement("String[] coreParts = normalizedCore.split(\"\\\\.\")")
+                         .addStatement("String[] serviceParts = normalizedService.split(\"\\\\.\")")
+                         .addCode("\n")
+                         .addStatement("int coreMajor = Integer.parseInt(coreParts[0])")
```
alextwoods: Are there ever cases we'd end up with non integer version parts?  I know in other SDKs we occasionally have like pre-release versions such as "2.0.0.pre".   Or are there ever times we would have less than 3 parts?

codegen/src/test/java/software/amazon/awssdk/codegen/poet/client/ServiceVersionInfoSpecTest.java
```diff
@@ -0,0 +1,63 @@
...
+        String[] parts = expectedContent.split("public static final String VERSION = \"");
+        if (parts.length == 2) {
+            String privateConstructor = parts[1].substring(parts[1].indexOf("\""));
+            expectedContent = parts[0] + "public static final String VERSION = \"" + currVersion
+                              + privateConstructor;
+        }
+
+        String actualContent = generateContent(serviceVersionInfoSpec);
+
+        assertThat(actualContent).isEqualTo(expectedContent);
```
alextwoods: Should we be using the `equalToIgnoringWhiteSpace` as we do in the poet matcher?

codegen/src/test/java/software/amazon/awssdk/codegen/poet/client/ServiceVersionInfoSpecTest.java
```diff
@@ -0,0 +1,63 @@
...
+    @Test
+    void testServiceVersionInfoClass() {
+        String currVersion = VersionInfo.SDK_VERSION;
+        ClassSpec serviceVersionInfoSpec = new ServiceVersionInfoSpec(ClientTestModels.restJsonServiceModels());
+
+        String expectedContent = loadFixtureFile("test-service-version-info-class.java");
+        String[] parts = expectedContent.split("public static final String VERSION = \"");
+        if (parts.length == 2) {
+            String privateConstructor = parts[1].substring(parts[1].indexOf("\""));
+            expectedContent = parts[0] + "public static final String VERSION = \"" + currVersion
```
alextwoods: I might be missing what this code is doing, but since you have effectively a template string (`{{VERSION}}`) in the test file, could you do something like `exepctedContent.replace("{{VERSION}}", currVersion)`

.changes/next-release/feature-AWSSDKforJavav2-382b343.json
```diff
@@ -0,0 +1,6 @@
+{
+    "type": "feature",
+    "category": "AWS SDK for Java v2",
+    "contributor": "",
+    "description": "Adding compatibility tests between sdk-core and service packages"
```
alextwoods: Nit: update to just adding the version info.

codegen/src/main/java/software/amazon/awssdk/codegen/poet/client/specs/VersionCompatibilityTestSpec.java
alextwoods: EDIT: Saw this was removed, ignore :-)

codegen/src/test/java/software/amazon/awssdk/codegen/poet/client/ServiceVersionInfoSpecTest.java
RanVaknin: nice. good call out.

codegen/src/test/java/software/amazon/awssdk/codegen/poet/client/ServiceVersionInfoSpecTest.java
```diff
@@ -0,0 +1,59 @@
...
+import software.amazon.awssdk.codegen.poet.ClassSpec;
+import software.amazon.awssdk.codegen.poet.ClientTestModels;
+import software.amazon.awssdk.codegen.poet.client.specs.ServiceVersionInfoSpec;
+import software.amazon.awssdk.core.util.VersionInfo;
+
+public class ServiceVersionInfoSpecTest {
+
+    // a fixture test that dynamically updates the generated fixture with the current version
+    // this is needed because every time codegen runs, the version will change.
+    // we need a way to generate the fixture, and then edit it in place with the current version and only then make the assertion.
```
alextwoods: Maybe add a comment about the {{VERSION}} template - if the test file is re-generated then the version in the file must be replaced by that template string. (thinking through what happens when I see this test case fail down the road and am trying to update it)

alextwoods (approved):

PR #6208, opened 2025-06-25
Add ec2InstanceProfileName configuration

S-Saranya1:
Adding ec2InstanceProfileName configuration to specify IMDS instance profile.

## Motivation and Context
Adding ec2InstanceProfileName configuration to specify IMDS instance profile for retrieving credentials. This improves performance by skipping the profile discovery step, eliminating one IMDS call.

## Modifications
- Added the ec2InstanceProfileName config in provider configuration, system properties, environment variables and in configuration files.
- Modified Ec2MetadataConfigProvider to parse ec2InstanceProfileName configuration.
- Modified the credentials fetching logic in InstanceProfileCredentialsProvider to include this new configuration.
- Added unit tests to test the new config. 

## Testing
- Added unit tests for all supported configuration methods.
- Added tests in the ExtendedApiTest file to verify the new profile name works with the Extended API.
- Added few test cases from the IMDS SEP which tests with ec2InstanceProfileName config.
- Performed one-off integration testing.

## Types of changes

## Checklist

## License

### DIFF
--- a/core/auth/src/main/java/software/amazon/awssdk/auth/credentials/InstanceProfileCredentialsProvider.java
+++ b/core/auth/src/main/java/software/amazon/awssdk/auth/credentials/InstanceProfileCredentialsProvider.java
@@ -205,7 +205,7 @@ public final class InstanceProfileCredentialsProvider
                 if (apiVersion == ApiVersion.UNKNOWN) {
                     apiVersion = ApiVersion.LEGACY;
                     return refreshCredentials();
-                } else if (ec2InstanceProfileName == null && configProvider.ec2InstanceProfileName() == null) {
+                } else if (resolveProfileName() == null) {
                     // Resolved profile name is invalid, reset it and try again
                     resolvedProfile = null;
                     
@@ -348,14 +348,16 @@ public final class InstanceProfileCredentialsProvider
         return configProvider.isMetadataV1Disabled();
     }
 
-    private String[] getSecurityCredentials(String imdsHostname, String metadataToken) {
-        if (ec2InstanceProfileName != null) {
-            return new String[]{ec2InstanceProfileName};
-        }
+    private String resolveProfileName() {
+        return ec2InstanceProfileName != null ? 
+               ec2InstanceProfileName : 
+               configProvider.ec2InstanceProfileName();
+    }
 
-        String configuredProfileName = this.configProvider.ec2InstanceProfileName();
-        if (configuredProfileName != null) {
-            return new String[]{configuredProfileName};
+    private String[] getSecurityCredentials(String imdsHostname, String metadataToken) {
+        String profileName = resolveProfileName();
+        if (profileName != null) {
+            return new String[]{profileName};
         }
 
         if (resolvedProfile != null) {
@@ -417,12 +419,9 @@ public final class InstanceProfileCredentialsProvider
          * Configure the EC2 instance profile name to use for retrieving credentials.
          * 
          * <p>When this is set, the provider will skip fetching the list of available instance profiles
-         * and use this name directly. This can improve performance by reducing the number of calls to IMDS.
-         * 
-         * <p>By default, this is not set and the provider will discover the instance profile name from IMDS.
+         * and use this name directly.
          * 
          * @param ec2InstanceProfileName The EC2 instance profile name to use
-         * @return This builder for method chaining
          */
         Builder ec2InstanceProfileName(String ec2InstanceProfileName);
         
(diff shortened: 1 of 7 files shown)


core/auth/src/main/java/software/amazon/awssdk/auth/credentials/InstanceProfileCredentialsProvider.java
```diff
@@ -118,6 +121,13 @@ private InstanceProfileCredentialsProvider(BuilderImpl builder) {
                                    .orElseGet(() -> ProfileFileSupplier.fixedProfileFile(ProfileFile.defaultProfileFile()));
         this.profileName = Optional.ofNullable(builder.profileName)
                                    .orElseGet(ProfileFileSystemSetting.AWS_PROFILE::getStringValueOrThrow);
+        this.ec2InstanceProfileName = builder.ec2InstanceProfileName;
+        
+        if (isBlank(ec2InstanceProfileName) && ec2InstanceProfileName != null) {
```
millems, line 126: This can be simplified to `StringUtils.isWhitespace(ec2InstanceProfileName)`

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/InstanceProfileCredentialsProvider.java
```diff
@@ -173,6 +183,11 @@ private RefreshResult<AwsCredentials> refreshCredentials() {
 
         try {
             LoadedCredentials credentials = httpCredentialsLoader.loadCredentials(createEndpointProvider());
+
+            if (apiVersion == ApiVersion.UNKNOWN) {
+                apiVersion = ApiVersion.EXTENDED;
+            }
```
millems, line 189: What is the purpose of this change?

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/InstanceProfileCredentialsProvider.java
```diff
@@ -186,19 +201,25 @@ private RefreshResult<AwsCredentials> refreshCredentials() {
...
+                    profileRetryCount++;
+                    if (profileRetryCount <= MAX_PROFILE_RETRIES) {
+                        log.debug(() -> "Profile name not found, retrying fetching the profile name again.");
+                        return refreshCredentials();
+                    }
+                } else {
+                    throw SdkClientException.builder()
+                                           .message("Invalid profile name")
+                                           .cause(e)
+                                           .build();
```
millems: Can we expand on this error message a bit? Specifically, what's the invalid value and what should they do about it?

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/Ec2MetadataConfigProvider.java
```diff
@@ -171,6 +189,22 @@ private static Optional<Long> fromProfileFileServiceTimeout(Supplier<ProfileFile
...
+    
+    // Profile file resolution for EC2 Instance Profile Name
+    private static Optional<String> fromProfileFileEc2InstanceProfileName(Supplier<ProfileFile> profileFile, String profileName) {
+        try {
+            return profileFile.get()
+                              .profile(profileName)
+                              .flatMap(p -> p.property(ProfileProperty.EC2_INSTANCE_PROFILE_NAME));
+        } catch (Exception e) {
+            return Optional.empty();
+        }
```
millems, line 206: What case is this guarding against?

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/InstanceProfileCredentialsProvider.java
S-Saranya1: Updated it.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/internal/Ec2MetadataConfigProvider.java
S-Saranya1: This is guarding against exceptions from aggregated profile suppliers, particularly NoSuchElementException like when there are invalid profiles or when suppliers throw exceptions during normal operation.

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/InstanceProfileCredentialsProvider.java
S-Saranya1: Initially, apiVersion is set to UNKNOWN. When ec2InstanceProfileName is set by the user, we skip the first GET request to IMDS in the getSecurityCredentials() method. In refreshCredentials(), the code attempts to load credentials using the extended API endpoint first. If successful, we update apiVersion to EXTENDED. This serves two purposes - It records that the extended API is supported for future reference so it prevents unnecessary attempts to use the legacy API in subsequent calls, as we now know the extended API works.(This is from IMDS SEP 2.1)

core/auth/src/main/java/software/amazon/awssdk/auth/credentials/InstanceProfileCredentialsProvider.java
S-Saranya1: Hmm, the StringUtils(software.amazon.awssdk.utils.StringUtils) doesn't have "isWhitespace" method, it only has "isBlank" method.
Apache Commons Lang StringUtils (from org.apache.commons.lang3.StringUtils) does have an isWhitespace method, but it's not available for this module.

PR #6212, opened 2025-06-25
Add useragent for service and version

RanVaknin:
## Motivation and Context
Implements User Agent 2.1 SEP by adding service-specific API metadata to user agent strings.

Example: 
```java
// user agent now includes the service identifier and the version with which the request was sent from:
"aws-sdk-java/2.26.22-SNAPSHOT ua/2.1 api/DynamoDB#2.26.22-SNAPSHOT os/Mac_OS_X#14.6.1 lang/java#21.0.2 md/OpenJDK_64-Bit_Server_VM#21.0.2+13-LTS md/en_US"
```

## Modifications
- Added a new constant `USER_AGENT` in `ServiceVersionInfo` to carry the service id + version.
- Modified client constructors to include API metadata in user agent
- Modified `SdkUserAgentBuilder` to include API metadata in correct position
- Added unit tests for user agent format validation, and order validation (following the SEP)

## Testing
- Added fixture tests for useragent constant code generation
- Added user agent builder tests with API metadata scenarios


.changes/next-release/feature-AWSSDKforJavav2-e1e404e.json
```diff
@@ -0,0 +1,6 @@
+{
+    "type": "feature",
+    "category": "AWS SDK for Java v2",
+    "contributor": "",
+    "description": "Adding the version of the used service to the useragent"
```
alextwoods: Nit on wording - this could imply the version of the service api (eg the latest version of ec2 api right now is `2016-11-15`).  Maybe "Add the sdk service client version to the useragent"

codegen/src/main/java/software/amazon/awssdk/codegen/poet/client/specs/ServiceVersionUserAgentSpec.java
```diff
@@ -0,0 +1,74 @@
...
+    public ServiceVersionUserAgentSpec(IntermediateModel model) {
+        this.poetExtension = new PoetExtension(model);
+        this.model = model;
+    }
+
+    @Override
+    public TypeSpec poetSpec() {
+        TypeSpec.Builder builder = TypeSpec.classBuilder("ServiceVersionUserAgent")
+                                           .addModifiers(Modifier.PUBLIC, Modifier.FINAL)
+                                           .addAnnotation(PoetUtils.generatedAnnotation())
```
alextwoods: Does it make sense to apply @SdkInternalApi to this?

codegen/src/main/java/software/amazon/awssdk/codegen/poet/client/specs/ServiceVersionUserAgentSpec.java
```diff
@@ -0,0 +1,74 @@
...
+    private final IntermediateModel model;
+
+    public ServiceVersionUserAgentSpec(IntermediateModel model) {
+        this.poetExtension = new PoetExtension(model);
+        this.model = model;
+    }
+
+    @Override
+    public TypeSpec poetSpec() {
+        TypeSpec.Builder builder = TypeSpec.classBuilder("ServiceVersionUserAgent")
```
alextwoods: I'm wondering if we should call this something like `ServiceClientVersionUserAgentMetadata` - super long, but I don't want to conflate the service api version with the client version

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/client/specs/test-service-version-user-agent-class.java
```diff
@@ -0,0 +1,15 @@
+package software.amazon.awssdk.services.json.internal;
+
+import java.lang.String;
+import software.amazon.awssdk.annotations.Generated;
+
+@Generated("software.amazon.awssdk:codegen")
+public final class ServiceVersionUserAgent {
```
zoewangg: Is there any reason we need a class for it? Can we just create a string constant in the client impl class?

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/client/specs/test-service-version-user-agent-class.java
```diff
@@ -0,0 +1,15 @@
...
+  /**
+   * Returns a user agent containing the service and version info
+   */
+  public static final String USER_AGENT = "Json_Service#{{VERSION}}";
```
zoewangg: IIRC, the key should be the service ID

zoewangg (changes requested):

codegen/src/main/java/software/amazon/awssdk/codegen/poet/client/specs/ServiceVersionUserAgentSpec.java
RanVaknin: It's already generated under the `/internal` folder, but I guess I could add that?

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/client/specs/test-service-version-user-agent-class.java
RanVaknin: By key do you mean the variable name `USER_AGENT` should be the name of the service? Or are you referring to the service identifier before the #? 

The `Json_Service` part is the fixture file equivalent of the service id:
codegen/src/main/java/software/amazon/awssdk/codegen/poet/client/specs/ServiceVersionUserAgentSpec.java

This will result in 

public static final String USER_AGENT = "Dynamodb#2.26.1"
Later on we append the `api` prefix so the useragent is: `api/Dynamodb#2.26.1`

zoewangg: Oh is `Json_Service` *the* service ID of this fake service? I someone expected it to be shorter, nvm then!

codegen/src/test/resources/software/amazon/awssdk/codegen/poet/client/specs/test-service-version-user-agent-class.java
RanVaknin: Since that `USER_AGENT` constant will get generated with a new version with every release, it will be very difficult to re-write all the 30+ fixture files that pertain to client and asyncClient to programmatically inject that version at build time. 

If we have a single class under the internal path, it only has one fixture file that has a {{VERSION}} placeholder, we can inject that code generated version at build time:
See ServiceVersionUserAgentSpecTest.java

zoewangg: I see, can we pass a supplier then, that way we don't need to worry about the version 
`.option(SdkClientOption.API_METADATA_SUPPLIER, () -> apiMetadata()).build();`

RanVaknin: I'm not sure I understand how the supplier gets around the issue? 

Wherever that constant is defined, we will need to be able to dynamically inject the current version at build time.
If that constant is defined in a single fixture file (`test-service-version-user-agent-class.java`), then we can have only one fixture test file that can handle the version injection (`ServiceVersionUserAgentSpecTest.java`).

If we add that constant into the client builder spec classes (async and sync), then we will need to update the 30+ fixture files for those clients, and also add logic to their test suites to be able to dynamically inject that version at build time, the same way `ServiceVersionUserAgentSpecTest` does that. 

Is the concern that we are generating an extra class just for this one constant? For reference the `VersionInfo.java` file we created in sdk-core is about 400 bytes. Assuming we have 500 aws services, this adds +- 200kilobytes to the overall size of the SDK.

An alternative I can think of is that we can add this new constant into ServiceVersionInfo.java file which already has similar logic?

zoewangg: Actually, it doesn't have to be `Supplier`,  can we just do 

`
.option(SdkClientOption.API_METADATA, formatApiMetadata(ServiceVersionInfo.VERSION)
`

alextwoods (approved):
I'm out of office the rest of the day - approving so I dont block this - pending resolution of: 
* https://github.com/aws/aws-sdk-java-v2/pull/6212/files#r2169894743
* https://github.com/aws/aws-sdk-java-v2/pull/6212#discussion_r2169970100

codegen/src/main/java/software/amazon/awssdk/codegen/poet/client/SyncClientClass.java
```diff
@@ -249,6 +255,11 @@ private Stream<MethodSpec> operations(OperationModel opModel) {
         return methods.stream();
     }
 
+    private String transformServiceId(String serviceId) {
+        // According to User Agent 2.0 spec, replace spaces with underscores
+        return serviceId.replace(" ", "_");
+    }
```
zoewangg: Can we share the code in both sync and async client class?

core/sdk-core/src/main/java/software/amazon/awssdk/core/client/builder/SdkDefaultClientBuilder.java
```diff
@@ -290,6 +292,7 @@ private SdkClientConfiguration mergeGlobalDefaults(SdkClientConfiguration config
                                                   .option(USER_AGENT_PREFIX, "")
                                                   .option(USER_AGENT_SUFFIX, "")
                                                   .option(CRC32_FROM_COMPRESSED_DATA_ENABLED, false)
+                                                  .option(API_METADATA, "")
```
zoewangg: Is there any reason we are setting an empty value as default?

core/sdk-core/src/test/java/software/amazon/awssdk/core/internal/useragent/SdkUserAgentBuilderTest.java
```diff
@@ -20,6 +20,7 @@
 import static software.amazon.awssdk.core.internal.useragent.UserAgentConstant.HTTP;
 import static software.amazon.awssdk.core.internal.useragent.UserAgentConstant.INTERNAL_METADATA_MARKER;
 import static software.amazon.awssdk.core.internal.useragent.UserAgentConstant.IO;
+import static software.amazon.awssdk.core.internal.useragent.UserAgentConstant.API_METADATA;
```
zoewangg, line 23: Can we add some functional tests (end-to-end tests with mock server) as well? Maybe in `BusinessMetricsUserAgentTest`?

codegen/src/main/java/software/amazon/awssdk/codegen/poet/client/AsyncClientClass.java
```diff
@@ -229,7 +229,14 @@ private MethodSpec constructor(TypeSpec.Builder classBuilder) {
                         .addStatement("this.clientHandler = new $T(clientConfiguration)", AwsAsyncClientHandler.class)
                         .addStatement("this.clientConfiguration = clientConfiguration.toBuilder()"
                                       + ".option($T.SDK_CLIENT, this)"
-                                      + ".build()", SdkClientOption.class);
+                                      + ".option($T.API_METADATA, $S + \"#\" + $T.VERSION)"
+                                      + ".build()",
+                                      SdkClientOption.class,
+                                      SdkClientOption.class,
```
zoewangg, line 236: Duplicate `SdkClientOption.class`

core/sdk-core/src/main/java/software/amazon/awssdk/core/client/builder/SdkDefaultClientBuilder.java
RanVaknin: In this case `API_METADATA` gets overridden in the client constructor with the actual service name and version. The SEP doesn't call a default behavior, therefore using an empty value seems reasonable and follows the other option conventions.

codegen/src/main/java/software/amazon/awssdk/codegen/poet/client/AsyncClientClass.java
RanVaknin: Why is this a concern? 

Generated code seems reasonable:

```java
    protected DefaultProtocolRestJsonWithConfigAsyncClient(SdkClientConfiguration clientConfiguration) {
        this.clientHandler = new AwsAsyncClientHandler(clientConfiguration);
        this.clientConfiguration = clientConfiguration.toBuilder().option(SdkClientOption.SDK_CLIENT, this)
                .option(SdkClientOption.API_METADATA, "AmazonProtocolRestJsonWithConfig" + "#" + ServiceVersionInfo.VERSION)
                .build();
        this.protocolFactory = init(AwsJsonProtocolFactory.builder()).build();
        this.executor = clientConfiguration.option(SdkAdvancedAsyncClientOption.FUTURE_COMPLETION_EXECUTOR);
    }
```

core/sdk-core/src/main/java/software/amazon/awssdk/core/client/builder/SdkDefaultClientBuilder.java
zoewangg: Hmm, for cases where users use old service that don't override service name and version, it'd be empty, which seems a bit odd. IMO, null and empty are semantically different and I'm leaning towards setting it to null to be more explicit about when the feature is not configured.

codegen/src/main/java/software/amazon/awssdk/codegen/poet/client/AsyncClientClass.java
zoewangg: Oh I missed `.option(SdkClientOption.SDK_CLIENT, this)` and thought we only needed one. nvm then!

codegen/src/main/java/software/amazon/awssdk/codegen/poet/client/SyncClientClass.java
RanVaknin: done

core/sdk-core/src/main/java/software/amazon/awssdk/core/client/builder/SdkDefaultClientBuilder.java
RanVaknin: done

zoewangg (approved):

PR #6228, opened 2025-07-02
URI bounded cache for DynamoDB account id based endpoint

davidh44:
## Motivation and Context

Continuation of https://github.com/aws/aws-sdk-java-v2/pull/6087 
- Only update is replacing `LruCache` with a random-eviction `BoundedCache` in `SdkUri`
  - https://github.com/aws/aws-sdk-java-v2/commit/bc51b1a459bd30ab80782e7c7caed3a54318e0b3

## Modifications

- Added a SdkURI class which caches calls to the various URI constructors methods.
- Added customization configuration to codegen to enable usage of SdkURI class in generated endpoint resolution code.
- Enabled this config only for DynamoDB.

## Testing

- Unit test
- Canary testing in progress

## Types of changes


utils/src/main/java/software/amazon/awssdk/utils/uri/SdkUri.java
```diff
@@ -153,15 +145,6 @@ private boolean isAccountIdUri(String s) {
         return false;
     }
 
-    private void logCacheUsage(boolean containsKey, URI uri) {
```
RanVaknin: Was logging only useful for LRU? Can we not benefit from seeing the usage of the BoundedCache as well?

utils/src/main/java/software/amazon/awssdk/utils/cache/bounded/BoundedCache.java
```diff
@@ -0,0 +1,126 @@
...
+        this.cache = new ConcurrentHashMap<>();
+        this.cacheLock = new Object();
+    }
+
+    /**
+     * Get a value based on the key. If the value exists in the cache, it's returned.
+     * Otherwise, the value is calculated based on the supplied function {@link Builder#builder(Function)}.
+     */
+    public V get(K key) {
+        V value = cache.get(key);
```
RanVaknin, line 64: can `key` ever be null? should we add a null check avoid cache.get(null)?

utils/src/main/java/software/amazon/awssdk/utils/cache/bounded/BoundedCache.java
```diff
@@ -0,0 +1,126 @@
...
+
+            cache.put(key, newValue);
+            return newValue;
+        }
+    }
+
+    /**
+     * Clean up the cache by removing an unspecified entry
+     */
+    private void cleanup() {
```
RanVaknin, line 92: This cleanup implementation removes a single entry every time the cache is at capacity. That means it becomes less efficient when the cache becomes full because every new entry will follow with a cleanup() call.

Consider this:

```
- cache.put(key) (capacity 1)
- cache.put(key) (capacity 2)
- cache.put(key) (capacity 3)
...
...
...
- cache.put(key) (capacity 99)
- cache.put(key) (capacity 100)
- cache is at max capacity
- cleanup() removes 1
<cycle>
- cache.put(key) (capacity 99)
- cache.put(key) (capacity 100)
- cache is at max capacity
- cleanup() removes 1
- <cycle>
- cache.put(key) (capacity 99)
- cache.put(key) (capacity 100)
- cache is at max capacity
- cleanup() removes 1
```

Did we consider an alternative cleanup strategy?

RanVaknin (changes requested):
Just a few questions

utils/src/main/java/software/amazon/awssdk/utils/uri/SdkUri.java
davidh44: Removed previously since BoundedCache didn't have containsKey method, added method and added back logging

utils/src/main/java/software/amazon/awssdk/utils/cache/bounded/BoundedCache.java
davidh44: Currently, key will never be null when invoked internally by the SDK. Adding null check for future safeguard

utils/src/main/java/software/amazon/awssdk/utils/cache/bounded/BoundedCache.java
davidh44: Good callout, updated to batch evict 10 entries when full

PR #6236, opened 2025-07-04
Run s3 regression tests only when detecting changes related to s3

L-Applin:
## Modifications
- Added a new Job for change detection called check-s3-related-changes that runs on every PR. This job checks if any files in specific directories were modified. Pattern to detect changes in specific modules:
     • core
     • services/s3
     • services-custom/s3-transfer-manager
     • http-client-spi
     • http-clients

- Updated all S3 regression test jobs to depend on the new check job, and run only if changes are detected in the specified modules

### DIFF
--- a/.github/workflows/s3-regression-tests.yml
+++ b/.github/workflows/s3-regression-tests.yml
@@ -5,10 +5,6 @@ on:
   push:
     branches:
       - master
-    paths-ignore:
-      - '**.md'
-      - '.all-contributorsrc'
-      - 'docs/**'
 
 permissions:
   id-token: write


.github/workflows/s3-regression-tests.yml
```diff
@@ -0,0 +1,157 @@
...
+on:
+  pull_request:
+  merge_group:
+  push:
+    branches:
+      - master
+    paths-ignore:
+      - '**.md'
+      - '.all-contributorsrc'
+      - 'docs/**'
```
zoewangg: This is no longer needed, right?

zoewangg (approved):

Issue #5968, opened 2025-03-18
EC2 `DescribeInstances` query uses locale-dependent formatting


DaveCTurner:
### Describe the bug

When calling `Ec2Client#describeInstances` with one or more filter expressions, the filters contain numbers formatted using the current locale rather than the root locale. Thus for instance running in the `ne` (Nepalese) local the query uses `Filter.%E0%A5%A7.Name` rather than `Filter.1.Name`, where `%E0%A5%A7` is the UTF-8 encoding for `U+0967 DEVANAGARI DIGIT ONE` which is the rendering for the integer `1` in this locale. The full HTTP request in this case is as follows:

```
POST / HTTP/1.1
Host: 127.0.0.1:60608
amz-sdk-invocation-id: a5d80d84-b502-096b-bef6-da85818cb947
amz-sdk-request: attempt=1; max=10
Authorization: AWS4-HMAC-SHA256 Credential=ec2_key/20250318/es-test-region/ec2/aws4_request, SignedHeaders=amz-sdk-invocation-id;amz-sdk-request;content-length;content-type;host;x-amz-content-sha256;x-amz-date, Signature=749ddaeccbe94b749d6c1e7446a070dcca7ec2845068053883f28fdd9129ded3
Content-Type: application/x-www-form-urlencoded; charset=utf-8
User-Agent: aws-sdk-java/2.30.38 md/io#sync md/http#Apache ua/2.1 os/Mac_OS_X#15.3.2 lang/java#23 md/OpenJDK_64-Bit_Server_VM#23+37 md/vendor#Eclipse_Adoptium md/en_GB cfg/auth-source#stat m/D
x-amz-content-sha256: 1380541b646459c65b2316cb6c26f03a41760ebe57c48ecbe55697c5da1b3938
X-Amz-Date: 20250318T142438Z
Content-Length: 239
Connection: Keep-Alive

Action=DescribeInstances&Version=2016-11-15&Filter.%E0%A5%A7.Name=instance-state-name&Filter.%E0%A5%A7.Value.%E0%A5%A7=running&Filter.%E0%A5%A7.Value.%E0%A5%A8=pending&Filter.%E0%A5%A8.Name=tag%3Astage&Filter.%E0%A5%A8.Value.%E0%A5%A7=prod
```

EC2 rejects this with a `400 Bad request`:

```
[...]

PR #6238, opened 2025-07-07
Fix EC2 DescribeInstances query use locale-dependent formatting

Fred1155:
## Motivation and Context

User reported the issue https://github.com/aws/aws-sdk-java-v2/issues/5968. Currently we are render the filters using locale-dependent numerals, UTF-8-encoded. For example, for instance running in the ne (Nepalese) local the query uses ``Filter.%E0%A5%A7.Name`` rather than ``Filter.1.Name``, where ``%E0%A5%A7`` is the UTF-8 encoding for ``U+0967 DEVANAGARI DIGIT ONE`` which is the rendering for the integer 1 in this locale. The expected behavior is in all locales, we should render the filters using ASCII digits 0 to 9 only.  
## Modifications

The current code uses ``String.format("%s.%d", path, i + 1``) which is locale-dependent. When formatting the ``%d`` (decimal integer), Java uses the system's default locale to determine how to display numbers.  
``Locale.ROOT`` is Java's locale-neutral locale. By adding that to String.format, it will always produce ASCII digits regardless of system locale.  This issue only appear for Java version 9 and above. 
## Testing

Added unit test for ListQueryMarshaller class and also did a local integration test using the self-repro code provided in the original issue. 
## Types of changes

## Checklist

## License

### DIFF
--- a/.changes/next-release/bugfix-AWSQueryProtocol-1398d13.json
+++ b/.changes/next-release/bugfix-AWSQueryProtocol-1398d13.json
@@ -1,0 +1,6 @@
+{
+    "type": "bugfix",
+    "category": "AWS Query Protocol",
+    "contributor": "",
+    "description": "Fixed the bug that EC2 DescribeInstances query uses locale-dependent formatting"
+}
--- a/core/protocols/aws-query-protocol/src/main/java/software/amazon/awssdk/protocols/query/internal/marshall/ListQueryMarshaller.java
+++ b/core/protocols/aws-query-protocol/src/main/java/software/amazon/awssdk/protocols/query/internal/marshall/ListQueryMarshaller.java
@@ -16,6 +16,7 @@
 package software.amazon.awssdk.protocols.query.internal.marshall;
 
 import java.util.List;
+import java.util.Locale;
 import software.amazon.awssdk.annotations.SdkInternalApi;
 import software.amazon.awssdk.core.SdkField;
 import software.amazon.awssdk.core.traits.ListTrait;
@@ -31,7 +32,8 @@ public class ListQueryMarshaller implements QueryMarshaller<List<?>> {
         listTrait.isFlattened() ?
         String.format("%s.%d", path, i + 1) :
         String.format("%s.%s.%d", path, listTrait.memberFieldInfo().locationName(), i + 1);
-    private static final PathResolver EC2_QUERY_PATH_RESOLVER = (path, i, listTrait) -> String.format("%s.%d", path, i + 1);
+    private static final PathResolver EC2_QUERY_PATH_RESOLVER = (path, i, listTrait) -> String.format(Locale.ROOT, "%s.%d", path
+        , i + 1);
 
     private static final EmptyListMarshaller AWS_QUERY_EMPTY_LIST_MARSHALLER =
         (context, path) -> context.request().putRawQueryParameter(path, "");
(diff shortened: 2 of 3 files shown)


.changes/next-release/bugfix-AWSQueryProtocol-1398d13.json
```diff
@@ -0,0 +1,6 @@
+{
+    "type": "bugfix",
+    "category": "AWS Query Protocol",
+    "contributor": "",
+    "description": "Fixed the bug that EC2 DescribeInstances query uses locale-dependent formatting"
```
alextwoods: Description nit:  I'd lean towards describing the new behavior as well.  Something like "Use ROOT locale when formatting query paths".

core/protocols/aws-query-protocol/src/main/java/software/amazon/awssdk/protocols/query/internal/marshall/ListQueryMarshaller.java
```diff
@@ -31,7 +32,8 @@ public class ListQueryMarshaller implements QueryMarshaller<List<?>> {
         listTrait.isFlattened() ?
         String.format("%s.%d", path, i + 1) :
         String.format("%s.%s.%d", path, listTrait.memberFieldInfo().locationName(), i + 1);
-    private static final PathResolver EC2_QUERY_PATH_RESOLVER = (path, i, listTrait) -> String.format("%s.%d", path, i + 1);
+    private static final PathResolver EC2_QUERY_PATH_RESOLVER = (path, i, listTrait) -> String.format(Locale.ROOT, "%s.%d", path
```
millems: Would it be simpler to just do `path + "." + (i+1)`? 

String.format is relatively slow compared to building the string manually, anyway.

core/protocols/aws-query-protocol/src/main/java/software/amazon/awssdk/protocols/query/internal/marshall/ListQueryMarshaller.java
```diff
@@ -31,7 +32,8 @@ public class ListQueryMarshaller implements QueryMarshaller<List<?>> {
         listTrait.isFlattened() ?
         String.format("%s.%d", path, i + 1) :
         String.format("%s.%s.%d", path, listTrait.memberFieldInfo().locationName(), i + 1);
```
millems: What about other places we're doing %d for string formatting in paths, like this?

Fred1155:
> Overall change looks good - Are there other areas where we are formatting URL components that we should make a similar fix for?

Good suggestion! I think technically this fix could be applied to all path resolver that uses ``String.format("%d")``?

core/protocols/aws-query-protocol/src/main/java/software/amazon/awssdk/protocols/query/internal/marshall/ListQueryMarshaller.java
Fred1155: Thanks for the suggestion! will check other path resolvers that uses ``String.format("%d")``. Just did a test and using concatenation also fixes the locale problem.

millems (approved):

core/protocols/aws-query-protocol/src/main/java/software/amazon/awssdk/protocols/query/internal/marshall/ListQueryMarshaller.java
```diff
@@ -16,6 +16,7 @@
 package software.amazon.awssdk.protocols.query.internal.marshall;
 
 import java.util.List;
+import java.util.Locale;
```
alextwoods: Unused import now I believe

alextwoods (approved):

PR #6244, opened 2025-07-08
URI Cache for DynamoDB Account Id based Endpoints

davidh44:
## Motivation and Context

Continuation of https://github.com/aws/aws-sdk-java-v2/pull/6087 and https://github.com/aws/aws-sdk-java-v2/pull/6228
- Replacing `LruCache` with a random-eviction `BoundedCache` in `SdkUri`
- Cache size of 150
- When full, performs batch eviction of 10 entries (arbitrarily chosen)

## Modifications

- Added a SdkURI class which caches calls to the various URI constructors methods.
- Added customization configuration to codegen to enable usage of SdkURI class in generated endpoint resolution code.
- Enabled this config only for DynamoDB.

## Testing

- Unit test
- Canary tests

## Types of changes

### DIFF
--- a/.changes/next-release/feature-AWSSDKforJavav2-b405876.json
+++ b/.changes/next-release/feature-AWSSDKforJavav2-b405876.json
@@ -1,6 +1,6 @@
 {
     "type": "feature",
-    "category": "Amazon DyanmoDB",
+    "category": "Amazon DynamoDB",
     "contributor": "",
-    "description": "Enable caching calls to URI constructors for account-id based endpoints"
+    "description": "Enable caching results to URI constructors for account-id based endpoints"
 }
--- a/.changes/next-release/feature-AWSSDKforJavav2-b405876.json
+++ b/.changes/next-release/feature-AWSSDKforJavav2-b405876.json
@@ -1,0 +1,6 @@
+{
+    "type": "feature",
+    "category": "Amazon DyanmoDB",
+    "contributor": "",
+    "description": "Enable caching calls to URI constructors for account-id based endpoints"
+}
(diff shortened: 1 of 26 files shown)


RanVaknin (approved):
Thanks for addressing the comments. This looks good

.changes/next-release/feature-AWSSDKforJavav2-b405876.json
```diff
@@ -0,0 +1,6 @@
+{
+    "type": "feature",
+    "category": "Amazon DyanmoDB",
```
dagnir: Even though we have additional caching for DDB, do we need a separate entry for "AWS SDK for Java v2" since the change is global?

.changes/next-release/feature-AWSSDKforJavav2-b405876.json
```diff
@@ -0,0 +1,6 @@
...
+    "contributor": "",
+    "description": "Enable caching calls to URI constructors for account-id based endpoints"
```
dagnir: nit: we cache the results, not the calls

utils/src/main/java/software/amazon/awssdk/utils/cache/bounded/BoundedCache.java
```diff
@@ -0,0 +1,146 @@
...
+            }
+
+            cache.put(key, newValue);
+            cacheSize.incrementAndGet();
+            return newValue;
+        }
+    }
+
+    /**
+     * Clean up the cache by batch removing random entries of evictionBatchSize
```
dagnir: minor: there was a similar comment on the internal CR, but "random entries" isn't guaranteed, since the ConcurrentHashMap doesn't say anything about that. Can we just say "unspecified"?

dagnir (changes requested):

utils/src/main/java/software/amazon/awssdk/utils/cache/bounded/BoundedCache.java
davidh44: Updated to "unspecified"

.changes/next-release/feature-AWSSDKforJavav2-b405876.json
davidh44: Added separate changelog entry

dagnir (approved):

PR #6245, opened 2025-07-08
Expand protocol test coverage for json1.0, query and cbor

alextwoods:
This PR does three things (detailed in modifications section below):
1. Expand protocol test coverage for json1.0, query and cbor using directly translated smithy protocol tests.
2. Add support for new protocol test runner features to support the new tests. 
3. Minor codegen updates to handle the protocol test models

## Motivation and Context
This is a followup/expansion of #6230.

This change improves our protocol test coverage for the json1.0, query and cbor protocols and ensures that all cases covered by the official Smithy protocol tests are run.

## Modifications
**SDK Behavior changes**:
* fix codegeneration for enums with numeric string values (eg "0" -> "VALUE_0").  There are NO existing AWS services using these - they are only present in protocol tests and without this change cause a compilation error.
* support XmlNamespace with non-nested uri (adds new constructor)

**Protocol Test Runner updates**:
* add a cborEquals body matcher (equivalent functionality to jsonEquals).  Ensures that encoded bodies are functionally equal while allowing differences in the exact cbor bytes.
* add mustContain matchers to header and query params.
* jsonEquals body matcher - relax equality on whole number double vs integer (eg: we now consider `1.0` to be equal `1`)
* queryEquals body matcher - adds a new body matcher for query that ensures encoded body (or request parameters) are functionally equal.
* add resolvedHost assertions
* support `host` given (allowing configuring custom endpoint)
* support requests with hostPrefixes/labels added (requires workarounds for wiremock - we need to avoid dns lookups being done for endpoints like "foo.localhost"
* Dynamic parsing of expected timestamp values - allowing fractional seconds with double values + either second or millisecond epoch values.  Legacy tests use epoch milliseconds and smithy tests use epoch seconds.

**Protocol Test Suite changes/Additions**:
* Add a new query protocol service (under smithy-query) - used for the ported smithy query tests.
* Replace the old query protocol test suite with the new official ported smithy tests.
* Add a new json1.0 service + official ported smithy tests
* Replace the protocol tests for cbor with the official ported smithy tests.

## FAQ
**Why add a new query protocol service instead of replacing the existing one?**
The existing query model/service is used in a large number of custom/hand written tests cases (eg: [QueryExceptionTests](https://github.com/aws/aws-sdk-java-v2/blob/master/test/protocol-tests/src/test/java/software/amazon/awssdk/protocol/tests/exception/QueryExceptionTests.java)).  Adding a new service for the new protocol test suite allows us to leave these existing tests alone.

## Types of changes

## Checklist

## License


test/protocol-tests-core/pom.xml
```diff
@@ -137,6 +137,10 @@
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-annotations</artifactId>
         </dependency>
+        <dependency>
```
alextwoods, line 140: Required to support the cborEquals body matcher - we use our internal namespaced CborFactory in the protocol, but the dependency here in the test package should be okay (since we depend on jackson core similarly above).

L-Applin: should we instead use `third-party-jackson-dataformat-cbor`?

test/protocol-tests-core/src/main/java/software/amazon/awssdk/protocol/asserts/marshalling/HeadersAssertion.java
```diff
@@ -49,13 +55,19 @@ protected void doAssert(LoggedRequest actual) throws Exception {
...
+            assertMustContainHeaders(actual.getHeaders());
+        }
     }
 
     private void assertHeadersContains(HttpHeaders actual) {
         contains.forEach((expectedKey, expectedValues) -> {
             assertTrue(String.format("Header '%s' was expected to be present. Actual headers: %s", expectedKey, actual),
                        actual.getHeader(expectedKey).isPresent());
             List<String> actualValues = actual.getHeader(expectedKey).values();
+            if (expectedKey.equalsIgnoreCase("Content-Type") && actualValues.size() == 1) {
```
L-Applin, line 69: It's probably my lack of context on this, but why is this special case for charset header needed?

alextwoods: Added a comment in code for this as well, but our marshalling adds charset to content-type.  This is valid but not included in the protocol tests.

test/protocol-tests-core/pom.xml
alextwoods: I wanted to use our cbor lib, but as far as I can tell, its not possible to use with a Jackson ObjectMapper that just lets us get JsonNodes to compare.
I based the CborBodyAssertion on the existing [JsonBodyAssertion](https://github.com/aws/aws-sdk-java-v2/blob/master/test/protocol-tests-core/src/main/java/software/amazon/awssdk/protocol/asserts/marshalling/JsonBodyAssertion.java#L21) which also uses the standard jackson ObjectMapper (rather than our third-party-jackson dependencies).

I'm not sure if theres a better way to do this though

test/protocol-tests-core/src/main/java/software/amazon/awssdk/protocol/asserts/marshalling/JsonBodyAssertion.java
```diff
@@ -37,8 +41,39 @@ public JsonBodyAssertion(String jsonEquals) {
...
+     * We serialize some numbers (in particular epoch timestamps) as doubles such as 123.000.
+     * In protocol tests, these values are parsed as longs.  This conversion insures that
+     * 123.000 will equal 123.
+     */
+    public static JsonNode convertWholeNumberDoubleToLong(JsonNode node) {
+        if (node.isDouble()) {
+            double value = node.doubleValue();
+            if (value % 1 == 0) {
+                if (value >= Integer.MIN_VALUE && value <= Integer.MAX_VALUE) {
+                    return new IntNode((int) value);
```
RanVaknin, line 58: >In protocol tests, these values are parsed as longs.

Why are we using int here? Can we not treat all of these doubles just as longs?

alextwoods: Great question - the `assertEquals` above will fail if the Nodes are not of the same type, so a IntNode(123) != LongNode(123), so this uses the same logic as the ObjectMapper to ensure the node types match what was parsed.

codegen/src/main/java/software/amazon/awssdk/codegen/naming/DefaultNamingStrategy.java
```diff
@@ -302,6 +302,11 @@ public String getEnumValueName(String enumValue) {
         // Special cases
         result = result.replace("textORcsv", "TEXT_OR_CSV");
 
+        // leading digits, add a prefix
+        if (result.matches("^\\d.*")) {
+            result = "VALUE_" + result;
```
RanVaknin, line 307: I understand the java limitation of not being able to have a number as an enum value, but can you explain how this substitution works in practice?

RanVaknin (approved):

codegen/src/main/java/software/amazon/awssdk/codegen/naming/DefaultNamingStrategy.java
alextwoods: When the value has leading digits (eg: `0` or `123_some_value`) the regex will match, and then we just prefix it with "VALUE", so we end up with `VALUE_0` and `VALUE_123_SOME_VALUE` for those cases.

PR #6249, opened 2025-07-10
Add PresignedUrlManager support to sync S3 client codegen

jencymaryjoseph:
## Motivation and Context
`PresignedUrlManager` is a new API that users need to access through the S3 client. This PR extends the codegen framework to generate `PresignedUrlManager` support for S3 clients. 

## Modifications
Added `presignedUrlManagerSupported` flag to `CustomizationConfig`.
Extended codegen to generate presignedUrlManager() method in sync client interfaces and implementations.

## Testing
Added codegen unit tests for pre-signed URL manager generation.

## Types of changes

## Checklist

- [] I have added a changelog entry. Adding a new entry must be accomplished by running the `scripts/new-change` script and following the instructions. Commit the new file created by the script in `.changes/next-release` with your changes.

## License

### DIFF
--- a/codegen/src/main/java/software/amazon/awssdk/codegen/poet/client/DelegatingSyncClientClass.java
+++ b/codegen/src/main/java/software/amazon/awssdk/codegen/poet/client/DelegatingSyncClientClass.java
@@ -232,4 +232,20 @@ public class DelegatingSyncClientClass extends SyncClientInterface {
                          .addStatement("return delegate.serviceClientConfiguration()")
                          .build();
     }
+
+    @Override
+    protected void addPresignedUrlManagerMethod(TypeSpec.Builder type) {
+        if (model.getCustomizationConfig().getPresignedUrlManagerSupported()) {
+            ClassName returnType = poetExtensions.getPresignedUrlManagerInterface();
+            MethodSpec method = MethodSpec.methodBuilder("presignedUrlManager")
+                                          .addAnnotation(Override.class)
+                                          .addModifiers(PUBLIC)
+                                          .returns(returnType)
+                                          .addJavadoc("Creates an instance of {@link $T} object with the "
+                                                      + "configuration set on this client.", returnType)
+                                          .addStatement("return delegate.presignedUrlManager()")
+                                          .build();
+            type.addMethod(method);
+        }
+    }
 }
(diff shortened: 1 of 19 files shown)


codegen/src/main/java/software/amazon/awssdk/codegen/poet/client/DelegatingSyncClientClass.java
```diff
@@ -232,4 +232,20 @@ protected MethodSpec serviceClientConfigMethod() {
                          .addStatement("return delegate.serviceClientConfiguration()")
                          .build();
     }
+
+    @Override
+    protected void addPresignedUrlManagerMethod(TypeSpec.Builder type) {
```
zoewangg, line 237: I thought we are only going to generate this in async client, is that not the case?

jencymaryjoseph: The delegating client needs the presignedUrlManager implementation because `DelegatingS3Client` implements the `S3Client` interface - without it, we get a compilation error "CLASS DelegatingS3Client.presignedUrlManager cannot be default" since interface methods can't be default methods in this case, so we need to maintain the delegation pattern by forwarding the call to the delegate.

jencymaryjoseph: No, we are implementing presignedUrlManager support for both sync and async S3 clients. While the initial discussion might have focused on async clients, the final design has both sync and async implementations.

> I thought we are only going to generate this in async client, is that not the case?

joviegas: Presigned URLs in S3 are not exclusively limited to multipart upload scenarios . They serve a broader range of use cases that extend beyond just multipart operations.

Presigned URLs can be generated for various S3 operations, including single-part file uploads and downloads. This flexibility is particularly valuable because many file operations don't require the complexity of multipart handling, especially for smaller files or straightforward download scenarios.

The decision to make this functionality available in the synchronous client addresses a practical developer need. By providing presigned URL generation in the sync client, developers who are already using synchronous operations can access this feature without introducing additional dependencies. This approach eliminates the need to add Netty dependencies or maintain a separate async client solely for presigned URL operations, which would add unnecessary complexity to their application architecture.

This implementation strategy provides developers with consistent access to presigned URL functionality regardless of their chosen client type, making the feature more accessible and reducing the overhead of managing multiple client configurations.

jencymaryjoseph:
Initial implementation will be limited to S3 async client to manage complexity and deliver core functionality efficiently. We are deferring S3 sync client implementation due to additional complexities around multipart downloads and byte-range based functionality.

PR #6255, opened 2025-07-14
Add AsyncPresignedUrlManager for S3 GetObject operations with presigned URLs

jencymaryjoseph:
Implements `DefaultAsyncPresignedUrlManager` to support asynchronous GetObject operations using presigned URLs.

## Motivation and Context
Enables async S3 GetObject operations via presigned URLs, addressing the need for non-blocking presigned URL operations.

## Modifications
Added  async API `AsyncPresignedUrlManager` and its implementation `DefaultAsyncPresignedUrlManager` 

## Testing
Added comprehensive unit tests covering success responses, error handling, request configuration with range headers, and invalid URL scenarios using MockAsyncHttpClient.

## Types of changes

## Checklist

## License

### DIFF
--- a/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/presignedurl/DefaultAsyncPresignedUrlManager.java
+++ b/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/presignedurl/DefaultAsyncPresignedUrlManager.java
@@ -64,6 +64,7 @@ public final class DefaultAsyncPresignedUrlManager implements AsyncPresignedUrlM
     private final AsyncClientHandler clientHandler;
     private final AwsS3ProtocolFactory protocolFactory;
     private final SdkClientConfiguration clientConfiguration;
+    private final List<MetricPublisher> metricPublishers;
     private final AwsProtocolMetadata protocolMetadata;
     
     public DefaultAsyncPresignedUrlManager(AsyncClientHandler clientHandler,
@@ -72,8 +73,11 @@ public final class DefaultAsyncPresignedUrlManager implements AsyncPresignedUrlM
                                            AwsProtocolMetadata protocolMetadata) {
         this.clientHandler = clientHandler;
         this.protocolFactory = protocolFactory;
-        this.clientConfiguration = clientConfiguration;
         this.protocolMetadata = protocolMetadata;
+        this.clientConfiguration = updateSdkClientConfiguration(clientConfiguration);
+        this.metricPublishers = Optional.ofNullable(
+            this.clientConfiguration.option(SdkClientOption.METRIC_PUBLISHERS))
+            .orElse(Collections.emptyList());
     }
 
     @Override
@@ -88,13 +92,9 @@ public final class DefaultAsyncPresignedUrlManager implements AsyncPresignedUrlM
                 .range(presignedUrlGetObjectRequest.range())
                 .build();
 
-        SdkClientConfiguration updatedClientConfiguration = updateSdkClientConfiguration(this.clientConfiguration);
-        List<MetricPublisher> metricPublishers = Optional.ofNullable(
-            updatedClientConfiguration.option(SdkClientOption.METRIC_PUBLISHERS))
-            .orElse(Collections.emptyList());
         MetricCollector apiCallMetricCollector = metricPublishers.isEmpty() ?
             NoOpMetricCollector.create() : MetricCollector.create("ApiCall");
-        
+
         try {
             apiCallMetricCollector.reportMetric(CoreMetric.SERVICE_ID, "S3");
             //TODO: Discuss if we need to change OPERATION_NAME as part of Surface API Review
@@ -102,7 +102,8 @@ public final class DefaultAsyncPresignedUrlManager implements AsyncPresignedUrlM
 
             Pair<AsyncResponseTransformer<GetObjectResponse, ReturnT>, CompletableFuture<Void>> pair =
                     AsyncResponseTransformerUtils.wrapWithEndOfStreamFuture(asyncResponseTransformer);
-            asyncResponseTransformer = pair.left();
+            AsyncResponseTransformer<GetObjectResponse, ReturnT> finalAsyncResponseTransformer = pair.left();
+            asyncResponseTransformer = finalAsyncResponseTransformer;
             CompletableFuture<Void> endOfStreamFuture = pair.right();
 
             HttpResponseHandler<GetObjectResponse> responseHandler = protocolFactory.createResponseHandler(
@@ -116,16 +117,15 @@ public final class DefaultAsyncPresignedUrlManager implements AsyncPresignedUrlM
                             .withProtocolMetadata(protocolMetadata)
                             .withResponseHandler(responseHandler)
                             .withErrorResponseHandler(errorResponseHandler)
-                            .withRequestConfiguration(updatedClientConfiguration)
+                            .withRequestConfiguration(clientConfiguration)
                             .withInput(internalRequest)
                             .withMetricCollector(apiCallMetricCollector)
                             // TODO: Deprecate IS_DISCOVERED_ENDPOINT, use new SKIP_ENDPOINT_RESOLUTION for better semantics
                             .putExecutionAttribute(SdkInternalExecutionAttribute.IS_DISCOVERED_ENDPOINT, true)
                             .withMarshaller(new PresignedUrlGetObjectRequestMarshaller(protocolFactory)),
                                                                                         asyncResponseTransformer);
-            CompletableFuture<ReturnT> whenCompleteFuture = null;
-            AsyncResponseTransformer<GetObjectResponse, ReturnT> finalAsyncResponseTransformer = asyncResponseTransformer;
-            whenCompleteFuture = executeFuture.whenComplete((r, e) -> {
+            
+            CompletableFuture<ReturnT> whenCompleteFuture = executeFuture.whenComplete((r, e) -> {
                 if (e != null) {
                     runAndLogError(log, "Exception thrown in exceptionOccurred callback, ignoring",
                                    () -> finalAsyncResponseTransformer.exceptionOccurred(e));
--- a/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/presignedurl/DefaultAsyncPresignedUrlManager.java
+++ b/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/presignedurl/DefaultAsyncPresignedUrlManager.java
@@ -1,0 +1,154 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.services.s3.internal.presignedurl;
+
+import static software.amazon.awssdk.core.client.config.SdkClientOption.SIGNER_OVERRIDDEN;
+import static software.amazon.awssdk.utils.FunctionalUtils.runAndLogError;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.awscore.exception.AwsServiceException;
+import software.amazon.awssdk.awscore.internal.AwsProtocolMetadata;
+import software.amazon.awssdk.core.async.AsyncResponseTransformer;
+import software.amazon.awssdk.core.async.AsyncResponseTransformerUtils;
+import software.amazon.awssdk.core.client.config.SdkAdvancedClientOption;
+import software.amazon.awssdk.core.client.config.SdkClientConfiguration;
+import software.amazon.awssdk.core.client.config.SdkClientOption;
+import software.amazon.awssdk.core.client.handler.AsyncClientHandler;
+import software.amazon.awssdk.core.client.handler.ClientExecutionParams;
+import software.amazon.awssdk.core.exception.SdkClientException;
+import software.amazon.awssdk.core.http.HttpResponseHandler;
+import software.amazon.awssdk.core.interceptor.SdkInternalExecutionAttribute;
+import software.amazon.awssdk.core.metrics.CoreMetric;
+import software.amazon.awssdk.core.signer.NoOpSigner;
+import software.amazon.awssdk.metrics.MetricCollector;
+import software.amazon.awssdk.metrics.MetricPublisher;
+import software.amazon.awssdk.metrics.NoOpMetricCollector;
+import software.amazon.awssdk.protocols.xml.AwsS3ProtocolFactory;
+import software.amazon.awssdk.protocols.xml.XmlOperationMetadata;
+import software.amazon.awssdk.services.s3.internal.presignedurl.model.PresignedUrlGetObjectRequestWrapper;
+import software.amazon.awssdk.services.s3.model.GetObjectResponse;
+import software.amazon.awssdk.services.s3.model.InvalidObjectStateException;
+import software.amazon.awssdk.services.s3.model.NoSuchKeyException;
+import software.amazon.awssdk.services.s3.model.S3Exception;
+import software.amazon.awssdk.services.s3.presignedurl.AsyncPresignedUrlManager;
+import software.amazon.awssdk.services.s3.presignedurl.model.PresignedUrlGetObjectRequest;
+import software.amazon.awssdk.utils.CompletableFutureUtils;
+import software.amazon.awssdk.utils.Pair;
+
+/**
+ * Default implementation of {@link AsyncPresignedUrlManager} for executing S3 operations asynchronously using presigned URLs.
+ */
+@SdkInternalApi
+public final class DefaultAsyncPresignedUrlManager implements AsyncPresignedUrlManager {
+    private static final Logger log = LoggerFactory.getLogger(DefaultAsyncPresignedUrlManager.class);
+
+    private final AsyncClientHandler clientHandler;
+    private final AwsS3ProtocolFactory protocolFactory;
+    private final SdkClientConfiguration clientConfiguration;
+    private final AwsProtocolMetadata protocolMetadata;
+    
+    public DefaultAsyncPresignedUrlManager(AsyncClientHandler clientHandler,
+                                           AwsS3ProtocolFactory protocolFactory,
+                                           SdkClientConfiguration clientConfiguration,
+                                           AwsProtocolMetadata protocolMetadata) {
+        this.clientHandler = clientHandler;
+        this.protocolFactory = protocolFactory;
+        this.clientConfiguration = clientConfiguration;
+        this.protocolMetadata = protocolMetadata;
+    }
+
+    @Override
+    public <ReturnT> CompletableFuture<ReturnT> getObject(
+            PresignedUrlGetObjectRequest presignedUrlGetObjectRequest,
+            AsyncResponseTransformer<GetObjectResponse, ReturnT> asyncResponseTransformer)
+            throws NoSuchKeyException, InvalidObjectStateException,
+                   AwsServiceException, SdkClientException, S3Exception {
+
+        PresignedUrlGetObjectRequestWrapper internalRequest = PresignedUrlGetObjectRequestWrapper.builder()
+                .url(presignedUrlGetObjectRequest.presignedUrl())
+                .range(presignedUrlGetObjectRequest.range())
+                .build();
+
+        SdkClientConfiguration updatedClientConfiguration = updateSdkClientConfiguration(this.clientConfiguration);
+        List<MetricPublisher> metricPublishers = Optional.ofNullable(
+            updatedClientConfiguration.option(SdkClientOption.METRIC_PUBLISHERS))
+            .orElse(Collections.emptyList());
+        MetricCollector apiCallMetricCollector = metricPublishers.isEmpty() ?
+            NoOpMetricCollector.create() : MetricCollector.create("ApiCall");
+        
+        try {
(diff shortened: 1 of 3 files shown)


services/s3/src/main/java/software/amazon/awssdk/services/s3/presignedurl/AsyncPresignedUrlManager.java
```diff
@@ -0,0 +1,81 @@
...
+import software.amazon.awssdk.services.s3.model.NoSuchKeyException;
+import software.amazon.awssdk.services.s3.model.S3Exception;
+import software.amazon.awssdk.services.s3.presignedurl.model.PresignedUrlGetObjectRequest;
+
+/**
+ * Interface for executing S3 operations asynchronously using presigned URLs. This can be accessed using
+ * {@link S3AsyncClient#presignedUrlManager()}.
+ */
+@SdkPublicApi
+public interface AsyncPresignedUrlManager {
```
millems, line 34: Have we bikeshed the naming yet?

services/s3/src/main/java/software/amazon/awssdk/services/s3/presignedurl/AsyncPresignedUrlManager.java
```diff
@@ -0,0 +1,81 @@
...
+import software.amazon.awssdk.services.s3.S3AsyncClient;
+import software.amazon.awssdk.services.s3.model.GetObjectResponse;
+import software.amazon.awssdk.services.s3.model.InvalidObjectStateException;
+import software.amazon.awssdk.services.s3.model.NoSuchKeyException;
+import software.amazon.awssdk.services.s3.model.S3Exception;
+import software.amazon.awssdk.services.s3.presignedurl.model.PresignedUrlGetObjectRequest;
+
+/**
+ * Interface for executing S3 operations asynchronously using presigned URLs. This can be accessed using
+ * {@link S3AsyncClient#presignedUrlManager()}.
```
millems, line 31: Is there a plan to be able to create one directly? That could be useful, for example if the user doesn't have credentials and doesn't want to get some.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/presignedurl/DefaultAsyncPresignedUrlManager.java
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    @Override
+    public <ReturnT> CompletableFuture<ReturnT> getObject(
+            PresignedUrlGetObjectRequest presignedUrlGetObjectRequest,
+            AsyncResponseTransformer<GetObjectResponse, ReturnT> asyncResponseTransformer)
+            throws NoSuchKeyException, InvalidObjectStateException,
+                   AwsServiceException, SdkClientException, S3Exception {
+
+        PresignedUrlGetObjectRequestWrapper internalRequest = PresignedUrlGetObjectRequestWrapper.builder()
```
millems, line 90: What purpose does this wrapper serve?

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/presignedurl/DefaultAsyncPresignedUrlManager.java
```diff
@@ -0,0 +1,154 @@
...
+                .url(presignedUrlGetObjectRequest.presignedUrl())
+                .range(presignedUrlGetObjectRequest.range())
+                .build();
+
+        SdkClientConfiguration updatedClientConfiguration = updateSdkClientConfiguration(this.clientConfiguration);
+        List<MetricPublisher> metricPublishers = Optional.ofNullable(
+            updatedClientConfiguration.option(SdkClientOption.METRIC_PUBLISHERS))
+            .orElse(Collections.emptyList());
```
millems: Can we do this in the constructor?

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/presignedurl/DefaultAsyncPresignedUrlManager.java
```diff
@@ -0,0 +1,154 @@
...
+            CompletableFuture<Void> endOfStreamFuture = pair.right();
+
+            HttpResponseHandler<GetObjectResponse> responseHandler = protocolFactory.createResponseHandler(
+                GetObjectResponse::builder, new XmlOperationMetadata().withHasStreamingSuccessResponse(true));
+
+            HttpResponseHandler<AwsServiceException> errorResponseHandler = protocolFactory.createErrorResponseHandler();
+
+            CompletableFuture<ReturnT> executeFuture = clientHandler.execute(
+                    new ClientExecutionParams<PresignedUrlGetObjectRequestWrapper, GetObjectResponse>()
+                            .withOperationName("PresignedUrlGetObject")
```
millems: This doesn't match the operation name from the metrics or the service's operation name. Should we just use the service's operation name?

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/presignedurl/DefaultAsyncPresignedUrlManager.java
```diff
@@ -0,0 +1,154 @@
...
+                            .withProtocolMetadata(protocolMetadata)
+                            .withResponseHandler(responseHandler)
+                            .withErrorResponseHandler(errorResponseHandler)
+                            .withRequestConfiguration(updatedClientConfiguration)
+                            .withInput(internalRequest)
+                            .withMetricCollector(apiCallMetricCollector)
+                            // TODO: Deprecate IS_DISCOVERED_ENDPOINT, use new SKIP_ENDPOINT_RESOLUTION for better semantics
```
millems, line 123: How are we tracking TODOs, to ensure they get fixed before release?

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/presignedurl/DefaultAsyncPresignedUrlManager.java
```diff
@@ -0,0 +1,154 @@
...
+                            .putExecutionAttribute(SdkInternalExecutionAttribute.IS_DISCOVERED_ENDPOINT, true)
+                            .withMarshaller(new PresignedUrlGetObjectRequestMarshaller(protocolFactory)),
+                                                                                        asyncResponseTransformer);
+            CompletableFuture<ReturnT> whenCompleteFuture = null;
+            AsyncResponseTransformer<GetObjectResponse, ReturnT> finalAsyncResponseTransformer = asyncResponseTransformer;
+            whenCompleteFuture = executeFuture.whenComplete((r, e) -> {
```
millems: Can `whenCompleteFuture` be declared inline?

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/presignedurl/DefaultAsyncPresignedUrlManager.java
```diff
@@ -0,0 +1,154 @@
...
+                            .withErrorResponseHandler(errorResponseHandler)
+                            .withRequestConfiguration(updatedClientConfiguration)
+                            .withInput(internalRequest)
+                            .withMetricCollector(apiCallMetricCollector)
+                            // TODO: Deprecate IS_DISCOVERED_ENDPOINT, use new SKIP_ENDPOINT_RESOLUTION for better semantics
+                            .putExecutionAttribute(SdkInternalExecutionAttribute.IS_DISCOVERED_ENDPOINT, true)
+                            .withMarshaller(new PresignedUrlGetObjectRequestMarshaller(protocolFactory)),
+                                                                                        asyncResponseTransformer);
+            CompletableFuture<ReturnT> whenCompleteFuture = null;
+            AsyncResponseTransformer<GetObjectResponse, ReturnT> finalAsyncResponseTransformer = asyncResponseTransformer;
```
millems: We could declare finalAsyncResponseTransformer inline to where asyncResponseTransformer is reassigned, and not need to have this line here.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/presignedurl/DefaultAsyncPresignedUrlManager.java
```diff
@@ -0,0 +1,154 @@
...
+        this.clientConfiguration = clientConfiguration;
+        this.protocolMetadata = protocolMetadata;
+    }
+
+    @Override
+    public <ReturnT> CompletableFuture<ReturnT> getObject(
+            PresignedUrlGetObjectRequest presignedUrlGetObjectRequest,
+            AsyncResponseTransformer<GetObjectResponse, ReturnT> asyncResponseTransformer)
+            throws NoSuchKeyException, InvalidObjectStateException,
+                   AwsServiceException, SdkClientException, S3Exception {
```
millems, line 88: Is there an opportunity to share code with the DefaultS3AsyncClient instead of copying it? That would greatly simplify keeping them in sync (since one is code generated) as well as the addition of support for new operations in the future.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/presignedurl/DefaultAsyncPresignedUrlManager.java
```diff
@@ -0,0 +1,154 @@
...
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.services.s3.internal.presignedurl;
```
millems, line 16: Have we talked about whether this feature is universally useful enough to include in the default S3 package? It feels like it might be worth keeping the core S3 client artifact small and building this in a different package. That may not be an issue if we can do more code sharing with the generated client (I commented this below).

services/s3/src/test/java/software/amazon/awssdk/services/s3/internal/presignedurl/DefaultAsyncPresignedUrlManagerTest.java
```diff
@@ -0,0 +1,373 @@
...
+import software.amazon.awssdk.regions.Region;
+import software.amazon.awssdk.services.s3.model.GetObjectResponse;
+import software.amazon.awssdk.services.s3.model.InvalidObjectStateException;
+import software.amazon.awssdk.services.s3.model.NoSuchKeyException;
+import software.amazon.awssdk.services.s3.model.S3Exception;
+import software.amazon.awssdk.services.s3.presignedurl.model.PresignedUrlGetObjectRequest;
+import software.amazon.awssdk.testutils.service.http.MockAsyncHttpClient;
+
+/**
+ * Tests for {@link DefaultAsyncPresignedUrlManager} using MockAsyncHttpClient to verify HTTP interactions.
```
millems, line 74: Can we add some integration tests as well?

services/s3/src/test/java/software/amazon/awssdk/services/s3/internal/presignedurl/DefaultAsyncPresignedUrlManagerTest.java
```diff
@@ -0,0 +1,373 @@
...
+
+        SdkClientConfiguration clientConfiguration = getDefaultSdkConfigs();
+        protocolMetadata = AwsProtocolMetadata.builder()
+                                              .serviceProtocol(AwsServiceProtocol.REST_XML)
+                                              .build();
+        protocolFactory = initProtocolFactory(clientConfiguration);
+        clientHandler = new AwsAsyncClientHandler(clientConfiguration);
+
+        presignedUrlManager = new DefaultAsyncPresignedUrlManager(
+            clientHandler, protocolFactory, clientConfiguration, protocolMetadata);
```
millems, line 112: Once it's implemented, we should test creation of this from the API that the customer uses. When we do that, we need to test all of the configuration overrides they may be trying. Retries, user agent, etc. That's probably dozens of tests, but it'll be important to make sure all of the functionality works and is safe against regression.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/presignedurl/DefaultAsyncPresignedUrlManager.java
jencymaryjoseph: The wrapper serves as an internal adapter to convert `PresignedUrlGetObjectRequest` into a format compatible with SDK's execution pipeline while avoiding exposure of unnecessary S3Request configurations like credentials provider, signers,.. that don't apply to pre-signed URL operations.

services/s3/src/main/java/software/amazon/awssdk/services/s3/presignedurl/AsyncPresignedUrlManager.java
jencymaryjoseph: No, we haven't had a naming discussion yet. We will have a surface API review soon where we can discuss the naming convention for `AsyncPresignedUrlManager` and others.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/presignedurl/DefaultAsyncPresignedUrlManager.java
jencymaryjoseph: I chose "`PresignedUrlGetObject`" to clearly distinguish pre-signed URL operations in metrics and logs, enabling better troubleshooting and performance analysis. However, we can align with service operation names for consistency if preferred.

services/s3/src/main/java/software/amazon/awssdk/services/s3/presignedurl/AsyncPresignedUrlManager.java
jencymaryjoseph: During the design phase, we decided that `AsyncPresignedUrlManager` must be accessed through `S3AsyncClient` since it relies on the `ClientHandler` component to execute requests and maintain core SDK functionality like retries, metrics collection, and error handling.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/presignedurl/DefaultAsyncPresignedUrlManager.java
jencymaryjoseph: During the design phase, we decided to keep `AsyncPresignedUrlManager` in the default S3 package following a similar pattern to `BatchManager`, as it's a core S3 utility. Additionally, since the implementation requires sharing the `ClientHandler` from `S3Client` for maintaining SDK functionality (like retries and metrics), keeping it in the same package allows for better integration. This approach was finalized in the design to ensure consistent behavior with other SDK features while maintaining proper access to internal components.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/presignedurl/DefaultAsyncPresignedUrlManager.java
jencymaryjoseph: While some aspects of `getObject` implementation could potentially be shared, the current design intentionally keeps pre-signed URL operations separate from `DefaultS3AsyncClient` since they require fundamentally different request handling (bypassing signing and endpoint resolution). Additionally, since `DefaultS3AsyncClient` is code generated, integrating shared code would require modifications to the code generation process, which would add complexity. The separation helps maintain cleaner architecture and easier maintenance of requirements for pre-signed URL operations.

services/s3/src/test/java/software/amazon/awssdk/services/s3/internal/presignedurl/DefaultAsyncPresignedUrlManagerTest.java
jencymaryjoseph: Yes, I planned to add the integration tests in a follow-up change once the code generation changes for adding `presignedUrlManager`() to `S3AsyncClient` are completed. The current PR focuses on the core implementation and unit tests.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/presignedurl/DefaultAsyncPresignedUrlManager.java
jencymaryjoseph: I've created tickets to track the TODO items.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/presignedurl/DefaultAsyncPresignedUrlManager.java
jencymaryjoseph: I've implemented the suggested change by declaring finalAsyncResponseTransformer inline where `asyncResponseTransformer` is reassigned. However, I needed to keep a separate declaration in the catch block because `asyncResponseTransformer` is modified in the try block, making it not effectively final.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/presignedurl/DefaultAsyncPresignedUrlManager.java
jencymaryjoseph: Thanks Matt, I've moved the client configuration update and metric publishers initialization to the constructor.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/presignedurl/DefaultAsyncPresignedUrlManager.java
jencymaryjoseph: I've updated the code to declare whenCompleteFuture inline with its initialization, eliminating the unnecessary null assignment.

services/s3/src/test/java/software/amazon/awssdk/services/s3/internal/presignedurl/DefaultAsyncPresignedUrlManagerTest.java
jencymaryjoseph: If we're keeping this functionality in the default S3 package as discussed above, should we still add regression tests?

services/s3/src/main/java/software/amazon/awssdk/services/s3/presignedurl/AsyncPresignedUrlManager.java
millems: Okay, let's keep an eye out for feedback on this. I suspect people might want to go from having a presigner to generate presigned requests to having a presignedUrlManager to execute those requests, which isn't possible in this model.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/presignedurl/DefaultAsyncPresignedUrlManager.java
millems: Got it!

By the way, is there a concern that you are creating a `PresignedUrlGetObjectRequest` when we already have a `PresignedGetObjectRequest` that comes from the presigner? It feels weird that we have two different representations with slightly different names and no way to convert between them.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/presignedurl/DefaultAsyncPresignedUrlManager.java
millems: It feels weird to be "making up" service operations, since the service is the owner for them. This is still fundamentally a get-object-request, just a presigned one. Maybe there's a different way we could indicate in the metrics and logs that this is a presigned get object request instead of a modeled one?

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/presignedurl/DefaultAsyncPresignedUrlManager.java
millems: That makes sense for now. We can consider a refactor in the future if it becomes a burden.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/presignedurl/DefaultAsyncPresignedUrlManager.java
jencymaryjoseph: Okay, yeah, forgot to mention, i changed the operation name to "GetObject".

services/s3/src/test/java/software/amazon/awssdk/services/s3/internal/presignedurl/DefaultAsyncPresignedUrlManagerTest.java
millems: The concern I'm hoping regression tests would mitigate is someone changing something in the shared core code, only thinking about the S3 client. They'd test the S3 client with their change and think everything's fine. We'd like tests here to break and tell them "your change to retries broke a use-case in the presigned URL manager".

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/presignedurl/DefaultAsyncPresignedUrlManager.java
millems: I still see `PresignedUrlGetObject` here, maybe the commit hasn't been pushed yet?

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/presignedurl/DefaultAsyncPresignedUrlManager.java
jencymaryjoseph: `PresignedGetObjectRequest` is the output of the presigning process containing the full HTTP details, while `PresignedUrlGetObjectRequest` is a request model for executing requests with an already pre-signed URL, allowing for additional request customization like range headers and potential future headers without requiring re-signing.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/presignedurl/DefaultAsyncPresignedUrlManager.java
jencymaryjoseph: missed that, pushed the commit now.

millems (approved):
LGTM, let's follow-up offline on how to increase the cohesion between the presign URL generation APIs and the execution APIs.

jencymaryjoseph:
> LGTM, let's follow-up offline on how to increase the cohesion between the presign URL generation APIs and the execution APIs.

Added a task JAVA-8383 to track this.

Issue #2914, opened 2021-12-13
using KinesisAsyncClient: An exceptionCaught() event was fired. java.io.IOException: The channel was closed before the protocol could be determined.


filipglojnari:
### Describe the bug

`java.io.IOException: The channel was closed before the protocol could be determined.` exception is thrown when using the aws-sdk KinesisAsyncClient. We start multiple threads and they are all publishing to the kinesis stream via `KinesisAsyncClient.putRecord()`.  After some time (~6h) the application is running,  we can see the warn logs, followed by exception that causes them. 

`WARN  i.n.channel.DefaultChannelPipeline - An exceptionCaught() event was fired, and it reached at the tail of the pipeline. It usually means the last handler in the pipeline did not handle the exception.`

`java.io.IOException: The channel was closed before the protocol could be determined.`

### Expected behavior

This should not be thrown to the user.

### Current behavior

Warn and exception are thrown after publishing (for longer than ~6h) to kinesis stream from multi-thread env without knowing what exactly caused the exception from our code. Full message and stacktrace:

`2021-11-02 09:46:59,487 4017786 [aws-java-sdk-NettyEventLoop-1-1] WARN  i.n.channel.DefaultChannelPipeline - An exceptionCaught() event was fired, and it reached at the tail of the pipeline. It usually means the last handler in the pipeline did not handle the exception.`

      java.io.IOException: The channel was closed before the protocol could be determined.
              at software.amazon.awssdk.http.nio.netty.internal.http2.Http2SettingsFrameHandler.channelUnregistered(Http2SettingsFrameHandler.java:58)
              at io.netty.channel.AbstractChannelHandlerContext.invokeChannelUnregistered(AbstractChannelHandlerContext.java:198)        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelUnregistered(AbstractChannelHandlerContext.java:184)        at io.netty.channel.AbstractChannelHandlerContext.fireChannelUnregistered(AbstractChannelHandlerContext.java:177)  
              at io.netty.channel.DefaultChannelPipeline$HeadContext.channelUnregistered(DefaultChannelPipeline.java:1388)       
[...]

PR #6258, opened 2025-07-15
ChannelClosedException when server close connection right after the last http2 stream

huajiang-tubi:
## Motivation and Context

```mermaid
sequenceDiagram
    participant aws AS AWS Service
    participant channel AS Channel Event Loop
    participant pool AS HttpOrHttp2ChannelPool.eventLoop

    aws->>+channel: 1. LastHttpContent
    channel->>+channel: 2. ResponseHandler.finalizeResponse
    channel->>+aws: 3. Http2ResetFrame
    channel->>+channel: 4. HttpOrHttp2ChannelPool.release
    channel->>pool: 5. doInEventLoop(() -> release0(channel, promise))
    deactivate channel
    deactivate channel
    deactivate channel

    aws->>+channel: 6. Close The Connection
    channel->>+channel: 7. Http2ConnectionHandler.channelInactive
    channel->>+channel: 8. MultiplexedChannelRecord.closeAndExecuteOnChildChannels
    deactivate channel
    deactivate channel
    deactivate channel


    pool->>+pool: 9. HttpOrHttp2ChannelPool.release0(channel, promise)
    pool->>+channel: 10. doInEventLoop(() -> MultiplexedChannelRecord.closeAndReleaseChild)
    channel->>+channel: 11. childChannels.remove
    deactivate channel
    channel-->>-pool:
    deactivate pool
```

1. The AWS service returns the final part of the response.

2. ResponseHandler.finalizeResponse is invoked to complete processing of the response.

3. The client sends an RST_STREAM frame to acknowledge the completion.

4. It then calls release on the channel pool. Since the channel pool consists of multiple layers, this invocation eventually reaches HttpOrHttp2ChannelPool.release.

5. HttpOrHttp2ChannelPool.release needs to access the protocolImpl field, which is only safely accessed from the pool’s event loop. To ensure thread safety, it submits a task to perform the release within that event loop.

6. Meanwhile, the server receives the reset frame and decides to immediately close the connection (No idea why. It's a question for the service developer).

7. This triggers channelInactive on the Http2ConnectionHandler.

8. As a result, MultiplexedChannelRecord.closeAndExecuteOnChildChannels is called. It detects that there are still unreleased child channels because the task submitted in step 5 has not yet been executed. This leads to a ClosedChannelException being thrown and logged as an error.

9. The release task is now running on the pool's event loop, but a little bit too late.

It may also be the root cause of https://github.com/aws/aws-sdk-java-v2/issues/2914
 
## Modifications

The protocolImpl, assigned within the HttpOrHttp2ConnectionPool's event loop, can be safely used in the channel’s event loop as long as it’s visible. Since it is assigned only once, there's no risk of calling release on an incorrect instance.

If protocolImpl is either not yet assigned or not visible to the channel event loop, we fall back to the safe path. The visibility issue is rare—typically, the time between acquiring and releasing an HTTP/2 stream channel exceeds the memory synchronization delay between threads. However, it can still occur in practice. Therefore, this PR serves more as a proof of concept to validate that analysis rather than a final fix.

A potentially better solution would be to wrap the HTTP/2 stream channel at acquisition time and store a reference to the underlying pool within that wrapper, using it to perform the release instead of relying on protocolImpl in HttpOrHttp2ConnectionPool. Unfortunately, I don’t currently have the time to implement that.

### DIFF
--- a/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/HttpOrHttp2ChannelPool.java
+++ b/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/http2/HttpOrHttp2ChannelPool.java
@@ -194,9 +194,12 @@ public class HttpOrHttp2ChannelPool implements SdkChannelPool {
 
     @Override
     public Future<Void> release(Channel channel, Promise<Void> promise) {
-        doInEventLoop(eventLoop,
-            () -> release0(channel, promise),
-                      promise);
+        // If protocolImpl != null, it’s already visible, so we call protocolImpl.release
+        // directly in the channel event loop. Otherwise — whether unassigned or not yet
+        // visible — we fall back to the safe path. Since protocolImpl is assigned only
+        // once, there's no risk of calling it on an incorrect instance.
+        if (protocolImpl != null) protocolImpl.release(channel, promise);
+        else doInEventLoop(eventLoop, () -> release0(channel, promise), promise);
         return promise;
     }
 


dagnir:
Hi @huajiang-tubi! Thank you for the PR, and apologies for the delayed response.

I am a bit confused by the description of the issue though. In your timing diagram, 3 has the client sending a [`rst_stream`](https://httpwg.org/specs/rfc7540.html#RST_STREAM) to the service. However, this only happens from the client side if the `Subscriber` cancels the `Subscription` https://github.com/aws/aws-sdk-java-v2/blob/9ad92603df921b32a581914bb3ce4a0cbd5ec2b5/http-clients/netty-nio-client/src/main/java/software/amazon/awssdk/http/nio/netty/internal/ResponseHandler.java#L294-L297, which signals that the client does not want to read the rest of the response content; this does not normally happen. Likewise, the service doesn't normally send a `rst_stream` if the request handled successfully (which seems like it is in your timing diagram given by the `LastHttpResponse`).

Are your perhaps describing two different scenarios? i.e. either the server or the client sends the `rst_stream`?

PR #6261, opened 2025-07-15
Fix service version info to only register major and minor versions

RanVaknin:
Follow up to #6212 
Changes `ServiceVersionInfo` to be generated with a `major.minor.x` format instead of `major.minor.patch` this is to avoid updating this file for every service for every release. 

Lowering the update cadence is done to avoid rate limiting in our release infrastructure. This comes at the cost of having less fine-grained useragent info.

### DIFF
--- a/codegen/src/test/java/software/amazon/awssdk/codegen/utils/VersionUtilsTest.java
+++ b/codegen/src/test/java/software/amazon/awssdk/codegen/utils/VersionUtilsTest.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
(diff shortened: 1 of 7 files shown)


codegen/src/test/java/software/amazon/awssdk/codegen/utils/VersionUtilsTest.java
```diff
@@ -0,0 +1,34 @@
...
+package software.amazon.awssdk.codegen.utils;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.junit.Test;
+
+public class VersionUtilsTest {
+
+    @Test
+    public void serviceVersionInfo_redactPatchVersion() {
```
alextwoods: probably worth adding a test case for versiosn that don't match the regex + a non 2.x version

alextwoods (approved):

zoewangg (approved):

PR #6262, opened 2025-07-15
Add ChunkedEncodedPublisher

dagnir:
## Motivation and Context
This publisher supports encoding binary streams using chunked encoding. This publisher fills the same purpose that ChunkedEncodedInputStream does, but for async bodies.

Note: currently unused anywhere. Will be part of supporting streaming signing.

## Testing
 - New unit tests
 - TCK test

## Types of changes

## Checklist

## License

### DIFF
--- a/core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/signer/chunkedencoding/ChunkedEncodedPublisher.java
+++ b/core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/signer/chunkedencoding/ChunkedEncodedPublisher.java
@@ -1,0 +1,250 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.http.auth.aws.internal.signer.chunkedencoding;
+
+import java.nio.ByteBuffer;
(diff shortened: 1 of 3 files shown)


core/http-auth-aws/src/main/java/software/amazon/awssdk/http/auth/aws/internal/signer/chunkedencoding/ChunkedEncodedPublisher.java
```diff
@@ -0,0 +1,250 @@
...
+                for (int i = 0; i < nBufferedChunks; i++) {
+                    ByteBuffer slice = byteBuffer.slice();
+                    int maxBytesToCopy = Math.min(chunkBuffer.remaining(), slice.remaining());
+                    slice.limit(maxBytesToCopy);
+
+                    chunkBuffer.put(slice);
+                    if (!chunkBuffer.hasRemaining()) {
+                        chunkBuffer.flip();
+                        chunks.add(chunkBuffer);
+                        chunkBuffer = ByteBuffer.allocate(chunkSize);
```
L-Applin, line 203: Do we need synchronization around this? What if another thread is putting bytes in the `chunkBuffer` while re-allocating it?

dagnir: No synchronization needed, since the spec requires publishers to externally synchronize signals to a subscriber; i.e. only one thread should be calling `onNext` at a time

L-Applin: I see, thanks, it's been a while with pub/sub I'm just getting back into it. 

On another note, would be possible to reset and reuse the same buffer here? Instead of allocating a new one for each chunk?

dagnir: No, that's not possible. `chunkBuffer` represents the current chunk that we're building, by splitting/combining the buffers given to onNext. Once it's full it gets added to the list of chunks that we send to the downstream subscriber. There's no requirement that the subscriber has to consume the buffer before returning from `subscriber.onNext(chunks)`. We have to allocate a new buffer for each chunk

L-Applin (approved):

Issue #505, opened 2018-05-25
Methods and parameters should be annotated for nullability


xenomachina:
If methods and parameters were annotated for nullability (ie: either `@Nullable` or `@NotNull`) it would help Java programmers who use static analysis tools, and would be a huge help to Kotlin programmers using the Java SDK.

Nullability annotations in Java unfortunately suffer from a bit of a [too many standards](https://xkcd.com/927/) problem, but JetBrains has [a list of nullability annotations supported by the Kotlin compiler as well as other advice that should help with migrating to using nullability annotations](https://kotlinlang.org/docs/reference/java-interop.html#nullability-annotations).

dagnir:
Related: #47

0xTadash1:
The "too many standards" issue has largely been resolved. The Java ecosystem is converging on JSpecify (v1.0.0 released), which is now supported by Spring Framework, Guava, JUnit, IntelliJ, Kotlin, and others.

Is there any possibility of adopting JSpecify (`@NullMarked` / `@Nullable`) in the AWS SDK?

Currently, without these annotations, the SDK's return values have unspecified nullness. NullAway, a major JSpecify static analysis tool, optimistically does not enforce null checks on unspecified types (similar to platform types in Kotlin). This causes false negatives during static analysis and hides potential NPEs.

Applying `@NullMarked` to packages or classes makes everything non-null by default, so you only need to add `@Nullable` to APIs that actually return null. This reduces the burden of adding annotations compared to older approaches.

lbkulinski:
Hello. Following up to see if there is any further discussion about this. In IntelliJ I get a warning ([which I reported](https://youtrack.jetbrains.com/issue/IDEA-387330/NullMarked-class-gives-warning-Optional.ofNullable-with-non-null-argument-should-be-replaced-with-Optional.of)) about `getItem`'s return value being non-nullable because my class is `@NullMarked`. Thus in my example below, `Optional.ofNullable()` isn't needed when in reality it is.

Ideally the classes in the SDK would be marked for nullability using [JSpecify](https://jspecify.dev/)'s annotations. Thoughts? 

Example code:
```
@Repository
@NullMarked
public final class ChampionRepository {
    private final DynamoDbTable<Champion> champions;

    @Autowired
    public ChampionRepository(
        DynamoDbEnhancedClient dynamoDbClient,
        @Value("${app.aws.dynamodb.tables.champions}") String tableName
    ) {
[...]

PR #6280, opened 2025-07-18
Update guidelins and add kiro steering docs

zoewangg:
## Motivation and Context
Add kiro steering docs. I added existing guidelines from docs folder and internal docs. I also added additional guidelines that came to my mind.


.kiro/steering/async-programming-guidelines.md
```diff
@@ -0,0 +1,133 @@
...
+- **Avoid blocking operations**: Never use `get()` or `join()` inside a CompletableFuture chain as it defeats the purpose of asynchronous execution
+- **Handle exceptions properly**: Always include exception handling with `exceptionally()` or `handle()` methods
+```java
+   CompletableFuture.supplyAsync(() -> fetchData())
+       .exceptionally(ex -> {
+           logger.error("Error processing data", ex);
+           return fallbackData();
+       }, executor);
+```
+- **Use appropriate completion methods**:
```
L-Applin: what about `whenComplete`? We use it a lot

.kiro/steering/aws-sdk-java-v2-general.md
```diff
@@ -0,0 +1,162 @@
...
+- Maintain backward compatibility when modifying existing APIs
+
+## Code Style Standards
+
+- Follow Java coding conventions and the existing style in the codebase
+- Use meaningful variable, method, and class names that clearly indicate purpose
+- Include comprehensive Javadoc for public APIs
+- Keep methods short and focused on a single responsibility
+- Limit the number of method parameters (ideally 3 or fewer)
+- Use appropriate access modifiers (private, protected, public)
```
L-Applin: will Kiro know what is "appropriate" or should we elaborate?

.kiro/steering/aws-sdk-java-v2-general.md
```diff
@@ -0,0 +1,162 @@
...
+  
+#### Classes that instantiate other classes
+
+- If the class's primary purpose is to return instances of another class:
+  - If the "get" method has no parameters:
+    - If the class implements `Supplier`: `{Noun}Supplier` (e.g. `CachedSupplier`)
+    - If the class does not implement `Supplier`: `{Noun}Provider` (e.g. `AwsCredentialsProvider`)
+  - If the "get" method has parameters: `{Noun}Factory` (e.g. `AwsJsonProtocolFactory`)
+
+#### Service-specific classes
```
L-Applin: Does this section seems a bit restrictive? We don't want any other options?

.kiro/steering/aws-sdk-java-v2-general.md
```diff
@@ -0,0 +1,162 @@
...
+      public String toString() {
+          return ToString.builder("YourClassName")
+                         .add("fieldName1", fieldValue1)
+                         .add("fieldName2", fieldValue2)
+                         .build();
+      }
+      ```
+  - `equals()`: Compare all fields for equality, including proper null handling
+  - `hashCode()`: Include all fields in the hash code calculation
+- Consider using IDE-generated implementations
```
L-Applin: SDK `ToString` utils for `toString()`?

.kiro/steering/aws-sdk-java-v2-general.md
```diff
@@ -0,0 +1,162 @@
...
+- `defaultXXX()` when returning an instance with default settings (e.g., `BackoffStrategy.defaultStrategy()`)
+
+## Use of Optional
+
+- `Optional` **MUST NOT** be used when the result will never be null
+- For return types:
+  - `Optional` **SHOULD** be used when it is not obvious to a caller whether a result will be null
+  - `Optional` **MUST NOT** be used for "getters" in generated service model classes
+- For member variables: `Optional` **SHOULD NOT** be used
+- For method parameters: `Optional` **MUST NOT** be used
```
alextwoods: Should we provide guidance to use @Nullable *instead* or just not doing it?

.kiro/steering/aws-sdk-java-v2-general.md
```diff
@@ -0,0 +1,162 @@
...
+
+## Object Methods (toString, equals, hashCode)
+
+- All public POJO classes **MUST** implement `toString()`, `equals()`, and `hashCode()` methods
+- When adding new fields to existing POJO classes, these methods **MUST** be updated to include the new fields
+- Implementation guidelines:
+  - `toString()`: Include class name and all fields with their values
```
alextwoods: Possible additional guideline - do not include fields that could be considered sensitive such as credentials

.kiro/steering/aws-sdk-java-v2-general.md
alextwoods: Is this guidance useful for kiro? I think we often use the intelliJ auto-generated implementations, but not sure that will apply well here.  I think the guidance maybe should be either: use `java.util.Objects` helpers or write explicit code.

RanVaknin:
Should we add a mixed version compatibility guidance? 


```
## Version Compatibility

- **Default Behavior Substitution**: Never replace a conditional code path with an unconditional one without thorough compatibility testing. When a feature flag is removed and a previously optional code path becomes the only path, this constitutes a high-risk change that can break compatibility with older service modules.

- **Method Invocation Pattern Changes**: Be extremely cautious when changing which methods core components call on service interfaces. Even if the interface method exists in older versions, the implementation might throw UnsupportedOperationException or behave differently than expected by newer core components.

- **Interface Default Methods**: When adding methods to shared interfaces that will be called by core components:
  - Implement graceful fallback behavior in core when methods might not be properly implemented when possible.
  - Test with older service versions to verify runtime behavior, not just compile-time compatibility

- **Behavioral Assumptions**: Explicitly document and test assumptions about how service modules implement interface methods. Core components should not make new assumptions about service module behavior without ensuring backward compatibility.

- **Risk Assessment for Code Cleanup**: What appears to be simple "cleanup" or "optimization" can have significant compatibility implications. Always evaluate the runtime behavior changes, not just the code structure changes.
```

.kiro/steering/async-programming-guidelines.md
```diff
@@ -0,0 +1,133 @@
+---
+title: Asynchronous Programming Guidelines for AWS SDK v2
```
dagnir, line 2: Doesn't have to be in this PR but I think some guidance around locking and shared resources would be good too. For example w.r.t  issues we had around making sure `SdkUri`/`BoundedCache` are performant

.kiro/steering/async-programming-guidelines.md
```diff
@@ -0,0 +1,133 @@
...
+
+```java
+    // Avoid when possible - blocks the current thread
+    String result = future.get();
+
+    // Better - use callbacks
+    future.thenAccept(result -> processResult(result));
+```
+- **Add stacktrace to exceptions**: When using `CompletableFuture#join`, use `CompletableFutureUtils#joinLikeSync` to preserve stacktraces
+- **Don't ignore results**: Never ignore the result of a new `CompletionStage` if the parent stage can fail (unless you're absolutely sure it's safe to)
```
dagnir: Might be good to add an example for this one, it might not be immediately apparent. e.g.

```java
        CountDownLatch latch = new CountDownLatch(1);

        CompletableFuture<URL> future = new CompletableFuture<>();
        
        future.thenRun(url -> {
            downloadUrl(url);
            latch.countDown();
        });
        
        latch.await();
```

.kiro/steering/testing-guidelines.md
```diff
@@ -0,0 +1,186 @@
...
+## Mocking Guidelines
+
+- Mock external dependencies, not the class under test
+- Only mock what you need to control for the test
+- Prefer constructor injection to make dependencies mockable
+- Use Mockito's verify() to ensure interactions with mocks are as expected
+- Avoid excessive stubbing - it can make tests brittle and hard to maintain
+- [Wiremock](https://wiremock.org/) is commonly used to mock server responses in functional tests
+
+## Testing Asynchronous Code
```
dagnir: Another useful method is sometimes we run hundreds/thousands of iterations of a test to ensure that none of the iterations fail, which is helpful for testing race conditions

davidh44 (approved):

.kiro/steering/aws-sdk-java-v2-general.md
zoewangg: This is actually from the existing guideline

.kiro/steering/aws-sdk-java-v2-general.md
zoewangg: There is actually a feature request for https://github.com/aws/aws-sdk-java-v2/issues/505

.kiro/steering/aws-sdk-java-v2-general.md
zoewangg: Removed this for now, will follow up in a separate PR

.kiro/steering/async-programming-guidelines.md
zoewangg: Ack. We can follow up in a separate PR

.kiro/steering/testing-guidelines.md
zoewangg: Ack. We can follow up in a separate PR

PR #6282, opened 2025-07-20
Add validations for upload in s3 mulitpart client

zoewangg:
### Motivation and Context
The S3 multipart upload client lacked proper validation of upload parameters, which could lead to runtime failures or incorrect behavior when publishers provide malformed data. This addresses potential issues where:
• AsyncRequestBody parts lack content length information
• Individual parts exceed the configured part size
• The actual number of parts doesn't match expected calculations
• Publisher implementations continue sending data after completion

#### Validation Enhancements
• **Content Length Validation**: Ensure all AsyncRequestBody parts have content length present
• **Part Size Validation**: Verify each part (except the last) matches the expected part size exactly
• **Last Part Size Validation**: Validate the final part size matches the calculated remainder or full part size
• **Part Count Validation**: Ensure the total number of parts received matches the expected count
• **Publisher State Guards**: Add isDone checks in onNext to prevent processing after completion

#### Code Structure Improvements
• **Enhanced MpuRequestContext**: Added expectedNumParts field to track anticipated part count
• **Improved Error Handling**: Better exception chaining in MultipartUploadHelper.failRequestsElegantly
• **Consistent Naming**: Renamed variables for clarity (contentLength → totalSize, partCount → expectedNumParts)

## Testing
Added tests 

## Types of changes

## Checklist

## License

### DIFF
--- a/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/multipart/KnownContentLengthAsyncRequestBodySubscriber.java
+++ b/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/multipart/KnownContentLengthAsyncRequestBodySubscriber.java
@@ -32,6 +32,7 @@ import org.reactivestreams.Subscription;
 import software.amazon.awssdk.annotations.SdkInternalApi;
 import software.amazon.awssdk.core.async.AsyncRequestBody;
 import software.amazon.awssdk.core.async.listener.PublisherListener;
+import software.amazon.awssdk.core.exception.SdkClientException;
 import software.amazon.awssdk.services.s3.model.CompleteMultipartUploadResponse;
 import software.amazon.awssdk.services.s3.model.CompletedPart;
 import software.amazon.awssdk.services.s3.model.PutObjectRequest;
@@ -54,10 +55,10 @@ public class KnownContentLengthAsyncRequestBodySubscriber implements Subscriber<
     private final AtomicBoolean failureActionInitiated = new AtomicBoolean(false);
     private final AtomicInteger partNumber = new AtomicInteger(1);
     private final MultipartUploadHelper multipartUploadHelper;
-    private final long contentLength;
+    private final long totalSize;
     private final long partSize;
-    private final int partCount;
-    private final int numExistingParts;
+    private final int expectedNumParts;
+    private final int existingNumParts;
     private final String uploadId;
     private final Collection<CompletableFuture<CompletedPart>> futures = new ConcurrentLinkedQueue<>();
     private final PutObjectRequest putObjectRequest;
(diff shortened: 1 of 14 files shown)


joviegas:
1. How are these validations handled with S3 Crt client ?
2. Do we have or should we write a test suite to test S3Multipart client and S3Crt Client to check if the behaviour is same for these validations

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/multipart/KnownContentLengthAsyncRequestBodySubscriber.java
```diff
@@ -145,21 +142,23 @@ public void onSubscribe(Subscription s) {
...
     @Override
     public void onNext(AsyncRequestBody asyncRequestBody) {
-        if (isPaused) {
+        if (isPaused || isDone) {
             return;
         }
 
-        if (existingParts.containsKey(partNumber.get())) {
-            partNumber.getAndIncrement();
+        int currentPartNum = partNumber.getAndIncrement();
```
joviegas, line 152: Can you please help me to understand why earlier we used to do contains on get, now we first increment and then do containsKey check.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/multipart/KnownContentLengthAsyncRequestBodySubscriber.java
```diff
@@ -179,6 +178,49 @@ public void onNext(AsyncRequestBody asyncRequestBody) {
         subscription.request(1);
     }
 
+    private void validatePart(AsyncRequestBody asyncRequestBody, int currentPartNum) {
```
joviegas: I see we already have JUnit tests for each class. However, I was wondering how we end up with the validation failures below from an external API perspective, such as when users pass invalid AsyncRequestBody or files get corrupted in transit. Is it possible to write end-to-end test cases for these scenarios so that we can test them with UnknownContentLength publisher or with S3CrtClient?

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/multipart/KnownContentLengthAsyncRequestBodySubscriber.java
zoewangg: I don't know why we did that earlier, but the reason I changed is to avoid another atomic integer get call (micro perf optimization)

zoewangg:
> How are these validations handled with S3 Crt client ?
> Do we have or should we write a test suite to test S3Multipart client and S3Crt Client to check if the behaviour is same for these validations

CRT team is working on adding the same validations on CRT side. I'm not sure if we want to add tests on our side for CRT-based S3 client since I'd expect tests to be added on their side.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/multipart/KnownContentLengthAsyncRequestBodySubscriber.java
zoewangg: Yup, good point, added

joviegas (approved):

PR #6289, opened 2025-07-22
Use CRT response file in low level CRT S3AsyncClient for getObject

alextwoods:
Use CRT response file in low level CRT S3AsyncClient for getObject.

## Motivation and Context
https://github.com/awslabs/aws-crt-java/pull/825 added support for setting the response file and allowing CRT to write directly to the file.  That can improve memory usage by avoiding passing the chunks into java where they are then written to the file.

Note that this change does *not* change the Transfrer manager behavior - TransferManager will still use the Java based in memory transform. 

## Modifications
This follows the implementation for requestPath (the inverse feature, allowing paths to be set for putObject, see: https://github.com/aws/aws-sdk-java-v2/pull/4379).

*  Implement getObject(*, Path) in the DefaultS3CrtAsyncClient.  Sets internal execution attributes for response file.
* S3CrtAsyncHttpClient - plumb through execution attributes

## Testing
New and existing s3 tests, including TM + regression tests. 

## Types of changes

## Checklist

## License


test/architecture-tests/src/test/java/software/amazon/awssdk/archtests/CodingConventionWithSuppressionTest.java
```diff
@@ -50,7 +50,9 @@ public class CodingConventionWithSuppressionTest {
         Arrays.asList(ArchUtils.classNameToPattern(EmfMetricLoggingPublisher.class),
                       ArchUtils.classNameToPattern(MetricEmfConverter.class),
                       ArchUtils.classNameToPattern(MakeHttpRequestStage.class),
-                      ArchUtils.classNameToPattern("software.amazon.awssdk.services.s3.internal.crt.S3CrtResponseHandlerAdapter")));
+                      ArchUtils.classNameToPattern("software.amazon.awssdk.services.s3.internal.crt.S3CrtResponseHandlerAdapter"),
+                      ArchUtils.classNameToPattern(
```
alextwoods, line 54: This file follows the logging pattern of the other response transformers, see for example: [FileAsyncResponseTransformer](https://github.com/aws/aws-sdk-java-v2/blob/master/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/async/FileAsyncResponseTransformer.java#L167)

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/DefaultS3CrtAsyncClient.java
```diff
@@ -396,6 +419,10 @@ public void afterMarshalling(Context.AfterMarshalling context,
                             executionAttributes.getAttribute(SdkInternalExecutionAttribute.REQUEST_CHECKSUM_CALCULATION))
                        .put(RESPONSE_CHECKSUM_VALIDATION,
                             executionAttributes.getAttribute(SdkInternalExecutionAttribute.RESPONSE_CHECKSUM_VALIDATION))
+                       .put(S3InternalSdkHttpExecutionAttribute.RESPONSE_FILE_PATH,
+                            executionAttributes.getAttribute(RESPONSE_FILE_PATH))
+                       .put(S3InternalSdkHttpExecutionAttribute.RESPONSE_FILE_OPTION,
+                            executionAttributes.getAttribute(RESPONSE_FILE_OPTION))
```
zoewangg, line 425: Duplicate?

zoewangg: Can customers configure this attribute? What's the default behavior? Is it the same as what we have now?

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/CrtResponseFileResponseTransformer.java
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    @Override
+    public void onResponse(ResponseT response) {
+        this.response = response;
+    }
+
+    @Override
+    public void onStream(SdkPublisher<ByteBuffer> publisher) {
+        publisher.subscribe(new OnCompleteSubscriber(cf, this::exceptionOccurred));
```
zoewangg, line 59: Will this method be invoked? If not, should we just invoke exceptionOccurred here with UnsupprotedoperationException?

alextwoods: onStream is invoked, but the `onNext` method of the subscriber should never be invoked.

alextwoods: I've updated the comments to hopefully make this more clear + cleaned up the subscriber implementation

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/DefaultS3CrtAsyncClient.java
alextwoods: I don't think these are duplicate anywhere?

This PR doesn't use the RESPONSE_FILE_OPTION - but the Transfer Manager pause/cancel I think needs to (well depending on how we eventually implement it...).

Customer's shouldn't be able to configure this attribute themselves (its all @sdk internal api), but our transfer manager code may need to.

This defaults to CREATE_OR_REPLACE - looking at the documentation on getObject, it suggests that the default behavior should be to fail if the file already exists, but my testing shows thats not true, the file is replaced...

zoewangg: >I don't think these are duplicate anywhere? 

Ah, I have bad eyes 👁️! 

> This defaults to CREATE_OR_REPLACE - looking at the documentation on getObject, it suggests that the default behavior should be to fail if the file already exists, but my testing shows thats not true, the file is replaced...

Gotcha, we should fix the documentation then...

alextwoods: Created a backlog item to track that.

services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/crt/CrtResponseFileResponseTransformer.java
```diff
@@ -0,0 +1,110 @@
...
+        }
+
+        @Override
+        public void onSubscribe(Subscription s) {
+            if (this.subscription != null) {
+                s.cancel();
+                return;
+            }
+            this.subscription = s;
+            // do not request data from the subscription since body is written directly to file
```
zoewangg, line 90: Should we still request data to be compliant to the reactive streams spec? Hopefully the publisher should be able to handle it?

zoewangg: On second thought, NVM

zoewangg (approved):

PR #6295, opened 2025-07-23
Add support for lambda traces via threadLocal

RanVaknin:
## Motivation and Context

This implementation uses SLF4J MDC (which uses ThreadLocal under the hood) to read the trace ID value stored in `AWS_LAMBDA_X_TraceId` that would be set by the Lambda runtime (this would be added later, probably [here](https://github.com/aws/aws-lambda-java-libs/blob/5a0f8ad9e9b8ff61ba626e71c4b5de62134e82e4/aws-lambda-java-runtime-interface-client/src/main/java/com/amazonaws/services/lambda/runtime/api/client/AWSLambda.java#L235)).

### DIFF
--- a/core/aws-core/src/main/java/software/amazon/awssdk/awscore/interceptor/TraceIdExecutionInterceptor.java
+++ b/core/aws-core/src/main/java/software/amazon/awssdk/awscore/interceptor/TraceIdExecutionInterceptor.java
@@ -20,6 +20,7 @@ import org.slf4j.MDC;
 import software.amazon.awssdk.annotations.SdkProtectedApi;
 import software.amazon.awssdk.awscore.internal.interceptor.TracingSystemSetting;
 import software.amazon.awssdk.core.interceptor.Context;
+import software.amazon.awssdk.core.interceptor.ExecutionAttribute;
 import software.amazon.awssdk.core.interceptor.ExecutionAttributes;
 import software.amazon.awssdk.core.interceptor.ExecutionInterceptor;
 import software.amazon.awssdk.http.SdkHttpRequest;
@@ -34,19 +35,27 @@ public class TraceIdExecutionInterceptor implements ExecutionInterceptor {
     private static final String TRACE_ID_HEADER = "X-Amzn-Trace-Id";
     private static final String LAMBDA_FUNCTION_NAME_ENVIRONMENT_VARIABLE = "AWS_LAMBDA_FUNCTION_NAME";
     private static final String CONCURRENT_TRACE_ID_KEY = "AWS_LAMBDA_X_TraceId";
+    protected static final ExecutionAttribute<String> CACHED_TRACE_ID = new ExecutionAttribute<>("CachedTraceId");
+
+    @Override
+    public void beforeExecution(Context.BeforeExecution context, ExecutionAttributes executionAttributes) {
+        String traceId = MDC.get(CONCURRENT_TRACE_ID_KEY);
+        if (traceId != null) {
+            executionAttributes.putAttribute(CACHED_TRACE_ID, traceId);
+        }
+    }
 
     @Override
     public SdkHttpRequest modifyHttpRequest(Context.ModifyHttpRequest context, ExecutionAttributes executionAttributes) {
         Optional<String> traceIdHeader = traceIdHeader(context);
         if (!traceIdHeader.isPresent()) {
-            Optional<String> lambdafunctionName = lambdaFunctionNameEnvironmentVariable();
-            Optional<String> traceId = traceId();
+            Optional<String> lambdaFunctionName = lambdaFunctionNameEnvironmentVariable();
+            Optional<String> traceId = traceId(executionAttributes);
 
-            if (lambdafunctionName.isPresent() && traceId.isPresent()) {
+            if (lambdaFunctionName.isPresent() && traceId.isPresent()) {
                 return context.httpRequest().copy(r -> r.putHeader(TRACE_ID_HEADER, traceId.get()));
             }
         }
-
         return context.httpRequest();
     }
 
@@ -54,9 +63,10 @@ public class TraceIdExecutionInterceptor implements ExecutionInterceptor {
         return context.httpRequest().firstMatchingHeader(TRACE_ID_HEADER);
     }
 
-    private Optional<String> traceId() {
-        if (TracingSystemSetting.AWS_LAMBDA_MAX_CONCURRENCY.getStringValue().isPresent()) {
-            return Optional.ofNullable(MDC.get(CONCURRENT_TRACE_ID_KEY));
+    private Optional<String> traceId(ExecutionAttributes executionAttributes) {
+        Optional<String> traceId = Optional.ofNullable(executionAttributes.getAttribute(CACHED_TRACE_ID));
+        if (traceId.isPresent()) {
+            return traceId;
         }
         return TracingSystemSetting._X_AMZN_TRACE_ID.getStringValue();
     }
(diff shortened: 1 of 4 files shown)


core/aws-core/src/main/java/software/amazon/awssdk/awscore/interceptor/TraceIdExecutionInterceptor.java
```diff
@@ -32,6 +33,7 @@
 public class TraceIdExecutionInterceptor implements ExecutionInterceptor {
     private static final String TRACE_ID_HEADER = "X-Amzn-Trace-Id";
     private static final String LAMBDA_FUNCTION_NAME_ENVIRONMENT_VARIABLE = "AWS_LAMBDA_FUNCTION_NAME";
+    private static final String CONCURRENT_TRACE_ID_KEY = "AWS_LAMBDA_X_TraceId";
```
millems: This capitalization is weird. Was this specific format chosen for a reason?

core/aws-core/src/main/java/software/amazon/awssdk/awscore/internal/interceptor/TracingSystemSetting.java
```diff
@@ -24,7 +24,9 @@
 @SdkInternalApi
 public enum TracingSystemSetting implements SystemSetting {
     // See: https://github.com/aws/aws-xray-sdk-java/issues/251
-    _X_AMZN_TRACE_ID("com.amazonaws.xray.traceHeader", null);
+    _X_AMZN_TRACE_ID("com.amazonaws.xray.traceHeader", null),
+    // Environment variable to detect Lambda multi concurrency mode ("elevator"). This value is set by the Lambda runtime.
+    AWS_LAMBDA_MAX_CONCURRENCY("aws.lambda.maxConcurrency", null);
```
millems: Do we need to have this separate system setting or can we just check if CONCURRENT_TRACE_ID_KEY is set. In other words, what does this system setting provide that the existence of the thread local value being set does not?

millems (commented):
Will async clients be addressed in a separate PR? This won't work for async retries (which are delegated to a retries thread pool), calls that happen within future chaining (which are delegated to a future completion executor thread pool), etc.

We'll want to find where calls might switch threads (i.e. anywhere we're using futures) and then preserve the thread local setting across that thread change.

core/aws-core/src/main/java/software/amazon/awssdk/awscore/interceptor/TraceIdExecutionInterceptor.java
RanVaknin: I don't have context about _why_ it was chosen, but according to the Lambda team ,this is the format they are going to use.

core/aws-core/src/main/java/software/amazon/awssdk/awscore/internal/interceptor/TracingSystemSetting.java
RanVaknin: That's a great call out. Since `AWS_LAMBDA_MAX_CONCURRENCY`'s value is irrelevant (only whether its set or not) I think like you said, we can just use existence of the thread local value being set or not to determine if we are in a concurrent env or not.

core/aws-core/src/main/java/software/amazon/awssdk/awscore/interceptor/TraceIdExecutionInterceptor.java
trivenay: Happy to follow AWS SDK standards - We can tweak our string accordingly

core/aws-core/src/main/java/software/amazon/awssdk/awscore/internal/interceptor/TracingSystemSetting.java
trivenay: A alternative would be to not use the `AWS_LAMBDA_MAX_CONCURRENCY` env var at all and try to get the value from MDC. If it's not set then fallback to reading it from System properties

RanVaknin: Fixed.

core/aws-core/src/main/java/software/amazon/awssdk/awscore/interceptor/TraceIdExecutionInterceptor.java
M-Elsaeed: Did we agree on a string? ex. `AWS_LAMBDA_X_TRACEID`?

millems: `AWS_LAMBDA_X_TRACEID` or `AWS_LAMBDA_X_TRACE_ID` sounds good to me!

core/aws-core/src/main/java/software/amazon/awssdk/awscore/interceptor/TraceIdExecutionInterceptor.java
```diff
@@ -32,27 +34,40 @@
...
+    private static final ExecutionAttribute<String> CACHED_TRACE_ID = new ExecutionAttribute<>("CachedTraceId");
```
millems: Probably just TRACE_ID? This is cached in the sense that anything stored in memory is cached. Probably not the right term here.

core/aws-core/src/main/java/software/amazon/awssdk/awscore/interceptor/TraceIdExecutionInterceptor.java
M-Elsaeed: Let's go with `AWS_LAMBDA_X_TRACE_ID`🤝🏻

core/aws-core/src/test/java/software/amazon/awssdk/awscore/interceptor/TraceIdExecutionInterceptorTest.java
```diff
@@ -133,5 +206,6 @@ private SdkHttpRequest modifyHttpRequest(Context.ModifyHttpRequest context) {
     private void resetRelevantEnvVars(EnvironmentVariableHelper env) {
         env.remove("AWS_LAMBDA_FUNCTION_NAME");
         env.remove("_X_AMZN_TRACE_ID");
+        env.remove("AWS_LAMBDA_MAX_CONCURRENCY");
```
millems: No longer used

test/codegen-generated-classes-test/src/test/java/software/amazon/awssdk/services/TraceIdTest.java
```diff
@@ -56,4 +61,40 @@ public void traceIdInterceptorIsEnabled() {
...
+                mockHttpClient.stubResponses(
+                    HttpExecuteResponse.builder()
+                                       .response(SdkHttpResponse.builder().statusCode(500).build())
+                                       .responseBody(AbortableInputStream.create(new StringInputStream("{}")))
+                                       .build(),
+                    HttpExecuteResponse.builder().response(SdkHttpResponse.builder().statusCode(200).build())
+                                       .responseBody(AbortableInputStream.create(new StringInputStream("{}")))
+                                       .build());
+
+                client.allTypes().join();
```
millems, line 95: We should also add a test sending a request in the result future:

```java
client.allTypes()
      .thenRun(() -> client.allTypes() /* Did this request have the right trace ID? */)
```

I suspect we'll need to plumb the thread local context to there as well (and clear it, so that reuses of that thread don't have it set).

core/aws-core/src/test/java/software/amazon/awssdk/awscore/interceptor/TraceIdExecutionInterceptorTest.java
RanVaknin: done

core/aws-core/src/main/java/software/amazon/awssdk/awscore/interceptor/TraceIdExecutionInterceptor.java
RanVaknin: done

core/aws-core/src/main/java/software/amazon/awssdk/awscore/interceptor/TraceIdExecutionInterceptor.java
RanVaknin: done

millems (approved):

RanVaknin:
flakey integ test is the only failure `S3TransferManagerDownloadPauseResumeIntegrationTest` Merging.

PR #6298, opened 2025-07-24
Replace httpclient.execute call with httpclient.executeOpen as mentioned in deprecation notes of httpclient.execute

joviegas:
## Motivation and Context

The Apache HttpClient 5 `execute()` method is deprecated with a warning to use `executeOpen()` instead when we need to keep the response object open after request execution. Our `Apache5HttpClient` implementation requires keeping the response open because we wrap it in an `AbortableInputStream` for proper resource management.

This change eliminates the deprecation warning and ensures we're using the recommended API for our use case.

## Modifications

1. **Apache5HttpClient.java**:
   - Replaced deprecated `httpClient.execute()` with `httpClient.executeOpen()`
   - Added `determineTarget()` method that uses `RoutingSupport.determineHost()` to extract the target host from the request
   - Added proper exception handling to convert `HttpException` to `ClientProtocolException`
   - Added necessary imports for `RoutingSupport` and `ClientProtocolException`

2. **Apache5SdkHttpClient.java**:
   - Added `executeOpen()` method to the delegate implementation to support the new API

3. **MetricReportingTest.java**:
   - Updated mock setup to use `executeOpen()` instead of deprecated `execute()`
   - Cleaned up unused imports
   - Minor formatting improvements

## Testing

- All existing unit tests pass without modification
- Updated `MetricReportingTest` to mock the new `executeOpen()` method
- Verified that the response handling and `AbortableInputStream` wrapping continue to work correctly
- No functional changes to the HTTP client behavior - this is purely an API migration

## License

### DIFF
--- a/http-clients/apache5-client/src/main/java/software/amazon/awssdk/http/apache5/Apache5HttpClient.java
+++ b/http-clients/apache5-client/src/main/java/software/amazon/awssdk/http/apache5/Apache5HttpClient.java
@@ -286,14 +286,16 @@ public final class Apache5HttpClient implements SdkHttpClient {
         }
     }
 
-    private static HttpHost determineTarget(final ClassicHttpRequest request) throws IOException {
+    /**
+     * Determines the target host from the request using Apache HttpClient's official routing support utility.
+     */
+    private static HttpHost determineTarget(ClassicHttpRequest request) throws IOException {
         try {
             return RoutingSupport.determineHost(request);
-        } catch (final HttpException ex) {
+        } catch (HttpException ex) {
             throw new ClientProtocolException(ex);
         }
     }
-
 
     private HttpUriRequestBase toApacheRequest(HttpExecuteRequest request) {
         return apacheHttpRequestFactory.create(request, requestConfig);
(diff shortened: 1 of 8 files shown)


http-clients/apache5-client/src/main/java/software/amazon/awssdk/http/apache5/Apache5HttpClient.java
```diff
@@ -275,14 +278,25 @@ private HttpExecuteResponse execute(HttpUriRequestBase apacheRequest, MetricColl
...
     }
 
+    /**
+     * Determines the target host from the request using Apache HttpClient's official routing support utility.
+     */
+    private static HttpHost determineTarget(ClassicHttpRequest request) throws IOException {
+        try {
+            return RoutingSupport.determineHost(request);
+        } catch (HttpException ex) {
+            throw new ClientProtocolException(ex);
```
Fred1155, line 296: Just wonder why do we need to convert to ClientProtocolException here?

Fred1155: Discussed offline, this is the same logic and preserves the existing behavior.

joviegas: Thanks Bole , as discussed pasting the link https://github.com/apache/httpcomponents-client/blob/97fd9319e1960a026bd0020237932be7f958b997/httpclient5/src/main/java/org/apache/hc/client5/http/impl/classic/CloseableHttpClient.java#L101-L107 where it did same for execute call internally

joviegas:
> We can also add an test to check the thrown exception since we extracted the logic outside but that's optional

I investigated adding test cases to verify the exception handling in determineTarget(), but it's not feasible to trigger this code path in practice. The SDK performs validation at the request builder level and in higher-level components, ensuring that only valid targets reach the low-level HTTP client. Any request that would cause RoutingSupport.determineHost() to throw an HttpException is already rejected by the SDK's validation layer before reaching the Apache client.

Issue #5876, opened 2025-02-10
Use Overridable `Ec2MetadataClient` for IMDS-backed Providers


commiterate:
### Describe the feature

Make IMDS-backed providers accept an `Ec2MetadataClient` instance which respects [IMDS client](https://docs.aws.amazon.com/sdkref/latest/guide/feature-imds-client.html) configurations.

### Use Case

Various IMDS-backed providers like the `InstanceProfileRegionProvider` and `InstanceProfileCredentialsProvider` make IMDS requests using an `HttpURLConnection` via the SDK protected (`@SdkProtectedApi`) `software.amazon.awssdk.regions.util.HttpResourcesUtils` ([def](https://github.com/aws/aws-sdk-java-v2/blob/fbb1f40459daa698587913af85566583ec3e7a6a/core/regions/src/main/java/software/amazon/awssdk/regions/util/HttpResourcesUtils.java)) class. This is instead of the public `Ec2MetadataClient` ([javadoc](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/imds/Ec2MetadataClient.html)).

In particular:

- `InstanceProfileRegionProvider`
  - Uses the internal `software.amazon.awssdk.regions.internal.util.EC2MetadataUtils` ([def](https://github.com/aws/aws-sdk-java-v2/blob/fbb1f40459daa698587913af85566583ec3e7a6a/core/regions/src/main/java/software/amazon/awssdk/regions/internal/util/EC2MetadataUtils.java)) class which uses the `HttpResourcesUtils` class.
- `InstanceProfileCredentialsProvider`
  - Uses the `HttpResourcesUtils` class directly.
  - __Hardcodes the IMDS token TTL to 21,600 seconds ([code](https://github.com/aws/aws-sdk-java-v2/blob/fbb1f40459daa698587913af85566583ec3e7a6a/core/auth/src/main/java/software/amazon/awssdk/auth/credentials/InstanceProfileCredentialsProvider.java#L75)).__

These may not respect certain IMDS client configurations (e.g. IMDS session token TTL) nor have features like IMDS session token caching + auto-refresh (some fetch a new token every time).

Switch to the `Ec2MetadataClient` to de-duplicate IMDS functionality.

### Proposed Solution

Add builders for all providers and have an `ec2MetataClient` function on the builder. For example:

```java
import software.amazon.awssdk.http.urlconnection.UrlConnectionHttpClient
[...]

PR #6301, opened 2025-07-28
Use Ec2MetadataClient in defaults mode

S-Saranya1:
AutoDefaultsModeDiscovery changes to use Ec2MetadataClient

## Motivation and Context

This change migrates IMDS-backed providers from using internal utilities to the public Ec2MetadataClient API.

- AutoDefaultsModeDiscovery currently uses the internal EC2MetadataUtils class via HttpResourcesUtils, which doesn't respect certain IMDS client configurations 
- Lacks features like IMDS session token caching and auto-refresh 
- It replaces the basic HttpURLConnection with a HTTP client that offers much better debugging capabilities
- Uses separate IMDS implementation instead of the public Ec2MetadataClient

This PR specifically addresses the Auto defaults mode discovery component, which is one of three planned migrations (along with InstanceProfileCredentialsProvider and InstanceProfileRegionProvider).

Github Feature Request: https://github.com/aws/aws-sdk-java-v2/issues/5876

## Modifications

- Migrated AutoDefaultsModeDiscovery class to use Ec2MetadataClient from EC2MetadataUtils.
- Created Ec2MetadataSharedClient, which creates a shared HTTP client that is shared among the providers.
- Added DefaultEc2MetadataClientWithFallback for IMDSv1 compatibility.
- Added imds module dependency to aws-core module's pom.xml

## Testing

- All existing tests pass - AutoDefaultsModeDiscoveryTest continues to validate original functionality
- Added "AutoDefaultsModeDiscoveryEc2MetadataClientTest" which has unit tests that test the migration from EC2MetadataUtils to Ec2MetadataClient.

## Types of changes

## Checklist

## License

### DIFF
--- a/core/aws-core/pom.xml
+++ b/core/aws-core/pom.xml
@@ -209,7 +209,7 @@
         <dependency>
             <groupId>software.amazon.awssdk</groupId>
             <artifactId>imds</artifactId>
-            <version>2.31.73-SNAPSHOT</version>
+            <version>${awsjavasdk.version}</version>
             <scope>compile</scope>
         </dependency>
     </dependencies>
--- a/core/aws-core/pom.xml
+++ b/core/aws-core/pom.xml
@@ -206,6 +206,12 @@
             <artifactId>rxjava</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>imds</artifactId>
+            <version>2.31.73-SNAPSHOT</version>
+            <scope>compile</scope>
+        </dependency>
     </dependencies>
     <build>
         <plugins>
--- a/core/aws-core/src/main/java/software/amazon/awssdk/awscore/internal/defaultsmode/AutoDefaultsModeDiscovery.java
+++ b/core/aws-core/src/main/java/software/amazon/awssdk/awscore/internal/defaultsmode/AutoDefaultsModeDiscovery.java
@@ -19,6 +19,7 @@ import java.util.Optional;
 import software.amazon.awssdk.annotations.SdkProtectedApi;
 import software.amazon.awssdk.awscore.defaultsmode.DefaultsMode;
 import software.amazon.awssdk.core.SdkSystemSetting;
+import software.amazon.awssdk.core.exception.SdkClientException;
 import software.amazon.awssdk.imds.Ec2MetadataClient;
 import software.amazon.awssdk.imds.Ec2MetadataRetryPolicy;
 import software.amazon.awssdk.imds.internal.Ec2MetadataSharedClient;
@@ -96,7 +97,7 @@ public class AutoDefaultsModeDiscovery {
 
             String ec2InstanceRegion = client.get(EC2_METADATA_REGION_PATH).asString();
             return Optional.ofNullable(ec2InstanceRegion);
-        } catch (Exception exception) {
+        } catch (SdkClientException e) {
             return Optional.empty();
         } finally {
             if (client != null) {
--- a/core/aws-core/src/main/java/software/amazon/awssdk/awscore/internal/defaultsmode/AutoDefaultsModeDiscovery.java
+++ b/core/aws-core/src/main/java/software/amazon/awssdk/awscore/internal/defaultsmode/AutoDefaultsModeDiscovery.java
@@ -88,15 +88,20 @@ public class AutoDefaultsModeDiscovery {
             return Optional.empty();
         }
 
+        Ec2MetadataClient client = null;
         try {
-            Ec2MetadataClient client = Ec2MetadataSharedClient.builder()
-                                                              .retryPolicy(Ec2MetadataRetryPolicy.none())
-                                                              .build();
+            client = Ec2MetadataSharedClient.builder()
+                                            .retryPolicy(Ec2MetadataRetryPolicy.none())
+                                            .build();
 
             String ec2InstanceRegion = client.get(EC2_METADATA_REGION_PATH).asString();
             return Optional.ofNullable(ec2InstanceRegion);
         } catch (Exception exception) {
             return Optional.empty();
+        } finally {
+            if (client != null) {
+                Ec2MetadataSharedClient.decrementAndClose();
+            }
         }
     }
 
--- a/core/aws-core/src/main/java/software/amazon/awssdk/awscore/internal/defaultsmode/AutoDefaultsModeDiscovery.java
+++ b/core/aws-core/src/main/java/software/amazon/awssdk/awscore/internal/defaultsmode/AutoDefaultsModeDiscovery.java
@@ -19,8 +19,10 @@ import java.util.Optional;
 import software.amazon.awssdk.annotations.SdkProtectedApi;
 import software.amazon.awssdk.awscore.defaultsmode.DefaultsMode;
 import software.amazon.awssdk.core.SdkSystemSetting;
+import software.amazon.awssdk.imds.Ec2MetadataClient;
+import software.amazon.awssdk.imds.Ec2MetadataRetryPolicy;
+import software.amazon.awssdk.imds.internal.Ec2MetadataSharedClient;
 import software.amazon.awssdk.regions.Region;
-import software.amazon.awssdk.regions.internal.util.EC2MetadataUtils;
 import software.amazon.awssdk.utils.JavaSystemSetting;
 import software.amazon.awssdk.utils.OptionalUtils;
 import software.amazon.awssdk.utils.SystemSetting;
@@ -81,9 +83,17 @@ public class AutoDefaultsModeDiscovery {
     }
 
     private static Optional<String> queryImdsV2() {
+
+        if (SdkSystemSetting.AWS_EC2_METADATA_DISABLED.getBooleanValueOrThrow()) {
+            return Optional.empty();
+        }
+
         try {
-            String ec2InstanceRegion = EC2MetadataUtils.fetchData(EC2_METADATA_REGION_PATH, false, 1);
-            // ec2InstanceRegion could be null
+            Ec2MetadataClient client = Ec2MetadataSharedClient.builder()
+                                                              .retryPolicy(Ec2MetadataRetryPolicy.none())
+                                                              .build();
+
+            String ec2InstanceRegion = client.get(EC2_METADATA_REGION_PATH).asString();
             return Optional.ofNullable(ec2InstanceRegion);
         } catch (Exception exception) {
             return Optional.empty();
--- a/core/aws-core/src/test/java/software/amazon/awssdk/awscore/internal/defaultsmode/AutoDefaultsModeDiscoveryEc2MetadataClientTest.java
+++ b/core/aws-core/src/test/java/software/amazon/awssdk/awscore/internal/defaultsmode/AutoDefaultsModeDiscoveryEc2MetadataClientTest.java
@@ -1,0 +1,249 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.awscore.internal.defaultsmode;
+
+import static com.github.tomakehurst.wiremock.client.WireMock.aResponse;
+import static com.github.tomakehurst.wiremock.client.WireMock.get;
+import static com.github.tomakehurst.wiremock.client.WireMock.put;
+import static com.github.tomakehurst.wiremock.client.WireMock.stubFor;
+import static com.github.tomakehurst.wiremock.client.WireMock.getRequestedFor;
+import static com.github.tomakehurst.wiremock.client.WireMock.putRequestedFor;
+import static com.github.tomakehurst.wiremock.client.WireMock.urlEqualTo;
+import static com.github.tomakehurst.wiremock.client.WireMock.verify;
+import static org.assertj.core.api.Assertions.assertThat;
+
+import com.github.tomakehurst.wiremock.junit.WireMockRule;
+import java.lang.reflect.Field;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import software.amazon.awssdk.awscore.defaultsmode.DefaultsMode;
+import software.amazon.awssdk.core.SdkSystemSetting;
+import software.amazon.awssdk.http.SdkHttpClient;
+import software.amazon.awssdk.imds.internal.Ec2MetadataSharedClient;
+import software.amazon.awssdk.regions.Region;
+import software.amazon.awssdk.testutils.EnvironmentVariableHelper;
+import software.amazon.awssdk.utils.Lazy;
+
+/**
+ * Tests specifically for AutoDefaultsModeDiscovery's migration to use Ec2MetadataClient.
+ * These tests verify that the migration from EC2MetadataUtils to Ec2MetadataClient works correctly.
+ */
+public class AutoDefaultsModeDiscoveryEc2MetadataClientTest {
+    private static final EnvironmentVariableHelper ENVIRONMENT_VARIABLE_HELPER = new EnvironmentVariableHelper();
+
+    @Rule
+    public WireMockRule wireMock = new WireMockRule(0);
+
+    @Before
+    public void setup() {
+        System.setProperty(SdkSystemSetting.AWS_EC2_METADATA_SERVICE_ENDPOINT.property(),
+                           "http://localhost:" + wireMock.port());
+        ENVIRONMENT_VARIABLE_HELPER.remove(SdkSystemSetting.AWS_EC2_METADATA_DISABLED.environmentVariable());
+    }
+
+    @After
+    public void cleanup() {
+        wireMock.resetAll();
+        ENVIRONMENT_VARIABLE_HELPER.reset();
+        System.clearProperty(SdkSystemSetting.AWS_EC2_METADATA_SERVICE_ENDPOINT.property());
+    }
+
+    @Test
+    public void autoDefaultsModeDiscovery_shouldUseSharedHttpClient() throws Exception {
+        // Stub successful IMDS responses
+        stubFor(put("/latest/api/token")
+                    .willReturn(aResponse().withStatus(200).withBody("test-token")));
+        stubFor(get("/latest/meta-data/placement/region")
+                    .willReturn(aResponse().withStatus(200).withBody("us-east-1")));
+
+        ENVIRONMENT_VARIABLE_HELPER.set("AWS_DEFAULT_REGION", "us-west-2");
+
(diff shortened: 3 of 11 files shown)


core/aws-core/pom.xml
```diff
@@ -206,6 +206,12 @@
             <artifactId>rxjava</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>imds</artifactId>
+            <version>2.31.73-SNAPSHOT</version>
```
L-Applin: remove version from dependency

core/aws-core/src/main/java/software/amazon/awssdk/awscore/internal/defaultsmode/AutoDefaultsModeDiscovery.java
```diff
@@ -81,9 +83,17 @@ private static DefaultsMode compareRegion(String region, Region clientRegion) {
...
+        }
+
         try {
-            String ec2InstanceRegion = EC2MetadataUtils.fetchData(EC2_METADATA_REGION_PATH, false, 1);
-            // ec2InstanceRegion could be null
+            Ec2MetadataClient client = Ec2MetadataSharedClient.builder()
+                                                              .retryPolicy(Ec2MetadataRetryPolicy.none())
+                                                              .build();
+
+            String ec2InstanceRegion = client.get(EC2_METADATA_REGION_PATH).asString();
```
L-Applin, line 98: Even if it doesn't really do anything, I think we might still want to  call `.close()` on the `Ec2MetadataClient` instance here after completing the fetch

core/imds/src/main/java/software/amazon/awssdk/imds/internal/BaseEc2MetadataClient.java
```diff
@@ -60,6 +60,10 @@ protected BaseEc2MetadataClient(DefaultEc2MetadataAsyncClient.Ec2MetadataAsyncBu
         this(builder.getRetryPolicy(), builder.getTokenTtl(), builder.getEndpoint(), builder.getEndpointMode());
     }
 
+    protected BaseEc2MetadataClient(DefaultEc2MetadataClientWithFallback.Ec2MetadataBuilder builder) {
```
L-Applin: Do we really need this constructor in the base class? Can't it only be in the `DefaultEc2MetadataClientWithFallback`?

core/aws-core/src/test/java/software/amazon/awssdk/awscore/internal/defaultsmode/AutoDefaultsModeDiscoveryEc2MetadataClientTest.java
L-Applin, line 1: Can we use Junit 5 (ie `org.junit.jupiter.api.Test`)?

core/imds/src/main/java/software/amazon/awssdk/imds/internal/DefaultEc2MetadataClientWithFallback.java
L-Applin, line 1: Can we add tests for the `DefaultEc2MetadataClientWithFallback` class standalone (not as part of Defaults Mode Discovery)? Specifically to test the fallback behaviour itself?

core/aws-core/src/test/java/software/amazon/awssdk/awscore/internal/defaultsmode/AutoDefaultsModeDiscoveryEc2MetadataClientTest.java
```diff
@@ -0,0 +1,247 @@
...
+
+        AutoDefaultsModeDiscovery discovery = new AutoDefaultsModeDiscovery();
+        DefaultsMode result = discovery.discover(Region.US_EAST_1);
+
+        // Should fall back to IMDSv1 and return IN_REGION
+        assertThat(result).isEqualTo(DefaultsMode.IN_REGION);
+
+        // Verify both token request and region request were made
+        verify(putRequestedFor(urlEqualTo("/latest/api/token")));
+        verify(getRequestedFor(urlEqualTo("/latest/meta-data/placement/region")));
```
L-Applin: Is there a way to verify for a v1 fallback request more specifically?

L-Applin (changes requested):

core/imds/src/main/java/software/amazon/awssdk/imds/internal/Ec2MetadataSharedClient.java
```diff
@@ -0,0 +1,84 @@
...
+import software.amazon.awssdk.imds.Ec2MetadataClient;
+import software.amazon.awssdk.imds.Ec2MetadataRetryPolicy;
+import software.amazon.awssdk.utils.AttributeMap;
+import software.amazon.awssdk.utils.Lazy;
+
+/**
+ * Creates Ec2MetadataClient instances using a shared HTTP client internally.
+ * This provides resource efficiency by sharing a single HTTP client across all IMDS-backed providers
+ */
+@SdkInternalApi
```
L-Applin: should be `@SdkProtectedApi` since it is accessed from another module.

core/aws-core/pom.xml
S-Saranya1: Updated it.

core/aws-core/src/main/java/software/amazon/awssdk/awscore/internal/defaultsmode/AutoDefaultsModeDiscovery.java
S-Saranya1: Added the logic to close the client.

core/aws-core/src/test/java/software/amazon/awssdk/awscore/internal/defaultsmode/AutoDefaultsModeDiscoveryEc2MetadataClientTest.java
S-Saranya1: Sure, updated the test file.

core/imds/src/main/java/software/amazon/awssdk/imds/internal/DefaultEc2MetadataClientWithFallback.java
S-Saranya1: Added DefaultEc2MetadataClientWithFallbackTest to test the fallback behaviour.

core/aws-core/src/test/java/software/amazon/awssdk/awscore/internal/defaultsmode/AutoDefaultsModeDiscoveryEc2MetadataClientTest.java
S-Saranya1: Yeah, modified the test to see if the request is not including token header which is a v1 fallback.

core/imds/src/main/java/software/amazon/awssdk/imds/internal/Ec2MetadataSharedClient.java
S-Saranya1: Updated.

core/aws-core/src/main/java/software/amazon/awssdk/awscore/internal/defaultsmode/AutoDefaultsModeDiscovery.java
```diff
@@ -81,12 +84,25 @@ private static DefaultsMode compareRegion(String region, Region clientRegion) {
...
-            String ec2InstanceRegion = EC2MetadataUtils.fetchData(EC2_METADATA_REGION_PATH, false, 1);
-            // ec2InstanceRegion could be null
+            client = Ec2MetadataSharedClient.builder()
+                                            .retryPolicy(Ec2MetadataRetryPolicy.none())
+                                            .build();
+
+            String ec2InstanceRegion = client.get(EC2_METADATA_REGION_PATH).asString();
             return Optional.ofNullable(ec2InstanceRegion);
-        } catch (Exception exception) {
+        } catch (SdkClientException e) {
```
millems, line 100: What was this change for?

core/aws-core/src/test/java/software/amazon/awssdk/awscore/internal/defaultsmode/AutoDefaultsModeDiscoveryEc2MetadataClientTest.java
```diff
@@ -0,0 +1,271 @@
...
+    @RegisterExtension
+    static WireMockExtension wireMock = WireMockExtension.newInstance()
+                                                               .options(wireMockConfig().dynamicPort().dynamicPort())
+                                                               .configureStaticDsl(true)
+                                                               .build();
+
+    @BeforeAll
+    static void setupClass() {
+        System.setProperty(SdkSystemSetting.AWS_EC2_METADATA_SERVICE_ENDPOINT.property(),
+                           "http://localhost:" + wireMock.getPort());
```
millems, line 58: Should we reset this as well after the class?

core/imds/src/main/java/software/amazon/awssdk/imds/internal/DefaultEc2MetadataClientWithFallback.java
```diff
@@ -0,0 +1,403 @@
...
+
+        String value = uncheckedInputStreamToUtf8(abortableInputStream);
+        return new Token(value, ttlDuration);
+    }
+
+    protected static final class Ec2MetadataBuilder implements Ec2MetadataClient.Builder {
+
+        private Ec2MetadataRetryPolicy retryPolicy;
+        private java.net.URI endpoint;
+        private java.time.Duration tokenTtl;
```
millems: Why not import these?

core/imds/src/main/java/software/amazon/awssdk/imds/internal/Ec2MetadataSharedClient.java
```diff
@@ -0,0 +1,102 @@
...
+        
+        private Builder() {
+        }
+        
+        public synchronized Builder retryPolicy(Ec2MetadataRetryPolicy retryPolicy) {
+            this.retryPolicy = retryPolicy;
+            return this;
+        }
+        
+        public synchronized Ec2MetadataClient build() {
```
millems: This synchronized uses a different lock from the containing class's synchronized. Maybe we should just manually create a reentrant lock and share that?

core/imds/src/main/java/software/amazon/awssdk/imds/internal/Ec2MetadataSharedClient.java
```diff
@@ -0,0 +1,102 @@
...
+import software.amazon.awssdk.imds.Ec2MetadataClient;
+import software.amazon.awssdk.imds.Ec2MetadataRetryPolicy;
+import software.amazon.awssdk.utils.AttributeMap;
+
+/**
+ * Creates Ec2MetadataClient instances using a shared HTTP client internally.
+ * This provides resource efficiency by sharing a single HTTP client across all IMDS-backed providers
+ */
+@SdkProtectedApi
+public final class Ec2MetadataSharedClient {
```
millems, line 34: Instead of having a separate protected class and API for this, should it be a feature of Ec2MetadataClient? E.g. if they do `Ec2MetadataClient.builder().build()`, we use a shared HTTP client instead of a new one? We'd still use reference counting to close the shared HTTP client when no metadata clients are using it anymore.

That would remove the need for another protected API we need to maintain/remember to use, and improve the creation time of multiple Ec2MetadataClients for customers.

core/imds/src/main/java/software/amazon/awssdk/imds/internal/RequestMarshaller.java
```diff
@@ -59,12 +59,14 @@ public SdkHttpFullRequest createTokenRequest(Duration tokenTtl) {
 
     public SdkHttpFullRequest createDataRequest(String path, String token, Duration tokenTtl) {
         URI resourcePath = URI.create(basePath + path);
-        return defaulttHttpBuilder()
+        SdkHttpFullRequest.Builder builder = defaulttHttpBuilder()
```
millems: nit: s/defaultt/default/

core/imds/src/main/java/software/amazon/awssdk/imds/internal/DefaultEc2MetadataClientWithFallback.java
```diff
@@ -0,0 +1,403 @@
...
+import software.amazon.awssdk.utils.OptionalUtils;
+import software.amazon.awssdk.utils.ToString;
+import software.amazon.awssdk.utils.Validate;
+import software.amazon.awssdk.utils.cache.CachedSupplier;
+import software.amazon.awssdk.utils.cache.RefreshResult;
+
+/**
+ * An Implementation of the Ec2Metadata Interface with IMDSv1 fallback support.
+ * This client is identical to DefaultEc2MetadataClient but provides automatic fallback
+ * to IMDSv1 when IMDSv2 token retrieval fails (except for 400 errors).
```
millems, line 58: Should this just be a feature of `Ec2MetadataClient` as well? `Ec2MetadataClient.builder().v1FallbackAllowed(true).build()`

It would remove the duplicated code between the two implementations, make the usage of v1 something customers can choose to use in their own Ec2MetadataClients, and reduce the number of APIs we need to manage/remember to use.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/loader/DefaultSdkHttpClientBuilder.java
```diff
@@ -24,7 +24,7 @@
 /**
  * Utility to load the default HTTP client factory and create an instance of {@link SdkHttpClient}.
  */
-@SdkInternalApi
+@SdkProtectedApi
 public final class DefaultSdkHttpClientBuilder implements SdkHttpClient.Builder {
```
millems, line 31: Darn, I see we already released something that uses this, so this is just backfilling it to mark it as protected. If we needed it to be protected, we should have moved it out of the internal package or ideally not have made it protected. Too late for that.

core/imds/src/main/java/software/amazon/awssdk/imds/internal/Ec2MetadataSharedClient.java
S-Saranya1: If we make Ec2MetadataClient.builder().build() calls use a shared HTTP client, then client.close() would only decrement the reference count instead of actually closing resources, this would be a behavior change for users right? They expect their client to be fully closed but it's not closing. Alternatively, if we kept user-created clients separate from internal provider clients, we'd need to add a new public API feature to Ec2MetadataClient for providers to use shared clients, but that would be a public API change and users could also access.

core/imds/src/main/java/software/amazon/awssdk/imds/internal/DefaultEc2MetadataClientWithFallback.java
S-Saranya1: That would be a public API change, and we only want IMDSv1 fallback for backward compatibility in internal providers so didn't go for that approach.

core/aws-core/src/main/java/software/amazon/awssdk/awscore/internal/defaultsmode/AutoDefaultsModeDiscovery.java
S-Saranya1: To fix a SpotBugs warning that complained about catching Exception when no exception was thrown and client.get() actually throws SdkClientException (and its subclasses), so changed this to more specific exception.

core/aws-core/src/test/java/software/amazon/awssdk/awscore/internal/defaultsmode/AutoDefaultsModeDiscoveryEc2MetadataClientTest.java
S-Saranya1: Done.

core/imds/src/main/java/software/amazon/awssdk/imds/internal/DefaultEc2MetadataClientWithFallback.java
S-Saranya1: Updated it.

core/imds/src/main/java/software/amazon/awssdk/imds/internal/RequestMarshaller.java
S-Saranya1: Updated.

core/imds/src/main/java/software/amazon/awssdk/imds/internal/Ec2MetadataSharedClient.java
S-Saranya1: Yes, Thanks for this, updated to use reentrant lock.

core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/loader/DefaultSdkHttpClientBuilder.java
S-Saranya1: Yeah :(  
Added the comment about it.

core/imds/src/main/java/software/amazon/awssdk/imds/internal/BaseEc2MetadataClient.java
S-Saranya1: BaseEc2MetadataClient constructor is private, so can't access it from DefaultEc2MetadataClientWithFallback, we could make the constructor protected and then remove this builder-specific constructor from this class.

L-Applin: mmmh I see. Yeah, that's my fault, that's how I coded it in the past. Now that I am looking back at it, I think it was not ideal for  `DefaultEc2MetadataClient` and `DefaultEc2MetadataAsyncClient` to have their own constructor in the parent class. 

I think it would now be best to have the `BaseEc2MetadataClient` constructor be protected and call
`super(builder.getRetryPolicy(), builder.getTokenTtl(), builder.getEndpoint(), builder.getEndpointMode());` from the inherited class `DefaultEc2MetadataClientWithFallback`.

S-Saranya1: Updated it!

L-Applin (approved):
Matt still has some concerns, but I think they were all part of the design that we reviewed while he was on vacation. It's all compromises for not introducing the v1 fallback available as a public API on the IMDS, which is what we agreed on during design.

core/aws-core/src/main/java/software/amazon/awssdk/awscore/internal/defaultsmode/AutoDefaultsModeDiscovery.java
millems: Should we do `RuntimeException`, instead? This would catch the same exceptions as before, but not upset spotbugs. Unless we want a potential behavior change?

core/imds/src/main/java/software/amazon/awssdk/imds/internal/Ec2MetadataSharedClient.java
millems: If we changed the default to use a shared client, how would that behavior change negatively affect the customer?

core/imds/src/main/java/software/amazon/awssdk/imds/internal/DefaultEc2MetadataClientWithFallback.java
millems: If we can reduce the number of protected APIs and simplify the code, it might be worth having v1 fallback be a public API.

I'm concerned about the duplication and additional protected surface area the current approach creates. Can you think of other ways to reduce it, without making a public API for enabling v1 fallback?

